PR #33, opened 2018-06-14
Add possibility to pass reducer to Peregrine initiation

leon0v:


leon0v:
My bad accidentally PR'ed to the original repo in stead of the forked one.

PR #40, opened 2018-06-15
Add a MiniCart component

jimbo:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

Add a new component, `MiniCart`.

## Additional information

**This PR follows #39, so it is blocked by that PR. Do not merge until #39 is merged.**

It will also add a `Button` component, which is needed for icon buttons because `button` elements can't be grid containers ([Chrome/Safari bug](https://bugs.chromium.org/p/chromium/issues/detail?id=375693)).

### DIFF
--- a/packages/venia-concept/src/components/Button/index.js
+++ b/packages/venia-concept/src/components/Button/index.js
@@ -1,0 +1,31 @@
+import { Component, createElement } from 'react';
+import PropTypes from 'prop-types';
+
+import classify from 'src/classify';
+import defaultClasses from './button.css';
+
+class Button extends Component {
+    static propTypes = {
+        classes: PropTypes.shape({
+            content: PropTypes.string,
+            root: PropTypes.string
+        }),
+        type: PropTypes.oneOf(['button', 'reset', 'submit'])
+    };
+
+    static defaultProps = {
+        type: 'button'
+    };
+
+    render() {
+        const { children, classes, type, ...restProps } = this.props;
+
+        return (
+            <button className={classes.root} type={type} {...restProps}>
+                <span className={classes.content}>{children}</span>
+            </button>
+        );
+    }
+}
+
+export default classify(defaultClasses)(Button);
--- a/packages/venia-concept/src/components/Icon/icon.js
+++ b/packages/venia-concept/src/components/Icon/icon.js
@@ -13,8 +13,8 @@ class Icon extends Component {
     };
 
     render() {
-        const { classes, name } = this.props;
-        const svg = Feather.icons[name].toSvg();
+        const { attrs, classes, name } = this.props;
+        const svg = Feather.icons[name].toSvg(attrs);
         const fn = () => ({ __html: svg });
 
         return <span className={classes.root} dangerouslySetInnerHTML={fn()} />;
--- a/packages/venia-concept/src/components/MiniCart/miniCart.js
+++ b/packages/venia-concept/src/components/MiniCart/miniCart.js
@@ -1,0 +1,67 @@
+import { Component, createElement } from 'react';
+import PropTypes from 'prop-types';
+import { List } from '@magento/peregrine';
+
+import classify from 'src/classify';
+import Button from 'src/components/Button';
+import Icon from 'src/components/Icon';
+import toMap from 'src/util/toMap';
+import Product from './product';
+import Trigger from './trigger';
+import mockData from './mockData';
+import defaultClasses from './miniCart.css';
+
+class MiniCart extends Component {
+    static propTypes = {
+        classes: PropTypes.shape({
+            root: PropTypes.string
+        })
+    };
+
+    static defaultProps = {
+        // TODO: remove when connected to graphql
+        data: toMap(mockData, v => [v.id, v])
+    };
+
+    render() {
+        const { classes, data, isOpen } = this.props;
+        const className = isOpen ? classes.open : classes.closed;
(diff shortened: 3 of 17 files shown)


packages/venia-concept/src/components/Button/index.js
```diff
@@ -0,0 +1,31 @@
...
+    static propTypes = {
+        classes: PropTypes.shape({
+            content: PropTypes.string,
+            root: PropTypes.string
+        }),
+        type: PropTypes.oneOf(['button', 'reset', 'submit'])
+    };
+
+    static defaultProps = {
+        type: 'button'
```
DrewML: I don't feel super strongly about this. However, for devs or designers that are used to writing  plain HTML, the default `type` being `button` is going to be surprising, since the sets the default as `submit`

packages/venia-concept/src/components/Icon/icon.js
```diff
@@ -13,8 +13,8 @@ class Icon extends Component {
     };
 
     render() {
-        const { classes, name } = this.props;
-        const svg = Feather.icons[name].toSvg();
+        const { attrs, classes, name } = this.props;
+        const svg = Feather.icons[name].toSvg(attrs);
```
DrewML: Note: This `feather-icons` library is [64.67kb minified](https://unpkg.com/feather-icons@4.7.3/dist/), and forces every icon into our bundle, along with a bunch of duplicated Babel helpers. Gzipped the lib is still 15.7kb. We should look for an alternative

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -0,0 +1,50 @@
+import { Component, Fragment, createElement } from 'react';
+import PropTypes from 'prop-types';
+
+import classify from 'src/classify';
+import defaultClasses from './product.css';
+
+class Product extends Component {
+    static propTypes = {
+        classes: PropTypes.shape({
```
DrewML: This is missing some classes used in `render` and `get options`

packages/venia-concept/src/components/Button/index.js
brendanfalkowski: Same

packages/venia-concept/src/components/Icon/icon.js
brendanfalkowski: For what it's worth, never used an icon set on client project. At most they use 10 icons and still need another 10 arrows/misc figures we have to hand-draw. It's simpler to draw/pluck them without importing a massive icon set.

packages/venia-concept/src/components/Button/index.js
jimbo: As you've emphasized elsewhere, this is optimal for the most common use case. In Venia and most modern apps, most buttons should have a type of `button`. From MDN:

> The button has no default behavior. It can have client-side scripts associated with the element's events, which are triggered when the events occur.

The cases in which we're using a button to submit a form, without that submission being intercepted by client scripts, will be far more rare. (Ironically this may be one of them.) I think supplying `type="submit"` in exceptional cases is reasonable.

packages/venia-concept/src/components/MiniCart/product.js
jimbo: Added missing prop-types for this and others. 👍

packages/venia-concept/src/components/Button/index.js
```diff
@@ -0,0 +1,31 @@
...
+    };
+
+    render() {
+        const { children, classes, type, ...restProps } = this.props;
+
+        return (
+            <button className={classes.root} type={type} {...restProps}>
+                <span className={classes.content}>{children}</span>
```
DrewML: Is the wrapper `span` needed? Might make more sense to let the consumer decide all the children

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -0,0 +1,76 @@
...
+                    renderItem={Product}
+                    items={data}
+                    classes={productListClasses}
+                />
+                <div className={classes.summary}>
+                    <dl className={classes.totals}>
+                        <dt className={classes.subtotalLabel}>
+                            Subtotal (4 Items)
+                        </dt>
+                        <dd className={classes.subtotalValue}>$528.00</dd>
```
DrewML: Can we use the `<Price />` component from Peregrine here?

jimbo: Sure, I'll pull that in. 👍

packages/venia-concept/src/components/Button/index.js
jimbo: Yes, unfortunately. In fact, that `span` is the only reason `Button` is a component.

- `<button>` can't be a flex or grid container in some browsers
- `<button>` has a lot of semantic value, so we still want to use it
- `<button>` can only contain phrasing content, not flow content, so no `<div>`
- so we have to wrap content in `<span>`
-

DrewML: Ah cool, missed that. Thanks for clarifying 👍

jimbo:
> Just want to see that <Price /> component swap and then 👍

Done. I put the `Price` component in for the subtotal as well as the product prices, so I updated the mock data to match the API format as well.

PR #52, opened 2018-06-18
Super basic implementation of category page using apollo client for data fetching

### LINKED ISSUE #20: Add live GraphQL data via Apollo bindings to components [WIP]
_From @ericerway on May 11, 2018 18:52_

As a developer, I'd like to be able to specify my GraphQL queries on a per-component basis, so I do not need to think about orchestrating data fetching needs at an application level. 

**Acceptance Criteria**:
- The `venia-pwa-concept` theme is setup with react-apollo's provider that allows components in the tree to specify queries to be fetched, based on suggesting implementation path in [the `react-apollo` docs](https://www.apollographql.com/docs/react/)
- At least one component in the theme fetches live data through GraphQL, using Apollo Client's React bindings
- Basic documentation is provided specifying how to attach a query to a React component

_Copied from original issue: magento-research/venia-pwa-concept#69_

DrewML:
~~**Note** This branch surfaced an issue with the unit/integration test setup in the monorepo. Freezing work on this branch until I've completed https://github.com/magento-research/pwa-studio/issues/54.~~

Closes #20 

## This PR is a:

[ ] New feature
[X] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

Same code from https://github.com/magento-research/venia-pwa-concept/pull/89, now moved over and tests fixed.


packages/peregrine/package.json
```diff
@@ -13,7 +13,7 @@
     "url": "https://github.com/magento-research/peregrine"
   },
   "scripts": {
-    "build": "babel src -d dist --ignore test.js",
+    "build": "babel src -d dist --ignore '*.test.js,*.spec.js'",
```
DrewML: No need to build the test files for distributing. But I also caught this generating built test artifacts, so jest would run all tests twice. Random tech debt from monorepo move

packages/venia-concept/jest.config.js
```diff
@@ -4,5 +4,8 @@ module.exports = {
     moduleNameMapper: {
         '\\.css$': 'identity-obj-proxy',
         '^src/(.+)': '<rootDir>/src/$1'
-    }
+    },
+    // We don't ship Peregrine with CSJ support (intentional), so transpile
+    // used Peregrine code in tests
+    transformIgnorePatterns: ['node_modules/(?!@magento/peregrine)']
```
DrewML: Letting `jest` take care of transpiling `peregrine` code used in `venia` during tests, because we don't want to a second persistent process running to build artifacts to disk just for tests

packages/peregrine/package.json
jimbo: We were using `.spec.js` for most of initial development, IIRC. When did we start having `.test.js` files, and why? Should change them.

DrewML: > When did we start having .test.js files

[October 31, 2017](https://github.com/magento-research/peregrine/commit/531a1deb36c0f35a17dfe68688f9003a5f58d4b3). We can change it in a separate PR

packages/peregrine/src/List/__tests__/items.spec.js
```diff
@@ -6,11 +6,38 @@ import { Items } from '..';
...
-const items = Object.entries({
-    a: { id: 'a', val: '10' },
-    b: { id: 'b', val: '20' },
-    c: { id: 'c', val: '30' }
-});
+const items = [
+    [
+        1,
+        {
+            id: 1,
```
jimbo: The tests for `List` and its child components have been updated as of #62, using almost the same sample data you've got here. You can probably delete these test changes.

Also, I think it's better to have `id` be a `string` instead of a `number`. Is that something we can accomplish with our query?

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -1,51 +1,83 @@
...
+    }
+`;
 
 class Category extends Component {
     static propTypes = {
-        classes: PropTypes.shape({
-            gallery: PropTypes.string,
-            root: PropTypes.string,
-            title: PropTypes.string
+        id: number,
```
jimbo: Likewise, can `id` be a `string`?

packages/venia-concept/src/components/Gallery/item.js
```diff
@@ -18,29 +18,36 @@ const ItemPlaceholder = ({ children, classes }) => (
...
+            name: string.isRequired,
+            small_image: string.isRequired,
+            price: shape({
+                regularPrice: shape({
+                    amount: shape({
+                        value: number.isRequired,
+                        currency: string.isRequired
+                    }).isRequired
+                }).isRequired
+            }).isRequired
```
jimbo: If an item doesn't have a price, will this full shape be present anyway? Will it just be `$0.00` all the way down?

packages/peregrine/src/List/__tests__/items.spec.js
DrewML: You can't cast in a GraphQL query, so we'd have to cast on the client anytime that data came in. That would be a PITA, because any query that takes `id` as an arg would have to be cast back to a number.

packages/venia-concept/src/RootComponents/Category/category.js
DrewML: See https://github.com/magento-research/pwa-studio/pull/52#discussion_r197236280

packages/venia-concept/src/components/Gallery/item.js
DrewML: May just be something we have to find out. I'd expect it to always be present, but the schema doesn't have enough info to really tell me.

It's not marked as a non-nullable field in the schema, but I haven't seen that team make much use of non-nullable fields anywhere.

Pinging @eric-bohanon

jimbo:
@DrewML Ok, if you resolve conflicts (probably with the `List` tests) I can approve this.

DrewML:
@jimbo updated. Luckily conflict was easy

packages/venia-concept/src/components/Gallery/item.js
paliarush: We are currently not using non-nullable fields in GraphQL schema, but it makes sense to consider this as an improvement. @DrewML feel free to create an issue in magento/graphql-ce project listing any fields that you believe should be made non-nullable.

Issue #41, opened 2018-06-18
Make `List` component usage more natural with shape of APIs response payload


DrewML:
## This issue is for the following packages:

[ ] `venia-concept`
[ ] `pwa-buildpack`
[X] `peregrine`
[ ] `pwa-module`

## This issue is a:

[X] Bug
[X] Feature suggestion
[ ] Other (Please Specify)

### Environment

| Question                    |    Answer     |
| --------------------------- | :-----------: |
| Magento version             | n/a |
| Operating System + version  | n/a |
| node.js version (`node -v`) | n/a |
| npm version (`npm -v`)      | n/a |

## Description
The `List` component in Peregrine currently accepts an `items` prop, which is defined as:

```js
PropTypes.oneOfType([
   PropTypes.instanceOf(Map),
    PropTypes.arrayOf(PropTypes.array)
]).isRequired
```

The [call to `Array.from`](https://github.com/magento-research/pwa-studio/blob/434b895cec413258ea181590b007b508ecedf1cf/packages/peregrine/src/List/items.js#L61) masks an important error. If you pass in data straight from the API, which is in the shape:

```js
[{ id: 1 }, { id: 2 }]
```

you'll get an error about trying to destructure something that isn't an iterable. This is because `Array.from(exampleArrayAbove)` translates to a multi-dimensional array where each nested array has an Object as the value, but our code attempts to destructure `[key, item]` from the collection.

## Expected result:
I would expect, now that we know the shape of the data from the API, that we would want to make the expected shape in our components match 1:1.

## Possible solutions:
I'd like to switch to just accepting an array of objects for the sake of simplicity. We did not know the interface shape from GQL before creating this component, but we do now. We should optimize for the common case, and let people transform `Map` >> `Array` if they're doing some transformations from `Array` >> `Map` with the API response.

jimbo:
Thanks for opening the issue. I'll walk through the design first, and then we can look at alternatives.

**Keys**

Like the `Map` constructor, the `List` component expects an iterable whose elements are key-value pairs (arrays with two elements). Like a `Map`, the collection of components returned by a `List` must be both indexed (order may be important) and keyed (each item **must** be accompanied by a unique key).

If you have a collection of data that is not keyed, such as a one-dimensional array of objects, that collection can't be transformed into a keyed list of components unless you do one of two things:

1. Transform the collection into a keyed one before passing it in
2. Pass a transform function that can be used to generate keys

At present, `List` is designed for the former.

**Data**

[...]

PR #62, opened 2018-06-20
Revert List component to old API

jimbo:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

Change the `List` component's API, such that it only expects `items` to be an iterable, instead of a stricter, `Map`-compatible, two-dimensional iterable. This change will make it more expedient to pass collections of data straight from the API to `List` without having to transform them first.

## Additional information

Since we can't safely mandate that all collections consist of objects with an `id` property, `List` must also accept a key-generation function, `getItemKey`. This function will default to `({ id }) => id`, which plucks the `id` prop from a plain object, covering the most common case. For flat collections of strings, such as tags, users can supply an identity function; for more unusual collections where the primary key is something other than `id`, they can supply a custom function.

**This PR is marked [WIP] until this API change is approved. Once approved, instances of `List` in this repo have to be updated.**


packages/peregrine/src/validators/iterable.js
```diff
@@ -0,0 +1,25 @@
+const isIterable = obj => typeof obj[Symbol.iterator] === 'function';
+
+function optionalIterable(props, propName, componentName) {
+    const prop = props[propName];
+
+    if (prop != null && !isIterable(prop)) {
+        return new Error(
+            `Invalid prop \`${propName}\` supplied to \`${componentName}\`. Validation failed.`
```
DrewML: Can we make this more specific? Doesn't seem to call out that we're expecting an iterable

jimbo: @DrewML Updated. 👍

DrewML:
Thanks for doing this! Good that it now works seamlessly with our payloads, but can be useable by others with differently-shaped data.

jimbo:
No problem, it's just a different approach to the same solution.

Fixes #41.

PR #90, opened 2018-06-26
feat(data): Add live GraphQL data to product detail page

### LINKED ISSUE #87: Add live GraphQL data via Apollo bindings to Product Detail component

[x] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-module`
[ ] `pwa-devdocs`

## This issue is a:

[ ] Bug
[x] Feature suggestion
[ ] Documentation issue
[ ] Other (Please Specify)

### Environment

## Description

The Category page is now pulling live data from GraphQL, but the product detail page is still using mock data. Add a GraphQL query for product details, using the URL to resolve the product record.

## Expected result:

Navigating to a canonical product URL should produce a product detail page in Venia with live data.

zetlen:
Basic implementation of a GraphQL query for product details. Builds on #52 by replicating the inline query declaration. Had to plumb out the child components for the new data shape; in doing so, I made a few reusable functions.

Closes #87.

## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

 - Added GraphQL query to
 `packages/venia-concept/src/RootComponents/Product.js`.
     - Resolves from URL by using the `url_key` in a GraphQL query.
     - ~Can also resolve by SKU.~
     - Modified prop types and render method to accommodate live data shape.
 - ~Added `<Currency />` component whose signature matches the Magento GraphQl `Money` type.~
    - ~Uses the `window.Intl` standard object to format.~
 - Modified the `Gallery` and `ProductImageCarousel` components to use new data shape.
 - Moved shared constant data URIs to a single `src/shared` folder, to replicate placeholder logic.
 - ~Created a shared `propShapes.js` file containing commonly used prop type expressions.~
 - Anticipating that `url_key` would be a common way to navigate, I made a `url_key` utility function.
 - Added a `makeProductMediaPath` utility function, for turning product image file paths from API responses into relative URLs.
   - Though [magento/graphql-ce/issues/88](https://github.com/magento/graphql-ce/issues/88) is still a problem for production, I found that **when `magento-sample-data` is installed, it symlinks into the `pub/media` folder so you can use simpler URLs.**
   - You can see this for yourself with `ls -l <magento-root>/pub/media/catalog/product`.
   - So I added a `makePathPrepender` function, which we'll later use often, that can create functions like `makeProductMediaPath`.
   - I hardcoded `/media/catalog/products` in the code, but I also added an environment variable to `.env` and `webpack.config.js` for configuring that URL per instance.

### Todos
 - Optimize queries with fragments
 - Centralize queries in query file to be preprocessed
 - Make link to product detail on category page
 - Resolve media URL issue
 - Test with image galleries


packages/venia-concept/src/components/Currency/index.js
```diff
@@ -0,0 +1,38 @@
+import { Component, createElement } from 'react';
+import PropTypes from 'prop-types';
+
+export default class Currency extends Component {
```
DrewML: Were you aware that the [`Price` component](https://github.com/magento-research/pwa-studio/blob/8539d8752fbe2f24efec41d882efed82d06df959/packages/peregrine/src/Price/Price.js) exists? Not clear if you just missed it, if this replaces it, or if this is in addition to it?

packages/venia-concept/src/components/Currency/index.js
```diff
@@ -0,0 +1,38 @@
...
+        value: PropTypes.number.isRequired,
+        currency: PropTypes.string.isRequired,
+        tagName: PropTypes.string
+    };
+
+    static defaultProps = {
+        tagName: 'span'
+    };
+
+    guessLocalLanguage() {
```
DrewML: We don't need to guess here. The spec advises implementers to use the [environment's default](https://www.ecma-international.org/ecma-402/2.0/#sec-defaultlocale). We should trust the browser to do the right thing, unless we have evidence they don't.

Default is accomplished by passing in `undefined` for the first positional argument to `Intl.NumberFormat`. See implementation [here](https://github.com/magento-research/pwa-studio/blob/8539d8752fbe2f24efec41d882efed82d06df959/packages/peregrine/src/Price/Price.js#L23)

packages/venia-concept/src/shared/propShapes.js
```diff
@@ -0,0 +1,19 @@
+import { shape, number, string, bool } from 'prop-types';
```
DrewML: I'm very reluctant to add a "shared proptypes" file. We did this at my previous employer while using GraphQL. We ended up with dozens of permutations of similar-but-slightly-different shapes for each component. We also found it encouraged over-fetching of data since you didn't have to make a 1-off version of a shared proptype.

The other downside was that you no longer can just look at the file and see the shape of data. You now need to jump through the abstraction to the shared prop files, and hop back and forth between the component and the proptype implementation.

packages/venia-concept/src/components/ProductImageCarousel/thumbnailList.js
```diff
@@ -5,17 +5,25 @@ import { List } from '@magento/peregrine';
...
         classes: PropTypes.shape({
             root: PropTypes.string
         }),
-        items: PropTypes.arrayOf(PropTypes.object).isRequired
+        items: PropTypes.arrayOf(mediaGalleryEntry).isRequired
     };
 
     render() {
-        return <List renderItem={Thumbnail} {...this.props} />;
+        // linear-time sort is possible when we have a numeric 'position' prop
```
DrewML: Few notes:

1. The comment in the top of `render` doesn't seem necessary - let's remove it.
2. We may want to reconsider doing this in `render`. Could be memoized, but we can always address this later
3. It does not seem ideal to make the client-side do this sorting. Are we sure this list isn't already sorted when returned from the API? If not, we should log a feature request for it to be ordered.

packages/venia-concept/src/util/getUrlKey.js
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Extract the basename from a URL object, which corresponds to the `url_key`
+ * property of a Magento 2 GraphQL entity. Mostly used to get the url_key
+ * from a recently navigated-to URL.
+ *
+ * @param {URL} url
+ * @returns {string} A string for use as the `url_key` in a GraphQL query.
+ */
+export default function getUrlKey(url = window.location) {
```
DrewML: Don't think we need any of this - we can already get the product's ID for the product page query. The `urlResolver` query has a `canonical_url` field - this will always be in the format `catalog/product/view/id/{integer}`.

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -1,97 +1,137 @@
...
 
+import { mediaGalleryEntry, price } from 'src/shared/propShapes'
+
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+
+const productDetailQuery = gql`
+query productDetail($sku: String, $urlKey: String) {
+  productDetail: products(filter: { sku: {eq: $sku}, or: {url_key: {eq: $urlKey }}}, pageSize: 1, currentPage: 1) {
+    total_count
```
DrewML: Seems like multiple fields in this query aren't used in this PR afaict. Can we audit and remove unused fields?

packages/venia-concept/src/components/ProductImageCarousel/thumbnailList.js
DrewML: I just checked on my local and that list already comes back from the API sorted on `position`

packages/venia-concept/src/components/Currency/index.js
zetlen: This is in addition! Not all representations of currency are prices. Subtotals, taxes, difference amounts, etc.

packages/venia-concept/src/util/getUrlKey.js
```diff
@@ -0,0 +1,17 @@
...
+    // The URL key is the last path segment.
+    const basename = url.pathname.slice(url.pathname.lastIndexOf('/') + 1);
+    // TODO: this may be configurable, but Magento SEO urls appear to always
```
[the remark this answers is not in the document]
zetlen: Is it commonly changed though? And is the current setting available through an API somewhere?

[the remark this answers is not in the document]
DrewML: I couldn't answer either of those questions, but it turns out we don't need this code. See https://github.com/magento-research/pwa-studio/pull/90#discussion_r198437197

packages/venia-concept/src/components/Currency/index.js
zetlen: I should have looked at the Price component, but I didn't. It looks like the Price component is a better implementation of this; I just think it should be called `Currency`, since not every representation of a currency is a price (e.g. subtotals, difference amounts, etc). I'll switch to this component and separately advocate for its renaming.

packages/venia-concept/src/components/Currency/index.js
zetlen: Good, that's good news--though I'm removing this implementation anyway.

packages/venia-concept/src/shared/propShapes.js
zetlen: We have to find a compromise on introducing high-level abstraction. They're not always necessary, no, but it's necessary to create some. We have to lower cognitive load for developers, and busy, deep inline definitions are cognitive load.

I see how shared propType shapes could become annoying if you define shared shapes too early, but we will have a shared prop shapes file. *It's the equivalent of GraphQL fragments.* (In fact, we might later put them together in a convenient wrapper, but not yet.) We can debate what belongs there, but we will need higher-level objects. This exports actual immutable prop types, not plain objects for `PropTypes.shape()`. If a dev needs a different shape, they have to make it inline. Whereas if they need the same shape, they use the shared shape.

It may make things a little harder for maintainers and code reviewers in exchange, but in cases where that's a tradeoff, our project's mission is to lean in the third party dev's direction. We should keep this file.

packages/venia-concept/src/components/ProductImageCarousel/thumbnailList.js
zetlen: I did this because of the presence of a `position` member in the first place. In my experience, something like that means that collection order is not significant. I'm gonna ask about this real quick.

packages/venia-concept/src/RootComponents/Product/Product.js
zetlen: Sure!

packages/venia-concept/src/util/getUrlKey.js
DrewML: Been meaning to do this anyway. Opened a PR to feed this to the RootComponent https://github.com/magento-research/pwa-studio/pull/93

packages/venia-concept/src/components/Gallery/item.js
```diff
@@ -112,6 +113,9 @@ class GalleryItem extends Component {
     /**
      * Product images are currently broken and pending a fix from the `graphql-ce` project
      * https://github.com/magento/graphql-ce/issues/88
+     *
+     * When using sample data, which symlinks to bypass cache,
+     * you can simple prepend /media/catalog/product/, which we'll do from .env.
```
jimbo: Good to know. 👍

packages/venia-concept/src/shared/images.js
```diff
@@ -1,5 +1,7 @@
 // corresponds to a 300x372 transparent png
 // keep in sync with constants above
 // TODO: generate this programmatically?
-export const imagePlaceholderUri =
+export const transparentPlaceholder =
     'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABkAAAAfCAQAAAC4ua71AAAAGklEQVR42mNkIBkwjmoZ1TKqZVTLqJYRpgUAaP0AIAQAObYAAAAASUVORK5CYII=';
+
+export const grayPlaceholder = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAQAAAAFCAQAAADIpIVQAAAADklEQVR42mNkgAJGIhgAALQABsHyMOcAAAAASUVORK5CYII=';
```
jimbo: None of the original placeholders were gray; they were all transparent. Have you changed one of them to gray or is this misnamed?

packages/venia-concept/src/components/ProductImageCarousel/thumbnailList.js
jimbo: Makes sense to me. I don't see a problem with this. If items have a `position` prop, that ought to be the sort order regardless of how the API returns them.

Though I agree it should at least be memoized a bit (i.e., cache only the last result of filter + reduce).

packages/venia-concept/src/shared/propShapes.js
jimbo: I have to agree with @DrewML. This is a path to the dark side. If the product is ever ripe for this optimization, that'll be clear in the future and it'll make a good PR then.

As you said, if a developer needs a different shape, they have to make it inline. That's going to encourage them to use the wrong shape for convenience. I believe we want people to write granular, specific, custom queries on a per-component basis, not use the most convenient existing query.

packages/venia-concept/src/util/getUrlKey.js
zetlen: Thanks for that! I was halfway through writing that just now.

However, we still need the `url_key`, because of https://github.com/magento/graphql-ce/issues/86. We could also open an issue asking for the `urlResolver` to be an interface and we could add inline fragments for things like `sku`, but that would take longer than just doing it this way.

packages/venia-concept/src/components/ProductImageCarousel/thumbnailList.js
zetlen: I implemented this as a `shouldComponentUpdate`! Good for me.

I also checked with the GQL team, and they confirmed that the returned order of the collections is not guaranteed to be position order, and `position` should win. They should change it if they can, and we should map GQL query results somewhere in our data layer to ensure that array order IS significant, but that's a future optimization.

Lastly, the linear-time sort comment seems like a nice reassurance to me; it's much faster than the `O(nlogn)` of `Array.prototype.sort()`.

packages/venia-concept/src/shared/images.js
zetlen: Looks like I made a bad assumption based on some CSS, and the fact that those two base64 strings are slightly different. But examining them...they're both transparent and they only differ by size. 
image
image

Whoops! I'll just use the one `transparentPlaceholder`.

zetlen:
@DrewML @jimbo Thanks for all your input so far. It's especially important for this to merge soon, so I have to ask for you to do followup reviews somewhat quickly. I realize that travel and schedule makes that hard, but I still have to ask.

DrewML:
@zetlen I can review again in the next few hours once I'm in Valencia

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -1,96 +1,151 @@
...
-                    </section>
-                    <section className={classes.related}>
-                        <h2 className={classes.relatedTitle}>
-                            <span>Similar Products</span>
-                        </h2>
-                    </section>
-                </article>
+                <Query
+                    query={productDetailQuery}
+                    variables={{ urlKey: getUrlKey() }}
```
DrewML: Can we add a `TODO` that links to an issue on the [`graphql-ce`](https://github.com/magento/graphql-ce) repo with a request to add a filter by entity ID?

packages/venia-concept/src/components/Gallery/item.js
DrewML: Will this leave broken images for those not using the sample data? Anton said there is a convention we can use to get _any_ product photo with the URL we get back from GraphQL. It will be the full size (for now), but we wouldn't have to hack it

packages/venia-concept/src/components/ProductImageCarousel/carousel.js
```diff
@@ -4,34 +4,62 @@ import PropTypes from 'prop-types';
...
+                position: PropTypes.number.isRequired,
+                disabled: PropTypes.bool,
+                file: PropTypes.string.isRequired
             })
-        )
+        ).isRequired
     };
 
+    shouldComponentUpdate({ images }) {
+        return (
```
DrewML: I don't think this should be a `shouldComponentUpdate`. This actually will run (in most cases) just as frequently as `render`. So now, if nothing has changed, we're looping over all the images _3 times_ (the filter in `render`, the sort in `render`, and the `some` in here).

[The recommended approach](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization) for possibly-expensive derivations from state in `render` is to use memoization. The better alternative is `PureComponent`, but we're writing these components for third parties, so we can't guarantee they'll always pass in an array with a new identity.

/cc @jimbo to correct me if I'm wrong about anything here

packages/venia-concept/src/components/ProductImageCarousel/carousel.js
```diff
@@ -4,34 +4,62 @@ import PropTypes from 'prop-types';
...
+                    image.disabled !== disabled ||
+                    image.file !== file
+                );
+            })
+        );
+    }
+
     render() {
         const { classes, images } = this.props;
+        // the order of the array is not guaranteed to be the position order,
```
DrewML: Can we link to the issue you mentioned creating to ensure the API returns a sorted list?

packages/venia-concept/webpack.config.js
```diff
@@ -104,6 +104,9 @@ module.exports = async function(env) {
                     phase === 'production' || enableServiceWorkerDebugging
                         ? serviceWorkerFileName
                         : false
+                ),
+                'process.env.MAGENTO_BACKEND_PRODUCT_MEDIA_PATH': JSON.stringify(
```
DrewML: This needs some form of documentation

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -1,96 +1,151 @@
...
+                    {({ loading, error, data }) => {
+                        if (error) return <div>Data Fetch Error</div>;
+                        if (loading) return <div>Fetching Data</div>;
+
+                        const product = data.productDetail.items[0];
+                        const { regularPrice } = product.price;
+
+                        return (
+                            <article className={classes.root}>
+                                <section className={classes.title}>
```
DrewML: Most of the keys we're doing lookups on with `classes` are missing in the proptypes definition for this component.

packages/venia-concept/src/components/Gallery/item.js
zetlen: Yeah it would. And that would be awesome--we can hit it next sprint.

packages/venia-concept/src/components/ProductImageCarousel/carousel.js
zetlen: Memoizing!

packages/venia-concept/src/RootComponents/Product/Product.js
zetlen: CSS classes are too fast-changing to solidify as a type; they're just a dictionary of strings, so I changed the prop type to that.

packages/venia-concept/src/RootComponents/Product/Product.js
zetlen: I put it above the query definition.

packages/venia-concept/src/RootComponents/Product/Product.js
DrewML: I think you should discuss this with @jimbo before deviating. There is a good reason these are documented the way they are. They're a public API

jimbo: > There is a good reason these are documented the way they are. They're a public API.

This. If you're writing CSS, you're targeting classnames. If those classnames are changing, your CSS needs changing. That's an interface.

> CSS classes are too fast-changing to solidify as a type

This really shouldn't be the case with CSS modules. Since the solution to a problem is never "add a classname to it", the classnames shouldn't really change unless a structure change necessitates it. As with any API, if you're changing it too frequently, you should probably be thinking about it more.

zetlen: Those are fair points; the `objectOf` prop type seems a bit useless as a consequence. Let's add all the classnames in a later PR; the missing `classes` prop types [predate this PR](https://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/RootComponents/Product/Product.js#L15-L17), and we don't want this one to get too large. I'll revert the change I made and get rid of the `objectOf` proptype.

jimbo: It's ok, I fixed it for you. 🙂

PR #93, opened 2018-06-27
Pass ID of entity to rendered RootComponent

DrewML:
## This PR is a:

[ ] New feature
[X] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

This passes the ID of the entity being viewed to the RootComponent. This should make it easier to craft GraphQL queries based on the ID, without needing to do manual URL parsing of the customer-facing URLs in the individual roots.

image

### DIFF
--- a/packages/peregrine/src/Router/MagentoRouteHandler.js
+++ b/packages/peregrine/src/Router/MagentoRouteHandler.js
@@ -36,18 +36,22 @@ export default class MagentoRouteHandler extends Component {
             apiBase,
             __tmp_webpack_public_path__
         })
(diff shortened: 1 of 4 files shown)


zetlen:
It looks good to me too, and is definitely the right API for RootComponents. But do we need to parse the canonical URL? I believe that `id` is an Int field in `urlResolver`!

DrewML:
lol, don't know how I missed the ID field. Will update once I get to the hotel in Valencia and have wifi

DrewML:
Updated to use the `id` field from the `urlResolver` query.

PR #104, opened 2018-06-29
[WIP] Checkout Workflow

jimbo:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

Add a basic checkout workflow inside the minicart.

## Additional information

Work in progress. Do not merge.

### DIFF
--- a/packages/venia-concept/src/components/Checkout/exit.js
+++ b/packages/venia-concept/src/components/Checkout/exit.js
@@ -12,7 +12,7 @@ class Exit extends Component {
             footer: string,
             root: string
         }),
-        resetCheckout: func
+        resetCheckout: func.isRequired
     };
 
     render() {
--- a/packages/venia-concept/src/components/Checkout/exit.js
+++ b/packages/venia-concept/src/components/Checkout/exit.js
@@ -1,0 +1,32 @@
+import { Component, createElement } from 'react';
+import { func, shape, string } from 'prop-types';
+
+import classify from 'src/classify';
+import ResetButton from './resetButton';
+import defaultClasses from './exit.css';
+
+class Exit extends Component {
+    static propTypes = {
+        classes: shape({
+            body: string,
+            footer: string,
+            root: string
+        }),
+        resetCheckout: func
+    };
+
+    render() {
+        const { classes, resetCheckout } = this.props;
+
+        return (
+            <div className={classes.root}>
+                <div className={classes.body}>Thank you for your order!</div>
+                <div className={classes.footer}>
+                    <ResetButton resetCheckout={resetCheckout} />
+                </div>
+            </div>
+        );
+    }
+}
+
+export default classify(defaultClasses)(Exit);
--- a/packages/venia-concept/src/components/Checkout/flow.js
+++ b/packages/venia-concept/src/components/Checkout/flow.js
@@ -1,5 +1,5 @@
 import { Component, createElement } from 'react';
-import { func, shape, string } from 'prop-types';
+import { func, oneOf, shape, string } from 'prop-types';
 
 import classify from 'src/classify';
 import Entrance from './entrance';
@@ -15,15 +15,17 @@ const stepMap = {
     ACCEPTED: 'STEP_3'
 };
 
+const stepEnum = Object.keys(stepMap);
+
 class Flow extends Component {
     static propTypes = {
         classes: shape({
             root: string
         }),
-        resetCheckout: func,
-        requestOrder: func,
-        status: string,
-        submitOrder: func
+        resetCheckout: func.isRequired,
+        requestOrder: func.isRequired,
+        status: oneOf(stepEnum).isRequired,
+        submitOrder: func.isRequired
     };
 
     render() {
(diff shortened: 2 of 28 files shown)


packages/venia-concept/src/components/Checkout/flow.js
```diff
@@ -0,0 +1,61 @@
...
+
+const stepMap = {
+    READY: 'STEP_1',
+    REQUESTING: 'STEP_1',
+    MODIFYING: 'STEP_2',
+    SUBMITTING: 'STEP_2',
+    ACCEPTED: 'STEP_3'
+};
+
+class Flow extends Component {
```
zetlen: This is a really good, really clear flow component!

packages/venia-concept/src/components/Checkout/flow.js
```diff
@@ -0,0 +1,61 @@
...
+                );
+                break;
+            }
+            case 'STEP_2': {
+                child = <Form status={status} submitOrder={submitOrder} />;
+                break;
+            }
+            case 'STEP_3': {
+                child = <Exit resetCheckout={resetCheckout} />;
+                break;
```
zetlen: We should have a default case that displays or throws an error here.

packages/venia-concept/src/components/Checkout/exit.js
```diff
@@ -0,0 +1,32 @@
...
+import defaultClasses from './exit.css';
+
+class Exit extends Component {
+    static propTypes = {
+        classes: shape({
+            body: string,
+            footer: string,
+            root: string
+        }),
+        resetCheckout: func
```
zetlen: Should this prop be required?

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -0,0 +1,46 @@
...
+
+    render() {
+        const { classes, status, submitOrder } = this.props;
+        const today = new Date().toDateString();
+
+        return (
+            <div className={classes.root}>
+                <div className={classes.body}>
+                    <Section label="Ship To">
+                        <span>Add Shipping Information</span>
```
zetlen: We always want to add TODO comments for unimplemented sections like this and below.

packages/venia-concept/src/components/Checkout/submitButton.js
```diff
@@ -0,0 +1,26 @@
...
+import { func, string } from 'prop-types';
+
+import Button from 'src/components/Button';
+
+const isDisabled = status => ['ACCEPTED', 'SUBMITTING'].includes(status);
+
+class SubmitButton extends Component {
+    static propTypes = {
+        status: string,
+        submitOrder: func
```
zetlen: This one definitely should be required, since it has no purpose without an onClick callback.

packages/venia-concept/src/components/Checkout/wrapper.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+const mapStateToProps = ({ checkout }) => ({ checkout });
+
+const mapDispatchToProps = dispatch => ({
+    resetCheckout: () => dispatch(resetCheckoutAction()),
+    requestOrder: () => dispatch(requestOrderAction()),
+    submitOrder: () => dispatch(submitOrderAction())
+});
```
zetlen: Is there an advantage to the dispatch callback, rather than the object form:

```js
const mapDispatchToProps = {
    resetCheckout: resetCheckoutAction,
    requestOrder: requestOrderAction,
    submitOrder: submitOrderAction
};
```
I ask not to be nitpicky, but because I've used the object literal pattern in other PRs and I want to know if you're aware of some drawback to it.

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -31,10 +33,26 @@ class MiniCart extends Component {
         data: mockData
     };
 
+    async componentDidMount() {
+        const {
```
zetlen: This async load pattern is intriguing; do you intend for this to become a pattern that we'd repeat in other components, or even abstract away?

packages/venia-concept/src/components/Checkout/exit.js
jimbo: Yes, probably. I'll do a sweep and see if I can make some props required. 👍

packages/venia-concept/src/components/Checkout/flow.js
jimbo: Good point. 👍

packages/venia-concept/src/components/Checkout/form.js
jimbo: True. I'll add some. 👍

packages/venia-concept/src/components/Checkout/wrapper.js
jimbo: No, there's not. I can probably simplify it to match your suggestion. 👍

packages/venia-concept/src/components/MiniCart/miniCart.js
jimbo: Actually, it's just sort of a convenient break point. I know that `MiniCart` will be mounted in an inactive/hidden state, so I can request some of its children asynchronously and have them rendered before the user activates and sees the component.

We're going to need a more sophisticated solution, probably involving redux actions, to really do this right, I think.

packages/venia-concept/src/components/Checkout/wrapper.js
jimbo: So there is a reason after all.

When a child component calls the function `submitOrder`, `submitOrder` needs to fire an async action (i.e., dispatch an async function). If we don't use `mapDispatchToProps` in this way, then calling `submitOrder` will not be able to fire an action, because `dispatch` will be undefined.

I think this process is correct.

PR #140, opened 2018-07-10
feat(checkout): Integrate live product, cart, checkout

zetlen:
Merges work from the add-to-cart functionality from a research branch
and the checkout proof-of-concept implemented in a pull request.

 - Implement a Magento REST API client in Peregrine.
 - Simplify redux middleware: use only `redux-thunk` and async/await to
 handle asynchronicity in action creators.
 - Split up actions and reducers into app, cart, and checkout areas.

## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary


packages/venia-concept/src/actions/app.js
```diff
@@ -0,0 +1,12 @@
+import timeout from 'src/util/timeout';
+import { drawer as drawerMs } from 'src/shared/durations';
+
+export const toggleDrawer = drawerName => async dispatch => {
```
DrewML: This doesn't need to be an async function. The code is the equivalent of:

```js
export const toggleDrawer = drawerName => async dispatch => {
   dispatch({
      type: 'TOGGLE_DRAWER',
      payload: drawerName
   });
   return Promise.resolve(timeout(drawerMs)); // unnecessary extra Promise
};
```

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -28,30 +31,38 @@ class MiniCart extends Component {
         })
     };
 
-    static defaultProps = {
-        // TODO: remove when connected to graphql
-        data: mockData
-    };
+    constructor(...args) {
+        super(...args);
+        this.loadAsyncDependencies();
```
DrewML: 2 notes:

1. Class constructors should never have side-effects.
2. Data fetching should be done in `componentDidMount`, which has been said frequently the closer we are to async react (Fiber) - https://reactjs.org/docs/react-component.html#componentdidmount.

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -20,27 +24,20 @@ class Product extends Component {
...
-            }).isRequired
-        })
+            item_id: number.isRequired,
+            name: string.isRequired,
+            price: number.isRequired,
+            product_type: string,
+            qty: number.isRequired,
+            quote_id: string,
+            sku: string.isRequired
+        }),
```
DrewML: `item` should be `isRequired`, since we're accessing properties on it without any safety checks

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -0,0 +1,112 @@
...
+
+import classify from 'src/classify';
+import Carousel from 'src/components/ProductImageCarousel';
+import Quantity from 'src/components/ProductQuantity';
+import RichText from 'src/components/RichText';
+import defaultClasses from './productFullDetail.css';
+
+class ProductFullDetail extends Component {
+    static propTypes = {
+        classes: objectOf(string).isRequired,
```
DrewML: These classes should be documented. See comment from @jimbo https://github.com/magento-research/pwa-studio/pull/90#discussion_r199540398

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -0,0 +1,112 @@
...
+                shape({
+                    label: string,
+                    position: number,
+                    disabled: bool,
+                    file: string.isRequired
+                })
+            ),
+            description: string
+        }).isRequired,
+        onClickAddToCart: func.isRequired
```
DrewML: I think `addToCart` would be a better name. Not crazy to think this could be used in the component for other interactions that trigger add to cart. Touch/enter/etc

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -172,4 +129,11 @@ class Product extends Component {
...
 
-export default classify(defaultClasses)(Product);
+export default connect(
+    ({ cart: { guestCartId } = {} }) => ({ guestCartId }),
+    {
+        addItemToCart,
+        getCartDetails,
+        toggleCart
+    }
+)(Product);
```
jimbo: Please consider defining `mapStateToProps` and `mapDispatchToProps` just before this export, as we've done in the other components. It's a bit less terse but more clear.

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -28,30 +31,38 @@ class MiniCart extends Component {
...
+    async loadAsyncDependencies() {
+        const [
+            CheckoutComponent,
+            checkoutReducer,
+            makeCartReducer
+        ] = (await Promise.all([
+            import('src/components/Checkout'),
+            import('src/store/reducers/checkout'),
+            import('src/store/reducers/cart')
+        ])).map(mod => mod.default);
```
jimbo: Good. `Promise.all()` is the way to go for sure.

packages/venia-concept/src/components/MiniCart/miniCart.js
jimbo: Correct on both counts. This was done in `componentDidMount` for a reason.

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -65,24 +76,56 @@ class MiniCart extends Component {
                     </Trigger>
                 </div>
                 <div className={classes.body}>
-                    <ProductList items={data} />
+                    {cartId ? (
+                        <ProductList
+                            currencyCode={cartCurrencyCode}
+                            items={cart.details.items}
+                        />
+                    ) : null}
```
jimbo: Please don't do multiline expressions in JSX, it defeats the purpose of the syntax. Consider [using a getter](https://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/components/ProductOptions/option.js#L25-L27) or a helper method, then [inject the result directly](https://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/components/MiniCart/miniCart.js#L82), as seen in this file and others.

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -65,24 +76,56 @@ class MiniCart extends Component {
...
                     <div className={classes.summary}>
-                        <dl className={classes.totals}>
-                            <dt className={classes.subtotalLabel}>
-                                <span>Subtotal (4 Items)</span>
-                            </dt>
-                            <dd className={classes.subtotalValue}>
-                                <Price currencyCode="USD" value={528} />
-                            </dd>
-                        </dl>
+                        {cartId && cart.totals && 'subtotal' in cart.totals ? (
```
jimbo: See previous comment.

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -65,24 +76,56 @@ class MiniCart extends Component {
...
+                details.currency &&
+                details.currency.quote_currency_code;
+            return {
+                cart,
+                cartId,
+                cartCurrencyCode
+            };
+        },
+        { getCartDetails }
+    )
```
jimbo: Please consider defining mapStateToProps and mapDispatchToProps just before this export, as we've done in the other components. It's a bit less terse but more clear.

packages/venia-concept/src/components/MiniCart/product.css
```diff
@@ -13,9 +13,10 @@
 .image {
     border-radius: 2px;
     box-shadow: inset 0 0 0 1px rgb(var(--venia-border));
+    background-color: rgb(var(--venia-grey));
     grid-area: image;
-    height: 100px;
-    width: 80px;
+    background-position: 50% 50%;
+    background-size: cover;
```
jimbo: I guess I can try to find a plugin that will alphabetize CSS declarations. Is is not clear that they're all alphabetized?

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -20,27 +24,20 @@ class Product extends Component {
...
+        currencyCode: string.isRequired
     };
 
     get options() {
         const { classes, item } = this.props;
-
+        if (!item.options) return [];
```
jimbo: Consider adding some whitespace.

```js
let foo = 'bar';

if (condition) {
  foo = 'baz';
}

return foo;
```

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -51,28 +48,35 @@ class Product extends Component {
...
+                        backgroundImage: `url(${makeProductMediaPath(
+                            item.image.file
+                        )})`
+                    }}
+                />
                 <div className={classes.name}>{item.name}</div>
-                <dl className={classes.options}>{options}</dl>
+                {options.length > 0 ? (
+                    <dl className={classes.options}>{options}</dl>
+                ) : null}
```
jimbo: Consider [using a getter](https://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/components/ProductOptions/option.js#L25-L27) or a helper method, then [inject the result directly](https://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/components/MiniCart/miniCart.js#L82).

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
jimbo: Agreed.

packages/venia-concept/src/shared/durations.js
```diff
@@ -0,0 +1,5 @@
+export const drawer = 192;
```
jimbo: Good idea. This is actually the duration of the drawer's closing animation, though. The opening animation has a different duration, so maybe this could be `drawerClose` or something.

packages/venia-concept/src/store/reducers/app.js
```diff
@@ -13,6 +14,12 @@ const reducer = (state = initialState, { payload, type }) => {
             };
         }
         default: {
+            if (error) {
+                return {
+                    ...state,
+                    error
+                };
+            }
```
jimbo: Since there are no breaks in this `switch` statement, does that mean this default case will always be reached? That is, if there is ever an `error`, this code will be executed? Interesting.

packages/venia-concept/src/store/reducers/cart.js
```diff
@@ -0,0 +1,76 @@
+import debounce from 'lodash.debounce';
+import BrowserPersistence from 'src/util/simplePersistence';
```
jimbo: I like the idea of abstracting our choice of persistence.

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -51,28 +48,35 @@ class Product extends Component {
...
-        const { classes, item } = props;
-        const amount = item.price.regularPrice.amount;
+        const { classes, item, currencyCode } = props;
 
         return (
             <li className={classes.root}>
-                <div className={classes.image} />
+                <div
+                    className={classes.image}
+                    style={{
```
jimbo: Please define this style object outside of the return statement, either as a `const` or a getter.

jimbo:
@zetlen Where did we land on async actions? I see that we're using `redux-thunk` (great!) but did you end up wanting to use `redux-promise-middleware`?

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -99,9 +98,12 @@ class Product extends Component {
         })
     };
 
-    render() {
-        const { classes } = this.props;
+    addToCart = async (item, quantity) => {
+        const { guestCartId } = this.props;
+        await this.props.addItemToCart({ guestCartId, item, quantity });
```
DrewML: This `await` looks superfluous, since we're not returning a value and the caller doesn't want on the resolution

packages/peregrine/docs/RestApi.md
```diff
@@ -0,0 +1,190 @@
+# REST API Clients
```
DrewML: I'm hesitant about introducing a large abstraction for REST API calls at this stage (our first integration to a non-GQL API). I'm sure people are getting tired of me beating this drum at this point, but I strongly believe that an abstraction should not be created over some other API until we've truly felt the pain of not having that abstraction. At that point, my preference is always going to be the minimal abstraction that covers the immediate use-cases.

For this particular abstraction, I have a few specific concerns.

1. This abstracts over the `fetch` API, but limits a consumer's ability to use other features exposed through `fetch`. Examples include having no control over the `credentials` option, the `headers` option, the `integrity` option, etc.

2. I don't see that we're using `addInterrupt` anywhere in this code. Once we've exposed an API we have to support it, and I'm not incredibly comfortable exposing an API we don't have a need for. There is a major hidden cost there.

I feel like I need help understanding what problem this `fetch` abstraction addresses in this specific PR. I've worked on several fairly large React/Redux CRUD applications, and I personally haven't seen the problems that this abstraction (I believe) is aiming to solve.

davemacaulay: _I'm not directly involved but I'm a strong believer in keeping things simple and I hate complex abstractions so_

From my previous experience at Gene: Magento already has _a lot_ of abstractions. They're taxing and confusing to learn and master, making the learning curve and time needed to work a new system much more confusion. On top of the complexity that PWA is already going to be, I'm strongly in favour of keeping things as simple as possible.

If we need this abstracted, for whatever reason, I feel like enough people use REST with Node that there is a community solution that is well understood and _may already be known_ by developers coming onto PWA (like maybe the one Microsoft wrote: https://github.com/Microsoft/typed-rest-client). Let's not continue the practice of Magento reinventing the wheel with everything we do.

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -172,4 +129,11 @@ class Product extends Component {
     }
 }
 
-export default classify(defaultClasses)(Product);
+export default connect(
+    ({ cart: { guestCartId } = {} }) => ({ guestCartId }),
```
DrewML: I personally find this difficult to visually parse without sitting and concentrating. Can we maybe try something less terse?

jimbo: Agreed. Nested destructuring is tough to read.

packages/venia-concept/src/components/MiniCart/product.css
zetlen: It was not, because I don't read letters, I read words. (Insert link to some kind of neuroscience article that absolves me here)

packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
```diff
@@ -0,0 +1,254 @@
+import M2ApiResponseError from './M2ApiResponseError';
```
DrewML: @zetlen Did you mean to `rm` this file in 42bf44d?

PR #153, opened 2018-07-13
feat(checkout): Add field components

jimbo:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

Add components to use for form fields. These fields will support advanced features, such as validation, and match the Venia design by default.

### DIFF
--- a/packages/venia-concept/src/components/Field/__tests__/control.spec.js
+++ b/packages/venia-concept/src/components/Field/__tests__/control.spec.js
@@ -1,0 +1,82 @@
+import { createElement } from 'react';
+import { configure, shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+
+import Control from '../control';
+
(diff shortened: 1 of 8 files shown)


jimbo:
> Can you give me like 100 words on comparison with existing things? I know this isn't solving the same problems as e.g. https://react-form.js.org/ or https://www.muicss.com/docs/v1/react/forms but it would be worthwhile to explain why we made a collection of fields from scratch.

The truth is, it is trying to solve those same problems; if we continued this work, it would eventually end up as fully featured as other options out there. Forms and fields are all about labels, values, and validation, and it gets complicated.

There are benefits to this approach—`Fragment` for lists, `Set` for selection, no `id` attributes necessary—but they're probably not critical at this point. On reflection, I'll close this PR and revisit if necessary down the road.

PR #156, opened 2018-07-16
Use Button component throughout the app

jimbo:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

Use the `Button` component throughout the app, instead of basic `<button>` elements. This allows us to add icons to any button at any time, and unifies the styling a bit more easily.


zetlen:
This looks great--you've just gotta rebase it.

jimbo:
> This looks great--you've just gotta rebase it.

Done! 👍

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -3,6 +3,7 @@ import { arrayOf, bool, func, number, shape, string } from 'prop-types';
 import { Price } from '@magento/peregrine';
 
 import classify from 'src/classify';
+import Button from 'src/button';
```
artKozinets: Will be better with this `import Button from 'src/components/Button';`

jimbo: Good catch. This line introduced a bug, which I had to fix with #185.

The bug made its way into `master` because I made changes to this PR after it was already approved. We'll be tightening up our processes to not let this kind of bug slip past again.

PR #164, opened 2018-07-20
feat: multicasting REST API client

zetlen:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

 - Added a REST API client to Peregrine, with features:
    - Implements Fetch API and takes the same options
    - Configured by default to handle Magento 2 REST API
    - Returns REST resources by default (parsed JSON), not HTTP responses
    - Identifies redundant, safe and idempotent requests and
    multicasts (shares) the highest-priority request to reduce network
    traffic for PWA performance
    - Exposes underlying `M2ApiRequest` constructor
        - `.abortRequest()` method cancels request using AbortSignal
        - `.getResponse()` method clones Response for multicasting Body
        stream
 - Refactored Venia cart actions to use Rest API client

## Summary

This fixes some (unfiled) issues present in #140, including:
 - Potential for separate components to make redundant REST calls
 (solved in GraphQL with batching and merging)
 - Redundant, non-DRY fetch calls

### DIFF
--- a/packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
+++ b/packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
@@ -11,7 +11,7 @@ const withDefaultHeaders = headerAdditions => {
         if (headerAdditions instanceof Headers) {
             /* istanbul ignore next: current phantomJS doesn't support */
             if (headerAdditions.entries) {
-                for (let [name, value] of headerAdditions.entries()) {
+                for (const [name, value] of headerAdditions) {
                     headers.append(name, value);
                 }
             } else if (headerAdditions.forEach) {
@@ -27,7 +27,7 @@ const withDefaultHeaders = headerAdditions => {
                 );
             }
         } else {
-            for (let [name, value] of Object.entries(headerAdditions)) {
+            for (const [name, value] of Object.entries(headerAdditions)) {
                 headers.append(name, value);
             }
         }
@@ -166,7 +166,7 @@ class M2ApiRequest {
                     throw e;
                 }
             )
-            .then(res => {
+            .then(response => {
                 // WHATWG fetch will only reject in the unlikely event
                 // of an error prior to opening the HTTP request.
                 // It pays no attention to HTTP status codes.
@@ -174,9 +174,9 @@ class M2ApiRequest {
                 // corresponding to status codes in the 2xx range.
                 // An M2ApiRequest will reject, passing server errors
                 // to the client, in the event of an HTTP error code.
-                if (!res.ok) {
+                if (!response.ok) {
                     return (
-                        res
+                        response
                             // The response may or may not be JSON.
                             // Let M2ApiResponseError handle it.
                             .text()
@@ -187,13 +187,13 @@ class M2ApiRequest {
                                 throw new M2ApiResponseError({
                                     method: this.opts.method,
                                     resourceUrl: this.resourceUrl,
-                                    res,
+                                    response,
                                     bodyText
                                 });
                             })
                     );
                 }
-                return res;
+                return response;
             });
     }
     /**
@@ -227,10 +227,10 @@ class M2ApiRequest {
             // swallow the AbortError we expect, and replace it with
             // the promise from our rolling request.
             if (error.name === 'AbortError') {
-                const newInFlightMatch = MulticastCache.match(this);
-                if (newInFlightMatch) {
+                const replacedInFlightMatch = MulticastCache.match(this);
+                if (replacedInFlightMatch) {
                     // There is a rolling request in the cache to override!
-                    return newInFlightMatch.getResponse();
+                    return replacedInFlightMatch.getResponse();
                 }
             }
             throw error;
@@ -255,8 +255,11 @@ export default M2ApiRequest;
  */
 export function request(resourceUrl, opts) {
     const req = new M2ApiRequest(resourceUrl, opts);
+
     req.run();
+
     const promise = req.getResponse();
+
     if (opts && opts.parseJSON === false) {
         return promise;
     }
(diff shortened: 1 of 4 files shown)


packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
```diff
@@ -0,0 +1,264 @@
...
+                });
+                /* istanbul ignore next: should never happen, trivial to test*/
+            } else {
+                console.warn(
+                    'Could not use headers object supplied to M2ApiRequest',
+                    headerAdditions
+                );
+            }
+        } else {
+            for (let [name, value] of Object.entries(headerAdditions)) {
```
jimbo: Likewise, `const` over `let` if possible?

packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
```diff
@@ -0,0 +1,264 @@
...
+     * network calls.
+     *
+     * The current logic for rolling requests is determined by the `cache`
+     * option. [Cache modes](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache)
+     * `reload` and `no-store` both indicate complete cache bypass. This
+     * logically implies that the user has just changed server state and wants
+     * to force retrieve an updated resource, so multicasting a prior request
+     * would not be appropriate--the response may not reflect the more recent
+     * change.
+     * @returns {boolean} True if the request is rolling.
```
jimbo: Great comments.

packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
```diff
@@ -0,0 +1,264 @@
...
+     */
+    isRolling() {
+        return this.opts.cache === 'no-store' || this.opts.cache === 'reload';
+    }
+    /**
+     * Make sure not to multicast POST requests which have a nonempty body,
+     * since they are unsafe and non-idempotent, so each call may mutate
+     * server-side state.
```
jimbo: Is this something that can be caught as an error or a warning, such that it might appear in a console? If so, is that happening yet?

packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
```diff
@@ -0,0 +1,264 @@
...
+        // Cache this request for future use.
+        MulticastCache.store(this);
+
+        const promise = this._fetch().catch(error => {
+            // Rolling requests cause prior matching requests to abort.
+            // Their consumers will get an unexpected error unless we
+            // swallow the AbortError we expect, and replace it with
+            // the promise from our rolling request.
+            if (error.name === 'AbortError') {
+                const newInFlightMatch = MulticastCache.match(this);
```
jimbo: Minor quibble, but can we rename this variable to not begin with `new`?

jimbo:
Overall this looks good and I think we can use it for API requests going forward. I want to base my checkout work on this so let's clean up a few minor things and get it merged. 👍

packages/venia-concept/src/actions/cart.js
```diff
@@ -31,121 +27,118 @@ const createGuestCart = async dispatch => {
...
-        payload = {
-            details,
-            totals
-        };
-    } catch (e) {
-        payload = e;
-        error = true;
+    // no cart populated? wait to show the drawer so it's not blank
+    if (!cart.id) {
+        await gettingCartDetails;
```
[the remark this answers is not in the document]
zetlen: Good catch. We can parallelize closing the open drawer and getting cart details.

packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
```diff
@@ -0,0 +1,264 @@
...
+const withDefaultHeaders = headerAdditions => {
+    const headers = new Headers({
+        'Content-type': 'application/json',
+        Accept: 'application/json'
+    });
+    if (headerAdditions) {
+        if (headerAdditions instanceof Headers) {
+            /* istanbul ignore next: current phantomJS doesn't support */
+            if (headerAdditions.entries) {
+                for (let [name, value] of headerAdditions.entries()) {
```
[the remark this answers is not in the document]
zetlen: `forEach` is not part of the standard. MDN doesn't make that clear for some reason, but we can't count on `forEach` being available, which is why I'm falling back to it only if I have to.

I will switch to `const` though.

packages/peregrine/src/RestApi/Magento2/MulticastCache.js
```diff
@@ -0,0 +1,69 @@
...
+}
+/**
+ * Remove a request from cache if it exists there.
+ * @param {M2ApiRequest} req
+ */
+export function remove(req) {
+    if (match(req) === req) {
+        inflight.delete(requestToKey(req));
+    }
+}
```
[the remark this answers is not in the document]
zetlen: `MulticastCache` is a singleton, so I implemented it as an ES6 module. It's equivalent at runtime to an object with function properties.

packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
zetlen: In this comment I mean [the REST definition of the term] 'unsafe' (http://restcookbook.com/HTTP%20Methods/idempotency/), which is not a bad condition; it just shouldn't be multicast. It can be overridden, but I don't think a warning is necessary. I figure if a dev forces `multicast: true` on a request, they know what they're doing.

zetlen:
@jimbo Thanks for the feedback--I made some corrections and clarifications and pushed an update. Feel free to merge if you approve (and it passes CI)!

PR #165, opened 2018-07-22
dev: Improve prod build and monorepo scripts

zetlen:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

 - Add more root-directory usability to monorepo scripts
 - Add `npm run watch:venia` mode that:
   - Simultaneously builds buildpack, peregrine, and venia
   - Hot reloads venia if either peregrine or venia change on disk
   - Fully restarts dev server if buildpack changes on disk
   - Interleaves and prefixes output
   - Doesn't use lerna to do this, because lerna's parallelizing console
   mode is not customizable enough
 - Fix production builds (by pegging them to very new browsers)

## Additional information

Production builds will need further optimization in the future.
TODO: Adjust production build for middle tier concept
TODO: Use performance analysis to optimize chunk layout

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
   },
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "build": "lerna --scope '@magento/{pwa-buildpack,peregrine}' exec -- npm run build",
-    "clean": "lerna --scope '@magento/{pwa-buildpack,peregrine}' exec -- npm run clean",
+    "build": "lerna --stream --scope '{@magento/pwa-buildpack,@magento/peregrine,theme-frontend-venia}' exec -- npm run build",
+    "clean": "lerna run clean && lerna clean --yes && rimraf ./node_modules",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "danger": "danger-ci",
     "lint": "eslint 'packages/**/*.js' --ignore-pattern node_modules",
@@ -22,12 +22,25 @@
     "prettier:check": "prettier --list-different 'packages/**/*.@(js|css)' '*.js'",
     "test": "jest",
     "test:debug": "node --inspect-brk node_modules/.bin/jest -i --watch",
-    "test:dev": "jest --watch"
+    "test:dev": "jest --watch",
+    "watch:all": "concurrently -c yellow.dim,green.dim,cyan.dim -n ␢,℗,☄︎  'npm run watch:buildpack' 'npm run watch:peregrine' 'npm run watch:venia-dev-server'",
+    "watch:buildpack": "cd packages/pwa-buildpack && npm run -s watch; cd - >/dev/null",
+    "watch:peregrine": "cd packages/peregrine && npm run -s watch; cd - >/dev/null",
+    "watch:venia": "cd packages/venia-concept && npm run -s watch; cd -",
+    "watch:venia-dev-server": "nodemon --exec npm --cwd packages/venia-concept start"
+  },
+  "nodemonConfig": {
+    "watch": [
+      "packages/pwa-buildpack/dist",
+      "packages/venia-concept/webpack.config.js"
+    ],
+    "delay": 1000
   },
   "devDependencies": {
     "@magento/eslint-config": "^1.0.0",
     "babel-eslint": "^8.2.3",
     "babel-plugin-syntax-dynamic-import": "^6.18.0",
+    "concurrently": "^3.6.0",
     "coveralls": "^3.0.1",
     "danger": "^3.7.18",
     "eslint": "^5.2.0",
@@ -39,7 +52,9 @@
     "greenkeeper-lockfile": "^2.4.0",
     "jest": "^23.4.0",
     "lerna": "^3.0.0-beta.21",
+    "nodemon": "^1.18.3",
     "prettier": "^1.13.5",
-    "prettier-check": "^2.0.0"
+    "prettier-check": "^2.0.0",
+    "rimraf": "^2.6.2"
   }
 }
--- a/packages/peregrine/.babelrc
+++ b/packages/peregrine/.babelrc
@@ -12,7 +12,7 @@
   "presets": [
     ["env", {
       "targets": {
-        "browsers": ["last 2 versions", "ie 11"]
+        "browsers": ["> 5%"]
       },
       "modules": false
     }]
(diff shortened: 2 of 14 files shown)


zetlen:
Feel free to clone and run `npm run watch:venia` and/or `npm run build && lerna run --stream --scope=theme-* build` to see the new hotness.

packages/peregrine/.babelrc
```diff
@@ -12,7 +12,7 @@
   "presets": [
     ["env", {
       "targets": {
-        "browsers": ["last 2 versions", "ie 11"]
+        "browsers": ["> 5%"]
```
jimbo: This should slim down the build. 👍

package.json
```diff
@@ -12,22 +12,34 @@
...
-    "clean": "lerna --scope '@magento/{pwa-buildpack,peregrine}' exec -- npm run clean",
-    "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
-    "danger": "danger-ci",
+    "reset": "lerna clean --yes && rm -rf ./node_modules && npm install",
+    "build": "lerna --stream --scope '@magento/{pwa-buildpack,peregrine}' exec -- npm run build",
     "postinstall": "lerna bootstrap && npm run build",
+    "watch:buildpack": "cd packages/pwa-buildpack && npm run -s watch",
+    "watch:peregrine": "cd packages/peregrine && npm run -s watch",
+    "watch:dev-server": "nodemon --exec npm --cwd packages/venia-concept start",
+    "watch:venia": "concurrently -c yellow.dim,green.dim,cyan.dim -n ␢,℗,☄︎  'npm run watch:buildpack' 'npm run watch:peregrine' 'npm run watch:dev-server'",
```
jimbo: Is this really the plan? Watch all of these at the same time? I'd rather just watch `venia`, personally. Can we make this one `watch:all` or something?

packages/peregrine/src/Router/resolveUnknownRoute.js
```diff
@@ -63,7 +63,13 @@ function remotelyResolveRoute(opts) {
  * @returns {Promise<{rootChunkID: number, rootModuleID: number}>}
  */
 function tempGetWebpackChunkData(pageType, webpackPublicPath) {
-    return fetch(new URL('roots-manifest.json', webpackPublicPath))
+    // In dev mode, `webpackPublicPath` may be a fully qualified URL.
+    // In production mode, it may be a pathname, which makes it unsafe
+    // to use as an API base. Normalize it as a full path using a DOM node
+    // as a native URL parser.
+    const parser = document.createElement('a');
```
jimbo: Can you elaborate? What does this workaround do that [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) doesn't?

packages/venia-concept/babel.config.js
```diff
@@ -21,9 +21,9 @@ const defaults = {
 // define preset-env config for production
 const presetEnvConfig = {
     targets: {
-        browsers: ['last 2 versions', 'ie >= 11'],
-        modules: false
-    }
+        browsers: ['> 5%']
+    },
+    modules: false
```
jimbo: is this a bug fix?

package.json
zetlen: This is definitely what you want. This has the exact same functionality that `npm start` in venia does today, with three key improvements:
1. If you change a Peregrine source file (e.g. moving a concept from Venia to Peregrine) it will hot reload and stay up to date, instead of requiring a manual Peregrine rebuild
2. If you change a Buildpack source file (e.g. changing the behavior of a plugin or DevServer), it will shut down the dev server, rebuild Buildpack, and restart the dev server to reflect the new changes.
3. You can run it from project root.

If you don't touch Peregrine or Buildpack, it's functionally the same.

packages/peregrine/src/Router/resolveUnknownRoute.js
zetlen: It's guaranteed to produce an absolute URL from a possibly-relative URL. The `webpackPublicPath` argument is relative in production and absolute in dev, so we need something that can handle both. URL objects cannot:

```js
// URL constructors deal in absolute URLs.
try {
  new URL('/hermano')
} catch(e) {
  console.warn(e.message, '/hermano'); // Failed to construct 'URL': Invalid URL
}

// They can use relative URLs if you supply a "base URL" as a second argument.
const absoluteBase = 'https://github.com/magento-research/pwa-studio/';
const url = new URL('pulls/165', absoluteBase);
console.log(url)
// Url { protocol: "https:" hostname: "github.com", pathname: "/magento-research/pwa-studio/pulls/165" }

// If the base URL is relative, they throw.
try {
  new URL('pulls/165', '/magento-research/pwa-studio/');
} catch(e) {
  console.warn(e.message, '/magento-research/pwa-studio/');
}

// Anchor tags will accept absolute or relative 'href' attributes:
const parserTag = document.createElement('a');
parserTag.setAttribute('href','/magento-research/pwa-studio');
// But their properties are always absolute:
console.log('parserTag.getAttribute("href")', parserTag.getAttribute("href"));
console.log('parserTag.href', parserTag.href); // "https://github.com/magento-research/pwa-studio/";

// So you can guarantee an absolute URL built with URL semantics
// if you can guarantee an absolute base, which we do with our parser.
// Which is a lot safer than string concatenation.

function appendPathToUrl(path, base) {
   const parser = document.createElement('a');
   parser.setAttribute('href', base);
   return new URL(path, parser.href).toString()
}

console.log(
  'true or false: using an anchor tag as a parser normalizes absolute URLs:'
  appendPathToUrl('pulls/165', '/magento-research/pwa-studio/') === 
  appendPathToUrl('pulls/165', 'https://github.com/magento-research/pwa-studio/')
);
// true
```

packages/venia-concept/babel.config.js
zetlen: Yep! For Webpack tree shaking.

package.json
jimbo: If you don't touch `peregrine` or `buildpack`, is it more expensive to watch them? Does it bog your system down more? That's my only concern.

jimbo: Yes, I think that would help. 👍

packages/peregrine/src/Router/resolveUnknownRoute.js
jimbo: I see, makes perfect sense to me now. Thanks for the explanation. 👍

zetlen:
@jimbo Made changes based on your feedback.

PR #166, opened 2018-07-22
WIP (DO NOT MERGE): PWADevServer as a decorator

zetlen:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary
PWADevServer began as a config generator, that would produce a
configuration object suitable to be assigned to the `devServer` property
of webpack config (which is only used by webpack-dev-server); However,
PWADevServer has grown to require information from other parts of the
config, and maybe also to modify the config. It made sense to
reimplement it as a decorator.

That said, we may not end up using PWADevServer in its current
incarnation, if we switch a middle tier concept less reliant on Webpack
dev tooling.

## Additional information

This is a work in progress published as a PR for consideration and
comment. Don't merge it!


zetlen:
No longer relevant due to imminent #248

PR #171, opened 2018-07-25
Category list component

artKozinets:
## This PR is a:

[X] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

New root Category list component to show subcategories from parent category.

## Issue
https://app.zenhub.com/workspace/o/magento-research/pwa-studio/issues/99

### How to use

Add to the page **`<CategoryList></CategoryList>`**
 
 
### Options

**`title`** - Category list component title, appears at the top of the block.
**`id`** - Main category id, to show child categories from.

### Example

 `<CategoryList title={'Shop by category'}></CategoryList>`

 `<CategoryList title={'Shop by category'} id={3}></CategoryList>`

### DIFF
--- a/packages/venia-concept/src/RootComponents/CMS/CMS.js
+++ b/packages/venia-concept/src/RootComponents/CMS/CMS.js
@@ -1,7 +1,18 @@
 import { createElement, Component } from 'react';
 
+import PropTypes from 'prop-types';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import Page from 'src/components/Page';
+import CategoryList from 'src/components/CategoryList';
+
 export default class CMS extends Component {
     render() {
-        return <div>CMS Page Stub</div>;
+
+        return (
+            <Page>
+                <CategoryList title={'Shop by category'} id={2}></CategoryList>
+            </Page>
+        );
     }
 }
--- a/packages/venia-concept/src/RootComponents/CMS/CMS.js
+++ b/packages/venia-concept/src/RootComponents/CMS/CMS.js
@@ -1,17 +1,12 @@
 import { createElement, Component } from 'react';
-
-import PropTypes from 'prop-types';
-import { Query } from 'react-apollo';
-import gql from 'graphql-tag';
 import Page from 'src/components/Page';
 import CategoryList from 'src/components/CategoryList';
 
 export default class CMS extends Component {
     render() {
-
         return (
             <Page>
-                <CategoryList title={'Shop by category'} id={2}></CategoryList>
+                <CategoryList title={'Shop by category'} id={2} />
             </Page>
         );
     }
--- a/packages/venia-concept/src/RootComponents/CMS/CMS.js
+++ b/packages/venia-concept/src/RootComponents/CMS/CMS.js
@@ -6,7 +6,7 @@ export default class CMS extends Component {
     render() {
         return (
             <Page>
-                <CategoryList title={'Shop by category'} id={2} />
+                <CategoryList title="Shop by category" id={2} />
             </Page>
         );
     }
--- a/packages/venia-concept/src/components/CategoryList/categoryList.css
+++ b/packages/venia-concept/src/components/CategoryList/categoryList.css
@@ -1,0 +1,48 @@
+.root {
+    padding: 1rem;
+}
+
+.header {
+    margin-bottom: 2rem;
+    text-align: center;
+}
+
+.title {
+    text-transform: uppercase;
+}
+
+.content {
+    display: grid;
+    grid-gap: 3rem 1rem;
+    grid-template-columns: repeat(auto-fit, 6rem);
+    justify-content: center;
+}
+
+.item {
+    display: block;
+    line-height: 1rem;
+    text-align: center;
+    width: 6rem;
+}
+
+.imageWrapper {
+    border-radius: 50%;
+    border: 1px solid rgb(var(--venia-border));
+    display: block;
+    height: 5rem;
+    margin: 0 auto 1rem auto;
+    overflow: hidden;
+    padding: .5rem;
+    width: 5rem;
+}
+
+.image {
+    height: auto;
+    max-height: 100%;
+    max-width: 100%;
+    width: auto;
+}
+
+.name {
+    display: block;
+}
--- a/packages/venia-concept/src/components/CategoryList/categoryList.js
+++ b/packages/venia-concept/src/components/CategoryList/categoryList.js
@@ -19,12 +19,6 @@ const categoryListQuery = gql`
     }
 `;
 
-// TODO: get baseUrl from graphql when it is ready
-const baseUrl = 'https://magento-venia.local.pwadev:8000';
-
-// TODO: get mediaUrl from graphql when it is ready
-const mediaUrl = 'https://magento-venia.local.pwadev:8000/pub/media';
-
 // TODO: get categoryUrlSuffix from graphql when it is ready
 const categoryUrlSuffix = '.html';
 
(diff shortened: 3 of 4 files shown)


packages/venia-concept/src/RootComponents/CMS/CMS.js
```diff
@@ -1,7 +1,13 @@
 import { createElement, Component } from 'react';
+import Page from 'src/components/Page';
+import CategoryList from 'src/components/CategoryList';
 
 export default class CMS extends Component {
     render() {
-        return <div>CMS Page Stub</div>;
+        return (
+            <Page>
+                <CategoryList title={'Shop by category'} id={2} />
```
zetlen: This is a small thing, but JSX allows you to use literal strings as attribute values without the expression braces. You have to use double quotes, so this could be:
```jsx
<CategoryList title="Shop by category" id={2} />
```
Soon we will create an internationalization system that finds and replaces these strings, so they should use a common syntax.

packages/venia-concept/src/components/CategoryList/categoryList.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.title {
+    text-transform: uppercase;
+}
+
+.content {
+    display: grid;
+    grid-gap: 3rem 1rem;
+    grid-template-columns: repeat(auto-fit, 6rem);
```
zetlen: @jimbo would give better advice here, but I think you might want to use a `fr` unit instead of `6rem`, if you want a predictable number of items in each row.

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -0,0 +1,105 @@
+import { Component, createElement } from 'react';
+import { string, number, shape } from 'prop-types';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import classify from 'src/classify';
+import defaultClasses from './categoryList.css';
+
+// TODO: get only active categories from graphql when it is ready
```
zetlen: Thank you for adding this TODO comment. This really helps everyone. :+1:

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -0,0 +1,105 @@
...
+                url_key
+                url_path
+                image
+            }
+        }
+    }
+`;
+
+// TODO: get baseUrl from graphql when it is ready
+const baseUrl = 'https://magento-venia.local.pwadev:8000';
```
zetlen: You should not need to hardcode the full domain! Relative paths from root should be fine. Are you working with Luma sample data?

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -0,0 +1,105 @@
...
+
+// TODO: get mediaUrl from graphql when it is ready
+const mediaUrl = 'https://magento-venia.local.pwadev:8000/pub/media';
```
zetlen: Same here; this variable should probably be `pub/media`.

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -0,0 +1,105 @@
...
+        title: string,
+        classes: shape({
+            root: string,
+            header: string,
+            content: string,
+            title: string,
+            item: string,
+            imageWrapper: string,
+            image: string,
+            name: string
```
zetlen: Excellent work implementing a public CSS classes API.

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -0,0 +1,105 @@
...
+        })
+    };
+
+    static defaultProps = {
+        id: 2
```
zetlen: This component probably shouldn't have a default ID. I'm guessing you made this for easy testing, but it should be removed.

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    render() {
+        const { id, classes } = this.props;
+
+        return (
+            <div className={classes.root}>
+                {this.props.title && (
```
zetlen: We have developed a pattern for nullable props that should conditionally render a portion of JSX. (@jimbo and I should probably document this pattern). This could be another component, but it seems too single-purpose; our pattern is to create a class getter that encapsulates this logic. An example [here](https://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/components/MiniCart/product.js#L72) shows a property `{this.options}` that you can simply embed in your JSX with no conditional, because the conditional is implemented [in the getter](https://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/components/MiniCart/product.js#L38-L51).

Here, you should add a class method:
```jsx
get header() {
  const { title, classes} = this.props;
  return title ? (
    <div className={classes.header}>
        <h2 className={classes.title}>
            <span>{title}</span>
        </h2>
    </div>
  ) : null;
}
```

And then in your render method right here, you replace this expression with
```jsx
<div className={classes.root>
  {this.header}
  <Query query={categoryListQuery} variables= {{ id }}>
```

(We recommend a getter instead of a plain accessor function, because a getter prevents you from passing arguments internally, which could put logic where it doesn't belong. A getter ensures that the method must have the same signature as `render()`, so that it could easily be moved into another component if necessary.)

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -0,0 +1,105 @@
...
+                <Query query={categoryListQuery} variables={{ id }}>
+                    {({ loading, error, data }) => {
+                        if (error) return <div>Data Fetch Error</div>;
+                        if (loading) return <div>Fetching Data</div>;
+                        if (data.category.children == '')
+                            return <div>Here are not any child categories</div>;
+
+                        return (
+                            <div className={classes.content}>
+                                {data.category.children.map((item, index) => (
```
zetlen: How would you feel about implementing the inner function of `.map` here as another Component instead of inline JSX? It's not mandatory, but it's something to think about.

packages/venia-concept/src/components/CategoryList/categoryList.css
artKozinets: We have mobile, tablet, desktop view.
My idea was to have fixed width of items + margin between them.
Center aligned, no matter which device we are talking about.

screen shot 2018-07-26 at 17 39 51
screen shot 2018-07-26 at 17 40 23

If you have some other solution please let me know.

packages/venia-concept/src/components/CategoryList/categoryList.js
artKozinets: Idea was to have preset of the default magento category.
"Root Catalog" - id="1"
"Default Category" - id="2"

packages/venia-concept/src/components/CategoryList/categoryList.js
artKozinets: Yep, I use magento 2.3-develop + sampledata

packages/venia-concept/src/components/CategoryList/categoryList.js
artKozinets: Yep, awesome idea.

zetlen:
@artKozinets Changes look great and I'm ready to merge. Can you please sign the Contributor License Agreement posted [above](https://github.com/magento-research/pwa-studio/pull/171#issuecomment-408454136)?

artKozinets:
@zetlen Done.

packages/venia-concept/src/components/CategoryList/categoryList.css
jimbo: The approach used here achieves the design they're aiming for, so it looks good to me. Thanks for embracing grid. 👍

packages/venia-concept/src/components/CategoryList/categoryList.js
jimbo: I know we've merged this already, but yes—in general we try to avoid performing work inside JSX expressions. Eventually, I'd like to cover that with a lint rule. Certainly not a dealbreaker, though, so thanks for your understanding. 😄

Issue #8, opened 2018-06-13
PWADevServer should be configurable to skip trust settings and hostfile update


PWAStudioBot:
_From @zetlen on May 16, 2018 20:41_

PWADevServer attempts to ensure OS-level trust of its managed SSL certificates, using [@magento/devcert](https://github.com/magento-research/devcert). There may be cases, especially in non-interactive build environments that still run `development` mode, where we want to skip trust settings.

PWADevServer also attempts to update the hostfile, which requires elevated permissions. This should also be skippable.

Steps to reproduce:

1. Add `PWADevServer` to a webpack.config file.
2. Run webpack in a non-interactive process.

Expected results: Dev mode initializes.
Actual results: Hangs forever during PWADevServer instantiation.

_Copied from original issue: magento-research/pwa-buildpack#36_

ericerway:
Review carried to Sprint 22.

Issue #5, opened 2018-06-13
PWADevServer should automatically detect EADDRINUSE port conflicts


PWAStudioBot:
_From @zetlen on April 25, 2018 0:17_

Right now, PWADevServer ensures unique domains per project only at provisioning time. It should also ensure unique ports at startup time.

_Copied from original issue: magento-research/pwa-buildpack#25_

zetlen:
Will be resolved by #175.

PR #175, opened 2018-07-25
feat(dev): Make PWADevServer host/SSL optional

zetlen:
PWADevServer attempts to create a unique domain name, create local
SSL certificates, and tell the OS to trust those certificates, for
every new project. It also tries to confirm they exist on every run!

This causes many problems for users under some common conditions:
- No administrative access to local machine
- No OpenSSL installed, or wrong OpenSSL installed
- OS cannot be scripted to trust certificates
- Developer uses Firefox, which uses its own cert store

Additionally, some bugs in the implementation have caused some
developers' projects to enter an unusable state.

This change makes those features optional and adds clearer documentation around them.

## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

- Adds `provideUniqueHost` flag to PWADevServer configuration.
 PWADevServer will no longer try to create or retrieve a custom domain
 name unless `provideUniqueHost` is in its configuration in
 `webpack.config.js` as either a custom string or `true`.
- Adds `provideSSLCert` flag to PWADevServer configuration. PWADevServer
will no longer try to create or retrieve a trusted SSL certificate
unless `provideSSLCert: true` is in its configuration in
`webpack.config.js`.
- Modifies custom domain name creation strategy to ensure uniqueness
based on a hash of the full local path, rather than using the local
flat file database.

## Additional information
We created these features for the needs of the developer working on
several PWAs at once on their local machine, so that they don't have to
set up manual SSL every time, and they have no conflicts with Service
Workers. This could be considered "bonus functionality", as it's not
critical to the setup of a minimum viable PWA. It was meant to establish
our focus on developer experience, and articulate the parts of developer
setup that PWA Studio can "own".

*However, we soon learned that we could not maintain all scenarios for
automated setup and continue to make progress with shopper-facing
features*. We still really want to support and automate all of these
scenarios, but for now, our implementations are a hindrance and we are
turning them off by default.

### DIFF
--- a/packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
+++ b/packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
@@ -1,5 +1,6 @@
 const debug = require('../util/debug').makeFileLogger(__filename);
 const { join } = require('path');
+const { createHash } = require('crypto');
 const url = require('url');
 const express = require('express');
 const GlobalConfig = require('../util/global-config');
@@ -14,17 +15,14 @@ const { lookup } = require('../util/promisified/dns');
 const { find: findPort } = require('../util/promisified/openport');
 const runAsRoot = require('../util/run-as-root');
 const PWADevServer = {
+    DEFAULT_NAME: 'my-pwa',
+    DEV_DOMAIN: 'local.pwadev',
     validateConfig: optionsValidator('PWADevServer', {
-        id: 'string',
         publicPath: 'string',
         backendDomain: 'string',
         'paths.output': 'string',
         'paths.assets': 'string',
         serviceWorkerFileName: 'string'
-    }),
-    hostnamesById: new GlobalConfig({
-        prefix: 'devhostname-byid',
-        key: x => x
     }),
     portsByHostname: new GlobalConfig({
         prefix: 'devport-byhostname',
@@ -61,12 +59,12 @@ const PWADevServer = {
         }
     },
     async findFreePort() {
-        const inUse = await PWADevServer.portsByHostname.values(Number);
-        debug(`findFreePort(): these ports already in use`, inUse);
+        const reserved = await PWADevServer.portsByHostname.values(Number);
+        debug(`findFreePort(): these ports already reserved`, reserved);
         return findPort({
             startingPort: 8000,
             endingPort: 9999,
-            avoid: inUse
+            avoid: reserved
         }).catch(e => {
             throw Error(
                 debug.errorMsg(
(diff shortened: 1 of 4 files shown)


packages/pwa-devdocs/src/pwa-buildpack/reference/pwa-dev-server/index.md
```diff
@@ -6,25 +6,41 @@ A utility for configuring a development OS and a `webpack-dev-server` for PWA de
...
-* Customizes the `webpack-dev-server` instance to:
-    * Proxy all asset requests not managed by webpack to the Magento store
-    * Emulate the public path settings of the Magento store
-    * Automatically switch domain names in HTML attributes
-    * Debug or disable ServiceWorkers
+## Basic Features
+PWADevServer creates a `devServer` which is optimized for a Magento API-backed
+PWA development workflow, including:
+
+* **Hot reload in the browser whenever you save a change**
```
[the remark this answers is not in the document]
zetlen: Slightly changed this because HMR does both the refresh and the inline replacement.

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=175) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=175) before we can accept your contribution.<br/><hr/>**James Zetlen** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

zetlen:
@jcalcaben Revised my docs per your comments. Thank you! I like them much better now.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -75,84 +73,97 @@ const PWADevServer = {
...
+                        `package.json has ${pkg.name} "name" field!`
+                    );
+                }
+                name = pkg.name;
+            } catch (e) {
+                console.warn(
                     debug.errorMsg(
-                        `findFreeHostname: Unable to find a free hostname after 9 tries. You may want to delete your database file at ${GlobalConfig.getDbFilePath()} to clear out old developer hostname entries. (Soon we will make this easier and more automatic.)`
-                    )
+                        `createUniqueSubdomain(): Using default "${name}" prefix. Could not autodetect theme name from package.json`
```
jimbo: The name of this method appears to be `getUniqueSubdomain`, not `createUniqueSubdomain`.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -75,84 +73,97 @@ const PWADevServer = {
...
-            debug(`findFreeHostname: ${maybeHostname} bound to port`, exists);
-            if (times > 9) {
-                throw Error(
+            const pkgLoc = join(process.cwd(), 'package.json');
+            try {
+                // eslint-disable-next-line node/no-missing-require
+                const pkg = require(pkgLoc);
+                if (!pkg.name) {
+                    throw new Error(
+                        `package.json has ${pkg.name} "name" field!`
```
jimbo: Could we rewrite this error to be a little more clear?

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -75,84 +73,97 @@ const PWADevServer = {
...
+        const digest = dirHash.digest('base64');
+        // Base64 truncated to 5 characters, stripped of special characters,
+        // and lowercased to be a valid domain, is about 36^5 unique values.
+        // There is therefore a chance of a duplicate ID and host collision,
+        // specifically a 1 in 60466176 chance.
+        return `${name}-${digest.slice(0, 5)}`
+            .toLowerCase()
+            .replace(/[^a-zA-Z0-9]/g, '-')
+            .replace(/^-+/, '');
+    },
```
jimbo: Nice and clear. 👍

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -175,6 +186,33 @@ const PWADevServer = {
                 );
             }
         };
+        let devHost;
+        if (config.id) {
+            devHost = await PWADevServer.provideCustomHost(config.id);
+        } else if (config.provideUniqueHost) {
+            devHost = await PWADevServer.provideUniqueHost(
+                config.id || config.provideUniqueHost
```
jimbo: Wouldn't this statement always evaluate to `true`, given the conditions?

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
zetlen: Perils of using static refactoring tools. Thanks!

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
zetlen: The way I've structured this, that error gets immediately caught, and then handled along with any other errors that might occur trying to `require()` the package.json file. So the errors that throw will look like:

Error loading package.json:
```
getUniqueSubdomain: Using default "my-pwa" prefix. Could not autodetect theme name from package.json Error: cannot find module './package.json'`
```

package.json exists but somehow has no "name" field:
```
`getUniqueSubdomain: Using default "my-pwa" prefix. Could not autodetect theme name from package.json Error: package.json has undefined "name" field!`
```

I can add a comment explaining this if you like, but I think the errors that emit are clear enough.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
zetlen: Thanks, good catch.

zetlen: It's actually possibly a string. But I changed the behavior so that `config.id` and `config.provideUniqueHost` are mutually exclusive (and `config.id` is just preserving legacy behavior), so I don't need the first part of that conditional./

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
jimbo: No, my concern was that if `pkg.name` is falsy but not `undefined`, the sentence won't make much sense, since this assumes it's an adjective. 😅 

If `undefined` is the only possible falsy value for `pkg.name`, then this doesn't need to be a template string. Just write `has undefined "name" field`. Otherwise, let's make it a bit more flexible.

```
// pkg.name: ''
package.json has "name" field!

// alternative
package.json is missing a "name" value!
```

zetlen: I had not considered the null string. Changing!

zetlen:
@jimbo Fixed that error message for ya.

zetlen:
Closes #8.

zetlen:
Closes #5.

zetlen:
Went ahead and merged this to take some work of Jimmy's plate--it's already been extensively reviewed.

PR #186, opened 2018-07-30
style(css): Configure prettier to format CSS

jimbo:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, Prettier will format CSS (in addition to JS, as it currently does).

## Additional information

Currently, CSS is neither linted nor formatted. As a result, our files have a mixture of two-space and four-space indentation, single quotes and double quotes, etc. Prettier is currently not formatting these files because we have explicitly instructed it to only format JS files. This PR changes that configuration to include CSS files as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,8 +21,8 @@
     "test": "jest",
     "test:debug": "node --inspect-brk node_modules/.bin/jest -i --watch",
     "test:dev": "jest --watch",
-    "prettier": "prettier --write 'packages/**/*.js' '*.js'",
-    "prettier:check": "prettier --list-different 'packages/**/*.js' '*.js'"
+    "prettier": "prettier --write 'packages/**/*.@(js|css)' '*.js'",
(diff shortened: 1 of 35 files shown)


jimbo:
I would consider `prettier.config.js` to be authoritative here; `.editorconfig` should be a reflection of the same settings.

Prettier does not have CSS-specific options. We could override our current option values, but I don't see any reason to do that. If our formatter is set to use four-space indentation and single quotes, it should do so for both JS and CSS files.

jimbo:
And for full disclosure, my own current personal preference would be two-space indentation and double quotes in all files. However, this project uses four-space and single quotes. This PR simply enforces the existing settings for consistency.

zetlen:
I'm in favor. In a later PR we should set these things authoritatively from one file. We could do it in the prettier config a little more easily.

PR #204, opened 2018-08-02
Create new topic for global style guide

### LINKED ISSUE #169: Venia Global Styles documentation

[ ] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-module`
[X] `pwa-devdocs`

## This issue is a:

[ ] Bug
[ ] Feature suggestion
[X] Documentation issue
[ ] Other (Please Specify)

## Description

As a Venia component developer, I want to read documentation that lists the specific style values the Venia theme uses to create a component that fits within the Venia theme.

## Expected result:

A developer is able to use a global style reference topic to create Venia-themed components.

## Possible solutions:

Create a new Global Styles documentation topic.

jcalcaben:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[X] Documentation

## Summary

When this pull request is merged, it will create the Global Style Guide topic for Venia.

## Additional information

This PR closes #169.


soumya-ashok:
- The section at the top with the title "General Rules" could use a different title. Maybe "Global Rules" or "Global Requirements"?
- This is a long page with sub-headings. Maybe a list of a-links at the top in case the reader wants to navigate to or is looking for a specific section?
- Colors section, I'd put the tip at the top since that would be useful to know right-off
- Fonts - I'd specify that it is an open source Google font, and include a screenshot that shows the weight range
- Modify "Root font size" to say "Size Specifications" and have it look more like a sub heading under "Fonts"?
- "The Venia theme uses a limited set of font weights in its design. Do not use anything other than the following weights" - Maybe reword this to say "We recommend using no more than 3 font weight variations for optimal performance" and put this in a green tip? Continue with "Venia uses the following :"?
- For Grid layouts, I'd recommend providing a visual that shows the measurements rather than putting it in a sentence. Would be much easier to grasp immediately - I can provide this if needed
- Animation specs could be in a table with two columns - type and duration? or something like that
- Icons - would include a screenshot here

soumya-ashok:
- Is the hierarchy of the page topics optimal? I would think that Layout should precede Colors or Fonts.
- Could we re-title the heading "Font" to "Typography"?
- Under size specifications, could we not include the root font size in the table? Since the table is the most obvious visually, people may miss reading the content or ignore it thinking it is an introduction. And instead of presenting examples, could we not just list them as pixel conversions of the root size?
- For the font screenshot, could something like the attached be more valuable as a reference? 
screen shot 2018-08-08 at 2 48 07 pm

soumya-ashok:
- The placement of the note for borders is at the bottom of the section, and for the other topics is at the top. It would be better to be consistent in the placement of notes within a section
- The Z index note seems like it should have the note treatment since it is an important standard?
- "Use the following guidelines to make component animations feel efficient and quick." I would like the reason why we're crafting animations in this way to be because it is optimal for the context. Sometimes, like in the case of confirmation messages or checkmarks, a 'quick' animation may not be the best choice to call attention, so I would rather not use these qualifying words. If we are able to present a more relevant justification, then we can word it that way.
- On Icons, perhaps we should present the specs first followed by the examples?

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
```diff
@@ -0,0 +1,151 @@
...
+| CSS class         | RGB value                     | Usage            |
+| ----------------- | ----------------------------- | ---------------- |
+| `black`           | `rgb(0, 0, 0)`                |                  |
+| `venia-border`    | `rgb(224, 224, 224)`          | Border color     |
+| `venia-grey`      | `rgb(246, 246, 246)`          |                  |
+| `venia-teal`      | `rgb(0, 134, 139)`            | Buttons          |
+| `venia-text`      | `rgb(33, 33, 33)`             | General text     |
+| `venia-text-alt`  | `rgb(117, 117, 117)`          | Alternative text |
+| `venia-text-hint` | `rgb(158, 158, 158)`          | Hint text        |
+| `mask.root`       | `rgb(0, 0, 0)` at 50% opacity | Mask overlay     |
```
jimbo: `mask.root` is not a global style, nor is it a classname that people can apply, so we should consider removing it from this list for now. In the future, though, we could capture part of this style as a custom property and update this document at that time.

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
```diff
@@ -0,0 +1,151 @@
...
+## Colors
+
+{: .bs-callout .bs-callout-tip}
+***Tip:** Color values are expressed in RGB format because it translates easier to `rgba()` values than HEX format.*
+
+The Venia theme uses the following color classes in its design:
+
+| CSS class         | RGB value                     | Usage            |
+| ----------------- | ----------------------------- | ---------------- |
+| `black`           | `rgb(0, 0, 0)`                |                  |
```
jimbo: `black` might be a global style, but unlike all of these others, it's not a custom property. I'm not sure I'd put it in this list.

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
```diff
@@ -0,0 +1,151 @@
...
+| `venia-border`    | `rgb(224, 224, 224)`          | Border color     |
```
jimbo: We should consider prefixing all of these with `--`, since that's how they're named as custom properties.

```css
.foo {
  border-color: rgb(var(--venia-border));
}
```

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
```diff
@@ -0,0 +1,151 @@
...
+| `border-width`  | `1px` | The width of an element's border            |
+{:style="table-layout:auto"}
+
+{: .bs-callout .bs-callout-info}
+***Note:** For buttons, the `border-radius` is half the height of the button.
+For example, a button with a height of `3rem` should have a `border-radius` of `1.5rem`.*
+
+### Layers
+
+Regarding layers, do not use the Z-Index to re-order items within a layer unless it is absolutely necessary.
```
jimbo: Probably lowercase and backtick `z-index` here. Maybe reword a bit, too?

```
Avoid using `z-index` to re-order items; instead, rely on document order
where possible. If `z-index` must be used, ensure that it is only used to
re-order sibling elements within a single layer.
```

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Animations
+
+Use the following guidelines to make component animations feel efficient and quick.
+
+| Animation property                      | Value                                   |
+| --------------------------------------- | --------------------------------------- |
+| Speed                                   | 8-16 milliseconds or 1 frame per second |
```
jimbo: I'm confused by this line. `16ms` does correspond to roughly `1 fps`, but it's not a value that people should use. Maybe suggest in the paragraph above that people should use increments of `16ms`?

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
```diff
@@ -0,0 +1,151 @@
...
+| `border-radius: 1.5rem;` | `border-radius: 24px;` |
+| `font-size: 0.75rem;`    | `font-size: 12px;`     |
+{:style="table-layout:auto"}
+
+### Font weights
+
+The Venia theme uses a limited set of font weights in its design.
+For optimal component performance, use no more than 3 font weight variations.
+
+![Muli font weights][]{:width="350px"}
```
jimbo: This shows all of the font weights. Is there a way to only show the three we use?

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
brendanfalkowski: Text color variables that are shades of pure gray should probably use `rgba(0, 0, 0, 0.5)`. Otherwise the middle grays tend to wash out against a non-white background as it darkens, but with a black-based `rgba()` they'll get darker and retain relative contrast.

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
brendanfalkowski: Not sure it's practical to have 600+700, but no italic style. The budget should change or I'd drop the 600 to suit the common bold/700 need.

Typeface weights aren't that unique, so this could be a pain to change across projects if Venia chooses a non-700 weight for the "bold" visual style.

soumya-ashok: Thanks for the comment @brendanfalkowski we will revisit the specs.

jimbo: I think `600` is generally the proper weight for semibold UI text, regardless of the font. It's at least true for every font in our font family. `700` is too strong for UI text.

Initially, I had suggested we only use two styles: normal `400` and semibold `600`. `700` is only included here because it may be useful for overlaying hero images. The Venia designs don't call for italics, and I don't anticipate that they will.

That said, CMS content will probably expect additional styles to be available, including italics. Perhaps we can load additional styles asynchronously. That's how a PWA should do it anyway. 😅

packages/pwa-devdocs/src/venia-pwa-concept/design/global-style/index.md
jimbo: As you might expect, these grays are actually pre-blended tints of black. The real data here would be opacity.

opacity | blended on white
--- | ---
`0.87` | `rgb(33, 33, 33)`
`0.54` | `rgb(117, 117, 117)`
`0.38` | `rgb(158, 158, 158)`

We don't use `rgba()` for text because of how text rendering works in browsers—text appears thinner when it's not fully opaque—but that does introduce the problem you pointed out, which is that the text washes out against non-white backgrounds. Ordinarily, the solution would be to set the text color when setting the background color:

```stylus
.foo
    background-color: rgb(236, 232, 245)
    color: blend(rgba(black, 0.87), @background-color)
```

And that's normally what I've done, but without a preprocessor, I'm not sure how to accomplish it.

brendanfalkowski: Thinning is a good point. I haven't worried about this as much since the effect isn't as pronounced with text > 16px and @2x screens becoming the norm. Made a CodePen to test, and it does look a bit bad on @1x screens:

https://codepen.io/brendanfalkowski/pen/djwNpy

To address wash-out, in design systems I've built there's usually a concept of "tones" as a token. They're not specific to components, but they makeup the majority of component coloration. The component applies a tone rather than defining hex/rgb colors.

- color-background
- color-border
- color-text-base
- color-text-subtle

screen shot 2018-08-09 at 12 58 49 pm

That type of organization helps avoid the mishaps of consistently changing text styles when the background changes.

Doesn't help doing color math without a preprocessor though.

PR #212, opened 2018-08-06
refactor(redux): Reorganize actions and reducers

jimbo:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, we'll once again use `redux-actions` to create reducers and actions. This will prepare us for incoming changes related to extensibility. It will also make them far more consistent.

## Additional information

Checkout components have been refactored to accommodate new app state structure. They should now be ready for editable forms to be added (and real cart modification requests to be made).


packages/venia-concept/src/actions/app.js
```diff
@@ -1,15 +1,30 @@
...
-export const closeDrawer = () => toggleDrawer(null);
+import { createActions } from 'redux-actions';
+
+import { store } from 'src';
+
+const prefix = 'APP';
+const actionTypes = ['TOGGLE_DRAWER'];
+
+const actions = createActions(...actionTypes, { prefix });
+export default actions;
```
jimbo: This is a standard way to create both action types (string constants that reducers look for) and their corresponding boilerplate action creators. Almost all sync action creators have a sort of identity signature:

- It returns an action object
- It defines a a `payload` prop on that object, whose value is the first argument
- If `payload` is an `Error`, it defines an `error` prop, whose value is `true`

[The docs suggest](https://redux.js.org/recipes/reducingboilerplate) that exporting action types and action creators can reduce boilerplate and keep action logic centralized. Here, the action creators generated by `createActions` are named such that they can act as action types.

```js
actions.toggleDrawer.toString(); // => 'APP/TOGGLE_DRAWER'
```

packages/venia-concept/src/actions/app.js
```diff
@@ -1,15 +1,30 @@
...
+        try {
+            const reducers = await Promise.all(payload);
+
+            reducers.forEach(({ default: reducer, name }) => {
+                store.addReducer(name, reducer);
+            });
+        } catch (error) {
+            console.log(error);
+        }
+    };
```
jimbo: Async action creators aren't actually action creators at all. They don't return an action object with a `type`, and they aren't handled by reducers. They're just higher-order functions useful for orchestrating dispatches.

Here, `loadReducers` doesn't even dispatch any actions. Should we still call it an async action creator? or a helper? Is there a difference, and does it matter?

packages/venia-concept/src/actions/cart.js
```diff
@@ -1,42 +1,61 @@
...
+    'UPDATE_DETAILS'
+];
+
+const actions = createActions(...actionTypes, { prefix });
+export default actions;
+
+/* async action creators */
 
 const { request } = RestApi.Magento2;
+const storage = new BrowserPersistence();
```
jimbo: Since actions are similar to events, a reducer can seem like a good place to "listen" for "events" and perform side effects such as local persistence. In fact, though, one of the few firm rules in Redux says that [reducers should perform no side effects](https://redux.js.org/basics/reducers).

> Things you should **never** do inside a reducer:
>
> - Mutate its arguments;
> - Perform side effects like API calls and routing transitions;
> - Call non-pure functions, e.g. `Date.now()` or `Math.random()`.

So I had to migrate the `storage` logic from the cart reducer here to the cart actions. It's largely unchanged. However, an alternate way to handle this logic on the "listening" side would be to put it in a middleware.

packages/venia-concept/src/actions/cart.js
```diff
@@ -163,21 +168,60 @@ async function fetchCartPart({ guestCartId, forceRefresh, subResource = '' }) {
...
+        }
+
+        // otherwise create a guest cart
         await dispatch(createGuestCart());
     }
 
-    // retrieve app state again
+    // retrieve the new guest cart from state
     return getState().cart.guestCartId;
 }
```
jimbo: Here, we have another example of the ambiguity around async action creators.

- An async action creator is a function that receives a payload and returns a thunk
- A thunk is a function that receives `dispatch` and `getState`, performs side effects, and returns `undefined`

Here, we have a function that's essentially a naked thunk—it's not the product of a higher-order function—and it both performs side effects (using `dispatch`) and returns a value (using `getState`). Currently, it's exclusively used by async action creators, since it's of little use to components due to its shape. What should we do with this function?

packages/venia-concept/src/actions/checkout.js
```diff
@@ -1,80 +1,101 @@
...
+    ORDER: {
+        SUBMIT: null,
+        ACCEPT: null,
+        REJECT: null
+    }
 };
 
-const enterSubflow = () =>
+const actions = createActions(actionMap, ...actionTypes, { prefix });
+export default actions;
```
jimbo: `createActions`, provided by `redux-actions`, has a [wonderful, optional API](https://redux-actions.js.org/api/createaction#createactionsactionmap-identityactions) that allows us to nest action creators. While it's a bit overkill for most action types, I think it's perfect for action types associated with API requests, which often fit a request-success-error pattern.

packages/venia-concept/src/reducers/checkout.js
```diff
@@ -1,70 +1,85 @@
...
+
 const initialState = {
-    shippingInformation: false,
-    status: 'READY',
-    subflow: null
+    editing: null,
+    step: 'cart',
+    submitting: false,
+    valid: false
 };
```
jimbo: Storing validity here as a simple boolean, `valid`, is definitely an oversimplification. This will get much more complex as we start to track client- and server-side validation of various parts of the form and the workflow.

packages/venia-concept/src/actions/checkout.js
zetlen: Yeah, let's go with prior art here. There should be simple recipes for the sequence of action dispatch API calls and other common interactions (say, additional logic loaded via a chunk?), like:

```js
import { lifecycle } from '@magento/peregrine';

const cartLifecycle = lifecycle('CART');

const cartApiLifecycle = cartLifecycle.requests({
    GET_DETAILS(state) {
       return {
         path: `/guest-carts/${state.guestCartId}`
      }
    },
    ADD_ITEM(state, item) {
       return {
          path: `/guest-carts/${state.guestCartId}/items/`,
          method: 'POST',
          body: item
      }
  },
  async (path, method, body) => M2ApiRequest.request(path, { method, body })
);

const [getDetails, addItem] = cartApiLifecycle.getActions();

const reducer = cartApiLifecycle.getReducer();

```

We don't need to make those abstractions yet, of course.

packages/venia-concept/src/components/Checkout/flow.js
```diff
@@ -1,79 +1,76 @@
...
+                    submitting,
+                    valid
+                };
+
+                return <Form {...stepProps} />;
             }
-            case 'STEP_3': {
-                child = <Exit resetCheckout={resetCheckout} />;
-                break;
+            case 'receipt': {
```
zetlen: This seems like a mismatch. Why's it here in this otherwise very consistent flow, which sets `child` and identifies itself as a step number?

packages/venia-concept/src/index.js
```diff
@@ -23,17 +23,4 @@ ReactDOM.render(
     document.getElementById('root')
 );
 
-if (process.env.SERVICE_WORKER && 'serviceWorker' in navigator) {
```
zetlen: This change seems unrelated.

packages/venia-concept/src/reducers/checkout.js
zetlen: Yeah, I think that reducers should be autogenerated at the same time as actions, see comment above.

packages/venia-concept/src/index.js
jimbo: Good catch. This shouldn't have been included in the PR.

jimbo: Fixed. 👍

packages/venia-concept/src/components/Checkout/flow.js
jimbo: Fixed. 👍

packages/venia-concept/src/actions/app.js
jimbo: When a reducer is added, the store actually replaces its root reducer, which instructs connected components to re-render (like an action does). And there's nothing that a reducer should do in response to `REDUCERS_ADDED`; even updating the list of current slices is pointless, since `getState()` would be the source of truth for that info.

I think we can improve the reducer-adding workflow—turn it into a pipeline, maybe—as we open up to extensibility. But for now I think this is actually fine.

Issue #30, opened 2018-06-13
root.phtml includes hardcoded reference to pub/static


PWAStudioBot:
_From @fooman on May 9, 2018 0:22_

Magento can be run with either 
1. the pub/ folder as the document root (most should use this one)
2. or from the Magento root folder.

Using the hardcoded `<script src="/pub/static/frontend/Magento/venia/en_US/client.js"></script>` in root.phtml assumes 2.

`<script src="<?= $this->getViewFileUrl('client.js')?>"></script>` will likely work better. It resolves to either option and also respects the versioning setting.

_Copied from original issue: magento-research/venia-pwa-concept#63_

PWAStudioBot:
_From @fooman on May 12, 2018 10:43_

Using getViewFileUrl() does not work unfortunately. There is a disconnect between what is the URL of the M2 instance and what is provided by node, currently hardcoded to https://magento-venia.local.pwadev:8000. In my trials I have for now resorted to using `<script src="https://magento-venia.local.pwadev:8000/static/frontend/Magento/venia/en_US/client.js"></script>`.

PWAStudioBot:
_From @erfanimani on May 26, 2018 1:39_

I was going to make a pull request with exactly the same fix you provided, and then stumbled upon this ticket. I still think using `getViewFileUrl` is much better than it is currently (because pretty much everyone will run into this issue). The only reason to link to node instead of the server Magento is running off, is the hot-reloading stuff right?

Additionally, running:

```
./node_modules/webpack/bin/webpack.js \
    --progress \
    --color \
    --env.phase development && find . | grep client\.js
```

for me outputs

```
./web/js/client.js
./web/js/client.js.map
```

Meaning that instead of referencing `client.js`, it should be referencing `js/client.js`, or am I missing something?

edit: to be clear, `<script src="<?php echo $this->getViewFileUrl('js/client.js') ?>"></script>` fixed the client.js 404 for me.

PWAStudioBot:
_From @erfanimani on May 26, 2018 10:39_

Just had a chat in CommEng, and disregard my previous comment. It seems you need to run Venia from the nodeJS server, which then proxies **some** requests through to the Magento backend, not the other way around (which I think is what you're doing @fooman?). Because client.js is provided by the proxy, it doesn't matter what the URL is. Whether you're running from `pub` or not shouldn't matter. Correct me if wrong..

PWAStudioBot:
[...]

PR #219, opened 2018-08-10
feat(dev): app shell detects env from proxy header

zetlen:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will enable the application shell to emit different script tags when in development vs production mode. When production mode is detected, it emits a script tag for the vendor module. When development mode is detected, it omits this tag. Production mode can work without 

## Additional information

This is a temporary hack; it shouldn't be necessary once app shells are generated on build. For now, it requires disabling of `page_cache` in admin; otherwise, you must `bin/magento cache:flush` every time you switch between dev and prod mode.


mhhansen:
**relates #30** 

I think will be useful to update this PR with the PHP view file call,
so it doesn't fire up another issue regarding `hardcoded references`
give a shout if want a hand updating this PR.

```
<?php echo $this->getViewFileUrl('js/vendor.js'); ?>
```

zetlen:
@mhhansen Sorry I've become so behind on PRs while I was working on the UPWARD spec. Have you tried `$this->getViewFileUrl('js/vendor.js')` and made it work in `root.phtml`? Since there is no actual block instance in the root file, we didn't have access to that method when we tried.

mhhansen:
No worries! :)
Yes, I tried indeed.. while I was doing some research for [issue #30](https://github.com/magento-research/pwa-studio/issues/30)
Worked for both `client.js` and `vendor.js` files, which are bundled/created at the same level.

That's also why I've created [PR#246](https://github.com/magento-research/pwa-studio/pull/246) for the [issue #30](https://github.com/magento-research/pwa-studio/issues/30) (left a bunch of research in there too, in case it helps others).

This PR will hit the `root.phtml` too, so either #246 needs to be merged/closed first and then this one on top.. or maybe it's better to just leave this one alive, update it with the `getViewFileUrl()` (to avoid introducing hardcoded references) and close/cancel #246.

LMK if you want me to do anything else (tests/changes and even try to rebase latest `master` branch on top and run this tests again) related to this or any other PR/issue related.

### This PR + `getViewFileUrl()`
This is the result having `venia` built for production:
(I know im serving without SSL/TSL, my valetplus went crazy last night I've disabled it)

01 getviewfileurl-renders-ok
02 vendor-client-200-ok


## getViewFileUrl() method available
I think this method is available because `root.phtml` file ends up being an instance of `Magento\Framework\View\Result\Page` which has that method available.

## sidenote
there's a deep research with workbox precache file that I did few weeks ago within the #30
([view comment](https://github.com/magento-research/pwa-studio/issues/30#issuecomment-414388792)) point 2.

zetlen:
@mhhansen Thank you for your detailed research! I'm closing this in favor of #246.

PR #220, opened 2018-08-10
fix(dev): merge 'assets' and 'output' paths

zetlen:
Fix the discrepancy between production and development mode paths which causes one or both to 404 on scripts. Do so by simplifying setup to just a single output directory.

## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[x] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

- Remove the concept of the "assets" directory entirely, and set the
"output" directory to be just "web/".
- Remove the old logo asset from /web/ (pre-Venia rebrand!)
- Gitignore entire "web" directory
- Set JS files to output to 'web/js' subdirectory by adding directory
prefix to `output.filename` and `output.chunkFilename`
- Modify tests, documentation, and type expectations to remove
`paths.assets`


packages/venia-concept/webpack.config.js
```diff
@@ -42,8 +41,8 @@ module.exports = async function(env) {
         output: {
             path: themePaths.output,
             publicPath: process.env.MAGENTO_BACKEND_PUBLIC_PATH,
-            filename: '[name].js',
-            chunkFilename: '[name]-[chunkhash].js',
+            filename: 'js/[name].js',
+            chunkFilename: 'js/[name]-[chunkhash].js',
```
jimbo: This change necessitates a corresponding change in `root.phtml`, doesn't it? The browser `404`s on these requests until I update the hardcoded `script src` in that file.

Further, I'm not sure this is the best way to do this. Couldn't we handle this with paths in `.env`?

- Expect that `MAGENTO_BACKEND_PUBLIC_PATH` end with `/js`?
- Look for two "output path" env vars—one for dev, one for prod?

Also, how is this possible? This one asset, `logo.svg`, is output to the root instead of `/js`.

wat

paths

zetlen: The original intent of this PR was to require that you update `.env` and add `/js` to the `MAGENTO_BACKEND_PUBLIC_PATH` var, but I forgot to include that part, which is my fault.

However, `root.phtml` should now automatically generate view URLs, which should allow us to stop using that environment variable in the first place.

At any rate, this is a large enough change that it ought to be in the 1.x release, which I'm putting together right now.

PR #221, opened 2018-08-10
fix(dev): MagentoRootComponentsPlugin dedupe fix

zetlen:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[x] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

- Fix MagentoRootComponentsPlugin detection of entry points so that only
user-defined entry points are considered, and not entry points issued by
downstream plugins (this fixes resource duplication in named chunks)
- Add phase detection for MagentoRootComponentsPlugin so it uses the
right strategy based on development/production mode
- Configure UglifyPlugin to correctly strip out
MagentoRootComponentsPlugin imports hack


packages/pwa-buildpack/src/WebpackTools/plugins/MagentoRootComponentsPlugin/index.js
```diff
@@ -46,21 +47,37 @@ class MagentoRootComponentsPlugin {
...
+
+                // Top-level modules injected by a downstream "issuer" are not
+                // entry points.
+                let isEntrySimpleTest = mod => !mod.issuer;
+                if (this.phase === 'development') {
+                    // Unless we're in development mode. In that case,
+                    // the best we can do is test if the entry is in node_modules.
+                    isEntrySimpleTest = mod =>
+                        mod.resource.indexOf('/node_modules/') !== -1;
+                }
```
jimbo: Super minor quibble:

```js
const isEntrySimpleTest = this.phase === 'development'
    ? mod => mod.resource.includes('/node_modules/')
    : mod => !mod.issuer;
```

zetlen: Thanks, I think I'd been staring at that too long.

zetlen:
Hard merging to 0.x for momentum's sake. Every time I do this something breaks, but having 17 open pull requests is also broken. 🔮

PR #222, opened 2018-08-10
fix(perf): Optimize logo.svg

zetlen:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will turn `logo.svg` from a 404KB Illustrator-enhanced layered SVG into a 3KB web-optimized SVG, with no change in appearance.


brendanfalkowski:
No idea how this works in React-world, but in my Gulp build tools there's always a task that runs every PNG/JPG/SVG in the frontend assets through optimization tools, so this is never manual.

Aside: probably check your Illustrator export too. An SVG of that file size is likely embedding the font file inside it (probably against the license). Normally you convert text to outlines before exporting so that can't happen.

zetlen:
@brendanfalkowski Yeah, we should probably put https://github.com/rpominov/svgo-loader in. Does your build process use svgo?

brendanfalkowski:
I use:
https://www.npmjs.com/package/gulp-imagemin

Which wraps:
https://www.npmjs.com/package/imagemin

Which wraps core optimizers:
https://github.com/imagemin/imagemin-gifsicle
https://github.com/imagemin/imagemin-jpegtran
https://github.com/imagemin/imagemin-optipng
https://github.com/imagemin/imagemin-svgo

It's easier than pulling in specific optimizers, so you can toss a mixed folder of assets and can manage them all.

Issue #28, opened 2018-06-13
Add Support for multiple types of connectors to Magento 2 instance


PWAStudioBot:
_From @Jakhotiya on May 14, 2018 13:42_

This repository doesn't work well without vagrant. We need to add support for trying out this code without needing vagrant. On my local I had to fix many url problems to get venia work. 

I would like to help but can somebody point me in right direction

_Copied from original issue: magento-research/venia-pwa-concept#70_

PWAStudioBot:
_From @DrewML on May 14, 2018 15:7_

Hi @Jakhotiya - appreciate the feedback. Can you outline what problems you had attempting to set this up without Vagrant?

The (admittedly sparse) setup docs go through the Vagrant setup process since that seems to be the most common among m2 users, but most of us on the team are using this with Valet+ on our hosts.

PWAStudioBot:
_From @Jakhotiya on May 14, 2018 17:47_

### Environment setup
magento 2.3-develop branch
nodejs v9
nginx php7.1 fpm mysql stack
store url on my local 23-develop.to (Couldn't use url of the dev server provided with this repo)

### Setting up the module and theme
First I downloaded the code from github and moved the module to app/code/Magento and venia theme
to app/design/frontend/Magento/. Then I ran npm install in theme directory to setup the dependencies. I ran npm start to start up the dev server. This generated client.js and other javascript files as expected. After stopping dev server I ran setup:upgrade to install the theme and the pwa module. I activated venia theme. After opening my store front these are issues I had to fix
1) **404 for client.js** (because the path was /pub/static/frontend/Magento/venia/en_US/client.js instead of /pub/static/frontend/Magento/venia/en_US/js/client.js)
2) client.js:1 Uncaught ReferenceError: webpackJsonp is not defined
    at client.js:1. To fix this I had to include vendor.js before client.js
3) Everything loads but screen remains black with "loading" text. I noticed response from http://23-develop.to/graphql an request was 
```
{"data":{"urlResolver":null}}
```
It would be great if we could move the module-pwa in a separate repository as pointed out in [issue](https://github.com/magento-research/venia-pwa-concept/issues/62)

I'm still working on the last issue. But overall I would love to streamline these things and document them.

PWAStudioBot:
_From @mujahid-emb on May 15, 2018 5:24_

[...]

PR #248, opened 2018-08-22
Add UPWARD specification, guide, tests, reference implementation

zetlen:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[x] Documentation

## Summary

When this pull request is merged, it will add the UPWARD specification, guide, test suite, and reference implementation in JavaScript. See the README in this pull request for details.


packages/upward-js/lib/ParameterValidator.js
```diff
@@ -0,0 +1,110 @@
+/**
+ * Return functions which return the empty string if validation passes, and a
+ * descriptive string if validation fails.
+ */
+const typeTests = {
+    oneOf(...allowed) {
+        return value =>
+            allowed.includes(value)
+                ? undefined
+                : `must be one of: ${allowed.join()}`;
```
jimbo: In the passing case, why does this return `undefined`, rather than `''` like all the rest of these tests?

packages/upward-js/lib/ParameterValidator.js
```diff
@@ -0,0 +1,110 @@
...
+                name,
+                descriptor,
+                test
+            });
+
+            if (itemErrors.length > 0) {
+                errors.push({
+                    name,
+                    errors: itemErrors
+                });
```
jimbo: It doesn't look like `errors` is defined at this point, so `errors.push` looks wrong to me here. Did you mean `report.errors.push()`?

zetlen:
Noting that I have to back-merge my integration branch into this.

zetlen:
Closes #28.

packages/peregrine/src/Router/resolveUnknownRoute.js
```diff
@@ -23,7 +21,34 @@ export default function resolveUnknownRoute(opts) {
...
+                );
+            try {
+                return handleResolverResponse(
+                    JSON.parse(preloaded.textContent)
+                ).then(x => {
+                    preloadDone = true;
+                    return x;
+                }, rejectPreload);
+            } catch (e) {
+                rejectPreload();
```
jimbo: It looks like `rejectPreload` accepts a single argument, `e`. It's not being called with any arguments here, though. Did you mean to pass `e` to it?

packages/peregrine/src/Router/resolveUnknownRoute.js
```diff
@@ -23,7 +21,34 @@ export default function resolveUnknownRoute(opts) {
             id: res.id,
             matched: true
         }));
-    });
+    }
+
+    if (!preloadDone) {
+        const preloaded = document.getElementById('url-resolver');
+        if (preloaded) {
```
jimbo: To be clear: if `#url-resolver` doesn't exist in the document, don't do anything?

packages/pwa-buildpack/src/WebpackTools/plugins/UpwardPlugin.js
```diff
@@ -0,0 +1,96 @@
...
+    //     }
+    // );
+
+    // app.use(proxy);
+
+    // UPWARD middleware must be asynchronously generated, but plugins must
+    // apply their changes synchronously. Use a proxy function to add a
+    // middleware which lazy loads UPWARD when it's time.
+    let upwardMiddleware;
+    let upwardMiddlewarePromise = new Promise((resolve, reject) => {});
```
jimbo: This variable, `upwardMiddlewarePromise`, appears to be unused. Is this dead code? How is this supposed to work?

packages/pwa-buildpack/src/WebpackTools/plugins/createAdminRestClient.js
```diff
@@ -0,0 +1,78 @@
...
+        const fetchAuthorized = () =>
+            fetchREST(pathname, addFetchOptions(fetchOptions));
+        return fetchAuthorized().then(
+            res =>
+                res.statusCode === 401
+                    ? getToken().then(fetchAuthorized)
+                    : res.json()
+        );
+    };
+};
```
jimbo: Clean and understandable. 👍 

Tests would be good.

packages/upward-js/bin/server
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env node
+require('dotenv').config();
+
+function castValue(value) {
+    const mapping = {
+        false: false,
+        true: true
+    };
```
jimbo: Consider using a `Map` for this kind of thing.

packages/upward-js/lib/Context.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+        this._data[base] = value;
+    }
+
+    forget(base) {
+        delete this._data[base];
+    }
+}
+
+module.exports = Context;
```
jimbo: Can you explain what this file is doing?

packages/upward-js/lib/ContextPath.js
```diff
@@ -0,0 +1,101 @@
+const debug = require('debug')('upward-js:ContextPath');
+const illegalPathChars = /(^[\.\[\]])|[^\.\w\$\/]/;
```
jimbo: Consider adding a comment either explaining the regex or giving an example.

packages/upward-js/bin/server
zetlen: It was so small, I thought it okay this once.

packages/upward-js/lib/ResolverVisitor.js
```diff
@@ -0,0 +1,107 @@
...
+
+        if (typeof defined !== 'object' || !this.getResolverFailure) {
+            throw new Error(`Unexpected value in config: ${defined}`);
+        } else {
+            throw new Error(this.getResolverFailure);
+        }
+    }
+    getResolverFor(defined, propertyName) {
+        let Resolver;
+        for (Resolver of ResolverList) {
```
jimbo: The scoping of `Resolver` is misleading here.

- `Resolver` is declared _as a mutable variable_ prior to the `for...of` block
- On each loop iteration, `Resolver` is reassigned and `Resolver.recognize()` is always called
- On each loop iteration, there may be an early return

The only time this arrangement is useful is if the code following the `for...of` block is interested in the value of `Resolver` _after the loop has completed_. Otherwise, we would expect to see `for (const Resolver of ResolverList) {}`.

However, the code following the `for...of` block ("subsequent code") is not interested in the value of `Resolver`. In fact, the subsequent code always reassigns `Resolver` before ever reading its value. This is counterintuitive because `Resolver` appears to have been specifically declared such that it is meaningful, yet it isn't.

My recommendation:

```es6
getResolverFor(defined, propertyName) {
    for (const Resolver of ResolverList) {
        // ...
    }

    const definedResolver = defined.resolver

    const Resolver = definedResolver
        ? ResolversByType[definedResolver]
        : ResolverList.find(({ telltale }) => defined.hasOwnProperty(telltale))

    if (Resolver) {
        return new Resolver(this)
    } else {
        this.getResolverFailure = definedResolver
            ? 'Unrecognized resolver type...'
            : 'Unrecognized configuration...'
    }
}

```

PR #254, opened 2018-08-26
Replace <a href=""> with <Link to="">

shakyShane:
Any link that currently works with a regular `<a>` should just be a `<Link>`
element instead. 

This PR adds the `<Link>` element in a few places to reduce the amount of full-page reloads.

## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

1. wraps the logo in a <Link> to allow navigation back to the store home
1. allow navigation from Category -> Product 
1. added the `IntlPolyfill` to the item spec, since Jest cannot handle the Price rendering in the item component, and we need to do a `mount` or `render` in order to make this test useful

## Summary

When this pull request is merged, it will prevent page reloads for some basic navigation items

### DIFF
--- a/packages/venia-concept/src/__mocks__/@magento/peregrine.js
+++ b/packages/venia-concept/src/__mocks__/@magento/peregrine.js
@@ -1,0 +1,8 @@
+import { createElement } from 'react';
+
+/**
+ * the Price component from @magento/peregrine
+ * has browser-specific functionality and cannot
+ * currently by rendered in the test environment
+ */
+export const Price = () => <div />;
--- a/packages/venia-concept/src/RootComponents/Category/category.js
+++ b/packages/venia-concept/src/RootComponents/Category/category.js
@@ -18,6 +18,7 @@ const categoryQuery = gql`
                     id
                     name
                     small_image
+                    url_key
                     price {
                         regularPrice {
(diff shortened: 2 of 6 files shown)


ericerway:
Thanks for the PR @shakyShane! We'll take a look right away.

shakyShane:
@jimbo no problem, I can look into that :)

packages/venia-concept/src/__mocks__/@magento/peregrine.js
```diff
@@ -0,0 +1,8 @@
+import { createElement } from 'react';
+
+/**
+ * the Price component from @magento/peregrine
+ * has browser-specific functionality and cannot
+ * currently by rendered in the test environment
+ */
+export const Price = () => <div/>;
```
shakyShane: @jimbo I managed to acheive mock the Price component in this manner, which seems really clean as it gives complete control in a deterministic place (the `__mocks__` folder)

shakyShane:
@jimbo removed the extra dependency 👍

jimbo:
Just pulled it down locally and gave it a whirl. Great work!

PR #277, opened 2018-09-13
Cms block component

artKozinets:
## This PR is a:

[X] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will make possible using content from Magento admin > Content > Blocks

## How to use

**1. Import component to** /pwa-studio/packages/venia-concept/src/RootComponents/CMS/CMS.js
```
 import CmsBlock from 'src/components/CmsBlock'; 
``` 

**2. Add component to** /pwa-studio/packages/venia-concept/src/RootComponents/CMS/CMS.js
```
 <CmsBlock identifiers="cms_block_identifier_1" /> 
``` 

**Optional**

**1. To have content from multiple Cms Blocks**
``` 
<CmsBlock identifiers={["cms_block_identifier_1", "cms_block_identifier_2", "cms_block_identifier_etc"]} /> 
``` 

## Visualization

cmsblock


packages/venia-concept/src/components/CmsBlock/cmsBlock.js
```diff
@@ -0,0 +1,57 @@
...
+    render() {
+        const { identifiers, classes } = this.props;
+        return (
+            <div className={classes.root}>
+                <Query query={cmsBlockQuery} variables={{ identifiers }}>
+                    {({ loading, error, data }) => {
+                        if (error) return <div>Data Fetch Error</div>;
+                        if (loading) return <div>Fetching Data</div>;
+                        if (data.cmsBlocks.items == '')
+                            return <div>Here is not any block(s)</div>;
```
nadavspi: It might be better to log something to the console and return null, don't you think? Should a user see this error?

zetlen:
This works! I'm thrilled.

packages/venia-concept/src/components/CmsBlock/cmsBlock.css
```diff
@@ -0,0 +1,17 @@
+.root {
+    padding: 1rem;
```
zetlen: With multiple CMSBlocks next to each other, I think this should be a margin value rather than padding.

packages/venia-concept/src/components/CmsBlock/cmsBlock.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+`;
+
+class CmsBlock extends Component {
+    static propTypes = {
+        classes: PropTypes.shape({
+            root: PropTypes.string,
+            content: PropTypes.string
+        }),
+        identifiers: PropTypes.oneOfType([PropTypes.string, PropTypes.array])
```
zetlen: Is it normal in Magento frontend to concatenate several CMS blocks together like this? Or are you allowing arrays here just because the GraphQL schema specifies a collection?

If it's not normal to concatenate blocks like this, then I think it would be best to treat this as a scalar "single-block" component, and show an error if multiple results are returned (which should never occur if you're using identifiers).

packages/venia-concept/src/components/CmsBlock/cmsBlock.js
```diff
@@ -0,0 +1,57 @@
...
+                    {({ loading, error, data }) => {
+                        if (error) return <div>Data Fetch Error</div>;
+                        if (loading) return <div>Fetching Data</div>;
+                        if (data.cmsBlocks.items == '')
+                            return <div>Here is not any block(s)</div>;
+
+                        return (
+                            <div className={classes.content}>
+                                {data.cmsBlocks.items.map((item, index) => (
+                                    <div
```
zetlen: I think that you should delegate to a function-as-child-component here. The arguments to the child component could be a div containing the block content, the identifier, and the title. That way, someone could use different iterative rendering strategies for different collections of components.

As it is, I have to use a CMS block like this:

```jsx
<CmsBlock identifiers={["block1", "block2", "block3"]} />
```

And this renders:

```html
<div class="root">
  <div class="content">
    <div>Content 1</div>
    <div>Content 2</div>
    <div>Content 3</div>
  </div>
</div>
```

That's fine as default behavior _(though the inner divs should have a classname!)_ but I would also like to be able to do something like this:

```jsx
<div>
  <h2>Quick Links</h2>
  <ul>
    <CmsBlock identifiers={['block1','block2','block3']}>
      {({ key, identifier, title }) => (
        <li key={key}>
          <a href={`#anchor-${identifier}`}>{title}</a>
        </li>
      )}
    </CmsBlock>
  </ul>
  <h2>Sections</h2>
  <CmsBlock identifiers={['block1','block2','block3']}>
    {({ key, identifier, title, content }) => (
      <div key={key} id={`anchor-${identifier}`}>
        <h3>{title}</h3>
        <div dangerouslySetInnerHTML={{ __html: item.content }}></div>
      </div>
    )}
  </CmsBlock>
</div>
```

Which would yield this:

```html
<div>
  <h2>Quick Links</h2>
  <div class="root">
    <ul>
      <li>
        <a href="#anchor-block1">Title 1</a>
      </li>
      <li>
        <a href="#anchor-block2">Title 2</a>
      </li>
      <li>
        <a href="#anchor-block3">Title 3</a>
      </li>
    </ul>
  </div>
  <h2>Sections</h2>
  <div id="anchor-block1">
    <h3>Title 1</h3>
    <div>Content 1</div>
  </div>
  <div id="anchor-block2">
    <h3>Title 2</h3>
    <div>Content 2</div>
  </div>
  <div id="anchor-block3">
    <h3>Title 3</h3>
    <div>Content 3</div>
  </div>
</div>
```

So your implementation here would just assume that `children` was a function (specified in `propTypes` of course), and do something like this:

```jsx
const renderBlock = this.props.children || defaultImpl;
return (
  <div className={classes.content}>
    {data.cmsBlocks.items.map((item, index) => renderBlock({ key: index, ...item }))}
  </div>
);
```

zetlen:
@artKozinets Just checking again on this PR; we're looking forward to merging it, but we need at least a quick review from you on our comments.

Thanks!

packages/venia-concept/src/components/CmsBlock/cmsBlock.css
artKozinets: Make sense, there are no any left / right paddings / margins in design.

packages/venia-concept/src/components/CmsBlock/cmsBlock.js
artKozinets: There is such behaviour in graphQl schema. )
That is why did it in a same way.

So, 100% agree with you, in case of simplicity let`s change it.

packages/venia-concept/src/components/CmsBlock/cmsBlock.js
artKozinets: That is awesome idea!
I like it.

zetlen:
@artKozinets Would you like to make the change I suggested? I think it should be pretty simple, since there's an example in the comment. It's okay if you don't have time--just let us know!

ericerway:
@zetlen let's make the suggested changes now and come back to @artKozinets when available to extend CMS functionality beyond MVP. This is a key part of the homepage experience and would build off the new category/thumbnail work nicely. Thanks everyone!

PR #291, opened 2018-09-19
Make cert creation more verbose

pcvonz:
## This PR is a:

[ ] New feature
[X] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When cert creation fails, the error output wasn't very usable. It would usually just throw a JSON parse error because cert creation and JSONification were happening on the same line. Now those two steps are separated, so if cert creation throws an error it'll be shown.

Example: On my operating system `ca-certificates-update` doesn't exist. I have to call `update-ca-trust`. That error was being obscured by the `JSON` parse error. 

## Additional information

Related issue: #235


zetlen:
Related to #175 but non-breaking, so this one gets merged.

pcvonz:
@zetlen That would definitely solve the platform specific issues. Sounds :fire:

PR #281, opened 2018-09-13
Implement basic login for Venia (Issue #249)

JStein92:
## This PR is a:

[x ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will allow basic login using existing user accounts via REST API.

Addresses #249 

login_real

## Additional information

- `signin_token` is retrieved from Magento REST API and is stored it local storage. If `signin_token` is present in local storage, it is added to all GraphQL and REST requests headers.

- Adds a generic `Input` component, which lets developers customize `error`, `success`, and `hint` messages and the conditional which will display each one. This allows real time form validation, like ensuring a password is x characters, etc.

- Adds `Account drawer` which is built into `navigation`.

- Adds new actions and reducers for `user` functionality, (ex. `SIGN_IN` and `SIGN_IN_ERROR`)

- Adds storybook-dist to eslint ignore

PR #293, opened 2018-09-20
 Account creation (issue #250)

JStein92:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

This pull request is an extension to PR https://github.com/magento-research/pwa-studio/pull/281 - login feature. 

Addresses issue #250 

When this pull request is merged, in addition to the changes of #281, it will enable users to create accounts through Magento's REST api.

 #### New components
- CreateAccount
- Checkbox
- Form

create_account_gif

## Additional information

- User is automatically logged in after creating an account through getUserDetails() in the root index.js

- signin_token expires after some time but stays in local storage, so it's possible to receive a 401 Unauthorized when the root index.js tries to get user details with an expired token. Automatic token refresh should be addressed in the future.

- Guest cart is merged into newly created account, but the carts will not show up because getCartDetails still only looks for guest cart ids. Changing this seemed out of scope for this issue so we left it as is.

- Style changes for Input component to better match Invision designs. 

- Automatic email availability check with debounce using Magento isEmailAvailable endpoint.


packages/venia-concept/src/components/CreateAccount/createAccount.js
```diff
@@ -0,0 +1,220 @@
...
+        return this.passwordConfirmError ? 'Passwords do not match' : '';
+    }
+
+    get passwordHelpType() {
+        return HelpTypes.error;
+    }
+
+    render() {
+        const { classes } = this.props;
+        const {
```
zetlen: I like this pattern. You've done all the validation right here, through destructuring some getters. Nicely done.

People newer to JS might not understand what's happening here; it would be good to add a comment.

packages/venia-concept/src/components/CreateAccount/createAccount.js
```diff
@@ -0,0 +1,220 @@
...
+            };
+            this.props.createAccount(newCustomer);
+        }
+    };
+
+    handleCheckboxChange = value => {
+        this.setState({ subscribe: value });
+    };
+
+    checkEmail = debounce(async email => {
```
zetlen: Do you want this debounced on the prototype, or debounced per instance? Not that there will be multiple instances active at once in the near future, but if there are, the debounce will only exist per-instance.

packages/venia-concept/src/components/CreateAccount/createAccount.js
```diff
@@ -0,0 +1,220 @@
...
+                    body: JSON.stringify(body)
+                }
+            );
+            this.setState({ emailAvailable: response, checkingEmail: false });
+        } catch (error) {
+            console.warn('err: ', error);
+        }
+    }, 300);
+
+    updateLastName = newLastName => {
```
zetlen: Nice job on these methods. It's a really good idea to keep `setState` itself behind some business logic.

packages/venia-concept/src/components/Form/form.js
```diff
@@ -0,0 +1,29 @@
+import { createElement, Component } from 'react';
```
zetlen: This is a great, sleek form component, but since https://github.com/magento-research/pwa-studio/commit/8f387edf83c5416a9e379e1be5557bda0d12309a is already in the codebase, we should switch to using `informed` here. It shouldn't be a big change, because you've factored out the base form logic fairly well.

I know this is a hassle, but as we do the heavy lifting to get our code into 1.x state, there are some functionality collisions and this is one of them.

packages/venia-concept/src/components/Input/input.js
```diff
@@ -0,0 +1,139 @@
+import { Component, createElement } from 'react';
```
zetlen: This input has a lot of things that the `informed` input does not. I like it a lot; however, if you can refactor it so it is based on the `informed` input, I would feel better about merging it. Let me know how you want to proceed.

packages/venia-concept/src/components/Navigation/navigation.js
```diff
@@ -26,46 +31,136 @@ const tiles = CATEGORIES.map(category => (
...
+                {createAccountForm}
             </aside>
         );
     }
 }
 
-export default classify(defaultClasses)(Navigation);
+const mapStateToProps = state => {
+    return {
+        isSignedIn: state['user']['isSignedIn'],
```
zetlen: I'm surprised the linter didn't turn these into dot-lookup props here. Why are they bracket lookups?

packages/venia-concept/src/components/SignIn/signIn.js
```diff
@@ -0,0 +1,93 @@
...
+            <div className={classes.root}>
+                <Form submitForm={onSignIn}>
+                    <Input
+                        onChange={this.updateUsername}
+                        helpText={'example help text'}
+                        label={'Username or Email'}
+                        required={true}
+                    />
+
+                    <Input
```
zetlen: When I load this field in my browser, Chrome gives me an accessibility warning:

image

And links me here: https://www.chromium.org/developers/design-documents/create-amazing-password-forms

It looks like the autocomplete attribute would be easy to add here, and would help mobile devices know to activate their password managers.

packages/venia-concept/src/components/SignIn/signIn.js
```diff
@@ -0,0 +1,93 @@
...
+        const { classes, showCreateAccountForm } = this.props;
+        const { onSignIn, errorMessage } = this;
+
+        return (
+            <div className={classes.root}>
+                <Form submitForm={onSignIn}>
+                    <Input
+                        onChange={this.updateUsername}
+                        helpText={'example help text'}
+                        label={'Username or Email'}
```
zetlen: Suggestion: If the user has filled out this field and then clicks "Create Account" instead, it should copy this value into the CreateAccount counterpart field.

packages/venia-concept/src/index.js
```diff
@@ -1,20 +1,50 @@
...
+
+store.dispatch(getUserDetails());
+
+const httpLink = createHttpLink({
+    uri: urlBase,
+    __tmp_webpack_public_path__: __webpack_public_path__
+});
+
+const authLink = setContext((_, { headers }) => {
+    // get the authentication token from local storage if it exists
```
zetlen: Here's another place where you would want to use a BrowserPersistence instance.

packages/venia-concept/src/index.js
```diff
@@ -1,20 +1,50 @@
...
+        headers: {
+            ...headers,
+            authorization: token ? `Bearer ${token}` : ''
+        }
+    };
+});
 
-const apolloClient = new ApolloClient();
+const apolloClient = new ApolloClient({
+    link: authLink.concat(httpLink),
```
zetlen: Excellent job using Apollo Client exactly as intended.

packages/venia-concept/src/reducers/mockAccount.json
```diff
@@ -0,0 +1,73 @@
+{
```
zetlen: Please move this and all other mocks/fixtures into a __tests__ subfolder like __fixtures__. Add to the Jest exclude pattern if you have to.

packages/venia-concept/src/reducers/mockAccount.json
```diff
@@ -0,0 +1,73 @@
...
+  "customer": {
+    "id": 0,
+    "group_id": 0,
+    "default_billing": "string",
+    "default_shipping": "string",
+    "confirmation": "string",
+    "created_at": "2018-09-13T18:26:34.662Z",
+    "updated_at": "2018-09-13T18:26:34.662Z",
+    "created_in": "string",
+    "dob": "1/1/2010",
```
zetlen: Your customer created their account when they were eight years old. Adorable!

packages/venia-concept/package.json
```diff
@@ -7,7 +7,7 @@
     "main": "src/index.js",
     "repository": "github:magento-research/pwa-studio",
     "bugs": {
-      "url": "https://github.com/magento-research/pwa-studio/issues"
+        "url": "https://github.com/magento-research/pwa-studio/issues"
```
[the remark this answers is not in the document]
JStein92: Our editors are set to follow the editor config - not sure why it wasn't caught in the prettier. Thanks!

packages/venia-concept/src/actions/user.js
```diff
@@ -0,0 +1,121 @@
+import { RestApi } from '@magento/peregrine';
+import BrowserPersistence from 'src/util/simplePersistence.js';
```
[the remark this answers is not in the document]
JStein92: Done! We're exporting it the same way as RestAPI is exported from Peregrine.  :+1:

packages/venia-concept/src/components/SignIn/signIn.js
JStein92: Nice catch - I had verbose warnings off in Chrome so I missed this.

packages/venia-concept/src/components/Form/form.js
JStein92: No worries - it was easy to replace!

packages/venia-concept/src/components/Input/input.js
JStein92: So I implemented 'Form' from Informed like I mentioned above, and that was great, but the `Text` component of Informed (their version of `input`) doesn't seem to add anything here, and it actually makes some things more cumbersome and less readable. 

For example, I built the feature where the email/username you enter on signIn gets carried over to the createAccount input, but Informed's `Text` does not play well with `defaultValue`. Their version, `initialValue`, is loaded on component mount which makes the aforementioned feature much more difficult since the email/username is empty on mount. And I can't have a `defaultValue` with a `field`, which is frustrating because `field` is required.

I'm sure I could get this to work, but in the end it doesn't seem to add anything for the hassle.

If you would like, I can revisit this Monday. I'm happy to take another look.

JStein92: It looks like using Informed's <Text /> component actually makes some features more difficult (initialValue, for example, loads on mount so it doesn't allow us to easily copy the email a user enters from the `login` form to the `create account` form without doing some funky stuff).

I understand the appeal of using it, but @pcvonz and I are finding it to just be cumbersome at this point. Is it OK to omit using it for now?

packages/venia-concept/src/components/Navigation/navigation.js
JStein92: Whoops, just a carry over from programming too much in python. We will change it to dot lookup. :+1:

packages/venia-concept/src/components/CreateAccount/createAccount.js
JStein92: I'm ok leaving the debounce on the prototype - this way we are ensuring that the requests be limited by the debounce, instead of sending a new request every key press. Is that OK? I'm open to changing it.

zetlen: In your code, the debounce is on the instance. The class assignment shorthand means that this method is created per-instance.

It's totally fine--just wanted to make sure it was intentional. Usually the mistake goes in the other direction; a debounce on the prototype causes an arguably worse problem where individual calls from individual instances are lost. I put together a [Codepen to demonstrate.](https://codepen.io/zetlen/pen/zmGbyq?editors=0012)

JStein92:
@zetlen Ok, this pull request has been updated to meet the release/2.0 standards. : ) Also now includes tests (just basic coverage) for user actions/asyncActions. 

That was quite a merge!

zetlen:
It sure was, @JStein92 . It sure was. Thank you so, so much. I guess this closes #250 now!

PR #309, opened 2018-09-27
fix: buildpack handles http, venia forces https

zetlen:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[x] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will allow Buildpack, specifically PWADevServer and DevServerNotifierPlugin, to handle and display unsecure HTTP domains, which it will do by default unless devServer is configured with `provideSSLCert: true`.

It will also update Venia's Webpack config to add `provideSSLCert: true`.

### DIFF
--- a/packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
+++ b/packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
@@ -217,7 +217,7 @@ const PWADevServer = {
             );
         }
         devServerConfig.publicPath = url.format({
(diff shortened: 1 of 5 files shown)


jimbo:
I know this is already merged, but this does fix the bug I was experiencing, in which the server wouldn't respond at all.

It's still running on port `8001`, so there's another bug somewhere in the port-finding logic.

PR #318, opened 2018-10-02
Update queries for latest 2.3 schema and add validator script

zetlen:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[x] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

As of https://github.com/magento/graphql-ce/pull/132 the Product schema has changed; the `small_image` is now [an object with `url` and `path` parameters](https://github.com/magento/magento2/commit/646f2481823837a8bf9095c82c4e635c8aa69392#diff-4f95d8c79d48f23ec9c02fe228d006f5). This PR updates the corresponding queries, prop types, and tests to handle this change.

When merged, this will make Venia _incompatible with Magento 2.3 prior to [this commit](https://github.com/magento/magento2/commit/2bfc225897d551112527f64260f389686a6a91f2)_.

For developer discoverability, this PR adds a script which validates queries against the connected Magento instance: `npm run-script validate:venia:gql`.

image

Error example:

image

In a later PR, it would be a good idea to run this validator every time a query changes, or when a Magento instance is connected...


zetlen:
Because this validation is dependent on a connected Magento instance, I haven't integrated it into the test suite or CI. Right now, we don't have a connected Magento instance for either.

scripts/watch-all.js
```diff
@@ -0,0 +1,89 @@
...
+                dep => e.toString().indexOf(dep) !== -1
+            );
+            if (failedDeps.length === 0) {
+                // something unexpected happened
+                prerequisites.forEach(dep => spinner.error(dep));
+            } else {
+                failedDeps.forEach(dep => spinner.error(dep));
+            }
+            throw e;
+        }
```
jimbo: Consider avoiding two-arity `.then`, [which does not catch errors thrown by the resolve handler](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#Chaining).

```
// bad
execa()
  .then(
    () => {},
    err => {}
  )

// good
execa()
  .then(() => {})
  .catch(err => {})

File: packages/venia-concept/src/RootComponents/Category/category.js [hunk 2]
```diff
@@ -35,7 +37,7 @@ const categoryQuery = gql`
 
 class Category extends Component {
     static propTypes = {
-        id: number,
+        id: oneOfType([number, string]),
```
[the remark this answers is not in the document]
zetlen: This was a mistaken cherry-pick from another branch; I'm removing it :)

scripts/watch-all.js
zetlen: Wasn't supposed to be in this PR. But I'll change it in the other one. :)

pcvonz:
Works fine on my end!

PR #217, opened 2018-08-09
Custom loader and 404

pcvonz:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

- Add the ability to use a custom router
- Add the ability to add a custom component which is shown when the page is loading
- Add the ability to add a custom 404 page 

The custom loader fixes #14

The custom 404 page fixes #13

## Additional information

Hopefully I interpreted this correctly:

> We could separate MagentoRouter from the Peregrine implementation, expose it directly as a Peregrine component, and make a Peregrine config option for "router implementation", which the developer might supply in a render callback (or just a Router constructor which satisfies a certain prop interface!)
> 
> MagentoRouter could add a loadingView prop which takes a Component constructor
> MagentoRouter could accept a loading view which takes a render callback, with props about loading progress

Exported the MagentoRouter from Peregrine and added properties for a customLoader and a component to display when a route doesn't resolve anywhere. 

The last bit about loading progress isn't implemented yet. I can tackle that if someone could tell me where to start :+1:

PR #324, opened 2018-10-03
Update routing

jimbo:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will refactor the interfaces provided by Peregrine and consumed by Venia, so that Venia ultimately has more control over what is rendered. Specifically, this should enable theme authors to determine what is rendered while the app is loading or when a page is not found.

## Additional information

This is an alternative approach to the same issue addressed in #217. In Paul's proposed approach, the Peregrine API expands to accept custom components, and Peregrine renders them when appropriate. In my proposed approach, the Peregrine API shrinks to accept less configuration, instead providing context to Venia, which Venia uses to render whatever it wants.


packages/peregrine/src/Router/MagentoRouteHandler.js
```diff
@@ -1,74 +1,128 @@
...
+        } catch ({ message }) {
+            const symbol = message === '404' ? NotFound : InternalError;
+
+            // we don't have a matching RootComponent, but we've checked for one
+            // so associate the appropriate error case with this location
+            this.setRouteComponent(pathname, symbol);
+        }
+    }
+
+    setRouteComponent(pathname, RootComponent, meta) {
```
pcvonz: Should it be `setRootcomponent`?

packages/peregrine/src/Router/MagentoRouteHandler.js
```diff
@@ -1,74 +1,128 @@
...
+        this.setState(({ componentMap }) => ({
+            componentMap: new Map(componentMap).set(pathname, {
```
pcvonz: How did I not know this map object existed :skull:

shakyShane:
This seems the correct approach IMO 👍

packages/peregrine/src/Router/MagentoRouteHandler.js
jimbo: Haha, I can see the confusion. It was named `setRouteComponent` because it's setting the component for the route. The prior function is `getRouteComponent,` which gets a component for the route.

Do you think they need to be renamed?

packages/venia-concept/src/components/Main/main.js
```diff
@@ -8,21 +8,33 @@ import defaultClasses from './main.css';
...
-        isMasked: PropTypes.bool
+        isMasked: bool
     };
 
+    get classes() {
+        const { classes, isMasked } = this.props;
+        const suffix = isMasked ? '_masked' : '';
+
+        return ['page', 'root'].reduce(
+            (r, v) => ({ ...r, [v]: classes[`${v}${suffix}`] }),
```
pcvonz: Small style change change: Rename the arguments from `r` to `accum` and `v` to `value`. Would make it a little easier to read.

packages/venia-concept/src/util/ensureDirUri.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Given a URI, will always return the same URI with a trailing slash
+ * @param {string} uri
+ */
+export default function ensureDirUri(uri) {
```
pcvonz: Nice addition, I always forget the ending `/` :slightly_smiling_face:

packages/peregrine/src/Router/MagentoRouteHandler.js
jimbo: It's from ES2015, so it's still kind of new. [You're one of today's lucky 10,000!](https://xkcd.com/1053/) Plain objects still have plenty of uses, but `Map` has a great API.

Here, the `.set()` method returns the `Map` itself, so it can be chained.

```es6
const foo = new Map
  .set('a', 1)
  .set('b', 2)

console.log(foo.size) // logs 2
```

Also, `Map` is [iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols), and the default iterator is the same shape that the [constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) expects, so you can quickly clone a `Map` by feeding it to the constructor as we're doing here.

```es6
const foo = new Map()
  .set('a', 1)
  .set('b', 2)

const bar = new Map(foo)

expect(bar).toEqual(foo) // pass: `foo` and `bar` contain equivalent keys and values
expect(bar).toBe(foo) // fail: `foo` and `bar` are not the same object
```

packages/peregrine/src/Router/MagentoRouteHandler.js
pcvonz: We pronounce 'route' and 'root' the same here which has led to some confusion :smile: 
`setComponentToRoute` or something similar would clear up any confusion.

PR #332, opened 2018-10-05
Run child packages build automatically after installing dependencies

Igloczek:
## This PR is a:
Enhancement/Optimization

## Summary
I think it will be nice to automate the building of own packages dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "node": ">=8.0.0"
   },
   "scripts": {
+    "postinstall": "npm run build",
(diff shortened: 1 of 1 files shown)


zetlen:
Thanks, @Igloczek, but I have two concerns with this:

 - In another PR, #339, we've added a similar script which builds only pwa-buildpack and peregrine. Those are the only ones that need a full build to get started, and it's way faster.

- The `postinstall` hook runs both on install locally and installation as a dependency. PWA Studio will never be a dependency as an entire monorepo, but I still think it's a best practice to use the `prepare` script instead.

PR #338, opened 2018-10-08
Explicitly run build on first install

rowan-m:
Fixed the buildpack error that turns up on first install.


DrewML:
Thanks @rowan-m!

PR #354, opened 2018-10-12
fix(venia): rm staleGuestCartId on error

### LINKED ISSUE #344: Infinite loop with Venia add to cart
Experiencing an infinite loop in devtools when trying to add to cart when using a Cloud instance of 2.3.0.

zetlen:
Fixes #344.

The error was caused by a stale guestCartId whose corresponding cart in the API was no longer available. Since we don't have GraphQL for sessions yet, the guestCartId is stored in localStorage as well as in the Redux store. We were clearing it from localStorage on error, but forgetting to clear it from the store itself.

This fixes that!

## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[x] Bugfix
[ ] Test for existing code
[ ] Documentation

### DIFF
--- a/packages/venia-concept/src/util/omit.js
+++ b/packages/venia-concept/src/util/omit.js
@@ -1,0 +1,35 @@
+/**
+ * Return a copy of an object with one named property omitted.
+ * Short version of lodash.omit.
+ *
+ * Useful when a reducer needs to drop something from state, without using
+ * the `delete` keyword to mutate the existing state. The `delete` keyword
+ * is also a deopt.
+ *
+ * @param {object} obj -- Object to copy
+ * @param {string} omitKey -- Property name to omit
+ * @returns {object} Copy of the object with the specified key omitted.
+ *
+ * @example
+ *
+ * ```js
+ * const musketeers = {
+ *   athos: true,
+ *   porthos: true,
+ *   aramis: true
+ * };
+ *
+ * console.lot(omit(musketeers, 'athos'));
+ *
+ * {
+ *   porthos: true,
+ *   aramis: true
+ * }
+ */
+
+module.exports = (obj, omitKey) =>
+    Object.entries(obj).reduce(
+        (out, [key, value]) =>
+            key === omitKey ? out : ((out[key] = value), out),
+        {}
+    );
--- a/packages/venia-concept/src/reducers/cart.js
+++ b/packages/venia-concept/src/reducers/cart.js
@@ -3,6 +3,8 @@ import { handleActions } from 'redux-actions';
 import actions from 'src/actions/cart';
 import checkoutActions from 'src/actions/checkout';
 
+import omit from 'src/util/omit';
+
 export const name = 'cart';
 
 const initialState = {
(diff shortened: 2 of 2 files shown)


packages/venia-concept/src/util/omit.js
```diff
@@ -0,0 +1,35 @@
...
+ * @example
+ *
+ * ```js
+ * const musketeers = {
+ *   athos: true,
+ *   porthos: true,
+ *   aramis: true
+ * };
+ *
+ * console.lot(omit(musketeers, 'athos'));
```
Starotitorov: I believe it should be console.log()

packages/venia-concept/src/util/omit.js
```diff
@@ -0,0 +1,35 @@
...
+ * {
+ *   porthos: true,
+ *   aramis: true
+ * }
+ */
+
+module.exports = (obj, omitKey) =>
+    Object.entries(obj).reduce(
+        (out, [key, value]) =>
+            key === omitKey ? out : ((out[key] = value), out),
```
Starotitorov: I propose to follow functional paradigm and do not mutate function arguments.

```
(out, [key, value]) => key === omitKey ? out :  { ...out, [key]: value }
```

packages/venia-concept/src/util/omit.js
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * console.lot(omit(musketeers, 'athos'));
+ *
+ * {
+ *   porthos: true,
+ *   aramis: true
+ * }
+ */
+
+module.exports = (obj, omitKey) =>
```
Starotitorov: In general, James, are you sure that we need to add custom implementation of omit function to the project? My personal opinion, that it will be better to add lodash or underscore to the project. This libraries contains a lot of useful methods, that we will need in our app in the future like omit. If we decide to write our own implementations of such methods, than we will have to maintain our code, cover it with tests, write docs etc. Lodash is very popular library. The only thing we should remember is to avoid imports `import _ from 'lodash';` and just write, for example, `import omit from 'lodash/omit';`

Starotitorov: By the way, it will be useful to add ability to pass an array of omitKeys to the function.

zetlen: In my experience, Lodash adds a lot of file size, even when using tree-shaking. There are a few places in Venia and Peregrine where a Lodash method might be useful, but so far we have avoided its use. But I could be persuaded!

If you think that Lodash would ultimately help the codebase, then I would love for you to create a branch, refactor those parts of Venia and Peregrine to use Lodash, make sure they do not increase bundle size, and open a pull request!

packages/venia-concept/src/util/omit.js
zetlen: In this implementation, the spread operator creates a new object and does not mutate the existing one.

However, there are other places where we do mutate intermediate results during a reduce operation. I consider these acceptable because they are local mutations only, to an object which is not exposed to the outside closure.

zetlen:
@Starotitorov We may still decide to use Lodash in a few places, because you're right that we need to maintain as little code as possible. But for this PR, I realized that the standard in other reducers was to use an ES7 object spread and set the property to null, so I changed it to that. There's no longer any need for an omit function!

Please review again when you have time.

mhhansen:
will double check in short to see if #344 stills reproduce

pcvonz:
looks good! :+1:

PR #402, opened 2018-10-22
enhancement: category image fallback to first product image

### LINKED ISSUE #364: Homepage thumbnail images for Categories in Venia (PWA)

[X] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-devdocs`
[ ] `upward-js`
[ ] `upward-spec`

## This issue is a:

[X] Bug
[ ] Feature suggestion
[ ] Documentation issue
[ ] Other (Please Specify)

### Environment

## Description

Thumbnail images are missing from each of the Venia categories. These can replaced programmatically by the first image in each category.

## Expected result:

Thumbnail images shown.

## Possible solutions:

Possible GraphQL/REST API coverage needs. Need to review options.

zetlen:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will fallback images to icons in category lists, when the categories do not have an `image` assigned. The fallback uses the `small_image` of the first product in the category. Fixes #364 .

## Additional information

In creating a fallback URL, I had to write logic to generate media URLs for category images and product images. This meant refactoring the `MAGENTO_BACKEND_PRODUCT_MEDIA_PATH` environment variable, and on reflection, I decided to remove it. In future versions of PWA Studio, image URLs will come from a CDN or a third-party optimization service, so this code will eventually be deprecated and/or changed.


packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -66,30 +77,51 @@ class CategoryList extends Component {
...
-                                                />
-                                            )}
-                                        </span>
-                                        <span className={classes.name}>
-                                            {item.name}
-                                        </span>
-                                    </Link>
-                                ))}
+                                {data.category.children.map((item, index) => {
+                                    let imagePath;
```
Starotitorov: I believe it will be better to move logic of making imagePath to separate helper function. In general, I propose to create file helpers.js in component folder and move some local helper functions, which is used only in this component there. In this case we can call helper `getImagePath`;

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -66,30 +77,51 @@ class CategoryList extends Component {
...
+                                        ) {
+                                            imagePath = makeProductMediaPath(
+                                                firstProduct.small_image.path
+                                            );
+                                        }
+                                    }
+                                    return (
+                                        <Link
+                                            className={classes.item}
+                                            to={`/${
```
Starotitorov: The same, we can move this logic to helpers, we can add helper function `buildItemUrl`

zetlen:
@Starotitorov I've refactored `CategoryList` to include a `CategoryTile` child component which uses a getter to retrieve its image path. This matches best practice for our other components.

Your Query component idea is excellent; we have thought about similar things, and https://github.com/magento-research/pwa-studio/pull/217 is a partial implementation of this API. For now, I've written my query to match others in our project. What do you think?

Starotitorov:
@zetlen Actually, I like the idea of helpers more. This approach has several benefits:
1. If you need it you can easily reuse this helper function somewhere else without refactoring of existing code and extracting class method.
2. When you wrote helper function and the code includes some complex calculations you can easily create memoized version of the function, the only thing you need to do is to pass your function to memoize function and export the result.

zetlen:
Helper functions are great! We have several. But this `previewImage` logic is specific to category icons, and I don't expect a need to reuse it in other situations.

We use getters to separate sub-units of rendering logic, _specifically because_ you cannot pass arguments to getters. Therefore, getter implementations have the same contract as `render()` methods, which do not receive arguments. They only require component props to be present, and therefore they keep all relevant data inside the React props pattern, rather than passing it through additional functions and adding dependencies on additional signatures.

PR #384, opened 2018-10-18
Standarise .env file naming convention

php4umagento:
## This PR is a:

[ ] New feature
[X ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ X] Documentation

## Summary

When this pull request is merged, it will rename example.env into .env.dist
 This is more widely used approach

## Additional information

Also changed documentation and circleci test

PR #429, opened 2018-10-30
Add new release cloud environment, add environment validation

zetlen:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[x] Test for existing code
[ ] Documentation

## Summary

- Rename `example.env` to `env.dist` following convention (closing #384)
- Replace the default `MAGENTO_BACKEND_URL` value in `env.dist` with the new "release" cloud environment: https://release-utkfd3a-dpfibs5yviagi.us-3.magentosite.cloud/
- Add more environment variable validation, so that new users of the project see clearer instructions for enabling their development environment after install
- Switching to a schema-validated environment object required that the UpwardPlugin and `upward-js` no longer use `process.env` directly
- Reverted parts of #318 and restored the simple `small_image: String` GraphQL query structure to compatible with 2.3.

## Additional information

I had to revert the `small_image` schema change because it turns out that the change to the schema has not made it into Magento 2 `2.3.0-release`. We are no longer developing on Magento 2 `2.3-develop`.

### DIFF
--- a/packages/venia-concept/validate-environment.js
+++ b/packages/venia-concept/validate-environment.js
@@ -1,0 +1,94 @@
+function validateEnvironment(env) {
+    const { readFileSync } = require('fs');
+    const path = require('path');
+    const dotenv = require('dotenv');
+    const chalk = require('chalk');
+
+    let parsedEnv;
+    const envPath = path.join(__dirname, '.env');
+    try {
+        parsedEnv = dotenv.parse(readFileSync(envPath));
+        console.log(
+            chalk.green(
+                `Using environment variables from ${chalk.greenBright('.env')}`
+            )
+        );
+        if (env.DEBUG || env.NODE_DEBUG) {
+            console.log(
+                '\n  ' +
+                    require('util')
+                        .inspect(parsedEnv, {
+                            colors: true,
+                            compact: false
+                        })
+                        .replace(/\s*[\{\}]\s*/gm, '')
+                        .replace(/,\n\s+/gm, '\n  ') +
+                    '\n'
+            );
+        }
+    } catch (e) {
+        if (e.code === 'ENOENT') {
+            console.log(
+                chalk.redBright(
+                    `\nNo .env file in ${__dirname}\n\tYou may need to copy 'env.dist' to '.env' to begin, or create your own '.env' file manually.`
+                )
+            );
+        } else {
+            console.log(
+                chalk.redBright(`\nCould not retrieve and parse ${envPath}.`, e)
+            );
+        }
+    }
+
+    const envalid = require('envalid');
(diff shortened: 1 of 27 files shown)


packages/venia-concept/validate-environment.js
```diff
@@ -0,0 +1,94 @@
+function validateEnvironment(env) {
+    const { readFileSync } = require('fs');
+    const path = require('path');
+    const dotenv = require('dotenv');
+    const chalk = require('chalk');
+
+    let parsedEnv;
+    const envPath = path.join(__dirname, '.env');
+    try {
+        parsedEnv = dotenv.parse(readFileSync(envPath));
```
vitalics: It a good practice to use also encoding for parsing `.env` files, because it can be an error for people who used another encoding
``` js
parsedEnv = dotenv.parse({path: readFileSync(envPath), encoding: 'utf-8'});
```

packages/venia-concept/validate-environment.js
```diff
@@ -0,0 +1,94 @@
...
+                'Tell the upward-js prod server to log the bound URL to stdout. Useful in testing and discovery scenarios, but may be disabled in production.',
+            default: true
+        })
+    });
+}
+
+module.exports = validateEnvironment;
+
+if (module === require.main) {
+    validateEnvironment(prcoess.env);
```
jahvi: Looks like a typo for `process.env`

packages/venia-concept/validate-environment.js
mhhansen: thanks @vitalics !
In that scenario, it won't be applied even being `utf8` the [default encoding](https://www.npmjs.com/package/dotenv#encoding)? (I'm just curious...)
maybe will be better to specify that at `require().config(xxxx)` level ?
cause we're using the `.env` in a lot of places, maybe it's a good chance to standardize this.

If that's the case, regarding `utf-8` value:
would make sense to specify as `utf8` and be aligned on how [the lib specifies this param](https://github.com/clakech/dotenv/blob/master/lib/main.js#L48) in their docs?
(despite that `readFileSync()` and `Buffer` from node _understands_ both values).

what do you think?

vitalics: @mhhansen 
Yes, by default it's  `utf-8` encoding, but what about the cross various solution. Read more about this on [this ](http://unicode.org/faq/utf_bom.html) article in next section: `Which of the UTFs do I need to support?`.
Quote: `UTF-8 is most common on the web. UTF-16 is used by Java and Windows. UTF-8 and UTF-32 are used by Linux and various Unix systems.`

I agree with you that we can make it on `require` level for defining it visually, it will be more clarify and understandable for others persons on this project

PR #451, opened 2018-11-05
feat(buildpack): Upgrade to Webpack 4 and refactor plugins

zetlen:
## This PR is a:

[x] New feature
[x] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

Upgrade to Webpack 4, speeding up and shrinking builds and giving us
more sophisticated optimization and bundling techniques.

Summary of changes:
 - Updated Webpack and its dependencies:
   - Loaders and plugins
   - Dev server
   - Replaced uglifyjs with more maintained `terser` for minification
   - webpack-cli is now bundled separately
 - Modified Venia webpack.config.js to use Webpack 4 mode, optimization, and plugin configuration
 - Refactored MagentoRootComponentsPlugin to use a generated module with a map of RootComponent names to dynamic import functions, achieving the same chunk splitting without using a loader to embed dummy code
    - TODO: skipped large test suite for expediency, should unskip
 - Refactored MagentoRouteHandler to no longer use __tmp_webpack_public_path__ hack that earlier plugin required
 - Modified UpwardPlugin to adapt to changed Webpack virtual FS interface

## Additional information

This is the **beginning** of a Webpack 4 upgrade. Many more
optimizations are possible with Webpack 4 and they should appear in
upcoming pull requests. This is a minimal changeset for easiest code
review.


shakyShane:
@zetlen looks good - can we limit the PR to *just* this webpack 4 upgrade (to enable a timely merge) - adding additional optimizations once it's landed?

packages/venia-concept/package.json
```diff
@@ -14,23 +14,18 @@
   },
   "homepage": "https://github.com/magento-research/pwa-studio/tree/master/packages/venia-concept#readme",
   "scripts": {
-    "build": "npm run -s validate-queries && webpack --color --progress --profile --env.phase production",
+    "build": "rimraf dist && npm run -s validate-queries && webpack --color --progress --profile --env.mode production",
```
jimbo: Does `&&` work on Windows? My cross-platform npm is a bit rusty.

packages/venia-concept/webpack.config.js
```diff
@@ -105,30 +106,48 @@ module.exports = async function(passedEnv) {
...
+                        chunks: 'all'
+                    }
+                }
+            }
+        }
     };
-    if (phase === 'development') {
-        config.devtool = 'eval-source-map';
+    if (mode === 'development') {
+        config.devtool = 'inline-source-map';
```
jimbo: Why? The original setting, `eval-source-map`, seems to be the same but strictly faster.

packages/venia-concept/webpack.config.js
```diff
@@ -12,15 +12,15 @@ const {
 } = require('@magento/pwa-buildpack');
 const path = require('path');
 
-const UglifyPlugin = require('uglifyjs-webpack-plugin');
+const TerserPlugin = require('terser-webpack-plugin');
```
jimbo: What's the justification for this change?

zetlen:
@shakyShane This PR is large because of some technical debt in MagentoRootComponentsPlugin: it was tightly coupled to Webpack internal, private interfaces. Upgrading Webpack without changing the RootComponentsPlugin would have broken the app, making the PR unmergeable. I recognize this is huge, but there's no good way around it.

One way around it would be to rewrite MagentoRootComponentsPlugin so that it is loosely coupled, _and_ so that it has strategies for Webpack 3 and 4, and I considered doing that--but that PR alone would look strange and unnecessary by itself, and would be hard to review. Plus, once Webpack 4 merged, the Webpack 3 strategies would be dead code.

So I know this is huge, but believe it or not, it's a minimal changeset I extracted from a much larger branch.

packages/pwa-buildpack/src/WebpackTools/MagentoResolver.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
         return {
             modules: [options.paths.root, 'node_modules'],
             mainFiles: ['index'],
-            extensions: ['.js', '.json']
+            extensions: ['.mjs', '.js', '.json', '.graphql']
```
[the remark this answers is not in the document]
zetlen: We don't, but `node_modules` does, and with Webpack 4 tree shaking set up correctly, we will directly access those modules rather than using their Babelified `dist` versions.

packages/pwa-buildpack/src/WebpackTools/plugins/makeMagentoRootComponentsPlugin/index.js
```diff
@@ -0,0 +1,126 @@
...
+            .map(entry => entry.join(':'))
+            .join(',\n')}
+};
+const key = ${toRootComponentMapKey.toString()};
+export default function fetchRootComponent(type, variant = 'default') {
+    return rootComponentsMap[key(type, variant)]().then(m => m.default || m);
+};
+`;
+    }
+}
```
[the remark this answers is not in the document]
zetlen: You know, I actually kind of like it, but maybe I've been looking at it too long. This basic idea of building a "dynamic import dispatch" file is promising to me, because it lets us set up a big registry of dynamic imports which Webpack can detect and chunk, entirely through filesystem analysis, but we could still set up an arbitrary API for other modules to trigger the dynamic import. I think this kind of thing could be the basis for future extensibility concepts. Though I suppose we'd build it with proper codegen.

packages/venia-concept/package.json
zetlen: It does.

packages/venia-concept/webpack.config.js
zetlen: Whoops, didn't mean to check this in.

packages/venia-concept/webpack.config.js
zetlen: The [terser homepage](https://github.com/terser-js/terser) explains it pretty well. Webpack 4 tree shaking requires a minifier that understands ES6 natively. Without Terser, this PR would not produce the smaller builds it produces!

shakyShane:
@zetlen no arguments from me 👍

packages/pwa-buildpack/src/WebpackTools/plugins/makeMagentoRootComponentsPlugin/index.js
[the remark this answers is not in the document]
jimbo: Okay. Mostly, I just don't want to have to maintain code that's written in a string. This plugin has a limited job, though, and it's doing it well. I think this will be fine.

packages/venia-concept/webpack.config.js
jimbo: Yeah, I can see now that they simply gave up on `uglify-es`. 😕

PR #492, opened 2018-11-12
Remodel address form from checkout, add validation for the state field

AlexeyKaryka:
Update Input component, address form from the checkout, add validation for the state field

## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ X ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

Add validation block to the address form and show details about invalid address form fulfilling there.

## Summary

When this pull request is merged, it will fix issue with saving address in checkout https://github.com/magento-research/pwa-studio/issues/462 and Common Input does not handle initial value correctly https://github.com/magento-research/pwa-studio/issues/416

### DIFF
--- a/packages/venia-concept/src/components/Checkout/address.js
+++ b/packages/venia-concept/src/components/Checkout/address.js
@@ -1,28 +1,20 @@
-import React, { Component, Fragment } from 'react';
-import { Form, Text } from 'informed';
+import React, { Component } from 'react';
+import { Form } from 'informed';
 import memoize from 'memoize-one';
-import { func, shape, string, bool } from 'prop-types';
+import PropTypes from 'prop-types';
 
 import classify from 'src/classify';
+import Input, {
+    createComplexValidator,
+    defaultIsRequired
+} from 'src/components/Input';
 import Button from 'src/components/Button';
-import Label from './label';
-import { invalidStateMessage } from './constants';
-import { isString } from 'util';
+import { fields, fieldsLabels } from './constants';
+import { validatorStateField } from './helpers';
 import defaultClasses from './address.css';
 
-const fields = [
-    'city',
-    'email',
-    'firstname',
-    'lastname',
-    'postcode',
-    'region_code',
-    'street',
-    'telephone'
-];
-
 const filterInitialValues = memoize(values =>
-    fields.reduce((acc, key) => {
+    Object.keys(fields).reduce((acc, key) => {
         acc[key] = values[key];
         return acc;
     }, {})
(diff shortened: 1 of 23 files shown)


zetlen:
Hey @AlexeyKaryka, could you please link to the issue this resolves? You can leave a comment in a commit message like "Fixes #NNN" with the issue number. Then, we can easily open the deployed app and do UAT.

Starotitorov:
@zetlen here is a link to the issue https://github.com/magento-research/pwa-studio/issues/462, sorry

zetlen:
Oh excellent, thank you!

Starotitorov:
I guess it will be better to use asField HOC provided by informed and wrap our input component in this HOC instead of passing such properties as fieldValue, setFieldValue to common Input component in the implementation of getInput method.

Starotitorov:
@AlexeyKaryka In general, it is really great that you removed `value` and `dirty` properties from local state of our Input component because informed provide us this information and there is no sense to duplicate it.

packages/venia-concept/src/components/Checkout/address.js
```diff
@@ -1,52 +1,104 @@
...
+        const { classes } = this.props;
+        const inputClasses = {
+            input: classes.textInput,
+            root: classes.inputRoot,
+            rootFocused: classes.inputRootFocused
+        };
+        const validator = this.getValidator(fieldName);
+
+        return (
+            <div className={classes[fieldKey]}>
```
pcvonz: Need to add a `key` to this div.

packages/venia-concept/src/components/Checkout/flow.css
```diff
@@ -10,6 +10,7 @@
     bottom: 5.5rem;
     box-shadow: 0 -1px rgb(var(--venia-border));
     display: grid;
+    grid-template-columns: 50% 50%;
```
pcvonz: Don't think we want the checkout options to look like this:
image

packages/venia-concept/src/components/CreateAccount/createAccount.js
```diff
@@ -43,25 +47,28 @@ class CreateAccount extends Component {
...
 
     get isIncompleteOrInvalid() {
         // A catch-all to check if the form is valid so the customer can attempt to create an account
         return (
             this.hasEmailError ||
             this.hasPasswordConfirmError ||
-            !this.state.email ||
-            !this.state.firstName ||
-            !this.state.lastName
+            !this.getFieldValue(fields.email) ||
```
pcvonz: It'd be great if all of this was checked through `Informed`. Separate issue probably.

packages/venia-concept/src/components/CreateAccount/createAccount.js
```diff
@@ -174,14 +192,15 @@ class CreateAccount extends Component {
     }
 
     onCreateAccount = () => {
+        const { getFieldValue } = this;
```
pcvonz: You can just pass `formState` into this function. I'm not sure why all of these fields are kept in `state`.

packages/venia-concept/src/components/Input/input.css
```diff
@@ -25,29 +17,61 @@
...
+}
+
 .root {
     padding: 0.5rem 1.5rem;
+    --input-right-padding: 1rem;
+}
+
+.rootFocused {
+    composes: root;
+    background-color: rgb(var(--venia-teal-alt));
```
pcvonz: Follow: https://github.com/magento-research/pwa-studio/issues/517. Should have a `background-color`, etc.

packages/venia-concept/src/components/Input/input.js
```diff
@@ -82,52 +111,83 @@ class Input extends Component {
...
-                    title={title}
-                    autoComplete={autoComplete}
-                    onChange={this.handleChange}
-                    onFocus={this.focusTextInput}
-                    onBlur={this.blurTextInput}
-                    field={field}
-                />
-                {helpText}
+                <div className={classes.inputContainer}>
+                    <BasicText
```
pcvonz: I think we should have the option to pass in any `Informed` component, just wrapping `BasicText` really limits what we can do with `Informed`. This `Input` component could be made a lot more generic. 

For example, I think that the State field should use the `Select` component from `Informed` instead. Removes the need to have that error display too.

packages/venia-concept/src/components/CreateAccount/createAccount.js
pcvonz: You can just use `formState` much how you're passing `formState` into `handleFormState` here. I'm not sure why the whole form is duplicated into the component `state`.

packages/venia-concept/src/components/Checkout/flow.css
pcvonz: Why was this added? It changed the layout of each section of checkout.

packages/venia-concept/src/components/Input/input.js
```diff
@@ -82,52 +111,83 @@ class Input extends Component {
...
+                onClick={this.resetValue}
+            >
+                <Icon name={name} attrs={attrs} />
+            </button>
+        );
+    }
+
+    resetValue = () => {
+        const { fieldApi } = this.props;
+        fieldApi.setValue(null);
```
Starotitorov: You need to call onChange here.

PR #494, opened 2018-11-12
chore(deploy): Adjust production build and serve for cloud compat

zetlen:
## This PR is a:

[ ] New feature
[x] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

- Add a special case in Venia staging server for production mode, which
skips attempted custom host and uses $PORT var
- Add a special case in Venia query validator for production mode, which
skips query validation
- Refactor usage of NODE_ENV env var to production only

## Additional information

Until this PR is merged, cloud integration efforts will be blocked. Hoping for immediate review!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2404,6 +2404,7 @@
 			"version": "6.5.5",
 			"resolved": "https://registry.npmjs.org/ajv/-/ajv-6.5.5.tgz",
 			"integrity": "sha512-7q7gtRQDJSyuEHjuVgHoUa2VuemFiCMrfQc9Tc08XTAc4Zj/5U1buQJ0HU6i7fKjXU09SVgSmxa4sLvuvS8Iyg==",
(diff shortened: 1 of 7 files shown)


pcvonz:
I'm kind of confused about the difference between `validEnv` and `process.env`. If it's production do we skip the `.env` file?

zetlen:
@pcvonz This is a hack which could use some improvement. The core issue is that Webpack and Envalid both default `NODE_ENV` to `production` if it's blank.

We need to have four distinct modes: `development`, `test`, `staging`, and `production`, and we need to declare the mode for a command in a consistent way.

We'll open an issue about this (I'm at Chrome Developer Summit, otherwise I would) and we'd welcome a pull request. But I merged this hack to enable our Cloud team to push a deploy that didn't run the custom host or query validation.

PR #496, opened 2018-11-13
Implement create account page

Starotitorov:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will add separate create account page. We will navigate user from order confirmation page to this page.

https://github.com/magento-research/pwa-studio/issues/502


packages/venia-concept/src/components/AppShell/renderRoutes.js
```diff
@@ -0,0 +1,16 @@
...
+import ErrorView from 'src/components/ErrorView';
+import CreateAccountPage from 'src/components/CreateAccountPage';
+
+const renderRoutingError = props => <ErrorView {...props} />;
+
+const renderRoutes = () => (
+    <Switch>
+        <Route exact path="/create-account" component={CreateAccountPage} />
+        <Route render={() => <Page>{renderRoutingError}</Page>} />
+    </Switch>
```
jimbo: Are we certain that it's safe to intercept `/create-account` and define it as a client-side route? Afaik, routing is owned by the Magento instance.

Also, we already have a Create Account workflow in the left drawer. Why do we need a new page with the same workflow?

packages/venia-concept/src/components/CreateAccount/createAccount.js
```diff
@@ -20,7 +20,16 @@ class CreateAccount extends Component {
             createAccountError: PropTypes.string
         }),
         createAccountError: PropTypes.object,
-        createAccount: PropTypes.func
+        createAccount: PropTypes.func,
```
jimbo: It looks like this component now expects to receive `props.onSubmit` rather than `props.createAccount`, so this prop type needs to be updated.

packages/venia-concept/src/components/CreateAccountPage/helpers.js
```diff
@@ -0,0 +1,9 @@
+const validCreateAccountParams = ['email', 'firstName', 'lastName'];
+
+export const getCreateAccountInitialValues = () => {
+    const params = new URLSearchParams(window.location.search);
+    return validCreateAccountParams.reduce(
+        (values, param) => ({ ...values, [param]: params.get(param) }),
+        {}
+    );
+};
```
jimbo: This would be a good file to have tests for, since it has edges. Consider at least verifying that it doesn't throw errors in common cases.

packages/venia-concept/src/components/Navigation/container.js
```diff
@@ -18,7 +19,7 @@ const mapStateToProps = ({ catalog, user }) => {
     };
 };
 
-const mapDispatchToProps = { closeDrawer, getAllCategories };
+const mapDispatchToProps = { closeDrawer, getAllCategories, createAccount };
```
jimbo: Why did we move this action creator out of the `CreateAccount` container, only to pass it back in from two different places?

packages/venia-concept/templates/index.mst
```diff
@@ -0,0 +1,9 @@
+{{> templates/doctype-and-head-start}}
+    <title>Venia</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    {{> templates/preloads}}
+    {{> templates/scripts}}
+  </body>
+</html>
```
jimbo: What exactly is this file? When does it render?

packages/venia-concept/src/components/AppShell/renderRoutes.js
Starotitorov: I discussed the question about create separate route for create account page with James, we decided to implement this for now.
Out current Create Account workflow is bad from SEO perspective, it is important to have direct urls for Create Account and Sign In pages.
Initially the problem was to navigate user from order confirmation step of checkout in the right drawer to create account page in the left drawer. Current implementation does not allow to do that, because we manage drawer state locally.
@zetlen proposed to implement a routed drawer, but this solution requires a lot of effort to implement it.
Just to summarize, in this pull request I added code to implement business requirement with navigating user to Create Account page from Order Confirmation and this separate Create Account page is better from SEO perspective.

packages/venia-concept/src/components/CreateAccount/createAccount.js
Starotitorov: Yeah, definitely, I will change this.

packages/venia-concept/src/components/CreateAccountPage/helpers.js
Starotitorov: I agree

packages/venia-concept/src/components/AppShell/renderRoutes.js
jimbo: Okay.

This does satisfy the business requirement of navigating the user after confirmation. I would prefer to open the left drawer to Create Account, but this approach of rendering a new page is acceptable for now. We can treat this approach as tech debt and have someone else revisit it later as part of a broader look at routing and transitions.

Starotitorov:
@jimbo, I believe this issue should be fixed separately because it is not a part of the story.

Starotitorov:
In general, I created an issue on github https://github.com/magento-research/pwa-studio/issues/416 some time ago.

Starotitorov:
@jimbo I passed initialValues to Form component according to your comment, but if we need a quick fix, we need to rollback this and pass initialValue to each field separately and merge pull request from Aliaksei

jimbo:
@Starotitorov Yeah, after looking at the changesets required, I agree that we should fix inputs in a separate PR. We can pick up where we left off in #593.

PR #498, opened 2018-11-13
feat(deploy): Venia can be built and deployed separately

zetlen:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will enable a Venia deployment that does not require cloning, hoisting, and building the entire pwa-studio repository.

- Change all Venia devDependencies to dependencies, so it can be build
from node_modules with only "production deps"
- Add a "venia" CLI to the package which builds and launches a server

This change enables a downstream user to create a package which launches
Venia as simple as this:

```sh
$ npm install -g @magento/venia-concept
$ NODE_ENV=production venia build
$ NODE_ENV=production MAGENTO_BACKEND_URL=https://magento.internal venia start
```

### DIFF
--- a/packages/venia-concept/bin/venia.js
+++ b/packages/venia-concept/bin/venia.js
@@ -1,0 +1,23 @@
+#!/usr/bin/env node
+
+const execa = require('execa');
+
+const cwd = require('path').resolve('../', __dirname);
+
+function npmRun(cmd) {
+    execa.shell(`npm run -s ${cmd}`, {
+        cwd,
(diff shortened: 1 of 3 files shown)


zetlen:
Merging to enable cloud deployment. Thanks for your patience. Since I'm skipping the review process, I could have pushed directly to the default branch. Instead, I did a PR so that we could review and if necessary revert changes.

PR #507, opened 2018-11-15
Update global footer to consume data from the cms

AlexeyKaryka:
## This PR is a:

[ ] New feature
[ X ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

Add footer ability to consume data from content blocks. This pr directly depends on https://github.com/PMET-public/module-venia-cms-sample-data/pull/2


packages/venia-concept/src/components/CmsBlock/cmsBlock.js
```diff
@@ -15,7 +17,27 @@ class CmsBlockGroup extends Component {
...
+        history: object
+    };
+
+    componentDidMount() {
+        this.interceptLinksClicks();
+    }
+
+    interceptLinksClicks = () => {
+        const { history, classes } = this.props;
+        document.getElementsByClassName(classes.root)[0].addEventListener(
```
Starotitorov: You can use React refs.

packages/venia-concept/src/components/CmsBlock/cmsBlock.js
```diff
@@ -62,4 +84,4 @@ class CmsBlockGroup extends Component {
     }
 }
 
-export default classify(defaultClasses)(CmsBlockGroup);
+export default classify(defaultClasses)(withRouter(CmsBlockGroup));
```
Starotitorov: use compose function here

packages/venia-concept/src/components/CmsBlock/cmsBlock.js
AlexeyKaryka: Yeah, thanks for the good idea

zetlen:
@jimbo I agree with you on all points, especially the first; the icon interpolation mixes code and data inappropriately and can't be merged (as well as being incompatible with #733). So we will need to update this code.

However, 2) and 3) will have to happen in some form, and this implementation is mostly good. We must support the TinyMCE-powered rich content in product description fields, CMS pages, and blocks _at least_, since that's part of the promised MVP core functionality, and it's necessary to make a homepage. So we have `dangerouslySetInnerHTML` and capture events right now because there are no other reasonably performant solutions for this. The simplest would be to validate DOM with a document fragment before injecting it, but then you're doing at least a DOM parse in the constructor and a large string compare in `shouldComponentUpdate`, on the main thread, with every render. I think that's a recipe for jank. The XSS vulnerability is real, but the existing storefront has the same vulnerability, so that's just the water we swim in.

Here's the future, though: 

- Parsing the HTML in a worker and getting back JSON to map into components
- Unifying that interface with a parser for the PageBuilder storage format

I have an [architecture proposal](https://github.com/magento/architecture/pull/26) for the unified interface, as well as an [update with a workaround](https://github.com/magento/architecture/blob/af592f781e990f048a845a1da459b0b734bd043f/proposals/rich-content-graphql-api.md#pwa-studio-mvp-workaround) in case we have to do PageBuilder support without a real GraphQL API.

I'm reopening this because we do have to have a solution for intercepting links. @jimbo I think we should move forward with a `dangerouslySetInnerHTML` implementation and backlog a spike to see if the workaround I describe in the above proposal can be done without a huge perf hit.

For now, we need to do to this PR:

- [ ] Remove the icon interpolation
- [ ] Capture click events more safely (using direct subscription if possible)

jimbo:
This is a good exploration of _how_ we could do this, but we don't _want_ to do this at the moment. I'm going to close this; we can revisit in the future.

Issue #452, opened 2018-11-05
Use GraphQL for Category Navigation


awilcoxa:
## This issue is for the following packages:

[x ] `venia-concept`
[ ] `pwa-buildpack`
[x ] `peregrine`
[ ] `pwa-devdocs`
[ ] `upward-js`
[ ] `upward-spec`

## This issue is a:

[ ] Bug
[x ] Feature suggestion
[ ] Documentation issue
[ ] Other (Please Specify)

### Environment

| Question                    |    Answer     |
| --------------------------- | :-----------: |
| Magento version             |  |
| Operating System + version  |  |
| node.js version (`node -v`) |  |
| npm version (`npm -v`)      |  |

## Description

During demo on 11.5 it was suggested that store categories are available in GraphQL. If true, we should replace the hard coded categories in in Venia with the available category endpoints for GraphQL. 

{Placeholder ticket, needs more detail from @jimbo}

PR #533, opened 2018-11-21
Graphql categories

Serunde:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will enable full product category navigation from the app's left side drawer using real data from graphql. 

Addresses #452 
navigationdemo2

## Additional information

This solution makes recursive queries in order to populate the full menu. If the graphql schema is updated to expose the entire category tree with one query, this logic can be replaced.


ennostuurman:
hey @Serunde (endures?) . Pulled this PR in locally and (no surprise) works great (tested with M2 fixtures sample data). Please note your branch is not synced with latest release/2.0 branch.  
This is an important fix for a MVP IMO (cc @ericerway ) .

ericerway:
Agreed @ennostuurman -- this just came in and is on our list to review/merge for MVP, which would replace the currently mocked navigation in Venia. Thanks for your feedback! /cc @jimbo @zetlen for visibility.

packages/venia-concept/src/components/Navigation/navigation.js
```diff
@@ -179,15 +181,17 @@ class Navigation extends PureComponent {
...
     };
 
     setRootNodeIdToParent = () => {
-        const { categories } = this.props;
-
-        this.setState(({ rootNodeId }) => ({
-            rootNodeId: categories[rootNodeId].parentId
+        this.setState(() => ({
+            rootNodeId: this.state.parentId
         }));
```
jimbo: When using current state to construct the next state, use the first argument to the function here.

```es6
this.setState(({ parentId }) => ({ rootNodeId: parentId }))
```

packages/venia-concept/src/components/Navigation/navigation.js
```diff
@@ -207,7 +211,7 @@ class Navigation extends PureComponent {
         const { isCreateAccountOpen, isSignInOpen, rootNodeId } = state;
         const { classes, closeDrawer, isOpen, rootCategoryId } = props;
         const className = isOpen ? classes.root_open : classes.root;
-        const isTopLevel = !rootNodeId || rootNodeId === rootCategoryId;
+        const isTopLevel = !rootNodeId || rootNodeId == rootCategoryId;
```
jimbo: Generally we prefer strict equality. Does this address a particular requirement?

ericerway:
@serunde would love to get this updated and merged this week. Do you think you'll have a chance to review? Thanks so much for all of your help. 🙌

Serunde:
@ericerway Yes, I have a working version of this now, was just checking in with @jimbo for feedback on a couple details. Should be ready to go very soon!

packages/venia-concept/src/components/Navigation/navigation.js
Serunde: The ids being returned from graphql were strings. I've switched back to strict equality here (in latest changes), but I'm using `parseInt()` to set the `rootNodeId` now. Will that work, or should I be using a different method for conversion?

packages/venia-concept/src/components/Navigation/categoryTree.js
```diff
@@ -21,45 +24,81 @@ class Tree extends Component {
...
+            currentId
+        } = this.props;
 
-        const leaves = childNodeIds.reduce((elements, id) => {
-            const childNode = nodes[id];
-            const { childrenData, position } = childNode;
-            const isLeaf = !childrenData.length;
-            const elementProps = { nodeId: id, nodes };
+        return rootNodeId ? (
+            <Query query={navigationMenu} variables={{ id: rootNodeId }}>
```
jimbo: I would consider moving this query up one level, into `Navigation`.

packages/venia-concept/src/components/Navigation/categoryTree.js
```diff
@@ -21,45 +24,81 @@ class Tree extends Component {
...
-        if (nodes[rootNodeId].urlPath) {
-            leaves.push(
-                <li key={rootNodeId}>
-                    <Leaf
-                        nodeId={rootNodeId}
-                        nodes={nodes}
-                        onNavigate={onNavigate}
-                    >
-                        {({ node }) => `All ${node.name}`}
-                    </Leaf>
```
jimbo: The reason for the `children` function prop was this right here. I was rendering an `All {Dresses}` leaf. Did we remove this?

packages/venia-concept/src/components/Navigation/navigation.js
```diff
@@ -179,15 +180,23 @@ class Navigation extends PureComponent {
         }));
     };
 
-    setRootNodeId = rootNodeId => {
-        this.setState(() => ({ rootNodeId }));
+    setCurrentPath = currentPath => {
+        const path = currentPath.split('/');
+        const rootNodeId = parseInt(path[path.length - 1]);
```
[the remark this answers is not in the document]
Serunde: I like it!

packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
```diff
@@ -0,0 +1,196 @@
...
+                    ]
+                }
+            }
+        }
+    }
+];
+
+const classes = {
+    root: 'a',
+    inactive: 'b'
```
zetlen: Now that we've merged #552, you don't need this particular hack.

packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
```diff
@@ -0,0 +1,196 @@
...
+};
+
+function wait(callback) {
```
zetlen: You can replace this with the `waait` library we now have.

packages/venia-concept/src/components/Navigation/categoryLeaf.css
```diff
@@ -6,6 +6,11 @@
     width: 100%;
 }
 
+.inactive {
+    position: fixed;
```
zetlen: I worry that this combination will cause an invisible click target which shadows other targets. Have you made sure that inactive elements are totally out of the viewport in all cases?

packages/venia-concept/src/components/Navigation/categoryTree.js
zetlen: Jimmy's right! I also notice that you called the class `Tree`, rather than `CategoryTree`. It does seem like this contains a lot of generic tree logic, which might be applicable to more than just categories. Should it perhaps become a generic navigation tree which takes a render prop? That would make it even easier to detach the query from it, since the propTypes would also be detached.

packages/venia-concept/src/components/Navigation/categoryTree.js
```diff
@@ -21,45 +24,81 @@ class Tree extends Component {
...
 
-            const element = isLeaf ? (
-                <Leaf {...elementProps} onNavigate={onNavigate} />
-            ) : (
-                <Branch {...elementProps} onDive={updateRootNodeId} />
-            );
+                    const branches = [];
 
-            elements[position - 1] = <li key={id}>{element}</li>;
+                    const children = data.category.children.sort((a, b) => {
```
zetlen: I can't wait until Magento GraphQL ensures that ordering is significant, so we don't have to do this.

packages/venia-concept/src/components/Navigation/categoryTree.js
```diff
@@ -21,45 +24,81 @@ class Tree extends Component {
...
+
+                        return <li key={node.id}>{element}</li>;
+                    });
+
+                    const branchClass =
+                        currentId == rootNodeId
+                            ? classes.branch
+                            : classes.inactive;
+                    return (
+                        <div>
```
zetlen: I feel like this div should have a class, but if it's totally styleable without, then nevermind.

packages/venia-concept/src/queries/getCategoryList.graphql
```diff
@@ -3,9 +3,13 @@ query categoryList($id: Int!) {
     category(id: $id) {
         id
         children {
+            id
             name
             url_key
             url_path
+            product_count
```
zetlen: Are we using `product_count` anywhere? Can't find it in the source.

jimbo: I don't see it either. Removing it.

packages/venia-concept/src/components/Navigation/categoryTree.js
jimbo: It doesn't serve any purpose, so it shouldn't even be a `div`. Changing it to a `Fragment`.

packages/venia-concept/src/components/Navigation/categoryLeaf.css
jimbo: Yeah, we should revisit this as part of the upcoming accessibility epic.

packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
jimbo: True. There are a lot of tests written this way for now.

packages/venia-concept/src/components/Navigation/categoryTree.js
jimbo: Probably, but we can revisit it.

PR #546, opened 2018-11-26
Lighthouse ci

vitalics:
## This PR is a:

[ ] New feature
[X] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[X] Test for existing code
[ ] Documentation

## Summary

lighthouse ci integration

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ workflows:
 # configuration nodes for reuse.
 common_settings:
   docker: &docker_setup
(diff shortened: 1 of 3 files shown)


zetlen:
I have obtained a Lighthouse CI API key, @vitalics ! I want to integrate Lighthouse CI, but for now we need to limit it to the default branch and not to PRs, since PRs are not deployed.

zetlen:
We're going to go a different direction with this. The Lighthouse team will introduce a GitHub service at some point; of that we can be fairly sure.

PR #359, opened 2018-10-15
Product options

pcvonz:
## This PR is a:

[X] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will add configurable options to product pages. 

Fixes #278 and #283

## Additional information

I believe disabling certain options in the `<ProductEdit />` component can be its own issue, so I haven't added it into this pull request. Wrestled with a couple implementations, but I didn't want to delve too much deeper into it without discussing it further with other PWA contributors. I'll document what I know about the UX of the implementation in Storybook.

Issue #283, opened 2018-09-14
Add support for Configurable Products and option selection to Venia


zetlen:
## This issue is for the following packages:

[X] `venia-concept`
[ ] `pwa-buildpack`
[x] `peregrine`
[ ] `pwa-module`
[ ] `pwa-devdocs`

## Description

![image](https://user-images.githubusercontent.com/2686596/45492048-e8644a00-b730-11e8-8c7e-40f42c6242f9.png)

As a shopper, I want to choose a configurable product (colors, sizes) on Venia so that I can make an informed purchase.

### Acceptance Criteria
 - Configurable products configuration options
 - Components accept props for internal display state and provide hooks for event listeners
 - Storybook stories implemented for demonstration
 - Storybook states exist for selected, out of stock, and metadata display
 - Test coverage exists

#### Assumptions
- Completed MVP for Checkout of Venia working on local
- REST/GraphQL coverage to retrieve sample product data (configurable examples)

## Tasks
- [ ] Enhance GraphQL API calls to include option information
- [ ] Enhance REST API calls to Cart to include option selection
- [ ] Create state management modules (actions, reducers) for product configuration state and integrate
- [ ] Create container components which bind to API data and action creators
- [ ] Integration Option selector presentation components from #278 into containers
- [ ] Integrate containers into Product Detail page
- [ ] Testing and documentation (as appropriate)

zetlen:
Feel free to add Storybook to Venia for the purposes of the story! All dev deps without CLI commands (or annoying peer dependencies) have been pulled to the repo root `package.json` as of #223. So you could install `@storybook/react` in Venia and get it going pretty easily.

PR #550, opened 2018-11-27
Product options

### LINKED ISSUE #278: Implement Product Option Selector Presentation UI for Venia

[X] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-module`
[ ] `pwa-devdocs`

## Description

image

As a shopper, I want to choose a configurable product (colors, sizes) on Venia so that I can make an informed purchase.

### Acceptance Criteria
 - Venia prototype selectors are implemented from mock
 - Components accept props for internal display state and provide hooks for event listeners
 - Storybook stories implemented for demonstration
 - Storybook states exist for selected, out of stock, and metadata display
 - Test coverage exists

#### Assumptions
- Completed MVP for Checkout of Venia working on local
- REST/GraphQL coverage to retrieve sample product data (configurable examples)

## Tasks
- [ ] Evaluation of Option Selector prototypes that do exist for code reuse
- [ ] Implementation of presentation components (receiving props, not binding to API data) that implement Venia option selector UI
  - [ ] Numeric value selector (e.g. Size)
  - [ ] Color swatches
  - [ ] Text values
  - [ ] All options: Show out-of-stock indicator
- [ ] Testing and documentation (as appropriate)

jimbo:
## This PR is a:

- [x] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will enable shoppers to select values for configurable options and purchase configurable products.

## Additional information

- Discovery and initial work by @pcvonz 
- Rework of #359 
- Closes #278, #283

### DIFF
--- a/packages/venia-concept/src/RootComponents/Product/Product.js
+++ b/packages/venia-concept/src/RootComponents/Product/Product.js
@@ -1,12 +1,71 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
 import { bool, shape, number, arrayOf, string } from 'prop-types';
 
 import { addItemToCart } from 'src/actions/cart';
 import ProductFullDetail from 'src/components/ProductFullDetail';
 import getUrlKey from 'src/util/getUrlKey';
-import getProductDetail from '../../queries/getProductDetail.graphql';
+
+const query = gql`
+    query productDetail($urlKey: String, $onServer: Boolean!) {
+        productDetail: products(filter: { url_key: { eq: $urlKey } }) {
+            items {
+                sku
+                name
+                price {
+                    regularPrice {
+                        amount {
+                            currency
+                            value
+                        }
+                    }
+                }
+                description
+                media_gallery_entries {
+                    label
+                    position
+                    disabled
+                    file
+                }
+                ... on ConfigurableProduct {
+                    configurable_options {
+                        attribute_code
+                        attribute_id
+                        id
+                        label
+                        values {
+                            default_label
+                            label
+                            store_label
+                            use_default_value
+                            value_index
+                        }
+                    }
+                    variants {
+                        product {
+                            fashion_color
+                            fashion_size
+                            id
+                            media_gallery_entries {
+                                disabled
+                                file
+                                label
+                                position
+                            }
+                            sku
+                            stock_status
+                        }
+                    }
+                }
+                meta_title @include(if: $onServer)
+                meta_keyword @include(if: $onServer)
+                meta_description @include(if: $onServer)
+            }
+        }
+    }
+`;
 
 /**
  * As of this writing, there is no single Product query type in the M2.3 schema.
@@ -59,7 +118,7 @@ class Product extends Component {
     render() {
         return (
             <Query
-                query={getProductDetail}
+                query={query}
                 variables={{ urlKey: getUrlKey(), onServer: false }}
             >
                 {({ loading, error, data }) => {
--- a/packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
+++ b/packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
@@ -1,14 +1,15 @@
-import React, { Component } from 'react';
+import React, { Component, Suspense } from 'react';
 import { arrayOf, bool, func, number, shape, string } from 'prop-types';
 import { Price } from '@magento/peregrine';
 
 import classify from 'src/classify';
 import Button from 'src/components/Button';
 import Carousel from 'src/components/ProductImageCarousel';
-import Options from 'src/components/ProductOptions';
 import Quantity from 'src/components/ProductQuantity';
 import RichText from 'src/components/RichText';
 import defaultClasses from './productFullDetail.css';
+
+const Options = React.lazy(() => import('../ProductOptions'));
 
 class ProductFullDetail extends Component {
     static propTypes = {
@@ -129,8 +130,12 @@ class ProductFullDetail extends Component {
         }));
     };
 
+    get fallback() {
+        return <div>Loading...</div>;
+    }
+
     get productOptions() {
-        const { handleSelectionChange, props } = this;
+        const { fallback, handleSelectionChange, props } = this;
         const { configurable_options } = props.product;
         const isConfigurable = Array.isArray(configurable_options);
 
@@ -139,10 +144,12 @@ class ProductFullDetail extends Component {
         }
 
         return (
-            <Options
-                options={configurable_options}
-                onSelectionChange={handleSelectionChange}
-            />
+            <Suspense fallback={fallback}>
+                <Options
+                    options={configurable_options}
+                    onSelectionChange={handleSelectionChange}
+                />
+            </Suspense>
         );
     }
 
--- a/packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
+++ b/packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
@@ -51,19 +51,40 @@ class ProductFullDetail extends Component {
         addToCart: func.isRequired
     };
 
-    state = { optionSelections: new Map(), quantity: 1 };
+    static getDerivedStateFromProps(props, state) {
+        const { configurable_options } = props.product;
+        const optionCodes = new Map(state.optionCodes);
+
+        // if this is a simple product, do nothing
+        if (!Array.isArray(configurable_options)) {
+            return;
+        }
+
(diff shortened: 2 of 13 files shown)


packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -1,12 +1,71 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
 import { bool, shape, number, arrayOf, string } from 'prop-types';
 
 import { addItemToCart } from 'src/actions/cart';
 import ProductFullDetail from 'src/components/ProductFullDetail';
 import getUrlKey from 'src/util/getUrlKey';
-import getProductDetail from '../../queries/getProductDetail.graphql';
```
jimbo: These queries appear to be cached at some layer, and I couldn't figure out how to bust that cache. Inline queries don't appear to suffer from this problem.

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -1,41 +1,54 @@
...
+        if (onSelectionChange) {
+            onSelectionChange(attribute_id, selection);
+        }
     };
 
     get listComponent() {
-        return this.props.type === 'color' ? SwatchList : TileList;
+        const { attribute_code } = this.props;
+
+        return attribute_code === 'fashion_color' ? SwatchList : TileList;
```
jimbo: This is a weak point. We're going to need an explicit field, from the API, that tells us whether to use a swatch or not.

packages/venia-concept/src/components/ProductOptions/swatch.js
```diff
@@ -1,38 +1,67 @@
...
+    }
+
     render() {
-        const { classes, item } = this.props;
-        const { id, name } = item;
-        const style = { '--swatch-color': id };
+        const { icon, props } = this;
+        const {
+            classes,
+            hasFocus, // eslint-disable-line
```
jimbo: This is silly. We need to update our linter config so unused destructured props are acceptable when rest props are being declared.

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -1,12 +1,71 @@
...
+                            media_gallery_entries {
+                                disabled
+                                file
+                                label
+                                position
+                            }
+                            sku
+                            stock_status
+                        }
+                    }
```
jimbo: While it's necessary to fetch variants, this is expensive. It would be a big optimization to only fetch fewer variants.

packages/venia-concept/src/components/ProductOptions/option.js
pcvonz: I'll open an issue in the graphql-ce repository if you think that's necessary. :+1:

packages/venia-concept/src/RootComponents/Product/Product.js
zetlen: It looks like there's a big problem with `babel-plugin-import-graphql` and it's documented here: https://github.com/detrohutt/babel-plugin-import-graphql/blob/master/README.md#install You have to clear the babel cache every time you change a query.

That's unsustainable--we may have drop this dependency and rely on `graphql-tag/loader` in Webpack instead. That'll be annoying, because then we also have to add [this Jest mapper](https://github.com/remind101/jest-transform-graphql). But it's worth it to avoid the bug.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -49,18 +51,85 @@ class ProductFullDetail extends Component {
...
+            }));
+
+            const item = variants.find(({ product: variant }) => {
+                for (const [id, value] of optionSelections) {
+                    const { attribute_code: code } = configurable_options.find(
+                        ({ attribute_id }) => attribute_id === id
+                    );
+
+                    if (variant[code] !== value) {
+                        return false;
```
zetlen: This of course is the code we would rather the server do. It's `O(2^n)`, growing exponentially slower with each additional option and variant. Super dangerous. We should add a comment noting that we're aware of this and it's temporary.

If it's perversely slow for the time being, we could optimize this by lazy-building objects instead of arrays. We'll cross that bridge when we come to it.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -5,6 +5,7 @@ import { Price } from '@magento/peregrine';
 import classify from 'src/classify';
 import Button from 'src/components/Button';
 import Carousel from 'src/components/ProductImageCarousel';
+import Options from 'src/components/ProductOptions';
```
zetlen: Can we dynamically import this instead, on demand in the `productOptions` getter?

packages/venia-concept/src/components/ProductOptions/option.js
zetlen: Proper swatch API support is coming in https://github.com/magento/graphql-ce/pull/241. Let's leave a comment here to link to that issue.

packages/venia-concept/src/components/ProductOptions/swatch.js
zetlen: I sympathize, because it feels fast and simple to use a language-native way of blacklisting properties when copying an object. But I don't think we should use this idiom. It's very unintuitive to the newbie. Instead, can we pick a subset of props to pass?

packages/venia-concept/src/components/ProductOptions/tile.js
```diff
@@ -1,38 +1,39 @@
...
     };
 
     render() {
-        const { classes, item } = this.props;
-        const { name } = item;
+        const {
+            classes,
+            hasFocus, // eslint-disable-line
+            isSelected,
+            item,
```
zetlen: This shares a lot of code with `Swatch` and `SwatchList`. Should we consider a wrapper element or a List variant that has a focus concept?

pcvonz: I'm curious to know the reasoning behind why these are being expanded out again. In my original pull request these were accidentally merged back in (from the `default-pragma` branch). Instead of creating separate components, there was one component (`option.js`) which would load different CSS files based on the option type.

packages/venia-concept/src/components/ProductOptions/swatch.js
jimbo: Sure.

packages/venia-concept/src/RootComponents/Product/Product.js
jimbo: Good to know. Let's tackle separately. 👍

packages/venia-concept/src/RootComponents/Product/Product.js
jimbo: Sure. 👍

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -49,18 +51,85 @@ class ProductFullDetail extends Component {
...
-        this.props.addToCart({
-            item: this.props.product,
-            quantity: this.state.quantity
-        });
+    addToCart = () => {
+        const { props, state } = this;
+        const { optionSelections, quantity } = state;
+        const { addToCart, product } = props;
+        const {
+            __typename: productType,
```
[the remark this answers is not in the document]
jimbo: We can do that. 👍

packages/venia-concept/src/components/ProductOptions/option.js
jimbo: Done. 👍

packages/venia-concept/src/components/ProductOptions/swatch.js
jimbo: To clarify, we obviously won't whitelist which props get passed to the `button` element; that's why we're using rest props in the first place. I'll explicitly exclude `hasFocus`, though.

pcvonz:
Storybook is missing again. But I think that can be handled in a separate issue.

packages/venia-concept/src/components/ProductOptions/tile.js
jimbo: @zetlen They only seem to share code because they're implementations of the same interface. The shared code has already been consolidated in `List` and `Item`, in Peregrine.

@pcvonz When I [originally created](https://github.com/magento-research/pwa-studio/pull/62) `SwatchList`, `TileList`, and their respective children, the UI part of the product options story was basically complete. The rendered output of `Swatch` and `Tile` are different, so they're intentionally separate components.

PR #551, opened 2018-11-27
Fix category tile presentation

jimbo:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [x] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will fix the presentation of the category tiles.

## Additional information

Tiles should use `background-image`, so that we can use CSS to control the presentation and make use of imperfect images.

### DIFF
--- a/packages/venia-concept/src/components/CategoryList/categoryTile.css
+++ b/packages/venia-concept/src/components/CategoryList/categoryTile.css
@@ -6,22 +6,21 @@
 }
 
 .imageWrapper {
+    background-image: var(--venia-image);
+    background-position: 50% 50%;
+    background-size: cover;
     border-radius: 50%;
-    border: 1px solid rgb(var(--venia-border));
+    box-shadow: 0 0 0 1px rgb(var(--venia-border));
     display: block;
     height: 5rem;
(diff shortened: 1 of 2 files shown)


packages/venia-concept/src/components/CategoryList/categoryTile.css
```diff
@@ -6,22 +6,21 @@
...
 }
 
 .image {
-    border-radius: 50%;
-    height: auto;
-    max-height: 100%;
-    max-width: 100%;
-    width: auto;
+    height: 100%;
+    opacity: 0;
```
jimbo: Hide the image, but not from screen readers. It should be there, and if you right click the tile, the context menu should give you actions related to the image element (e.g., `View image`).

packages/venia-concept/src/components/CategoryList/categoryTile.css
```diff
@@ -6,22 +6,21 @@
 }
 
 .imageWrapper {
+    background-image: var(--venia-image);
+    background-position: 50% 50%;
+    background-size: cover;
```
jimbo: This is what we want.

PR #260, opened 2018-09-04
Checkout selectors (Issues #241 and #242)

JStein92:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will

- Address https://github.com/magento-research/pwa-studio/issues/241 and https://github.com/magento-research/pwa-studio/issues/242

- Let a shopper choose a shipping method and payment method pulled from Magento's REST API
- Add a generic component `Selector`, that renders `Sections` the user can choose from
- Moves ` subtotal` to shared component

Additional info

- We reworked form.js to use conditional rendering with `getters`

PR #556, opened 2018-11-28
Shipping and Payments

supernova-at:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will bring #260 up to date with the latest redux pipeline and form handling.

## TODO

- [x] Selecting a Payment Method actually updates the store
- [x] Selecting a Shipping Method actually updates the store
- [x] Update existing Unit Tests
- [x] Add new Unit Tests
- [x] Removing the last item from the cart causes a fatal error
- [x] Console warning about `select` needing a `value`
- [x] Shipping Information doesn't persist after closing mini cart
- [x] Losing some information when toggling the cart. `toggleCart` in `cart/asyncActions` is calling `dispatch(getCartDetails())`, and that's wiping out our state info. Note to self: this happens after submitting payment info
- [x] Submitting an order fails with message `Please check the billing address information`
- [ ] Update the price after submitting shipping method


supernova-at:
I am working on this one currently:

> The following unit tests did not pass 😔. All tests must pass before this PR can be merged
> packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js

packages/venia-concept/src/components/Checkout/shippingForm.js
```diff
@@ -0,0 +1,100 @@
...
+        const {
+            availableShippingMethods,
+            classes,
+            shippingMethod,
+            submitting
+        } = this.props;
+
+        // TODO: fix this hack that gets around React warnings about items having unique keys.
+        // We have to add a 'value' prop due to the Select component's getItemKey function.
+        const selectableShippingMethods = availableShippingMethods.map(
```
jimbo: This doesn't look like a hack to me. One way or another, you have to map your objects to `<option>` elements that have the correct `value` attribute, and they might as well key off of `value`, since it'll be unique for each option.

Realistically though, we should either

1. deprecate our `Select` component in favor of the one from `informed`, or
2. change our `Select` component to render the one from `informed`

Don't have to do that in this PR, though. And you'd still have to map data to `<option>` elements.

packages/venia-concept/src/components/Checkout/form.css
```diff
@@ -9,6 +9,15 @@
     composes: footer from './flow.css';
 }
 
+.informationPrompt {
+    color: orangered;
+    text-transform: capitalize;
+}
+.informationPrompt--disabled {
+    color: silver;
```
jimbo:
```suggestion
    color: rgb(var(--venia-text-hint));
```

packages/venia-concept/src/components/Checkout/form.css
```diff
@@ -9,6 +9,15 @@
     composes: footer from './flow.css';
 }
 
+.informationPrompt {
+    color: orangered;
```
jimbo:
```suggestion
    color: rgb(var(--venia-orange));
```

You should define this color as a custom property called `--venia-orange` in `src/index.css`, and set its value to `241, 99, 33`.

jimbo:
@supernova-at You'll also want to rebase this onto `release/2.0`.

jimbo:
Another comment, overall: the checkout components are getting too large and complex. That's not the fault of anyone working on this feature, either. Next week, we can discuss how to approach refactoring it.

packages/venia-concept/src/components/Checkout/shippingForm.js
supernova-at: I dig that, yeah.

Without having looked at its API yet, I vote for using the one from `informed` until we have a legitimate need for wrapping it.

packages/venia-concept/src/index.css
```diff
@@ -7,6 +7,7 @@ html {
     --venia-border: 224, 224, 224;
     --venia-font: Muli, -apple-system, BlinkMacSystemFont, sans-serif;
     --venia-grey: 246, 246, 246;
+    --venia-orange: 241, 99, 33;
```
jimbo: Look at you, respecting an alphabetized list. 😍

supernova-at: I was so excited when I noticed that about this codebase 😆. It's one of those things that always feels nitpicky to bring up but really makes reading and finding things so much easier!

supernova-at:
Okay, this guy is almost ready but needs a solution for the sequence of API calls to make during the checkout process.

# The Problem

Currently, the checkout flow is as follows:

| User Action | Resulting API Call | Notes |
| --- | --- | --- |
| Submit Shipping Address | `shipping-information` | `shipping-information` also wants a billing address and shipping method, so we hardcode those (to shipping address and "flatrate" respectively). |
| Submit Payment Method | `set-payment-method` | - |
| Submit Shipping Method | `collect-totals` | I picked this one because it's description says "Set shipping/billing methods and additional data for cart and collect totals for guest.". There is no `set-shipping-method` endpoint. |
| Submit Order | `order` | - |

This results in the following oddities:
  1.  The order always fails with a `200 Bad Request: Please check the billing address information` due to missing billing information.
  1. Refreshing the cart (by opening and closing the cart in the UI) after submitting payment information causes the user to have to start over again by entering their shipping address.

# The Culprit 

`set-payment-method` is the thing that wipes out the billing address information on the backend, causing `order` to fail and _refresh_ to return a cart with no address.

## How does LUMA do it?

The checkout flow in the [LUMA Demo](http://uxdemo.corp.magento.com/ce/) is as follows:

| User Action | Resulting API Call | Notes |
| --- | --- | --- |
| Set Shipping Address and Shipping Method | `shipping-information` | LUMA knows the list of available shipping methods by calling `estimate-shipping-methods` with a `country_id` of "US". |
| Set Payment Method | `payment-information` | This also places the order. Description from the API docs: _Set payment information and place order for a specified cart._ |

# Possible Solutions

Here are some possible solutions for the Venia checkout sequence:

## Option A

This is as close to the LUMA checkout process as possible.

### Overview

  1. `estimate-shipping-information`
  1. `shipping-information`
  1. `payment-information`

| User Action | Resulting API Call | Notes |
| --- | --- | --- |
| Create a Cart | `estimate-shipping-information` | Hardcode `country_id` = "US" |
| Submit Shipping Address | n/a | Save the shipping address to the store or storage but don't actually POST to `shipping-information` until we also have a shipping method. |
| Submit Payment Method | n/a | Save the payment method to the store or storage but don't POST to `set-payment-information`. |
| Submit Shipping Method | `shipping-information` | Submit the saved address from `Submit Shipping Address` and the selected shipping method. |.
| Submit Order | `payment-information` | Submit the saved payment method from `Submit Payment Method` to place the order. |

### Downsides

  1. Since our UX is different from LUMA's, we have to store some information in the client (in the Redux store), and not communicate with the server after each step. This introduces the potential to lose / wipe out information if we refresh the cart from server data.
  1. The hardcoded `country_id` of "US" in the `estimate-shipping-methods` call is probably fine for now, but we'll want to update this based on the user's actual location in the future.

## Option B

This option deviates from LUMA in the way we submit payment information.

### Overview

  1. `estimate-shipping-information`
  1. `set-payment-method`
  1. `shipping-information`
  1. `order`

### Details

| User Action | Resulting API Call | Notes |
| --- | --- | --- |
| Create a Cart | `estimate-shipping-information` | Hardcode `country_id` = "US" |
| Submit Shipping Address | n/a | Save the shipping address to the store or storage but don't actually POST to `shipping-information` until we also have a shipping method. |
| Submit Payment Method | `set-payment-method` | We don't have to worry about wiping out an existing saved billing address because we haven't submitted one yet. |
| Submit Shipping Method | `shipping-information` | Submit the saved address from `Submit Shipping Address` and the selected shipping method. |.
| Submit Order | `order` | Place the order. |

### Downsides

  1. Since our UX is different from LUMA's, we have to store some information in the client (in the Redux store), and not communicate with the server after each step. This introduces the potential to lose / wipe out information if we refresh the cart from server data.
  1. The hardcoded `country_id` of "US" in the `estimate-shipping-methods` call is probably fine for now, but we'll want to update this based on the user's actual location in the future.

## Option C

This option acknowledges that we're going to lose billing information and re-submits it.

### Overview

  1. `estimate-shipping-information`
  1. `shipping-information`
  1. `set-payment-information`
  1. `shipping-information`
  1. `order`

### Details

| User Action | Resulting API Call | Notes |
| --- | --- | --- |
| Create a Cart | `estimate-shipping-information` | Hardcode `country_id` = "US" |
| Submit Shipping Address | `shipping-information` | We need to save the address to the store or storage here so that we can re-submit it later. `shipping-information` also wants a shipping method, so we hardcode it to "flatrate" in this call. |
| Submit Payment Method | `set-payment-method` | This will wipe out the billing address we just submitted. |
| Submit Shipping Method | `shipping-information` | Submit the saved address along with the user's choice of shipping method |.
| Submit Order | `order` | Place the order. |

### Downsides

  1. More calls to the server.
  1. We're submitting to `shipping-information` multiple times, which feels hacky and incorrect.
  1. We have to store some information in the client (in the Redux store) which introduces the potential to lose / wipe out information if we refresh the cart from server data.
  1. The hardcoded `country_id` of "US" in the `estimate-shipping-methods` call is probably fine for now, but we'll want to update this based on the user's actual location in the future.

## Option D

This option submits the payment method and shipping method using the `collect-totals` endpoint.

### Overview

  1. `estimate-shipping-information`
  1. `shipping-information`
  1. `collect-totals`
  1. `order`

| User Action | Resulting API Call | Notes |
| --- | --- | --- |
| Create a Cart | `estimate-shipping-information` | Hardcode `country_id` = "US" |
| Submit Shipping Address | `shipping-information` | `shipping-information` also wants a billing address and shipping method, so we hardcode those (to shipping address and "flatrate" respectively). |
| Submit Payment Method | n/a | Save the payment method to the store or storage but don't send anything to the server. |
| Submit Shipping Method | `collect-totals` | Submit the saved payment from `Submit Payment Method` and the selected shipping method to get new order totals that include shipping costs. |.
| Submit Order | `order` | Place the order. |

### Upsides

  1. We could probably use `collect-totals` as part of the other options too but this will update the cart total to include shipping costs based on the selected shipping method.

### Downsides

  1. We have to store the payment method in the client (in the Redux store), and not send it to the server. This introduces the potential to lose / wipe out information if we refresh the cart from server data.
  1. The hardcoded `country_id` of "US" in the `estimate-shipping-methods` call is probably fine for now, but we'll want to update this based on the user's actual location in the future.

zetlen:
@supernova-at That was the best solutioneering comment I've ever seen. Just wanted to mention that before anything else. I'll look into a couple API options and then come back with my personal preference.

supernova-at:
> @supernova-at That was the best solutioneering comment I've ever seen.

</img>

zetlen:
@supernova-at Since all of your proposed solutions involve keeping some data in the Redux store, we should conclude that we'll have to do that. We have a little hack for cart images which uses a `BrowserPersistence` utility (a wrapper around LocalStorage that we can later refactor to use better APIs), and I think that we should experiment with storing that client-only order config in a BrowserPersistence instance on order ID. It's not great, but it looks like we'll have to do that.

All of your proposed solutions also mention the hardcoded `country_id` as a drawback. I agree that this is gross, but since all the possible solutions seem to require it, we can omit it from our judgment as well.

So, looking past those inevitabilities, your options have these drawbacks:

A) None

B) Deviates from the known-good Luma checkout process

C) Additional and repetitive network traffic

D) Deviates from the known-good Luma checkout process

Is there a compelling reason to go for Options B or D instead of A? They seem to all be functionally equivalent if we assume that the quote endpoint will accept these updates in any order. If we're gonna deviate from Luma, *we should make sure that the quote endpoint will always accept these any order.* If we go with Option A, we don't have to do that.

Thoughts?

supernova-at:
> Since all of your proposed solutions involve keeping some data in the Redux store, we should conclude that we'll have to do that.

This is a good point that I didn't realize was the case until I enumerated them.

> Is there a compelling reason to go for Options B or D instead of A?

I don't think so.

In addition to ensuring we could deviate from LUMA from an API perspective (the quote endpoint), I was unsure whether we should. The only real reason we would is because our UX is different. If we can solve for that using `BrowserPersistence`, then I think Option A (replicating LUMA) is our best bet.

jimbo:
Venia's checkout flow was optimized for user experience, in terms of capturing user input on a mobile device. Submitting data to the API at every step is not required, just as it's not required on Luma today. I'm fine with making the same calls Luma makes (`Option A`).

supernova-at:
This PR is ready for final review.

I should mention one small remaining oddity, which is that the app will "remember" the user's inputs in the checkout flow across refreshes for address and payment method but not shipping method.

To reproduce this:
  1. Enter the checkout flow and enter input for address, payment method, and shipping method
  1. Refresh the page
  1. Open the cart
  1. See that the address and payment method are present, but the shipping method needs to be selected again

This is because we're storing the address and payment method in local storage (preserved across refreshes), but not the shipping method. The shipping method gets submitted to the API (`shipping-information`), but it does not come back to us when we request the cart's details.

supernova-at:
I have fixed the shipping method oddity described in my previous comment.

supernova-at:
PR Updated:
  * Addresses last bit of code feedback: minor fixes and cleanups

PR #581, opened 2018-12-03
Vitali/bundlesize

vitalics:
## This PR is a:

- [x] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

1. Adds a dependency on [bundlesize](https://www.npmjs.com/package/bundlesize), a build tool that can check files to ensure they are under a given size. 
2. Adds an npm script, `bundlesize` for e-z size checking.
3. Use bundlesize npm command to check build size in CI.
4. Adds a pre-commit hook to check bundlesize.

[Image from Gyazo](https://gyazo.com/789791c48f47616b12f5782da90ac95b)

## Additional information

`bundlesize@0.16.0` and on has a dependency on `brotli-size` which requires `iltorb` binary be installed or found. This fails in docker which is sort of documented in an issue [here](https://github.com/siddharthkp/bundlesize/issues/202). To avoid this issue we can use `0.15.x` instead.

bundlesize's real benefit of is the github/ci integration. Any of us could write a script that checks some filesize and ensures it hasn't exceeded some limit. So my 

**Open question to the team**: Do we want this CI step or do we want to just write some script that checks on pre-commit or pre-push?


zetlen:
@vitalics I added the token as you requested, then pushed a couple of changes to use the token and produce real output. Thanks for experimenting with this. I'm not sure we need it, though; it's nice to see it as a GitHub check, but [Webpack Performance Budgets](https://webpack.js.org/configuration/performance/#performance-maxassetsize) are built in and we can use Webpack's own build to fail when a bundle is too large. Is there an additional feature of `bundlesize` that I'm not noticing?

vitalics:
@zetlen unfortunately, Webpack Performance Budgets are not integrated with GitHub natively, and for solve this, need to write integration layer, it's better to use an existing solution, to reduce unexpected bugs, GitHub api researching etc.
Of course, we can use another solution, but it takes time for further integration with service, and the faster solution is to use `bundlesize` library since it faster and working solution for us

vitalics:
@zetlen before the merge, please squashed them or click the `squash and merge` button
image
, thanks

sharkySharks:
RE: 
> I'm trying to understand why this was necessary.

Honestly, adding another dependency without a clear idea of the benefit makes me question the pr. Do we need this change? I'm not sure we do.

.circleci/config.yml
```diff
@@ -75,8 +75,9 @@ jobs:
       - run:
           name: DangerCI
           command: npm run danger
+      - run:
+          name: Bundle size analysis
+          command: npx -p iltorb bundlesize --debug
```
sharkySharks: can we get this command in a package.json script? i'll be porting over functionality to a different CI platform in a few weeks and it is also better practice to have these commands in a reusable place, rather than straight in the config.

sirugh:
> Honestly, adding another dependency without a clear idea of the benefit makes me question the pr. Do we need this change? I'm not sure we do.

:) Sure. That's why I'm investigating this PR and trying to understand whats included. What I've found so far is that `iltorb` is a requirement of `brotli-size` which is the library that `bundlesize` uses to calculate brotli compression stats. You can see an open issue regarding this very thing [here](https://github.com/siddharthkp/bundlesize/issues/202).

Looking into that issue:

> If iltorb isn't able to find a pre-compiled binary, it has to build using node-gyp, which requires Python and g++ be installed - which often aren't in Docker images, or in Windows environments.

So I think @vitalics added that LOC to compile the iltorb binary for `brotli-size`, which is fine -- I just would like to have a more detailed comment explaining that OR try to find some way to include the `iltorb` dependency in the `npx` command, like `npx -p iltorb bundlesize --debug` (which [failed](https://circleci.com/gh/magento-research/pwa-studio/2747) fwiw).

.circleci/config.yml
sirugh: Sure.

sharkySharks:
Cool 😎

screen shot 2019-01-17 at 12 40 04 pm

Do we only care about the venia-concept bundlesize?

LGTM - we just want to make sure to squash all those commits 
screen shot 2019-01-17 at 12 41 41 pm

sharkySharks:
this could easily be added to a prepush hook, is the expectation that folks should reduce bundlesize before submitting a PR? then prepush would be a good place for it. i think we want to keep this in CI, but if you also wanted to add it to prepush because you want devs to get that feedback sooner than in a PR then you can add it there as well.

sirugh:
> Do we only care about the venia-concept bundlesize?

I think so? It's the only runtime/browser asset AFAIK.

> this could easily be added to a prepush hook

Done, though I added it as a pre-commit hook, which the team may find encumbering. I say we give it a try and see how it feels.

Issue #553, opened 2018-11-27
Add variant information to cart item display


zetlen:
## This issue is for the following packages:

[x] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-devdocs`
[ ] `upward-js`
[ ] `upward-spec`

## This issue is a:

[ ] Bug
[x] Feature suggestion
[ ] Documentation issue
[ ] Other (Please Specify)

## Description

With the merging of #550 the storefront supports configurable products, but the cart item display doesn't reflect the chosen configuration:

![image](https://user-images.githubusercontent.com/1643758/49111379-8b6b3100-f255-11e8-98db-94640412d0a0.png)

As a shopper, I want to see my chosen configuration options in my cart, so I can keep track of what I'm purchasing.

@soumya-ashok has made a visual comp: https://magento.invisionapp.com/share/HNPBRBHP29D#/screens

ericerway:
Assigned to @JStein92 - let's groom this in realtime based on your interested in taking it.

ennostuurman:
hey @JStein92 I have created PR #582 for this . Understood from @ericerway that you'd been doing work for this already as well which I wasn't aware off.  I definitely would like to compare our results or maybe we could even combine both?

ericerway:
I'd love for @JStein92 to finish this out but get a quick review from @ennostuurman who appeared to show interest at the same time in an effort to work on a quick starter story. We'll find more "good first issue" stories for everyone -- good problem to have. Thanks!

JStein92:
Ok, just pushed a prettier pass to #558 (whoops) - looking forward to a review :)

PR #582, opened 2018-12-03
Feature/553 add variant to minicart item

ennostuurman:
## This PR is a:

[x] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will add product variants to the minicart item details dynamically , see issue #553 . 

## Additional information

- this is a first setup, there's some overlap between the added totals items and the item details prop (i.e. item_id, price and product name) , so maybe code can be more optimized, on the other hand options are only available from the totals response object and items details contains images reference etc.. 
- removed mockdata
- did some additional styling as per mockup as well. See screenprint.
- if first setup is okay then next step is to add tests.

magento-pwa-studio-issue-553-minicart-items-variants-revised


packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -47,23 +72,33 @@ class Product extends Component {
...
-            </dl>
-        ) : null;
+        const { classes, item, totalsItems } = this.props;
+        const { item_id } = item;
+        const regex = /[\[\]']+/g; // regex to to remove [] , use it to check if options are really empty
+        return totalsItems
+            .filter(totalsItem => totalsItem.item_id === item_id)
+            .map(totalsItem => {
+                const { options } = totalsItem;
+                return options && options.replace(regex, '').length > 0 ? (
```
supernova-at: I'm a little confused by the need for `regex`. Is `options` a string or an array?

Looking at [the Magento REST API documentation](https://devdocs.magento.com/swagger/) (`GET /V1/guest-carts/{cartId}/totals`), it should be a `string`:

```
"options": "string",
```

This code makes me assume that `options` can sometimes be a stringified version of an array? With values like "['foo', 'bar', 'baz']" ?

If that's the case, and you're trying to determine if options is empty as the comment on line 77 indicates, I think a simple `options === '[]'` could reduce the complexity here.

In other words, this line could become

```js
return options && options !== '[]'
```

and we could drop the regex.

Alternately, you could `JSON.parse(options)` ahead of time and use the parsed version throughout this section. That might be even better.

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -47,23 +72,33 @@ class Product extends Component {
...
+                        ))}
+                    </dl>
+                ) : null;
+            });
     }
 
     styleImage(image) {
         return {
-            height: imageHeight,
+            minHeight: imageHeight, // min-height instead of height so image will always align with grid bottom
```
supernova-at: I don't have a problem with setting `min-height` but do we want to still set `height` also?

Also it's a little bit of a red flag that we're relying on heights to properly align things. We should use [align-items](https://css-tricks.com/snippets/css/complete-guide-grid/#prop-align-items) for alignment.

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -63,6 +63,7 @@ class MiniCart extends Component {
                 showEditPanel={this.showEditPanel}
                 currencyCode={cartCurrencyCode}
                 items={cart.details.items}
+                totalsItems={cart.totals.items}
```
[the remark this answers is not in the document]
ennostuurman: @supernova-at 

Both are currently fetched with the M2 REST API , see /packages/venia-concept/src/actions/cart/asyncActions . There's some overlap between the two but we also need both to be able to display images (only available in details.items) and options (only available in totals.items).

As I see it, the cart.details.items contains the more basic information for card display and actions, like id, sku, price and image. The totals items contains more detailed information that you could use to enrich cart display or to add functionality (totals.items contains a.o. the options used here but also tax amount etc..).

packages/venia-concept/src/components/MiniCart/product.js
ennostuurman: This setup is mainly caused by the fact that we use grid here and the images are background images. So we get the following:

- Setting an explicit height to the image will restrict the image, so it won't fill the complete grid area. Screenprint: explicit height:
magento-pwa-studio-pull-582-image-height-instead-of-min-height
- Setting no height, won't always render the complete image (for example in case of simple products). screenprint: no height 
magento-pwa-studio-pull-582-image-no-height
- so min-height seems the best fit here.

packages/venia-concept/src/components/MiniCart/product.js
ennostuurman: yeah much better, updated accordingly, thx for the suggestions!

packages/venia-concept/src/components/MiniCart/product.js
supernova-at: Thanks for the explanation 👍

JStein92:
@supernova-at @ennostuurman No worries, I will wait to rebase on release/2.0!

ennostuurman:
@supernova-at thank you! Appreciate your review and helpful comments. 

I resolved the failing test, and added a variant test. However commented it out for now because the test fails where it shouldn't. I will try to fix this, but maybe you or someone could help me or review it so we get that fixed as well.

And with regard to the PR confusion, no problems. I picked the issue up because of the "help wanted" label and the fact that the issue hadn't been assigned yet according to the Github GUI. So assigning the issue right after it has been promised to someone might prevent this. But I don't complain, this is a voluntary contribution to Open Source happy to able to contribute and learn at the same time ;-)  And the advantage of two solutions is that it is interesting to see two different approaches that maybe can complement each other.

ennostuurman:
- did some additional trials to fix variants test - not succeeded (yet)
- updated PR with latest staging, build deploy fails now likely because of missing jest-transform-graphql , see https://github.com/magento-research/pwa-studio/commit/992e4ecfc02a6a3f80e87838d1fd44ec45638f73

ennostuurman:
@supernova-at thx for updating with latest staging. I see that build deploy still fails. On my end everything seems fine after running new npm install because of new  jest-transform-graphql package. Maybe that's the build issue.

Let me know if that's the case or if I should fix something on my end.

supernova-at:
Hi @ennostuurman ,

> did some additional trials to fix variants test - not succeeded (yet), see additional comments in product.spec.js

I had to fiddle around with it a bit, but [here's a potential solution](https://gist.github.com/supernova-at/20bc07731ce34a9d63f8f863dfa1c322), let me know what you think.

Note that I edited `totalsItems.options` and the `Product variants are rendered` test.

> updated PR with latest staging, build deploy fails now likely because of missing jest-transform-graphql , see 992e4ec

I saw this same error and was able to resolve by running

```
rm -rf node_modules
npm install
```

in the root directory of `pwa-studio`. Please try this and let me know if it resolves your issue.

supernova-at:
> On my end everything seems fine after running new npm install 

🎉 😃

ennostuurman:
@supernova-at implemented the tests as you suggested. All tests pass, so thank you for this fix!

I rebased branch with latest staging and pushed again. 

Build deploy fails but that should be a matter of updating node_module on the build server.

ennostuurman:
@supernova-at all checks passed (needed to run prettier once more ;-) !

supernova-at:
Thanks for your contribution, @ennostuurman !

PR #586, opened 2018-12-03
Update docs

sharkySharks:
Small changes to update the documentation and to state the upper bound of the node engine version in package.json

## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[X] Documentation

## Summary

When this pull request is merged, it will update documentation and the node engine field in package.json

### DIFF
--- a/pwa-devdocs/README.md
+++ b/pwa-devdocs/README.md
@@ -23,13 +23,13 @@ Use the following instructions to build a local instance of the documentation.
     ```
     cd pwa-studio
     ```
-3.  Run the following command to install the required node modules for the project:
-    ```
-    npm install
-    ```
-4.  Change into the **pwa-devdocs** directory using the following command:
+3.  Change into the **pwa-devdocs** directory using the following command:
     ```
     cd packages/pwa-devdocs
+    ```
+4.  Run the following command to install the required node modules for the project:
+    ```
+    npm install
     ```
 5.  Run the following command to install the required Ruby libraries:
     ```
--- a/pwa-devdocs/package-lock.json
+++ b/pwa-devdocs/package-lock.json
@@ -1,0 +1,8348 @@
+{
+  "name": "pwa-devdocs",
+  "version": "2.0.0-rc2.0.5",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/node": {
+      "version": "10.12.12",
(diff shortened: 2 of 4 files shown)


pwa-devdocs/README.md
```diff
@@ -23,13 +23,13 @@ Use the following instructions to build a local instance of the documentation.
...
+3.  Change into the **pwa-devdocs** directory using the following command:
     ```
-    npm install
+    cd packages/pwa-devdocs
     ```
-4.  Change into the **pwa-devdocs** directory using the following command:
+4.  Run the following command to install the required node modules for the project:
     ```
-    cd packages/pwa-devdocs
+    npm install
```
supernova-at: Does it not work the way it was previously described?

We have a [Dependency Management](https://github.com/magento-research/pwa-studio/wiki/Dependency-Management) doc in the Wiki that states:

> The NPM scripts are designed so that nearly all operations, even ones for individual packages, should be run from the root directory.

package.json
```diff
@@ -9,7 +9,7 @@
     "url": "https://github.com/magento-research/pwa-studio/issues"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=8.0.0 <=11.0.0"
```
supernova-at: Should this be `<=9.0.0` ?

I know there's still work to be done to get us fully valid on Node 10 - should an update to `<= 11.0.0` wait until that work is complete?

sharkySharks: Not sure, going by what @zetlen is saying for the project. right now without the addition of the upper limit we're saying that anything above 8 is working, so need some clarity on that.

supernova-at: Totally agree, I'm just pointing out that I don't know if we know that 10 is working. Maybe we set our upper limit to 9 until we're sure.

sharkySharks: I am running 11 and it is working, but that is part of the issue is that right now everyone has different local environments so it is difficult to confirm any of these versions are working as expected. How are we determining a version is cleared as working?

pwa-devdocs/README.md
jcalcaben: The doc site is a unique project in the mono repo and is self contained. So running `npm install` inside the docs directory makes sense and is the correct approach in my opinion.

PR #588, opened 2018-12-03
add configurable information to minicart item display - issue #553

JStein92:
## This PR is a:

[ x ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

When this pull request is merged, it will add configurable options to the minicart item display.

## Additional information

I added a max height + scroll to the configurable options block. So if an item has 4+ options, it will scroll instead of making the entire item display larger.

I also changed it so the quantity block is flex-end, to make it consistent with @soumya-ashok's mockup for configurable items.

image


packages/venia-concept/src/actions/cart/asyncActions.js
```diff
@@ -231,9 +231,11 @@ export const getCartDetails = (payload = {}) => {
 
             // for each item in the cart, look up its image in the cache
             // and merge it into the item object
+            // then assign its options from the totals subResource
             if (imageCache && Array.isArray(items) && items.length) {
-                items.forEach(item => {
+                items.forEach((item, i) => {
                     item.image = item.image || imageCache[item.sku] || {};
+                    item.options = totals.items ? totals.items[i].options : [];
```
supernova-at: Are we guaranteed that the order of entries in `details.items` is the same as the order in `totals.items`?

If not, we might be mixing up `options` here.

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -48,19 +50,26 @@ class Product extends Component {
 
     get options() {
         const { classes, item } = this.props;
+        const options = item.options ? JSON.parse(item.options) : [];
 
-        return item.options && item.options.length > 0 ? (
+        return options && options.length > 0 ? (
             <dl className={this.props.classes.options}>
```
supernova-at: I know you didn't change this line but `classes` has already been destructured out of `this.props` and can be used directly here.

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -48,19 +50,26 @@ class Product extends Component {
...
-                {item.options.map(({ name, value }) => (
-                    <Fragment key={name}>
-                        <dt className={classes.optionName}>{name}</dt>
-                        <dd className={classes.optionValue}>{value}</dd>
+                {options.map(({ label, value }) => (
+                    <Fragment key={`${name}${value}`}>
```
supernova-at:
```suggestion
                    <Fragment key={`${label}${value}`}>
```

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -48,19 +50,26 @@ class Product extends Component {
...
+                        </dt>
                     </Fragment>
                 ))}
             </dl>
         ) : null;
     }
 
+    get modal() {
+        const { classes } = this.props;
+        return this.state.isOpen ? <div className={classes.modal} /> : null;
```
supernova-at: Be sure to add `classes.modal` to `propTypes` as well 👍

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -86,19 +95,24 @@ class Product extends Component {
...
-                        readOnly
-                    >
-                        <option value={item.qty}>{item.qty}</option>
-                    </select>
-                    <span className={classes.quantityOperator}>{'×'}</span>
-                    <span className={classes.price}>
-                        <Price currencyCode={currencyCode} value={item.price} />
-                    </span>
+                    <div className={classes.quantityRow}>
+                        <select
```
supernova-at: We do have a [Select](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/Select/select.js) component if you need it.

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -86,19 +95,24 @@ class Product extends Component {
...
+                        <span className={classes.price}>
+                            <Price
+                                currencyCode={currencyCode}
+                                value={item.price}
+                            />
+                        </span>
+                    </div>
                 </div>
-                <div className={this.state.isOpen ? classes.modal : ''} />
+                {modal}
```
supernova-at: Nice 👍

packages/venia-concept/src/actions/cart/asyncActions.js
JStein92:
```suggestion
                    item.options = totals.items.find(t => t.item_id === item.item_id).options;
```
Interesting point - although it does seem to always be in the same order, I suppose I don't know for sure that it will always be. How about matching the item Id instead?

packages/venia-concept/src/components/MiniCart/product.js
JStein92: good catch!

packages/venia-concept/src/components/MiniCart/product.js
JStein92: ah, thank you for catching that

packages/venia-concept/src/components/MiniCart/product.js
JStein92: I actually didn't add/change the selector, just added a wrapper around it for formatting. If it's OK with you, I'd like to leave the Select component change to another PR.

JStein92:
@supernova-at - thank you for the feedback. I've included a merge of release/2.0 into the latest commit. Lmk what you think!

packages/venia-concept/src/actions/cart/asyncActions.js
```diff
@@ -241,9 +241,16 @@ export const getCartDetails = (payload = {}) => {
 
             // for each item in the cart, look up its image in the cache
             // and merge it into the item object
+            // then assign its options from the totals subResource
             if (imageCache && Array.isArray(items) && items.length) {
                 items.forEach(item => {
                     item.image = item.image || imageCache[item.sku] || {};
+
+                    item.options =
+                        totals && totals.items
```
supernova-at: Since this conditional doesn't change with each `item`, you could pull this out of the `forEach` if you wanted. Doesn't block a 👍 though.

JStein92: Oh of course! Ok, I'll make that quick change and get this pushed. Thanks again!

JStein92: @supernova-at should be good to go now!

supernova-at:
On `hold` until #582 lands.

supernova-at:
Hi @JStein92 , #582 has just been merged and I have removed the `hold` label from this PR 🎉 

A few quick thoughts:
  * I like your approach of merging the `options` in `cart/asyncActions.js` as opposed to having the React application worry about it. This definitely seems like a data-layer concern.
  * I'd love for `MiniCart/product.js` to not have to know that it needs to `JSON.parse` the item's `options`. What do you think about `parse`ing them in `cart/asyncActions.js`?
  * Please do keep (and possibly update) the unit test
  * There are some conflicts to resolve

Thanks - let me know if you have any questions 👍

JStein92:
@supernova-at implemented your suggested changes - let me know what you think!

packages/venia-concept/src/actions/cart/asyncActions.js
```diff
@@ -241,9 +241,18 @@ export const getCartDetails = (payload = {}) => {
...
             if (imageCache && Array.isArray(items) && items.length) {
+                const validTotals = totals && totals.items;
                 items.forEach(item => {
                     item.image = item.image || imageCache[item.sku] || {};
+
+                    item.options = validTotals
+                        ? JSON.parse(
+                              totals.items.find(t => t.item_id === item.item_id)
+                                  .options
+                          )
```
supernova-at: There are three layers of Inception here (😛 ), let's reduce the mental overhead by using an `if` statement instead of a ternary (`?`). Something like:

```js
let options = [];
if (validTotals) {
  const matchingItem = totals.items.find(t => t.item_id === item.item_id);
  if (matchingItem && matchingItem.options) {
    options = JSON.parse(matchingItem.options);
  }
}
item.options = options;
```

packages/venia-concept/src/components/MiniCart/__tests__/product.spec.js
```diff
@@ -7,15 +7,25 @@ import Section from '../section';
...
 
 const item = {
     item_id: 1,
     name: 'Product 1',
     price: 10,
     qty: 1,
     sku: 'TEST1',
-    image: 'test.jpg'
+    image: 'test.jpg',
+    options: [
```
supernova-at: > Since these options will be present on the `item` because we merged them from `totals` in the data-layer, you can remove `totalsItems` from `product.js` entirely.

As a corollary to this, we can remove all references to `totalsItems` from this file too 👍

packages/venia-concept/src/components/MiniCart/__tests__/product.spec.js
```diff
@@ -61,33 +71,36 @@ test('passed functions are called from nested `Section` components', () => {
...
-                .at(index)
-                .text()
-        ).toContain(option.label);
-        expect(
-            wrapper
-                .find('dd')
-                .at(index)
-                .text()
-        ).toEqual(option.value);
+    wrapper.find(`.${classes.optionLabel}`).forEach((l, i) => {
```
supernova-at: I'm fine with `i`, but let's change `l` to something more meaningful. I'm assuming it means `label`?

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -15,19 +15,26 @@ class Product extends Component {
...
         }),
         item: shape({
             item_id: number.isRequired,
             name: string.isRequired,
+            options: arrayOf(
+                shape({
+                    label: string,
+                    value: string
+                })
+            ),
```
supernova-at: Since these options will be present on the `item` because we merged them from `totals` in the data-layer, you can remove `totalsItems` from `product.js` entirely.

packages/venia-concept/src/actions/cart/asyncActions.js
JStein92: Haha good idea - ternaries can get a bit hard to read.

supernova-at:
> @supernova-at thank you for the kind feedback and patience! I've made the requested changes. 🎉

No, thank *you*!

PR #589, opened 2018-12-03
Remove underscore

jimbo:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will remove the `underscore` dependency. It was only pulled in for `debounce` in a single component (`CreateAccount`), and we already have `lodash.debounce`.


packages/venia-concept/src/components/CreateAccount/__tests__/createAccount.spec.js
```diff
@@ -41,13 +41,7 @@ jest.mock('@magento/peregrine', () => {
     };
 });
 
-jest.mock('underscore', () => {
-    return {
-        debounce: cb => {
-            return cb;
-        }
-    };
-});
```
sharkySharks: I love when there are more deletions that additions in a 🔴PR 🔴

PR #593, opened 2018-12-04
Remodel input, adjust components which uses it

AlexeyKaryka:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[ ] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

Update Input component, adjust createAccount, signIn

## Summary

### DIFF
--- a/packages/venia-concept/src/components/Checkout/formContainer.js
+++ b/packages/venia-concept/src/components/Checkout/formContainer.js
@@ -1,0 +1,14 @@
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+import classify from 'src/classify';
+import Form from './form';
+import defaultClasses from './form.css';
+
+const mapStateToProps = ({
+    checkout: { isAddressIncorrect, incorrectAddressMessage }
+}) => ({ isAddressIncorrect, incorrectAddressMessage });
+
+export default compose(
+    classify(defaultClasses),
+    connect(mapStateToProps)
+)(Form);
--- a/packages/venia-concept/src/components/Input/helpers.js
+++ b/packages/venia-concept/src/components/Input/helpers.js
@@ -1,0 +1,15 @@
+import { isString } from 'util';
+
+const emptyFieldMessage = 'This field cannot be empty';
+export const isRequired = value =>
+    isString(value) && value.length > 0 ? null : emptyFieldMessage;
+
+export const createComplexValidator = validators => (value, values = null) => {
+    for (let validator of validators) {
+        const validatorOutput = validator(value, values);
+        if (validatorOutput) {
+            return validatorOutput;
+        }
+    }
+    return null;
+};
--- a/packages/venia-concept/src/components/Input/input.js
+++ b/packages/venia-concept/src/components/Input/input.js
@@ -1,29 +1,30 @@
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
+import { compose } from 'redux';
 import { PropTypes } from 'prop-types';
+import { BasicText, asField } from 'informed';
 import defaultClasses from './input.css';
 import classify from 'src/classify';
-import { Text } from 'informed';
+import Icon from 'src/components/Icon';
+import { resetButtonIcon, HelpTypes } from './constants';
 
-export const HelpTypes = {
-    hint: 'hint',
-    error: 'error',
-    success: 'success'
-};
-
-class Input extends Component {
+//TODO: try to incapsulate default validatorRequired in Input component
+// so you don't need to pass validator apart from required prop
+export class Input extends Component {
     static propTypes = {
         classes: PropTypes.shape({
-            helpText: PropTypes.string,
             hint: PropTypes.string,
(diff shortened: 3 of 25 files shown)


packages/venia-concept/src/components/Input/helpers.js
```diff
@@ -0,0 +1,22 @@
...
+    for (let i = 0; i < validators.length; i++) {
+        const validatorOutput = validators[i](value, values);
+
+        if (validatorOutput) {
+            return validatorOutput;
+        } else if (i + 1 < validators.length) {
+            // If current iteration is on the last validator
+            // and this validator hasn't returned error
+            // we need to return null as result of whole complex valdiator
+            continue;
```
Starotitorov: You do not need to do this. It is better just to return null after this for cycle.

packages/venia-concept/src/components/Input/helpers.js
```diff
@@ -0,0 +1,22 @@
+import { isString } from 'util';
+
+const emptyFieldMessage = 'This field cannot be empty';
+export const defaultIsRequired = value =>
```
Starotitorov: Could you just change the name of the function to isRequired?

packages/venia-concept/src/components/Input/helpers.js
```diff
@@ -0,0 +1,22 @@
...
+    isString(value) && value.length > 0 ? null : emptyFieldMessage;
+
+export const createComplexValidator = validators => (value, values = null) => {
```
Starotitorov: I believe in general it is better to at least move this to util/ folder. Or create services/ folder and create validator.js file there.

packages/venia-concept/src/components/Input/input.js
```diff
@@ -32,48 +33,76 @@ class Input extends Component {
         required: PropTypes.bool,
         title: PropTypes.string,
         autoComplete: PropTypes.string,
-        helpText: PropTypes.string,
         helpType: PropTypes.string,
+        helpText: PropTypes.string,
         field: PropTypes.string.isRequired,
-        onChange: PropTypes.func
+        onChange: PropTypes.func,
+        validate: PropTypes.func,
```
Starotitorov: You do not need this prop types definition, asField hoc manages this

packages/venia-concept/src/components/SignIn/signIn.js
```diff
@@ -1,12 +1,18 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import get from 'lodash/get';
```
Starotitorov: Right now we did not define lodash as a dependency in our package.json, but I think that `get` function from lodash is extremely useful. @zetlen, @jimbo Could we discuss adding lodash.get https://github.com/lodash/lodash/tree/4.4.2-npm-packages/lodash.get to the project?

packages/venia-concept/src/components/SignIn/signIn.js
```diff
@@ -17,44 +23,48 @@ class SignIn extends Component {
...
         signIn: PropTypes.func
     };
 
-    state = {
-        password: '',
-        username: ''
-    };
+    constructor(props) {
+        super(props);
+        this.formState = {};
```
Starotitorov: In general, it is a bad idea to put the data to this object directly. In this case component will not rerender when something changes

Starotitorov: In general, you do need to store something in local state. Form component from informed can retrieve function as a children. So, formState will be passed to this function as a property.

packages/venia-concept/src/components/SignIn/signIn.js
```diff
@@ -76,22 +86,15 @@ class SignIn extends Component {
     }
 
     onSignIn = () => {
-        const { username, password } = this.state;
-        this.props.signIn({ username: username, password: password });
+        const username = this.getFieldValue(fields.username);
```
Starotitorov: If you pass onSignIn function to onSubmit will receive `values` object as an argument.

packages/venia-concept/src/components/Checkout/formContainer.js
```diff
@@ -0,0 +1,14 @@
...
+import classify from 'src/classify';
+import Form from './form';
+import defaultClasses from './form.css';
+
+const mapStateToProps = ({
+    checkout: { isAddressIncorrect, incorrectAddressMessage }
+}) => ({ isAddressIncorrect, incorrectAddressMessage });
+
+export default compose(
+    classify(defaultClasses),
```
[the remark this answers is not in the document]
AlexeyKaryka: @jimbo I've seen that in all components with containers classify wraps common component in container's export, not in common component's export, could you please describe why this approach has been chosen?

packages/venia-concept/src/components/Input/helpers.js
AlexeyKaryka: https://github.com/magento-research/pwa-studio/pull/593#discussion_r238682240

PR #600, opened 2018-12-05
Fixed minicart close button being shifted outside window

codeAdrian:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [x] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will fix an issue described here:
https://github.com/magento-research/pwa-studio/issues/565

This issue was caused by applying horizontal margin to a minicart footer which already has horizontal margin applied by elements inside it, causing the content to shift outside the viewport and affecting element width. This issue was fixed by removing parent margin.

## Additional information

Issue reported here: https://github.com/magento-research/pwa-studio/issues/565


tjwiebell:
@codeAdrian - Great fix, thank you for this PR. Not a blocker to merging, but did you find any reason why this margin rule existed? Checkout flow looks perfect, so I'm suspecting just a remnant from a re-factor.

codeAdrian:
> @codeAdrian - Great fix, thank you for this PR. Not a blocker to merging, but did you find any reason why this margin rule existed? Checkout flow looks perfect, so I'm suspecting just a remnant from a re-factor.

Thank you very much. I also assumed that this line remained from the refactoring, since the button set below also has a margin set, so the whole layout ended up with double margin which didn't make any sense to me. Personally, I tend to use padding since they can handle these edge-case situations. Let me know if you have any more questions.

PR #602, opened 2018-12-05
Introduces Jest setup file that configures the React 16 Enzyme adapter

supernova-at:
## This PR is a:

- [ ] New feature
- [x] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will alleviate developers from having to configure an Enzyme adapter at the top of every single test file.

With this change, Jest will run the the given setup file (in the `peregrine` and `venia-concept` packages) once at the start of each test file. We set up the Enzyme adapter there instead.

## Additional information

I'm definitely open to feedback on the name / location of these files.


packages/peregrine/jest.config.js
```diff
@@ -1,11 +1,12 @@
...
     displayName: 'Peregrine',
     setupFiles: [
         '<rootDir>/scripts/shim.js',
         '<rootDir>/scripts/fetch-mock.js'
     ],
-    browser: true,
-    testURL: 'https://localhost/',
-    coveragePathIgnorePatterns: ['scripts/*', 'node_modules', 'src/index.js'],
-    testPathIgnorePatterns: ['dist', 'node_modules']
+    setupTestFrameworkScriptFile: '<rootDir>/scripts/jest-setup.js',
```
supernova-at: The addition of this property is really the only change in this file. The rest are just alphabetical order changes.

supernova-at:
> Actually, I would consider only having one of these files. Call it jest.setup.js and put it in the root of the monorepo. See if that works.

Good thought - unfortunately, that did not work 😢 . I ran into some Babel issues and then a weird "Configured enzyme adapter did not inherit from the EnzymeAdapter base class" that the internet didn't help me solve (`venia-concept` and root `package.json` both declare `^3.7.0`).

In any event, while the duplication is not ideal, `peregrine` and `venia-concept` do both have their own `jest.config.js` which overrides the root `jest.config.js`. Perhaps in the future these newe `jest.setup.js` files introduced might diverge too.

packages/venia-concept/jest.config.js
```diff
@@ -21,8 +20,10 @@ module.exports = {
         // improves
         '^@magento/peregrine(/*(?:.+)*)': join(__dirname, '../peregrine/src/$1')
     },
+    testPathIgnorePatterns: ['dist', 'node_modules'],
+    testURL: 'https://localhost/',
     // Have Jest use Babel to transpile Peregrine imports in tests, since
     // our cross-package tests in the monorepo should all operate on `src`
     transformIgnorePatterns: ['node_modules/(?!@magento/peregrine)'],
-    testPathIgnorePatterns: ['dist', 'node_modules']
+    setupTestFrameworkScriptFile: '<rootDir>/jest.setup.js'
```
supernova-at: More alphabetical changes here. The only real addition is the `setupTstFrameworkScriptFile` property.

supernova-at:
> The following file(s) were not formatted with prettier. Make sure to execute npm run prettier locally prior to committing.

Is there a way to get this check to re-run? I amended a commit and had to force push, and I think that's why it didn't run again.

zetlen:
Figured out what was unstable about the `configureHost` test (it was making hashes from filesystem paths and then testing them with a bad regex) and fixed it in this branch. Good to merge!

PR #624, opened 2018-12-07
update node engine version range

sharkySharks:
Need to update the node version to reflect current support.

## This PR is a:

- [ ] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

package.json only

## Summary

When this pull request is merged, it will update the node version in package.json to better reflect our node version engine range that the app supports. I only tested this on mac os mojave. I will be spinning up a vm with windows 10 to see what bugs may arise in that environment.


jcalcaben:
can you also update the following doc page with the correct Node version:
`/pwa-devdocs/src/venia-pwa-concept/setup/index.md`

package.json
```diff
@@ -9,7 +9,7 @@
     "url": "https://github.com/magento-research/pwa-studio/issues"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=10.3.0"
```
sharkySharks: So we may want to actually bump this to `10.14.1` since that is the latest LTS version of node
https://nodejs.org/en/download/releases/

supernova-at: This is saying we "work on `10.3.0` and above" so I think this is good 👍

supernova-at: Outdated already! 😃

supernova-at:
The status of this guy is that I am unsure if we should pin to `>=10.3.0` or jump all the way to `>=10.14.1`. In my opinion if it works for all things greater than `10.3.0` then `>=`10.3.0` makes sense. 

We should only have to update this number to latest when we make a breaking change.

I think @jimbo wanted to chime in on this PR as well.

PR #628, opened 2018-12-10
Fixed an issue where pagination styles would break checkout layout

codeAdrian:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [x] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will fix an issue where category pagination style would cause the checkout section (in cart drawer) layout to break as reported here:

https://github.com/magento-research/pwa-studio/issues/627

Screenshot of the issue that is fixed with this PR:

screen shot 2018-12-10 at 8 15 56 am

## Additional information

Issue was caused by adding a global button style for pagination which caused the layout to break: https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/Pagination/pagination.css#L14

I have cleaned up the markup a bit (used CSS-modules "composes" instead of class chaining) and fixed the issue by replacing the button selector with a class selector.


jimbo:
> I have cleaned up the markup a bit (used CSS-modules "composes" instead of class chaining)

Nice. We definitely prefer `composes` over class concatenation. 👍 

> and fixed the issue by replacing the button selector with a class selector.

Yeah, we shouldn't have let this through code review. Thanks for catching this one.

packages/venia-concept/src/components/Pagination/navButton.css
```diff
@@ -1,8 +1,10 @@
-.arrow {
+.buttonArrow {
```
jimbo: It would probably have been fine to leave these names as `.arrow` and `.inactive`, but this also works.

PR #599, opened 2018-12-05
Fixed an issue where initial user details fetch would return underfined

codeAdrian:
## This PR is a:

- [ ] New feature
- [ ] Enhancement/Optimization
- [ ] Refactor
- [x] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will fix an issue which is reported here:

https://github.com/magento-research/pwa-studio/issues/549

In summary: When user logs in and refreshes the page, the initial load returns "undefined undefined" as a first name and last name of the user. This PR fixes it by calling the function that returns user's details if they are signed in.

## Additional information

Reported issue: 
https://github.com/magento-research/pwa-studio/issues/549

PR #638, opened 2018-12-10
Forgot Password Redux Refactor

supernova-at:
## This PR is a:

- [ ] New feature
- [x] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will move the `resetPassword` API call out of the `ForgotPassword` component and into Redux, for consistency with the rest of the application.

Actually making the API call is out of scope at the moment, but when it does come in scope it should be an easy change, as this PR puts all of the plumbing in place.

## Additional information

This PR is based on #513 .

#### `currentUser` in `reducers/user.js`

I fixed an inconsistency here, as `currentUser` was not being used outside of this file. [This return](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/reducers/user.js#L28-L33) was dumping the current user's details directly into the user state object, but that didn't appear to be the intended purpose.

I fixed this to update `currentUser` when we receive sign in details, and updated [the rest of the application](https://github.com/magento-research/pwa-studio/blob/supernova/forgot_password_redux_refactor/packages/venia-concept/src/components/Navigation/container.js#L16) appropriately.


supernova-at:
This PR is on `hold` until #513 lands.

supernova-at:
Removing the `hold` label on this since #513 has landed.

supernova-at:
PR Updated:
  * I accidentally incorporated changes from #599. Since they are unrelated, I backed them out

packages/venia-concept/src/components/Navigation/navigation.js
```diff
@@ -33,13 +33,21 @@ class Navigation extends PureComponent {
             userMore: string,
             userName: string
         }),
+        closeDrawer: func.isRequired,
+        completePasswordReset: func.isRequired,
+        createAccount: func.isRequired,
```
supernova-at: I added `closeDrawer` and `createAccount` as required functions in `propTypes` because they are passed in and called directly in this file.

PR #639, opened 2018-12-10
Search Page

supernova-at:
## This PR is a:

- [x] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will add a Search Page and functionality.

## Additional information

This PR brings #464 up to date and addresses some outstanding review comments.


supernova-at:
There is [one outstanding comment](https://github.com/magento-research/pwa-studio/pull/464/files/8c8c5a41e35fbf4f27cc416c7a40aed2e29be97e#r233245629) on #464 about using a `<Route>` component in `<SearchBar>`.

supernova-at:
Pending checks, this PR is good to go.

I did notice a bug while performing repeated searches however - watch the images disappear:

shared images

I did some investigation and it turns out that this only happens when an item is returned by both searches. Notice it working as intended when no items are shared:

no shared images

The culprit is the "all images are done loading" logic in `Gallery/items.js`, which doesn't take into account images that have **already** loaded from a previous search.

Working on a fix now.

codeAdrian:
Great work @supernova-at , that also puzzled me (I am working on an autocomplete feature based on the search branch). Nice to hear that you have managed to determine the cause.

jimbo:
> The culprit is the "all images are done loading" logic in Gallery/items.js, which doesn't take into account images that have already loaded from a previous search.

Yeah, that bit of code was a bit experimental. We could take it out.

supernova-at:
> I put in a quick fix to Gallery/item.js but I honestly think Gallery/items.js could be greatly simplified.
> 
> I was trying to be conscious of the fact that you had work on top of this one so I tried not to change too much out from underneath you :)

After talking it over with @jimbo , I went ahead and simplified `Gallery/items.js` after all.  `<Gallery>` no longer waits for all of its images to load before showing any of them. The result is that the images load as they become available:

gallery images

jimbo:
Obviously we lose a little visual choreography from this change, but I think it's something we can revisit later.

packages/venia-concept/src/RootComponents/Search/search.js
```diff
@@ -0,0 +1,69 @@
...
+                        return (
+                            <div className={classes.noResult}>
+                                No results found!
+                            </div>
+                        );
+
+                    return (
+                        <article className={classes.root}>
+                            <div className={classes.totalPages}>
+                                <span>{data.products.total_count} ITEMS</span>
```
jimbo: Consider making this one string rather than two nodes. Also, text generally shouldn't be all-caps in the source; that's a presentation detail, so we should use CSS to transform it.

```suggestion
                                <span>{`${data.products.total_count} items`}</span>
```

packages/venia-concept/src/actions/app/actions.js
```diff
@@ -1,6 +1,12 @@
 import { createActions } from 'redux-actions';
 
 const prefix = 'APP';
-const actionTypes = ['TOGGLE_DRAWER', 'SET_ONLINE', 'SET_OFFLINE'];
+const actionTypes = [
+    'TOGGLE_DRAWER',
+    'SET_ONLINE',
+    'SET_OFFLINE',
+    'TOGGLE_SEARCH',
+    'EXECUTE_SEARCH'
```
jimbo: We should probably have a new domain for search, one way or another.

packages/venia-concept/src/actions/app/asyncActions.js
```diff
@@ -5,3 +5,12 @@ export const toggleDrawer = name => async dispatch =>
 
 export const closeDrawer = () => async dispatch =>
     dispatch(actions.toggleDrawer(null));
+
+export const toggleSearch = () => async dispatch =>
+    dispatch(actions.toggleSearch(null));
```
jimbo: If `toggleSearch` doesn't use a payload, we shouldn't dispatch `null` here.

packages/venia-concept/src/components/App/renderRoutes.js
```diff
@@ -3,11 +3,16 @@ import { Switch, Route } from 'react-router-dom';
...
+import Search from 'src/RootComponents/Search';
 
 const renderRoutingError = props => <ErrorView {...props} />;
+const search = () => <Search />;
 
 const renderRoutes = () => (
     <Switch>
+        <Route exact path="/search.html">
+            <Page>{search}</Page>
+        </Route>
```
jimbo: If we're handling the route `/search.html` here, why would we render a `Page`? I don't think `Search` should be a RootComponent if it's a client route rather than a server one.

Can we just render a `<SearchPage>` component instead of `<Page>` here?

packages/venia-concept/src/components/CategoryList/__tests__/categoryList.spec.js
```diff
@@ -85,7 +85,7 @@ test('displays category tiles', async () => {
                                 id: 17,
                                 name: 'baz',
                                 url_key: 'baz-url.html',
-                                url_path: '/bar-url.html',
+                                url_path: '/baz-url.html',
```
jimbo: I'll never understand why the first three default "random" words include two that look so alike.

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -31,7 +31,6 @@ class CategoryList extends Component {
 
     render() {
         const { id, classes } = this.props;
-
```
jimbo: Why?

packages/venia-concept/src/components/Header/header.js
```diff
@@ -6,6 +6,9 @@ import classify from 'src/classify';
 import Icon from 'src/components/Icon';
 import CartTrigger from './cartTrigger';
 import NavTrigger from './navTrigger';
+import SearchTrigger from './searchTrigger';
+
+import SearchBar from 'src/components/SearchBar';
```
jimbo: This seems like a good candidate for lazy loading—with [`React.lazy()`](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/MiniCart/miniCart.js#L18) and [`Suspense`](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/MiniCart/miniCart.js#L128-L130).

packages/venia-concept/src/components/Header/header.js
```diff
@@ -44,21 +49,18 @@ class Header extends Component {
...
                     </div>
                 </div>
-                <div className={classes.searchBlock}>
-                    <input
-                        className={classes.searchInput}
-                        type="text"
-                        placeholder="I'm looking for..."
-                    />
-                </div>
+                <SearchBar isOpen={searchOpen} classes={classes} />
```
jimbo: This doesn't look right. Are you passing the `classes` object from `header.css` straight through to `SearchBar`?

packages/venia-concept/src/components/Header/searchTrigger.js
```diff
@@ -0,0 +1,62 @@
...
+
+                        return <EnsureOpenSearch {...props} />;
+                    }}
+                />
+            </Fragment>
+        );
+    }
+}
+
+class EnsureOpenSearch extends Component {
```
jimbo: Multiple component definitions per file is something we try to avoid, especially when they're classes. Could we move this to a new file (in the same directory)?

packages/venia-concept/src/components/SearchBar/searchBar.css
```diff
@@ -0,0 +1,72 @@
+.root {
+    background-color: rgb(var(--venia-grey));
+    align-items: center;
+    display: flex;
+    justify-content: center;
+    width: 100%;
+    transition-property: opacity, transform, height, padding;
```
jimbo: 🙅 

The only properties we should be animating are the GPU-accelerated ones: `opacity` and `transform`. Other properties cause reflows and repaints when they change, which interfere with smooth animation. [Read more.](https://www.smashingmagazine.com/2016/12/gpu-animation-doing-it-right/)

If there's not an immediately clear solution, let's remove the transition for now and add it later. I don't want to hold this PR up because of an animation.

packages/venia-concept/src/components/SearchBar/searchBar.js
```diff
@@ -0,0 +1,153 @@
+import querystring from 'querystring';
```
jimbo: Is this a new dependency? I don't see it in our `package.json`.

packages/venia-concept/src/components/SearchBar/searchBar.js
```diff
@@ -0,0 +1,153 @@
...
+                />
+            </div>
+        );
+    }
+}
+
+/**
+ * This class seeds the value of the given input ref with the search query from the URL.
+ */
+export class SeedSearchInput extends Component {
```
jimbo: Likewise, please move this component definition to a new file in the same directory.

packages/venia-concept/src/components/App/renderRoutes.js
supernova-at: This is still outstanding.

packages/venia-concept/src/components/Header/header.js
supernova-at: Good catch, totally not needed.

packages/venia-concept/src/components/SearchBar/searchBar.css
supernova-at: Awesome, thanks for the info!

packages/venia-concept/src/components/SearchBar/searchBar.js
supernova-at: Thanks, I meant to flag this - it's actually [built in to Node](https://nodejs.org/api/querystring.html), but [modern browsers](https://caniuse.com/#search=URLSearchParams) include a much easier way to do it in the browser using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get).

supernova-at:
There's a bug I have to squash in `EnsureOpenSearch` component that is causing search to stay open. I put the `hold` label on this until I fix it.

supernova-at:
> There's a bug I have to squash in EnsureOpenSearch component that is causing search to stay open. I put the hold label on this until I fix it.

So this was a fun one to track down, the problem was that the `EnsureOpenSearch` component that is supposed to get mounted when you first hit the search page was being delayed.

We ended up with a flow like this:

| User Action | Search is Open? | Notes |
| --- | ---- | --- |
| None / Initial State | False | - | 
| Toggle Search (Open Search Bar) | True | - |
| Execute Search | True | App navigates to `search.html` |
| Toggle Search (Close Search Bar) | False | The `EnsureOpenSearch` component wouldn't mount until right now. It sees that the search is not open, and would fire another `toggleSearch` to open it, resulting in the user attempting to close the search and the app closing it but then immediately opening it again |

What we really want is for `EnsureOpenSearch` to fire when the user first hits the search page (`/search.html`), like so:

| User Action | Search is Open? | Notes |
| --- | ---- | --- |
| None / Initial State | False | - | 
| Toggle Search (Open Search Bar) | True | - |
| Execute Search | True | App navigates to `search.html`. `EnsureOpenSearch` mounts now (and has no effect) |
| Toggle Search (Close Search Bar) | False | Search bar closes |

It turns out that React Router's `withRouter` higher order component [includes an important note](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md#important-note): 

> `withRouter` does not subscribe to location changes like React Redux's `connect` does for state changes ... This means that `withRouter` does not re-render on route transitions unless its parent component re-renders.

Here is the relevant component tree:

`App Container` > `App` > `Main` > `Header` > `EnsureOpenSearch`

To fix it we had to update a `prop` or `state` to tell `Header` to re-render when the `location` changed.

Since `App Container` is already "connected" to the router via `withRouter`:

```js
export default compose(
    withRouter,
    connect(
        mapStateToProps,
        mapDispatchToProps
    )
)(App);
```

We just pass `location` down from `App` to `Main` and `Header`. 

Here is some [good further reading](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md).

packages/venia-concept/src/components/App/renderRoutes.js
supernova-at: Done

packages/venia-concept/src/components/SearchBar/searchBar.css
supernova-at: I removed them. It does look a bit janky now but we can definitely revisit.

packages/venia-concept/src/actions/app/actions.js
jimbo: No, we can adjust later.

supernova-at:
PR Updated:
  * Reverts passing down `location` from `App` and instead exports `RouteConsumer` from `Peregrine` and uses that. This allows us to pick up changes in context, specifically the `location`.

supernova-at:
We ran into a little architectural discussion on this one, documenting for posterity:

# Why does `App` need `withRouter` ?

It doesn't, but removing it causes a bug when trying to `Link` to other pages (the homepage, product pages) from the Search page:

transition to homepage bug

An explanation and recommended solution follows.

## Current State

In `App/container.js`, we export the following:

```js
export default compose(
    withRouter,
    connect(
        mapStateToProps,
        mapDispatchToProps
    )
)(App);
```

`withRouter` is a [react-router HOC](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/withRouter.js) that passes [route props](https://reacttraining.com/react-router/web/api/Route/route-props) to a wrapped component.

## But doesn't it already get those in `index.js`?

Here is the relevant snippet from [venia-concept/index.js](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/index.js#L55-L57): 

```jsx
<Router apiBase={apiBase}>
    <App />
</Router>
```

If we dig into [Router](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/peregrine/src/Router/Router.js#L23-L31) we find:

```jsx
<Router {...routerProps}>
    <Route>
        {routeProps => (
            <Provider value={{ apiBase, ...routeProps }}>
                {children}
            </Provider>
        )}
    </Route>
</Router>
```

If we manually merge those together, we get:

```jsx
<Router {...routerProps}>
    <Route>
        {routeProps => (
            <Provider value={{ apiBase, ...routeProps }}>
                <App />
            </Provider>
        )}
    </Route>
</Router>
```

And it becomes clear that though `routeProps` are available, they are never passed to `App`!

## Recommended Solution

### Add a `Consumer` to `index.js`

React's [Context API](https://reactjs.org/docs/context.html#contextconsumer) also gives access to a `Consumer` that we could pair with the `Provider` we already have. 

In `index.js` instead of:

```jsx
<Router apiBase={apiBase}>
    <App />
</Router>
```

We could use the `Consumer`:

```jsx
<Router apiBase={apiBase}>
    <RouteConsumer>
        {routeProps => <App {...routeProps} />}
    </RouteConsumer>
</Router>
```

and `App` would always get updated `routeProps` whenever they change.

## Possible Solutions

### Keep `withRouter`

This is redundant since we have the `routeProps` at the parent (`Provider`) level already.

### Have Components manually subscribe to `RouteConsumer` via `static contextType`

I won't go in detail on this one since it doesn't scale well and we shouldn't consider it a viable solution.

supernova-at:
^ It should be noted that other `Link`s in the application work even after removing `withRouter` and not adding `RouteConsumer` to `index.js`.

Since it appears to be a Search page-specific bug, we may not even need to add `RouteConsumer` to `index.js`, although it does seem like that was the original intent and so we probably want to keep it anyway.

supernova-at:
Okay, so an alternate way to do this would be to leave `index.js` as is:

```jsx
ReactDOM.render(
    <ApolloProvider client={apolloClient}>
        <ReduxProvider store={store}>
            <Router apiBase={apiBase}>
                <App />
            </Router>
        </ReduxProvider>
    </ApolloProvider>,
);
```

and change `renderRoutes` from:

<Route exact path="/search.html" component={Search} />

to:

```jsx
const renderSearch = props => <Search {...props} />;

<Route exact path="/search.html" render={() => <Page>{renderSearch}</Page>} />
```

There was a bit of feedback that we didn't want to wrap `<Search>` in a `<Page>` here, but I don't fully understand the why yet.

This way may be preferable if we don't want to wrap the entire `<App>` in a `<RouteConsumer>` in `index.js`. As it stands now, every time the route changes `<App>` will be passed new `routeProps` and will re-render the entire app. There are performance concerns there of course, but I'm unsure if that's actually the desired behavior.

jimbo:
@supernova-at Thanks, Andy. Well summarized.

## Route props as context

React Router now uses the new context API to make `history`, `location`, and `match` available to any component that wants them. In this system, `Router` is a `Provider` and `Route` is a `Consumer`.

### Why do we have our own context then?

We need `MagentoRouteHandler` to be renderable at any depth in the tree without updates being blocked by any ancestor, so we need to make `routeProps` available via context. To accomplish this, we define and provide our own context in `Router` and consume it in `Page`, which in turn renders `MagentoRouteHandler`. Note that this work predates the React Router change.

Now that React Router uses the new context API, our context work in `Router` is probably redundant. The only additional property being passed by our context is `apiBase`, which could probably be an environment variable or something.

### How do we get route props then?

You should be able to use `Route` from React Router as normal. If a component needs route props, it should wrap itself with `<Route>` as recommended.


```jsx
// before
<Router>
  <App />
</Router>

// after
<Router>
  <Route component={App} />
</Router>
```

### Should we provide route props to `App`?

Not unless it needs them. The point of React Router's context-based architecture is that you don't need to inject route props at the top of the tree; `Router` *is* the top of the tree, and any child at any level underneath it can wrap itself in `Route` to consume route props. In this case, `Search` should probably wrap itself, rather than `App` as shown above.

packages/venia-concept/src/components/SearchBar/searchBar.js
```diff
@@ -0,0 +1,119 @@
...
+        this.setClearIcon(searchQuery);
+
+        if (
+            this.props.isOpen &&
+            (event.type === 'click' || event.key === 'Enter') &&
+            searchQuery !== ''
+        ) {
+            this.props.executeSearch(searchQuery, this.props.history);
+        }
+    };
```
jimbo: I think we should use `Form` and `onSubmit` for this. Listening for specific keyboard events to emulate native behavior is a sign that we're overcomplicating things.

packages/venia-concept/src/components/SearchBar/searchBar.js
```diff
@@ -0,0 +1,119 @@
...
+
+    clearSearch = () => {
+        this.searchRef.current.value = '';
```
jimbo: Likewise, this is not how we control inputs in React. We don't write to the DOM—React does.

Here, `value` is a piece of state; whatever that state is should be applied to the input  during `render()`. The role of `clearSearch()` should only be to set the `value` state to `""`.

packages/venia-concept/src/components/Header/header.js
```diff
@@ -44,21 +55,32 @@ class Header extends Component {
...
+                            {this.searchIcon}
+                        </SearchTrigger>
+                        <Route
+                            exact
+                            path="/search.html"
+                            render={() => {
+                                const { searchOpen, toggleSearch } = this.props;
+                                if (!searchOpen) {
+                                    toggleSearch();
+                                }
```
jimbo: > Mutations, subscriptions, timers, logging, and other side effects are not allowed inside the main body of a function component (referred to as React’s render phase). Doing so will lead to confusing bugs and inconsistencies in the UI.

We don't dispatch actions from a `render` function. Here we're actually in *two* render functions. 😅 

This needs to be moved to a lifecycle method.

supernova-at:
PR Updated:
  * Moved the logic for getting the value of a query string parameter into a utility method, since we'll likely re-use that
  * Moved the logic that force opened the search to the `<Search>` component's `componentDidMount`, which makes logical sense
  * Removed exporting `RouteConsumer` from Peregrine, since we're no longer using it.

Note that `<Link>`s on the Search page are not working. The easy solution is to wrap `<App>` in `withRouter` or similar, but that doesn't explain why they work everywhere else in Venia.

supernova-at:
There's a big change to `package-lock.json` now but I don't know how to get around that. The Storybook PR and this PR added new dependencies.

PR #654, opened 2018-12-12
[TASK] Pull Copyright value from Magento and display in Footer

lewisvoncken:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will dynamically show the Year and Store Name in the footer:
image

## Additional information
Related to PR https://github.com/magento/graphql-ce/pull/297


gil--:
It looks like this will depend on a new PR to the graphql codebase. How are we currently approaching developing off unreleased core-code. New code in Magento 2.3-develop is not in a official release (Maybe 2.3.1 someday) and since PWA runs on a separate release cycle, how do we insure features present in the PWA (especially related to graphql API) work with the latest Magento release (2.3.0 at the time of posting) and don't require the use of running off the develop branch.

packages/venia-concept/src/components/Footer/footer.js
```diff
@@ -61,7 +64,34 @@ class Footer extends Component {
...
+                            }
+                            if (data.storeConfig.name == null || data.storeConfig.name.length === 0) {
+                                return (
+                                    <span>© Magento {year}. All rights reserved.</span>
+                                );
+                            }
+
+                            return (
+                                <span>© {data.storeConfig.name} {year}. All rights reserved.</span>
+                            );
```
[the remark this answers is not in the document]
lewisvoncken: done

packages/venia-concept/src/queries/getStoreConfigData.graphql
```diff
@@ -0,0 +1,7 @@
+query storeConfigData {
+    storeConfig {
+        id
+        code,
+        name
+    }
```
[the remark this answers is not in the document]
lewisvoncken: only code can be removed because id is required for the query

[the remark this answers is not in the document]
lewisvoncken: done, removed code

packages/venia-concept/src/components/Footer/footer.js
```diff
@@ -17,6 +19,7 @@ class Footer extends Component {
 
     render() {
         const { classes } = this.props;
+        const year = new Date().getFullYear();
```
supernova-at: Looking back over this, I just wanted to point out that a Component's `render` function gets called a lot. Since the year is unlikely to change, we could pull the `year` out into a class- / module-level member variable and re-use it.

The call to `new Date()` isn't expensive enough to make a significant impact so this won't block a 👍 .

zetlen: @supernova-at Ah, but consider the case where the user is navigating the store _at the strike of midnight on January 1_. The cached year would not update correctly until a hard refresh re-instantiated the component. Unacceptable!!!

jimbo:
Resolved conflicts in this one. Marking as `hold` until we're on a version of GraphQL that has `StoreConfig.name`.

packages/venia-concept/src/components/Footer/footer.js
```diff
@@ -81,14 +80,8 @@ class Footer extends Component {
...
                             }
-                            const storeName =
-                                data.storeConfig.name || 'Magento';
 
-                            return (
-                                <span>
-                                    © {storeName} {year}. All rights reserved.
-                                </span>
-                            );
+                            return <span>{data.storeConfig.copyright}</span>;
```
sirugh: If this is undefined that's ok?

tjwiebell: GraphQL will always return a string or null, undefined should not be possible.

sirugh: Yea, I think I meant that maybe we should not even render a span.

tjwiebell:
Waiting to merge until @zetlen has had a chance to review.

zetlen:
Unfortunately, since we still don't have the store name according to https://github.com/magento/graphql-ce/pull/297#issuecomment-468400193, we can't merge this yet. However, the Copyright is available in the storeInfo query. Is that sufficient for this, do you think?

For what it's worth, I am advocating internally for us to change our practices and permit additive changes to GraphQL in patch releases of Magento.

sirugh:
> However, the Copyright is available in the storeInfo query. Is that sufficient for this, do you think?

What does the `Copyright` field normally contain? If people sometimes include the date & store name then I think this is fine, although maybe we should merge this and keep the issue open with a blocker linked to the GQL issue.

tjwiebell:
@zetlen - I should have updated the name and description of the PR, but I actually took a different direction that doesn't rely on this GraphQL PR. I'm not sure if it was added in 2.3.1, but there is a `copyright` field in the `storeConfig` schema that pulls this entire value from Magento (from `Content > Design > Configuration > Footer > Copyright`). Please review my commit, and I'll think you'll find this should be ready to merge.

https://github.com/magento-research/pwa-studio/pull/654/commits/3d37006aa17f3a941ec34e207b872658d9bd92d2

Issue #617, opened 2018-12-07
Add indicator to shopping cart in Venia header


ericerway:
## This issue is for the following packages:

[X] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-module`
[ ] `pwa-devdocs`

## Description

As a shopper, I want to know if I have items in my Venia shopping cart for purchase so that I know what to do next as a shopper online.

(insert updated design here)

#### Assumptions
- Working shopping cart; ability to capture # of items (backend) for future consideration

## Tasks
- [ ] Update Venia header with stateless shopping cart (empty) 
- [ ] Update Venia header with shopping cart filled with >1 item (indicator) 
- [ ] Review, test and UAT updated shopping cart across states
- [ ] Update documentation (as appropriate)

ericerway:
Updated post-grooming and assigned to @ennostuurman -- @soumya-ashok can fast follow with any updates when she has time to reflect an indicator vs. counter approach. Thanks!

ennostuurman:
hi @soumya-ashok 

I did development on this one today and it's technically almost ready , only need to write some additional unit tests and do the layout.

**Functionality:**
In this  solution the cart indicator will show number of items in the cart, i.e. 1x product A + 2 x product B = 3, in case of an empty cart the counter will not be visible.

Can you confirm the above is correct functionality?

**Layout**
There's a discrepancy right now between the used header cart icon , see screenprint, and the designs as per the Venia documentation, see https://magento-research.github.io/pwa-studio/venia-pwa-concept/component/header/ . 

Can you confirm what is the preferred layout for this component / functionality?

Thank you.

-----

Current cart icon Venia theme:
![magento-pwa-studio-issue-617-cart-icon](https://user-images.githubusercontent.com/1906257/49814419-dfa00600-fd69-11e8-9f7b-73b3886cb8ae.png)

Layout as per documentation:
![magento-pwa-studio-issue-617-documentation-cart-icon](https://user-images.githubusercontent.com/1906257/49814607-48877e00-fd6a-11e8-9e6e-dd15043c3386.png)

ennostuurman:
@soumya-ashok thank you I adjusted my previous  accordingly and issued PR ^^ .

Screenprint cart icon empty cart:

[...]

PR #664, opened 2018-12-13
Feature/617 cart item indicator

ennostuurman:
## This PR is a:

- [X] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

- Add visual indication to user that the cart is filled.
- Requirements see issue #617 

## Summary

- Show an outline cart icon when cart is empty
- Show an filled cart icon when items have been added to cart

## Additional information

- did not yet add unit tests because not sure whether we want to test SVG output (filled or not).
- this functionality works fine, however I noticed when debugging propTypes.isRequired (which logs a warning on the itemsCount prop) that items_count at first render returns undefined. Looks like getCartDetails() does not return value and header component renders multiple times. Did some first tests to see if I could fix this, e.g. adding ` async componentdidMount await getCartDetails() ` . But this didn't fix it (yet).


packages/venia-concept/src/components/Header/header.js
```diff
@@ -20,11 +22,34 @@ class Header extends Component {
...
+
+    get cartIcon() {
+        const itemsCount = this.props.itemsCount || false;
+        return !itemsCount || itemsCount === 0 ? (
+            <Icon name="shopping-cart" attrs={{ stroke: '#212121' }} />
+        ) : (
+            <Icon
+                name="shopping-cart"
+                attrs={{ stroke: '#212121', fill: '#212121' }}
+            />
```
supernova-at: Lastly on this function, since the only thing changing is whether or not we add a `fill` attribute, we could extract that piece of logic out and simplify things a bit:

What do you think of something like:

```jsx
get cartIcon() {
    const svgAttributes = {
        stroke: 'rgb(var(--venia-text))',
    };

    if (this.props.itemsCount > 0) {
        svgAttributes.fill = 'rgb(var(--venia-text))';
    }

    return (
        <Icon
            name="shopping-cart"
            attrs={svgAttributes}
        />
    );
}
```

?

packages/venia-concept/src/components/Header/header.js
```diff
@@ -1,4 +1,6 @@
 import React, { Component } from 'react';
+import { connect } from 'react-redux';
```
supernova-at: I'd like @jimbo to chime in here too but from what I've gathered so far we want to be very deliberate where we `connect` components.

In this case, an alternative approach to `connect`ing this component would be to add `itemsCount ` to the connected component `App/container.js`, and pass it down to this component via props (you'll have to pass it through `Main/main.js` too).

There's a balancing act here to be sure but it does simplify things quite nicely when done well.

ennostuurman: @supernova-at thanks again for all the suggestions :+

1. I get it your point, and indeed it's a balancing act. But isn't that the reason we want to use Redux, i.e. not having to pass props down to children through components that don't require the props at all?  I find passing down props often less clear than just having a connected component in which you just 'select' a state object with mapStateToProps.

2. In a first setup I had this logic in the cartTrigger.js which is already a connected component.  I however moved the cart counter to header.js because I felt that cart counter should not be a part of cartTrigger.js which is an action based component (trigger an action)  This said I even thought about a complete separate cartCounter.js  component which might solve the header component re-render issue that I faced and which is easier to isolate / fix in a single purpose component.

Please let me know what you prefer and I'll refactor. Will refactor anyhow again because I will be adding the items counter back in again, see issue #617

packages/venia-concept/src/components/Header/header.js
ennostuurman: yeah much DRY'er ;-)

packages/venia-concept/src/components/Header/header.js
```diff
@@ -20,11 +22,34 @@ class Header extends Component {
             searchTrigger: PropTypes.string,
             secondaryActions: PropTypes.string,
             toolbar: PropTypes.string
-        })
+        }),
+        // isRequired won't work because initially itemsCount is undefined
+        // seems page re-renders before cart getDetails() returns value
+        itemsCount: PropTypes.number
```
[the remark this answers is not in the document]
ennostuurman: got it, good point!

packages/venia-concept/src/components/Header/header.js
ennostuurman: @supernova-at see my previous comment ^^, what do you think, shall I add the counter to cartTrigger instead, so we don't introduce an additional `connect`ed component?

supernova-at: Sorry for the delay, @ennostuurman - vacation called 😄 

I think absolutely a `cartCounter.js` component would be awesome.

supernova-at:
Looks like there is a conflict in `Header/header.js` as well.

packages/venia-concept/src/components/Header/header.js
ennostuurman: @supernova-at I updated PR according to new specs, see https://github.com/magento-research/pwa-studio/issues/617#issuecomment-451650645 . 

Unfortunate a peregrine/package-lock.json made it into this PR, I always try to keep them out, this on e slipped through tho and I couldn't manage to remove it despite rebasing my branch w/ latest release/2.0 .

Could you please review? Thx!

ennostuurman: @supernova-at FYI => I updated PR with latest staging, fixed header.js conflict (recent react-feather changes) .

supernova-at:
This is looking great!

Some last things:

1. We will want to remove `packages/peregrine/package-lock.json`. You can just delete it from disk, `git add .` and then commit and push that change.
1. Even when there are no items in the cart, the cart appears black / filled in for me. Is this intentional?
1. I thought I saw that once the count goes over `10` the display should show `10+`. Am I mistaken? 

Thanks again!

packages/venia-concept/src/components/Header/cartTrigger.js
```diff
@@ -13,32 +17,47 @@ class Trigger extends Component {
...
     }
 }
 
+const mapStateToProps = state => {
+    const { cart } = state;
+
+    return {
+        cart
+    };
+};
```
supernova-at: If you want, you can simplify this with:

```js
const mapStateToProps = ({ cart }) => ({ cart });
```

ennostuurman:
@supernova-at thanks for the review and suggestions. 

I made following changes:

1) refactored mapStateToProps to use destructuring as you suggested
2) added outline only layout for empty cart including test for outline / filled layout. There was some confusion around the latest designs, but Soumya confirmed outline layout, see https://github.com/magento-research/pwa-studio/issues/617#issuecomment-453596488  . 

Re you question about the counter showing 99+ . Eventually we decided to always show the actual quantity, , see https://github.com/magento-research/pwa-studio/issues/617#issuecomment-448216969 . The layout is fool proof for a 9999 counter so that should be fine.

Can you please review again? Thx!

ennostuurman:
hey @supernova-at I just updated this PR with latest release/2.0 branch changes. In case you have time and you're brains haven't been totally sucked up by Braintree ;-) , could you please review (see my earlier comments ;-) .? Thks!

PR #707, opened 2018-12-21
Fix grid layout breaking in safari in checkout

AlexeyKaryka:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [ X ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will fix grid layout breaking in checkout in safari


jimbo:
In Safari, an element whose height is implicitly or explicitly set to a flexible unit will instead have its height calculated as `auto` unless its containing element has an explicit height.

Although the minicart's height is fixed here using absolute positioning, and its second grid row should have an implicit height of `1fr` as specified by `grid-template-rows`, Safari treats it as `auto` instead, causing the row to not consume any of the free space available to it. Setting an explicit height on the container doesn't change the calculated height, it just handles Safari's bug.

awilcoxa:
closing as PR is merged

Issue #695, opened 2018-12-20
UAT: Checkout Drawer Incorrect Styling


awilcoxa:
There are currently various inconsistencies within the checkout drawer, including text and styling issues. Issues should be resolved to adhere to current design. 

Issue 11: https://wiki.corp.magento.com/display/FF/Venia+UX+Issues 

Corrections Include: 
Prompt text color #C0123F
Get it by pre-populated with a default shipping method
Place order changed to "Confirm Order"
Disabled button style for "Confirm Order" updated to "medium" size, light gray, fill style
Cancel button added - this will lead the user back to the first cart state
Currently, the rows have to be filled in order from top to bottom. The original intent was for the user to be able to begin where they wanted to, so all rows should be clickable / editable
The confirm order button converts to an enabled state (dark gray fill) when all required information is filled out
Once a section is filled out, an edit icon needs to display next to it like on "get it by"

## This issue is for the following packages:

- [x ] `venia-concept`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [x ] Bug
- [ ] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

awilcoxa:
Current: 

![image](https://user-images.githubusercontent.com/44207440/50298319-7a70a280-0444-11e9-85d9-8ece7ceb7e35.png)

Correct: 

![image](https://user-images.githubusercontent.com/44207440/50298331-8197b080-0444-11e9-9206-7119c8b92f2a.png)

supernova-at:
Assigning to myself, as I am taking over this ticket.

PR #716, opened 2018-12-27
Checkout form updates

AlexeyKaryka:
## This PR is a:

- [X] New feature
- [ ] Refactor
- [X] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will update styles and behavior of checkout form:
1) Make all checkout options available for editing from the beginning
2) Add edit icon  to the option after it's filling
3) Change submit button text to "Confirm Order"
4) Add cancel button that closes checkout form and leave drawer open

WHA THIS PR DOESN'T MAKE FROM ISSUE TASKS:
1)Get it by  isn't pre-populated with a default shipping method - because I haven't found 'default shipping method' and also the only available option 'flat rate' isn't able to be chosen, after clicking I get 400 error.


packages/venia-concept/src/components/Button/button.css
```diff
@@ -1,18 +1,20 @@
...
     font-size: 0.875rem;
-    height: 3rem;
-    min-width: 8.75rem;
+    height: 2.125rem;
+    min-width: 9rem;
     padding: 0 1.5rem;
     transition: opacity 256ms linear;
 }
 
-.root:hover {
```
supernova-at: Do we no longer have a hover state on buttons?

packages/venia-concept/src/components/Button/button.css
```diff
@@ -1,18 +1,20 @@
 .root {
     composes: root from '../clickable.css';
     background-color: white;
-    border: 1px solid currentColor;
-    border-radius: 1.5rem;
-    color: rgb(var(--venia-teal));
+    border: 1px solid black;
+    border-radius: 6.25rem;
+    color: black;
```
supernova-at: Note that these are the styles for the base `Button` component, which means this change updates all the buttons in the entire storefront to this style.

Are we no longer using the `--venia-teal` anywhere?

packages/venia-concept/src/components/Checkout/form.css
```diff
@@ -10,11 +10,7 @@
 }
 
 .informationPrompt {
-    color: rgb(var(--venia-orange));
-    text-transform: capitalize;
-}
-.informationPrompt--disabled {
```
supernova-at: I think we might have to keep this.

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -182,17 +188,10 @@ class Form extends Component {
...
-            isShippingInformationReady,
-            paymentTitle
-        } = this.props;
+        const { classes, isPaymentMethodReady, paymentTitle } = this.props;
 
         if (!isPaymentMethodReady) {
-            const promptClass = isShippingInformationReady
-                ? classes.informationPrompt
-                : classes['informationPrompt--disabled'];
+            const promptClass = classes.informationPrompt;
```
supernova-at: You don't even need this variable anymore, just

```jsx
return <span className={classes.informationPrompt}>Add Billing Information</span>;
```

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -204,17 +203,10 @@ class Form extends Component {
...
-            isShippingMethodReady,
-            shippingTitle
-        } = this.props;
+        const { classes, isShippingMethodReady, shippingTitle } = this.props;
 
         if (!isShippingMethodReady) {
-            const promptClass = isPaymentMethodReady
-                ? classes.informationPrompt
-                : classes['informationPrompt--disabled'];
+            const promptClass = classes.informationPrompt;
```
supernova-at: Same here, this variable isn't needed.

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -238,6 +230,11 @@ class Form extends Component {
         return <div className={classes.root}>{children}</div>;
     }
 
+    cancelHandler = () => {
+        const { closeCheckoutForm } = this.props;
+        closeCheckoutForm();
```
supernova-at: Please add `closeCheckoutForm` to this component's `propTypes`. 

Since we're calling it directly here, please also make it `required`.

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -144,19 +145,23 @@ class Form extends Component {
         return (
             <Fragment>
                 <div className={classes.body}>
-                    <Section label="Ship To" onClick={this.editAddress}>
+                    <Section
+                        label="Ship To"
+                        filledOption={isShippingInformationReady}
```
supernova-at: On first glance, it's not clear what `filledOption` means / wants.

Can we think of a more informative name for this prop?

packages/venia-concept/src/components/Checkout/section.js
```diff
@@ -15,7 +20,12 @@ class Section extends Component {
             icon: string
         }),
         label: node,
-        selectedOption: bool
+        selectedOption: bool,
+        filledOption: bool
+    };
+
+    defaultProps = {
```
supernova-at:
```suggestion
     static defaultProps = {
```

packages/venia-concept/src/components/Checkout/submitButton.js
```diff
@@ -17,8 +17,8 @@ class SubmitButton extends Component {
         const disabled = isDisabled(submitting, valid);
 
         return (
-            <Button disabled={disabled} onClick={submitOrder}>
-                Place Order
+            <Button filled={true} disabled={disabled} onClick={submitOrder}>
```
supernova-at: Just for my own education, do we now have four possible button states?

1. Not filled, Not disabled
2. Not filled, disabled
3. Filled, Not disabled
4. Filled, disabled

packages/venia-concept/src/components/Checkout/wrapper.js
```diff
@@ -4,7 +4,7 @@ import { array, bool, func, object, oneOf, shape, string } from 'prop-types';
 
 import { Util } from '@magento/peregrine';
 import { getShippingMethods } from 'src/actions/cart';
-import {
+import actions, {
```
supernova-at: Looks like you're only using `actions.reset`, let's just add `reset` to the list of named imports.

supernova-at: Also, it doesn't look like `src/actions/checkout` exports a function named `reset` ?

supernova-at:
From #695:

**Done and Verified**

- [x] Prompt text color #C0123F
- [x] Place order changed to "Confirm Order"
- [x] Disabled button style for "Confirm Order" updated to "medium" size, light gray, fill style
- [x] Cancel button added - this will lead the user back to the first cart state
- [x] Once a section is filled out, an edit icon needs to display next to it like on "get it by"

**To Be Determined**

- [ ] Get it by pre-populated with a default shipping method

> I haven't found 'default shipping method'

I also am not familiar enough to know if we can just hardcode `flatrate` in as the default shipping method initially. If all Magento instances support `flatrate`, perhaps we can. 

But definitely once we get the results back from [getShippingMethods](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/actions/cart/asyncActions.js#L287) we should be able to default to the first option. We may have to move this call to earlier in the flow to have it available initially.

<hr />

- [ ] Currently, the rows have to be filled in order from top to bottom. The original intent was for the user to be able to begin where they wanted to, so all rows should be clickable / editable

>  the only available option 'flat rate' isn't able to be chosen, after clicking I get 400 error.

Correct, unfortunately to my knowledge the design goal is not technically possible. In order get the list of available shipping methods to choose from, you must first have provided a shipping address to the Magento backend.

cc @tjwiebell for a sanity check on this one.

<hr/>

- [ ] The confirm order button converts to an enabled state (dark gray fill) when all required information is filled out

I am seeing black instead of dark gray on this one.

supernova-at:
There are some things not hooked up correctly after the cancel button is clicked:

cancel button click

packages/venia-concept/src/components/Button/button.css
AlexeyKaryka: @supernova-at This is a good question, at one hand I support your opinion regarding that not every button will have such a style, but on other hand looks like the whole design of application changes it's major color from --venia-teal to black, so it preferable to retain such button in one place, not changing color from --venia-teal to black in each button that need it

supernova-at:
Closing this PR as it is superseded by #799.

Issue #652, opened 2018-12-12
[Bug]: Double slash in request


vitalics:
## This issue is for the following packages:

- [X] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [X] Bug
- [ ] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

### Environment

| Question                    |    Answer     |
| --------------------------- | :-----------: |
| Magento version             |  |
| Operating System + version  |  |
| node.js version (`node -v`) |  |
| npm version (`npm -v`)      |  |

## Description
When we send a request with `graphql`, we get a next image url: e.g. `/v/a/va10-ct_main_2.jpg`
but in a request we're sending request with double slashes before product
![image](https://user-images.githubusercontent.com/8816260/49874479-a6ca6480-fe2f-11e8-8710-b564729bdefb.png)

## Expected result:
Sending canonical request string. In this case: `URL/media/catalog/product/v/a/va10-ct_main_2.jpg`

## Additional Information:
It's not a critical bug

PR #724, opened 2019-01-03
Adds more robust unit tests to makeMediaPath

supernova-at:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will update the unit tests for `makeMediaPath` so that they more fully test the intended functionality.

Changes to this module should now cause its tests to fail if - for example - they introduce double slashes as reported in issue #652 (resolved by #676). This PR is essentially a followup to #676.

## Additional information

I have done a self code review and left some comments in this PR where the reviewer should focus.

### DIFF
--- a/packages/venia-concept/src/util/__tests__/makeMediaPath.spec.js
+++ b/packages/venia-concept/src/util/__tests__/makeMediaPath.spec.js
@@ -4,7 +4,7 @@ const fooTrailingSlash = '/foo/';
 const fooNoTrailingSlash = '/foo';
 
 describe('makeProductMediaPath', () => {
-    test('works as intended', () => {
+    test('normalizes the product path of resources with and without trailing slashes', () => {
         const expected = '/media/catalog/product/foo';
 
         const trailingSlashResult = makeProductMediaPath(fooTrailingSlash);
@@ -16,7 +16,7 @@ describe('makeProductMediaPath', () => {
 });
 
 describe('makeCategoryMediaPath', () => {
-    test('works as intended', () => {
+    test('normalizes the category path of resources with and without trailing slashes', () => {
         const expected = '/media/catalog/category/foo';
 
         const trailingSlashResult = makeCategoryMediaPath(fooTrailingSlash);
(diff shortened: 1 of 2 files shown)


packages/venia-concept/src/util/makeMediaPath.js
```diff
@@ -1,6 +1,7 @@
 const trailingSlashRE = /\/+$/;
 const wrappingSlashRE = /^\/*(.+?)\/*$/;
-export function makePathPrepender(prefix) {
+
+function makePathPrepender(prefix) {
```
supernova-at: Please note that we are no longer exporting this `makePathPrepender` higher order function (it is not used anywhere outside of this file).

packages/venia-concept/src/util/__tests__/makeMediaPath.spec.js
```diff
@@ -1,13 +1,28 @@
...
+import { makeCategoryMediaPath, makeProductMediaPath } from '../makeMediaPath';
+
 const fooTrailingSlash = '/foo/';
 const fooNoTrailingSlash = '/foo';
 
-test('makePathPrepender makes a function that returns a path', () => {
-    const makePath = makePathPrepender(fooNoTrailingSlash);
-    expect(makePath('nothing')).toBe('/foo/nothing');
+describe('makeProductMediaPath', () => {
+    test('works as intended', () => {
```
jimbo: This is a super lazy test name. 🤔

packages/venia-concept/src/util/makeMediaPath.js
jimbo: It not being used doesn't necessarily mean it shouldn't be exported (and tested). I'll approve this particular change, but in general we should export and test more units.

packages/venia-concept/src/util/__tests__/makeMediaPath.spec.js
supernova-at: 😄

PR #729, opened 2019-01-07
mock TTY setting for domain configuration test

sharkySharks:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [x] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will fix a bug where the test that is being changed was failing due to the state of the pre-push hook being run in a non-interactive shell. I also added `prettier` to the pre-push hook since there doesn't seem to be a reason not to.

## Additional information

`process.stdin.isTTY` is being mocked in the test previous to the one changed, so it makes sense to also mock it in the following test, since these tests will be run in both interactive and non-interactive shells.


packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
```diff
@@ -177,7 +177,7 @@ test('child node correctly sets new root and parent ids', async () => {
         </MockedProvider>
     );
 
-    await wait();
+    await wait(200);
```
sharkySharks: this change was added to fix flakiness where it wouldn't finish resolving in time for the assertion

packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
```diff
@@ -153,7 +153,7 @@ test('renders with product data', async () => {
         </MockedProvider>
     );
 
-    await wait();
+    await wait(500);
```
sharkySharks: added this to prevent flakiness, the test needed a little more time to resolve before the assertion. reran the tests in ci 3 times and did not see it come up again, but we may want to keep an eye on this one to see if it still causes issues. Saw the flakiness both locally as well as up in ci.

sharkySharks:
This fixes #731

packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
jimbo: I don't think this is the right solution. If `await wait()` isn't working, we need to find something else.

What we're trying to do here is wait for all promises to resolve, which may be several ticks. I noticed that other tests sometimes tell Jest to run timers and such, so that may be the key. We should find the minimal instruction and use it consistently.

sharkySharks:
@jimbo I removed the test change I made but the test is failing in CI, so would you rather skip the test or have the non-ideal `wait(500)` in there?

jimbo:
@sharkySharks After diving into the issue, it looks like Jest just doesn't have a good answer for ensuring promises have been resolved before executing a test. While they have some solutions in flight, there's nothing clear-cut and nothing we could use right now while we wait.

So I'm fine with `wait(500)` for now.

sirugh:
> @sharkySharks After diving into the issue, it looks like Jest just doesn't have a good answer for ensuring promises have been resolved before executing a test. While they have some solutions in flight, there's nothing clear-cut and nothing we could use right now while we wait.
> 
> So I'm fine with `wait(500)` for now.

At the least I think we should leave a brief comment on the `wait(500)` line or leave a TODO regarding the decision.

sharkySharks:
Alright y'all, @jimbo @sirugh this is ready for another review when you get a chance.

packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
sharkySharks: @zetlen here is the quick fix I put up for this, we can add to the issue more details if you have ideas for a better solution, but I wanted to unblock folks from its flakiness for now.

zetlen: There's a fix for this that seems to work well in #738. @sharkySharks if you can review that PR and merge it, this change might be unnecessary!

Issue #376, opened 2018-10-16
Review icon fonts/library for performance improvements [SPIKE]


ericerway:
## This issue is for the following packages:

[X] `venia-concept`
[ ] `pwa-buildpack`
[X] `peregrine`
[ ] `pwa-devdocs`
[ ] `upward-js`
[ ] `upward-spec`

## This issue is a:

[ ] Bug
[ ] Feature suggestion
[ ] Documentation issue
[X] Other (Please Specify)

### Environment

| Question                    |    Answer     |
| --------------------------- | :-----------: |
| Magento version             |  |
| Operating System + version  |  |
| node.js version (`node -v`) |  |
| npm version (`npm -v`)      |  |

## Description

Continued review of Javascript size should include our use of feather-icons for Venia to ensure high performance for our reference storefront in PWA.

## Expected result:

[ ] Summary of results; recommended improvements
[ ] Creation of new issues/stories (as appropriate)

## Possible solutions:

[ ] Optimization of current library to remove unused icons (preferred)
[ ] Review of alternate icon fonts/libraries for performance

sirugh:
Spitballing some notes/ideas/planning around this:

* FeatherIcons seems to give you all or nothing in regards to its exports. It doesn't look like you can import specific icons.
* We can use webpack stats to compare various icon package sizes (if the publisher doesn't already list it) and then pick whichever is "lightest".
* Because of the "dynamic" nature of `Icon` component we probably cannot utilize tree-shaking OOTB. With a different icon package we may be able to do away with the `Icon` component and import specific icons.

*Open Question*

* Why did we use feathericons in the first place? For the SVG output? Size? Just looking for some background.

ericerway:
Thanks for the readout @sirugh -- @jimbo can help with the rationale and collaboration on which option to move forward with on this one.

PR #733, opened 2019-01-08
SPIKE: Evaluating Icon library

sirugh:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

* Add a script, `stats:venia`, that will build a production bundle of Venia and display a visual representation of various file sizes within the bundle.
* Replace usage of [feather-icons](https://feathericons.com/) with a react-ified library, [react-feather](https://github.com/carmelopullara/react-feather).
* Address #376. 

## Additional information
**Before**
[Image from Gyazo](https://gyazo.com/35de69bb95f4c18e5ad3d4755fb10148)

**After**
[Image from Gyazo](https://gyazo.com/e4ccb87d7e037a57dc57f197042bd171)


packages/venia-concept/src/RootComponents/Search/search.js
```diff
@@ -63,7 +63,7 @@ export class Search extends Component {
                     </Query>
                 </small>
                 <Icon
-                    name="x"
+                    name="X"
```
sirugh: You'll see a lot of this in the PR. This name is now just a direct correlation to the name exported by the `react-feather` lib.

packages/venia-concept/src/components/Icon/icon.js
```diff
@@ -14,10 +55,16 @@ class Icon extends Component {
...
+        // Use the name provided so that in the case of an unknown icon we throw
+        // an appropriate error.
+        const IconComponent = Icons[name] || name;
+        return (
+            <span className={classes.root}>
+                <IconComponent
+                    size={attrs && attrs.width}
+                    color={attrs && attrs.color} />
+            </span>
+        );
```
sirugh: The above is really the bulk of this change (aside from the stats addition). Does it make sense to dynamically import like so or should we put the onus of icon import/utilization on the components and do away with this abstraction? Aside from having all the icon imports in one place this component does provide a default styling.

packages/venia-concept/src/components/Icon/icon.js
```diff
@@ -1,10 +1,51 @@
 import React, { Component } from 'react';
-import Feather from 'feather-icons';
 import PropTypes from 'prop-types';
 
 import classify from 'src/classify';
 import defaultClasses from './icon.css';
 
+// TODO: Tree-shaking is broken in react-feather right now, see the issue below.
+//   When this is fixed you can import without using full path.
+//   https://github.com/carmelopullara/react-feather/issues/6
```
sirugh: If we want to avoid having to do full path imports we could switch to [react-icons](https://www.npmjs.com/package/react-icons). I guess the icon font lib we use is really a UX call? @soumya-ashok any opinions?

packages/venia-concept/src/components/Icon/icon.js
sirugh: For example, per @DrewML: 
```js
// Inside MyComponent.js
import { SomeIcon } from 'react-feather';
import Icon from '../Icon';

<Icon src={SomeIcon} />
```

sharkySharks:
re prettier formatting errors above: i added `prettier` to the pre-push script so you shouldn't get those after #729  is merged

packages/venia-concept/src/components/Icon/icon.js
sirugh: I'm going to go ahead and do this ^

packages/venia-concept/src/components/Icon/icon.js
```diff
@@ -13,11 +15,16 @@ class Icon extends Component {
...
-        const { attrs, classes, name } = this.props;
-        const svg = Feather.icons[name].toSvg(attrs);
-        const fn = () => ({ __html: svg });
+        const { attrs, classes, src: IconComponent } = this.props;
 
-        return <span className={classes.root} dangerouslySetInnerHTML={fn()} />;
+        return (
+            <span className={classes.root}>
+                <IconComponent
+                    size={attrs && attrs.width}
```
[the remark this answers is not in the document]
zetlen: @sirugh That would be nice, thanks! Let's make it a union type so that either a `{ height, width }` or an int works.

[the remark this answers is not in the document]
sirugh: I don't think `react-feather` takes height and width unfortunately. For example, [this icon](https://github.com/carmelopullara/react-feather/blob/master/src/icons/align-justify.js#L5-L18) takes `color, `size`, and `otherProps` which are spread. Though now that I look at it I wonder if that would spread/overwrite width if provided.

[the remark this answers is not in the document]
zetlen: My bad, I misunderstood. I guess what I mean is something like:

```suggestion
                    size={this.props.size && Number(this.props.size) || attrs && Number(attrs.width)}
```

which would allow:
```jsx
<Icon src={Foo} size={16} />
<Icon src={Foo} size="16" />
<Icon src={Foo} attrs={{ width: 16 }} />
<Icon src={Foo} attrs={{ width: "16" }} />
```

sirugh:
>name imported icons according to their usage. 

Sure, but we'd have to re-alias the default. For example:

```js
import { default as CloseIcon } from 'react-feather/dist/icons/x';
``` 

Then again, it may make it that much easier to switch out the dependency if we go with a different icon library later. I can do this tomorrow morning.

jimbo:
> Sure, but we'd have to re-alias the default.

Not really. When you import a default export, you're already re-aliasing it. Its alias is `default`. 😅 

```es6
// the following two lines are equivalent
import Bar from "foo"
import { default as Bar } from "foo"
```

packages/venia-concept/src/components/Header/header.js
```diff
@@ -4,6 +4,9 @@ import { Link, Route } from 'react-router-dom';
 
 import classify from 'src/classify';
 import Icon from 'src/components/Icon';
+import { default as SearchIcon } from 'react-feather/dist/icons/search';
+import { default as MenuIcon } from 'react-feather/dist/icons/menu';
+import { default as ShoppingCartIcon } from 'react-feather/dist/icons/shopping-cart';
```
jimbo: Now just simplify all these and we're good. 😅 

```es6
import SearchIcon from "react-feather/dist/icons/search"
```

Issue #696, opened 2018-12-20
UAT: Mask is not applied to background view in Checkout


awilcoxa:
During the checkout flow when the right hand navigation drawer is open, a mask should be applied to the cart as indicated in design. 

Issue 14: https://wiki.corp.magento.com/display/FF/Venia+UX+Issues

## This issue is for the following packages:

- [x ] `venia-concept`
- [ ] `pwa-buildpack`
- [x ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [ x] Bug
- [ ] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

awilcoxa:
Current: 

![image](https://user-images.githubusercontent.com/44207440/50298730-75602300-0445-11e9-9a91-39a385991321.png)


Correct: 

![image](https://user-images.githubusercontent.com/44207440/50298741-7b560400-0445-11e9-80ca-aadcc002b030.png)

PR #744, opened 2019-01-10
Updates checkout flow with Braintree Web Drop In for Credit Card processing

supernova-at:
## This PR is a:

- [x] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Issue

Closes https://github.com/magento-research/pwa-studio/issues/349.

## Summary

When this pull request is merged, it will update the payment section of the checkout flow with a credit card form instead of the simple dropdown.

cc-updated

## Additional information

I integrated [Braintree's Web Drop In](https://github.com/braintree/braintree-web-drop-in), which is built on the new Web Payments / Payment Request API browser standard.

Luckily for us, the [drop in's browser support](https://braintree.github.io/braintree-web/current/#browser-support) matches or exceeds Venia's documented browser support (`>5%`).

### Out of Scope

The following features are out of scope for this ticket.

#### Address form validation errors
This will be done in #770.

#### Dynamic settings from M2 backend

Venia still requests and stores the available payment methods in Redux because we anticipate adding this support in the future, but the UI does not respect or care about them: it will always show the credit card payment form.

#### PayPal Integration

Enabling PayPal in the Braintree Web Drop In should be a straightforward follow up ticket. Here is that ticket: #747.

### Notes for the Reviewer(s)

This PR introduces a new environment variable section for braintree integration to `.env.dist`. When verifying locally, be sure to update your local `.env` file with this new section and then `npm run build`.

Additionally, in order to verify locally you'll have to make the following edit after checkout:
In `venia-concept/src/actions/checkout/asyncActions.js`, replace line 151 as follows:

```js
// from
method: paymentMethod.code

// to
method: 'checkmo'
```

This is because the current cloud instance that we're pointing to does not have Braintree enabled and can only accept the `checkmo` payment method.

### Requires Follow Up

After this PR is merged and deployed to the demo site https://magento-venia.now.sh/, we will have to update our Magento cloud instance to accept Braintree payments.


packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -0,0 +1,102 @@
...
+        ) {
+            // Our parent is telling us to request the payment nonce.
+            this.requestPaymentNonce();
+        }
+    }
+
+    render() {
+        if (!this.state.isSupported) {
+            throw new Error('Web Payments is not supported by this browser.');
+        }
```
supernova-at: This is actually still a vestige of when I thought I'd need to provide a fallback UI in case this browser did not support the Payment Request API. Since that is not a concern (see PR write up), I can rip this out.

packages/venia-concept/src/components/Checkout/flow.js
```diff
@@ -47,16 +46,15 @@ class Flow extends Component {
         isPaymentMethodReady: bool,
         isShippingInformationReady: bool,
         isShippingMethodReady: bool,
-        paymentMethod: string,
-        paymentTitle: string,
+        paymentDisplayPrimary: string,
```
supernova-at: The `paymentMethod` was used to seed the payment method dropdown with the current selection when it was re-opened.

Since that is no longer a thing, we no longer need `paymentMethod`.

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -1,59 +1,45 @@
...
             paymentMethod: string
         }),
-        paymentMethod: string,
         submit: func.isRequired,
         submitting: bool
     };
 
-    static defaultProps = {
-        availablePaymentMethods: [{}]
+    // TODO: don't actually keep the state here, move to Redux
```
supernova-at: This TODO may not be accurate.

Either I can move this state to Redux, or I can remove this TODO.

packages/venia-concept/src/components/ErrorBoundary/index.js
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    render() {
+        if (this.state.hasError) {
+            return this.props.fallbackUI;
+        }
+
+        return this.props.children;
+    }
+}
```
supernova-at: This is also leftover from when I was providing a fallback UI.

I'm fine including or excluding this file. If we keep it I'll add unit tests to document its usage. Basically:

```jsx
<ErrorBoundary fallbackUI={this.someFunctionThatReturnsJSX}>
  <SomeComponentThatMightThrow />
</ErrorBoundary>
```

packages/venia-concept/src/reducers/checkout.js
```diff
@@ -12,8 +12,11 @@ const storedShippingMethod = storage.getItem('shippingMethod');
 
 const initialState = {
     editing: null,
-    paymentMethod: storedPaymentMethod && storedPaymentMethod.code,
-    paymentTitle: storedPaymentMethod && storedPaymentMethod.title,
+    paymentCode: storedPaymentMethod && storedPaymentMethod.code,
+    paymentDisplayPrimary:
+        storedPaymentMethod && storedPaymentMethod.display_primary,
+    paymentDisplaySecondary:
+        storedPaymentMethod && storedPaymentMethod.display_secondary,
```
supernova-at: `paymentCode` is a much less ambiguous name for what this value holds.

I split `paymentTitle` into `paymentDisplayPrimary` and `paymentDisplaySecondary` to facilitate the desired UI.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -0,0 +1,102 @@
...
+    componentDidUpdate(prevProps) {
+        if (
+            this.braintree &&
+            this.props.isRequestingPaymentNonce &&
+            !prevProps.isRequestingPaymentNonce
+        ) {
+            // Our parent is telling us to request the payment nonce.
+            this.requestPaymentNonce();
+        }
+    }
```
supernova-at: I'm open for comments on this approach. The parent needed some way to trigger this call on the child and I was trying to prefer a declarative approach to an imperative one.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
supernova-at: Resolved by new commit.

packages/venia-concept/src/components/Checkout/flow.js
```diff
@@ -47,16 +46,14 @@ class Flow extends Component {
         isPaymentMethodReady: bool,
         isShippingInformationReady: bool,
         isShippingMethodReady: bool,
-        paymentMethod: string,
-        paymentTitle: string,
+        paymentData: object,
```
supernova-at: `availablePaymentMethods` and `paymentMethod` are no longer used, and the UI needs more than just the `paymentTitle`.  That information is stored in `paymentData`.

supernova-at:
PR Updated:
  * Adds unit tests to new `braintreeDropin` component

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -0,0 +1,90 @@
...
+ * https://github.com/braintree/braintree-web-drop-in
+ * https://braintree.github.io/braintree-web-drop-in/docs/current/index.html
+ * https://developers.braintreepayments.com/guides/drop-in/overview/javascript/v3
+ */
+
+import React, { Component } from 'react';
+import { bool, func } from 'prop-types';
+import dropin from 'braintree-web-drop-in';
+
+const BRAINTREE_SANDBOX_TOKEN = 'sandbox_8yrzsvtm_s2bg8fs563crhqzk';
```
sirugh: Since venia is a concept/reference perhaps this sort of thing should live in a top-level config/keys file with a big header like `EDIT THESE TOKENS WITH YOUR OWN DEETS`?

Alternatively if this is a one-off, maybe just add a comment :P

supernova-at: Great call, I'll move this to `.env`.

supernova-at: Moved in https://github.com/magento-research/pwa-studio/pull/744/commits/6b85da632da0cecf05793b4b0e5e7dc4b49ffdd9.

supernova-at:
@zetlen , @jimbo , @sirugh I added this to the description, but:

> This PR introduces a new environment variable section for braintree integration to .env.dist. When verifying locally, be sure to update your local .env file with this new section.

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -196,9 +192,22 @@ class Form extends Component {
             return <span className={promptClass}>Add Billing Information</span>;
         }
 
+        let primaryDisplay = '';
+        let secondaryDisplay = '';
+        if (paymentData) {
+            primaryDisplay = paymentData.details.cardType;
```
sirugh: Is `details` always an object? What is `cardType` for check/MO?

packages/venia-concept/src/components/Checkout/form.css
```diff
@@ -18,6 +18,19 @@
     text-transform: capitalize;
 }
 
+.paymentDisplayPrimary {
+    text-transform: capitalize;
+}
+
+.paymentDisplaySecondary {
+    /* The ::first-letter pseudo element below only works on block elements */
```
sirugh: Interesting approach :P My mind immediately jumps to i18n and rtl languages but not sure if we care for Venia or even if that matters. Is there capitalization in rtl? Would browsers apply `::first-letter` in rtl to the right-most character?

packages/venia-concept/src/components/Checkout/flow.js
sirugh: In that case maybe update to use `shape` like some of the other props?

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -1,59 +1,45 @@
...
+    state = {
+        isRequestingPaymentNonce: false
     };
 
     render() {
-        const {
-            availablePaymentMethods,
```
sirugh: I realize now that support for multiple payment methods is being removed in this PR so some of my other concerns around the API request for payment methods is moot.

I really don't have enough context on the Venia project yet to know, but maybe @ericerway or @awilcoxa could let us know whether or not we should keep this around?

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -0,0 +1,86 @@
...
+
+class BraintreeDropin extends Component {
+    static propTypes = {
+        isRequestingPaymentNonce: bool,
+        onError: func.isRequired,
+        onSuccess: func.isRequired
+    };
+
+    componentDidMount() {
+        this.createDropinInstance().then(instance => {
```
sirugh: Can we use async await here?

```js
async componentDidMount {
  this.braintreeInstance = await this.createDropinInstance()
}
```

packages/venia-concept/src/components/Checkout/braintreeDropin.js
sirugh: Another approach would be to surface the instance's availability to the wrapping form like this guy did [here](https://github.com/Cretezy/braintree-web-drop-in-react). Check out `onInstance` and how he ties the dropin to the form. Kind of how you did but with `onSuccess`.

soumya-ashok:
@sirugh here are the payment screens mocks -
https://magento.invisionapp.com/share/HVP22925B2Q
https://magento.invisionapp.com/share/R3P229LXAWQ
Let me know if you'd like to discuss.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -0,0 +1,86 @@
...
+ * https://github.com/braintree/braintree-web-drop-in
+ * https://braintree.github.io/braintree-web-drop-in/docs/current/index.html
+ * https://developers.braintreepayments.com/guides/drop-in/overview/javascript/v3
+ */
+
+import React, { Component } from 'react';
+import { bool, func } from 'prop-types';
+import dropin from 'braintree-web-drop-in';
+
+const { BRAINTREE_TOKEN } = process.env;
```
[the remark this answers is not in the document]
supernova-at: This one's okay because `process.env` gets replaced with an object that has `BRAINTREE_TOKEN` as a key:

https://github.com/magento-research/pwa-studio/blob/464691ac5d9f2a247e5deb9c3d5413c3bf5a8648/packages/venia-concept/webpack.config.js#L121-L135

packages/venia-concept/src/components/Checkout/form.js
supernova-at: For right now it is - check/MO has been completely removed from the application.

If/when other payment methods get added this will definitely have to be updated to handle them.

packages/venia-concept/src/components/Checkout/form.css
supernova-at: I found this [CSS Tricks](https://css-tricks.com/almanac/selectors/f/first-letter/) post informative

packages/venia-concept/src/components/Checkout/braintreeDropin.js
supernova-at: Dang good find I didn't even see that. Perhaps this can be a future thing, I think ours is fine for now.

supernova-at:
PR Updated:

✅ `Submission error causes dropin to be too tall for the slide out`

packages/venia-concept/src/components/Checkout/paymentsForm.css
```diff
@@ -21,8 +26,23 @@
...
+/*
+ * On error, the Braintree container increases in height
+ * in order to show an error message.
+ *
+ * Unfortunately because of a z-index on an internal element (the one
+ * selected below), it overlaps and displays above the minicart heading.
+ *
+ * Here we unset the z-index on the braintree internal element to cause it
+ * to scroll behind the minicart heading, as intended.
+ */
```
jimbo: Great comment. 👍

packages/venia-concept/src/components/Checkout/braintreeDropin.js
jimbo: This kind of workflow will get much simpler with hooks very shortly. As long as this works as expected, it should be fine for now.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -0,0 +1,84 @@
...
+            this.props.isRequestingPaymentNonce &&
+            !prevProps.isRequestingPaymentNonce
+        ) {
+            // Our parent is telling us to request the payment nonce.
+            this.requestPaymentNonce();
+        }
+    }
+
+    render() {
+        return <div id="dropin-container" />;
```
jimbo: We should probably use an `id` that is more specific to `braintree` here, I would think.

packages/venia-concept/src/components/Checkout/__tests__/braintreeDropin.spec.js
```diff
@@ -0,0 +1,45 @@
+// NPM.
```
jimbo: Appreciate the thought, but let's not litter our files with these comments.

Eventually we'll write a lint rule if we want to make our conventions for import ordering more explicit and consistent.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -0,0 +1,84 @@
...
+        } catch (e) {
+            // BrainTree will update the UI with error messaging,
+            // but signal that there was an error.
+            console.warn(`Invalid Payment Details. \n${e}`);
+            this.props.onError();
+        }
+    };
+}
+
+export default BraintreeDropin;
```
jimbo: Is this component not wrapped with `classify`? Why don't we put any classes on the root `div` here?

packages/venia-concept/src/components/MiniCart/miniCart.css
```diff
@@ -1,11 +1,13 @@
 .root {
+    --minicart-height: 100vh;
+    --minicart-header-height: 3.5rem;
```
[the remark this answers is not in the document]
supernova-at: [paymentsForm.css](https://github.com/magento-research/pwa-studio/blob/supernova/payments/packages/venia-concept/src/components/Checkout/paymentsForm.css#L7-L11) needed to know them in order to set a `max-height` so that it could scroll.

If there's a better way to do this definitely let me know.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
supernova-at: Fixed with update 👍

supernova-at:
PR Updated:
  * I believe this bad boy is feature complete and on par with the mocks!

https://magento.invisionapp.com/share/HVP22925B2Q#/screens
https://magento.invisionapp.com/share/R3P229LXAWQ#/screens

That said, there are a few things that differ from the mocks:
  * `Clear Card` button: Braintree doesn't support this
  * The Billing Address fields are slightly different, as required by M2 backend

soumya-ashok:
@supernova-at Great work. A few comments
1. When the user enters the checkout flow, they no longer have access to their shopping cart. The drawer where the shopper enters their information - there should be a mask on the shopping cart at this point. https://magento.invisionapp.com/share/VWP228WG3CB
2. The main call to action is on the right "Confirm order" and the cancel action to the left at the bottom of the drawer - if the shopper clicks cancel, then the mask is removed and they are back in their shopping cart
3. The color for the prompt text to add shipping / billing in the drawers is #C0123F
4. Once the shopper taps into the drawer, they enter a "full screen" overlay where the shopping cart header is not seen. https://magento.invisionapp.com/share/HVP22925B2Q
5. When the "Billing same as shipping" box selection is removed, it looks like the CC information needs to be re-entered. Are we able to preserve the information the user has already entered?
6. In all three drawer-detail views, the primary call to action "Save" needs to be on the right and cancel on the left.
7. "Save" could be a misleading label. I would recommend "Use Card" for the billing screen and "Use Address" as the primary cta labels.
8. Once information has been entered for all 3 drawers, the design has an 'edit' icon at the right end of each drawer to indicate that the information can be edited - https://magento.invisionapp.com/share/7MP229V2U8W

Please let me know if you have questions about any of the comments.

supernova-at:
- [x] When the user enters the checkout flow, they no longer have access to their shopping cart. The drawer where the shopper enters their information - there should be a mask on the shopping cart at this point. https://magento.invisionapp.com/share/VWP228WG3CB

Covered by issue #696.

- [x] The main call to action is on the right "Confirm order" and the cancel action to the left at the bottom of the drawer - if the shopper clicks cancel, then the mask is removed and they are back in their shopping cart

Covered by issue #696.

- [x] The color for the prompt text to add shipping / billing in the drawers is #C0123F

Contained in issue #695.

- [x] Once the shopper taps into the drawer, they enter a "full screen" overlay where the shopping cart header is not seen. https://magento.invisionapp.com/share/HVP22925B2Q

Created #784 to address this.

- [x] When the "Billing same as shipping" box selection is removed, it looks like the CC information needs to be re-entered. Are we able to preserve the information the user has already entered?

Added commit https://github.com/magento-research/pwa-studio/pull/744/commits/4c821c375ce0f8159974305bfcec21841cca3653 to address this.

- [ ] In all three drawer-detail views, the primary call to action "Save" needs to be on the right and cancel on the left.

Added commit https://github.com/magento-research/pwa-studio/pull/744/commits/8783e95d96b7caf0bd5f38672e60df1a67cf3821 to address this.

- [ ] "Save" could be a misleading label. I would recommend "Use Card" for the billing screen and "Use Address" as the primary cta labels.

Added commit https://github.com/magento-research/pwa-studio/pull/744/commits/8783e95d96b7caf0bd5f38672e60df1a67cf3821 to address this.

- [x] Once information has been entered for all 3 drawers, the design has an 'edit' icon at the right end of each drawer to indicate that the information can be edited - https://magento.invisionapp.com/share/7MP229V2U8W

Covered by issue #695.

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -1,90 +1,230 @@
...
+                initialFormValues = {
+                    addresses_same: true
+                };
+            } else {
+                // The addresses are not the same, populate the other fields.
+                initialFormValues = {
+                    ...initialValues,
+                    // Note: we have to 'rename' this property - it doesn't match
+                    // the form field's name ("addresses_same").
+                    addresses_same: initialValues.sameAsShippingAddress
```
sirugh:
```suggestion
                    addresses_same: false
```

Would be identical, no?

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -1,90 +1,230 @@
...
-            return;
+    setPaymentNonce = value => {
+        this.setState({
+            isRequestingPaymentNonce: false
+        });
+
+        // Build up the billing address payload.
+        const formValue = this.formApi.getValue;
+        let billingAddress = {
+            sameAsShippingAddress: formValue('addresses_same') || false
```
sirugh: I'm guessing this `sameAsShippingAddress` code was intended to prevent you from having to make an async/store request for the shipping address within the component, but I feel like if you managed to pull that to this level it would be a little cleaner. In fact I think it would remove the need for much of the code around `sameAsShippingAddress` checks.

```js
if (formValue('addresses_same') {
  billingAddress = await actions.retrieveShippingAddress()
}
```

Of course this may be asking for something that isn't recommended to do in redux, I'm really not sure.

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -215,16 +273,35 @@ export function formatAddress(address = {}, countries = []) {
...
-async function retrieveAddress() {
-    return storage.getItem('address');
+async function clearBillingAddress() {
+    return storage.removeItem('billing_address');
 }
 
-async function saveAddress(address) {
-    return storage.setItem('address', address);
+async function retrieveBillingAddress() {
+    return storage.getItem('billing_address');
```
sirugh: If you keep the `sameAsShippingAddress` stuff around you could move the `removeInvalidKeysFromAddress` step here. Or you could scrub the boolean from the initial setting below:

```js
if (address.isSameAsShippingAddress) {
  return storage.setItem(storage.getItem('shipping_address')
}
```

etc.

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -1,90 +1,221 @@
...
+            </Form>
         );
-        const initialValue =
-            paymentMethod || availablePaymentMethods[0].code || '';
+    }
+
+    /*
+     *  Class Properties.
+     */
+    get billingAddressFields() {
```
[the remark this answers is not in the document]
supernova-at: Eesh, yeah we should have some validation here.

I think it does make sense to do them at the same time, or at the very least take the same approach.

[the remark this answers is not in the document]
sirugh: I added a comment to #770 asking for this.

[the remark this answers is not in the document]
supernova-at: Oh. Oops, I just added a commit that adds validation 😝 .  I'll comment over there too.

packages/venia-concept/src/components/Checkout/paymentsForm.js
supernova-at: Fixed in https://github.com/magento-research/pwa-studio/pull/744/commits/edbf60386b5a24d8e483365a477255e531ae8e81.

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -1,90 +1,230 @@
...
-                    <Button priority="high" type="submit" disabled={submitting}>
-                        Save
+                    <Button className={classes.button} onClick={this.cancel}>
+                        Cancel
+                    </Button>
+                    <Button
+                        className={classes.button}
+                        priority="high"
+                        type="submit"
+                        disabled={submitting}
```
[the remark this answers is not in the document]
supernova-at: Now that the validation is in place, you can no longer submit with an empty address. I could definitely do some work to disable this button but it is less urgent now.

supernova-at:
I added the `requires followup` label just now because after this gets merged we have to update our Magento Cloud Instance to accept Braintree payments.

babel.config.js
```diff
@@ -36,7 +36,7 @@ const config = {
             ]
         },
         test: {
-            plugins,
+            plugins: [...plugins, ['babel-plugin-dynamic-import-node']],
```
jimbo: This plugin is [required](https://jestjs.io/docs/en/webpack#using-with-webpack-2) for Jest to parse dynamic imports as requires.

packages/venia-concept/__mocks__/braintree-web-drop-in.js
```diff
@@ -0,0 +1,7 @@
+const mockCreate = jest.fn();
+
+const dropin = {
+    create: mockCreate
+};
+
+module.exports = dropin;
```
jimbo: Since this is an external module and we want to mock its methods, I've created a [manual mock](https://jestjs.io/docs/en/manual-mocks#mocking-node-modules) for it.

packages/venia-concept/src/components/Checkout/__tests__/braintreeDropin.spec.js
```diff
@@ -18,26 +13,17 @@ const props = {
     onSuccess: mockSuccess
 };
 
-beforeAll(() => {
-    dropin.create = mockCreate;
-});
-
-beforeEach(() => {
-    // We clear this mock so we don't blow away the implementation.
-    mockCreate.mockClear();
```
jimbo: We've configured Jest to automatically clear mocks after every test, so you don't need to call this.

packages/venia-concept/src/components/Checkout/__tests__/braintreeDropin.spec.js
```diff
@@ -1,15 +1,10 @@
 import React from 'react';
 import TestRenderer from 'react-test-renderer';
+import waitForExpect from 'wait-for-expect';
 import dropin from 'braintree-web-drop-in';
-jest.mock('braintree-web-drop-in');
```
jimbo: Since I've defined a manual mock for this module, we no longer need to call `jest.mock()` for it.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -86,7 +84,10 @@ class BraintreeDropin extends Component {
         );
     }
 
-    createDropinInstance = () => {
+    createDropinInstance = async () => {
+        // import the dropin API
+        const { default: dropin } = await import('braintree-web-drop-in');
+
```
jimbo: I'm pretty sure `React.lazy` only works with components, so we'll do this dynamic import manually.

PR #753, opened 2019-01-14
Yarn & Babel 7

jimbo:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will migrate from NPM to Yarn. In the process, it will also upgrade the repo to Babel `7.2.0`.

### Getting Started

```sh
# install yarn via homebrew
brew install yarn

# install packages
yarn

# build peregrine and pwa-buildpack
yarn build:dev

# watch venia and start dev server
yarn watch:venia
```

### Motivation

Package management is currently one of our biggest pain points. Installing and building packages is slow, and our current workflow requires us to reinstall `node_modules` and rebuild every package on a regular basis. These operations should be faster and used less frequently.

Furthermore, as a monorepo, `pwa-studio` has a complex dependency tree. Our packages share a number of dependencies, and some of them depend on other packages within the monorepo. We currently use `lerna` to hoist dependencies into the monorepo root, but `lerna` isn't particularly fast. Worse, each package in this repository should list its own dependencies accurately, but we've blurred the lines between `dependencies` and `devDependencies` just to enable `lerna` to inform us about mismatched dependencies. As a result, most of our `package.json` files are missing dependencies, and our root `package.json` file contains unused dependencies.

### Yarn

Fortunately, `yarn` can help us solve both problems.

#### Hoisting

[Yarn Workspaces](https://yarnpkg.com/en/docs/workspaces/) directly address our monorepo problems. When `pwa-studio` is configured as a monorepo with `workspaces`, running `yarn install` from the root will install, hoist, and dedupe the dependencies listed in each workspace's `package.json` file, plus any dependencies listed in the root `package.json` file. We no longer need to manually update the root when a dependency changes within a workspace.

Workspaces also simplify our package scripts:

Command | Effect
:--- | :---
`yarn workspaces run clean` | run the `clean` script within each package
`yarn workspace @magento/venia-concept run watch` | run the `watch` script within `packages/venia-concept`

#### Speed

But most importantly, `yarn` is fast.

Operation | npm | yarn
:--- | ---: | ---:
`clean` | 0:27 | 0:03
`install` | 0:51 | 0:16
`build` | 0:25 | 0:06
**Total** | **1:43** | **0:25**


package.json
```diff
@@ -12,148 +18,48 @@
...
     "test:debug": "node --inspect-brk node_modules/.bin/jest -i --no-cache",
     "test:dev": "jest --watch",
-    "validate-queries": "npx lerna run validate-queries -- -s",
+    "validate-queries": "cd packages/venia-concept && yarn run validate-queries",
     "watch:all": "node scripts/watch-all.js",
-    "watch:buildpack": "cd packages/pwa-buildpack && npm run -s watch; cd - >/dev/null",
-    "watch:peregrine": "cd packages/peregrine && npm run -s watch; cd - >/dev/null",
-    "watch:venia": "cd packages/venia-concept && npm run -s watch; cd - >/dev/null"
-  },
-  "husky": {
```
sharkySharks: i know you aren't a huge fan of git hooks but can we leave this in?

packages/peregrine/package.json
```diff
@@ -7,12 +7,13 @@
   "description": "The core runtime of Magento PWA",
   "main": "dist/index.js",
   "scripts": {
-    "build": "npx babel src --out-dir dist --ignore '__tests__/,__mocks__/,__fixtures__/' --source-maps --copy-files",
+    "build": "npx babel src --out-dir dist --root-mode 'upward' --ignore '__tests__/,__mocks__/,__fixtures__/' --source-maps --copy-files",
```
zetlen: Should we be using `npx` with `yarn`? Looks like [the jury is still out](https://github.com/yarnpkg/yarn/issues/3937). I'm not strictly sure it's necessary here anymore, anyway.

package.json
```diff
@@ -12,148 +18,48 @@
...
-    "webpack": "~4.25.1",
-    "webpack-babel-env-deps": "~1.4.3",
-    "webpack-bundle-analyzer": "~3.0.3",
-    "webpack-cli": "~3.1.2",
-    "webpack-dev-server": "^3.1.14",
-    "workbox-webpack-plugin": "~3.6.1",
-    "write-file-webpack-plugin": "~4.2.0",
-    "xml": "~1.0.1"
-  },
-  "dependencies": {}
```
sharkySharks: did all these dependencies just get removed? 😳

packages/venia-concept/webpack.config.js
```diff
@@ -77,7 +79,11 @@ module.exports = async function(env) {
                     use: [
                         {
                             loader: 'babel-loader',
-                            options: { ...babelOptions, cacheDirectory: true }
+                            options: {
+                                cacheDirectory: false,
```
zetlen: Did we try a cache directory and it was buggy?

packages/venia-concept/webpack.config.js
```diff
@@ -77,7 +79,11 @@ module.exports = async function(env) {
...
+                                envName: mode,
+                                rootMode: 'upward'
```
zetlen: Lol, upward.

packages/pwa-buildpack/src/ExtensionComponentWrap/.babelrc.js
```diff
@@ -0,0 +1,31 @@
...
+    'not firefox 47',
+    'not firefox 48',
+    'not firefox 51',
+    'not firefox 52',
+    'not ios 8.1',
+    'not ios 9.3',
+    'not safari 5.1',
+    'not safari 9.1'
+];
+
```
zetlen: Seems like this could be DRYer--could you not just import this `browsers` array from the root babelrc.js?

packages/venia-concept/yarn-error.log
```diff
@@ -0,0 +1,12112 @@
+Arguments:
```
sharkySharks: something tells me we want to add this file to the `.gitignore`

zetlen:
Regarding CircleCI config: https://circleci.com/docs/2.0/yarn/ seems pretty straightforward. I know that the cache parts are pretty weird, but fortunately this Yarn doc has some stuff you can cut and paste right over what we have today.

package.json
```diff
@@ -12,148 +18,48 @@
     "node": ">=10.14.1"
   },
   "scripts": {
-    "build": "npx lerna run --stream build -- -s",
-    "clean:all": "npx lerna run clean && npx lerna clean --yes && rimraf ./node_modules",
-    "clean:dist": "npx lerna run clean",
+    "build": "yarn workspaces run build",
+    "build:dev": "yarn workspaces run build:dev",
+    "clean:all": "yarn workspaces run clean && rimraf ./node_modules",
```
sharkySharks: would we also want to clear the yarn [cache](https://yarnpkg.com/lang/en/docs/cli/cache/#toc-yarn-cache-clean)?

sharkySharks:
Getting this warning when running `yarn install`:

screen shot 2019-01-14 at 10 21 26 am

sirugh:
I left a few comments but I was also unable to run the `watch:venia` command I get this output after first running `yarn install`:

```
❯ yarn run watch:venia
yarn run v1.13.0
$ cd packages/venia-concept && yarn run watch; cd - >/dev/null
$ webpack-dev-server --progress --color --env.mode development
Using environment variables from .env
internal/modules/cjs/loader.js:605
    throw err;
    ^

Error: Cannot find module 'babel-helper-module-imports'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:603:15)
    at Function.Module._load (internal/modules/cjs/loader.js:529:25)
    at Module.require (internal/modules/cjs/loader.js:657:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/rugh/code/pwa-studio/packages/pwa-buildpack/dist/magento-layout-loader/babel-plugin-magento-layout.js:4:24)
    at Module._compile (internal/modules/cjs/loader.js:721:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:732:10)
    at Module.load (internal/modules/cjs/loader.js:620:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:560:12)
    at Function.Module._load (internal/modules/cjs/loader.js:552:3)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
✨  Done in 1.26s.
```

package.json
sharkySharks: I think I have more questions around when to use `yarn workspaces` and when to leverage our use of `lerna bootstrap` to install all of our package dependencies, and running commands like `lerna run clean` to remove all the package node_modules folders for us. Isn't lerna our monorepo package manager? 🤔

sirugh:
There are still quite a few references to `npm` in docs and scripts. Can you update these accordingly?

packages/venia-concept/webpack.config.js
jimbo: No, this was supposed to be temporary while I was upgrading, to make sure it didn't interfere. I'll set this back to `true` now.

packages/pwa-buildpack/src/ExtensionComponentWrap/.babelrc.js
jimbo: Yes, it should be. I'll either do that or make a separate module for it.

jimbo:
> Do I need to do anything other than yarn install in root?

Yes, you need to build `peregrine` and `pwa-buildpack` (same as you do already). I've added steps to the PR description.

package.json
```diff
@@ -12,148 +18,48 @@
...
-    "babel-plugin-syntax-jsx": "~6.18.0",
-    "babel-plugin-transform-class-properties": "~6.24.1",
-    "babel-plugin-transform-es2015-modules-commonjs": "~6.26.0",
-    "babel-plugin-transform-object-rest-spread": "~6.26.0",
-    "babel-plugin-transform-react-jsx": "~6.24.1",
-    "babel-plugin-transform-react-remove-prop-types": "~0.4.13",
-    "babel-plugin-transform-runtime": "~6.23.0",
-    "babel-preset-env": "~1.6.1",
-    "babel-runtime": "~6.26.0",
-    "boxen": "~2.0.0",
```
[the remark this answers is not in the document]
jimbo: Yes, that part of the documentation will need to be changed. With Yarn Workspaces, packages from each workspace's `package.json` are hoisted.

package.json
jimbo: A main objective of this change is to stop using `lerna` for most of our monorepo tasks—especially for install, build, and clean operations. The goal is to have each package describe its own dependencies **accurately**.

Run `yarn` from the root directory and it will install all dependencies for the monorepo. Similarly, `yarn build:dev` and `yarn:clean` from the root will use `workspaces` to run that command in each workspace. It should be faster.

packages/venia-concept/yarn-error.log
jimbo: Correct, and done. 👍

package.json
jimbo: Yes, I'll add it back as a very last step when this PR is approved.

package.json
jimbo: No; they were relocated to the individual packages. Each package should list all of its own dependencies correctly. With `yarn workspaces`, even though those packages are hoisted to the root `node_modules`, they don't need to be added to the root `package.json`.

packages/peregrine/package.json
sirugh: The only reason to npx babel would be if we don't have it installed as a dependency of this repo.

I think if you install as a dep you can do away with `npx` here. I guess it just depends on the reason.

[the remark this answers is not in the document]
jimbo: I've updated these to not use `npx`.

packages/pwa-buildpack/src/ExtensionComponentWrap/.babelrc.js
jimbo: Done.

babel.config.js
```diff
@@ -0,0 +1,73 @@
+// precise browser targeting
+// reduces the amount of transpilation necessary
+// more: https://twitter.com/jamiebuilds/status/1022568923726639104
+const browsers = [
```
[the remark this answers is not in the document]
jimbo: Done.

packages/venia-concept/webpack.config.js
jimbo: Done.

lerna.json
```diff
@@ -1,11 +1,5 @@
...
-    "packages": [
-        "packages/peregrine",
-        "packages/pwa-buildpack",
-        "packages/upward-js",
-        "packages/upward-spec",
-        "packages/venia-concept"
-    ],
-    "npmClient": "npm"
+    "npmClient": "yarn",
+    "useWorkspaces": true
```
sharkySharks: can we just remove this file now that we are going to use yarn workspaces?

package.json
sharkySharks: sounds good, i think from our discussion I'd like to remove lerna as a tool then if we are going to switch to using these workspaces and clear it out of our dependencies.

jimbo:
[ ] Circle
[ ] now.sh

lerna.json
sharkySharks: looks like lerna.json is obsolete now, same question as above, can we remove this file now? It looks like you removed lerna from everywhere so far but [this file](https://github.com/magento-research/pwa-studio/blob/release/2.0/scripts/watch-all.js#L180). We probably want to make that change as well.

package.json
sharkySharks: still would like this not to be removed.

README.md
```diff
@@ -67,23 +64,25 @@ See our [Troubleshooting][] guide if you run into any problems.
...
-## Things not to do
+## Tips
 
-* Our monorepo is set up so that `npm install` can cross-link dependencies (such as Venia's dependency on Buildpack) without any extra tools. Do not run `lerna bootstrap`.
-* All devDependencies are installed in the repository root. This means that **all scripts must be run from repository root**; otherwise, the locally installed CLI commands they use will not be available.
-* Production dependencies are sometimes installed in child packages, but for some projects, such as Venia, it makes no sense to have production dependencies, because of bundling.
-* Don't check in a big change to `package-lock.json`, and don't check in any `package-lock.json` files but the root one.
-* Make sure to run `npm run prettier` and `npm run lint` before any commit you intend to push. You may want to set up a [Git hook] for this.
+* Don't run `npm install`! Since this project has been configured with [Yarn Workspaces][], run `yarn install` to properly install, hoist, and cross-link dependencies.
+* Don't check in `package-lock.json`. There is only one lockfile, `yarn.lock`, and it's in the root directory.
```
supernova-at: ❤️

Issue #668, opened 2018-12-14
Images should be lazy-loaded and responsive


rowan-m:
## This issue is for the following packages:

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [ ] Bug
- [x] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

## Description

Images are currently loaded via an `<img>` tag. On simpler, lighter pages where the majority of the content is within the first viewport, this is fine. However on heavier, longer pages such as a category listing this has a significant performance impact.

## Expected result:

Images should be loaded on demand and at an appropriate size for the viewer.

## Possible solutions:

#398 addresses serving the appropriate format, but the front-end still needs to take advantage of that. This should be achieved via the `<picture>` element to specify multiple `<source>` options for both optimal image formats and sizes.
See https://developers.google.com/web/fundamentals/design-and-ux/responsive/images for more detail.

Lazy loading images should make use of `IntersectionObserver` to check for proximity of elements to the viewport and dynamically setting the `src` and `srcset` dynamically to trigger loading. Images should have at the least as static fallback / loading image but generating low fidelity previews may also be possible.
See https://developers.google.com/web/fundamentals/performance/lazy-loading-guidance/images-and-video/ for more detail.

Lazy loading probably wants to be quite conservative as product images are key to the experience, so don't wait until the element is in view to load but trigger it when they are close the viewport.

There's a fair amount of behaviour here that needs to encapsulated, so it feels like this deserves its own component. There may also be existing React components here that would be appropriate to bring in.

vitalics:
@rowan-m for react we're used redux and trunk for asynchronous interaction. In this case I presume  we should get an image from store after triggering action(e.g. `getImagesFromBackend()`) and before getting an answer from store, use variable `isLoading` and `isLoaded` which declared on store.

real34:
Gatsby image is one interesting component implementing this: https://www.gatsbyjs.org/packages/gatsby-image/

There are rooms for improvement though, but it would be a good idea to see what could be shared between both implementations.

awilcoxa:
related to: https://github.com/magento/pwa-studio/issues/1958

tjwiebell:
[...]

Issue #636, opened 2018-12-10
EPIC: Accessibility improvements for Venia (Lighthouse)


ericerway:
From pwa-studio created by [ericerway](https://github.com/ericerway) : magento-research/pwa-studio#594  

## This issue is for the following packages:

[X] `venia-concept`
[ ] `pwa-buildpack`
[X] `peregrine`
[ ] `pwa-devdocs`
[ ] `upward-js`
[ ] `upward-spec`

## This issue is a:

[ ] Bug
[X] Feature suggestion
[ ] Documentation issue
[ ] Other (Please Specify)

### Environment

| Question                    |    Answer     |
| --------------------------- | :-----------: |
| Magento version             |  |
| Operating System + version  |  |
| node.js version (`node -v`) |  |
| npm version (`npm -v`)      |  |

## Description

As a shopper, I want to experience a fully accessible Venia storefront regardless of my abilities so that I can achieve WCAG 2.0 recommendations and Google Lighthouse recommendations.

https://www.w3.org/WAI/standards-guidelines/wcag/

## Expected result:

Updates to elements, colors, and language on key Venia experiences to improve accessibility as defined Google Lighthouse.

![image.png](https://images.zenhubusercontent.com/5a71f21a4b5806bc2bc51012/cdeea5a5-1adf-4099-8188-581547e6c1a2)

Categories: (affected)
- Elements Have Discernible Names
- Color Contrast is Satisfactory
- Page Specifies Valid Language

Pages: (affected)
- Homepage
- Category
- Product Details

## Tasks
- [ ] Update Venia home, category, and PD pages with recommended fixes from Lighthouse
- [ ] Review Google Lighthouse scores for Accessibility to verify (Lighthouse 4.0.0-alpha.1)
- [ ] Update documentation (as appropriate)

## Possible solutions:

Updates to Venia and UI components (peregrine) as appropriate.

dmtrkad:
@ericerway , Are there any PBIs in "Ready for development" status that could be picked by us into this epic?

ericerway:
Not yet, we have determined that this needs to be further groomed and associated with the core Peregrine (components) work first.

ericerway:
@awilcoxa do we need both epics for accessibility or can we consolidate? (reference: #623)

awilcoxa:
closing as duplicate

PR #757, opened 2019-01-14
feat(images): Image and other performance optimizations

### LINKED ISSUE #398: Advanced Image optimization for Magento with UPWARD

[ ] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-devdocs`
[ ] `upward-js`
[ ] `upward-spec`

## This issue is a:

[ ] Bug
[x ] Feature suggestion
[ ] Documentation issue
[ ] Other (Please Specify)

## Description

Today, Magento PWAs retrieve media assets (images, video, pdf, etc.) from Magento APIs (GraphQL/REST) with no optimization for file or image size. This creates performance issues on the display side for PWA. There are systems that exist for retrieval of resized/compressed images but they are not currently accessible to PWA due pre-existing dependencies and requirements. 

## Expected result:

A Magento Progressive Web App can retrieve compressed and/or resized images on demand in order to ensure best-in-class performance. This optimization should apply to the initial load of a PWA as well as throughout the entire customer journey/use. 

## Possible solutions:

Solutions that require discovery/investigation effort: 
1. Escalate requirement/need with API teams to ensure long-term scalable solution with single ownership. Risk: requires buy-in across functional teams for prioritization/delivery leading to slower velocity. 
2. Write middleware within UPWARD  to provide media optimization/resizing. Risk: potential throw away work that is redundant to existing services. Requires knowledge/experience with may not be readily available. 
3. Make a single purpose module. Risk: Large amount of effort and tech debt. 

Chosen solution: **2**. Write middleware within UPWARD.

### Tasks
- [ ] Add binary and JSON data handling to UPWARD file and data spec.
- [ ] Add resolvers for transformation.
- [ ] Add image compression to strategies and implement in upward-js.

zetlen:
## This PR is a:

- [x] New feature
- [x] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

- Add image resizing and compression via two strategies, one when Fastly
is available and one when Node is running
- Add gzip/brotli compression
- Replace image tags throughout Venia with new ResponsiveImage component
using srcsets
- Adjust Venia CSS and JS implementations to work with responsive images
- Make some other small changes for improved Lighthouse scores

### Summary

PWA Studio now provides image optimization and response compression services.

This went through **many** iterations, but we arrived at a solution that doesn't burden the UPWARD spec with extra functionality. Instead, Venia uses UPWARD as a router, to declare what image optimization URLs it expects, and then redirects to different locations based on the environment (Fastly for Enterprise Cloud production, the local Node `express-sharp` transformer for all other scenarios).

### Architecture

It's important to note that **the onboard image optimization is a feature of Buildpack**, and Venia is built to expect it in development and staging mode. Specifically, Venia implements the URL generation in two places: `src/util/makeMediaUrl.js`, and `venia-upward.yml`. The logic in `makeMediaUrl.js` generates URLs that `venia-upward.yml` can route to the appropriate image optimization services.

### Usage

In practice, developers should not have to generate these URLs by hand. They can use `src/util/makeMediaUrl.js` to get a URL directly, but more often, they can use the `ResponsiveImage` abstraction to generate images with `srcset` and `sizes` attributes, which helps them use native browser features to optimize.

### Reviewing

I've named the maximum number of reviewers for this, as it is a pretty big change.

In doing this refactor, I've tried several different means of centralizing logic, and they're all on display here in one form or another; though I think it's all right to have variety, I would totally expect to be asked to consolidate the approaches into one. They include:

- Declaring media conditions inline
- Declaring media conditions in a top-level component and prop drilling down to the image render
- Declaring media conditions in a CSS file using the CSS Modules `:export {}` feature
- Using CSS media queries alone
- Combining img srcsets, sizes, and CSS media queries
- Setting background images and hiding image tags
- Using several different kinds of units in the media conditions

I welcome feedback on all this, including the API of `ResponsiveImage` as you see here.

Lastly, I also want feedback about the choice to use 301 redirects to render images. If the redirects are to the same host, they won't result in another DNS lookup, but they do cause a slight delay; a further improvement to UPWARD should allow us to rewrite URLs instead of redirecting them. Other suggestions would be appreciated!

### Issues
Closes #398.
Addresses *some* of #668, but lazy loading still needs work.
Addresses a couple accessibility issues in #636.


packages/venia-concept/src/components/Main/main.js
```diff
@@ -33,9 +33,9 @@ class Main extends Component {
 
         return (
             <main className={classes.root}>
-                <Header />
-                <div className={classes.page}>{children}</div>
-                <Footer />
+                <Header style={{ flex: 1 }} />
```
zetlen: This is not ideal, but it was necessary to fix a bad stretch problem with the parent container that was blocking my media query tests.

I need to inject styles into the `Header` from the parent component. If `classify` allowed us to compose classnames by joining them with spaces, then I could inject extra styles into `Header` just by knowing the classnames, e.g.

main.css:
```css
.mainheader {
  flex: 1
}
```

```jsx
<Header classes={{ header: mainheader }} />
```

packages/venia-concept/src/components/ProductImageCarousel/thumbnail.js
```diff
@@ -15,15 +15,23 @@ class Thumbnail extends Component {
 
     render() {
         const { classes, item } = this.props;
-        const src = item.file
-            ? makeProductMediaPath(item.file)
-            : transparentPlaceholder;
```
zetlen: I went back and forth on whether to add transparent placeholder functionality to `ResponsiveImage`, but I ended up worrying that the component was doing too much.

packages/venia-concept/src/util/makeMediaUrl.js
```diff
@@ -0,0 +1,37 @@
...
+
+const absoluteUrl = /^https?:\/\//i;
+
+const joinUrls = (base, url) =>
+    (base.endsWith('/') ? base.slice(0, -1) : base) +
+    '/' +
+    (url.startsWith('/') ? url.slice(1) : url);
+
+const mediaPathPrefixes = {
+    product: '/media/catalog/product/',
```
zetlen: These should be provided from env by a Webpack EnvironmentPlugin, so that we can swap out URL generation schemes from a central location.

packages/venia-concept/templates/preloads.mst
```diff
@@ -1,3 +1,9 @@
 {{#urlResolver}}
 <script type="application/json" id="url-resolver">{ "type": "{{type}}", "id": "{{id}}" }</script>
 {{/urlResolver}}
+<script>
+    const fontStylesheet = document.createElement('link');
```
zetlen: As a link tag in the head, this font was a blocking resource. This does run the risk of a FOUC, but I'd rather have that than a blocking load.

packages/venia-concept/templates/scripts.mst
```diff
@@ -1,2 +1,2 @@
-{{^isDevelopment}}<script defer src="/js/vendor.js"></script>{{/isDevelopment}}
+<script defer src="/js/vendor.js"></script>
```
zetlen: Developer builds now generate a vendor chunk.

packages/venia-concept/validate-environment.js
```diff
@@ -3,6 +3,21 @@ function validateEnvironment(env) {
     const { str, bool, url } = envalid;
 
     const validation = {
+        IMAGE_CACHE_EXPIRES: str({
```
zetlen: I'd like to increasingly move all of our configuration APIs into environment variables, so that Webpack config doesn't have to be as long.

packages/venia-concept/venia-upward.yml
```diff
@@ -21,6 +24,31 @@ response:
...
+    headers:
+      inline:
+        Location:
+          engine: mustache
+          provide:
+            width: $match.$1
+            path: $match.$2
+          template:
+            when:
+              - matches: env.USE_FASTLY
```
zetlen: This may not be the environment variable that ends up available, but it's easy to swap out.

scripts/watch-all.js
```diff
@@ -54,10 +54,12 @@ function whenQuiet(childProcess, timeout = 1000) {
 
         // wait until stdout is quiet for a while before resolving this promise.
         const resolveDebounced = debounce(() => {
+            childProcess.stderr.on('data', () => resolveDebounced());
```
zetlen: I noticed that when Buildpack or Peregrine failed the initial build, the watch:all script didn't display errors. This fixes that.

zetlen:
https://magento-venia-sgbwkcbdhc.now.sh/

[image](https://magento-venia-sgbwkcbdhc.now.sh/)

🛬

packages/venia-concept/venia-upward.yml
```diff
@@ -21,6 +24,31 @@ response:
...
+                pattern: '[^0]'
+                use:
+                  inline: '/{{path}}?auto=webp&format=pjpg&'
```
pcvonz: Do we want to use webp? 

Firefox 65 just added support and Safari doesn't support it at all. There is a polyfill we could use.

.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ common_settings:
 
   install_packages: &install_packages
     name: Install NPM packages from lockfile
-    command: npm ci --ignore-scripts && npx lerna bootstrap --hoist --no-ci
+    command: npm ci --ignore-scripts && npx lerna bootstrap --hoist --no-ci && npm rebuild
```
zetlen: Added `npm rebuild` because this original command didn't run the install scripts that build native extensions, like `sharp`.

packages/venia-concept/src/components/Main/main.js
codeAdrian: https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/Header/header.js#L46

I think I've done something similar in custom theme. From this file, we can see that either `classes.open` or `classes.closed` will be applied to the Header HTML element. So you can override these 2 classes from parent by doing the following:

- `<Header classes={{ open: classes.mainheaderOpen, close:  classes.mainheaderClose}} />`
- In the main.css file where mainheaderOpen and mainheaderClose are located, compose the class from Header component (`open` and `close` respectively)

Let me know if this works. I will check tomorrow directly on your branch just to be sure and get back to you.

codeAdrian: To follow up on my suggestion, I have checked this locally and it works. Let me know if this solution is alright:

This is an example of overriding header class only. Same technique can be applied for others.

main.css
```css
.headerOverride {
    composes: root from 'src/components/Header/header.css';
    flex: 1;
}

```
main.js
```jsx
<Header
     classes={{
          open: this.props.classes.headerOverride,
          closed: this.props.classes.headerOverride
     }}
/>
```

Note: both close and open classes inherit from root class in header.css, so I think that they can be safely removed.

Result:

screen shot 2019-01-15 at 2 50 59 pm

rowan-m:
Taking a peek at this now. Definitely looking better! Verified performance improvements on https://www.webpagetest.org/result/190115_K2_7643452a2796eb3efe22b315d60b977d/ with the perf score now at 74 with no image related issues listed.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -85,6 +87,9 @@ const PWADevServer = {
                         })
                     )
                 );
+            },
+            before(app) {
+                addImgOptMiddleware('/scale', app, config.env);
```
rowan-m: The `/scale` value here feels a bit fragile. It looks like it's coupled to the `ImageOpt` configuration in `/packages/venia-concept/venia-upward.yml` but I had to go up and down the diff a few times to work out what was connected where. Could this be specified in that configuration instead?

packages/venia-concept/src/components/ResponsiveImage/ResponsiveImage.js
```diff
@@ -0,0 +1,166 @@
...
+ * Optionally, takes a render callback so you can use other DOM structures
+ * besides literal img tags to display responsive images. The img tag is
+ * required, however.
+ */
+
+import React, { Component } from 'react';
+import { arrayOf, func, number, oneOf, string } from 'prop-types';
+import makeMediaUrl from 'src/util/makeMediaUrl';
+
+class ResponsiveImage extends Component {
```
rowan-m: I'm not sure this component is working correctly from looking at https://magento-venia-etnammagpk.now.sh/

It looks like the inline style image on the surrounding `span` is always overriding the `img`. It does't seem to resize (presumably as it's the low-res preview?) but the underlying `img` also doesn't look correctly sized.

For example, disable the `opacity: 0` in DevTools and it should show.
Before:
screenshot 2019-01-15 at 16 36 28

rowan-m: After:
screenshot 2019-01-15 at 16 38 22

packages/venia-concept/src/components/ResponsiveImage/ResponsiveImage.js
```diff
@@ -0,0 +1,166 @@
...
+         * Updating the `style` property with a CSS variable does not trigger a
+         * repaint, so we have to modify the string in the style attribute.
+         */
+        let callback;
+        if (typeOfArg === 'function') {
+            callback = propToSet;
+        } else if (typeOfArg === 'string') {
+            const styleRegexp = new RegExp(`${propToSet}:\\s*[^;]+(;?)`);
+            callback = (element, currentSrc) => {
+                const declaration = `${propToSet}: url("${currentSrc}")`;
```
rowan-m: Would be good to investigate some alternatives here that don't (unless I'm missing something) result in a minimum of two requests per image: one for the low-res preview set here, one for the actual image. Maybe this could create an inline preview, i.e. super low-res then CSS blurred or just a background colour?
image

zetlen: Actually it's both better and worse: currently, UPWARD cannot rewrite URLs while proxying, so instead we're using 301 redirects--one for each image. They are to the same host, at least, but that's definitely two TCP handshakes. It's a tech debt item.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
zetlen: That's a good idea. We should push more config to `venia-upward.yml` in general.

We have two goals here that are sometimes opposed:

- To centralize config on a single resolution path and contract (env vars -> UPWARD  YAML -> defaults -> code)
- To be as zero-config as possible

It makes sense to me to have a sensible default route for the onboard, dev-and-staging-only image opto service. But it should definitely be clearly documented, and there should be a very friendly configuration debugging mode, where you can see what comes from where.

packages/venia-concept/src/components/ResponsiveImage/ResponsiveImage.js
zetlen: Yes, that needs some work. With the new art direction capabilities of `srcset` and `<picture>`, I'm not sure we need to use CSS backgrounds here.

I left them in because I wanted to surgically refactor image tags rather than whole components if I could help it--plus, this type of usage forced me to think harder about the API of the ResponsiveImage tag. However, it may not be necessary to deviate from the sensible default of image + srcset, so if you think I should, I can simplify this code and remove the render prop feature.

I also observed while trying to demo this that resizing the viewport doesn't seem to trigger a dynamic update to the `currentSrc` of the image. I haven't gone and StackOverflown that yet, but I was wondering if that was Chrome's behavior on purpose. Maybe it thinks that img tags semantically should use only one source per the lifetime of the element, to conserve bandwidth. Any ideas, @rowan-m ? (I'll update this later if I learn this on my own through independent research).

brendanfalkowski: I think for `srset` the browser will apply the largest image it has in cache rather than request a smaller image during resize (even though it more accurately fits the media query). The images are assumed to be equivalent content, and the browser makes the swap call.

Whereas for `<picture>` it will swap when told because that's semi-intended for art direction (i.e. different content per media query) and the browser must swap to maintain the correct presentation.

packages/venia-concept/src/util/makeMediaUrl.js
```diff
@@ -0,0 +1,37 @@
+import memoize from 'lodash/memoize';
+
+const absoluteUrl = /^https?:\/\//i;
+
+const joinUrls = (base, url) =>
+    (base.endsWith('/') ? base.slice(0, -1) : base) +
+    '/' +
+    (url.startsWith('/') ? url.slice(1) : url);
```
supernova-at: [Node's path module](https://nodejs.org/api/path.html#path_path_normalize_path) might be able to help here.

packages/venia-concept/src/util/makeMediaUrl.js
```diff
@@ -0,0 +1,37 @@
+import memoize from 'lodash/memoize';
```
supernova-at: We pull in [lodash.debounce](https://github.com/magento-research/pwa-studio/blob/release/2.0/package.json#L69) as its own dependency, should we do that for `memoize` as well?

packages/venia-concept/templates/preloads.mst
supernova-at: I went down a whole rabbit hole of [interesting research](https://codepen.io/tigt/post/async-css-without-javascript#the-future-5) on this but it does seem like this is the recommended way to accomplish async loading a stylesheet 👍 .

packages/venia-concept/venia-upward.yml
supernova-at: Is this documented anywhere?

PR #761, opened 2019-01-15
Update buttons

jimbo:
## This PR is a:

- [ ] New feature
- [x] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will update buttons to be consistent with the updated design.


ericerway:
@jimbo could this improve our button accessibility around naming or will that be done in Peregrine? Would be great to close the gap on Lighthouse here but understand if that's for a different effort.
image

jimbo:
@ericerway We could certainly go through Venia and add [`aria-label`](https://developers.google.com/web/tools/lighthouse/audits/button-name) to each button.

PR #768, opened 2019-01-18
feat(dev): Onboard image resizing middleware

zetlen:
## This PR is a:

- [x] New feature
- [ ] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will add dynamically compressed and resized images to Venia. It includes:

- [Buildpack] Add an image optimization middleware to Utilities
- [Buildpack] Enhance PWADevServer to use new image optimization middleware
- [Venia] Enhance staging server script to use new image optimization middleware
- [Venia] Add and document additional environment variables to configure/debug image optimization middleware
- [Venia] Replace images in existing components with optimized images (hardcoded sizes instead of the additional logic in #757)
- [Venia] Add UPWARD configuration to route image requests to a CDN in production, where the middleware will not be present

## Additional information

This is a subset of #757 containing only the middleware and the smallest possible Venia changes. #757 should also slim down into just the ResponsiveImage dynamic component, so it can be evaluated separately.


packages/venia-concept/.env.dist
```diff
@@ -22,8 +22,14 @@ MAGENTO_BACKEND_URL="https://release-dev-rxvv2iq-zddsyhrdimyra.us-4.magentosite.
...
+#   By default these are `/media/catalog/product` and `/media/catalog/category`,
+#   but if your Magento instance does something different, they're configurable
+#   here.
+#   MAGENTO_BACKEND_PRODUCT_MEDIA_PATH="/media/catalog/product"
+#   MAGENTO_BACKEND_CATEGORY_MEDIA_PATH="/media/catalog/category"
+#
 #   These legacy environment variables from 1.x are no longer necessary:
-#   - `MAGENTO_BACKEND_PRODUCT_MEDIA_PATH` : Now that we are using both category and
 #      product URLs, we are hardcoding this until we have an image optimization
 #      solution.
```
jimbo: Take another look at this comment; it looks like part of it was removed.

jimbo:
Yep, this seems to work just fine. Of course, the images look awful now, but I assume that's something we'll solve with a `ResponsiveImage` component.

packages/pwa-buildpack/src/WebpackTools/plugins/UpwardPlugin.js
```diff
@@ -10,13 +10,24 @@ const httpsAgent = new https.Agent({ rejectUnauthorized: false });
...
+        const oldBefore = devServer.before;
+        devServer.before = (app, ...rest) => {
+            app.use(upward.bestPractices());
+            if (oldBefore) oldBefore(app, ...rest);
+        };
         const oldAfter = devServer.after;
         devServer.after = (app, ...rest) => {
-            app.use((req, res, next) => this.handleRequest(req, res, next));
             if (oldAfter) oldAfter(app, ...rest);
+            app.use((req, res, next) => this.handleRequest(req, res, next));
```
sirugh: Why did you move this line after 29?

zetlen: @sirugh PWADevServer now exposes a builtin image resizing middleware at `/img` (though that route is configurable with an env var). Venia is designed to run on PWADevServer in development mode. If PWADevServer is absent, then the route falls to UPWARD to resolve and then the addition to `venia-upward.yml` handles it (by forwarding it to Fastly if possible).

zetlen:
@sirugh Doesn't seem to be broken anymore. Feel like trying again?

packages/pwa-buildpack/src/WebpackTools/plugins/UpwardPlugin.js
sirugh: Gotcha. Because line 27 will be falsy in non dev.

zetlen:
@jimbo @sirugh Added some comments and quick fixes. Ready for re-review.

I don't _want_ the images to look terrible and they don't have to, we just need to adjust the sizes!

jimbo:
Looks good. 👍

PR #770, opened 2019-01-20
PWA-749 checkout address form validation

### LINKED ISSUE #749: Required fields are not indicated in checkout and on placing order throws error

- [X] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [X] Bug
- [ ] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

### Environment

| Magento version             | 2.3 |
| Operating System + version  | Ubuntu 16.04 |
| node.js version (`node -v`) | 8.12 |
| npm version (`npm -v`)      | 6.4.1 |

## Description
When entering shipping address in checkout, don't enter telephone field which is a required field.
Proceed on and fill in checkout form. When the order is about to placed the rest endpoint throws telephone is needed in shipping address. On entering telephone only billing address is changed and not shipping address. So even after filling telephone order could not be placed.
When shipping method is re-selected the order can be placed.

## Expected result:
- telephone input should be validated as required field.
- saving shipping address should save shipping address in quote.
- need a checkbox to indicate same as shipping address like magento 2 ordinary storefront and if that's not checked a billing address input should be created.
- error message should be displayed that checkout step is failed

VitaliyBoyko:
This closes #749. Required fields are not indicated in checkout and on placing order throws an error 

## This PR is a:

- [ ] New feature
- [x] Refactor
- [x] Bugfix
- [ ] Test for existing code
- [ ] Documentation

Added validation to the address form.

## Summary
image


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=770) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=770) before we can accept your contribution.<br/>

sirugh:
Can we hold off on this until #744 is merged? There are address forms that require validation in that Pr and it would be nice to have that handled in this PR as well.

zetlen:
I like this PR! It's well organized and clear. Thanks @VitaliyBoyko ! Now that we've merged #744 can you please resolve the conflict in the test file and confirm that the fix still works?

supernova-at:
@VitaliyBoyko now that #744 is merged, check out the [shared form validator functions](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/util/formValidators.js), I added them so that we can share validator functions across components.

We may want to put in a special case in [hasLengthAtLeast](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/util/formValidators.js#L9-L13) for required fields though, I think we can improve the UX there.

I'm thinking something like:

```js
export const hasLengthAtLeast = (value, minimumLength) => {
    if (!value) {
        return 'Is required.';
    }
    else if (value.length < minimumLength) {
        return `Must contain at least ${minimumLength} character(s).`;
    }
};
```

sirugh:
Is there a way to order validators or have multiples with `informed`?

supernova-at:
> I would keep it in the separate validators hasLengthAtLeast and isNotEmpty.

I thought about this too and I'm really not opposed to it. I think the readability is better if we put the "is required" message in `isNotEmpty` but it's more composable if we put it in `hasLengthAtLeast` (we don't have to copy messages all over the place).

Also, for a field that must be filled in I think "Is required" is a better error message than "must contain at least 1 character(s).".

> I suggest using two levels of validators.
> The first level will be common for all components as @supernova-at has implemented in packages/venia-concept/src/util/formValidators.js
> The second level is file validators.js within the certain component. We inject there needed validator or validators from util/formValidators.js for each field, adding custom error messages and other business logic applicable to that field.

I feel your pain on this one and likely if this were a real app we'd want the ability to customize them to that level but let's keep it simple and consistent for now and just keep the one level of validators.

VitaliyBoyko:
Hi @supernova-at @sirugh 
I've reviewed current react validate API and researched for solutions in other applications an libraries and figured out that the only one way to have multiple validations on a field it's make modifications on both the field and the form components.
Instead of this, I suggest using a function which will wrap multiple validators into a single callback what is more elegant in my opinion.
Here is [initial implementation](https://github.com/magento-research/pwa-studio/pull/770/files#diff-af035a9408272f297ea0e98183ecea93R1) and [usage](https://github.com/magento-research/pwa-studio/pull/770/files#diff-1959c0a18eaa236cc210e4e522cc98f0R142).

Let me know if the solution acceptable. If so, I gonna adjust and fix test and we are ready to go.
Thanks!

PR #774, opened 2019-01-23
[Bug] Handle title and meta tags for navigation on client side

vasilii-b:
## This PR is a:

- [ ] New feature
- [ ] Refactor
- [X] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

1. Solve issue magento-research/pwa-studio#422
2. Adds page default title from System Configuration
3. Appends prefix and suffix (from System Configuration) to the page title if necessary
4. Adds category title, meta title, meta keywords and meta description to the current App state
5. Adds product title, meta title, meta keywords and meta description to the current App state
6. Introduces the `react-helmet` package to `venia-concept`

## Additional information
⚠️ This PR is dependent on one opened PR magento/graphql-ce#318. 
The thing is that that PR solves GraphQl issue for getting information about store design configuration, like `default title`, `title prefix`, `title suffix`, and so on, which are used to solve mentioned issue above.

----
Would be nice to have someone "instruct" me on what's wrong in this PR, how we can improve the current approach and give some advices for future PR.

P.S. this is my 1st PR and commitment to PWA

### DIFF
--- a/packages/venia-concept/package.json
+++ b/packages/venia-concept/package.json
@@ -78,6 +78,7 @@
     "react": "~16.6.1",
     "react-apollo": "~2.2.4",
     "react-dom": "~16.6.1",
+    "react-helmet": "^5.2.0",
     "react-redux": "~5.0.7",
     "react-router-dom": "~4.4.0-beta.6",
     "react-test-renderer": "~16.6.1",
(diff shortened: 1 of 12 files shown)


zetlen:
Thank you for chipping in! I see the benefits of rendering with React as much as possible, but I'm not sure this is the time to try it. This change would remove the server-rendered metadata that gives us SEO and SMO.

PR #788, opened 2019-01-28
feat(devex): Refactor Venia to be portable and importable

zetlen:
## Description

**Change the build configuration and dependency graph of Venia and Peregrine so that community developers can use individual Venia components in external apps.**

- Add "partially transpiled" build targets to Venia and Peregrine, producing ES Modules that external build processes can parse, _and_ that web packaging tools can tree-shake.
  - Add and refactor package scripts and Babel configuration
  - Add module resolver plugin to Babel so it transpiles package-relative paths
  - Add export directives to Peregrine package.json and Venia package.json to tell dependency managers where to find different transpilations and/or raw source
- Refactor Venia components to pull their context-driven dependencies from a single aggregate file of such deps, "drivers"
  - Reroute all imports of `<Query />`, `connect()`, `<Link >`, `<Route />`, and others through the central drivers file
  - Consolidate Context Providers into an also-exported `<VeniaAdapter />`
  - Adjust babel config to map the drivers file to a friendly, unique dep name `@magento/venia-drivers`
- Refactor Webpack plugins and other build time configuration to streamline builds and reduce custom code
  - Add doc comments
 

## Motivation and Context
PWA Studio needs to be modular and its components need to be usable and mix-and-matchable at release time. We don't offer "inheritance" in the way of Magento themes, so we have to live up to our promise of modularity and composability. We sacrificed some of that for velocity, and it's time to get that back.

## How Has This Been Tested?
Refactored unit tests and restored coverage.
Tested all changed code.
Ran all NPM scripts.

I built an example app to show how these changes enable import of Venia components á la carte:
[image](https://venia-consuming-app.now.sh)

https://venia-consuming-app.now.sh/

Its source code is at https://github.com/magento-research/venia-consumer-example and a build artifact is inspectable at https://venia-consuming-app.now.sh/_src . Check it out!

feature, documentation

peregrine, pwa-buildpack, venia-concept

## Checklist:


packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -79,7 +80,9 @@ class Product extends Component {
         return {
             minHeight: imageHeight, // min-height instead of height so image will always align with grid bottom
             width: imageWidth,
-            backgroundImage: `url(${makeProductMediaPath(image.file)})`
+            backgroundImage: `url(${new ResourceUrl(
+                makeProductMediaPath(image.file)
+            ).toString()})`
```
jimbo: We could omit `.toString()` in places like this, right? It should be automatically called for expressions in template literal placeholders.

packages/pwa-buildpack/.babelrc.js
```diff
@@ -1,4 +1,6 @@
 const config = {
+    // Ignore __mocks__, __stories__, etc.
+    ignore: [/\/__\w+__\//],
```
jimbo: That helps. 👍

packages/peregrine/src/Router/MagentoRouteHandler.js
```diff
@@ -77,6 +79,13 @@ export default class MagentoRouteHandler extends Component {
             location: { pathname }
         } = this.props;
 
+        // Depending on the environment, the fetchRootComponent global can be
+        // either an ES module with a `default` property or a plain CJS module.
+        const fetchRoot =
+            'default' in fetchRootComponent
+                ? fetchRootComponent.default
+                : fetchRootComponent;
```
jimbo: Consider checking for `__esModule` instead of `default`. (?) I'm pretty sure that's the convention.

packages/peregrine/src/Router/MagentoRouteHandler.js
```diff
@@ -1,7 +1,9 @@
 import React, { Component } from 'react';
 import { func, shape, string } from 'prop-types';
 
-import fetchRootComponent from 'FETCH_ROOT_COMPONENT';
+// 2019-01-28 Removed virtual `FETCH_ROOT_COMPONENT` import. It's much cleaner
+// to inject a "fetchRootComponent" global at build time, so that's what we
+// changed the MagentoRootComponentsPlugin to do.
```
jimbo: That should help. 👍

package.json
```diff
@@ -45,9 +46,11 @@
   "devDependencies": {
     "@magento/eslint-config": "~1.3.0",
     "babel-eslint": "~10.0.1",
+    "babel-plugin-module-resolver": "^3.1.3",
     "bundlesize": "~0.17.1",
     "chalk": "~2.4.2",
     "chokidar": "~2.0.4",
+    "concurrently": "~4.1.0",
```
jimbo: Is this being used anywhere?

package.json
```diff
@@ -45,9 +46,11 @@
   "devDependencies": {
     "@magento/eslint-config": "~1.3.0",
     "babel-eslint": "~10.0.1",
+    "babel-plugin-module-resolver": "^3.1.3",
```
jimbo: Consider `--tilde` here.

packages/peregrine/src/Router/MagentoRouteHandler.js
zetlen: I just Googled a bit and I haven't found anything conclusive about that, but __esModule seems like an internal Babel thing to me.

package.json
zetlen: Yeah, in Venia's `build` script.

package.json
zetlen: Whoa, how'd that get in there?

zetlen: Looks like Yarn doesn't respect our .npmrc, so I added a .yarnrc to set the default save prefix.

zetlen:
@jimbo Thanks for taking a look. What would you suggest instead of `@magento-venia/drivers`?

jochienabuurs:
@zetlen would the same approach be viable for https://github.com/magento-research/pwa-studio/tree/release/2.0/packages/peregrine

zetlen:
@jochienabuurs It would! The only reason I did this for Venia first is that Peregrine is about to go through some major API changes after our release. For now, large Peregrine changes are on hold, and Venia is the more useful and feature-complete of the two.

However, this PR also makes the existing Peregrine components easier to import.

zetlen:
Switched to `@magento/venia-drivers` based on @jimbo 's observation that `@magento/` is our consistent NPM namespace. Good catch! Any other @magento-research/core-pwa-devs have input? I'd like to expedite this

packages/venia-concept/package.json
```diff
@@ -6,23 +6,26 @@
...
-    "prebuild": "yarn run clean",
+    "clean": "rimraf dist esm",
+    "prepublishOnly": "yarn run build",
     "start": "node server.js",
     "start:debug": "node --inspect-brk ./node_modules/.bin/webpack-dev-server --progress --color --env.mode development",
     "stats": "webpack-bundle-analyzer dist/build-stats.json",
     "storybook": "start-storybook -p 9001 -c .storybook",
     "storybook:build": "build-storybook -c .storybook -o storybook-dist",
-    "validate-queries": "node ./validate-queries.js",
+    "validate-queries": "true || node ./validate-queries.js",
```
sirugh: Oh hey...what?

packages/venia-concept/src/components/MiniCart/product.js
jimbo: Naming that method `toString` is the magic part, though. 😉

package.json
jimbo: Looks like it's used in the Peregrine's build script, not Venia's. In either case, the dependency should be listed as a `devDependency` of the package itself, not here in the root.

```sh
yarn workspace @magento/peregrine add --dev --tilde concurrently
```

packages/peregrine/src/Router/MagentoRouteHandler.js
jimbo: My mistake. I thought this was inside a plugin, but it's just inside a component.

jimbo:
@zetlen Thanks for responding to the feedback. Take a look through the comments again. One or two minor changes remain, then we're good.

package.json
zetlen: @jimbo Thanks. You shouldn't need the tilde anymore now that we've got a .yarnrc.

Speaking of which: One downside of Yarn is that it doesn't warn you on the command line if two packages have slightly different semver strings for the same dependency. I think we should strive to not let our dependencies between packages drift apart.

jimbo:
@zetlen This is approved and ready to go.

I'd like to note here that we should probably find a way to do something about the noisy warnings about module resolution. Same for the missing peer dependencies. Let's figure out what our options are.

PR #792, opened 2019-01-30
feature: Possibility to close checkout form

### LINKED ISSUE #608: No way to close "checkout" panel in minicart

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [x] Bug
- [ ] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

## Description

When a user selects "checkout", they are locked into the checkout process and can not close out of the panel.

image

## Expected result:

A user should be able to back out of the checkout/selection process, especially since it covers the bottom half of the minicart.

@soumya-ashok @jimbo Please confirm if this is a bug, or an intended feature. Feel free to close if this is not a bug. :)

### LINKED ISSUE #848: Close Cart View and "Continue Shopping"
A user in the cart view should be able to click a button, besides the `x`, that takes them back to the page they were on (category/product/etc). This could be a button that reads `Continue Shopping` and is located to the left of the "Checkout" button in the cart view.

rogyar:
## Description
This PR introduces the possibility to close checkout and continue shopping. It seems like the approach was designed previously but the reset button was not added to the form. This PR adds the "Continue Shopping" button which allows closing both drawer and checkout pane (there was no possibility to close checkout pane previously)

## Related Issue
Closes #848 

 ## Motivation and Context
This change is required since there's no possibility to close the checkout pane and put the cart back to the previous state (without checkout form opened)

## How Has This Been Tested?
- Add a product to the shopping cart
- Open the shopping cart and click on the checkout button
- You should see "Continue Shopping" button.
- By clicking on "Continue Shopping" the drawer is closed
- If you open the shopping cart once again, the checkout information pane is closed

image


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=792) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=792) before we can accept your contribution.<br/>

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -177,6 +180,7 @@ class Form extends Component {
                     </Section>
                 </div>
                 <div className={classes.footer}>
+                    <ResetButton resetCheckout={resetCheckout} />
```
sirugh: After looking at the [ResetButton](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/Checkout/resetButton.js) component I feel like it's an extra abstraction that we don't need. Can we delete that component entirely and just implement this as  `<Button onClick={resetCheckout}>Continue Shopping</Button>`?

soumya-ashok:
@sirugh When they're in the checkout flow, there needs to be a mask on the shopping cart. @supernova-at is aware of this. Whether the button reads "Cancel" or "Continue Shopping", the shopper will land in the shopping cart since that is where they initiated checkout from. If they want to continue shopping, they will need to exit the shopping cart first. Based on that, I think "Cancel" is a better fit for this context.

brendanfalkowski:
Maybe I'm crazy, but have you ever seen a "continue shopping" or "cancel" button right next to the "place order" button? I sure haven't.

Have you seen a "continue shopping" button anywhere except a Magento site? I have but not really.

This is a pet peeve for me, but I always remove the "continue shopping" button from the cart because:

- Best case — it's an opaque action because it tells users nothing about where it takes them.
- Worst case — it's a dead link if you access the cart directly (i.e. it depends on history to work).

I never understood the argument that shoppers reach their cart and either suddenly become lost or don't understand the "back" button. I can't help thinking this is just a vestigial tail of Magento 1 that won't disappear.

soumya-ashok:
@brendanfalkowski Thanks for the feedback. Here's the original design for the screen - https://magento.invisionapp.com/share/VWP228WG3CB

I agree that "Continue Shopping" next to 'Confirm Order' is an odd pattern in the checkout context, and I'm not sure why it was implemented this way.

brendanfalkowski:
@soumya-ashok — Thanks, I appreciate the dialog.

For this: `distinct modules for "cart" and "checkout" alongside a traditional mini-cart`

This is the model in M1 and M2 — well most sites, Amazon too. In most small screen apps/sites I've seen, they still treat cart and checkout as separate modules (sometimes mini-cart and cart are combined) rather than a layered/sidebar experience. It just feels like forcing the modules into layers adds cognitive load where it's already the norm that cart and checkout are distinct areas (i.e. full page).

For this: `Venia is currently only addressing the smaller retail store context, and we haven't yet delved into how this might translate to the more complex context of B2B.`

I hope Magento rethinks Venia's context and importance beyond a demo. Small retail sites aren't the target Magento client and that's not the average client since M2 was released. I know it's unfair to expect Venia to do everything at once, and you're paring down to build a demo. But practically speaking Venia is the basis for all PWA Studio based projects. You can't escape or ignore this fact even if it's explicitly not the intention.

Venia needs to be designed with consideration for the top-end of the complexity chain — even if you're aren't building that part today — I hope that's being discussed because it'll come back at you fast. Historically Magento has not significantly improved any module's frontend once released because that version-locks sites that made changes. So even now this has an effect on client projects (particularly in B2B) where we say: "hey, from what we're seeing of the early checkout work in PWA, I think we need to plan to build checkout from scratch instead of using what will eventually be provided".

I'm really not suggesting "just use a time machine, and build the right thing the first time". Everyone is doing their best, and it's appreciated. I'd just rather hear that strategically Magento isn't boxed into its own thinking that "Venia is just a small-site demo, it doesn't need to XYZ" because it won't be contained to that.

soumya-ashok:
@brendanfalkowski Okay, I was imagining the current M2 flow on Luma.
I understand your point about cognitive load, and will revisit the pattern with that in mind to see if I can preserve some of the original thought I was working towards while increasing clarity in the flow and areas of cart and checkout.

We are discussing as a team as to what use-case(s) Venia will ultimately serve and extend to. Your feedback is valuable, and we will definitely factor it into the discussion. But this being a product decision, perhaps @awilcoxa could weigh in.

Appreciate the discussion, it really helps me to be able to talk through the context!

sirugh:
As of the 2.0 release we added the "Back to Cart" button which returns a user to the cart view/mask. This changeset includes a "Continue Shopping" button which returns a user to the PDP/Category page. Take a look at the below gifs to see the difference. 

IMO we can close this PR but I want to get a sanity check from @soumya-ashok and/or @jimbo.

## Release 2.0

[Image from Gyazo](https://gyazo.com/11ac8074122bb59c4f9a0d63866635e5)

## This PR

[Image from Gyazo](https://gyazo.com/0632a20ff244c55821cbceff57927eb3)

soumya-ashok:
"Back to Cart" is what we had decided on, let's keep that. We could add "continue shopping" as an action in the mini cart to the left of "Checkout" but that's a separate discussion.

sirugh:
Per discussion with @soumya-ashok the *checkout view* should have a "back to cart" button and the "cart view" should have a "continue shopping" button. Closing this. The latter functionality will be a feature/issue later.

PR #229, opened 2018-08-13
Default pragma

pcvonz:
## This PR is a:

[ ] New feature
[ ] Enhancement/Optimization
[x] Refactor
[ ] Bugfix
[ ] Test for existing code
[ ] Documentation

## Summary

Switch back to the default pragma in our Babel configuration. More information on why this switch was made originally and why we're moving back are documented [here](https://github.com/magento-research/pwa-studio/issues/21)

To sum it up though: We need to add something to our compiler to let it know how to compile JSX. A decision was made to stray away from the standard way. This pull request undoes that and implements the standard way. (Hope that sums it up well)

**NOTE**: This pull request is tied to this another pull request in `magento-eslint`: 
https://github.com/magento-research/magento-eslint/pull/1

## Additional information

Fixes #21

PR #795, opened 2019-01-30
Changed createElement import

mage2pratik:
## Description

Change `createElement` import in each file (ex: `import { createElement } from 'react';` to `import React from 'react';`)

## This PR is a:

- [ ] New feature
- [x] Refactor
- [ ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

#229


supernova-at:
I am a little confused by this change:

It appears within a test called "Operations work when createElement is used instead of React.createElement", which does seem like a valid thing to test for in a function that transforms `createElement` calls: we want to catch all of them.

So I could just update the failing snapshot and this would pass, but I think that would make this particular test useless in that it is not testing what it should.

First, is the above correct?

Second, should we then just delete this test case? If so, do we actually care about testing calls to `createElement` that are achieved some other way than `import React from 'react'`?

jimbo:
> I am a little confused by this change:
> 
> It appears within a test called "Operations work when createElement is used instead of React.createElement", which does seem like a valid thing to test for in a function that transforms `createElement` calls: we want to catch all of them.
> 
> So I could just update the failing snapshot and this would pass, but I think that would make this particular test useless in that it is not testing what its name indicates it should.
> 
> First, is the above correct?

Good observation. You're correct, this is a test case for pragmas other than the default `React.createElement`.

> Second, should we then just delete this test case? If so, do we actually care about testing calls to `createElement` that are achieved some other way than `import React from 'react'`?

No, we shouldn't delete this test case; it's valuable as-is, so we should keep it and close this PR.

PR #797, opened 2019-01-31
feature: Layered Navigation (Filters)

codeAdrian:
## Description

When this PR is merged, it will: 
* Add a Layered navigation (Filters) modal where users can choose, apply and reset filters.
* Filters data is pulled from Products data. 
* GraphQL currently doesn't support filters being applied to queries.
* URL is being updated with applied filters
* Applied filters are initialized with URL data
* Config for each filter group (grid/list mode, search, etc.) is partially hard-coded in constants.js file
* Users are able to search the groups of filters (Discussed with @soumya-ashok in meeting). Currently this is applied only to "Fashion color" filters, but it can be enabled for any group.

Additional notes:
* Tests will be added when general components and structure is approved
* There are some issues with unmounting filter swatch components on change. I haven't been able to trace the issue and fix it. __Need asssistance__

## Related Issue

Closes https://github.com/magento-research/pwa-studio/issues/361

## Motivation and Context

Adding Layered navigation to category page.

## How Has This Been Tested?

Tested with Venia sample data on any product category.

Automatic tests will be added when general components and structure are approved.

screen shot 2019-01-31 at 1 42 02 pm

FEATURE

venia-concept

## Checklist:


codeAdrian:
Reason why build is failing:

> new Octokit({headers}) is deprecated. Use {userAgent, previews} instead. See https://github.com/octokit/rest.js#client-options

sirugh:
@codeAdrian can you resolve the conflicts please :)

codeAdrian:
> @codeAdrian can you resolve the conflicts please :)

Hello @sirugh , I am out of office until Tuesday. I will fix the conflict as soon as I'm back in the office.

sirugh:
@codeAdrian no worries, I went ahead and resolved it. However I'm noticing that the filter is not being applied upon clicking the "Apply" button. Any ideas? We can wait till you're back next Tuesday. Enjoy your vaca!

codeAdrian:
Hello @sirugh , thanks for the feedback.

Regarding the applying of filters, while I was working on this feature, my understanding was that API is not ready for applying filters, so we delayed that final step in layered navigation until GraphQL API has been updated (updated to 2.3.1).

https://github.com/magento/graphql-ce/pull/259

I see that the graphQL has been updated after I submitted the PR. If the backend endpoints are updated to latest develop, I can take a look at finishing the implementation once I get back.

sirugh:
I've also noticed that the menu that contains the back button is missing in the menu that slides in when clicking the top-left hamburger button.

packages/venia-concept/src/components/Main/main.css
```diff
@@ -2,7 +2,6 @@
     background-color: white;
     color: rgb(var(--venia-text));
     position: relative;
-    z-index: 1;
```
sirugh: This causes the main menu to appear incorrectly (without back button). Can we add this back? Was it removed intentionally?

packages/venia-concept/src/RootComponents/Category/categoryContent.js
```diff
@@ -21,13 +45,26 @@ class CategoryContent extends Component {
...
+                        </button>
+                    </div>
                 </h1>
                 <section className={classes.gallery}>
                     <Gallery data={items} title={title} pageSize={pageSize} />
                 </section>
                 <div className={classes.pagination}>
                     <Pagination pageControl={pageControl} />
                 </div>
+                <FilterModal
```
sirugh: I think the reason you are having z-index issues is that this `FilterModal` is a child of a category page and not of the main app like `Navigation` or `MiniCart`.

packages/venia-concept/src/components/Main/main.css
codeAdrian: This was intentional. Up until now, all modals were a child of the main container and the z-index worked as intended. 

But now the modal is part of the content which is nested deep into main container, so we had to remove the z-index on the main wrapper and add them to individual drawers. I have fixed the CSS issues and now the drawers appear correctly.

codeAdrian:
> I've also noticed that the menu that contains the back button is missing in the menu that slides in when clicking the top-left hamburger button.

Fixed with latest update

> Filter is not actually being applied after clicking "Apply". See below I added a filter for price but items > $100 still appear.

> Deleting all filters after applying some results in a weird scenario where you cannot "clear" them. See below I removed a single filter but the URL remains. I then click add it back but click "reset" and the url is reset.

Fixed with latest update. This was an UX issue. It is a bit unclear how the "Apply filters", "Reset Filters" and clearing individual filters from checkboxes or active filters list should interact with URL. Should it update immediately or should user apply the changes by clicking "Apply filters".

> Filter is not actually being applied after clicking "Apply". See below I added a filter for price but items > $100 still appear.

I have checked the GraphQL schema on magento 2.3-develop branch, and as far I can see, there is no way of applying attribute filters (color, size, etc.) on catalog page. Only price is supported at the moment.

https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/CatalogGraphQl/etc/schema.graphqls#L457

No info in the documentation: https://devdocs.magento.com/guides/v2.3/graphql/search-pagination.html

GraphQL playground. I have tried testing it with color filter attribute and It didn't work.

Screen Shot 2019-03-26 at 12 25 40 PM

@sirugh , let me know how if the fixes work as expected and how I should proceed with this feature.

packages/venia-concept/src/components/Main/main.css
sirugh: Thanks for the explanation! I did not know the history of the modals :)

sirugh:
As discussed with @codeAdrian there are a few things to note here. First, is a breaking issue that Adrian will get to tomorrow. Second, the GQL endpoints _always_ return the same filters and counts. We are assuming that means they are statically defined. This may be resolved in [this issue](https://github.com/magento/graphql-ce/issues/258).

This PR is purely UX and functional in terms of the UI -- it is not possible to actually apply the filters because GQL restrictions but we did add the UX and functional work that will make them work once GQL starts accepting the filters. Do we still want to merge this? 

@jimbo @tjwiebell @jimbo @supernova-at thoughts?

codeAdrian:
> As discussed with @codeAdrian there are a few things to note here. First, is a breaking issue that Adrian will get to tomorrow. Second, the GQL endpoints _always_ return the same filters and counts. We are assuming that means they are statically defined. This may be resolved in [this issue](https://github.com/magento/graphql-ce/issues/258).
> 
> This PR is purely UX and functional in terms of the UI -- it is not possible to actually apply the filters because GQL restrictions but we did add the UX and functional work that will make them work once GQL starts accepting the filters. Do we still want to merge this?
> 
> @jimbo @tjwiebell @jimbo @supernova-at thoughts?

Breaking issue with pagination has been fixed in the latest update.

packages/venia-concept/src/components/FilterModal/filterModal.js
```diff
@@ -0,0 +1,112 @@
...
+                title: title,
+                value: values[index]
+            }));
+        }
+
+        return urlFilterParams;
+    };
+
+    componentDidMount = () => {
+        const filterParams = this.getFilterParams(this.props.history.location);
```
sirugh: I think we need to store the filters in a single location such as redux and then use that to create the chosen filters within the modal. I think the store could initialize with the values from the URL but I don't think the component should do this every time.

packages/venia-concept/src/index.css
```diff
@@ -42,6 +42,10 @@ body {
     padding: 0;
 }
 
+.modalIsOpen {
```
sirugh: I couldn't find this used in the code anymore. Can we delete?

packages/venia-concept/src/RootComponents/Category/category.css
```diff
@@ -19,6 +18,24 @@
     height: 100vh;
 }
 
+.headerButtons {
+    display: flex;
+    justify-content: center;
```
sirugh: Can you add some space between the header items?

[Image from Gyazo](https://gyazo.com/bc29f3b7988147a88814d482c5958e95)

packages/venia-concept/src/components/FilterModal/FilterFooter/filterFooter.js
```diff
@@ -0,0 +1,114 @@
...
+    handleApplyFilters = () => {
+        const { history, chosenFilterOptions, closeModalHandler } = this.props;
+        history.push('?' + serialize(chosenFilterOptions));
+        closeModalHandler();
+    };
+
+    getFooterButtons = areOptionsPristine => {
+        const { classes } = this.props;
+
+        const resetButtonClass = areOptionsPristine
```
sirugh: Instead of using `getDerivedStateFromProps` we can simply check if there are any `chosenFilterOptions` when we render since the footer component is "connected" to the catalog store in redux.

packages/venia-concept/src/components/FilterModal/FilterList/filterList.js
```diff
@@ -0,0 +1,107 @@
...
+        return (
+            <List
+                items={items}
+                getItemKey={({ value_string }) => `item-${id}-${value_string}`}
+                render={props => (
+                    <ul className={layoutClass}>{props.children}</ul>
+                )}
+                renderItem={({ item }) => {
+                    const isActive = isFilterSelected(item);
+                    const swatchProps = {
```
sirugh: Can we rename this prop to `filterProps` since not all filter items are swatches?

packages/venia-concept/src/components/FilterModal/filterBlock.css
```diff
@@ -0,0 +1,59 @@
...
+    width: 100%;
+    display: flex;
+    justify-content: space-between;
+    padding-top: 0.75rem;
+    padding-bottom: 0.75rem;
+}
+
+.filterList {
+    display: none;
+    padding-right: 2.75rem;
```
sirugh: IMO it would be better to not apply the padding to the swatch filter, unless of course you are going off a mockup. 

[Image from Gyazo](https://gyazo.com/68cc5552282f6eade7ae0350ada9a96a)

packages/venia-concept/src/components/FilterModal/filterModal.js
```diff
@@ -0,0 +1,112 @@
...
+                        value
+                    });
+                });
+            }
+        }
+    };
+
+    render() {
+        const { classes, isModalOpen, closeModalHandler } = this.props;
+        const modalClass = isModalOpen ? classes.rootOpen : classes.root;
```
sirugh: We don't need to thread this through from the root of the app. Instead, you could connect this component to the redux store and check the drawer like:

```js
mapStateToProps = ({app}) => {
 return {
    isActive: app.drawer === 'filter',
    ..otherProps
  }
})
```

Although now that I look at it, maybe @jimbo has a reason why we should connect the `RootComponents` but not the children of those components... I need to hear from him >.<

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -54,16 +58,18 @@ class Category extends Component {
...
 
+        const queryVariables = {
+            id: Number(id),
+            onServer: false,
+            pageSize: Number(pageSize),
+            currentPage: Number(currentPage),
+            idString: String(id)
+        };
+
+        const isFilterModalOpen = drawer === 'filter';
```
sirugh: This is what I mean by not having to thread the prop down. Infact in the future if some other form of page besides a `category` page wants to apply some filters it will just need to use the `toggleDrawer('filter')` action and not need to pass this prop down.

packages/venia-concept/src/RootComponents/Category/categoryContent.js
```diff
@@ -21,13 +45,26 @@ class CategoryContent extends Component {
                         }}
                     />
                     <div className={classes.categoryTitle}>{categoryTitle}</div>
+                    <div className={classes.headerButtons}>
+                        <button
+                            onClick={openDrawer}
```
sirugh: For the same reason, could we just `mapDispatchToProps` here instead of having to thread down the handler?

packages/venia-concept/src/RootComponents/Category/categoryContent.js
```diff
@@ -21,13 +45,26 @@ class CategoryContent extends Component {
...
+                    </div>
                 </h1>
                 <section className={classes.gallery}>
                     <Gallery data={items} title={title} pageSize={pageSize} />
                 </section>
                 <div className={classes.pagination}>
                     <Pagination pageControl={pageControl} />
                 </div>
+                <FilterModal
+                    closeModalHandler={closeDrawer}
```
sirugh: For the same reason, could we just `mapDispatchToProps` within the `FilterModal` instead of having to thread down the handler?

packages/venia-concept/src/components/FilterModal/FilterList/filterSwatch.js
```diff
@@ -0,0 +1,86 @@
...
+                className={classes.root}
+                value={value_string}
+                data-group={group}
+                title={label}
+                onClick={toggleOption}
+            >
+                {isActive && (
+                    <Fragment>
+                        <span
+                            className={classes.swatchLabel}
```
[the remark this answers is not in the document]
sirugh: FWIW I feel like we could reuse a lot of the swatch component stuff here.

packages/venia-concept/src/components/FilterModal/constants.js
```diff
@@ -0,0 +1,24 @@
...
+    grid: 'grid',
+    list: 'list'
+};
+
+export const filterRenderOptions = {
+    fashion_color: {
+        mode: filterModes.swatch,
+        options: {
+            layout: filterLayouts.grid,
+            generateColor: true,
```
sirugh: Let's remove this from the code. It's mocked anyways. We can just assume we are going to generate colors until they are provided in the response.

packages/venia-concept/src/components/FilterModal/constants.js
```diff
@@ -0,0 +1,24 @@
...
+            searchable: true
```
sirugh: Is this another prop that could be removed when provided by the GQL? Maybe we just do away with it and assume always searchable until then.

packages/venia-concept/src/actions/catalog/asyncActions.js
```diff
@@ -1,4 +1,85 @@
...
+export const filterAdd = ({ group, title, value }, history) =>
+    async function thunk(dispatch, getState) {
+        const {
+            catalog: { chosenFilterOptions }
+        } = getState();
+
+        const oldState = chosenFilterOptions[group] || [];
+        const newState = oldState.concat({ title, value });
+
+        dispatch(actions.filterOption.add({ newState, group }));
```
sirugh: The `filterOption.add` and `filterOption.remove` functions are identical. Would it be OK to simplify into a single `filterOption.update` function that both this `filterAdd` and the below `filterRemove` function use?

packages/venia-concept/src/components/FilterModal/FilterFooter/filterFooter.js
```diff
@@ -0,0 +1,114 @@
...
+
+    resetFilterOptions = () => {
+        const { history, filterClear } = this.props;
+        history.push();
+        filterClear();
+    };
+
+    handleApplyFilters = () => {
+        const { history, chosenFilterOptions, closeModalHandler } = this.props;
+        history.push('?' + serialize(chosenFilterOptions));
```
sirugh: Can we use the `filterAdd` function here?

packages/venia-concept/src/components/FilterModal/filterBlock.css
codeAdrian: Padding is set in design: https://magento.invisionapp.com/share/S2Q41VCR45C#/screens/341558811

packages/venia-concept/src/reducers/catalog.js
```diff
@@ -9,7 +9,9 @@ export const initialState = {
     rootCategoryId: null,
     currentPage: 1,
     pageSize: 6,
-    prevPageTotal: null
+    prevPageTotal: null,
+    chosenFilterOptions: {},
+    appliedFilterOptions: {}
```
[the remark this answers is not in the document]
codeAdrian: You are right, thanks for catching that. It must have slipped through the previous conflicts I have resolved. Code has been updated with "prevPageTotal" references, reducers and tests removed.

packages/venia-concept/src/index.css
codeAdrian: Removed in latest update

packages/venia-concept/src/RootComponents/Category/category.css
codeAdrian: Added space in latest update

packages/venia-concept/src/components/FilterModal/FilterList/filterList.js
codeAdrian: Renamed in the latest update

packages/venia-concept/src/components/FilterModal/constants.js
codeAdrian: Removed "generateColor" references and config in latest update.

packages/venia-concept/src/components/FilterModal/constants.js
codeAdrian: I'm not sure if this is going to be returned by GraphQL out of the box. Searching through filters is not default Magento 2 functionality as far as I'm aware of. According to the design, it's only for color filters: https://magento.invisionapp.com/share/S2Q41VCR45C#/screens/341558811

codeAdrian:
> Alright! This is awesome and really good work. After getting those first few wrinkles out of the way I was able to give this a good look-see. I've left feedback on quite a few things. If you think any of it is not applicable or due to my misunderstanding (which is quite possible) please respond and let me know!
> 
> I would also like @jimbo to take a look at this PR and some of my feedback if possible. Some of the changes I requested may not fall in line with decided upon patterns that I may just not be aware of.
> 
> Cheers!

Thanks @sirugh for the detailed code review. I will go through the requested changes and resolve them one by one. I have resolved the easier ones first, so if they are okay, we can mark them as "resolved" so we can keep better track of requested changes.

packages/venia-concept/src/components/FilterModal/filterModal.js
codeAdrian: Great suggestion. Resolved in latest update. Let me know if everything works as expected.

packages/venia-concept/src/components/FilterModal/FilterFooter/filterFooter.js
codeAdrian: This code remained from previous contributor. I have resolved it in latest update.

packages/venia-concept/src/actions/catalog/asyncActions.js
codeAdrian: Resolved in latest update.

packages/venia-concept/src/components/FilterModal/FilterFooter/filterFooter.js
codeAdrian: This only updated the URL and closes the modal, it doesn't affect the redux state. I'm thinking on adding "appliedFilters" redux state that will keep track of applied filters which will affect the category page listing. We'll use that function here.

packages/venia-concept/src/reducers/catalog.js
```diff
@@ -9,7 +10,8 @@ export const initialState = {
     rootCategoryId: null,
     currentPage: 1,
     pageSize: 6,
-    prevPageTotal: null
+    chosenFilterOptions: getFilterParams(),
+    appliedFilterOptions: getFilterParams()
```
sirugh: This will get initialized but not updated. Intentional?

packages/venia-concept/src/reducers/catalog.js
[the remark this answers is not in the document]
sirugh: Wait why was `prevPageTotal` removed?

packages/venia-concept/src/reducers/catalog.js
codeAdrian: This will hold filters that are currently being applied on catalog product list. But since the API doesn't yet support it, I can remove it. Chosen filter option keep track of the filters that user edits in the modal and that will be applied to the catalog product list once user clicks "Apply filters".

packages/venia-concept/src/components/FilterModal/filterBlock.css
sirugh: Hmm it just seems...off in that view. Maybe it's better on different widths. Oh well, I'm not too concerned with this, just wanted to bring it up.

packages/venia-concept/src/components/FilterModal/constants.js
sirugh: Oh, I didn't even notice that this was a prop on `fashion_color`. I'm cool with this :)

packages/venia-concept/src/RootComponents/Category/categoryContent.js
```diff
@@ -21,13 +45,26 @@ class CategoryContent extends Component {
                         }}
                     />
                     <div className={classes.categoryTitle}>{categoryTitle}</div>
+                    <div className={classes.headerButtons}>
+                        <button
```
sirugh: We should conditionally render this button based on the `filters` value from data. We already return null within the modal if there are no filters, so we shouldn't even display the button in that case :)

packages/venia-concept/src/RootComponents/Category/categoryContent.js
sirugh: And maybe instead of worrying about returning null within the modal component we can conditionally render here as well?

packages/venia-concept/src/reducers/catalog.js
```diff
@@ -34,14 +36,22 @@ const reducerMap = {
...
+            chosenFilterOptions: {
+                ...state.chosenFilterOptions,
+                [group]: newState
+            }
+        };
+    },
+    [actions.filterOption.clear]: state => {
         return {
             ...state,
-            prevPageTotal: payload
```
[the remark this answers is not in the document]
codeAdrian: Sorry, my mistake. I will revert this commit. I didn't quite get which line was referenced. The add/remove diff confused me and I thought you were referring to that diff.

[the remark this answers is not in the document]
sirugh: Ah I was referring to the `appliedFilterOptions` line because I couldn't see it used anywhere.

packages/venia-concept/src/components/FilterModal/filterModal.css
```diff
@@ -0,0 +1,57 @@
...
+    position: fixed;
+    left: 0;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    background-color: white;
+    max-width: 640px;
+    transform: translate3d(-100%, 0, 0);
+    transition-duration: 192ms;
+    transition-timing-function: cubic-bezier(0.4, 0, 1, 1);
```
jimbo: I think there's a CSS var for this.

packages/venia-concept/src/components/FilterModal/filterBlock.js
```diff
@@ -0,0 +1,125 @@
...
+    getLayout = options => {
+        const { layout } = options ? options : {};
+        const { classes } = this.props;
+        switch (layout) {
+            case filterLayouts.grid:
+                return classes.layoutGrid;
+            default:
+                return classes.layout;
+        }
+    };
```
jimbo: Interesting approach. 👍

packages/venia-concept/src/components/FilterModal/filterBlock.css
jimbo: The padding is fine. If I were to implement it, I would probably use the `grid-template-columns` property to make the accordion header and accordion container each have two columns, and just have the container's second column be empty, but it's not super important. Layout is achieved here.

packages/venia-concept/src/components/FilterModal/FilterSearch/filterSearch.js
```diff
@@ -0,0 +1,88 @@
...
+            );
+
+            return (
+                <Fragment>
+                    {isSearchable && (
+                        <Form className={classes.filterSearch}>
+                            <TextInput
+                                placeholder={`Search for a specific ${name}`}
+                                fieldState={{ value: filterQuery }}
+                                onChange={handleFilterSearch}
```
jimbo: Consider using `onValueChange` here, so we don't have to interact with an event object or the DOM.

packages/venia-concept/src/components/FilterModal/FilterSearch/filterSearch.js
```diff
@@ -0,0 +1,88 @@
...
+        };
+
+        handleResetSearch = () => this.setState({ filterQuery: '' });
+
+        getFilteredItems = (items, filterQuery) =>
+            items.filter(
+                item =>
+                    item.label
+                        .toUpperCase()
+                        .indexOf(filterQuery.toUpperCase()) >= 0
```
jimbo: Consider using `.includes()`.

packages/venia-concept/src/components/FilterModal/FilterList/filterSwatch.js
```diff
@@ -0,0 +1,76 @@
...
+                            dangerouslySetInnerHTML={{
+                                __html: label
+                            }}
```
jimbo: I don't think we want to do this.

packages/venia-concept/src/components/FilterModal/FilterList/filterDefault.js
```diff
@@ -0,0 +1,55 @@
...
+                title={label}
+                onClick={toggleOption}
+            >
+                <span className={iconClassName}>
+                    {isActive && <Icon src={Checkmark} size={14} />}
+                </span>
+                <span
+                    dangerouslySetInnerHTML={{
+                        __html: label
+                    }}
```
jimbo: Why are we dangerously setting inner HTML here?

packages/venia-concept/src/actions/catalog/asyncActions.js
```diff
@@ -1,4 +1,85 @@
...
+
+        dispatch(actions.filterOption.update({ newState, group }));
+
+        if (history) {
+            const filters = { ...chosenFilterOptions, [group]: newState };
+            updateCatalogUrl(filters, history);
+        }
+    };
+
+export const filterRemove = ({ group, title, value }, history) =>
```
jimbo: Minor quibble: these names stick out to me. Why not `addFilter` and `removeFilter`, which fits the usual pattern of `verbObject`?

packages/venia-concept/src/actions/catalog/asyncActions.js
```diff
@@ -1,4 +1,85 @@
 import actions from './actions';
+import mockData from './mockData';
+
+export const serialize = (params, keys = [], isArray = false) => {
```
jimbo: I'll come back and review this in more depth later.

packages/venia-concept/src/reducers/catalog.js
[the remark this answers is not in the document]
codeAdrian: Commit has been reverted

packages/venia-concept/src/reducers/catalog.js
codeAdrian: I will remove this for now until we have a proper API support

packages/venia-concept/src/components/FilterModal/FilterList/filterDefault.js
codeAdrian: Some filters (like price) return HTML as their label value.

Screen Shot 2019-03-29 at 8 31 00 AM

Feel free to let me know if there is a better approach that we should use (like stripping HTML)

packages/venia-concept/src/RootComponents/Category/category.js
codeAdrian: Fixed in latest commit

packages/venia-concept/src/RootComponents/Category/categoryContent.js
codeAdrian: Fixed in latest commit.

packages/venia-concept/src/RootComponents/Category/categoryContent.js
codeAdrian: Fixed in latest commit.

packages/venia-concept/src/RootComponents/Category/categoryContent.js
codeAdrian: Fixed in latest commit.

packages/venia-concept/src/RootComponents/Category/categoryContent.js
codeAdrian: Fixed in latest commit

packages/venia-concept/src/components/FilterModal/filterModal.js
codeAdrian: Fixed in latest commit

packages/venia-concept/src/util/getFilterParamsFromUrl.js
```diff
@@ -0,0 +1,22 @@
+export const getFilterParams = () => {
+    const params = new URLSearchParams(window.location.search);
+    let titles,
+        values = [];
+
+    let urlFilterParams = {};
+
+    for (var key of params.keys()) {
```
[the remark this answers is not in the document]
codeAdrian: Linter didn't warn me about this, but thanks for catching this. I have fixed this in latest commit.

packages/venia-concept/src/actions/catalog/asyncActions.js
codeAdrian: Makes sense to me. Fixed in latest update.

packages/venia-concept/src/components/FilterModal/FilterList/filterSwatch.js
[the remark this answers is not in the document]
codeAdrian: Replaced with swatch component in latest update.

packages/venia-concept/src/components/FilterModal/FilterList/filterSwatch.js
codeAdrian: This component has been replaced with global swatch component in latest commit. This is not an issue anymore for Swatch component.

packages/venia-concept/src/components/FilterModal/FilterSearch/filterSearch.js
codeAdrian: `includes()` is case sensitive, this approach is case insensitive. Let me know if I should go with `includes`anyway.

packages/venia-concept/src/components/FilterModal/FilterSearch/filterSearch.js
codeAdrian: Fixed in latest commit.

codeAdrian:
> Fails
> 🚫	
> The following unit tests did _not_ pass 😔. All tests must pass before this PR can be merged
> 
> packages/venia-concept/src/RootComponents/Category/__tests__/categoryContent.spec.js
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [6922ad0](https://github.com/magento-research/pwa-studio/commit/6922ad016a50cea67277d9885ac8ec363c4ac680)

@sirugh @jimbo I might need help with fixing this test. It's a bit too complex for me.

It was caused by implementing this improvement: https://github.com/magento-research/pwa-studio/pull/797/files/468509a54281db1f1bcee18034e0c5c0d8f31a04#r269654614

packages/venia-concept/src/components/FilterModal/filterModal.css
codeAdrian: Fixed in latest update

packages/venia-concept/src/components/FilterModal/FilterList/filterDefault.js
sirugh: That's... an odd choice. @jimbo is that just an issue with our sample data or is there some practice of defining classes/html as copy at Magento?

packages/venia-concept/src/util/getFilterParamsFromUrl.js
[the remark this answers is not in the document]
sirugh: Huh. Looks like we use [this config](https://github.com/magento-research/magento-eslint) for our linting. I'll open an issue regarding this :)

sirugh:
> I might need help with fixing this test. 

@codeAdrian you may want to refer to other connected components to see how we test them. Another pattern I've used before is exporting the un-wrapped component as well and using that in the tests. The wrapping is just a way to provide props so if you go with that pattern you would just import that un-wrapped component within the test file and manually provide the props that would otherwise be passed by the wrapper.

We're basically doing that already, its just that all of our connected components are wrapped by a "Container" component where we do the connection. Check out the `Receipt` and `ReceiptContainer` components for an example of what I'm talking about.

codeAdrian:
> > I might need help with fixing this test.
> 
> @codeAdrian you may want to refer to other connected components to see how we test them. Another pattern I've used before is exporting the un-wrapped component as well and using that in the tests. The wrapping is just a way to provide props so if you go with that pattern you would just import that un-wrapped component within the test file and manually provide the props that would otherwise be passed by the wrapper.
> 
> We're basically doing that already, its just that all of our connected components are wrapped by a "Container" component where we do the connection. Check out the `Receipt` and `ReceiptContainer` components for an example of what I'm talking about.

I might go and create a wrapper around `categoryContent` component and only test the component. Great suggestion, thank you.

codeAdrian:
Test has been fixed as suggested by @sirugh . All tests now pass.

@jimbo let me know if I need to modify anything else.

codeAdrian:
Hello everyone, I have fixed the eslint issues. All tests now pass. @jimbo Any news on the reviews?

Just to let you know, I'll be out of office next week and will be unavailable.

jimbo:
@codeAdrian Thanks for updating the PR. I've finally got some time to take another look at this one.

packages/venia-concept/src/components/FilterModal/FilterList/filterDefault.js
codeAdrian: Fixed in latest commit:
- Added stripHtml function
- Removed Dangerously setting HTML functions

GraphQL only returns the HTML version of the label. There is no raw string label option implemented. So this needs to be fixed with sample data or GraphQL team.

packages/venia-concept/src/components/FilterModal/FilterSearch/filterSearch.js
codeAdrian: Thanks for the clarification. I have added `includes` instead of `indexOf` in the latest update.

packages/venia-concept/src/components/FilterModal/FilterList/filterDefault.js
jimbo: @codeAdrian Great, thank you. I'll make sure this gets brought up to the GraphQL team.

codeAdrian:
> @codeAdrian Thanks for being patient and addressing feedback with changes. This is a good PR and we've failed to do our part to get it merged.
> 
> I do have a couple more requests, and one of them may be a bit unreasonable.
> 
> **New modals and stacking order**
> 
> I'm still concerned about adjusting all the stacking orders (`z-index`) to accommodate our first deeply nested modal. The other modals—`Navigation` and `MiniCart`—were placed as siblings to the `<main>` element, and it would be nice if we could do that with `FilterModal` as well. I think I understand why we aren't doing that yet: these filters are the category page's local concern rather than the site's global concern like the others. Still, I think it should be possible.
> 
> Can I ask you to experiment with [`createPortal`](https://reactjs.org/docs/portals.html) here? When the category mounts, try creating an element as a sibling of the other modals, capturing a `ref` to it, and rendering the FilterModal into that element. Maybe create a new branch from thsi one; if you're stuck or unhappy with how it turns out, you don't have to push anything to this branch and overwrite your work here, and we can discuss how to move forward based on how you feel about the experiment. I'm just curious because this seems like the appropriate use case for portals.
> 
> **Design**
> 
> Did @soumya-ashok provide mockups for the filter modal? I'd like to see the design more closely match the others, with the same light gray header at the top and such. Also, I'd be interested in having this modal enter from the bottom edge rather than the left edge, which would make it centered at desktop resolutions. If existing designs contradict what I'm proposing, let me know and I'll discuss with Soumya.

Hi Jimmy, thanks for the feedback. I will look into using portals. I am familiar with them but never had the chance to use them. I will look into it and let you know when I made the improvements.

Regarding the design, I am following @soumya-ashok design posted here: https://github.com/magento-research/pwa-studio/issues/361#issuecomment-456539901

Let me know if you have any more suggestions.

Regards,
Adrian

soumya-ashok:
@jimbo and @codeAdrian 

The filters would be requested by the shopper from a category or search results page - https://magento.invisionapp.com/share/8ZPPEYD7BGS
Having them come up from the bottom was my intention as well.

The reason they do not have the gray top navigation bar is that they are not a global element in the same way that navigation or cart is in my mind. They are contextual to a couple of instances within the shopping flow which is why both filter and sort are visually different from the main nav. or cart.

All mockups for filters - https://magento.invisionapp.com/share/F2RRNA8WMBT

codeAdrian:
Hello everyone, turns out that the Portal implementation was really easy. I'm really happy how it turned out. PR has been updated with:
* Added Portal for FilterModal
* Reverted changes to z-index values

Let me know if I need to fix anything else.

packages/venia-concept/src/components/Modal/modal.js
```diff
@@ -0,0 +1,6 @@
+import { createPortal } from 'react-dom';
+
+const Modal = ({ children }) =>
+    createPortal(children, document.getElementById('root'));
```
sirugh: Do we care about keyboard focus for this iteration? If so we will need to do more work here.

See [here](https://reactjs.org/docs/portals.html#usage) where they discuss that it is important to manage keyboard focus with portals. And [here](https://reactjs.org/docs/accessibility.html#programmatically-managing-focus) where they have suggestions on how to proceed.

Aside from that, this is cool! Should we apply this to the other modals such as nav or the cart? cc @jimbo

codeAdrian: Hello @sirugh , thank your for your feedback. I haven't seen the keyboard focus improvements being done on other drawers.

However, I have implemented the keyboard focus on the filter modal. But I'm not really happy with the implementation. For some reason, focus wouldn't be switched to modal if the `focus()` method is not in `setTimeout` function, even with the `preventDefault` function.

You can see the implementation here: `src/RootComponents/Category/categoryContent.js`.

Do you have any suggestions on how to improve this?

jimbo: This is so simple—I love it. Two thoughts come to mind here:

- Rather than hardcoding the target to `#root`, we might consider accepting an element as a prop, with `#root` being the default
- I wonder if we can memoize the `document.getElementById()` call with `useMemo()`

packages/venia-concept/src/components/FilterModal/filterModal.js
```diff
@@ -0,0 +1,96 @@
...
+    closeDrawer: () => dispatch(closeDrawer())
+});
+
+export default compose(
+    connect(
+        mapStateToProps,
+        mapDispatchToProps
+    ),
+    classify(defaultClasses)
+)(FilterModal);
```
jimbo: Since we're calling `connect()` from `FilterModalContainer`, why are we also calling it here?

packages/venia-concept/src/components/Modal/modal.js
codeAdrian: Great suggestions. Element id prop and memoization has been added in the latest update.

packages/venia-concept/src/components/FilterModal/filterModal.js
codeAdrian: Good catch. I have moved the redux props to container in latest update.

packages/venia-concept/src/components/Modal/modal.js
jimbo: I had something *slightly* different in mind for the element prop—I wanted to cover a case in which you have an element stored as a ref and don't need to query the document—so I've pushed a small change. Overall this is just what I had in mind though. 👍

dpatil-magento:
@jimbo @codeAdrian @sirugh    Let me know your views on below observations - If anything major - fix , minor - I will create separate issue , not needed - then we can just ignore those.

1. Select any filter and unselect - Reset Filters and Apply Filters button does not get disabled.
2. Fashion Color - If I type a color - 
      - Don't see the text in search field,
      - Filtered colors are not correct. Ex Type - Rain,   Rain and Mint 2 options are displayed.
3. Keyboard tab order focus and the way focus should move needs to be implemented.
4. If I select so many filters which does not fit in window size, then there is no way user can see the hidden filters on desktop browser. On mobile user can see by swiping left. Is it good idea to wrap in next line?
5. Say I'm on Page 4 of a category and select a filter and then click Reset filter and close filter overlay. I'm back to Page 1.
6. Filter button can move some more up, just to make UI look good.  something like below ?

image

codeAdrian:
> @jimbo @codeAdrian @sirugh Let me know your views on below observations - If anything major - fix , minor - I will create separate issue , not needed - then we can just ignore those.
> 
> 1. Select any filter and unselect - Reset Filters and Apply Filters button does not get disabled.
> 2. Fashion Color - If I type a color -
>    
>    * Don't see the text in search field,
>    * Filtered colors are not correct. Ex Type - Rain,   Rain and Mint 2 options are displayed.
> 3. Keyboard tab order focus and the way focus should move needs to be implemented.
> 4. If I select so many filters which does not fit in window size, then there is no way user can see the hidden filters on desktop browser. On mobile user can see by swiping left. Is it good idea to wrap in next line?
> 5. Say I'm on Page 4 of a category and select a filter and then click Reset filter and close filter overlay. I'm back to Page 1.
> 6. Filter button can move some more up, just to make UI look good.  something like below ?

Hi @dpatil-magento , thank you for your feedback. Let me address each point separately.

1. Thanks for catching that. This issue has been fixed.
2. Color search issue has been fixed.
3. Jimmy has commented on this topic here https://github.com/magento-research/pwa-studio/pull/797#pullrequestreview-234262932. I suggest creating a new issue for this because it also includes navigation, minicart, checkout and account modals.
4. I have enabled a scrollbar to render as browser (or OS) default. Meaning that desktop will render it as a standard scrollbar, and mobile will render it in their own styled scrollbar (hidden or transparent). This ensures equal UX across all devices.
5. I am assuming if the user changes filters, the catalog item listing will change and user would like to start browsing from the start so they can view items they might have missed. Let me know if there is a UX reason that we might want to keep the current page when changing filters.
6. Styles have been adjusted in the latest update.

soumya-ashok:
@codeAdrian and @dpatil-magento - I need a little more context on 5. since I am unable to view the deployment.

5. If a shopper is browsing a category, and at some point decides to apply a filter, it makes sense to me that they are back at "Page 1" because they are in a new view within the category now. I think this is what you're saying as well Adrian?

4. I am not sure how much space was around the filter button but my intent was to actually provide a touch area that would be comfortable for someone with larger hands. Keeping active elements too close together can lead to accidental clicks. That being said we should not be adjusting UI padding / spacing one-off. I would like @jimbo to weigh in on this as well.

dpatil-magento:
@codeAdrian   I discussed with @soumya-ashok and 5 also needs be fixed. If the filters are engaged on page 3, but no filters are actually applied, when the filter screen is closed, the shopper would still be on the same page they engaged the filters from.

@jimbo @codeAdrian  Per discussion with @soumya-ashok - here are some more changes needed. Let Me know if we need to address as part of this ticket or create new ticket?
A. Filter screen is designed to be a full screen overlay
B. Category filter should not be displayed when user is already on Category page. It make sense to display Category when user is search results page.
C. The filter pill text size could be knocked down a point
D. Filters should be category aware. i.e  If I select a category then clothing styles filter option should be updated.
E.  Selected filter pills (header area on filter screen) be sticky so they can stay in the shopper's view?

jimbo:
1. Good suggestion. Conditionally disabling modal actions is the right idea. Good fix.
2. Good suggestion. Color search works now, but throws when clearing the field.
3. Good suggestion, but not now. Keyboard support should be addressed as a new issue.
4. It's fine to let the filter tag container scroll horizontally. No fix necessary.
5. Fair suggestion. Pagination should only change when the *applied filters* change, since that's when the data changes. If it overcomplicates the code, though, let me know and we can waive this requirement.
6. @dpatil-magento Let's reserve the minor style feedback for Soumya. 😅 It looks fine as-is, and we certainly don't want *less* spacing. I'll work with @soumya-ashok on how to adjust spacing and sizing for this page, but it won't block this PR.

jimbo:
A. The bottom sheet approach used here is pretty solid. I wouldn't make any changes to how the filter modal is presented here.
B. I don't have an opinion on this.
C. This is another minor style thing, but I disagree. We don't use `15px` text anywhere in Venia (other than text fields and only for Safari compatibility reasons), and `16px` is our most-preferred size. If anything, the tags would grow larger in order to be valid touch areas, but I think they're acceptable as-is.
D. Yes, it'd be nice if filters were category-aware, but I don't think we have that ability right now. I assume we simply get a flat list of potential filters and filter values from GraphQL? @codeAdrian
E. You mean within the modal? It's fine if we fix the height of the modal and allow the modal body to overflow with a vertical scrollbar, but I'd like to avoid any new sticky toolbars.

codeAdrian:
Hello everyone, this PR has been updated with the following fixes:
* Fixed the "Clear filter search" issue. This now works as intended.
* Fixed the page reset issue on applying and removing filters. The easiest way of fixing this was just to apply this fix for all cases.

To answer the comments:
* Regarding the category GraphQL query. There is an GraphQL issue which returns the same filters for any category or search query. So I would recommend to wait for GraphQL fix before making any adjustments to filters groups.
* Regarding the modal, it looks better when it covers a part of the bottom section of the screen. Also the transition when opening and closing looks much better when it doesn't cover the entire screen.
* @soumya-ashok can you provide me with the list of style adjustments I need to make, taking into account @jimbo and @dpatil-magento suggestions?

dpatil-magento:
@codeAdrian   Fixes on Category page looks good.  Some observations on **Search Result page** -

7. "Fashion Size" and "Price" filters are missing.
8. In Filter overlay I select a filter and click on Reset Filters button -  
      a- Filter Overlay closes with shadowed window instead of just clearing out the filters
      b- User is taken back to Home page instead of keeping them on Search page with results.
9. In Filter overlay I select a filter and click on Apply Filters button then user taken back to Home page

soumya-ashok:
@jimbo Per the design, the filter pill text size is actually smaller than body copy. We can keep the touch area the same but I think the text is too large relative to the size of the pill. Personal opinion.

On sticking the top of the filter modal - I think the visual treatment on the checkbox changing when selected is enough reinforcement for the shopper to know that their action has succeeded so we can scroll this area without too much negative impact to the user. I was originally thinking that if they had multiple filters applied it might be good to keep that within their view but they could just scroll to the top to see it if they really wanted to.

@codeAdrian 
With having the filters be full screen here are my thoughts - We have multiple accordions which collapse and expand within the area that already taken up vertical space and we cannot predict how many filters there will be within each accordion since this depends entirely on how the store is set up. Why would we not use the full height of the device and allow the user to see more filters at a time? I do not see an advantage in having this be a modal of restricted height.

codeAdrian:
> "Fashion Size" and "Price" filters are missing.
@dpatil-magento This is a Backend / GraphQL issue. We only show filters that the graphQL query returns.

> I do not see an advantage in having this be a modal of restricted height.
@soumya-ashok Modal height is not restricted. It can expand to the full height once more accordions are open. You can see how that looks right how. I can adjust it to make it full height tomorrow if it's going to look better.

codeAdrian:
Hello everyone, following fixes have been applied in the new update:
* Fixes for search results filtering issues
* Fix for spacing above Filter button

Let me know if you find anything else for me to fix and I'll fix it on Monday. If there is something that can be fixed sooner than that, feel free to push a fix to the PR.

dpatil-magento:
@codeAdrian   Fixes looks good.  One more scenario does not look correct - 
10- Apply a filter. Go back to filter overlay and uncheck the applied filter.   Url still has the filter in it and Apply/Reset Filter button is disabled.   There is no way user can remove the filter now.

Note - The above scenario works fine if I remove filter by clicking the  filter in filter header but again this breaks the flow when say user engage filter overlay for the first time and selects 2 filters and removes one filter via filter header. Check the URL remaining filter is already applied even though user has not clicked Apply filters button.

Is it good idea to make filters to take effects only when Reset Filters and Apply Filters buttons are clicked? 

@soumya-ashok  Please provide your inputs on above.

codeAdrian:
Hello everyone,
PR has been updated with the fix for the following scenario:
> Apply a filter. Go back to filter overlay and uncheck (via check box) the applied filter. Url still has the filter in it and Apply/Reset Filter button is disabled. There is no way user can remove the filter now.

Let me know if anything else needs fixing.

dpatil-magento:
@codeAdrian  Recent fix looks good.  Please check following scenario's :
11:  Apply filter on Search page. Now navigate to category page and go to filter overlay.  The previously applied filters on search page still show up on Category page filter.
12: On Filter overlay select filters and close filter overlay by clicking X (Close) icon or outside overlay area to close filter overlay without applying selected filters. Now Go back to filter overlay, previously selected filters are still in checked state.

codeAdrian:
Hello @dpatil-magento , I have updated the PR with the fixes you have reported. I have went through the all edge cases and scenarios I can think of and I haven't noticed anything out of place. Let me know if you find anything out of place.

dpatil-magento:
Thanks @codeAdrian all fixes looks good.

@jimbo @soumya-ashok  Please let me know if we need to create ticket for below which is pending from GraphQL just for tracking purpose. And also if you have any other suggestions for this ticket. 

D: Filters should be category aware

soumya-ashok:
@codeAdrian - Not sure if you implemented the full height on the modal for the mobile context, but I'm still seeing the old treatment. The modal height varying with the expanding accordions is a bit jarring. We would deal with filters differently on desktop, so I won't get into that here.

@dpatil-magento - I didn't understand your question exactly but unchecking the filters removes the tags within the filter modal. However, if I were to close the modal and reopen it, those old filters I removed are still there.

Applying filters isn't doing anything at the moment. I don't see a results page.

Also, when I select a color, the swatch expands to a rectangle. This should not be happening. Screenshot below.

Screen Shot 2019-05-16 at 2 44 52 PM

codeAdrian:
Hello @soumya-ashok thank you for feedback.

Regarding the filters not applying after changing, I'm not sure if the attribute filters input has been added to GraphQL backend. @jimbo , @sirugh feel free to let me know if I'm in the wrong. I've been checking out the Catalog GraphQL schema and didn't see the attribute filter input: https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/CatalogGraphQl/etc/schema.graphqls#L279

Regarding the style issues (full height filters modal and swatch width), they have been fixed in the latest update. Let me know if there is anything else that needs to be fixed.

dpatil-magento:
Thanks @codeAdrian fixes looks good.

sirugh:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  node` failed.
  </td>
    </tr>
  </tbody>
</table>



### Log


```sh
    at Object.dangerfilePath (/codebuild/output/src451119981/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/utils/fileUtils.js:26:11)
    at Object.<anonymous> (/codebuild/output/src451119981/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:92:42)
    at step (/codebuild/output/src451119981/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:33:23)
    at Object.next (/codebuild/output/src451119981/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:14:53)
    at fulfilled (/codebuild/output/src451119981/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:5:58)

```

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0e122d302bf0cdfc01db36461ece854019a87e93
</p>

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=797) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=797) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: Starotitorov<br/>:x: Adrian Bece<hr/>**Adrian Bece** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

jimbo:
@codeAdrian I've resolved the conflicts that were created when we merged #1211, which converted `Category` to a functional component using hooks. Take a look and see if you have any feedback; otherwise we'll try to merge this Monday.

codeAdrian:
@jimbo Thanks for resolving the conflicts. I really appreciate it. I have fixed a minor issue and cleaned up a commented-out piece of code that was leftover from the merge. I am happy to proceed with the merge on Monday. Cheers!

dpatil-magento:
Did another round of QA, all looks good. @jimbo - Good to merge.

PR #799, opened 2019-01-31
Checkout Form Updates

### LINKED ISSUE #695: UAT: Checkout Drawer Incorrect Styling

supernova-at:
## Description

At the heart of this PR is the ability for users to fill out the checkout sections (shipping address, payment method / billing address, and shipping method) in any order, and to cancel out of the checkout flow, and providing a default shipping method.

There are also some CSS changes that include the addition of edit icons for filled in sections, and proper button and text colors.

Lastly, this PR fixes a bug I found along the way where we were submitting an incorrect billing address if the user selected "Billing Address same as Shipping Address" but then going back and editing their shipping address.

## Related Issue

Closes #695.

## Motivation and Context

@soumya-ashok ran some user acceptance testing and came up with a list of things to fix related to checkout in #695. Each of the things she identified is fixed by this PR.

## How Has This Been Tested?

You'll want to run through checkout a bunch of different ways and ensure each path leads to a successful submission. For example:

1. Shipping address > Payment method (keep billing same as shipping) > Submit
1. Shipping address > Payment method (enter different billing address) > Submit
1. Payment method (keep billing same as shipping) > Shipping address > Submit
1. Payment method (enter different billing address) > Shipping address) > Submit

screen shot 2019-01-31 at 12 24 09 pm

Things to notice:
  1. The shipping method has a default
  1. Users can fill out the sections in any order
  1. There is a new "Cancel" button
  1. The prompt text color is updated
  1. Button says "Confirm Order" instead of "Place Order"
  1. "Confirm Order" button is gray when disabled

screen shot 2019-01-31 at 12 24 30 pm

Things to notice:
  1. Once a section is filled out, an edit icon appears
  1. "Confirm Order" button becomes enabled when all sections have been filled in
  1. "Confirm Order" button is black (standard button color) when enabled

BUG, FEATURE

This PR is 100% in `venia-concept`.

## Checklist:


packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,22 +1,22 @@
...
 import Section from './section';
 import ShippingForm from './shippingForm';
 import SubmitButton from './submitButton';
 
 import classify from 'src/classify';
+import Button from 'src/components/Button';
 import defaultClasses from './form.css';
 
-const { BrowserPersistence } = Util;
-const storage = new BrowserPersistence();
```
zetlen: It's always great to see one of these going away. 🙌

packages/venia-concept/src/components/Checkout/section.js
```diff
@@ -1,33 +1,40 @@
...
 import { bool, node, shape, string } from 'prop-types';
-import CheckIcon from 'react-feather/dist/icons/check';
+import EditIcon from 'react-feather/dist/icons/edit-2';
 
 import classify from 'src/classify';
 import Icon from 'src/components/Icon';
 import defaultClasses from './section.css';
 
+const editIconAttrs = {
+    color: 'black',
```
zetlen: Can you add a TODO for moving these attributes into a stylesheet? We want to colocate interdependent things like pixel width, but we don't have an agreed-upon means of doing it yet. (I favor using a CSS Modules [:export block](https://github.com/css-modules/icss#export), but any choice will need @jimbo to weigh in, and can come after release.

packages/venia-concept/src/components/Checkout/wrapper.js
```diff
@@ -16,53 +15,51 @@ import {
...
         beginCheckout: func.isRequired,
+        cancelCheckout: func.isRequired,
         cart: shape({
             details: object,
             guestCartId: string,
             shippingMethods: array,
             totals: object
         }),
         checkout: shape({
+            billingAddress: object, // TODO: shape
```
sirugh: Resolve this and other TODOs? Or if we need to wait for something else can we open an issue to resolve?

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -134,14 +158,21 @@ export const submitShippingMethod = payload =>
...
-        const desiredShippingMethod = payload.formValues.shippingMethod;
+        const paymentMethod = await retrievePaymentMethod();
         const shipping_address = await retrieveShippingAddress();
+        const shipping_method = await retrieveShippingMethod();
 
-        // Remove extranneous properties from the billing address because the
-        // Magento backend will reject them.
-        billing_address = removeInvalidKeysFromAddress(billing_address);
+        if (billing_address.sameAsShippingAddress) {
+            billing_address = shipping_address;
```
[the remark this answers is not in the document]
sirugh: This is the action associated with submitting the order, so I believe it occurs on the final click. Any changes upstream such as if the user goes back and changes from "same as shipping addr" to a manually entered billing address should update the prop and then this condition would be false.

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -134,14 +158,21 @@ export const submitShippingMethod = payload =>
...
+        } else {
+            billing_address = {
+                ...shipping_address,
```
sirugh: Why do we merge the address objects here?

packages/venia-concept/src/components/Checkout/wrapper.js
```diff
@@ -92,33 +88,29 @@ class CheckoutWrapper extends Component {
...
             submitShippingMethod
         };
 
         const { shippingMethods: availableShippingMethods } = cart;
-        const { paymentData, shippingMethod, shippingTitle } = checkout;
+        const { paymentData, shippingAddress, shippingMethod } = checkout;
 
         const miscProps = {
             availableShippingMethods,
+            havePaymentMethod: hasData(paymentData),
```
sirugh: I usually like `has` over `have`. Just... you know, incase you wanted to change it.

packages/venia-concept/src/actions/checkout/asyncActions.js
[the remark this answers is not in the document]
supernova-at: > Or, do all user changes to either address flow through this code, and so it's safe to do this?

Correct, and this changeset facilitates that. This now happens *only* when placing the order (we're in `submitOrder` here), so it is too late to change either address.

packages/venia-concept/src/components/Checkout/form.js
supernova-at: Yeah I moved this logic up to the wrapping component `Checkout/wrapper.js`, which is now the only one hooked up to Redux and aware of storage. It passes all that information down to the checkout form via `props`.

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -16,7 +12,21 @@ const { request } = RestApi.Magento2;
 const { BrowserPersistence } = Util;
 const storage = new BrowserPersistence();
 
-const INVALID_ADDRESS_KEYS = ['sameAsShippingAddress'];
+// Use the "flatrate" shipping method as the default.
+export const DEFAULT_SHIPPING_METHOD = {
+    amount: 5,
```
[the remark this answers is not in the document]
supernova-at: Good call, I'll have to think about that.

Ideally our default is simply the first shipping method returned from `estimate-shipping-methods`, but that happened too late in the process to default the checkout form with it, and it didn't make sense anywhere earlier in the process.

packages/venia-concept/src/components/Checkout/wrapper.js
supernova-at: Oops, thanks

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -25,6 +35,11 @@ export const beginCheckout = () =>
         dispatch(getCountries());
     };
 
+export const cancelCheckout = () =>
+    async function thunk(dispatch) {
+        dispatch(actions.reset());
```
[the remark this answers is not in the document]
sirugh: Speaking of, if we do this work here should we close #792?

packages/venia-concept/src/components/Checkout/wrapper.js
supernova-at: I don't feel too strongly about it aside from the fact that the consuming code is doing `if (!havePaymentMethod)`, which - in english - reads nicer than `if (!hasPaymentMethod)` but I'm really splitting hairs here 😄

packages/venia-concept/src/actions/checkout/asyncActions.js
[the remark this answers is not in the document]
supernova-at: Good ping, I may have misunderstood @soumya-ashok 's intent.

packages/venia-concept/src/components/Checkout/wrapper.js
```diff
@@ -16,53 +15,51 @@ import {
...
-        isShippingMethodReady(checkout)
+    const {
+        billingAddress,
+        paymentData,
+        shippingAddress,
+        shippingMethod
+    } = checkout;
+
+    return [billingAddress, paymentData, shippingAddress, shippingMethod].every(
+        hasData
```
jimbo: You can just use `.every(Boolean)` here.

packages/venia-concept/src/components/Checkout/section.js
jimbo: I like the idea of the export block (and the comment). Also agree that we should do it project-wide after the release.

packages/venia-concept/src/components/Checkout/section.js
```diff
@@ -1,33 +1,40 @@
...
+    width: 18
+};
+const EDIT_ICON = <Icon src={EditIcon} attrs={editIconAttrs} />;
```
jimbo: Good use of a static element here.

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -222,26 +202,56 @@ class Form extends Component {
...
+                year: 'numeric'
+            }
+        );
+
         return (
             <Fragment>
-                <strong>{shippingTitle}</strong>
+                <strong>{twoDaysFromNowDisplay}</strong>
+                <br />
+                <span>{shippingTitle}</span>
```
jimbo: I've already discussed with @soumya-ashok why we shouldn't do this.

If the shopper is ordering on a Friday, is this telling them it'll be there by Sunday? Should this function be aware of holidays? Does it know if the shipper ships on weekends? Of course not.

"Estimated ship date" is not a mathematical calculation; it's a logistical one. Worse, a shopper is likely to interpret it as a guarantee, even if we were to add an explicit disclaimer that it's not—and they'll be upset accordingly.

If we get a ship date back from the API, that's one thing, but this shouldn't be client-side. 😅

packages/venia-concept/src/components/Checkout/wrapper.js
supernova-at: Resolved by https://github.com/magento-research/pwa-studio/pull/799/commits/fb3ab5b146b877d59168c9d108a9adb38b48b982.

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -134,14 +158,21 @@ export const submitShippingMethod = payload =>
...
-        // Remove extranneous properties from the billing address because the
-        // Magento backend will reject them.
-        billing_address = removeInvalidKeysFromAddress(billing_address);
+        if (billing_address.sameAsShippingAddress) {
+            billing_address = shipping_address;
+        } else {
+            billing_address = {
+                ...shipping_address,
+                ...billing_address
+            };
```
jimbo: This seems like it could have unintended effects. Is it possible for `shipping_address` to have property values that `billing_address` does not?

```es6
const a = { first: "Austin", middle: "Danger", last: "Powers" }
const b = { first: "Vanessa", last: "Kensington" }
const c = { ...a, ...b }

log(c) // { first: "Vanessa", middle: "Danger", last: "Kensington" }
```

packages/venia-concept/src/components/Checkout/wrapper.js
jimbo: I've never seen `haveData` before. @sirugh is right, IMO.

packages/venia-concept/src/components/Checkout/form.js
supernova-at: Good points, I was unaware!

packages/venia-concept/src/actions/checkout/asyncActions.js
supernova-at: Yes, `shipping_address` definitely does have properties that `billing_address` does not (`firstname`, `lastname`, and `email`, specifically). I'm basically piggybacking those values from `shipping_address` here because the M2 will reject the `billing_address` if it doesn't contain them.

packages/venia-concept/src/components/Checkout/form.js
soumya-ashok: @supernova-at I'm sorry I missed the preset ship date - I was too caught up with all the other details. Thanks for your work on this!

packages/venia-concept/src/actions/checkout/asyncActions.js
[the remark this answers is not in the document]
soumya-ashok: @sirugh and @supernova-at This comes back to the point I was making and have made multiple times now. The cart is supposed to have a mask on it once the user enters checkout as far as the current design intent is concerned. The cancel is intended to take them back to the shopping cart. If we're not going to be able to make the cart unavailable when they enter checkout, we need to discuss an alternate flow.

packages/venia-concept/src/actions/checkout/asyncActions.js
sirugh: I think it'd be fine to just explicitly copy those props into the object here rather than spreading the entire shipping address in that case.

packages/venia-concept/src/components/Checkout/wrapper.js
supernova-at: I do like the simplicity of `.every(Boolean)` but `.every(hasData)` makes it explicit what we're testing for and avoids some mental overhead for the reader.

I'm fine with either, really, but unless you feel strongly I'd like to keep this as is.

packages/venia-concept/src/components/Checkout/form.js
supernova-at: I will update this to match the mock:

screen shot 2019-02-01 at 11 06 56 am

packages/venia-concept/src/actions/checkout/asyncActions.js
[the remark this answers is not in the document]
supernova-at: I was actually able to remove this entirely now that we're not defaulting the field 🎉 

https://github.com/magento-research/pwa-studio/pull/799/commits/7244054a1cbbf369e19126a3e22b77fa0800d24c

packages/venia-concept/src/reducers/checkout.js
```diff
@@ -24,8 +27,28 @@ const initialState = {
...
+        const storedPaymentMethod = storage.getItem('paymentMethod');
+        const storedShippingAddress = storage.getItem('shipping_address');
+        let storedShippingMethod = storage.getItem('shippingMethod');
+
+        // If we didn't have a stored shipping method already,
+        // use the default one and then save the default one to storage for next time.
+        if (!storedShippingMethod) {
+            storedShippingMethod = DEFAULT_SHIPPING_METHOD;
+            saveShippingMethod(DEFAULT_SHIPPING_METHOD);
+        }
```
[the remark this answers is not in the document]
supernova-at: I was able to remove this entirely in https://github.com/magento-research/pwa-studio/pull/799/commits/7244054a1cbbf369e19126a3e22b77fa0800d24c now that we're not defaulting the shipping method 👍

packages/venia-concept/src/components/Checkout/form.js
supernova-at: Resolved in https://github.com/magento-research/pwa-studio/pull/799/commits/7244054a1cbbf369e19126a3e22b77fa0800d24c.

packages/venia-concept/src/actions/checkout/asyncActions.js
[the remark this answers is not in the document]
supernova-at: Sounds like a plan 👍 .

I updated the button text to "Back to Cart" in https://github.com/magento-research/pwa-studio/pull/799/commits/9d8ddff723e373b8fd18c67307f25348466d207e.

I'll move on to #717 immediately after this.

packages/venia-concept/src/actions/checkout/asyncActions.js
supernova-at: Resolved in https://github.com/magento-research/pwa-studio/pull/799/commits/6b34063e9a9b2d4e48eacd4c544f6dc1770e9d9e.

packages/venia-concept/src/components/Checkout/wrapper.js
supernova-at: Resolved in https://github.com/magento-research/pwa-studio/pull/799/commits/bc395d699d381b578cd3d88bbc8b6ea35e728fca.

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -1,11 +1,7 @@
 import { RestApi, Util } from '@magento/peregrine';
 
 import { closeDrawer } from 'src/actions/app';
-import {
-    clearGuestCartId,
-    getCartDetails,
-    getShippingMethods
-} from 'src/actions/cart';
+import { clearGuestCartId, getShippingMethods } from 'src/actions/cart';
```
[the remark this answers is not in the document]
supernova-at: Resolved in https://github.com/magento-research/pwa-studio/pull/799/commits/af961b8fb613ba9fc8910fe93d492b2f3997fc3b.

packages/venia-concept/src/components/Checkout/shippingForm.js
```diff
@@ -37,11 +37,19 @@ class ShippingForm extends Component {
...
+        let selectableShippingMethods;
+
+        if (availableShippingMethods.length) {
+            selectableShippingMethods = availableShippingMethods.map(
+                ({ code, title }) => ({ label: title, value: code })
+            );
+            initialValue =
+                shippingMethod || availableShippingMethods[0].carrier_code;
+        } else {
+            selectableShippingMethods = [];
```
sirugh: You can default these values above and get rid of this block.

PR #803, opened 2019-02-01
Make image resize compatible with upward-php

tjwiebell:
## Description

upward-php variable `request.url.pathname` doesn't contain query parameters, and the regex pattern for image resizing wasn't matching, causing 404s for all image requests. It's assumed that pathname *SHOULD NOT* contain query params, and the upward config should be updated to reflect this.

## Related Issue

Closes magento-research/upward-php#53

## Motivation and Context

upward-php should be able to serve images with the image resize functionality in place

## How Has This Been Tested?

Built and ran Venia with both upward-js and upward-php implementations, and verified that images are correctly served.

BUG

venia-concept

## Checklist:


packages/venia-concept/venia-upward.yml
```diff
@@ -42,15 +42,15 @@ imageOpt:
           engine: mustache
           provide:
             width: $match.$1
-            path: $match.$2
+            path: request.url.query.url
```
supernova-at: Can we drop a comment directing to https://url.spec.whatwg.org/ here?

packages/venia-concept/venia-upward.yml
```diff
@@ -42,15 +42,15 @@ imageOpt:
...
           template:
             when:
               - matches: env.USE_FASTLY
                 pattern: '[^0]'
                 use:
-                  inline: '/{{path}}?auto=webp&format=pjpg&'
+                  inline: '{{path}}?auto=webp&format=pjpg&'
             default:
-              inline: '/{{path}}'
+              inline: '{{path}}'
```
supernova-at: What is the affect of these changes? Does something somewhere else need to change to reflect them? I'm assuming this worked at one point 😃

supernova-at:
This could use a look from @zetlen .

packages/venia-concept/venia-upward.yml
zetlen: Yeah, since we don't have a Fastly-enabled cloud instance to test on (@bbatsche you mentioned you have access to one...) we haven't tested this.

PR #806, opened 2019-02-01
Checkout Mask

### LINKED ISSUE #696: UAT: Mask is not applied to background view in Checkout

supernova-at:
## Description

This PR adds a mask to the background during checkout so that no other elements can be interacted with. The original PR is #717.

## Related Issue

Closes #696 .

## Motivation and Context

It prevents other elements such as the product list from being interacted with during checkout.

## How Has This Been Tested?

Open the MiniCart / checkout and see that there is a mask applied behind it, preventing inadvertent interaction with other elements on the page.

The mask can be dismissed by clicking it or by clicking the "Back to Cart" button.

screen shot 2019-02-01 at 4 00 36 pm

BUG / FEATURE

`venia-concept`

## Checklist:


supernova-at:
I should note that the "thin white line" is a bit odd at first:

screen shot 2019-02-01 at 4 00 36 pm

It appears because of the [box-shadow](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/MiniCart/miniCart.css#L8) on the MiniCart.

It looks even worse without it though 😄:

screen shot 2019-02-01 at 4 00 44 pm

packages/venia-concept/src/components/MiniCart/miniCart.css
```diff
@@ -47,6 +47,12 @@
 .footer {
     box-shadow: 0 -1px rgb(var(--venia-border));
     padding: 1.5rem 0 0.5rem;
+    background-color: white;
+}
+
+.footerMaskOpen {
+    composes: footer;
+    z-index: 3;
 }
```
zetlen:
```suggestion
/* Override Mask style to hide the gray effect when masking the cart during checkout */
.footerMask {
  composes: root_active from '../Mask/mask.css';
  opacity: 0;
}
```

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -231,6 +239,7 @@ class MiniCart extends Component {
                     </Trigger>
                 </div>
                 {miniCartInner}
+                <Mask isActive={isMiniCartMaskOpen} dismiss={cancelCheckout} />
```
zetlen:
```suggestion
                <Mask
                  className={isMiniCartMaskOpen ? defaultClasses.footerMask : undefined}
                  isActive={isMiniCartMaskOpen} dismiss={cancelCheckout} />
```

supernova-at:
> I think it's probably good enough to have an invisible mask that fires `cancelCheckout`, so you go back to the cart if you tap the cart area. It'll be two touches to actually interact with the cart UI, but I think people will figure that out fast.

Here it is in action:

invisible mask

> What do you think, @supernova-at ?

I really don't like that 😝

<hr />

Here's another option that splits the difference:

screen shot 2019-02-05 at 11 37 16 am

I do actually like this one, what do you think @zetlen @soumya-ashok ?

Here's the current implementation for comparison:

screen shot 2019-02-05 at 11 39 07 am

zetlen:
Thank you @supernova-at . I fully agree that my first idea doesn't work. The lighter mask looks awesome to me too. Are you overriding the Mask styles as in my suggestion, or (my preference) adding an opacity parameter to Mask?

supernova-at:
> Are you overriding the Mask styles as in my suggestion, or (my preference) adding an opacity parameter to Mask?

I added opacity to Mask, yeah. I'll push it shortly.

packages/venia-concept/src/components/Mask/mask.js
```diff
@@ -11,14 +11,20 @@ class Mask extends Component {
...
+
+    static defaultProps = {
+        opacity: 0.5
     };
 
     render() {
-        const { classes, dismiss, isActive } = this.props;
+        const { classes, dismiss, isActive, opacity } = this.props;
         const className = isActive ? classes.root_active : classes.root;
+        const style = isActive ? { opacity } : {};
```
jimbo: This has to be in CSS. Opacity is purely a presentational choice and has no business here.

jimbo:
> ...or (my preference) adding an opacity parameter to Mask?

As a strict rule, we don't put presentational choices in JS. The `Mask` component doesn't need to know if masking is translucent or opaque, accomplished with `opacity` or an alpha channel on `background-color`, or handled by a binary image. All it needs to know is what state it's in.

packages/venia-concept/src/components/MiniCart/miniCart.js
sirugh: I know I'm late to the party here but do we need a component for this? Could it just be a div with a `className` and an `onClick` rendered by a conditional? ie
```js
{ isOpen ? <div className={classes.mask} onClick={cancelCheckout}></div> : null }
```

PR #812, opened 2019-02-04
feat(upward): Add JSON file parsing to upward-js and asset introspection to Venia templates

zetlen:
## Description

The purpose of this PR is to help Venia templates autodetect the generated names of build assets, and use them when generating script and link tags, rather than hardcoding asset names and forcing Webpack to output those filenames.

- Add a JSON filetype to upward-js which makes FileResolver resolve a parsed JSON object, so that other UPWARD operations can use values in the JSON.
- Add a WebpackAssetManifest plugin to Venia's build pipeline to create a JSON file for UPWARD to read.
- Refactor Venia templates to create prefetch links and script tags dynamically, from the contents of the manifest.
- Refactor the MagentoRouteHandler urlResolver preload to use data attributes instead of handwritten JSON.
- Add specialized app shells for other page types

**Note: This is technically additional feature in upward-js and Venia, but I think it's part of production-readiness--especially since the UPWARD spec actually calls for this functionality, so it was technically a gap. If (and only if) @tjwiebell and @bbatsche feel confident about adding this to FileResolver, I recommend we merge it before release. Otherwise, it'll have to wait for a minor rev fast follow.**

## Related Issue

Closes #ISSUENUM.

## Motivation and Context

We can't continue to scale the app without helping the app shells generate assets. Almost all other React projects generate HTML documents using an asset manifest.

## How Has This Been Tested?
- Added tests for new functionality
- Changed tests for changed functionality
- Loaded from a brand new session
- Tried all navigations

FEATURE
upward-js
venia-concept
peregrine

## Checklist:


packages/venia-concept/venia-upward.yml
```diff
@@ -109,6 +127,17 @@ resource:
     inline:
       template: './templates/notfound-shell.mst'
 
+assetManifest:
+  when:
+    - matches: env.NODE_ENV
+      pattern: 'development'
+      use: ./asset-manifest.json
```
tjwiebell: I'm seeing an error in the log with this configuration in dev mode: `Unrecognized configuration. Could not match a resolver to bundles: \'assetManifest.bundles\'`. Did you intend to check-in a hard-coded asset manifest, or should this file be generated and exist in the root?

I think the conditional fall through the bundles is masking this, or I'm misunderstanding the logging output.

tjwiebell: Disregard, there is no error in the log, I had a rogue `console.log` from some old debugging. I still don't see this file in my project, but it doesn't seem to be causing any problems in the app.

packages/peregrine/src/Router/resolveUnknownRoute.js
```diff
@@ -3,25 +3,39 @@
  * with the assigned Root Component (and its owning chunk) from the backend
  * @param {{ route: string, apiBase: string, __tmp_webpack_public_path__: string}} opts
  */
+
+// Some M2.3.0 GraphQL node IDs are numbers and some are strings, so explicitly
+// cast numbers if they appear to be numbers
```
sirugh: Shouldn't this sort of interp happen in the API? I assumed the HTTP request stringifies anyways so I guess my question is why we need to coerce to a number at all in the client side.

packages/venia-concept/webpack.config.js
```diff
@@ -155,8 +178,7 @@ module.exports = async function(env) {
                         test: new RegExp(
                             `[\\\/]node_modules[\\\/](${libs.join('|')})[\\\/]`
                         ),
-                        name: true,
-                        filename: 'js/vendor.js',
+                        // name: true,
```
sirugh: Do we need this?

packages/venia-concept/venia-upward.yml
tjwiebell: Discussed with James, this is my fault. WebPack serves `./asset-manifest.json` via virtual filesystem, that is only started up with `yarn watch:` and not `yarn stage:` commands.

packages/venia-concept/webpack.config.js
zetlen: No, this can be removed.

packages/peregrine/src/Router/resolveUnknownRoute.js
zetlen: We switched from this gross hand-built JSON (since UPWARD doesn't serialize JSON) into using HTML data attributes to preload the URL resolver stuff. All HTML attributes are strings, so we need to coerce to the right variable types for the query--and _presently_, all our category and CMS pages have `Int` IDs. But, @paliarush has told me that the ID field might become a string, or even the native GraphQL ID type, in the future. So this is a way to allow for that future possibility, while casting recognizable numbers today.

sirugh: > we need to coerce to the right variable types for the query

And this is required because we form the query (at which point type matters) on the client side?

I guess I'm just confused because in HTTP requests `param=123` is a string and the casting happens on the server.

zetlen:
@sirugh Removed comment, ready for review.

packages/peregrine/src/Router/resolveUnknownRoute.js
zetlen: Exactly. Apollo creates the query after picking up the urlResolver details.

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=812) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=812) before we can accept your contribution.<br/>**4** out of **5** committers have signed the CLA.<br/><br/>:white_check_mark: zetlen<br/>:white_check_mark: sirugh<br/>:white_check_mark: tjwiebell<br/>:white_check_mark: supernova-at<br/>:x: cherdman

sirugh:
When I ran the build locally I saw the `asset-manifest.json` file updated though I'm not quite sure how to "test" these changes. Using this branch I was able to run the local dev server but I don't think I have a way to verify the productionization of this code. @dpatil-magento have you done any testing of this code in prod? I think if you can test this and sign off we can go ahead and merge.

packages/venia-concept/src/queries/getProductDetail.graphql
```diff
@@ -49,6 +49,8 @@ query productDetail($urlKey: String, $onServer: Boolean!) {
                 }
             }
             meta_title @include(if: $onServer)
+            # Yes, Products have `meta_keyword` and
+            # everything else has `meta_keywords`.
```
sirugh: 😭

dpatil-magento:
Completed regression tests with upward-js and upward-php. looks good.

jimbo:
@sirugh  @tjwiebell Thanks for reviewing this one.

@zetlen I just took a quick look through. Nice work, and good explanatory comments throughout. 👍

Issue #839, opened 2019-02-06
Venia Unit Test Coverage for 2.0


sirugh:
# What

Our test coverage for Venia is abysmal. This issue is to track our progress in updating the coverage numbers which, at time of writing, are less than 50%.

Additionally we should work to migrate tests from `enzyme` to `react-test-renderer` at the same time.

https://itnext.io/testing-react-16-3-components-with-react-test-renderer-without-enzyme-d9c65d689e88

# How

If you want to help, check out the Venia coverage by running `yarn test` from the root. Pick a file from `packages/venia-concept` with low/red coverage and add a comment here saying that you are going to work on it so that we don't double up.

For example, I would see the following and comment here that I am going to write tests for `TextArea`.

[![Image from Gyazo](https://i.gyazo.com/5a1ff717433e022d96d5f439bd1d5e2d.png)](https://gyazo.com/5a1ff717433e022d96d5f439bd1d5e2d)

While working on the tests I've found it easier to see the results by modifying both `package.json` and `jest.config.js` to make sure I'm only testing the files/components I care about. For example:

In *jest.config.js*, to only collect coverage for `ProductImageCarousel`:
```js
collectCoverageFrom: [
  ...
  'packages/*/{src,lib}/components/ProductImageCarousel/*.js', 
  ...
],
```

In *package.json*, to only test for `ProductImageCarousel`:
```js
"scripts": {
  ...
  "test:dev": "jest --watch packages/venia-concept/src/components/ProductImageCarousel",`
  ...
}
```

sirugh:
Speaking of, I'll be writing coverage for `TextArea`!

jimbo:
And if you reference this issue from a PR, make sure not to use a [keyword that would close it](https://help.github.com/articles/closing-issues-using-keywords/).

sirugh:
Now writing tests for `ProductImageCarousel`.

supernova-at:
I'll tackle `simplePersistence.js`.

sirugh:
Now writing tests for `ProductOptions`.

supernova-at:
Writing tests for `UpwardServerError.js`.

supernova-at:
Writing tests for `packages/venia-concept/src/RootComponents/CMS/CMS.js`

supernova-at:
Picking up `catalog/asyncActions.js`: #873.

sirugh:
Now writing tests for `Checkout` 😨

supernova-at:
[...]

PR #856, opened 2019-02-06
test: TextArea

sirugh:
## Description
Adds unit tests for Venia TextArea.

## Related Issue

Related to #839.

## How Has This Been Tested?
`yarn run test`

[Image from Gyazo](https://gyazo.com/a89896346c58304d428fb7ca3ee24363)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -149,7 +149,7 @@ const jestConfig = {
     collectCoverageFrom: [
         // Code directories
         'packages/*/{src,lib}/**/*.js',
-        '!packages/*/{src,lib}/**/index.js',
+        '!packages/venia-concept/{src,lib}/**/index.js',
         // Not node_modules
         '!**/node_modules/**',
         // Not __tests__, __helpers__, or __any_double_underscore_folders__
--- a/jest.config.js
+++ b/jest.config.js
@@ -149,6 +149,7 @@ const jestConfig = {
     collectCoverageFrom: [
         // Code directories
         'packages/*/{src,lib}/**/*.js',
+        '!packages/*/{src,lib}/**/index.js',
         // Not node_modules
         '!**/node_modules/**',
         // Not __tests__, __helpers__, or __any_double_underscore_folders__
--- a/jest.config.js
+++ b/jest.config.js
@@ -149,7 +149,6 @@ const jestConfig = {
     collectCoverageFrom: [
         // Code directories
         'packages/*/{src,lib}/**/*.js',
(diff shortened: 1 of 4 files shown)


jest.config.js
```diff
@@ -149,6 +149,7 @@ const jestConfig = {
     collectCoverageFrom: [
         // Code directories
         'packages/*/{src,lib}/**/*.js',
+        '!packages/*/{src,lib}/**/index.js',
```
sirugh: I added this line because our coverage numbers were checking index.js files which IMO do not warrant testing.

packages/venia-concept/src/components/TextArea/textArea.js
```diff
@@ -13,9 +13,11 @@ export class TextArea extends Component {
             input: string
         }),
         cols: oneOfType([number, string]),
+        field: string.isRequired,
```
sirugh: Turns out that `informed` expects each field to be passed this prop. Found this bug while writing these tests :)

jest.config.js
jimbo: Some of our packages other than `venia-concept` have real code in `index.js` though.

packages/venia-concept/src/components/TextArea/__tests__/textArea.spec.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import TestRenderer from 'react-test-renderer';
+import { Form } from 'informed';
+import TextArea from '../textArea';
+
+jest.mock('src/classify');
+
+const classes = {
+    input: {}
+};
```
jimbo: I wouldn't pass classes in. It's better to just mock `classify` (which you're doing) and use the default classes, I think.

packages/venia-concept/src/components/TextArea/textArea.js
jimbo: Yeah, `field` is required by all Informed fields. It's the key used for storing field state in form context.

jest.config.js
sirugh: Ok, should I make this more specific or remove it all together?

jest.config.js
```diff
@@ -149,6 +149,7 @@ const jestConfig = {
     collectCoverageFrom: [
         // Code directories
         'packages/*/{src,lib}/**/*.js',
+        '!packages/venia-concept/{src,lib}/**/index.js',
```
sirugh: @jimbo I can remove this if you think it's a bit too opinionated right now.

PR #859, opened 2019-02-06
Docker compose

### LINKED ISSUE #734: Get PWA Studio running in Docker container

- [ ] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [x] `all`

## This issue is a:

- [ ] Bug
- [x] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

### Environment

## Description

In order to support Windows users and reproduceable environment, getting PWA to run inside of a container with https support would be very helpful.

## Expected result:

Run a docker container with PWA that works on any OS that has docker on it.

## Possible solutions:

Same as above.

sharkySharks:
## Description
This pull request adds the ability run PWA inside of a docker container with a secure https protocol.

## Related Issue
Closes #734 

## Motivation and Context
Containerizing PWA will allow for more CI configuration in the future and help towards a more platform agnostic development environment. So far the setup script in this PR will only work on Mac, but in the future the hope is to get this working on Windows Docker as well.

## How Has This Been Tested?
Manually.

To check this out, follow the instructions under the [docs draft file](https://github.com/magento-research/pwa-studio/pull/859/files#diff-7305b4d3ecf8be58090e6ebdd084a9ceR1)

FEATURE
pwa-buildpack, venia-concept, pwa-devdocs

## Checklist:
- [ ] I have added tests to cover my changes. - not sure the best testing approach for this but open to suggestions.


docker/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Run the `docker/run-docker` script from the root of the repository to create a container running PWA with a secure https protocol. 
+
+This script will:
+
+* copy the `.env.docker` environment variables to `.env` to be easily consumable by `docker-compose`
+* add a custom domain, configured in `.env.docker`, to the host system's `/etc/hosts` file
+* generate a self-signed ssl/tls certificate and trust the certificate in the system keychain
+* run `docker-compose build` to build the container network
+* run `docker-compose up` to start the container running PWA at the customer domain with https
```
sharkySharks: `custom` not `customer`

docker/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+After `docker/run-docker` is executed from the root of the repository, the default confgiuration will have the PWA application running at `https://pwa-docker.local`.
```
sharkySharks: `configuration`

lerna.json
```diff
@@ -1,11 +1,7 @@
 {
```
sharkySharks: do we need this file anymore?

now.json
```diff
@@ -5,5 +5,6 @@
         "node": "10.x.x"
     },
     "public": true,
-    "version": 1
+    "version": 1,
+    "type": "npm"
```
sharkySharks: because a `Dockerfile` is present `now` will automatically try to do a docker deploy, but I didn't set this up for that purpose, but I can look into it if needed.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -18,19 +18,19 @@ const addImgOptMiddleware = require('../Utilities/addImgOptMiddleware');
...
-  to PWADevServer configuration options.
-
-  More options for this feature are described in documentation.
-`;
+        and location on disk, simply add:
+        ${chalk.whiteBright('provideSecureHost: true')}
+        to PWADevServer configuration options.
+        
+        More options for this feature are described in documentation.
+        `;
```
sharkySharks: changes above are from running `prettier`

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -44,8 +44,8 @@ const PWADevServer = {
             contentBase: false, // UpwardPlugin serves static files
             compress: true,
             hot: true,
-            host: '0.0.0.0',
-            port: await portscanner.findAPortNotInUse(10000),
+            host: config.host || '0.0.0.0',
+            port: config.port || (await portscanner.findAPortNotInUse(10000)),
```
sharkySharks: maybe these values should just be checked against `process.env` values rather than messing with them in [webpack.config.js](https://github.com/magento-research/pwa-studio/pull/859/files#diff-1acbde853507ec5d5dac59ef5301a593R176)

bbatsche:
Since `.local` is typically used for mDNS, should we default use one of the four TLDs reserved for development and testing: https://en.wikipedia.org/wiki/Top-level_domain#Reserved_domains ?

brendanfalkowski:
When Google steamrolled `.dev` I think `.test` became the standard practice. I use that even though I hate it.

sharkySharks:
@bbatsche great point. I'm for either `.localhost` since that is where it is being served in the container, or possibly `.test`. My only hesitation about the latter is that it is not being staged for testing per say, but maybe that is what it will be used for ultimately.

sharkySharks:
well that's disconcerting, the now deploy does not fail the check if there are 500 errors from the now server...

docker/run-docker
```diff
@@ -0,0 +1,136 @@
...
+#
+####################################################################################
+
+ENVFILE=./.env
+HOSTSFILE=/etc/hosts
+HOSTPRESENT=false
+
+main () {
+  # copy environment variables to .env in root for docker-compose to consume for build
+  cp ./docker/.env.docker $ENVFILE
```
sharkySharks: maybe not wipe out the current .env so that users can switch between both configurations

docker/run-docker
```diff
@@ -0,0 +1,136 @@
...
+  -sha1 \
+  -days 3650 \
+  -nodes \
+  -x509 \
+  -keyout ./docker/certs/$DOMAIN.key \
+  -out ./docker/certs/$DOMAIN.crt \
+  -config $OpenSSLConf
+
+  # add the certificate to the system keychain and trust the certificate
+  sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ./docker/certs/$DOMAIN.crt
```
sharkySharks: look into adding this to the login keychain rather than the system keychain

docker/run-docker
```diff
@@ -0,0 +1,136 @@
...
+
+start_docker () {
+  message "Clearing any running containers"
+  docker-compose down
+
+  message "Building PWA image"
+  docker-compose build
+
+  message "Starting Docker network and containers"
+  message "You may see some warnings that @magento/venia-drivers could not be \n resolved. This is normal and not an error."
```
sirugh: I think this is going to be resolved with #881.

docker/run-docker
sharkySharks: see if this can be added to the docker-compose setup (adding an environment file for building) so that it doesn't require tampering with the host file system environment

sharkySharks: added a clean up of the .env at the end of this script and a message for users to switch to a new window if they want a clean environment to work outside of the docker setup

docker/run-docker
sharkySharks: looked into adding the cert to the login keychain and while I successfully added it to the login keychain file and it was present in the keychain access gui the cert was not being trusted by the browser. despite a lot of google searching I could not get this to work in the login keychain.

docker/run-docker
sharkySharks: removed since this has been merged now

docker/run-docker
```diff
@@ -0,0 +1,142 @@
...
+start_docker () {
+  message "Clearing any running containers"
+  docker-compose down
+
+  message "Building PWA image"
+  docker-compose build
+
+  # clean up: delete .env that was added for the docker-compose build
+  # this ensures that when users open a new shell the .env values won't carry over to the new shell
+  if [ -f .env ]; then
```
sharkySharks: hmmm this isn't actually working anymore, need to think of a different solution for this.

docker/run-docker
```diff
@@ -0,0 +1,149 @@
...
+  docker-compose build
+
+  # clean up the environment in the background
+  clean_up_env &
+
+  message "Starting Docker network and containers"
+  docker-compose up
+}
+
+clean_up_env () {
```
sharkySharks: @tjwiebell found a better way to clear the environment.

docker/run-docker
```diff
@@ -0,0 +1,149 @@
...
+  HOSTMAP="127.0.0.1 $DOMAIN"
+
+  add_domain_to_hostfile
+
+  generate_certificate
+
+  start_docker
+}
+
+add_domain_to_hostfile () {
```
zetlen: The `add_domain_to_hostfile` and `generate_certificate` functions look like they work great, but they seem to be duplicate work that [devcert](https://github.com/davewasmer/devcert) already does when we [call it from the Buildpack.Utilities.configureHost() function](https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/pwa-buildpack/src/Utilities/configureHost.js#L75). It both updates the hostfile (cross-platform) and generates a CA and self-signed certificate. So could we just run a Node script here?

docker/makeHostAndCert.js
```js
const { Utilities: { configureHost } } = require('../packages/pwa-buildpack/');
const { writeFileSync } = require('fs');
const { join } = require('path');

const exactDomain = process.env.PWA_STUDIO_PUBLIC_PATH;

configureHost({
  exactDomain,
  addUniqueHash: false
})
.then(({ ssl: { key, cert }}) => {
  console.log('Hostfile updated and TLS certificate created and trusted. Saving...');
  save(key, 'key');
  save(cert, 'crt');
})
.catch(e => {
  console.error(e);
  process.exit(1);
});

function save(contents, ext) {
  const filePath = join('.', 'certs', `${exactDomain}.${ext}`);
  writeFileSync(filePath, 'utf8');
  console.log(`Saved ${filePath}`);
}
```

Then you could remove a lot of this code and just call `node makeHostAndCert.js` in this script.

When I run `run-docker`, I'm always on a machine with NodeJS, right? This is no knock on the shell script implementation itself, which looks fine, but the Node stuff is cross-platform and already tested.

now.json
zetlen: Thanks for catching this. Yeah, we don't want to invest too much in refactoring our Now presence, so this is the right thing to do.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
zetlen: That's odd...what version of Prettier?

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -44,8 +45,10 @@ const PWADevServer = {
             contentBase: false, // UpwardPlugin serves static files
             compress: true,
             hot: true,
-            host: '0.0.0.0',
-            port: await portscanner.findAPortNotInUse(10000),
+            host: process.env.PWA_STUDIO_HOST_DEVELOPMENT || '0.0.0.0',
```
zetlen: Thank you. Dunno how I missed this.

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+COPY packages/peregrine/package.json ./packages/peregrine/package.json
+COPY packages/pwa-buildpack/package.json ./packages/pwa-buildpack/package.json
+COPY packages/upward-js/package.json ./packages/upward-js/package.json
+COPY packages/upward-spec/package.json ./packages/upward-spec/package.json
+COPY packages/venia-concept/package.json ./packages/venia-concept/package.json
+COPY package.json yarn.lock babel.config.js browserslist.js ./
+
+# install dependencies with yarn
+RUN yarn install
+
```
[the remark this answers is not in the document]
sharkySharks: The idea behind copying over all the package.json dependency files independently is to utilize Docker's [cache](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache) so that the build is fast on subsequent builds. The idea is that you only add files needed for the build, then run a build, then add all the other files because, presumably, dependencies change less than other files and the cache will not be disrupted until there is a change in a file. So the more files in a cached layer, the more likely that there will be changes and therefore require a rebuild of that layer.

When an item is added to a post-install then it will need to be added here too. Right now we don't have any other dependencies needed for the build other than the files above the install line.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
sharkySharks: i don't have prettier globally on my computer so this is coming from the repo, I can change it back and try it again. Not sure why this triggered in the first place.

docker/run-docker
sharkySharks: I see what you are saying here. The node script would depend on the ability of the user to be able to successfully run an `npm install` locally in order to use the library successfully to make the certs and host. I don't think we have that ability yet so I was keeping everything in a shell script outside of requiring any repo dependencies. With docker, you wouldn't need to require `node` to be on any user's machine, you could work out of docker exclusively and the only requirement would be having docker on your machine and the ability to clone a repository.

sharkySharks: I like the idea of this being immediately cross platform using devcert with your node script version of the above, but I ran into this error where one of the dependencies is saying that it is not valid on Win32: 

```
Error: \\?\C:\Users\User\code\pwa-studio\node_modules\node-zopfli-es\lib\binding\zop fli.node is not a valid Win32 application.                                           \\?\C:\Users\User\code\pwa-studio\node_modules\node-zopfli-es\lib\binding\zopfli.nod
```

And on my mac I ran into an error where it couldn't find the module by the path `../packages/pwa-buildpack`, not sure why. Let me know if there was anything else you did to get this to run for you locally. 

If this can solve the cross platform issue for docker then I'm for it but I think there is still more work for that next piece. I like that a bash script is not dependent on any installed node dependencies since docker builds are supposed to help with setting up a repeatable environment, the current state of it does not support Windows yet though so there's that.

sirugh: So the windows issue should be resolved if you rebase with master now that #864 is merged. Then you just need to `yarn install --ignore-optional`.

As for the mac issues I'm not sure, but I did find it necessary to run `yarn install`, then `yarn build` and _then_ any other commands like `run watch` because if you don't build it doesn't make thelinks.

sharkySharks: @sirugh did you try the node script specifically? I am still getting errors on Windows when I just try to run the node script. I've added support for Windows in the bash script for now. It made it a little busier than I would like but it covers getting a working installation up without a lot of fuss on the user's side. Windows does not make it easy to automate editing the host file or trusting certs.

sharkySharks: So after more thought, I don't think that it is a good idea to require an npm install in order to run a node script to generate a certificate and edit the host file. It makes the Docker setup have more dependencies when the only dependency _should_ be docker. Because we need https it requires having a trusted cert for a domain and I think that should be the only requirement added to setup docker, otherwise requiring an install before the docker install ends up duplicating work.

sirugh: > did you try the node script specifically?

I don't know, which one?

sharkySharks: I was referring to the one that James posted above.

sharkySharks:
Recent updates to this PR since previous reviews: I added support for Windows to the setup script and just added support for command line options to make the `docker/run-docker` script easier to on multiple runs after initial setup, especially for Windows users since there is some manual configuration needed that I could not automate at this time. No more major changes will be made to the PR, aside from addressing comments for changes, per usual pr review standards.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -41,11 +42,14 @@ const PWADevServer = {
         debug('configure() invoked', config);
         PWADevServer.validateConfig('.configure(config)', config);
         const devServerConfig = {
+            public: process.env.PWA_STUDIO_PUBLIC_PATH || '',
```
sharkySharks: found this [gem](https://webpack.js.org/configuration/dev-server/#devserverpublic)

sharkySharks:
@tjwiebell @zetlen comments have been addressed. Cert creation is using `devcert` and `PWADevServer` has been refactored.

tjwiebell:
@sharkySharks - I'm seeing some odd behavior with hot code reloading, curious if you're able to re-create. It seems like the build is always one save behind, eg:

1) Edit `packages/venia-concept/index.js` and add `console.log('Hot Code Test')` after the render
2) Triggers build and refresh in browser, but no `Hot Code Test` to be found in console
3) Remove line you added to `index.js`
4) Triggers build and refresh in browser, and now `Hot Code Test` shows up in console

LMK if you have any questions.

sharkySharks:
Thanks for checking this out @tjwiebell . I am seeing that behavior too. I watch the Application Clear Storage panel and I see that the Cache Storage takes an extra second or two to update and then if I refresh the page the correct content shows up as expected. I will look into this.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
zetlen: Heyyy, good catch! I absolutely approve of this

packages/venia-concept/package.json
```diff
@@ -26,7 +26,8 @@
     "storybook": "start-storybook -p 9001 -c .storybook",
     "storybook:build": "build-storybook -c .storybook -o storybook-dist",
     "validate-queries": "node ./validate-queries.js",
-    "watch": "webpack-dev-server --progress --color --env.mode development"
+    "watch": "webpack-dev-server --progress --color --env.mode development",
+    "watch:docker": "webpack-dev-server --progress --color --env.mode development --host pwa-docker.localhost"
```
tjwiebell: I don't think a docker specific script is necessary, just pass runtime arguments

`yarn watch -- --host pwa-docker.localhost`

This might allow you to pull this value from `.env.docker` so devs don't have to manually make sure this value is configured correctly in all places.

sharkySharks: this was one thing I really wanted to pull in from env vars so that all the configuration is only in the `.env.docker` file. I will look into this and hopefully get it to pull in 🙏

Dockerfile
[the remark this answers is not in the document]
paveq: Is there a reason to run "yarn install" and copy these files in Dockerfile build phase at all? Perhaps at least for development environment this could be done on entrypoint script, so that image itself does not need to contain the application files.

Usually application files are only "burned" to the image when final deployable production image is created.

docker-compose.yml
```diff
@@ -0,0 +1,36 @@
+version: '3'
+services:
+  nginx-proxy:
+    image: jwilder/nginx-proxy:alpine
```
paveq: Not sure I agree with using jwilder/nginx-proxy. It seems to work in quite non-standard way of reading docker containers and automatically exposing them. While this is nice, it is slightly unexpected behaviour, and can take while to understand how it works.

Dockerfile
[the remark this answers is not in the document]
sharkySharks: in order to build you need to `yarn install`. without the build dependencies then it would throw errors when running `yarn run build`. above I have copied only the build dependencies, package.json and relevant files needed for the build. Below on line 28 I am copying over the remainder of the package files before building. All files in the repo are needed to create a build that can then be spun up with the `CMD`. 

If you look at what `yarn run build` is executing, it runs the build commands of all the directories in the workspace under `packages/`, therefore it is needed to run `yarn install` to get all relevant dependencies needed for the build scripts to execute, as well as copy over all the contents of each package.

docker-compose.yml
sharkySharks: Can you elaborate on the unexpected behavior you are experiencing? Spinning up these services together automatically puts them in the same network, thereby exposing them to one another. [links](https://docs.docker.com/compose/compose-file/#links) also connect the containers in a way that is natural to docker. 

I did not add any comments to this docker-compose file, I can add some here to give some more context on what is going on.

sharkySharks: One other thing that I like about this image is that it allows you to set up as many proxies as you like. You can even move the proxy up a layer in your dev environment and have it running on its own and whenever you have a project you need a proxy for then you can add it to the network and proxy as many projects through it at once. For the purposes of PWA that means that you can run multiple PWA storefronts through the proxy, essentially have as many PWA domains as you like that all have their own service workers assigned and running without collision because they are running in different containers and have different domains, etc.

packages/venia-concept/package.json
sharkySharks: this is fixed now!

sharkySharks:
Addressed comments. hard coded values for the host name and the .env file passed for configuration are now being set by environment variables in both `docker-compose` and `Dockerfile`. I also added more comments throughout the files for more clarity, as well as the ability to set a custom configuration file via cli arg. Instructions for that in the docker/README.md as well as the docs draft have been updated.

tjwiebell:
LGTM, but will defer approval of `pwa-buildpack` changes to @zetlen.

packages/pwa-buildpack/src/WebpackTools/PWADevServer.js
```diff
@@ -41,11 +43,18 @@ const PWADevServer = {
...
             contentBase: false, // UpwardPlugin serves static files
             compress: true,
             hot: true,
+            watchOptions: {
```
sharkySharks: had to add the `watchOptions` back in for [hot reloading](https://github.com/Atyantik/react-pwa/issues/42) to work properly

zetlen:
@sharkySharks @tjwiebell This is an existing race condition in all workflows; it only shows up more in Docker because Docker is slightly slower. We fix this by adding functionality to the ServiceWorker to deregister itself in hot mode. I have an old branch where I was experimenting with this and other SW integrations to the Webpack HMR system, but I need to refactor this branch in light of new features in Workbox 4.0 that can help us: https://github.com/GoogleChrome/workbox/releases

supernova-at:
Adding the `hold` label pending a demo to community on the 16th.

PR #870, opened 2019-02-07
bug: Extend propTypes through wrapped components

sirugh:
## Description

While writing some tests I noticed that some of the propTypes weren't actually being vetted. I believe it is because we don't "expose" the propTypes in the classify HOC.

## Motivation and Context

PropTypes help us avoid bugs. This seemed to hide propTypes from being used.

## How Has This Been Tested?

I built and ran the app and saw new propTypes errors. This is good because it shows errors that were previously hidden.

[Image from Gyazo](https://gyazo.com/581e0e64264c920ef62c0a0c59fbab3f)

bug, venia-concept

## Checklist:


supernova-at:
Interesting. I've not actually seen this before but it seems logical.

sirugh:
Per discussion with Jimmy this is intentional, and kind of obvious when I think about it >.<

Shouldn't need to thread proptypes to the HOC because it is what provides those props at times.

supernova-at:
> Shouldn't need to thread proptypes to the HOC because it is what provides those props at times.

Yeah, that makes sense. And I'm guessing it spreads other props it gets down to the component it is wrapping, so no need to have `propTypes` declared on the HOC.

PR #874, opened 2019-02-08
test: ProductOptions

sirugh:
## Description
Adds unit tests for Venia ProductOptions.

## Related Issue

Related to #839.

## How Has This Been Tested?
`yarn run test`

[Image from Gyazo](https://gyazo.com/2d0a83a9abe745f7b4768524546a9867)


packages/venia-concept/src/components/ProductOptions/tile.js
```diff
@@ -27,15 +33,20 @@ class Tile extends Component {
...
         const className = classes[getClassName('root', isSelected, hasFocus)];
         const { label } = item;
 
         return (
-            <button {...restProps} className={className}>
+            <button
+                className={className}
+                onBlur={onBlur}
+                onClick={onClick}
+                onFocus={onFocus}
```
jimbo: This is curious. What's the goal here?

packages/venia-concept/src/util/getRandomColor.js
```diff
@@ -0,0 +1,11 @@
...
+// M2 GraphQL doesn't currently support them
+const cache = new Map();
+const memoize = fn => key =>
+    cache.has(key) ? cache.get(key) : cache.set(key, fn(key)).get(key);
+
+const getRandomColor = () =>
+    Array.from({ length: 3 }, () => Math.floor(Math.random() * 255)).join(',');
+
+export const memoizedGetRandomColor = memoize(getRandomColor);
+export default getRandomColor;
```
jimbo: Not a bad function to extract into a utility. 👍

packages/venia-concept/src/util/__mocks__/getRandomColor.js
```diff
@@ -0,0 +1,7 @@
+const getRandomColor = () =>
+    Array.from({ length: 3 }, () => Math.floor(Math.random() * 255)).join(',');
+
+// The primary reason for the existence of this mock is to return a non-random
+// memoized color.
+export const memoizedGetRandomColor = () => '123,123,123';
+export default getRandomColor;
```
supernova-at: Is there any reason `getRandomColor` needs to have an actual implementation? Can it just do the same thing as `memoizedGetRandomColor` ?

A mock containing randomly generated anything is a red flag, and its less code we'll have to maintain in the end.

packages/venia-concept/src/components/ProductOptions/swatch.js
```diff
@@ -1,20 +1,12 @@
...
-// TODO: replace with actual swatch colors or images from API
-// M2 GraphQL doesn't currently support them
-const cache = new Map();
-const memoize = fn => key =>
-    cache.has(key) ? cache.get(key) : cache.set(key, fn(key)).get(key);
-
-const getRandomColor = () =>
-    Array.from({ length: 3 }, () => Math.floor(Math.random() * 255)).join(',');
-const memoizedGetRandomColor = memoize(getRandomColor);
+import { memoizedGetRandomColor } from '../../util/getRandomColor';
```
supernova-at: Can import from `src/util/getRandomColor` if you don't want the relative paths.

packages/venia-concept/src/components/ProductOptions/swatch.js
```diff
@@ -46,11 +45,12 @@ class Swatch extends Component {
...
             isSelected,
             item,
-            // eslint-disable-next-line
-            itemIndex,
-            style,
-            ...restProps
+            onBlur,
+            onClick,
+            onFocus,
+            style
```
supernova-at: Add `style` to `propTypes` too.

packages/venia-concept/src/components/ProductOptions/tile.js
sirugh: ListItems contain a prop called `itemIndex` which is relevant for some things that can be items and  not relevant for others such as `tile` or `swatch`. I switched from spreading the rest to only using the specific props here because it felt more intentional. The alternative would be to delete the `itemIndex` from the `restProps` before spreading, but that felt weird too.

packages/venia-concept/src/util/__mocks__/getRandomColor.js
sirugh: > Is there any reason `getRandomColor` needs to have an actual implementation?

Are you asking about this mock file or in the more general sense?

> A mock containing randomly generated anything is a red flag

Definitely, I just wanted to match the actual implementation's export signature. I could just do `export default memoizedGetRandomColor as getRandomColor;`. Whatcha think?

packages/venia-concept/src/components/ProductOptions/swatch.js
sirugh: Thanks!! I will make this change.

packages/venia-concept/src/components/ProductOptions/swatch.js
sirugh: Oops, was it not already? Will do.

packages/venia-concept/src/components/ProductOptions/tile.js
jimbo: The problem with picking props to intentionally apply is you're creating a very arbitrary whitelist. Deleting a single prop might feel clumsy, but it's not leaky.

The pattern originally in place excluded `itemIndex` from `restProps` without having to use the `delete` operator, without excluding all kinds of other props: data attributes, `id`, etc.

brendanfalkowski: Sidebar — just following this via emails, it was a little confusing reading `restProps` and thinking REST as in API. I'm not sure if that's a conventional name, but was thinking I wouldn't have been confused by `otherProps` after seeing the context.

packages/venia-concept/src/util/__mocks__/getRandomColor.js
supernova-at: Yep, love it 👍

packages/venia-concept/src/components/ProductOptions/tile.js
sirugh: @brendanfalkowski the idea is that those are the "rest of the props" and yeah, it's sort of a convention by virtue of the feature's name. See [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters).

sirugh:
Deployment failure due to unrelated issue. Merging.

PR #875, opened 2019-02-08
Adds unit tests to Catalog actions

### LINKED ISSUE #873: Unit Tests for Catalog `asyncActions.js`
Write Unit Tests for `venia-concept/src/actions/catalog/asyncActions.js`.

supernova-at:
## Description

This PR adds unit tests to the Catalog's `actions.js` and `asyncActions.js`.

## Related Issue

Closes #873 .

## Motivation and Context

This PR is a part of the larger effort to increase unit test coverage in PWA Studio.

## How Has This Been Tested?

`yarn test`.

TEST

`venia-concept`.
## Checklist:


packages/venia-concept/src/components/Pagination/pagination.js
```diff
@@ -167,10 +167,12 @@ class Pagination extends Component {
 
         const queryPage = Math.max(
             1,
-            ~~getQueryParameterValue({
-                location,
-                queryParameter: 'page'
-            })
+            Math.floor(
```
supernova-at: This change is completely unrelated to the rest of this PR but [since I had to spend time looking up what the ~~ operator does](https://stackoverflow.com/questions/5971645/what-is-the-double-tilde-operator-in-javascript), I thought I'd save all future readers the effort.

packages/venia-concept/src/actions/catalog/asyncActions.js
```diff
@@ -18,7 +18,6 @@ export const getAllCategories = () =>
 export const setCurrentPage = payload =>
     async function thunk(dispatch) {
         dispatch(actions.setCurrentPage.receive(payload));
-        window.scrollTo(0, 0);
```
supernova-at: Scrolling the window is a concern of the UI, not the action handler. See the corresponding change in `category.js`.

packages/venia-concept/src/components/Pagination/pagination.js
sirugh: It *is* faster. Maybe a comment instead :)

jimbo: You looked it up and now you know what it does. No need to change it. 👍

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -57,6 +57,13 @@ class Category extends Component {
         id: 3
     };
 
+    componentDidUpdate(prevProps) {
+        // If the current page has changed, scroll back up to the top.
+        if (this.props.currentPage !== prevProps.currentPage) {
+            window.scrollTo(0, 0);
+        }
+    }
```
jimbo: Any time props change, scroll to top? Could that affect more cases than just the intended one?

packages/venia-concept/src/actions/catalog/asyncActions.js
jimbo: Having `window.scrollTo()` in this particular action was a bit arbitrary—it only covered this one use case—but I don't actually think there's anything wrong with it being part of an async action. Ultimately, async actions are a natural place for side effects to live, and scrolling the window is just a side effect.

packages/venia-concept/src/RootComponents/Category/category.js
supernova-at: Only happens when the `currentPage` prop changes, not when any prop changes.

In english, only when the page changes do we scroll to the top. I think that's exactly what we want, no more / no less.

packages/venia-concept/src/components/Pagination/pagination.js
supernova-at: So I dug a little deeper and [wrote up a little something](https://codepen.io/supernova_at/post/javascript-s-double-bitwise-not-operator).

Turns out that `~~` is not a straight replacement for `Math.floor`, specifically for the case of negative floating point numbers.

So on one hand since we're pulling our input from the query string, we can't be certain that it will be a positive number. But on the other hand, negative page numbers don't make sense (let alone negative floating point page numbers) so I don't think we care about them.

So I reverted this change and added a comment, as @sirugh suggested.

I do want to voice my opinion though, that we should strongly consider readability when we write our code - especially in Venia, which is supposed to be a reference implementation.

Is `~~` faster? Sure. Is it more concise? Yeah. Do either of those things matter here? I'd argue that they do not.

If `~~` was a widely understood and commonly used paradigm in JavaScript I'd have less of a problem, but my gut tells me that the majority of the people who run across and try to understand this code are going to have to look up double bitwise NOT, and we can make this much more developer friendly.

supernova-at:
> Just one thing -- asyncActions.js#14 was not tested,

Good catch, thanks for bringing that up - I'm actually not sure how to force the error case. Here's the code:

```js
        try {
            // TODO: implement rest or graphql call for categories
            // `/rest/V1/categories` requires auth for some reason
            const { default: payload } = await import('./mockData');

            dispatch(actions.getAllCategories.receive(payload));
        } catch (error) {
            dispatch(actions.getAllCategories.receive(error));
        }
```

First, I think it makes more sense to come back and write a test for this path once we get a real implementation in there, and second, I legit don't know how to mock / force an error to be thrown there 😝 .

sirugh:
I'm good with that answer @supernova-at.

PR #892, opened 2019-02-13
fix: minicart contents no longer render until detail fetch is complete

sirugh:
## Description
Due to async nature of cart detail fetching we re-render the contents of the mini cart after clicking the cart itself. This manifested in a visual bug wherein a user could open the cart, click the kebab for an item and it would disappear.

The root cause of this was traced back to the way we initiated the details request and the way the mini cart component responded. In the above gif you can see the cart rendering before the `GET_DETAILS/RECEIVE` action resolves. When that action resolves it updates the state of the cart causing a re-render.

## Resolution

To resolve this bug I added a `loading` value to the cart state that initializes to `false` and is set to `true` when the `getDetails` request action begins. If the `loading` prop is true, we render the loading spinner instead of the cart contents. If you're wondering why the cart contents were visible on the first load after you click the cart for the first time it is because we actually make a getDetails request that populates the `cart.details` in order to know the number of items in the cart for the cart icon.

This could all probably use a re-work but this PR should primarily fix the visual issue.

## How Has This Been Tested?
Manually. See gif: 

[Image from Gyazo](https://gyazo.com/aeae4dd42f6962a935a8eab096cff458)

bug, venia-concept

## Checklist:


packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -231,17 +231,13 @@ class MiniCart extends Component {
     }
 
     render() {
-        if (this.props.loading) {
```
sirugh: This wasn't actually doing anything. Notice there was no `loading` prop in anything that created `<MiniCart />` nor was there anything in `mapStateToProps`.

zetlen: Good catch. There are a few dead branches sitting around, and it's nice to identify them one by one.

jimbo:
@sirugh Thanks; this works. 👍 

I'll note here that once this is in place, a separate change should eventually follow to restore the original add-to-cart flow. The cart should open as soon as the user clicks `Add to Cart`, it should enter the `loading` state, the cart should update, and finally it should exit the `loading` state.

PR #902, opened 2019-02-14
Fix GraphQL Incompatibility with Magento Cloud Pro

### LINKED ISSUE #897: [BUG] GraphQL Incompatibility with Magento Cloud Pro
**Describe the bug**
A bug in Magento 2.3.0 is causing some GraphQL queries in Venia to return empty results when deployed to the Magento Cloud Pro tier.

**To Reproduce**
1. Install Venia w/ upward-php in Magento Cloud Pro (Production)
2. Navigate to home page
3. Screen will go white
4. Reload, and screen will display "No categories found"

**Expected behavior**
You see the five default Category tiles

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

**Additional context**

**Possible solutions**
The CategoryTree query still returns the expected Categories, so it seems if your query includes a "join" on products, it tricks the pro instance into returning the correct results. We'll put a temporary patch in place that add this join to the Navigation query.

tjwiebell:
## Description

Implement work around to query structure so responses are consistent in the Cloud Pro environment. Expected to be temporary until magento/graphql-ce#374 is fixed. With this fix in place we were finally able to test Fastly image optimization and realized we missed passing the width attribute, so that minor change has been included as well.

## Related Issue

Closes #897 

## How Has This Been Tested?

`yarn test` and manual ui check that categories and navigation still work as before

## Checklist:

### DIFF
--- a/packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
+++ b/packages/venia-concept/src/components/Navigation/__tests__/categoryTree.spec.js
@@ -37,7 +37,14 @@ const mocks = [
                             url_path: 'test-category/child-1',
                             product_count: 5,
                             children_count: '2',
-                            path: '1/2'
+                            path: '1/2',
+                            productImagePreview: {
+                                items: [
+                                    {
+                                        small_image: 'media/child-1.jpg'
+                                    }
+                                ]
+                            }
                         },
                         {
                             id: 3,
@@ -48,7 +55,14 @@ const mocks = [
                             url_path: 'test-category/child-2',
                             product_count: 3,
                             children_count: '1',
-                            path: '1/3'
+                            path: '1/3',
+                            productImagePreview: {
+                                items: [
+                                    {
+                                        small_image: 'media/child-2.jpg'
+                                    }
+                                ]
+                            }
                         }
                     ]
                 }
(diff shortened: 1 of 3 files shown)


packages/venia-concept/src/queries/getCategoryList.graphql
```diff
@@ -1,15 +1,24 @@
 # TODO: get only active categories from graphql when it is ready
-query categoryList($id: Int!) {
+
+query categoryList($id: Int!, $forNavigation: Boolean!) {
     category(id: $id) {
         id
+        name @include(if: $forNavigation)
```
tjwiebell: Would have loved to use named fragments here, but Magento doesn't currently support them (magento/graphql-ce#147). Sorry for the verbose directives :confused:

tjwiebell:
Deployed to Magento Cloud Pro here: https://dd2prod.dummycachetest.com/

packages/venia-concept/src/queries/getCategoryList.graphql
jimbo: Agreed. We really need fragments.

PR #910, opened 2019-02-17
Setting the addToCart button to disable on request

### LINKED ISSUE #855: "Add to Cart" button does not handle multiple clicks properly.
Description:
When user clicks on Add to Cart button multiple times then Cart gets updated as separate product for each click.

1. Go to Venia storefront and navigate to any product details page.
2. Select size and color and now click on Add to Cart button multiple times.

Actual:
Cart gets updated for each click as separate product.

Possible Solution:
"Add to cart" button should get disabled until the product gets added to cart. Once product added then again "Add to Cart" button gets enabled.

Screen:

LucasCalazans:
## Description
When the user clicks on Add to Cart button multiple times then Cart gets updated as a separate product for each click.

## Related Issue
Closes #855.

## Motivation and Context
When the user clicks on the button they don't receive feedback, so they can think that nothing is happening.

## How Has This Been Tested?
Tested locally using the main branch (release/2.0).

BUG
venia-concept

## Checklist:


tjwiebell:
@zetlen - there's a note on the issue that you'd prefer to see this fixed at the component level, which would be a slightly larger undertaking. To the same theme of some other PRs, would it be okay to merge this as-is, with the expectation its refactored once that Button component is completed?

packages/venia-concept/src/actions/cart/asyncActions.js
```diff
@@ -105,14 +105,13 @@ export const addItemToCart = (payload = {}) => {
...
 
-            dispatch(actions.addItem.receive({ cartItem, item, quantity }));
-
             // 2019-02-07  Moved these dispatches to the success clause of
             // addItemToCart. The cart should only open on success.
             // In the catch clause, this action creator calls its own thunk,
             // so a successful retry will wind up here anyway.
             await dispatch(getCartDetails({ forceRefresh: true }));
             await dispatch(toggleDrawer('cart'));
+            dispatch(actions.addItem.receive({ cartItem, item, quantity }));
```
tjwiebell: Moved this action below drawer toggle, since there was a slight delay between the button going live and drawer sliding out (`getCartDetails` takes a few ms). Without the move, it still prevents the accidental double click, but there is a window of clicking again before the drawer opens.

sirugh: This is fine. I'm not sure what best practice says about dispatching actions within actions but since we're already doing it...tbh I don't think it's normal for stores to open a cart right after adding an item. What if the user wants to buy more things? Doing this is a bit opinionated of us. Sorry, just sort of rambling here. This is fine :D

sirugh:
CircleCI is actin a fool! I'm still good with this so long as tests pass and the changes works as expected.

PR #917, opened 2019-02-19
Create new Guest Cart when Checkout resets

### LINKED ISSUE #916: Continue Shopping after Checkout
**Describe the bug**

After submitting a successful order, users are unable to add items to the cart.

**To Reproduce**
1. Add an item to the cart
1. Checkout and submit an order
1. Click "Continue Shopping" button
1. Attempt to add an item to the cart
1. Observe error notification on screen (and in browser console)

**Expected behavior**

Users should be able to add items to their cart after submitting an order.

**Screenshot**

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

**Additional context**

**Possible solutions**
A `guestCartId` isn't being created again after checkout.

supernova-at:
## Description

This PR creates a new guest cart whenever the `checkout/reset` action is triggered.

The `Continue Shopping` button triggers this action and thus creates a new guest cart on click. The user can add items to the newly created cart without error.

#### Move the Dispatch

The following code appears when `try`ing to add an item to the cart:

```js
        } catch (error) {
            const { response, noGuestCartId } = error;

            dispatch(actions.addItem.receive(error));

            // check if the guest cart has expired
            if (noGuestCartId || (response && response.status === 404)) {
                // if so, then delete the cached ID...
                // in contrast to the save, make sure storage deletion is
                // complete before dispatching the error--you don't want an
                // upstream action to try and reuse the known-bad ID.
                await clearGuestCartId();
                // then create a new one
                await dispatch(createGuestCart());
                // then retry this operation
                return thunk(...arguments);
            }
        }
```

We could move the `dispatch(actions.addItem.receive(error));` line to **after** the "no guest cart" retry logic so that the error only gets dispatched when there is truly an unhandled error.

This "no guest cart" retry logic does appear in many places, however, so we'd likely want to make this change app-wide (in all the `asyncActions.js` files).

#### Handle the Error in the Reducer

This has the effect of signaling the backstop error catching reducer that we're handling the error (by retrying) and that it shouldn't pop up a notification.

## Related Issue

Closes #916 .

## Motivation and Context

This change allows users to add items to their cart after a successful purchase.

## How Has This Been Tested?

See the repro steps in #916.

BUG

`venia-concept`

## Checklist:


supernova-at:
Honestly I prefer this solution:

> #### Create a new Guest Cart during `checkout/reset`
> This action gets triggered when the `Continue Shopping` button is clicked.

but I wanted to get a solution up quickly to discuss the options.

sirugh:
I'm a little confused -- what was the actual source of the error?

supernova-at:
> I'm a little confused -- what was the actual source of the error?

Oh, sorry - each cart has an id. After submitting an order, we wipe out that cart's id. When you attempt to add a new item, the error is that you can't because you don't have a cart.

In the current code we catch that error, create a new cart, and then retry adding the item. But we don't do any of that in redux, which is where the new backstop error notification reducer is looking to see if the error got handled. *It* thinks that the error went unhandled, so it pops up the error notification.

The solution I prefer is to create a new cart when the "continue shopping" button is clicked.

sirugh:
> The solution I prefer is to create a new cart when the "continue shopping" button is clicked.
This way when you attempt to add an item, a cart already exists and everyone is happy.

I think that makes a lot of sense!

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -27,6 +30,7 @@ export const cancelCheckout = () =>
 export const resetCheckout = () =>
     async function thunk(dispatch) {
         await dispatch(closeDrawer());
+        await dispatch(createGuestCart());
```
jimbo: Consider guarding this with a check to see if `guestCartId` exists in state.

supernova-at: Good call, I actually purposefully didn't do that because `createGuestCart` does it itself:

```js
export const createGuestCart = () =>
        const { cart } = getState();

        // if a guest cart already exists, exit
        if (cart.guestCartId) {
            return;
        }

// more code
```

Issue #898, opened 2019-02-14
[bug]: Windows contributor workflow is broken


sirugh:
**Describe the bug**
There are several issues when trying to get the app to run locally or install. The installation bit should be handled with #864 but trying to develop locally is a headache.

**To Reproduce**
On a Windows 10 machine, checkout the app and run `yarn install`. Then try to run any of the normal script commands like `yarn run lint` or `yarn run watch:venia` or `yarn run build`.

**Expected behavior**
All root package scripts work including `lint`, `test`, `build`, `watch:venia`, `watch:all`, etc.

**Please complete the following device information:**
 - Device: PC
 - OS: Windows 10
 - NPM version6.4.1
 - Node Version 10.15.1
 - Yarn Version 1.13.0

**Additional context**
This should address the workflow for contributors of _this_ repo. Developers should still be able to add pwa-studio packages like `venia-concept` or `peregrine` and utilize them within their own workflows.

**Problems and Possible solutions**

| Problem | Output | Fix |
| ----------- | ----------- |---|
| Use of env vars breaks commands, ie. `BABEL_ENV`. | |Use [cross-env](https://github.com/kentcdodds/cross-env) for commands that use env variables. |
|`yarn build`| Assuming fix using cross-env, still has issues: [![Image from Gyazo](https://i.gyazo.com/41c57e80c79f1a9d340ae745699aec49.png)](https://gyazo.com/41c57e80c79f1a9d340ae745699aec49) |RootComponent path resolution. |
| `yarn run prettier:validate` |[![Image from Gyazo](https://i.gyazo.com/ffbc9202c6e76bd0674eb553a1948e4f.png)](https://gyazo.com/ffbc9202c6e76bd0674eb553a1948e4f)|Replace `'` with `"`|
| `yarn run watch:venia`| [![Image from Gyazo](https://i.gyazo.com/35fef052c87c162cd0fbfc8c5bfe7a46.png)](https://gyazo.com/35fef052c87c162cd0fbfc8c5bfe7a46)|RootComponent path resolution.|
| `yarn test` | This actually runs the tests but there are issues with path resolution/matching ||

awilcoxa:
Moved to for grooming as this needs review and planning from devops for PR #899

awilcoxa:
Not a high priority at this time, moving the issue to backlog and @sirugh will close the PR and the team will come back to this at a later time/date.

sirugh:
[...]

PR #929, opened 2019-02-20
bug: Cannot re-assign const variable

sirugh:
This code will fail if `markDepInvalid` is executed. However, it is only executed on machines that do not have optional deps installed such as windows.

Nevertheless it is still a bug.

### DIFF
--- a/packages/pwa-buildpack/src/Utilities/addImgOptMiddleware.js
+++ b/packages/pwa-buildpack/src/Utilities/addImgOptMiddleware.js
@@ -50,7 +50,7 @@ function addImgOptMiddleware(app, env = process.env) {
     if (missingDeps) {
         console.warn(
             `Cannot add image optimization middleware due to dependencies that are not installed or are not compatible with this environment:
-${listInvalidDeps()}
+${missingDeps}
 Images will be served uncompressed.
 
 If possible, install additional tools to build NodeJS native dependencies:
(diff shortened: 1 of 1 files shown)


sirugh:
Additionally, something we may want to fix in this PR: https://github.com/magento-research/pwa-studio/pull/929/files#diff-4038ce758492d17b6add3daa61c02f5eL53 this `listInvalidDeps` doesn't seem to exist. Maybe it was missed on a commit?

zetlen:
Oh heck, that's big. Let's fix it.

zetlen:
I fixed these changes in a different branch, zetlen/windows-compat, which I wasn't planning to PR until after release. [Here is the relevant file we could cherry-pick](https://github.com/magento-research/pwa-studio/blob/zetlen/windows-compat/packages/pwa-buildpack/src/Utilities/addImgOptMiddleware.js): I've written tests for it, so that might add some stability. We would need to factor my new logging concept out of those changes.

Or we could go with this PR. I'm leaning towards just merging this PR.

sirugh:
I only noticed cause I was tinkering on my Windows box. BTW you should just take over that #898 for the foreseeable future :)

supernova-at:
> Or we could go with this PR. I'm leaning towards just merging this PR.

Yeah, we can fast follow with unit tests and other stuff.

PR #936, opened 2019-02-22
[TASK] Allow urls without .html for CMS.

lewisvoncken:
## Related Issue

Closes #ISSUENUM.

## Checklist:


awilcoxa:
@lewisvoncken can you add additional detail here and include an issue so that the team can review?

awilcoxa:
placing on hold until detail is added

lewisvoncken:
@awilcoxa 
There is no issue related to this Pull Request but here is a description why this should be merged:

In Magento it is possible to use a suffix in the url which is a configuration option

By default this option is set to .html but only for Catalog Urls

Because of this the CMS page routing won't work correctly because these rewrites don't have .html 

To make CMS page routing work correctly it should allow urls without .html Besides that the functionality won't change and the support of a different suffix is also added through this change.

packages/venia-concept/venia-upward.yml
```diff
@@ -190,7 +190,7 @@ cmsPageResult:
 urlKey:
   when:
     - matches: request.url.pathname
-      pattern: '^/(?:(.*)\.html)?'
+      pattern: '^/(?:(.*))?'
```
zetlen: This is the right idea (until we have the store config introspection requested in https://github.com/magento/graphql-ce/issues/535) but it needs to go further down in the stack. Since CMS pages have no suffixes, they now live in the same namespace as some of our hardcoded routes, like `checkout`. Those hardcoded routes need to resolve first--otherwise, an administrator could accidentally break their store by writing an SEO Friendly URL that conflicts with a builtin PWA route, AND, UPWARD would place the urlResolver call unnecessarily for every single request.

PR #712, opened 2018-12-22
Create tooltip component, add to color swatch at product page

AlexeyKaryka:
## This PR is a:

- [ X ] New feature
- [ ] Enhancement/Optimization
- [ ] Refactor
- [ X ] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will add tooltip component and append it to color swatch on the product page

PR #956, opened 2019-02-25
SwatchTooltip component

### LINKED ISSUE #700: UAT: Color Swatches Missing Tooltips
On the Product Display Page when a product has color options/configurations and swatches are displayed, selecting a color does not correctly display a tooltip for color name. 

Issue 19: https://wiki.corp.magento.com/display/FF/Venia+UX+Issues

- [x ] `venia-concept`
- [ ] `pwa-buildpack`
- [x ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [x ] Bug
- [ ] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

supernova-at:
## Description

This PR adds a new `SwatchTooltip` component and uses it in the `ProductOptions` views (product page and edit mini cart item).

It updates the closed #712.

## Related Issue

Closes #700 .

## Motivation and Context

There was a UAT ticket for providing these tooltips.

## How Has This Been Tested?

`yarn test` and manually on the product pages (PDPs) and in the mini-cart edit item dialog.

image

FEATURE

`venia-concept`

## Checklist:


soumya-ashok:
Had a look at the implementation. The left and right padding around the label needs to be reduced, and the vertical padding increased. It also looks odd that the tip is off-center.

screen shot 2019-02-28 at 10 16 50 am

packages/venia-concept/src/components/ProductOptions/__tests__/__snapshots__/option.spec.js.snap
```diff
@@ -14,30 +14,50 @@ exports[`renders Option component correctly 1`] = `
...
-          "--venia-swatch-bg": "123,123,123",
+    >
+      <span
+        className="tooltip"
+      >
+        red
+      </span>
+      <button
+        className="root"
+        itemIndex={0}
```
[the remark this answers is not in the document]
supernova-at: > This unrecognized DOM attribute angers the React.

😄

supernova-at:
> Had a look at the implementation. The left and right padding around the label needs to be reduced, and the vertical padding increased. On second thought, off-center tip makes the tooltip look unbalanced.

Totally agree.

> Would centering the tip cause any issues with display when the tooltip needs to appear on the swatch at the left edge of the screen?

It would, yeah. Would you mind if we merged this now and I create a follow up ticket to address these concerns?

soumya-ashok:
@supernova-at Sure, as long as we fix the issues I'm fine with changes being addressed in a follow-up ticket.

supernova-at:
> As long as we fix the issues I'm fine with changes being addressed in a follow-up ticket.

@soumya-ashok Awesome - created #1024 for follow up.

supernova-at:
> we have an opportunity for an a11y win here

Any chance I could convince you to allow this in the followup ticket? 😄

sirugh:
>Any chance I could convince you to allow this in the followup ticket?

I don't think it's *that* much work to add. `role="tooltip"` and `aria-describedby="<id>"` are easy enough. The focus things are also easy I think, probably just removing z-index. Heck I've got downtime so I could look into it tomorrow if you want. Just let me know :D

jimbo:
> Had a look at the implementation. The left and right padding around the label needs to be reduced, and the vertical padding increased. On second thought, off-center tip makes the tooltip look unbalanced.

Agreed on all points.

supernova-at:
In progress....

- [x] The tooltip never receives focus. Focus stays on the textbox.
- [ ] The tooltip widget can shown via keyboard focus or by the onmouse event.
- [ ] The tooltip widget can be hidden by removing focus from the text box or by moving the mouse off the textbox.
- [ ] The tooltip widget can be hidden by pressing the Escape key
- [ ] The tooltip is only hidden via JavaScript and CSS selectors. If JavaScript is not available the tooltip is shown.
- [x] The element that serves as the tooltip container has role tooltip.
- [ ] The element that triggers the tooltip references the tooltip element with aria-describedby.

supernova-at:
PR Updated:
  * Incorporated accessibility changes
  * Updated #1024 , as I took care of some easy wins here as long as I was touching the code again anyway. All that's left in 1024 is the centering of the triangle, which warrants larger design discussions.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
```diff
@@ -0,0 +1,111 @@
...
+        const ariaEnhancedChildren = React.Children.map(children, child => {
+            if (child.type !== 'button') {
+                return child;
+            }
+
+            // The button triggers the tooltip.
+            // Update its accessibility accordingly.
+            const ariaEnhancedButton = React.cloneElement(child, {
+                'aria-describedby': this.uniqueId,
+                className: `${child.props.className} ${classes.trigger}`,
```
supernova-at: I'm a little hesitant about composing the classnames together like this, but I'm not sure why. Basically I just want to add this component's class onto whatever is already present.

For example, `root tile` becomes `root tile trigger`.

packages/venia-concept/src/components/clickable.css
```diff
@@ -4,7 +4,6 @@
     display: inline-flex;
     justify-content: center;
     line-height: 1;
-    padding: 0;
```
supernova-at: This is not really a necessary change, I can back it out if desired.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
sirugh: I'm a little confused about what the desired outcome is here. In practice this will check for the `classes.trigger` prop from the css we have created, except that in `swatchTooltip.css` there is no `trigger` class so this is `undefined`. Do we even need this class addition?

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
```diff
@@ -0,0 +1,111 @@
...
+     */
+    onFocus = () => {
+        // If a child receives focus, do not close the tooltip yet.
+        clearTimeout(this.timeoutId);
+
+        this.setState({ isShowing: true });
+    };
+
+    onKeyDown = event => {
+        if (event.key === 'Escape') {
```
sirugh:
```suggestion
        if (event.key === 'Escape' || event.key === 'Tab') {
```

This would also hide it if you are moused in and then press tab. This could present an issue wherein there are focusable children within the button, imagine some text within the swatch, and pressing `Tab` should move focus to the text and keep the tooltip above the swatch.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
supernova-at: Nice catch, thanks!

If I remember correctly it was adding the `margin` around the swatch, which I thought definitely didn't belong in the `SwatchTooltip`, so I refactored the CSS and the `trigger` class doesn't exist anymore.

I'll remove this.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
supernova-at: This suggestion would help solve the problem of multiple tooltips showing at the same time. This occurs when one Swatch has focus and another is being moused over.

But it may not be worth the effort.

1. We'd have to add logic to `SwatchTooltip` to workaround the new issue that it would present, which would be duplicating a lot of what `blur` is supposed to tell us.
1. Since this isn't a form, users usually aren't navigating with keyboard and mouse.

We could add some sort of `TooltipController` component that ensures that only one tooltip is showing at a time, but that seems like overkill too.

supernova-at:
PR Updated:
  * Removes erroneous `className` composition during `cloneElement`

packages/venia-concept/src/components/clickable.css
jimbo: Clickable is essentially a mixin you would use for some shared styles between links and buttons of all kinds. You use it by writing `composes: clickable` in your component CSS. As such, it doesn't *change* the padding of a button to `0`, it *resets* the padding to `0`, like a default. Your component would override that padding as needed.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
```diff
@@ -0,0 +1,110 @@
...
+                // provided to it with the args it receives -
+                // in this case, a React SyntheticEvent.
+                onBlur: over([child.props.onBlur, this.onBlur]),
+                onFocus: over([child.props.onFocus, this.onFocus]),
+                onKeyDown: over([child.props.onKeyDown, this.onKeyDown]),
+                onMouseOver: over([child.props.onMouseOver, this.onMouseOver]),
+                onMouseLeave: over([
+                    child.props.onMouseLeave,
+                    this.onMouseLeave
+                ])
```
jimbo: I've never seen this pattern before. Ordinarily, you would call the props callback from the class handler. Any reason we shouldn't do that here?

Also, by convention, class handlers are named `handleEvent` rather than `onEvent`.

```jsx
class Foo extends Component {
  handleClick = event => {
    const { id, onClick } = this.props

    // this component's business
    event.preventDefault()

    // its parent's business
    // note: event object is treated as private
    // callback gets only what it needs
    if (typeof onClick === "function") {
      onClick(id)
    }
  }

  render() {
    const { handleClick } = this

    return <button onClick={handleClick} />
  }
}
```

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
jimbo: Generally speaking, if the expected behavior is that only one tooltip can be open at a time, then it's not local state: it's shared state. So it sounds like this situation does call for a "controller". Context is good for this purpose.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.css
```diff
@@ -0,0 +1,42 @@
...
+
+    /* Children positioning. */
+    display: flex;
+    align-items: center;
+    justify-content: flex-start;
+    padding: 0 0.75rem;
+}
+
+/* The tooltip triangle that points down. */
+.tooltip:before {
```
jimbo: Love me some pseudo-element triangles.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
sirugh: At my last company we had a `FocusManager` component that handled this sort of thing. Probably a bit overkill for now.

supernova-at: @jimbo great points all around.

> if the expected behavior is that only one tooltip can be open at a time

This is the piece I'm iffy about. I mean, it sounds right but it is also a pretty rare occurrence so I lean towards "out of scope" here. When this thing makes it into Peregrine we can definitely generalize and make improvements to it.

What do you think?

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
supernova-at: Yeah that's a good call - the reason this particular case is different is because this `SwatchTooltip` component is the parent, not the child.

In other words:

```jsx
<SwatchTooltip>
  <Swatch />
</SwatchTooltip>
```

So for a given event handler - let's say `onClick` - `SwatchTooltip` (the parent) wants its handler executed while also not trampling on `Swatch` (the child)'s handler.

I suppose I could do something like

```js
    onMouseOver = () => {
        // parent business.
        this.setState({ isShowing: true });

        // child business.
        const child = this.getChild(); // some logic that uses React.Children to get the right one
        if (child.props.onMouseOver) {
          child.props.onMouseOver();
        }
    };
```

and repeat that pattern for each event handler.

I'm mostly just trying to avoid tightly coupling these components.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
jimbo: > When this thing makes it into Peregrine we can definitely generalize and make improvements to it.

Yeah, that's probably the right approach.

packages/venia-concept/src/components/SwatchTooltip/swatchTooltip.js
jimbo: > I'm mostly just trying to avoid tightly coupling these components.

Why's that? It seems to me that the tooltip should be a child of the `Swatch`, rendered by `Swatch`, and pretty coupled to it.

supernova-at: > Why's that? It seems to me that the tooltip should be a child of the Swatch, rendered by Swatch, and pretty coupled to it.

Good point, that would definitely make things much simpler here. In the future we may want a generic `Tooltip` component like this, but that's out of scope here.

sirugh:
We're going to merge this with the caveat that we try to figure out how to invert the parent-child relationship of the ToolTip component.

PR #990, opened 2019-03-04
Compatibility with Magento 2.3.1 (2.1.0 Edition)

### LINKED ISSUE #842: Add support for 2.3 develop
As a product, the existing version of PWA studio (Venia) should ensure that all GraphQL queries conform to new schema changes released in 2.3.1 in order to ensure compatibility.

### LINKED ISSUE #1001: Add mapping layer to support 2.3.1 query schemas
- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

Query schemas have changed in 2.3.1, but most component APIs should not change.

- Avoid changing the API (prop types) of "dumb" components
- In "smart" components, define functions to translate query responses
- Pass query responses through these functions on their way to "dumb" components

tjwiebell:
## Description

Make changes to Venia that supports the backwards incompatible changes to GraphQL being release with Magento 2.3.1. Additionally adds mapping layer into data queries so our presentational components maintain backwards compatibility.

## Related Issue

Closes #842 
Closes #1001 

## Motivation and Context

Venia must be compatible with 2.3.1 release

## How Has This Been Tested?

Manually tested locally and ran jest

## Checklist:


jimbo:
@tjwiebell I've opened #1001. Let's see how we can use functions to avoid changes to most components.

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -60,6 +24,15 @@ class Product extends Component {
         window.scrollTo(0, 0);
     }
 
+    // map Magento 2.3.1 schema changes to v2.0.0 proptype shape to maintain backwards compatibility
+    mapProduct(product) {
```
sirugh: I talked it over with @tjwiebell so I understand why these functions exist but I don't really like them :P I would prefer if we could do this in a single place such as a normalization/transform layer. I can see this getting unwieldy.

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=990) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=990) before we can accept your contribution.<br/>**2** out of **3** committers have signed the CLA.<br/><br/>:white_check_mark: tjwiebell<br/>:white_check_mark: jimbo<br/>:x: dpatil-magento

packages/venia-concept/src/RootComponents/Product/Product.js
supernova-at: Minor consistency issue here with how class functions are defined. Line 18 above uses a function expression instead.

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -60,6 +24,15 @@ class Product extends Component {
         window.scrollTo(0, 0);
     }
 
+    // map Magento 2.3.1 schema changes to v2.0.0 proptype shape to maintain backwards compatibility
```
supernova-at: Could make this comment a little more clear that "v2.0.0" refers to PWA and not Magento.

Edit: in all the places where this comment appears

packages/venia-concept/src/components/CategoryList/categoryList.js
```diff
@@ -30,6 +30,21 @@ class CategoryList extends Component {
         ) : null;
     }
 
+    // map Magento 2.3.1 schema changes to v2.0.0 proptype shape to maintain backwards compatibility
+    mapCategory(categoryItem) {
+        categoryItem.productImagePreview.items = categoryItem.productImagePreview.items.map(
```
supernova-at: I think we're pretty safe in assuming that none of `categoryItem`, `.productImagePreview`, or `.items` will be null or undefined, right?

packages/venia-concept/src/components/Gallery/gallery.js
```diff
@@ -13,21 +13,7 @@ class Gallery extends Component {
...
-                    regularPrice: shape({
-                        amount: shape({
-                            value: number.isRequired,
-                            currency: string.isRequired
-                        }).isRequired
-                    }).isRequired
-                }).isRequired
-            })
-        ),
+        data: array,
```
supernova-at: In `RootComponents/Product/Product.js` we blew away the `data` propType entirely.

I'm not clear on the difference between these two, and when I would use each approach.

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -30,11 +31,13 @@ class Option extends Component {
     };
 
     get listComponent() {
-        const { attribute_code } = this.props;
+        const { attribute_code, values } = this.props;
 
         // TODO: get an explicit field from the API
         // that identifies an attribute as a swatch
-        return attribute_code === 'fashion_color' ? SwatchList : TileList;
+        const optionType = getOptionType({ attribute_code, values });
```
supernova-at: Does `getOptionType` need `values` ?

packages/venia-concept/src/components/ProductOptions/getOptionType.js
```diff
@@ -0,0 +1,5 @@
+const customAttributes = {
+    fashion_color: 'swatch'
+};
+
+export default ({ attribute_code: code }) => customAttributes[code];
```
supernova-at: I know it's small but a unit test here would be good.

packages/venia-concept/src/components/SearchBar/autocomplete.js
```diff
@@ -65,6 +65,17 @@ class SearchAutocomplete extends Component {
 
     handleOnProductOpen = () => this.props.updateAutocompleteVisible(false);
 
+    // map Magento 2.3.1 schema changes to v2.0.0 proptype shape to maintain backwards compatibility
+    mapProducts(products) {
+        return products.map(product => {
+            if (typeof product.small_image === 'object') {
+                product.small_image = product.small_image.url;
+            }
```
supernova-at: Echoing @sirugh 's sentiments earlier, this does seem like the same transform each time so it could be nice to re-use this logic.

Understandable if that's a future nice to have.

packages/venia-concept/src/util/appendOptionsToPayload.js
```diff
@@ -18,15 +18,27 @@ const appendOptionsToPayload = (
...
+            // if any option selection fails to match any standard attribute
+            // and also fails to match any custom attribute
+            // then this isn't the correct variant
+            if (!matchesStandardAttribute && !matchesCustomAttribute) {
                 return false;
             }
         }
 
+        // otherwise, every option selection matched
+        // and this is the correct variant
```
supernova-at: Good comments 👍

packages/venia-concept/src/RootComponents/Product/Product.js
sirugh: Actually I'm not sure why that function even exists. Doesn't seem like we're using addToCart anywhere, instead we're using the connected prop `addItemToCart`.

sirugh: But I think you're right, we should be using the class function rather than prototype otherwise `this.mapFoo` won't work.

packages/venia-concept/src/components/CategoryList/categoryList.js
tjwiebell: Correct, GraphQL will always return an array.

packages/venia-concept/src/components/Gallery/gallery.js
tjwiebell: Sorry, some comments got lost re-creating the PR, but the change in Product was just cleanup. The Product component doesn't receive its data as a prop; Jimmy and I had no idea why that existed at all. This component only receives `id` and `addToCart`.

packages/venia-concept/src/components/ProductOptions/option.js
tjwiebell: From Jimmy's PR: https://github.com/magento-research/pwa-studio/pull/1005#discussion_r264333689

packages/venia-concept/src/components/SearchBar/autocomplete.js
tjwiebell: @jimbo can elaborate more, but I think the intention is that this code would be obsolete in the next major, and be removed. Given it's temporary, so we didn't want to over engineer a solution that would just be deleted in the near future.

packages/venia-concept/src/RootComponents/Product/Product.js
tjwiebell: Oh wow, good catch @sirugh, that function is a good candidate for removal. I'm not keen on what's @api yet; can we safely remove it in a minor?

And who do you mean is right? The `mapProduct` function is fine as-is or Andy's suggestion that we should use expressions? My understanding was that function expressions were cleaner than binding params in constructor, but that's solved by the new shorthand.

tjwiebell: See https://github.com/magento-research/pwa-studio/pull/990#discussion_r265708897

sirugh:
```js
class Foo {
  constructor() {
    	console.log(this);
  }
  first () {
    	console.log('first');
  }
  second = () => {
	console.log('second');
  }
}

const foo = new Foo() // "Foo { second: f }"
```

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -16,41 +15,6 @@ import productQuery from 'src/queries/getProductDetail.graphql';
...
-                        description: string,
-                        short_description: string,
-                        canonical_url: string
-                    })
-                ).isRequired
-            }).isRequired
-        })
-    };
-
     addToCart = async (item, quantity) => {
```
sirugh: @jimbo can we remove this function safely? It's not being used by anything in `pwa-studio` but it is technically part of the public API of this component.

packages/venia-concept/src/components/ProductOptions/getOptionType.js
tjwiebell: Coverage report has this covered. Is it still best practice to add an individual test for this file?

```
 venia-concept/src/components/ProductOptions                            |      100 |      100 |      100 |      100 |                   |
  getOptionType.js                                                      |      100 |      100 |      100 |      100 |                   |
  index.js                                                              |        0 |        0 |        0 |        0 |                   |
  option.js                                                             |      100 |      100 |      100 |      100 |                   |
  options.js                                                            |      100 |      100 |      100 |      100 |                   |
  swatch.js                                                             |      100 |      100 |      100 |      100 |                   |
  swatchList.js                                                         |      100 |      100 |      100 |      100 |                   |
  tile.js                                                               |      100 |      100 |      100 |      100 |                   |
  tileList.js                                                           |      100 |      100 |      100 |      100 |                   |
```

packages/venia-concept/src/util/makeUrl.js
```diff
@@ -1,7 +1,7 @@
 import memoize from 'lodash/memoize';
 
 // Tests if a URL begins with `http://` or `https://` or `data:`
-const absoluteUrl = /^(data|https)?:/i;
+const absoluteUrl = /^(data|https|http)?:/i;
```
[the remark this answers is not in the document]
tjwiebell: Added other two protocols to unit test

packages/venia-concept/src/RootComponents/Product/Product.js
tjwiebell: The `v` stands for Venia, come on 😁 (updated four occurrences of this comment)

packages/venia-concept/src/util/__tests__/makeUrl.spec.js
```diff
@@ -1,8 +1,15 @@
...
-    expect(makeUrl(absoluteUrl)).toBe(absoluteUrl);
+    const absoluteUrls = [
+        'data://example.com/foo.png',
+        'http://example.com/bar.png',
+        'https://example.com/baz.png'
+    ];
+
+    absoluteUrls.forEach(url => {
+        expect(makeUrl(url)).toBe(url);
+    });
```
supernova-at: 👍

packages/venia-concept/src/components/ProductOptions/getOptionType.js
supernova-at: I'm assuming it's covered because some other test exercises this code path.

I think it is best practice to have unit tests for any public function exported by a module. These are the discrete "units" that we want to ensure work as we expect.

If this code were to change in a breaking way, we'd know immediately via a unit test in a `getOptionType.spec.js` file, not a failing unit test in some other file that we'd then have to go debug and track it down to this function.

Further, who knows what that other file's unit test is actually checking for, so if this code were to break we may not even be alerted via a failing test.

That other code could get refactored to not use this function or something else too, but we'd still want to have confidence that it works.

packages/venia-concept/src/components/SearchBar/autocomplete.js
zetlen: That's exactly the idea. It's the same idea as our requirement that all REST calls specify the URL and stringify the payload inline. It's a way of emphasizing that the REST calls are temporary workarounds, and making it more obvious how big a change it is when they are removed.

packages/venia-concept/src/components/ProductOptions/getOptionType.js
zetlen: We don't need to unit test internals. I'm as guilty of this as anyone, but I say it's okay not to have an individual test per file.

packages/venia-concept/src/RootComponents/Product/Product.js
zetlen: The only reason to use arrow functions when defining class members is if that function will be passed directly as a value, usually to a React component prop. Let's definitely not be lazy about this--every arrow function definition is O(n), whereas every class member is O(1).

zetlen: @sirugh I thought about that too, but Tommy is right that this is the right way to do it for a workaround. If we create an abstract mapping layer, we're already halfway down the path to an ORM.

packages/venia-concept/src/components/ProductOptions/getOptionType.js
supernova-at: Slipped my mind that this was internal.  This definitely isn't blocking a 👍 , I was just responding to the best practice question. I'm cool with it as is 👍 .

packages/venia-concept/src/components/CategoryList/categoryList.js
zetlen: Small but important thing: multiple components may use this object, so we shouldn't mutate it.

Immutable version:

```js
mapCategory(categoryItem) {
  const { items } = categoryItem.productImagePreview;
  return {
    ...categoryItem,
    productImagePreview: {
      items: items.map(item => {
        const { small_image } = item;
        return {
          ...item,
          small_image: typeof small_image === 'object' ? small_image.url : small_image
        }
      })
    }   
  }
}
```

packages/venia-concept/src/components/Gallery/__tests__/gallery.spec.js
```diff
@@ -21,7 +23,9 @@ const items = [
     {
         id: 2,
         name: 'Test Product 2',
-        small_image: '/test/product/2.png',
+        small_image: {
+            url: '/test/product/2.png'
+        },
```
zetlen: Add a test here and in other relevant tests for 2.3.0 style data.

packages/venia-concept/src/components/Gallery/items.js
```diff
@@ -41,14 +32,25 @@ class GalleryItems extends Component {
             : defaultPlaceholders;
     }
 
+    // map Magento 2.3.1 schema changes to Venia 2.0.0 proptype shape to maintain backwards compatibility
+    mapGalleryItem(item) {
+        if (typeof item.small_image === 'object') {
```
zetlen: Immutable version:

```js
mapGalleryItem(item) {
  const { small_image } = item;
  return {
    ...item,
    small_image: typeof small_image === 'object' ? small_image.url : small_image
  }
}
```

Safer!

packages/venia-concept/src/components/SearchBar/autocomplete.js
zetlen: Immutable version:

```js
mapProducts(products) {
  return products.map(product => {
    const { small_image } = product;
    return {
      ...product,
      small_image: typeof small_image === 'object' ? small_image.url : small_image 
    }
  })
}
```

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -60,6 +24,15 @@ class Product extends Component {
         window.scrollTo(0, 0);
     }
 
+    // map Magento 2.3.1 schema changes to Venia 2.0.0 proptype shape to maintain backwards compatibility
+    mapProduct(product) {
+        if (typeof product.description === 'object') {
+            product.description = product.description.html;
```
zetlen: Immutable version:

```js
mapProduct(product) {
  const { description } = product;
  return {
     ...product,
      description: typeof description === 'object' ? description.html : description 
  }
}
```

packages/venia-concept/src/components/ProductOptions/getOptionType.js
jimbo: Yes to both. It's effectively internal and covered by other tests, so it's not necessary to have a unit test for it. But I'd like to have one anyway.

PR #1005, opened 2019-03-08
Accommodate custom attributes

### LINKED ISSUE #983: Change Product components to use attribute collections instead of Venia custom attrs
Venia currently uses a custom attribute `fashion_color` in its product detail query. This makes it strictly dependent on Venia sample data.

Like Luma, Venia is meant to be consumed in multiple contexts, so we need to change these queries to be agnostic and the components that use them to be agnostic.

Acceptance Criteria

- No direct custom attribute selection in any queries
- No switching off attribute names to pick options display
- Swatch display based on attribute metadata

jimbo:
## Description

Make `venia-concept` less coupled to the Venia sample data by accommodating custom attributes.

## Related Issue

Closes #983.

## Motivation and Context

**Custom attributes in queries**

Venia uses the custom attributes `fashion_color` and `fashion_size`. These product options are required on some configurable products; for such a product, a shopper must select values for these options in order to add the product to her cart. This configured product is called a `variant`.

Our queries include these custom attributes directly, because in Magento 2.3.0 there is no way to retrieve a list of them. In Magento 2.3.1, the `ConfigurableVariant` interface has an `attributes` field, so we can now retrieve that list (and remove `fashion_color` and `fashion_size` from our product queries).

**Attribute type**

There is one remaining area in which `venia-concept` references Venia's custom attributes directly. Within product options, the `Option` component is responsible for determining which component is used to render the list of selectable option values. In both 2.3.0 and 2.3.1, GraphQL does not expose `swatch` as an attribute type, so we check for `fashion_color` specifically and render a `SwatchList` in that case.

Since we still can't get this info from GraphQL, I've isolated that logic to a private function so we can potentially change it in the future without changing the component.

## Verification

Ensure configurable products can be added to cart and purchased.

## How Has This Been Tested?

Pending.

### DIFF
--- a/packages/venia-concept/src/components/ProductOptions/option.js
+++ b/packages/venia-concept/src/components/ProductOptions/option.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { arrayOf, func, object, shape, string } from 'prop-types';
 
 import classify from 'src/classify';
+import getOptionType from './getOptionType';
 import SwatchList from './swatchList';
 import TileList from './tileList';
 import defaultClasses from './option.css';
@@ -30,11 +31,13 @@ class Option extends Component {
     };
 
     get listComponent() {
-        const { attribute_code } = this.props;
+        const { attribute_code, values } = this.props;
 
         // TODO: get an explicit field from the API
         // that identifies an attribute as a swatch
-        return attribute_code === 'fashion_color' ? SwatchList : TileList;
+        const optionType = getOptionType({ attribute_code, values });
+
+        return optionType === 'swatch' ? SwatchList : TileList;
     }
 
     render() {
--- a/packages/venia-concept/src/components/ProductOptions/getOptionType.js
+++ b/packages/venia-concept/src/components/ProductOptions/getOptionType.js
@@ -1,0 +1,5 @@
+const customAttributes = {
+    fashion_color: 'swatch'
+};
+
+export default ({ attribute_code: code }) => customAttributes[code];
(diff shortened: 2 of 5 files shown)


tjwiebell:
Ignore my suggestion, issue seems to be in changes in `appendOptionsToPayload::selectedVariant`; this data is not being appended because that validation is failing.

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -30,11 +31,13 @@ class Option extends Component {
     };
 
     get listComponent() {
-        const { attribute_code } = this.props;
+        const { attribute_code, values } = this.props;
 
         // TODO: get an explicit field from the API
         // that identifies an attribute as a swatch
-        return attribute_code === 'fashion_color' ? SwatchList : TileList;
+        const optionType = getOptionType({ attribute_code, values });
```
sirugh: Do you need to pass `values` to this function? Didn't see it used in the implementation.

jimbo:
> Ignore my suggestion, issue seems to be in changes in `appendOptionsToPayload::selectedVariant`; this data is not being appended because that validation is failing.

Yeah, I had a small bug in a push I made after opening this PR. It should be fixed now. 👍

packages/venia-concept/src/components/ProductOptions/option.js
jimbo: It's not currently used in the implementation because the implementation is mocked. In the future, when `getOptionType` is capable of doing something, it will need both arguments.

PR #985, opened 2019-03-04
Fixed HTML special characters issue

yogeshsuhagiya:
## Description
Fixed HTML special characters issue for the product name on the category page and product view page.

## Related Issue
#984 

## Screenshots (if appropriate):
screenshot

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have linked an issue to this PR.
- [ ] I have indicated the change type and relevant package(s).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] All CI checks are green (linting, build/deploy, etc).
- [ ] At least one core contributor has approved this PR.

PR #1010, opened 2019-03-11
bugfix: Fix html special chars

### LINKED ISSUE #984: [BUG] HTML special characters not displyaing properly.
**Describe the bug**
HTML special characters like `&trade;` are not appearing properly on category and product detail page.

**To Reproduce**
1. Login into Magento 2 Backend
2. Goto 'Catalog > Products'
3. Search 'Carmina Earrings' product
4. Edit the product and replace the product name with `Carmina Earrings &trade;`
5. Now goto pwa studio's root directory and run the server `yarn run watch:venia`.
6. In navigation menu, click on *Accessories* menu and then open `Carmina Earrings` product.

**Expected behavior**
The product name should look like: `Carmina Earrings ™` instead of `Carmina Earrings &trade;`.

**Screenshots**
screenshot_list
screenshot_product_view

 - Device: [e.g. iPhone6, PC]
 - OS: [e.g. iOS8.1, Windows 10]
 - Magento Version
 - NPM version `npm -v`
 - Node Version `node -v`

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

**Additional context**

**Possible solutions**

sirugh:
Recreated based on #985.

## Description
Fixed HTML special characters issue for the product name on the category page and product view page.

## Related Issue
Closes #984 

screenshot

## Checklist:

### DIFF
--- a/packages/venia-concept/src/components/Gallery/item.js
+++ b/packages/venia-concept/src/components/Gallery/item.js
@@ -73,7 +73,7 @@ class GalleryItem extends Component {
                     {this.renderImage()}
                 </Link>
                 <Link to={resourceUrl(productLink)} className={classes.name}>
-                    <span>{name}</span>
+                    <span dangerouslySetInnerHTML={{ __html: name}} />
                 </Link>
                 <div className={classes.price}>
                     <Price
--- a/packages/venia-concept/src/components/Gallery/item.js
+++ b/packages/venia-concept/src/components/Gallery/item.js
@@ -73,7 +73,7 @@ class GalleryItem extends Component {
                     {this.renderImage()}
                 </Link>
                 <Link to={resourceUrl(productLink)} className={classes.name}>
-                    <span dangerouslySetInnerHTML={{ __html: name}} />
+                    <span dangerouslySetInnerHTML={{ __html: name }} />
                 </Link>
                 <div className={classes.price}>
                     <Price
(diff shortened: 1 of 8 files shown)


sirugh:
@yogeshsuhagiya I remade this PR for you. Please take over it for now. You should fill out the sections that are missing content in the PR description such as "Motivation and Context" and "How has this been tested". Providing this information makes it easier for us to understand your changes. Without it we will likely push this to the bottom of the stack and prioritize other issues that have more detail.

packages/venia-concept/src/components/Gallery/item.js
```diff
@@ -73,7 +73,7 @@ class GalleryItem extends Component {
                     {this.renderImage()}
                 </Link>
                 <Link to={resourceUrl(productLink)} className={classes.name}>
-                    <span>{name}</span>
+                    <span dangerouslySetInnerHTML={{ __html: name }} />
```
jimbo: This is not the right approach. It's a vector for XSS, and it doesn't reflect the purpose of this node (`name` should be a `utf-8` string, not HTML).

sirugh: We could probably close this PR and the associated issue then, right? Seems like it isn't a real bug if the data should be a `utf-8` string already.

jimbo:
Product names should be `utf-8` strings, so you should be able to include characters such as `™` directly in a name—without HTML entities.

Even if that weren't the case, though, we shouldn't just render database content as raw HTML, since that's what enables XSS. (If there are any other places in the app where that's happening, we should fix them.)

PR #1019, opened 2019-03-13
[FEATURE] When clicking on the search icon button, focus the search input

### LINKED ISSUE #1018: [FEATURE] When clicking on the button, a frame appears
The frame **(outline)** appears when clicking on the button (search icon and etc.).

**To Reproduce**
1. Go to 'Home'
2. Click on 'Search'
3. See bug

**Expected behavior**
The behavior when clicking on the elements, must be all exactly the same.

**Screenshots**

Dedicated search: (Dekstop)

Screenshot from 2019-03-13 13-32-29

On iPhone 6/7/8:

Screenshot from 2019-03-13 13-33-55

**Environment:**
 - Device: iPhone6/7/8, PC
 - OS: Ubuntu
 - Browser Chrome
 - Version 72.0.3626.81
 - NPM version `6.7.0`
 - Node Version `10.15.1`

- [X] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

**Possible solutions**
Remove the frame from the element.

DmitryShmaliuk:
## Description

Focus the search field when the search button is clicked.

## Related Issue

Closes #1018.

## Motivation and Context

When clicking the search button icon the field was not focused.

## How Has This Been Tested?

The behavior of the item has been tested on the PC as well as on mobile devices.

## Screenshots
Screenshot from 2019-03-13 13-57-12
Screenshot from 2019-03-13 13-54-21

**BUG**, pkg:venia-concept

## Checklist:

- [X] I have read the **CONTRIBUTING** document.
- [X] I have linked an issue to this PR.
- [X] I have indicated the change type and relevant package(s).
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


packages/venia-concept/src/index.css
```diff
@@ -99,6 +99,7 @@ button {
     padding: 0;
     touch-action: manipulation;
     user-select: none;
+    outline: none;
```
belbiy: It is considered a bad practice to remove the outline from an active element (http://www.outlinenone.com/).

I would start by asking why the outline on the button feels wrong. 

Perhaps, the search button should move the focus to the input when the button is tapped? 
Or the style of the outline does not meet the expectations?

DmitryShmaliuk: @belbiy good remark, thank you. I read the article you provided and I agree with you that the outline should be left on the desktop, but do we need it on mobile devices?

DmitryShmaliuk: It also seems to me that we need to move the focus after we have clicked on the active element. **For example:** Click on the search element, and after, the focus is transferred to the input field to search.

DmitryShmaliuk: [Removing that ugly :focus ring (and keeping it too)](https://hackernoon.com/removing-that-ugly-focus-ring-and-keeping-it-too-6c8727fefcd2)

belbiy: @DmitryShmaliuk, we cant just remove the focus ring for mobile. There's an interesting prototype for [input modality](https://www.oreilly.com/ideas/proposing-css-input-modality) that handles the keyboard-mouse input switching.

But in reality, i think that we can improve user experience greatly by doing simple things first:

1. For search button, move the focus to the input field, so the user does not need to make extra tap. Consider the same behavior for every button – tapping a button can move focus the the area that button controls. If the button opens a shopping cart, for example, move focus inside that modal, etc. This can be a feature of the modal component actually, receiving focus on open, possibly even "trapping" the focus inside itself until closed.

2. Styling the `:focus` state of the buttons and other active elements that they look consistent and not "weird" when focused.

 @soumya-ashok, what do you think?

soumya-ashok: @belbiy Agree with your feedback. I'd just be cautious with choosing the default focus area in a scenario like the cart where there are multiple focusable elements and there aren't traditional form fields in the same way search or checkout has. We should choose a default that causes the least friction for the user in this scenario.

DmitryShmaliuk: @belbiy, i agree with you, and i will try to implement the your points.

packages/pwa-buildpack/src/WebpackTools/plugins/makeMagentoRootComponentsPlugin/index.js
```diff
@@ -110,16 +110,16 @@ class MagentoRootComponentsPlugin {
...
-${Object.entries(rootComponentImporters)
-    .map(entry => entry.join(':'))
-    .join(',\n')}
-};
-const key = ${toRootComponentMapKey.toString()};
-export default function fetchRootComponent(type, variant = 'default') {
-    return rootComponentsMap[key(type, variant)]().then(m => m.default || m);
-};
-`;
+            const rootComponentsMap = {
```
sirugh: Template strings will contain tabs/spaces. That's probably why this was not indented. We may want to undo this and add a `eslint ignore` or something along those lines here.

PR #1029, opened 2019-03-17
Fix for CMS that may not have .html extensions

danrcoull:
Error: 
Cms pages without .html in the url key will never resolve
Any Other Pages without .html wont resolve

Fix:

Add a check count, and try and resolve both .html and not .html urls


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1029) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1029) before we can accept your contribution.<br/>

packages/peregrine/src/Router/resolveUnknownRoute.js
```diff
@@ -86,8 +86,10 @@ function remotelyResolveRoute(opts) {
  * @param {{ route: string, apiBase: string}} opts
  * @returns {Promise<{type: "PRODUCT" | "CATEGORY" | "CMS_PAGE"}>}
  */
+let checkCount = 0;
 function fetchRoute(opts) {
     const url = new URL('/graphql', opts.apiBase);
+    console.log(opts);
```
sirugh: Please remove this log.

packages/peregrine/src/Router/resolveUnknownRoute.js
```diff
@@ -107,6 +109,11 @@ function fetchRoute(opts) {
     })
         .then(res => res.json())
         .then(res => {
+            if(res.urlResolver == null && checkCount < 1) {
```
sirugh: It would be better to put this "retry" logic at the point of the invocation (ie. within `remotelyResolveRoute`) so we're not maintaining a global variable. You could do something similar to the following:

```js
let resolved = await fetchRoute(opts)
if (!resolved) {
  opts.route.replace('.html', '');
  resolved = await fetchRoute(opts)
}
return resolved;
```

I'm not actually sure if that would return something or throw... I leave it to you to figure out :P

danrcoull: Cheers for the idea, i dropped the current way and added a async await function, with a check for html extension to make sure it doesnt infiante loop

supernova-at:
Related to and maybe dependent on #936 .

packages/peregrine/src/Router/resolveUnknownRoute.js
```diff
@@ -77,10 +77,28 @@ function remotelyResolveRoute(opts) {
...
+ * @param {{ route: string, apiBase: string}} opts
+ * @returns {Promise<resolve>}
+ */
+async function checkRoute(opts)
+    {
+      let resolve = await fetchRoute(opts);
+      if (resolve === null) {
+        if (opts.route.endsWith('.html')) {
+          opts.route = opts.route.replace('.html', '');
+          resolve = await fetchRoute(opts)
```
zetlen: Let's say the user clicks on an external ink to `venia.com/accessories.html`. If that catalog page no longer exists, this function will attempt to look up the entity at `accessories`. If that's a CMS page, it will display the CMS page without changing the URL in the browser location. Is this what we want?

I think there may be a better place to remove the URL key than this.

danrcoull: Catalog categories have .html  aswell, so unless a cms page is called /checkout then thats the only time i think this wont work.

zetlen: What if there used to be an `/accessories.html`, but that is no longer a category, and instead there is a CMS page called `/accessories`? That's where I'm worried about a bug. This may need to just redirect instead of silently removing the `.html` from the active route. Am I missing something?

ericerway:
@zetlen can you take a look in review? Thanks!

zetlen:
Gotta close this in favor of UPWARD-driven and GraphQL `storeInfo` driven solutions, since the `.html` suffix is configurable per Magento instance.

PR #1055, opened 2019-03-24
[BUG] Thumbnails are requested on Venia PDP in mobile

### LINKED ISSUE #1041: [bug]: Thumbnails are requested on Venia PDP in mobile
**Describe the bug**
In the Venia Storefront in a mobile view the page requests the thumbnail images unnecessarily.

In the following screenshot you can see the displayed image (640px) but you also can see the two thumbnails (240px) being requested.
[Image from Gyazo](https://gyazo.com/9be2b5a1a6911bc062c9a9884dc3a683)

**To Reproduce**
1. Open a PDP in mobile view with the network tab open.
2. See that large (~640) and small (~240) images are requested.

**Expected behavior**
Only the necessary images are requested. On Desktop that would be large & small images but on mobile that should only be the large images.

 - Device: Macbook Pro
 - OS: Mojave
 - Browser: Chrome
 - Magento Version: 2.3.1
 - NPM version `npm -v` 6.7.0
 - Node Version `node -v` 11.6.0

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

**Possible solutions**
Avoid mounting thumbnail components when viewport is below a certain width.

DmitryShmaliuk:
## Description
In the Venia Storefront in a mobile view the page requests the thumbnail images unnecessarily.

In the following screenshot you can see the displayed image (640px) but you also can see the two thumbnails (240px) being requested.

## To Reproduce

1. Open a PDP in mobile view with the network tab open.
2. See that only the large (~640) image is requested.

## Related Issue

Closes #1041.

## Motivation and Context

Now on mobile devices, only the necessary images are requested. (640px) On Desktop that would be large & small images but on mobile that should only be the large images. (240px, 640px)

### **Mobile view:** 

Screenshot from 2019-03-24 09-54-20

### **Desktop view:**

Screenshot from 2019-03-24 09-54-34

**BUG**

**pkg: venia-concept**

## Checklist:


DmitryShmaliuk:
Hi guys, in this issue there is another solution, we can block requested from thumbnails image only mobile devices, or will be left everything as is now.

DmitryShmaliuk:
@sirugh okay. I'll fix the decision, taking into account your comment.

jimbo:
@DmitryShmaliuk Thanks for taking a stab at this. I concur with @sirugh that our goal should probably be to avoid rendering the `img` element on mobile, given that our CSS just hides it.

DmitryShmaliuk:
@sirugh I have a small problem, can you help me? When I open the Home page in PWA, I see the next error: `Data Fetch Error: GraphQL error: Field "small_image" of type "String" must not have a sub selection.`, but in my branch, changes in only one file, it is `thumbnail.js`. I don't know what the problem is, and work on an issue, stopped.

supernova-at:
> @sirugh and @jimbo I have a small problem, can you help me? When I open the Home page in PWA, I see the next error: Data Fetch Error: GraphQL error: Field "small_image" of type "String" must not have a sub selection., but in my branch, changes in only one file, it is thumbnail.js. I don't know what the problem is, and work on an issue, stopped.

@DmitryShmaliuk are you still running into this problem?

DmitryShmaliuk:
@supernova-at your recommendation, not works. Problem is not solved. I will try reinstalling all project.

tjwiebell:
@DmitryShmaliuk - that's a backend mismatch error; try replacing your `.env` with `.env.dist`, which should have the correct backend url.

sirugh:
@DmitryShmaliuk how are things going? We'd love to get this PR wrapped up. If you need any help, reach out to me on the public slack and we can work through it.

packages/venia-concept/src/components/ProductImageCarousel/thumbnail.js
```diff
@@ -20,6 +20,12 @@ class Thumbnail extends Component {
         itemIndex: PropTypes.number,
         onClickHandler: PropTypes.func.isRequired
     };
+    // TODO: When we implement hooks/mergeClasses from #1078 we can replace this
+    // with a useWindowSize hook and rerender when the width changes across the
+    // breakpoint. This is important when changing from portrait to landscape.
+    isDesktop = () => {
+        return window.innerWidth > 640;
```
tjwiebell: The media rule to display thumbnails is `@media (min-width: 1024px)`, so this still would request images and not display for widths between 641-1023.

```suggestion
        return window.innerWidth >= 1024;
```

sirugh: I think there is some confusion around this component. The thumbnails are actually dual mode. When the window width is > 640 we display actual thumbnail images. when the window is less than 640 we still display a thumbnail component but it's just the icons below the image:

[Image from Gyazo](https://gyazo.com/d743ffc0ef20db6d357ca3a7599ca577)

sirugh: A more long term solution that I don't think we should get into here would be to actually refactor these into separate components.

sirugh: Ok I was wrong -- 1024 is correct, although we still should probably separate these out eventually, IMO :)

PR #1080, opened 2019-03-27
[Documentation] hello upward tutorial

### LINKED ISSUE #1046: [Documentation] Setting up a simple UPWARD server
This beginner-level tutorial will guide the reader through the process of creating a simple UPWARD server. It will start by using the sample echo configuration found in the UPWARD-spec package.

Things to cover:

* How to import the UPWARD-js server
* Defining the UPWARD-spec file for the UPWARD-js server
* Showing the various ways of using the UPWARD-js library to get a server up and running

jcalcaben:
## Description

Creates a three part tutorial that show a "from scratch" approach of creating a web application backed by an UPWARD server.

## Related Issue

Closes #1046 

## Motivation and Context

There have been a lot of questions on how people can get started from scratch.
It also helps to teach basic UPWARD concepts.

## Verification Steps

React project created following the steps described in the tutorials.

## How Has This Been Tested?

n/a

n/a

documentation
devdocs

## Checklist:

- [ ] I have proposed a change level by adding one of the `version: Major`, `version: Minor`, or `version: Patch` labels based on the defined [Public API](https://magento-research.github.io/pwa-studio/technologies/versioning/).


pwa-devdocs/src/tutorials/hello-upward/using-template-resolver/index.md
```diff
@@ -0,0 +1,106 @@
...
++               title:
++                   resolver: inline
++                   inline: 'This is the page title!'
+```
+
+This new code replaces the InlineResolver with a TemplateResolver in the response body.
+This TemplateResolver configuration sets the rendering engine to `mustache`, since the templates created previously are in Mustache format.
+It also provides the `title` variable to the context during template render.
+
+Now, when you start the server and navigate to the root or `/hello-world` path, you get an actual HTML webpage instead of text.
```
supernova-at: The only visible difference is the page title, which appears in the browser tab.

Perhaps `hello-world.mst` should wrap the "Hello World!" text in an `<h1>` or similar to very clearly differentiate it from the inline example.

And / or we could add a line here that tells readers to "View Source" to see the actual HTML.

pwa-devdocs/src/tutorials/hello-upward/using-template-resolver/index.md
```diff
@@ -0,0 +1,106 @@
...
+This template partial closes the `head` tag and begins the `body` tag of the HTML response.
+
+**`close-document.mst`**
+
+```mustache
+  </body>
+</html>
+```
+
+This template partial closes the `body` and `html` tag for the HTML response.
```
supernova-at: I think the partials are overkill for this tutorial.

Since this isn't a tutorial on Mustache, I think the `hello-world.mst` template could just contain all of this.

pwa-devdocs/src/tutorials/hello-upward/adding-react/index.md
```diff
@@ -0,0 +1,176 @@
+---
+title: Adding React
+---
+
+This tutorial teaches you how to add webpack to your UPWARD project and use it to create a React application.
```
supernova-at: I think this topic can be made much simpler by getting rid of all the "extras" around React.

Since [React doesn't actually need JSX](https://reactjs.org/docs/react-without-jsx.html) you can get rid of the dependency on `babel` and `webpack` here.

Check it out:

## `hello-world.mst`

```jsx
{{> templates/open-document}}

<title>{{ title }}</title>
<script src="https://cdnjs.cloudflare.com/ajax/libs/react/16.8.6/umd/react.production.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.8.6/umd/react-dom.production.min.js"></script>

{{> templates/open-body}}

<div id="root">Loading...</div>

<script defer src="/app.js"></script>

{{> templates/close-document}}
```

Notice that the template is pulling in React and React DOM from CDNs. This eliminates the need to install them locally.

## `app.js`

```js
const App = () => {
    return React.createElement(
        'h1',
        null,
        'Hello World from React!'
    );
};

ReactDOM.render(
    React.createElement(App),
);
```

Notice that we aren't using JSX here and instead using the `React.createElement` function directly. This is [what Babel does](https://babeljs.io/en/repl) for you.

The only thing left to do is fix up the UPWARD spec a bit:

```yaml
status: response.status
headers: response.headers
body: response.body

response:
  resolver: conditional
  when:
        pattern: '^/?$'
        use: helloWorld
        pattern: '^/hello-world/?$'
        use: helloWorld
        pattern: '^/app.js$'
        use: appScript
  default: notFound

helloWorld:
  inline:
      status:
          resolver: inline
          inline: 200
      headers:
          resolver: inline
          inline:
              content-type:
                  resolver: inline
                  inline: 'text/html'
      body:
          resolver: template
          engine: mustache
          template: './templates/hello-world.mst'
          provide:
            title:
              resolver: inline
              inline: 'This is the page title!'
notFound:
    inline:
        status:
            resolver: inline
            inline: 404
        headers:
            resolver: inline
            inline:
                content-type:
                    resolver: inline
                    inline: 'text/string'
        body:
            resolver: inline
            inline: 'Page not found!'

appScript:
  inline:
    status:
      resolver: inline
      inline: 200
    headers:
      resolver: inline
      inline:
        content-type:
          resolver: inline
          inline: 'application/javascript'
    body:
      resolver: file
      file:
        resolver: inline
        inline: './app.js'
      encoding:
        resolver: inline
        inline: 'utf-8'
```
<hr />

Since this is an UPWARD tutorial, I think this greatly simplifies the topic and introduces readers to the `FileResolver` more quickly.

We probably want to include a paragraph about how production code should use webpack and babel but I don't think they need to be introduced in this tutorial.

pwa-devdocs/src/_scss/src/_code-block-style-overrides.scss
```diff
@@ -0,0 +1,13 @@
+.highlight .gd,
+.highlight .gi {
```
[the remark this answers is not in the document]
jcalcaben: These classnames come from the devdocs theme and/or jekyll itself, so we don't have control over the names. 😿

pwa-devdocs/src/tutorials/hello-upward/using-template-resolver/index.md
jcalcaben: I agree that partials add some complexity for a simple tutorial. However, this pattern is used in the Venia implementation. Readers will undoubtedly draw comparisons between what's in the Venia source and how we're telling them to do something.

pwa-devdocs/src/tutorials/hello-upward/adding-react/index.md
jcalcaben: Even though it's billed as an UPWARD tutorial, I also wanted the end result of doing this tutorial to be an environment that the reader can use as a starting point for their PWA Studio projects. This is something the community has been asking us for.

It's true that React doesn't require JSX, but most examples out there use JSX and ES6 syntax. I think it would be a bad experience for developers if we introduce how something is done in the tutorial and promote something very different in the implementation code.

pwa-devdocs/src/tutorials/hello-upward/adding-react/index.md
```diff
@@ -0,0 +1,176 @@
...
+
+Create a `src` directory in your project's root directory and a `hello-world.js` file inside that directory with the following content:
+
+```jsx
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+class App extends React.Component {
+    render() {
+        return <h1>Hello World!</h1>;
```
supernova-at: What about "Hello World from React!" ?

pwa-devdocs/src/tutorials/hello-upward/using-template-resolver/index.md
```diff
@@ -0,0 +1,107 @@
...
+{% raw %}
+
+```mustache
+{{> templates/open-document}}
+
+  <title>{{ title }}</title>
+
+  {{> templates/open-body}}
+
+    <b>Hello World!</b>
```
supernova-at: How about "Hello World from a Template!" ?

PR #1088, opened 2019-03-29
[dev]: Update eslint config version

sirugh:
This PR Updates our magento-eslint config dep to v1.4.0 which adds the `prefer-const` rule. This rule is fairly straightforward and you can see what it does in the code. That being said there were a few places it did not agree with (array destructuring) so I just ignored those lines.

### DIFF
--- a/packages/upward-js/lib/resolvers/TemplateResolver.js
+++ b/packages/upward-js/lib/resolvers/TemplateResolver.js
@@ -63,15 +63,13 @@ module.exports = class TemplateResolver extends AbstractResolver {
             providePromise
         ];
 
-        // eslint-disable-next-line prefer-const
-        let [engine, template, rootEntries] = await Promise.all(toResolve);
+        const [engineName, template, rootEntries] = await Promise.all(
+            toResolve
+        );
         debug('template retrieved, "%s"', template);
         debug('rootEntries retrieved, %o', rootEntries.map(([name]) => name));
 
-        if (!engine) {
-            engine = 'mustache';
-        }
-
+        const engine = engineName || 'mustache';
         const Engine = Engines[engine];
 
         if (!Engine) {
--- a/packages/upward-js/lib/resolvers/TemplateResolver.js
+++ b/packages/upward-js/lib/resolvers/TemplateResolver.js
@@ -63,6 +63,7 @@ module.exports = class TemplateResolver extends AbstractResolver {
             providePromise
         ];
 
+        // eslint-disable-next-line prefer-const
         let [engine, template, rootEntries] = await Promise.all(toResolve);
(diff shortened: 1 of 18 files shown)


packages/venia-concept/src/components/PurchaseHistoryPage/PurchaseHistory/purchaseHistoryContainer.js
```diff
@@ -3,14 +3,10 @@ import { transformItems } from 'src/selectors/purchaseHistory';
 import PurchaseHistory from './purchaseHistory';
 import actions, { getPurchaseHistory } from 'src/actions/purchaseHistory';
 
-const mapStateToProps = ({ purchaseHistory }) => {
-    let { isFetching, items } = purchaseHistory;
-
-    items = transformItems(items);
-
+const mapStateToProps = ({ purchaseHistory: { isFetching, items } }) => {
```
sirugh: Please double check my changes here!

jimbo: Nested destructuring is too difficult to read for too little benefit. I recommend reverting this change.

packages/upward-js/lib/resolvers/TemplateResolver.js
```diff
@@ -63,6 +63,7 @@ module.exports = class TemplateResolver extends AbstractResolver {
             providePromise
         ];
 
+        // eslint-disable-next-line prefer-const
```
jimbo: Instead of using this comment, these lines should probably be rewritten.

```es6
const [engineName, template, rootEntries] = await Promise.all(toResolve)
const engine = engineName || "mustache"
```

packages/venia-concept/src/components/PurchaseHistoryPage/PurchaseHistory/purchaseHistoryContainer.js
supernova-at: Resolved in https://github.com/magento-research/pwa-studio/pull/1088/commits/c12f9179443d38f9c434dfbfda04872e2033dad6.

packages/upward-js/lib/resolvers/TemplateResolver.js
supernova-at: Resolved in https://github.com/magento-research/pwa-studio/pull/1088/commits/c12f9179443d38f9c434dfbfda04872e2033dad6.

supernova-at:
PR Updated:
  * Updates based on code review feedback
  * Updates to failing tests due to assignment to `const` variable 🎉

jimbo:
@supernova-at Thanks for taking this over and making those changes. 👍

PR #1097, opened 2019-04-01
Removes ability to add configurable items to cart until options are chosen

### LINKED ISSUE #640: [Bug]: Error message for configurable product

- [X] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [X] Bug
- [ ] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

### Environment

## Description
No any error messages displayed if user try to add configurable product in a cart with messing configuration. Flow:
1) navigate to: https://magento-venia.now.sh/helena-cardigan.html
2) choose color configuration only
3) click `add to cart` button
4) cart is empty:
image
and request was failed
- general info: 
image
- payload: 
image
- response: 
image

## Expected result:
on the step 3 show error message for missing configuration 

## Possible solutions:
- Implement front-end validation for configurable product
- Implement back-end validation for configurable product (unsafe to show stack-trace in response, hacker can use this information and use them)

supernova-at:
## Description

This PR disables the "Add to Cart" button on configurable product description pages (PDPs) until all product options have been selected.

## Related Issue

Closes #640 .

### Configurable Products
1. Select a configurable product (ex: https://magento-venia-concept-cj55l.local.pwadev:8582/valeria-two-layer-tank.html).
1. See that the "Add to Cart" button remains disabled until you make selections for each product option (color, size, etc.).

### Non-Configurable Products
1. Select a non-configurable product (ex: https://magento-venia-concept-cj55l.local.pwadev:8582/augusta-necklace.html).
1. See that the "Add to Cart" button is always enabled. 

## How Have YOU Tested this?

I loaded up a configurable and non-configurable product, as described above.

`yarn test` passes.

Screen Shot 2019-04-01 at 2 24 05 PM

## Checklist:

## Further Details / Risks / Notes / Etc.

This is perhaps the simplest solution to this problem. Some other options would be:

### Client-side, per-field Validation

Other forms in the app (ex: Address Form) have the ability to perform client-side, per-field validation and show error messages inline on the form.

Unfortunately the PDP would need to be heavily refactored to allow for this, and a solution for the color swatches would need to be included as part of this (changing them to radio buttons instead of actual buttons).

**Why I didn't go this route**:
We would still want to stick to Venia's precedent and pattern of not enabling the "submit" button of the form until all required fields are filled out, and since there is no additional client-side validation on top of _this field was filled in_, the refactor seemed to be too much work for not enough payoff.

Edit: turns out this isn't much of a precedent or pattern. The checkout flow doesn't enable the submit button until billing, payment, and shipping are filled in but other forms have their submit buttons enabled and hooked up to client-side validation first.

### Capturing the Error in Redux

Currently our backstop reducer handles the error from the server that results from attempting to add a configurable item to your cart without specifying options.

Ideally, the `cart` reducer handles this itself and passes the error back up to the UI (via Redux and `mapStateToProps`) so that the UI can show a nice, in-context error message.
 
**Why I didn't go this route**:
By disabling the "Add to Cart" button until all options are selected, we sidestep this problem.

Because client-side validation can be circumvented, some consideration should be given to the question of "what happens if the call to the server is made some other way?", in which case the backstop reducer would still kick in and the original issue would persist.

But I don't think solving for this case is worth it: if someone intentionally sidesteps our UI to make a server call, they shouldn't expect a nice clean UI if it fails.

### DIFF
--- a/packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
+++ b/packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
@@ -14,11 +14,9 @@ import RichText from 'src/components/RichText';
 import defaultClasses from './productFullDetail.css';
 import appendOptionsToPayload from 'src/util/appendOptionsToPayload';
 import findMatchingVariant from 'src/util/findMatchingProductVariant';
+import isProductConfigurable from 'src/util/isProductConfigurable';
 
 const Options = React.lazy(() => import('../ProductOptions'));
-
-const productIsConfigurable = product =>
-    Array.isArray(product.configurable_options);
 
 class ProductFullDetail extends Component {
     static propTypes = {
@@ -38,6 +36,7 @@ class ProductFullDetail extends Component {
             title: string
         }),
         product: shape({
+            __typename: string,
             id: number,
             sku: string.isRequired,
             price: shape({
@@ -66,7 +65,7 @@ class ProductFullDetail extends Component {
         const optionCodes = new Map(state.optionCodes);
 
         // if this is a simple product, do nothing
-        if (!productIsConfigurable(props.product)) {
+        if (!isProductConfigurable(props.product)) {
             return null;
         }
 
@@ -90,10 +89,7 @@ class ProductFullDetail extends Component {
         const { props, state } = this;
         const { optionSelections, quantity, optionCodes } = state;
         const { addToCart, product } = props;
-        const isConfigurable = productIsConfigurable(product);
-        const productType = isConfigurable
-            ? 'ConfigurableProduct'
-            : 'SimpleProduct';
+        const productType = product.__typename;
 
         const payload = {
             item: product,
@@ -101,7 +97,8 @@ class ProductFullDetail extends Component {
             quantity
         };
 
-        if (productType === 'ConfigurableProduct') {
+        if (isProductConfigurable(product)) {
+            console.log('appending options to payload');
             appendOptionsToPayload(payload, optionSelections, optionCodes);
         }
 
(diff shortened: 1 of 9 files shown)


packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -17,6 +17,9 @@ import findMatchingVariant from 'src/util/findMatchingProductVariant';
 
 const Options = React.lazy(() => import('../ProductOptions'));
 
+const productIsConfigurable = product =>
```
tjwiebell: I know you grabbed this logic from the existing implementation, but I'm curious if we should leverage `__typename` from the query response instead. Documentation is a little light on why this field is in the response even though we're not explicitly requesting it, but it's there, and seems like the preferred pattern when you're dealing with interfaces. Think this might look cleaner?

`product.__typename === 'ConfigurableProduct'`

tjwiebell: The reason this appears to be "magic" is because of Apollo, it automatically adds `__typename` to queries for its own caching purposes. We may want to explicitly add it to the query and proptype so that's more clear.

supernova-at: Nice! I'll give it a shot - agreed it's a lot nicer.

supernova-at:
PR Updated:
  * Extracts out all `isProductConfigurable` logic to one utility function
  * Updates appropriate graphql queries to explicitly ask for `__typename`

For some reason I'm getting a weird error where I can't successfully add a configurable item to my cart due to mismatched `option_id`s. But I'm also getting it on `develop`, and I don't see any changes here that would cause it, so I'm thinking the problem is not in this PR.

Please do verify that you can successfully add a configurable item to your cart.

PR #1105, opened 2019-04-02
Prevent Rest api call on minicart sidebar if cart is empty

### LINKED ISSUE #1104: Mini-cart Sidebar: If cart is empty and clicked on mini-cart, execute unwanted Rest API call.
**Describe the bug**
I've noticed extra or unwanted Rest API call executed on mini-cart open action even cart is empty.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the pwa-studio storefront.
2. Open developer console and go to the network tab.
3. Click on the mini-cart icon at the top right section.
4. See you will find rest API call even cart is empty.

**Expected behavior**
If the cart is empty, the rest API call shouldn't be executed.

**Screenshots**
2019_04_02_Selection_002

 - Device: iPhone6/7/8, PC
 - OS: Ubuntu
 - Browser Chrome
 - Magento Version 2.3.1
 - PWA Studio Version: 2.1
 - NPM version 6.4.1
 - Node Version 10.15.2

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

vishal-7037:
## Description

I've prevented Rest API call if the cart is empty.

## Related Issue

Closes #1104.

## Verification Steps

1. Go to the pwa-studio storefront.
2. Open developer console and go to the network tab.
3. Click on the mini-cart icon at the top right section.
4. See you will find rest API call even cart is empty.

**Before Fixed:** 
2019_04_02_Selection_002

**After Fixed:**
2019_04_02_Selection_001

## Checklist:


packages/venia-concept/src/actions/cart/asyncActions.js
```diff
@@ -402,10 +403,11 @@ export const toggleCart = () =>
...
         // otherwise open the cart and load its contents
-        await Promise.all([
-            dispatch(toggleDrawer('cart')),
-            dispatch(getCartDetails())
-        ]);
+        cartActions = [dispatch(toggleDrawer('cart'))];
+        if (cart.details !== undefined && cart.details.items_count !== 0) {
+            cartActions = [dispatch(getCartDetails())];
+        }
+        await Promise.all(cartActions);
```
supernova-at: The previous code,

```
await Promise.all([
  dispatch(toggleDrawer('cart')),
  dispatch(getCartDetails())
]);
```

waited for both dispatch promises to resolve (or one of them to reject) before finishing execution and letting the thunk return.

The new code doesn't replicate that behavior, even when there is something in the cart. The following factors combine to make that the case:

**`cartActions` is reassigned**

We're trying to build `cartActions` up as an array of actions to take. By the time we get to the `Promise.all` (line 410), it should have one action (`toggleDrawer`) or two (`toggleDrawer` and `getCartDetails`), depending on the number of items in the cart.

Unfortunately, line 408 is _reassigning_ `cartActions` instead of adding to it, so the `Promise.all` is only ever passed one promise, never two.

That would be tricky to debug because

**`dispatch(toggleDrawer('cart'))` is executed immediately**

The statement `cartActions = [dispatch(toggleDrawer('cart'))];` (line 406) has the effect of `toggleDrawer` executing immediately. 

Combined with the previous bullet point, this results in this thunk **not** waiting for the `toggleDrawer` dispatch to complete, which doesn't match the previous implementation and likely wasn't an intended change (correct me if I'm wrong).

Currently this action doesn't take any time at all so we don't even notice it but imagine code gets added in the future that makes it take a few seconds to complete. The thunk would return and then _sometime later_ the `toggleDrawer` dispatch would complete. Instead, we don't want to return from this thunk until that action completes.

supernova-at: Note: It's fine that `dispatch(toggleDrawer('cart'))` is executed immediately, we just need to make sure we're passing that promise to `Promise.all`.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -897,6 +897,32 @@ describe('toggleCart', () => {
         expect(dispatch).not.toHaveBeenCalled();
     });
 
+    test('toggleCart thunk opens the drawer and cart is empty', async () => {
```
supernova-at:
```suggestion
    test('its thunk opens the drawer when cart is empty', async () => {
```

These tests are inside a `describe('toggleCart` block so you don't have to repeat the name of the function each time 👍

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -897,6 +897,32 @@ describe('toggleCart', () => {
...
+        getState.mockReturnValueOnce({
+            app: { drawer: 'cart' },
+            cart: { details: { items_count: 0 } }
+        });
+        await toggleCart()(...thunkArgs);
+
+        expect(dispatch).toHaveBeenNthCalledWith(1, expect.any(Function));
```
supernova-at: While `expect.any(Function)` is a perfectly valid test case, it isn't very friendly for the reader because there's no context or intention.

Either of the following would help out the next person to read this:

**Add a comment**
```js
// toggleDrawer
expect(dispatch).toHaveBeenNthCalledWith(1, expect.any(Function));
```

or **Give it a semantic name**

```js
const toggleDrawer = expect.any(Function);
expect(dispatch).toHaveBeenNthCalledWith(1, toggleDrawer);
```

We're working on making this more consistent throughout the tests 👍

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -897,6 +897,32 @@ describe('toggleCart', () => {
...
+        expect(dispatch).toHaveBeenCalledTimes(1);
+    });
+
+    test('toggleCart thunk opens the drawer and refreshes the cart', async () => {
```
supernova-at:
```suggestion
    test('its thunk opens the drawer and refreshes the cart when the cart has items', async () => {
```

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -897,6 +897,32 @@ describe('toggleCart', () => {
...
+        getState.mockReturnValueOnce({ app: { drawer: 'cart' }, cart: {} });
+        await toggleCart()(...thunkArgs);
+        expect(dispatch).toHaveBeenNthCalledWith(1, expect.any(Function));
+        expect(dispatch).toHaveBeenCalledTimes(1);
```
supernova-at: Let's split these assertions out into their own test. Something like "its thunk opens the drawer when the cart has no details".

packages/venia-concept/src/actions/cart/asyncActions.js
```diff
@@ -389,6 +389,7 @@ export const getCartDetails = (payload = {}) => {
 
 export const toggleCart = () =>
     async function thunk(dispatch, getState) {
+        let cartActions;
```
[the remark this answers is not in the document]
vishal-7037: I've removed declaration of `cartActions`

packages/venia-concept/src/actions/cart/asyncActions.js
```diff
@@ -402,10 +403,11 @@ export const toggleCart = () =>
         }
 
         // otherwise open the cart and load its contents
-        await Promise.all([
-            dispatch(toggleDrawer('cart')),
-            dispatch(getCartDetails())
-        ]);
+        cartActions = [dispatch(toggleDrawer('cart'))];
+        if (cart.details !== undefined && cart.details.items_count !== 0) {
```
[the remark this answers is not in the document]
vishal-7037: I've updated the changes with your suggestion

packages/venia-concept/src/actions/cart/asyncActions.js
vishal-7037: Okay,

For this, I've done some minor modification in the code, after modification I think the behaviour is replicated with previous code which you have mentioned in this.

Please see, I've tried to print both older and newer dispatch in the developer console.
2019_05_02_Selection_001

And In developer Console, It will return the same output.
2019_05_02_Selection_002

Please let me know your suggestion.

Thanks!

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
vishal-7037: I've applied suggested changes.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
vishal-7037: I've applied the suggested changes.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
vishal-7037: I've applied the suggested changes.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
vishal-7037: I've tried to split it but I have a little bit confused with test cases. So, Can you please help with this?

packages/venia-concept/src/actions/cart/asyncActions.js
supernova-at: Yep, you got it - thanks!

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -897,6 +897,36 @@ describe('toggleCart', () => {
...
+
+    test('its thunk opens the drawer when the cart has details', async () => {
+        getState.mockReturnValueOnce({
+            app: { drawer: 'cart' },
+            cart: { details: { items_count: 2 } }
+        });
+        await toggleCart()(...thunkArgs);
+        // toggleDrawer
+        expect(dispatch).toHaveBeenNthCalledWith(1, expect.any(Function));
+        expect(dispatch).toHaveBeenCalledTimes(1);
```
supernova-at: Shouldn't `dispatch` be called twice here? First to `toggleDrawer` and then to get the details?

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -922,12 +952,18 @@ describe('toggleCart', () => {
 
     test('its thunk opens the drawer and refreshes the cart', async () => {
         await toggleCart()(...thunkArgs);
+        // toggleDrawer
+        expect(dispatch).toHaveBeenNthCalledWith(1, expect.any(Function));
+        expect(dispatch).toHaveBeenCalledTimes(1);
```
supernova-at: Because you haven't mocked `getState` here, the call to `toggleCart` on line 954 will use the default mock return value declared at the top of the file:

```js
beforeAll(() => {
    getState.mockImplementation(() => ({
        app: { drawer: null },
        cart: { cartId: 'CART_ID' },
        user: { isSignedIn: false }
    }));
});
```

which makes these assertions a duplicate of your previous test case "its thunk opens the drawer when the cart has no details" (note that the `cart` above has no `details`). Therefore you can remove lines 954-957.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -922,12 +952,18 @@ describe('toggleCart', () => {
...
 
+        getState.mockReturnValueOnce({
+            app: { drawer: 'cart' },
+            cart: { details: { items_count: 2 } }
+        });
+        await toggleCart()(...thunkArgs);
+        // Refresh the cart
+        expect(dispatch).toHaveBeenNthCalledWith(2, expect.any(Function));
```
supernova-at: And because this setup and assertion duplicates your previous test "its thunk opens the drawer when the cart has items", you can drop this too.

Therefore you can drop this entire test 🎉

supernova-at:
Oh, one other thing that was mentioned that we want to be aware of here:

We want to preserve the functionality of having multiple browser tabs open and having the cart remain up to date on both.

That is to say, if I have browser tabs A and B open and I add an item to the cart in tab A, that item should show up in the cart in tab B.

Let's make sure we verify that functionality remains before approving this.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -897,6 +897,36 @@ describe('toggleCart', () => {
...
+
+    test('its thunk opens the drawer when the cart has no details', async () => {
+        getState.mockReturnValueOnce({ app: { drawer: 'cart' }, cart: {} });
+        await toggleCart()(...thunkArgs);
+        // toggleDrawer
+        expect(dispatch).toHaveBeenNthCalledWith(1, expect.any(Function));
+        expect(dispatch).toHaveBeenCalledTimes(1);
+    });
+
+    test('its thunk opens the drawer when the cart has details', async () => {
```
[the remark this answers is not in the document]
vishal-7037: Committed Suggested changes.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
vishal-7037: I've tried to fixed it.please review and let me know.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
vishal-7037: I've tried to fixed it.please review and let me know.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
vishal-7037: I've tried to fixed it.please review and let me know.

vishal-7037:
> Oh, one other thing that was mentioned that we want to be aware of here:
> 
> We want to preserve the functionality of having multiple browser tabs open and having the cart remain up to date on both.
> 
> In other words: if I have browser tabs A and B open and I add an item to the cart in tab A, that item should show up in the cart in tab B.
> 
> Let's make sure we verify that functionality remains before approving this change.

Hi, @supernova-at,

I've committed suggested changes and tried to fixed test cases. Please review it and let me know.

Also regarding the above concerned, I've checked the same case and seems working fine.
Can you please check it from your side?

Thank you.

packages/venia-concept/src/actions/cart/__tests__/asyncActions.spec.js
supernova-at: Hi @vishal-7037, I found the issue:

Lines 400-402 in `cart/asyncActions.js` are as follows:

```js
        if (app.drawer === 'cart') {
            return dispatch(closeDrawer());
        }
```

In each of these tests we are mocking `app.drawer` to equal "cart", so this conditional is getting hit and dispatching a `closeDrawer` action.

1. Please update these tests to set `app.drawer` to something other than "cart" (`null` would work).
2. Please update this specific test, as we expect `dispatch` to be called twice (`toggleDrawer` and `getCartDetails`).

Thanks!

supernova-at:
> Also regarding the above concerned, I've checked the same case and seems working fine.
Can you please check it from your side?

Hmm, it didn't work for me.  Here are the steps I performed:

1. `yarn watch:venia`
2. Open two browser tabs to local Venia site
3. Ensure cart has no items (may need to remove any items & refresh both tabs)
4. Add an item to the cart in Tab 1
5. Click the shopping cart icon to toggle the cart open in Tab 2 
Screen Shot 2019-05-06 at 2 22 19 PM
6. See that the item added in Step 4 does not appear in the cart

In Step 6 it is expected that the item should show up.

Please let me know if you can reproduce.

vishal-7037:
Hi @supernova-at,

After the merge with `develop` branch, I've got warnings and not able to work ahead.
I've ran `yarn install` command and update all packages also and then ran `yarn watch:venia` command.

and getting following warning.

2019_05_07_warning

2019_05_07_error_after_mergeing_with_develop

Please help me out so I can start work ahead.

Thanks!

supernova-at:
> Please help me out so I can start work ahead.

Ah, yes - we just merged a PR that changed some dependencies. Please perform the following commands and let me know if it works for you:

```
yarn clean:all
yarn install
yarn build
yarn watch:venia
```

supernova-at:
The team has decided to close this ticket as it breaks the multiple tab scenario described above.

Thanks for taking the initiative on this though!

PR #1153, opened 2019-04-22
Passes email, firstname, and lastname to checkout receipt properly

### LINKED ISSUE #830: After placing an order if user clicks on Create account button then create account page Email, First and last name field contains value as Null instead of values used during order placement.
Description:
Place an order as guest and click on "Create Account" button > Create account page displays with "null" as values for Email, First and Last name.

1. Go to Venia storefront as guest and Place an order.
2. Click on Create account button on order confirmation page.

Expected behavior:
Create account page should be displayed with pre-filled email, first and last name which are user for order placement.

Actual :
Create account page displayed with null values for email, first and last name.

Screen:

supernova-at:
## Description

After a successful guest checkout, users are prompted to create an account:

Screen Shot 2019-04-22 at 3 10 43 PM

The design is that their first name, last name, and email would be pre-set for them on the "Create Account" page, with that information being taken from the previous order's details:

Screen Shot 2019-04-22 at 3 14 23 PM

Unfortunately, after being sent to the "Create Account" page, these fields were filled with `null`s:

image

This PR fixes up some internal wiring so that those fields are once again pre-populated with the correct data (not `null`).

## Related Issue

Closes #830 .

## Verification Steps

Perform the following steps twice: Once for billing address same as shipping address, and once with billing address different than shipping address:

Screen Shot 2019-04-22 at 3 10 28 PM

1. Complete the checkout process as a guest user and submit your order, making a note of the values you enter for shipping address first name, last name, and email fields.
1. On the receipt screen, click to `Create an Account`
1. See that the first name, last name, and email fields are populated with the data entered in step 1.

📝 We don't have to test this for authenticated users because the "Create Account" button does not exist for them on the checkout receipt (they already have an account):

Screen Shot 2019-04-24 at 1 23 04 PM

## How Have YOU Tested this?

1. The steps above
1. `yarn test`

## Checklist:

### DIFF
--- a/packages/venia-concept/src/actions/checkout/__tests__/asyncActions.spec.js
+++ b/packages/venia-concept/src/actions/checkout/__tests__/asyncActions.spec.js
@@ -512,7 +512,7 @@ describe('submitOrder', () => {
             2,
             checkoutReceiptActions.setOrderInformation({
                 id: response,
-                billing_address: address
+                shipping_address: mockShippingAddress
             })
         );
         expect(dispatch).toHaveBeenNthCalledWith(
@@ -540,9 +540,7 @@ describe('submitOrder', () => {
             directory: { countries }
         };
 
-        getState
-            .mockImplementationOnce(() => mockState)
-            .mockImplementationOnce(() => mockState);
+        getState.mockImplementationOnce(() => mockState);
 
         mockGetItem
             .mockImplementationOnce(
@@ -562,7 +560,7 @@ describe('submitOrder', () => {
             2,
             checkoutReceiptActions.setOrderInformation({
                 id: response,
-                billing_address: address
+                shipping_address: mockShippingAddress
             })
         );
         expect(dispatch).toHaveBeenNthCalledWith(
--- a/packages/venia-concept/src/actions/checkout/__tests__/asyncActions.spec.js
+++ b/packages/venia-concept/src/actions/checkout/__tests__/asyncActions.spec.js
@@ -53,8 +53,9 @@ const paymentMethod = {
 
 beforeAll(() => {
     getState.mockImplementation(() => ({
-        cart: { guestCartId: 'GUEST_CART_ID' },
-        directory: { countries }
+        cart: { cartId: 'CART_ID' },
+        directory: { countries },
+        user: { isSignedIn: false }
     }));
 });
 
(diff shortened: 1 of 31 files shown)


packages/venia-concept/src/actions/checkout/__tests__/asyncActions.spec.js
```diff
@@ -540,9 +540,7 @@ describe('submitOrder', () => {
             directory: { countries }
         };
 
-        getState
-            .mockImplementationOnce(() => mockState)
-            .mockImplementationOnce(() => mockState);
```
supernova-at: The module under test no longer calls `getState` a second time.

packages/venia-concept/src/actions/checkout/asyncActions.js
```diff
@@ -244,9 +243,10 @@ export const submitOrder = () =>
             );
 
             dispatch(
-                checkoutReceiptActions.setOrderInformation(
-                    getOrderInformation(getState(), response)
-                )
+                checkoutReceiptActions.setOrderInformation({
+                    id: response,
+                    shipping_address
+                })
```
supernova-at: We have all the information we need to make this call in scope here, no need to go to the store to get it.

packages/venia-concept/src/selectors/cart.js
```diff
@@ -1,11 +1,3 @@
-export const getOrderInformation = (
-    { cart: { details: { billing_address } = {} } },
-    orderId
-) => ({
-    id: orderId,
-    billing_address
-});
```
supernova-at: This function was only used once, and that one place has been deleted. This can be removed too.

packages/venia-concept/src/actions/checkout/asyncActions.js
supernova-at: Importantly, this has changed to `shipping_address` instead of `billing_address` because that is where `firstname`, `lastname`, and `email` live.

If the user is creating an account it does make more sense to seed the Create Account form with the first name, last name, and email address from their billing address (as opposed to the shipping address, which might be different). 

Because the billing address does not contain this information, and the user can change the values on the Create Account form if they so desire, I have created a new ticket to address the desired design: #1154. I considered it out of scope for this bug ticket.

tjwiebell: I agree with the creation of the ticket to add those additional fields, but I don't see why you had to change the receipt to pull values from `shipping_address` in this scope. `email, firstname, lastname, telephone` are copied from shipping to billing, so the values are there, and it'll be one less thing they need to do in scope of #1154.

I quickly tested and didn't run into a problem. Was there a use case that necessitated that change?

supernova-at: Good points, I'll play with this.

supernova-at:
PR Updated: 
  * Reverts change that sourced first name, last name, and email from shipping address

PR #864, opened 2019-02-07
Make dependencies which require native extensions optional

jimbo:
## Description

Some project dependencies have native extensions written in other languages. These dependencies require [`node-gyp`](https://github.com/nodejs/node-gyp#installation), which in turn requires Python and a C/C++ compiler (msbuild.exe on Windows, gcc or clang on *nixes, etc).

Since these dependencies are required, any user or environment without these tools, such as a minimal Docker container or a fresh Windows/OSX install, will fail to install the project dependencies.

The affected dependencies are:
- `shrink-ray-current`, a middleware for compressing HTTP responses, used by `upward-js`
- `sharp`, a library for image compression and manipulation, used by the image optimization middleware in `pwa-buildpack`
- `bundlesize`, a tool for reporting changes in build asset filesize to a GitHub integration reporter

Since none of the above features are strictly critical for basic operations, this PR moves them to `optionalDependencies`, so that npm/yarn will not bail out when it can't build them.

When these optional dependencies are not installed:
- In a CI container which runs `yarn run bundlesize`, that step will fail. CircleCI NodeJS build containers do not appear to have this problem.
- In a development or production environment running the NodeJS server, HTTP responses and catalog images won't be compressed. In dev mode, this will affect performance measurements negatively. In production mode, it's not acceptable; some external proxy would have to do compression instead.

## Related Issue

#820

## How Has This Been Tested?

`yarn test`

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] I have linked an issue to this PR.
- [x] I have indicated the change type and relevant package(s).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] All CI checks are green (linting, build/deploy, etc).
- [ ] At least one core contributor has approved this PR.

PR #1165, opened 2019-04-24
ci: Adopt now.sh serverless and staging URLs

### LINKED ISSUE #1163: Deploy 2.1.0 to veniapwa.com
Right now we have this simple ass now for github integration. It deploys the default branch. We deploy develop to `venianow.pwa.com`. Unacceptable, we need to deploy 2.1.0.

Additionally, we'll need to ensure any and all related items to ensure demo site performance during upcoming Imagine are identified and either addressed as a part of this ticket or as separate items. @zetlen to weigh in on any additional effort needed.

zetlen:
## Description
Enable independent production/staging deploys using the [Now.sh staging URL feature](https://zeit.co/docs/v2/domains-and-aliases/aliasing-a-deployment?query=staging).

### New staging architecture
The staging URL feature enables predictable URLs for each deployed branch. This means that our master branch will always deploy to `venia-git-master.magento-research.now.sh`.

To use this new feature, we have to start using Now v2. This is a significant change to our staging server

- Our UPWARD server will now run in a [lambda](https://zeit.co/docs/v2/deployments/concepts/lambdas), with no guarantees of a persistent Node process between requests. Fortunately, UPWARD is designed to be stateless.
- Our build process is now specified using two [Now.sh v2 "Builders"](https://zeit.co/docs/v2/deployments/builders/overview/):
  - the [`@now/node`](https://zeit.co/docs/v2/deployments/official-builders/node-js-now-node/) builder for *installing, building, and serving* the project
  - the [`@now/static`](https://zeit.co/docs/v2/deployments/official-builders/static-now-static/) builder for copying static files needed at runtime, such as `packages/venia-concept/templates`, which the build process will not automatically copy over since they are not generated assets.

### New URL deployments

All branches will now deploy to predictable URLs based on their Git branches and users.  Then we can configure aliases like this:

| Deployment Type | Deployment URL | Aliased Domain | 
| - | - | - |
| Current release version | `https://venia-git-master.magento-research.now.sh` | `https://veniapwa.com` |
| Current "unstable" dev branch | `https://venia-git-develop.magento-research.now.sh` | `https://magento-venia-develop.now.sh` |

**There is one non-intuitive aspect to this.** We automatically deploy to Now.sh using the [Now for GitHub integration](https://zeit.co/docs/v2/integrations/now-for-github). That integration makes one assumption that isn't true for our project: namely, that the GitHub "default" branch always contains the "production" code. Our default branch is "develop", which encourages proper Git Flow behavior and correct targeting for pull requests. This means that Now for GitHub treats a push to `develop` as a "production" deployment.

Fortunately, this only shows up if you run `now` CLI commands directly, or when you read Now documentation. The effect for us is that deployment to `develop` will automatically alias to the friendly now.sh URL `magento-venia-develop.now.sh`.

## Related Issue

Closes #1163.

### Verification Steps
- [ ] Verify that this pull request has deployed to a predictable URL.
- [ ] Run a full end-to-end test suite on this deployment to test the new serverless architecture's performance.
- [ ] Do UAT to verify adequate performance.
- [ ] :warning: Further steps post-merge are necessary, see below.

### :warning: Required Next Steps

- [ ] **Merge this pull request to `develop`.**
- [ ] **Verify that `https://venia-git-develop.magento-research.now.sh` deployed as expected.**
- [ ] **Verify that `https://magento-venia-develop.now.sh` aliased as expected.**
- [ ] **Issue a patch release (2.1.1) of PWA Studio, so we can initiate a build from `master`.**
- [ ] **Verify that `https://venia-git-master.magento-research.now.sh` deployed as expected.**
- [ ] **Change the DNS configuration of `veniapwa.com` to point to `venia-git-master.magento-research.now.sh`.**
- [ ] **Wait 24 hours to start hitting `veniapwa.com` in your browser, to avoid caching DNS upstream.**

## How Have YOU Tested this?
Pushing this pull request should demonstrate now.sh v2 configuration. There should be a newly deployed branch with a predictable URL.

## Checklist:


supernova-at:
> Verify that this pull request has deployed to a predictable URL.

Is this URL correct: https://venia-git-zetlen/now-v2-and-staging.magento-research.now.sh ? If so, it's not resolving.

package.json
```diff
@@ -79,7 +77,7 @@
     "bundlesize": "~0.17.1"
   },
   "engines": {
-    "node": ">=10.7.0",
+    "node": "8.10.0 || >=10.7.0",
```
supernova-at: I think I saw in chat you mentioned that Now v2 only runs Node 8. Is that the reason for this change?

I know @sharkySharks spent a lot of time verifying what version of Node PWA / Venia actually works on; I don't know if `8.10.0` passed the test.

Not sure how to resolve if not, though.

now.json
```diff
@@ -1,10 +1,60 @@
...
+        },
+        {
+            "config": {
+                "includeFiles": [
+                    ".env.dist",
+                    "venia-upward.yml",
+                    "static/*",
+                    "templates/*"
+                ],
+                "maxLambdaSize": "50mb"
```
jimbo: 😰

package.json
jimbo: I also remember this being the source of a lot of difficulty with `now.sh`. I'm fine with this if it works, though; we can make it clear in the docs that we don't actually support `8.x` if that's the case.

packages/venia-concept/package.json
```diff
@@ -17,9 +17,10 @@
     "build:analyze": "yarn run clean && mkdir dist && yarn run validate-queries && yarn run build:stats",
     "build:dev": "echo 'Skipping venia-concept build...'",
     "build:esm": "BABEL_ENV=development babel src --out-dir esm --root-mode 'upward' --source-maps --copy-files",
-    "build:prod": "webpack -p --color --progress --profile --env.mode production",
+    "build:prod": "webpack -p --color --no-progress --profile --env.mode production",
```
jimbo: Can we just remove the `--no-progress` flag and have the same effect?

packages/venia-concept/server-lambda.js
```diff
@@ -0,0 +1,44 @@
...
+    const server = await createUpwardServer(config);
+    return server;
+}
+
+console.log('[VENIA] Launching staging server...\n');
+const appStarted = serve();
+
+// This part is for now.sh version 2 and other serverless deployments.
+// A lambda probably won't execute against more than one request, but juuuuust
+// in case it does, let's cache the server startup procedure.
```
jimbo: 👍

zetlen:
Been trying for 4 days and hundreds of deployments to get the lambda working with image compression and `libvips`. Giving up for now--if anyone else wants to take a look at it, be my guest.

now.json
sirugh: Removing sharp dropped this to 3mb soooo *shrug*

sirugh:
It sounds like we either need to:
* wait for the resolution of [this issue](https://github.com/zeit/now-builders/issues/348) and try again or 
* go with an alternative deployment solution like ec2 or
* use a hosted image service like upward PHP and hardwire to that (as a temporary stopgap).

As of standup yesterday it sounded like we were going with the third option. @zetlen @tjwiebell did ya'll talk after standup yesterday?

sharkySharks:
Not sure of the current discussions around this, but there are 2 deployed instances of PWA, one for staging and one for "prod", that can point to different branches in the repo that live up in AWS. I'm not sure if you want to spend a lot of time with Now.sh given that you have full control over your resources in the AWS account.

zetlen:
@sharkySharks I don't think the current AWS deployments do what we need to do, which is deploy for each pull request. It's hard to tell, though, because my account doesn't have permissions to set up deploy configurations. I can't inspect that area of the build.

At any rate, this is a nice lambda implementation that we could reuse for AWS Lambda in the future.

zetlen:
If you click on the "Show environments" link below, next to the "This branch was successfully deployed" notification, you'll go to the ephemeral URL. But this branch is indeed always aliased based on its branch name! Yay!

https://venia-git-zetlen-now-v2-and-staging.magento-research1.now.sh/

@supernova-at I think you guessed this URL almost right six days ago. Thing is my branch name contains a slash, so to make a valid domain name, now.sh changes it to a dash.

packages/venia-concept/package.json
sirugh: Yea I don't think this is an option. [Check it](https://webpack.js.org/api/cli/#debug-options).

zetlen: Even if it's not an option, it's good to be explicit about it. (I also think it may be an option, because most CLI parsers for Node apps, like yargs and commander, accept `--no-` prefixes on any boolean flag. This way, the explicit CLI option can override any other setting.

sirugh: `--progress false` would be the explicit way, I think, per the docs.

But that's cool about the `--no-` prefix. The more you no!

package.json
sirugh: Since this is a temporary measure to get a live demo instance working for Imagine I also think this is fine. @jcalcaben do you know if we call out the supported node version anywhere for the server side things?

jcalcaben: https://magento-research.github.io/pwa-studio/venia-pwa-concept/setup/

packages/venia-concept/server-lambda.js
```diff
@@ -0,0 +1,42 @@
...
+} = require('@magento/upward-js');
+
+async function serve() {
+    const config = Object.assign({}, envToConfig(validEnv), {
+        env: validEnv,
+        before: app => {
+            app.use(bestPractices());
+        },
+        bindLocal: false,
+        logUrl: false
```
sirugh: `bindLocal` and `logUrl` can be and are set from the environment variables passed in the `now.json` file. You could remove lines 15 and 16 I think.

now-build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -xe
+
+cd ../../
+export NODE_ENV="production"
```
sirugh: Why is this necessary? According to the docs you can set env variables, as you do, in `now.json` and I see that you are already setting `NODE_ENV` to production.

now.json
sirugh: As we saw, removing the image opt middleware lowered this size so you *should* be able to remove this and have it set back to the default of `5mb`.

https://zeit.co/docs/v2/deployments/concepts/lambdas/#maximum-bundle-size

now-build.sh
```diff
@@ -0,0 +1,6 @@
...
+yarn concurrently 'yarn workspace @magento/pwa-buildpack run build' 'yarn workspace @magento/peregrine run build:esm'
```
sirugh: Why not just `yarn run build`? Would essentially do all of the things on line 5 and 6 with some extra steps. Was it the extra steps we're trying to avoid here?

now.json
```diff
@@ -1,10 +1,68 @@
...
-        "node": "10.x.x"
+    "version": 2,
+    "name": "venia",
+    "alias": [
+        "magento-venia-develop"
+    ],
+    "scope": "magento-research",
+    "env": {
+        "UPWARD_JS_BIND_LOCAL": "false",
+        "USE_FASTLY": "1",
```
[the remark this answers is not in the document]
tjwiebell: Its usage was recently removed in #1183 and isn't needed anymore.

packages/upward-js/package.json
```diff
@@ -38,14 +38,15 @@
...
+    "graphql": "~14.2.1",
     "graphql-tag": "~2.10.1",
     "hogan.js": "~3.0.2",
     "http-proxy-middleware": "~0.19.1",
     "js-yaml": "~3.12.1",
     "lodash": "~4.17.11",
     "morgan": "~1.9.1",
-    "node-fetch": "~2.3.0"
+    "node-fetch": "~2.3.0",
+    "shrink-ray-current": "~4.0.0"
```
tjwiebell: This re-introduces the bug solved by #864, this can't be a hard dependency.

zetlen: [It can now](https://github.com/Alorel/shrink-ray/releases/tag/4.0.0)! Instead of having transitive dependencies on native bindings, we can now have direct dependencies. That's always better. So that's what the new optionalDependencies are!

now-build.sh
zetlen: Yeah, this was to speed up deployments. In deployments we don't need to build Venia's ES modules for distribution, and we don't need to validate queries. It's a speedup of almost half a minute!

packages/upward-js/package.json
tjwiebell: Neat! I thought the offending dependency was `node-gyp`, but it appears it's not a direct dependency of `shrink-ray-current`. Can we confirm this is working as expected on a machine w/o Python?

now-build.sh
sirugh: Oh my!

sirugh:
Assigned this back to @zetlen to address feedback. After that I'll review once more and move it along!

now.json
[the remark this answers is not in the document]
zetlen: Can we leave it? I'd like to use it for #1213 .

packages/peregrine/package.json
```diff
@@ -50,7 +50,7 @@
     "react-router-dom": "~5.0.0"
   },
   "engines": {
-    "node": ">=10.7.0",
```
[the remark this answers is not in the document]
zetlen: It is necessary for the build step to execute, I'm afraid. :(

packages/upward-js/package.json
zetlen: I don't think the container has Python, but I've tested with a local container lacking python as well and it doesn't choke there.

now.json
```diff
@@ -1,10 +1,68 @@
...
+        "MAGENTO_BACKEND_URL": "https://master-7rqtwti-xjmpcdjpe6kzm.us-4.magentosite.cloud/",
+        "NODE_ENV": "production"
```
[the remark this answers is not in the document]
zetlen: We don't, because that makes Yarn install only production dependencies. We specify `NODE_ENV=production` in the build _script_, but that's after install is complete.

now-build.sh
zetlen: It's not obvious, but this build script runs _after_ `yarn install`. If `NODE_ENV=production` is set during yarn install, it won't install devDependencies. So we need to wait until install is done before setting that variable.

zetlen:
@sirugh @tjwiebell @supernova-at Ready again for your review. I added some explanation for the now.json configurations. @sirugh The most recent build seems to be 13mb, so I set the lambda limit at 20mb.

now-build.sh
sirugh: I think a comment in this script would go a long way :) Thanks for clearing that up!

now.json
```diff
@@ -1,10 +1,66 @@
...
-    "alias": ["magento-venia", "veniapwa.com"],
-    "engines": {
-        "node": "10.x.x"
+    "version": 2,
+    "name": "venia",
+    "alias": [
+        "magento-venia-develop"
+    ],
+    "scope": "magento-research",
+    "env": {
```
tjwiebell: Need to add `BRAINTREE_TOKEN: "sandbox_8yrzsvtm_s2bg8fs563crhqzk"` to this list so payments will work

tjwiebell: It appears this is actually needed at build time, that's why copying doesn't solve the problem. It would make sense the bundle can't expect to have access to `process.env`, so the value has to be built in. Add to the build `env`, and that should resolve.

zetlen: I wasn't copying the .env file!

zetlen:
@tjwiebell @sirugh I added some comments and .env file copying to the build script. Confirmed that Braintree payment form is showing up now. Ready for re-review!

now-build.sh
```diff
@@ -1,6 +1,18 @@
...
+# NODE_ENV=production to optimize the build step.
 export NODE_ENV="production"
+
+# This step will happen _twice_ in yarn deploy: once for the static deploy and
+# once to compile the lambda. So instead of a full build, this command runs a
+# streamlined build to make only what is essential for deploy.
 yarn concurrently 'yarn workspace @magento/pwa-buildpack run build' 'yarn workspace @magento/peregrine run build:esm'
+# For the Venia build, any environment variables not set in now.json or other
+# scripts are filled in by copying the default .env file here.
+cp packages/venia-concept/.env.dist packages/venia-concept/.env
```
tjwiebell: Does this still receive the build environment variables in `now.json`? Just want to verify that build time `MAGENTO_BACKEND_URL` matches the runtime `MAGENTO_BACKEND_URL`. You've got my 👍 if the answer is yes.

zetlen: Yes, an extant variable always overrides :)

PR #1174, opened 2019-04-26
[Doc] Venia UPWARD spec file explainer

### LINKED ISSUE #1130: [doc]: Verbose comments/explainer on Venia's upward.yml
The [`upward.yml`](/magento-research/pwa-studio/blob/develop/packages/venia-concept/venia-upward.yml) is a great example of the UPWARD spec applied to a real project. However, if that's your introduction to it, it can be a little intimidating.

It would be helpful to have this file commented as an explainer / walkthrough of what each resolver is doing and why it's structured in this particular way.

Specific things to call out in the file:

- when it comes to naming - what's part of the spec / reserved, what's convention, what's developer choice. e.g. `response`, `proxy`, `resource` at the top-level sound like they might have built-in meanings.
- use of variables / aliases (or at least, that's what they look like). e.g. `cmsPage: cmsPageResult.data.cmsPage`, why is it specified where it is, what's it doing? `cmsPage` is used further up in the config in `resource` which is harder to read.

Some things that would be useful in a getting started / explainer:

- what kind of feature means I need to add something to the config? The `CMS_PAGE` section has some cross-over with the app-routing, but it's not immediately clear how that relates.
- does everything start from the top-level `resource` resolver and then branch under that, or do any requests fall through to be picked up by later resolvers?

cc @zetlen

jcalcaben:
## Description

Add verbose comments to Venia's UPWARD spec file to help explain what each piece does.

## Related Issue

Closes #1130 

## Verification Steps

1. Open Venia's UPWARD spec file
2. See comments

## How Have YOU Tested this?

n/a

n/a

## Checklist:


jcalcaben:
cc @rowan-m

packages/venia-concept/venia-upward.yml
```diff
@@ -1,43 +1,86 @@
+# This is the UPWARD specification file for the Venia storefront.
+# It is used by an UPWARD server implementation, such as upward-js and
+# upward-php, to describe how the server responds to requests.
```
supernova-at: Can we drop a link to the docs in here too?

packages/venia-concept/venia-upward.yml
```diff
@@ -1,43 +1,86 @@
...
       use: imageOpt
+    # Requests to graphql/rest endpoints, the media library, and cache are
+    # handled by the top-level 'proxy' object, which is a ProxyResolver
+    # that passes the request through to the backing Magento server.
     - matches: request.url.pathname
       pattern: '^/(graphql|rest|media|cache/)'
       use: proxy
+    # Requests to create an account are handled by the top-level 'index'
+    # object, which returns a generic app shell since the feature has
+    # not been implemented.
```
supernova-at: Is this accurate? Creating an account works in Venia I thought.

supernova-at:
I feel like @zetlen should have a look at this too.

packages/venia-concept/venia-upward.yml
jcalcaben: you're probably right about the feature being implemented. if that's the case, I don't know why this path is listed here specifically since it returns nothing but a 'Loading...' message.

packages/venia-concept/venia-upward.yml
```diff
@@ -68,97 +117,177 @@ isDevelopment:
...
+# HTML containing the PWA application shell.
+# For SEO purposes, the appropriate meta tags in the HTML head element are also
+# set based on information about the resource.
+# This object uses properties in the top-level 'resource' object to return the
+# appropriate response values.
 appShell:
   inline:
     status:
       when:
+        # A request for a valid resource results in a 'resource' object with
```
zetlen:
```suggestion
        # To test this clause, the ConditionalResolver must evaluate the
        # top-level `resource` object. UPWARD is lazily evaluated, so if this is
        # the first time `resource` was looked up during the request, then this
        # `matches` clause will trigger `resource` to begin resolving.
```

packages/venia-concept/venia-upward.yml
```diff
@@ -68,97 +117,177 @@ isDevelopment:
...
+        # a model property.
```
zetlen:
```suggestion
        # When `resource` is resolved, it will have a model property.
```

packages/venia-concept/venia-upward.yml
```diff
@@ -68,97 +117,177 @@ isDevelopment:
...
         env: env
 
+# The resource object provides information about the page type and which
+# template to use for a requested resource.
+# It uses a ConditionalResolver on the top-level 'urlResolver' object to
+# determine it's 'model' and 'template' properties.
+# See the definition of the 'urlResolver' object for details on how it determines
+# the page type based on the request URL.
 resource:
   when:
```
zetlen:
```suggestion
   when:
    # As with all other context values, `urlResolver` begins resolving the first
    # time its value is requested. If UPWARD has not yet queried the GraphQL
    # service before evaluating this match clause, then it does so now.
```

packages/venia-concept/venia-upward.yml
```diff
@@ -68,97 +117,177 @@ isDevelopment:
...
           template: './templates/generic-shell.mst'
   default:
+    # If none of the resolver conditions pass, no 'model' property is set and
+    # this resource object sets the template property to the notfound-shell
+    # template file.
     inline:
       template: './templates/notfound-shell.mst'
 
+# The assetManifest object is a FileResolver that returns JSON data from a
+# manifest.
```
zetlen:
```suggestion
# file generated by Webpack. This file contains a list of all the "bundle"
# and "chunk" filenames generated by the last Webpack build. Using this data,
# the templates can render dynamic script tags based on the filenames of the
# most recent build assets.
```

packages/venia-concept/venia-upward.yml
```diff
@@ -68,97 +117,177 @@ isDevelopment:
...
+#
+# NOTE: Extracting smaller objects from large or deeply nesting data objects
+# is a pattern used to enhance readability and avoid long "dot lookup" strings
 urlResolver: urlResolverResult.data.urlResolver
 
+# The urlResolverResult object is a ServiceResolver that returns data from a
+# GraphQL backing service, whose URL is resolved in the top-level 'magentoGQL'
+# object.
+#
+# This object retrieves information from Magento about the request path.
```
zetlen:
```suggestion
# This object retrieves information from Magento about the request path,
# so the PWA can respect redirects and routes controlled by the admin.
```

packages/venia-concept/venia-upward.yml
jcalcaben: @zetlen should I go ahead and remove the entry in this PR?

packages/venia-concept/venia-upward.yml
```diff
@@ -1,43 +1,90 @@
...
+    # object, which returns a generic app shell since the feature has
+    # not been implemented.
     - matches: request.url.pathname
       pattern: '^/create-account'
       use: index
-    # serve document root (blank urlKey)
+    # Requests for static resources provided by the application project, such
+    # as icons, manifests, and static images, are handled by the top-level
+    # 'static' object, which defines a DirectoryResolver that serves assets
+    # from the './static' folder in the project. The label 'static' is chosen
```
sirugh: Is the folder name configurable or are you saying here that static assets *must* be served from a `static` dir?

packages/venia-concept/venia-upward.yml
```diff
@@ -52,13 +99,20 @@ imageOpt:
             path: request.url.query.url
           template:
             when:
+              # Redirect to the Fastly service, which is presumed to be part
+              # the URL query, when the USE_FASTLY environment value is present
```
sirugh: I think this environment variable is no longer necessary. See [here](https://github.com/magento-research/pwa-studio/commit/1f707406eab90756ce7a22824c06af7a20caa5e6). I don't quite understand how it would be used now. @tjwiebell may have input.

packages/venia-concept/venia-upward.yml
```diff
@@ -68,97 +122,187 @@ isDevelopment:
   default:
     inline: false
 
+# A ProxyResolver object that passes a request to the backend Magento
+# server defined in the MAGENTO_BACKEND_URL environment variable.
+# An UPWARD server infers this object as a ProxyResolver due to the presence
+# of the 'target' property.
```
sirugh: I thought the upward server would know to use the `ProxyResolver` because of the "telltale" `proxy` on line 129. What does this mean then?

jcalcaben: 'proxy' is just the name of the context object. As far as I know an UPWARD server only infers resolver types through its properties, not it's name.

See: https://github.com/magento-research/pwa-studio/tree/develop/packages/upward-spec#resolver-type-inference

packages/venia-concept/venia-upward.yml
```diff
@@ -68,97 +122,187 @@ isDevelopment:
...
       provide:
         model: resource.model
         bundles: assetManifest.bundles
         urlResolver: urlResolver
         env: env
 
+# The resource object provides information about the page type and which
+# template to use for a requested resource.
+# It uses a ConditionalResolver on the top-level 'urlResolver' object to
+# determine it's 'model' and 'template' properties.
```
sirugh:
```suggestion
# determine its 'model' and 'template' properties.
```

packages/venia-concept/venia-upward.yml
sirugh: I don't think we should remove the entry as part of this documentation-only PR. I'd rather see a separate issue/PR made for that work _just incase_ removing this would introduce a bug.

packages/venia-concept/venia-upward.yml
sirugh: Great, thanks for the link!

packages/venia-concept/venia-upward.yml
tjwiebell: `static` is a reference to this definition below, but you could create another directory resolver definition to specify a custom directory, or modify the existing one.
```yaml
static:
  directory:
    inline: './static'
```

packages/venia-concept/venia-upward.yml
tjwiebell: `USE_FASTLY` is no longer in use. We should probably merge in mainline and update the docs; more along the lines of:

> Fastly Image Optimization parameters are preemptively added to proxied media requests to Magento, and will be handled by Fastly, if enabled. Otherwise, requests will fall through to Magento, those parameters will be ignored, and unoptimized images will be served.

packages/venia-concept/venia-upward.yml
```diff
@@ -68,97 +122,187 @@ isDevelopment:
...
+# shorthand syntax for a FileResolver.
 urlResolverResult:
   url: magentoGQL
   query: './src/queries/urlResolver.graphql'
   variables:
     inline:
       urlKey: request.url.pathname
 
+# The magentoGQL object uses a TemplateResolver to generate the URL for the
+# GraphQL endpoint using the MAGENTO_BACKEND_URL environment variable.
```
tjwiebell: We replaced template resolution of URLs with the new URL Resolver. Let's merge in mainline and update these sections.

packages/venia-concept/venia-upward.yml
zetlen: I thought my addition might make that clearer, but I guess not. 

> The label 'static' is chosen to match the directory, but the resolver is still manually configured to use that directory by name.

How could this be rephrased, @tjwiebell @sirugh ?

packages/venia-concept/venia-upward.yml
zetlen: I wanna bring it back, so we know when to base media URLs directly on the instance.

packages/venia-concept/venia-upward.yml
jcalcaben: Merged and updated

packages/venia-concept/venia-upward.yml
sirugh: This was just my lack of understanding. I see now that `use: static` uses that handler which is where the directory is configured.

tjwiebell:
Looking into failed now.sh deployment, but should be good to merge if that was just a service interruption. Will also wait for another 👍 from @sirugh since he reviewed a lot as well, and is the perfect audience to gauge if this describes the config well.

PR #1094, opened 2019-04-01
Fixed configurable media loading issue on product detail page

vishal-7037:
## Description

I've fixed configurable media loading issue and now, its behaviour is the same as luma theme.

## Related Issue

Closes #1093.

## Verification Steps

1. Go to any configurable product.
2. Click on configurable option of fashion color.

## Screenshots / Screen Captures (if appropriate):
**Before Fixed:**
2019_04_01_configurable_media_issue

**After Fixed:**
2019_04_01_configurable_media_fixed

## Proposed Labels for Change Type/Package

- [ ] major (e.g x.0.0 - a breaking change)
- [ ] minor (e.g 0.x.0 - a backwards compatible addition)
- [ ] patch (e.g 0.0.x - a bug fix)

## Checklist:

- [ ] I have updated the documentation accordingly, if necessary.
- [ ] I have added tests to cover my changes, if necessary.

PR #1189, opened 2019-05-01
Unit Test for ProductFullDetail Media Gallery

### LINKED ISSUE #1178: Add Unit test coverage to pwa-studio #1094
Unit test coverage is missing for pull request pwa-studio #1094

tjwiebell:
## Description

Adds unit test for media gallery changes merged in #1094. In this scope it was discovered that `ProductFullDetail` was being wrapped by redux inline, and not using a `container` component. We've also made that change in this scope. 

## Related Issue

Closes #1178

## Verification Steps

Run unit tests w/ coverage

## How Have YOU Tested this?

Ran unit test

## Checklist:


packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -18,7 +18,7 @@ import isProductConfigurable from 'src/util/isProductConfigurable';
 
 const Options = React.lazy(() => import('../ProductOptions'));
 
-class ProductFullDetail extends Component {
+export class ProductFullDetail extends Component {
```
tjwiebell: @jimbo - @sirugh and I wanted your opinion on this change. A common pattern in Venia is to use container components that handles the redux and classify HoCs, but this component currently does it inline, and that change would not be backward compatible. This change allows us to more easily unit test, and we know there are alternatives, but this seemed like the most simple backward compatible solution. Would you prefer either of the alternatives below?

1) Add container component, BIC approved.
2) Mock `classify`, `connect`, and `compose` and leave implementation as-is (ran into issues with this, so would need some guidance).

sirugh: 3. export the unwrapped class but as a name that identifies it as such (ie.` UnwrappedProductFullDetail`)

supernova-at: I don't think we should be afraid of breaking changes at this stage of the game (the next release will already be a major if #1169 goes in). I'd prefer the consistency of a container component here.

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
supernova-at: I think we're trying to move away from `enzyme` in our tests in favor of `react-test-renderer`.

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
```diff
@@ -0,0 +1,95 @@
...
+    const wrapper = shallow(
+        <ProductFullDetail
+            product={mockConfigurableProduct}
+            isAddingItem={false}
+            classes={{}}
+        />
+    );
+
+    expect(wrapper.instance().mediaGalleryEntries).toHaveLength(2);
+    wrapper.setState({
```
supernova-at: If you do switch to `react-test-renderer` you can use [instance.setState](https://reactjs.org/docs/test-renderer.html#testinstanceinstance) to replicate this functionality.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
tjwiebell: Neither of us really liked it, very confusing that `import ProductFullDetail` would be different from `import { ProductFullDetail }`. Made this change. Should we add this pattern to the standards and conventions wiki article?

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
tjwiebell: Enzyme still seems to be better at shallow rendering, so there may be an argument for keeping it strictly for non-DOM tests, but I've made this change.

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
tjwiebell: Done

PR #1193, opened 2019-05-02
[feature]: Adding a window size hook

### LINKED ISSUE #1192: [feature]: A window size hook!
Components (like Thumbnail) may want to do something when the window size changes such as make a new determination whether to render or not.

Implement a window size hook!

We can also implement this as multiple hooks for height and width but that's probably overkill

**Additional context**
https://codesandbox.io/s/qz590jkzv6

- [ ] `venia-concept`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
## Description
I added a window size hook so that components that need to do something based on window size (recalc/rerender/etc) can do so without having to reimplement the same functionality over and over.

I refactored the `Thumbnail` component to use the hook in determining whether or not to render the image.

## Related Issue

Closes #1192 .

## Verification Steps
1) Open the app using chrome dev tools and set your display to "iPad" in portrait mode.
2) Navigate to a PDP and see that no thumbnails are requested.
3) Turn the ipad to landscape mode
4) See that thumbnails are requested (240px image).

## How Have YOU Tested this?
Above verification steps.

- [x] major (e.g x.0.0 - a breaking change) * ONLY MAJOR BECAUSE `useDocumentListener` -> `useEventListener`

## Checklist:


packages/peregrine/src/hooks/__tests__/useWindowSize.spec.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { act } from 'react-test-renderer';
+
+import { useWindowSize } from '../useWindowSize';
+import createTestInstance from '../../util/createTestInstance';
```
sirugh: This test is based on @jimbo's work in #1078. Until that's merged these tests won't work.

packages/peregrine/src/hooks/__tests__/useWindowSize.spec.js
```diff
@@ -0,0 +1,40 @@
...
+    useWindowSize();
+
+    return <i />;
+};
+
+test('adds an event listener to the window on mount', () => {
+    createTestInstance(<Component />);
+
+    const spy = spies.get('addEventListener');
+    expect(spy).toHaveBeenCalledTimes(5);
```
sirugh: These numbers don't make sense to me. Even when isolated to just this test it still is called 5 times. Not sure why. Help?

packages/peregrine/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,31 @@
...
+        outerHeight: window.outerHeight,
+        outerWidth: window.outerWidth
+    };
+};
+
+/**
+ * A hook that will return inner and outer height and width values whenever
+ * the window is resized.
+ */
+export const useWindowSize = () => {
```
sirugh: After talking with @jimbo we're going to add this to a root context so we don't get the awful side effect of adding a new event listener whenever a component tries to use this.

So basically components will subscribe to the context rather than trying to use this hook directly.

jimbo: 🎉 

Right on; most components should call a hook that obtains window size from context, but we'll need to provide that context near the root of the app, and so all the code you've written here will be still be used to attach the listener and set the size in state.

packages/peregrine/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,31 @@
...
+    const [windowSize, setWindowSize] = useState(getSize());
+
+    const handleResize = () => {
+        setWindowSize(getSize());
+    };
+
+    useEffect(() => {
+        window.addEventListener('resize', handleResize);
```
tjwiebell: Could be an opportunity to generalize `useDocumentListener` to `useEventListener` and pass the object you want to subscribe to (window, document, element).

sirugh: That feels like overkill to me.

packages/venia-concept/src/classify.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 
 import getDisplayName from 'src/util/getDisplayName';
 
-const merge = (...args) => Object.assign({}, ...args);
+export const mergeClasses = (...args) => Object.assign({}, ...args);
```
sirugh: Another thing taken from #1078

packages/venia-concept/src/components/App/appWrapper.js
```diff
@@ -0,0 +1,27 @@
...
+import { useWindowSize } from '@magento/peregrine/src/hooks/useWindowSize';
+import App from './container';
+
+import AppContext from './context';
+
+/**
+ * A functional wrapper so that we can provide context to the entire app
+ * without having to rewrite the App file itself.
+ */
+function AppWrapper() {
```
sirugh: I wrote this temporarily since I didn't want to have to refactor `index` or `App` to be a functional component. App is close but it uses `getDerivedStateFromError` which is [not yet hook-ified](https://reactjs.org/docs/hooks-faq.html#how-do-lifecycle-methods-correspond-to-hooks):

> getDerivedStateFromError: There are no Hook equivalents for these methods yet, but they will be added soon.

packages/venia-concept/src/components/ProductImageCarousel/thumbnail.js
```diff
@@ -1,59 +1,56 @@
...
-    // with a useWindowSize hook and rerender when the width changes across the
-    // breakpoint. This is important when changing from portrait to landscape.
-    isDesktop = () => {
-        return window.innerWidth >= 1024;
-    };
-
-    onClickHandlerWrapper = () => {
-        const { onClickHandler, itemIndex } = this.props;
+function Thumbnail(props) {
+    const { windowSize } = useContext(WindowSizeContext);
```
sirugh: This isn't working. For some reason, `WindowSizeContext` is an object containing the exports from the WindowSizeContext file:

```js
WindowSizeContext === {
  WindowSizeContext: (...),
  WindowSizeContextProvider: (...),
  ...
}
```

What did I do wrong? I exported both as non-defaults in the context file and I'm importing by name here so why is it resolving the exported object?

sirugh: Or maybe it's working properly and I'm just not seeing the bug I wrote...

sirugh:
>  Consider renaming the component to just WindowSize.

Should it remain in the `hooks` folder?

sirugh:
Oh and it seems like this will break a lot of tests unless we add a fake WindowSize context to the test runner root. I'll look into that.

jimbo:
> > Consider renaming the component to just WindowSize.
> 
> Should it remain in the `hooks` folder?

I would think so. If It were up to me, the file would be `src/hooks/useWindowSize.js`, and it would export the following:

- `useWindowSize`, the hook for context consumers
- `WindowSize`, the component that wraps the context provider

Alternatively, the file could be `src/WindowSize/index.js`, similar to existing Peregrine components. In either case, we flatten all of this when we re-export from `src/index.js`, so none of this affects Peregrine consumers much. Go with what you think matches best.

packages/venia-concept/src/components/ProductImageCarousel/thumbnail.js
```diff
@@ -23,10 +25,10 @@ function Thumbnail(props) {
 
     const isDesktop = windowSize.innerWidth >= 1024;
 
-    function onClickHandlerWrapper() {
-        const { onClickHandler, itemIndex } = props;
+    const onClickHandlerWrapper = useCallback(() => {
```
jimbo: Conventionally,

- a callback received as a prop is called `onClick`
- a callback passed as a prop is called `handleClick`

```jsx
const { children, onClick } = props
const handleClick = useCallback(
  () => { onClick() },
  [onClick]
)

return (
  <button onClick={handleClick}>
    {children}
  </button>
)
```

sirugh: That makes sense :)

packages/peregrine/src/hooks/__tests__/useWindowSize.spec.js
sirugh: I think React is doing something here though I can't figure out what. The 4 calls before the 5th are `error` listeners.

packages/peregrine/src/hooks/useWindowSize.js
sirugh: I didn't realize we already had the `useDocumentListener` when I wrote that comment. I think I'll give it a shot :D

dpatil-magento:
QA Pass.

packages/peregrine/src/hooks/__tests__/useWindowSize.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    useEventListener: jest.fn()
+}));
+
+const Component = () => {
+    useWindowSize();
+
+    return <i />;
+};
+
+test.only('adds an event listener to the window on mount', () => {
```
supernova-at: Remove `test.only` here

packages/peregrine/src/hooks/__tests__/useWindowSize.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    createTestInstance(
+        <WindowSizeContextProvider>
+            <Component />
+        </WindowSizeContextProvider>
```
supernova-at: Excuse my lack of education on hooks but I thought part of the idea was that you didn't have to rely on context like this.

Why doesn't it work for `Component` to just `useWindowSize` directly?

supernova-at: Okay nevermind, I just reviewed the actual `useWindowSize.js` file. The reason is because we don't want to add duplicate event listeners, right? 

If so, 👍 but it does seem like this pattern could lead to many top-level context wrappers in the future. I'm cool solving that problem if/when we get there though.

jimbo: > If so, 👍 but it does seem like this pattern could lead to many top-level context wrappers in the future. I'm cool solving that problem if/when we get there though.

Yes, it will ideally lead to that. It's like Redux without the slicing.

packages/venia-concept/src/components/ProductImageCarousel/__tests__/carousel.spec.js
```diff
@@ -41,7 +42,11 @@ const defaultProps = {
 };
 
 test('renders the Carousel component correctly w/ sorted images', () => {
-    const component = testRenderer.create(<Carousel {...defaultProps} />);
+    const component = testRenderer.create(
```
jimbo: In general, all new tests should use `createTestInstance()`, which wraps this method with `act()` [to ensure consistency](https://reactjs.org/docs/test-utils.html#act).

sirugh:
This change renames the peregrine `useDocumentListener` which is a public API. Therefor `major` is required.

PR #1196, opened 2019-05-02
bug: Make css-font non render-blocking

### LINKED ISSUE #1195: [bug]: Muli font is a blocking resource!
**Describe the bug**
According to Lighthouse, the Muli font link is a blocking resource.

**To Reproduce**
1. Run the app, and open in Incognito
2. Run lighthouse.
3. See that it complains about a blocking call for the Muli font.

[Image from Gyazo](https://gyazo.com/46bb75cbb34c0b88409b22121240287a)

**Expected behavior**
Lighthouse doesn't complain, and we don't block on the Muli font load

**Possible solutions**
* Use webfont loader (requires javascript)
* Inline source of Muli font from Google
* Bundle font (currently impossible due to how we serve the core templates.

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
## Description
This is bad:
[Image from Gyazo](https://gyazo.com/46bb75cbb34c0b88409b22121240287a)

I hardcoded [the response](https://fonts.googleapis.com/css?family=Muli:300) from Google's font API into our mustache template to avoid the render blocking `<link>` inclusion. In the template I also include a new link with the `rel="preload"` attribute for the font that is requested later in the file which tells the browser to start the request for the font ASAP instead of waiting for the CSSOM (where it would normally find the font url) to be built.

So two things were done here:

1) **Inclusion of link rel="preload" for font**: Browser will request font asap instead of waiting for CSSOM.
2) **Include stylesheet inline**: This is done because CSS is a _render-blocking_ request while font is only a _text-blocking_ request. 

This is a pretty extreme optimization but it does improve the lighthouse scores. The image below should help you visualize the improvement (the css is now included in the `html` response).

Check out [this awesome documentation](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization) for more details about web font optimization.

## Related Issue

Closes #1195 .

## Verification Steps
Run the app through lighthouse and see that it no longer complains about the blocking resource:

[Image from Gyazo](https://gyazo.com/800bf57d92a3b8fd5893e54d9a9104da)

## How Have YOU Tested this?
* Used lighthouse and saw it no longer complains
* Ran app in chrome and saw that fonts were loaded and displayed properly (Muli is the venia-font)
* Blocked JS on page and reloaded and saw the no-script displayed properly.
* Opened app in Safari, Chrome, Firefox on Desktop 

_(Mocked up a fancier preload screen to show the font behavior. -- Z)_

### (Current) Linked stylesheet with blocking font
original font stylesheet

### Linked stylesheet with `display=swap`
display swap stylesheet

### Inlined stylesheet with `font-display: swap`
inlined font def

## Checklist:


packages/venia-concept/templates/default-font-include.mst
```diff
@@ -0,0 +1,30 @@
+{{!
```
sirugh: The contents of this file are copied from the API response: https://fonts.googleapis.com/css?family=Muli:300

That being said, according to the [docs](https://developers.google.com/fonts/docs/technical_considerations) different browsers solicit a different response from the API which may be a concern.

packages/venia-concept/templates/default-font-include.mst
```diff
@@ -0,0 +1,30 @@
...
+  via <link>. Apparently css urls are asynchronous.
+}}
+<style type="text/css">
+  /* vietnamese */
+  @font-face {
+    font-family: 'Muli';
+    font-style: normal;
+    font-weight: 300;
+    src: local('Muli Light'), local('Muli-Light'), url(https://fonts.gstatic.com/s/muli/v13/7Au_p_0qiz-adZnkOCr2z24PMFk-0g.woff2) format('woff2');
+    unicode-range: U+0102-0103, U+0110-0111, U+1EA0-1EF9, U+20AB;
```
zetlen: Seems like all these unicode ranges could be combined, since they're all the exact same otherwise.

sirugh: I'm new to unicode ranges so I'll have to read into what that means but I do want to say that I copied this directly from the API response for the google font.

If I customize this I'll want to add a comment saying as such so that if we go with a different font whomever updates it does the same thing.

sirugh:
Hilariously: https://twitter.com/szynszyliszys/status/1126261253280649216?s=21

They are shipping `font-display` as part of the API (which I also learned uses `rel=preconnect` though apparently [not always](https://twitter.com/addyosmani/status/1126380315994255360). We may be able to just add the new param to our `link` instead of having to host this ourselves!

zetlen:
Good catch, @sirugh. I was _there!!!!_ Yet I show up in this thread and you've already seen the news.

Do you wanna try the URL parameter instead and see if the blocking call warning in Lighthouse still goes away?

sirugh:
> Do you wanna try the URL parameter instead and see if the blocking call warning in Lighthouse still goes away?

I'm not sure when it is shipping but Addy claimed "soon" and when it goes live I'd love to try it. That being said I'm not sure if the render block will go away. I may have this wrong but I think `font-display` just lowers the time of blocked text painting rather than the render block so I think Lighthouse may still complain.

zetlen:
@sirugh OK, let's snooze this until next week when it maybe goes live!

sirugh:
@zetlen @jimbo I tested this using _just_ the original link but with `display=swap` and the render blocking remains. That makes sense though, as the browser has to fetch the css asset before it renders.

So I added the new response with `display=swap` which means we should now get a FOUT instead of FOIT at least.

zetlen:
@sirugh Is there a chance you could post a couple of gifs to compare and contrast the FOUT before `display=swap` with the FOIT afterwards? I like to use [licecap](https://www.cockos.com/licecap/) for quick GIF capturing.

sirugh:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  node` failed.
  </td>
    </tr>
  </tbody>
</table>



### Log


```sh
    at Object.dangerfilePath (/codebuild/output/src450921827/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/utils/fileUtils.js:26:11)
    at Object.<anonymous> (/codebuild/output/src450921827/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:92:42)
    at step (/codebuild/output/src450921827/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:33:23)
    at Object.next (/codebuild/output/src450921827/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:14:53)
    at fulfilled (/codebuild/output/src450921827/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:5:58)

```

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against cef24621ec7652723cb9f36693740a2d23ce3bda
</p>

zetlen:
@sirugh I added some screenshots :)

soumya-ashok:
@zetlen On the loading indicator, there's a standard size for the icon and text we use on other screens - would be good to maintain that consistency in this instance.

dpatil-magento:
@sirugh Site is not accessible, can you please check once?

sirugh:
@dpatil-magento try a redeploy? There's no reason that should have happened to only this PR.

dpatil-magento:
Verified with AWS deployment as there is no easy way to trigger re-deployment on now.sh.

PR #1207, opened 2019-05-06
Fixes Broken Forms

### LINKED ISSUE #1206: magento-venia.now.sh and veniapwa.com are broken
Steps - 
1. Go to https://magento-venia.now.sh or https://veniapwa.com/
2. Try to sign-in, create account or add product to cart and try placing order.

Issue: On clicking buttons submit (create account, forgot password) , Sign In, Use address (checkout flow)  nothing happens.
Expected - User should be able to complete above flows without any errors.

supernova-at:
## Description

Included in the recent merge of #1078 was an upgrade to the `informed` Form library.

[One of the things that changed](https://github.com/joepuzzo/informed/blob/master/CHANGELOG.md#changed) between the previous version we were on and the new one is the return value expected by validator functions:

> the validate prop now expects the validation function to return undefined if there is no error. Any other returned value (including falsey null, 0, etc will be treated as an error for the field.

Previously, some of our validators were returning `null` to indicate successful validation. I have updated them to return `undefined` instead.

## Related Issue

Closes #1206 .

## Verification Steps

Any form that has fields with validators on them is potentially impacted by this change. Please test the following forms specifically:

1. Sign In
1. Forgot Password
1. Create Account
1. Shipping Address (checkout flow)
1. Billing Address (checkout flow)

## How Have YOU Tested this?

1. `yarn test`
1. Verification steps described above

## Checklist:

### DIFF
--- a/packages/venia-concept/src/components/CreateAccount/validators.js
+++ b/packages/venia-concept/src/components/CreateAccount/validators.js
@@ -22,7 +22,7 @@ const isPasswordComplexEnough = (str = '') => {
 
 export const validators = new Map()
     .set('confirm', (value, values) => {
-        return value !== values.password ? 'Passwords must match.' : null;
+        return value !== values.password ? 'Passwords must match.' : undefined;
     })
     .set('email', value => {
         const trimmed = (value || '').trim();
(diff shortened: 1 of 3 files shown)


packages/venia-concept/src/components/CreateAccount/validators.js
```diff
@@ -22,21 +22,21 @@ const isPasswordComplexEnough = (str = '') => {
...
+        return value !== values.password ? 'Passwords must match.' : undefined;
     })
     .set('email', value => {
         const trimmed = (value || '').trim();
 
         if (!trimmed) return 'An email address is required.';
         if (!trimmed.includes('@')) return 'A valid email address is required.';
 
-        return null;
+        return undefined;
```
jimbo: Minor quibble: `return undefined` is [redundant](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/return#Syntax).
```suggestion
        return;
```

PR #1211, opened 2019-05-07
Category RootComponent Simple Re-Factor

### LINKED ISSUE #975: Peregrine: Category Component

tjwiebell:
## Description

Re-factor Category RootComponent and all components inside component directory to be functional components that use hooks.

## Related Issue

Closes #975

## Verification Steps

1. Go to the Bottoms Category
2. Verify that a product list renders

## How Have YOU Tested this?

Manually ran steps above and ran test suite

## Checklist:


packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -1,119 +1,97 @@
...
-                            data={data}
-                        />
-                    );
-                }}
-            </Query>
+    if (error) return <div>Data Fetch Error</div>;
+    // If our pagination component has mounted, then we have
+    // a total page count in the store, so we continue to render
+    // with our last known total
+    if (loading || !data)
```
tjwiebell: This change was odd to me; there seemed to be a state where `loading = false` but `data` was still null. Is this expected?

sirugh: Does it eventually resolve `data` or does it stay `null`?

tjwiebell: `data` does eventually resolve

sirugh: https://github.com/magento-research/pwa-https://github.com/magento-research/pwa-studio/blob/develop/packages/peregrine/src/hooks/useQueryResult.js#L24

I wonder what happens if you modify that default value to something else.
Maybe this is a bug in `useQueryResult`?

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -1,119 +1,97 @@
...
-            currentPage,
-            pageSize,
-            prevPageTotal,
-            setCurrentPage,
-            setPrevPageTotal
-        } = this.props;
+    const [queryResult, queryApi] = useQuery(categoryQuery);
+    const { data, error, loading } = queryResult;
+    const { runQuery, setLoading } = queryApi;
+    const classes = mergeClasses(defaultClasses, props.classes);
```
sirugh: As we move to functional components we should colocate the variables with their uses, IMO (that may be controversial). But I think this `classes` assignment can live right above the `return` where it is used. May as well go through and make sure other variables are assigned right before use as well.

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -1,119 +1,97 @@
...
+        setLoading(true);
+        runQuery({
+            variables: {
+                id: Number(id),
+                onServer: false,
+                pageSize: Number(pageSize),
+                currentPage: Number(currentPage)
+            }
+        });
+    }, [id, pageSize, currentPage, runQuery, setLoading]);
```
sirugh: By my understanding you only need to pass in variables for which you want to rerun the effect after changes. So for example here I would expect we only want to rerun this if the `currentPage`,  `id`, or `pageSize` changes. `runQuery` and `setLoading` shouldn't change as they are pre-defined functions.

I could also be wrong 😭

packages/venia-concept/src/RootComponents/Category/category.js
tjwiebell: Reviewed this with @jimbo, and he confirmed that `useQueryResult` has introduced a new "empty" state that we do need to handle (ideally we should have been handling it already). His Search component re-factor demonstrates the `useQueryResult` states a little better, since it has an "empty" state to handle.
```js
    if (error) {
        message = 'An error occurred while fetching results.';
    } else if (loading) {
        message = 'Fetching results...';
    } else if (!data) {
        message = 'Search for a product';
    } else if (!data.products.items.length) {
        message = 'No results were found.';
    } else {
        message = `${data.products.items.length} items`;
    }
```

jimbo: @tjwiebell Right, thanks for adding that. The empty state exists when you don't have an in-flight request (`loading` is `false`) but you haven't received any responses yet (so `data` is `null`).

The empty state in `Autocomplete` occurs when a user hasn't typed anything yet, since typing is what initiates the request. But here, it occurs when the component is first mounted, since the request is initiated *after* mount, as an effect. That may seem counterintuitive, but a React component isn't allowed to initiate requests before mounting (and for good reasons), so we do briefly have a state where there's neither a request nor a response yet.

packages/venia-concept/src/RootComponents/Category/container.js
```diff
@@ -0,0 +1,18 @@
...
+        pageSize: catalog.pageSize,
+        prevPageTotal: catalog.prevPageTotal
+    };
+};
+const mapDispatchToProps = { setCurrentPage, setPrevPageTotal };
+
+export default connect(
+    mapStateToProps,
+    mapDispatchToProps
+)(Category);
```
jimbo: I think this entire connection to redux could be removed. I'll investigate a bit more.

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -1,119 +1,97 @@
...
-                        ) : (
-                            loadingIndicator
-                        );
+    useEffect(() => {
+        window.scrollTo({
+            left: 0,
+            top: 0,
+            behavior: 'smooth'
+        });
+    }, [currentPage]);
```
jimbo: Great use of `useEffect()` here. Window scrolling is absolutely a side effect, and we only want to do it when the `currentPage` changes. This could eventually be a candidate for a utility hook.

jimbo: It's also correct to do this at this point in the component, before any early returns. 👍

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -1,119 +1,97 @@
...
-            </Query>
+    if (error) return <div>Data Fetch Error</div>;
+    // If our pagination component has mounted, then we have
+    // a total page count in the store, so we continue to render
+    // with our last known total
+    if (loading || !data)
+        return pageControl.totalPages ? (
+            <CategoryContent pageControl={pageControl} pageSize={pageSize} />
+        ) : (
+            loadingIndicator
```
jimbo: I don't think we want to show a loading indicator when we're not loading. If a request was not initiated, this would be loading forever. Consider returning `null` or something when we're in that empty state.

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -50,30 +47,21 @@ const Category = props => {
...
-        return pageControl.totalPages ? (
-            <CategoryContent pageControl={pageControl} pageSize={pageSize} />
-        ) : (
-            loadingIndicator
-        );
+    useEffect(() => {
+        if (data) {
+            setTotalPages(data.category.products.page_info.total_pages);
+        }
+    }, [data]);
```
sirugh: I think we only need to check this if the `total_pages` change rather than all of the `data` prop. Would prevent an unnecessary re-render if other props change but the total pages stays the same.

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -9,20 +9,17 @@ import { loadingIndicator } from 'src/components/LoadingIndicator';
...
+    const { setCurrentPage, setTotalPages } = paginationApi;
 
     const pageControl = {
-        currentPage: currentPage,
+        currentPage,
         setPage: setCurrentPage,
-        updateTotalPages: setPrevPageTotal,
-        totalPages: prevPageTotal
+        updateTotalPages: setTotalPages,
+        totalPages
```
sirugh: Similar to how I shared the Toast state could we also create a `Pagination` context and use it to share these props rather than passing them down? We already have the hooks so it feels like a pretty simple win. Not sure if you already talked about that with @jimbo though.

packages/venia-concept/src/RootComponents/Category/category.js
tjwiebell: The initialState of `data = null`, and React's dependency checking didn't seem intelligent enough to do `!== undefined` down the chain, so this wasn't possible w/o triggering a `Cannot read property 'category' of null` error.

packages/venia-concept/src/RootComponents/Category/category.js
tjwiebell: In talking with @jimbo, we decided to leave the prop drilling in place, and when we re-factor `Paginiation`, we'll come back and fix it here.

packages/venia-concept/src/RootComponents/Category/category.js
sirugh: You could pull out the prop before the hook use:

```js
let total_pages = null;
if (data) {
  { category: { products : { page_info: total_pages } } } } = data;
}
useEffect(() => { setTotalPages(total_pages) }, [total_pages])
```

packages/venia-concept/src/RootComponents/Category/category.js
zetlen: He's right!

packages/venia-concept/src/RootComponents/Category/category.js
zetlen: The line `const { total_pages } = data && data.category.products.page_info` might be more readable, but otherwise, that's exactly it. JavaScript deserves a proper coalesce operator and it doesn't have one.

packages/peregrine/src/hooks/usePagination.js
```diff
@@ -0,0 +1,14 @@
...
+    const [currentPage, setCurrentPage] = useState(0);
+    const [totalPages, setTotalPages] = useState(null);
+
+    const paginationState = { currentPage, totalPages };
+    const api = useMemo(() => ({ setCurrentPage, setTotalPages }), [
+        setCurrentPage,
+        setTotalPages
+    ]);
+
+    return [paginationState, api];
```
zetlen: Clean and beautiful.

packages/venia-concept/src/RootComponents/Category/category.js
zetlen: Should we add `pageSize` to the dependent property list? I think we should scroll to top when that changes.
```suggestion
    }, [currentPage, pageSize]);
```
Even the `id` property might be a good idea; it would be cool to reuse most of the UI when you navigate between entities of the same type.

packages/venia-concept/src/RootComponents/Category/category.js
sirugh: This is an alternative.

```js
    const total_pages = data ? data.category.products.page_info.total_pages : null;
    useEffect(() => {
        setTotalPages(total_pages);
    }, [total_pages]);

```

tjwiebell: I'm not a fan of this change and think y'all are over-optimizing (useState sets with identical values are noop), but made it anyway; I hope you're happy 😭

packages/venia-concept/src/RootComponents/Category/category.js
tjwiebell: Moved this logic into the hook above, that has all these same dependencies.

sirugh:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  node` failed.
  </td>
    </tr>
  </tbody>
</table>



### Log


```sh
    at Object.dangerfilePath (/codebuild/output/src997859200/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/utils/fileUtils.js:26:11)
    at Object.<anonymous> (/codebuild/output/src997859200/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:92:42)
    at step (/codebuild/output/src997859200/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:33:23)
    at Object.next (/codebuild/output/src997859200/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:14:53)
    at fulfilled (/codebuild/output/src997859200/src/github.com/magento-research/pwa-studio/node_modules/danger/distribution/commands/danger-runner.js:5:58)

```

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8b13a9fa6d90d3bdf9981d2b7582c6667bde233d
</p>

dpatil-magento:
@sirugh  I'm bit confused on purpose of new build pipeline. Need to discuss.

@tjwiebell  Observations - 
1. Be on page any page other than 1 and reload browser, page 1 is being displayed. Or just try accessing https://venia-8cmy79725.now.sh/venia-bottoms.html?page=4 , instead pf requested page 4, page 1 is being displayed. Page 1 should only display if page 4 does not exists.

2. Click on sub-category which has just 1 product (Shop the look > retire-your-lbd). For fraction of seconds entire page skeleton shows loading and then 1 product is displayed. I can show this behavior if not well explained here.

3. Even though category has just 1 product, URL has "page=1" parameter, not sure if its expected? https://venia-8cmy79725.now.sh/shop-the-look/retire-your-lbd.html?page=1

packages/venia-concept/src/components/Pagination/pagination.js
```diff
@@ -171,28 +166,31 @@ class Pagination extends Component {
...
-            // Bitwise NOTing any number x yields -(x + 1). For example, ~-5 yields 4.
-            // Importantly, it truncates any fractional component of x. For example, ~-5.7 also yields 4.
-            // For positive numbers, applying this operator twice has the same effect as Math.floor.
-            ~~getQueryParameterValue({
-                location,
-                queryParameter: 'page'
-            })
-        );
+        // Indicator that our data has been loaded and we know enough to bounds check
+        if (totalPages) {
```
sirugh: What is the case in which we would be loaded but not have `totalPages`?

tjwiebell: Somehow a render of pagination is entering the event loop where `usePagination` is still in it's initialState (`totalPages = null`), but this is a very good question. Persistent pagination logic in `category.js` might be conflicting with this queryParams logic. With the assumption that the pagination would go through its own re-factor, I was going for a quick fix, but I can look into it more.

tjwiebell: Fixed by f209db6; digging into it, the child component previously never rendered until totalPages was set, which I've now restored by using that as our loading check.

packages/venia-concept/src/components/Pagination/pagination.js
```diff
@@ -27,11 +27,6 @@ class Pagination extends Component {
     };
 
     componentDidMount() {
-        // updateTotalPages pushes the current page count of a category query to
-        // redux so it knows how many tiles to render even in the Query
-        // component's loading state
-        const { updateTotalPages, totalPages } = this.props.pageControl;
-        updateTotalPages(totalPages);
```
sirugh: We don't need to do this anymore?

tjwiebell: Nope. We're grabbing this from the queryResult now, instead of calculating ourselves in `category.js` and prop drilling, so it's already set.

dpatil-magento:
@tjwiebell QA pass but need final review approval to get merged.

Issue #1184, opened 2019-04-30
[feature]: Remove 301 Redirect for Image Optimization


tjwiebell:
**Is your feature request related to a problem? Please describe.**
`makeUrl.js` currently captures Product and Category image requests, and outputs them in the `/img/resize` format so `upward-js` image optimization middleware has an easily identifiable pattern to capture. If you're not using this middleware, this doubles the requests needed to fetch images, since it has to pass through UPWARD twice.

**Describe the solution you'd like**
Instead of letting UPWARD handle the 301 redirects when onboard IO is disabled, we should just let `makeUrl` handle this logic, and generate media URLs (optionally with Fastly params added too). 

**Additional context**
You can go to the below instance to see this in action:

https://master-7rqtwti-xjmpcdjpe6kzm.us-4.magentosite.cloud

**Please let us know what packages this feature is in regards to:**
- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

tjwiebell:
@zetlen - In the scope of the below PR, we're now hard-coding Fastly params in all image requests that don't get intercepted by middleware, making `USE_FASTLY` obsolete. Do we want to stick with this logic, or bring back this environment variable in this issue scope?

https://github.com/magento-research/pwa-studio/pull/1183/files#diff-18caf8bff8d62967d69295861abd143c

PR #1213, opened 2019-05-07
enhancement: unify fastly and local media URLs (no more 301s for images in production)

zetlen:
## Description

Simplify the creation and consumption of image URLs, so they work the
same way in all environments.

This removes the extra 301 redirects to optimized images in production.

- Add a "preload" data to the HTML tag, which renders the Magento
backend URL and the image optimization source (`backend` or `origin`) as data attributes
- Modify `makeUrl()` to query those attributes and:
    - if `IMAGE_OPTIMIZING_ORIGIN` is `backend`, indicating Fastly or any other optimization service is enabled on the backend, then force ALL media URLs to use the backend as the base URL and render as absolute.
   - if `IMAGE_OPTIMIZING_ORIGIN` is `onboard`, indicating the Web server must optimize images, then force ALL media URLs to to render as relative.
- Modify the backend `addOptImgMiddleware()` feature to work with Fastly-style URL parameters, replicating a subset of its features.
- Remove image optimization logic from UPWARD entirely, since it is now
delegated to other layers.

## Related Issue

Resolves the ongoing performance implication of 301 redirects in production.
Resolves the HTTP/2 redirect ding that Lighthouse gives the changes in #1165 .

## Verification Steps

1. Run a staging or development server.
1. Open the network tab in dev tools. Filter for images.
1. Go to the homepage.
1. Observe that the images are compressed and that they have headers indicating that PWA Studio runtime compressed them.
1. Visit a production cloud server running Fastly IO.
1. Go to the homepage.
1. Observe that the image URLs have the same structure, but their headers indicate they were compressed with Fastly instead.

## How Have YOU Tested this?
In Unit Tests: Updated unit tests for makeURL, and updated the snapshots which included makeURL output
For UAT: Locally, but not in Fastly IO yet. Since we let those URLs through unmolested, it should work great.

## Checklist:


zetlen:
Fixes #1184.

packages/venia-concept/src/util/makeUrl.js
```diff
@@ -1,8 +1,8 @@
-// ensure there's exactly one slash between segments
-const joinUrls = (base, url) =>
-    (base.endsWith('/') ? base.slice(0, -1) : base) +
-    '/' +
-    (url.startsWith('/') ? url.slice(1) : url);
+const absoluteUrlRE = /^(?:https?|data|file):/;
+
+// If the root template supplies the backend URL at runtime, use it directly
```
tjwiebell: Why pull this value at runtime instead of baking it into the build w/ `process.env.MAGENTO_BACKEND_URL`?

packages/venia-concept/src/util/makeUrl.js
```diff
@@ -40,35 +39,38 @@ const resizeBase = joinUrls(
  * @param {number} props.width - the desired resize width of the image
  */
 const makeOptimizedUrl = (path, { type, width } = {}) => {
-    const { location } = window;
-    const urlObject = new URL(path, location.href);
-    const params = new URLSearchParams(urlObject.search);
+    const isAbsolute = absoluteUrlRE.test(path);
```
tjwiebell: Starting in `2.3.1`, GraphQL started returning all images as absolute URLs, which is why we had to do away with this logic. This results in only the `width` parameter being appended to these URLs, but it does not add the other two for `webp` or `pjpg` fallback. You will need to come up with another detection method other than absolute URL.

packages/venia-concept/src/util/makeUrl.js
zetlen: I try to avoid embedding highly volatile strings into the bundles, so that we can reuse bundles between deployments. Any env var that can vary from individual deployment to deployment should maybe come out of runtime.

packages/venia-concept/src/util/makeUrl.js
zetlen: Thanks for the tip. We now have a boolean env var for determining whether to use the backend URL or relative URLs to do image optimization. This replaces `USE_FASTLY` and I think it's clearer. What do you think?

packages/venia-concept/src/util/makeUrl.js
```diff
@@ -39,36 +38,42 @@ const mediaBases = new Map()
...
 
-    if (!isAbsolute && type && mediaBases.has(type)) {
+    if (type && mediaBases.has(type)) {
         const mediaBase = mediaBases.get(type);
         // prepend media base if it isn't already part of the pathname
         if (!urlObject.pathname.includes(mediaBase)) {
             urlObject.pathname = mediaBase + urlObject.pathname;
         }
+        // add image optimization parameters and optionally change origin
+        if (type.startsWith('image-')) {
```
tjwiebell: With this nested inside of `if (type && mediaBases.has(type))` we're back to only processing types in the Map. This gnarly nested if should bring that functionality back:

```js
if (type) {
    if (mediaBases.has(type)) {
        const mediaBase = mediaBases.get(type);
        if (!urlObject.pathname.includes(mediaBase)) {
            urlObject.pathname = mediaBase + urlObject.pathname;
        }
    }
    ...
}
```

packages/venia-concept/src/util/makeUrl.js
tjwiebell: I really like this change, great improvement.

packages/venia-concept/src/util/makeUrl.js
zetlen: Very good catch, thank you. Switched.

packages/venia-concept/templates/open-document.mst
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" data-image-optimizing-origin="{{env.IMAGE_OPTIMIZING_ORIGIN}}" data-backend="{{env.MAGENTO_BACKEND_URL}}">
```
sirugh: Would it be better to store these in a `meta` tag rather than directly on the root?

PR #1198, opened 2019-05-03
Create PWA tool

zetlen:
## Description

Refactor Buildpack as a CLI utility and add a `npm init @magento/pwa` starter kit.

#### Buildpack CLI

```
$ buildpack --help
buildpack <command>

Commands:
  buildpack create-env-file <directory>     Generate a .env file in the provided
                                            directory to store project
                                            configuration as environment
                                            variables
  buildpack init-custom-origin <directory>  Get or create a secure, unique
                                            hostname/port combination and a
                                            trusted SSL certificate for local
                                            development.
  buildpack init-project <template>         Create a PWA project in <directory>
  <directory>                               based on <template>.
  buildpack validate-env <directory>        Check the current environment,
                                            including .env file if present, to
                                            ensure all required values are
                                            present.

Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]
```

The big feature is the subcommand `init-project`, which through a combination of arguments and flags will create a new project directory.

```
$ buildpack init-project --help
buildpack init-project <template> <directory>

Create a PWA project in <directory> based on <template>.

Positionals:
  template   Name of a "template" to clone and customize. Currently the
             "venia-starter" template is supported: `buildpack init-project
             venia-starter <directory>     [required] [choices: "venia-starter"]
  directory  Name or path to a directory to create and fill with the project
             files. This directory will be the project root. [string] [required]

Project configuration:
  --backendUrl, -b  URL of the Magento 2.3 instance to use as a backend. Will be
                    added to `.env` file.
  --customOrigin    Create a custom secure host and certificate for this
                    project. Required for full PWA functionality. Requires
                    administrator privileges.          [boolean] [default: true]

Metadata:
  --name, -n    Short name of the project to put in the package.json "name"
                field. Uses <directory> by default.
  --author, -a  Name and (optionally <email address>) of the author to put in
                the package.json "author" field.

Package management:
  --install    Install package dependencies after creating project
                                                       [boolean] [default: true]
  --npmClient  NPM package management client to use.
                                       [choices: "npm", "yarn"] [default: "npm"]

Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]
```

- **TODO**: Add more subcommands to simpify scripts
- **TODO**: Tests and docs

This command is non-interactive. The friendly facade on top of it is the `create-pwa` package.

#### `npm init @magento/pwa`

NPM initializer, like `create-react-app`.

image

Could be expanded to other project templates.

### Changes
[...]

PR #1217, opened 2019-05-07
Refactor driver usage to improve Venia portability

zetlen:
_This PR is a cherry-pick from #1198 to make that PR more digestible._

It is number 3 of 6 in the chain of cherry-picked PRs. It needs to be merged before the next 3:

4. #1219 
5. #1224
6. #1198

## Description
When trying to use Venia components in a context with another app, it's easy to get React errors claiming you're trying to use a `Link` element without a `Router` in context. This mostly happens because we forgot a couple of spots where we are hardcoding a particular instance of React Router instead of pulling in the instance provided by `@magento/venia-drivers`. This PR does the refactoring and prop drilling necessary to pass drivers around, so that an app can share routing with other components.

## Related Issue

Closes #ISSUE_NUMBER.

## Verification Steps

1. Ensure `yarn build` is successful
1. Run `yarn watch:venia` and load up your local Venia in your browser.
1. Ensure the main (CMS) page works.
1. Click through to the Create Account page, ensure that works.
1. Execute a search and ensure that works.
1. Hit a page that doesn't exist (ex: `/noexist.html`) and ensure you see the "Something went wrong. Please try again" message.
1. Sign In and ensure you can pull up your account by clicking on your user information:

Screen Shot 2019-05-09 at 10 13 16 AM

## Checklist:


packages/peregrine/package.json
```diff
@@ -33,8 +32,8 @@
     "@babel/plugin-syntax-dynamic-import": "~7.2.0",
     "@babel/plugin-syntax-jsx": "~7.2.0",
     "@babel/plugin-transform-react-jsx": "~7.3.0",
-    "@babel/plugin-transform-runtime": "~7.3.4",
-    "@babel/runtime": "~7.3.4",
+    "@babel/plugin-transform-runtime": "~7.4.4",
+    "@babel/runtime": "~7.4.2",
```
supernova-at: Since it's listed in `peerDependencies` do we need `@babel/runtime` here?

packages/venia-concept/package.json
```diff
@@ -47,14 +44,14 @@
     "apollo-link-http": "~1.5.11",
     "braintree-web-drop-in": "~1.16.0",
     "graphql-cli": "~3.0.11",
-    "graphql-cli-validate-magento-pwa-queries": "~1.0.0",
+    "graphql-cli-validate-magento-pwa-queries": "~1.1.0-beta.1",
```
supernova-at: Was this change for testing purposes or do we actually want to commit it?

packages/venia-concept/src/components/App/renderRoutes.js
```diff
@@ -11,7 +11,11 @@ const renderRoutes = () => (
     <Switch>
         <Route exact path="/search.html" component={Search} />
         <Route exact path="/create-account" component={CreateAccountPage} />
-        <Route render={() => <Page>{renderRoutingError}</Page>} />
+        <Route
+            render={() => (
+                <Page using={{ Route, Router }}>{renderRoutingError}</Page>
```
supernova-at: More of a curiosity question: is it possible for the `Route` component being rendered here and the `Route` being passed in the `using` prop to `Page` to be different? Would you ever actually want to do that?

pwa-devdocs/src/peregrine/routing/index.md
```diff
@@ -144,6 +145,6 @@ class App extends Component {
 [`Page.js`]: https://github.com/magento-research/pwa-studio/blob/master/packages/peregrine/src/Page/Page.js
 [`react-router`]: https://github.com/ReactTraining/react-router
 [React Context]: https://reactjs.org/docs/context.html
-[ErrorView]:t://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/components/ErrorView/errorView.js 
+[ErrorView]: https://github.com/magento-research/pwa-studio/blob/master/packages/venia-concept/src/components/ErrorView/errorView.js
```
supernova-at: image

packages/peregrine/package.json
zetlen: Yeah, for development.

packages/venia-concept/src/components/App/renderRoutes.js
zetlen: It's _possible_. And honestly it's easier to just drill the props through. `<Page />` is its own component and could be used separately by something in the future, so rather than try to make a new ContextProvider and think about all those use cases, I thought props would be simpler code.

zetlen:
@supernova-at Thanks for reminding me about the beta version I left in there. I removed it and answered your question about passing Route around. I'd love it if you took another look (at this one before my others if possible) :)

dpatil-magento:
@zetlen    Home page does not load https://venia-e2aw4xl3e.now.sh/      console - Uncaught TypeError: Cannot read property 'location' of undefined

supernova-at:
> @zetlen Home page does not load https://venia-e2aw4xl3e.now.sh/ console - Uncaught TypeError: Cannot read property 'location' of undefined

I verified on my local machine. Not sure where the disconnect is there. `location` is likely React Router's `location`.

packages/peregrine/package.json
jimbo: @supernova-at  Entries in `peerDependencies` aren't installed when you run `yarn` (or `npm install`). They're needed for tests, though, so we have to list them in `devDependencies` as well.

zetlen:
@dpatil-magento I've refactored this to be much simpler. It appears to be working now in staging--please validate if you have the time!

@supernova-at @jimbo No more passing the Route and Router pair around. I figured out what was causing the realm mismatch: it was just bad versioning and `withRouter`.

dpatil-magento:
@zetlen Thanks, testing this one now.

packages/venia-concept/src/drivers/index.js
```diff
@@ -1,5 +1,5 @@
 export { Query } from 'react-apollo';
-export { Link, Redirect, Route } from 'react-router-dom';
+export { Link, Redirect, Route, Switch, withRouter } from 'react-router-dom';
```
jimbo: Consider `export *` here, maybe? This is such a major dependency—I think whitelisting only some of its exports qualifies as a leaky abstraction.

If a consumer of `venia-concept` wanted to import `useRouter()`, for example, they'd encounter the same "conflicting context" issues that this change is addressing.

packages/pwa-buildpack/package.json
```diff
@@ -40,6 +40,7 @@
     "babel-loader": "~8.0.5",
     "boxen": "~3.0.0",
     "chalk": "~2.4.2",
+    "css-loader": "^2.1.1",
```
jimbo: Is this actually a dependency of buildpack now?

jimbo: Also:
```suggestion
    "css-loader": "~2.1.1",
```

packages/venia-concept/package.json
zetlen: For testing. Removing it.

packages/venia-concept/src/drivers/index.js
zetlen: I'd rather export these things one by one to assist in tree shaking.

jimbo: Yeah, I ordinarily wouldn't recommend `export *` for that reason. I think this counts as a serious flaw in this approach, though. If we're not going to `export *`, then we should whitelist all of its exports. And we should consider doing the same for the other third-party libraries being exported here as well.

PR #1218, opened 2019-05-07
[feature]: Toasts

### LINKED ISSUE #783: [FEATURE] Toasts.
- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

The banner used to display online/offline text currently exists as a component that displays some text related specifically to whether the app is functioning in online or offline mode. It does not allow for any other text.

The toast/notification banner is useful as a generic component powered by some sort of event system in which components throughout the app could say something like `displayToast(type, message)` and the toast component would handle the display.

Some considerations for dev/ux:

* What states do we care about? Will "info" and "error" suffice or do we need warning, error, etc?
* How do we want to handle multiple toasts emitted at once? **Per advice on material and spectrum, never display multiple at once, queue and display once at a time.**
* How do we want to handle dismissal? Should they be on a timer? Should the user be able to dismiss manually? **Probably both. Soumya will provide mocks.**
* Should we power this via redux/context or use React Portals? **[Here's](https://codesandbox.io/s/qzn5om4y5q) an example using React Hooks.**
* How do we key the list of messages displayed? **timestamp**
* We should be conscious of a11y such as tab/keyboard nav and voiceover.

**Mocks**

[Image from Gyazo](https://gyazo.com/b17131ae4473b25366c7973c71e975e6)

(more coming later from @soumya-ashok)

**Additional context**

https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/App/app.js#L26

https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/OnlineIndicator/onlineIndicator.js

https://github.com/magento-research/pwa-studio/blob/release/2.0/packages/venia-concept/src/components/OnlineIndicator/onlineIndicator.css

https://codesandbox.io/s/w60xo8o3k7

sirugh:
# TODO LIST

- [x] ~CSS Transitions for toast addition and removal~ Decided that pop-in/out was fine.
- [x] Refactor error notification work from #871 to use toasts.
- [x] If identical toast is repeated _reset timeout_ and _flash the opacity_ to indicate reset
- [x] Dismissable toasts should also auto remove after _7 seconds_
- [x] Show online/offline indicator _along with_ online/offline toast.
- [x] Add a warning type (yellow?) toast.
- [x] export from peregrine properly
- [x] Cleanup (console logs, etc)
- [x] Tests
- [x] Handle desktop toasts (currently always centered) cc @soumya-ashok 

## Description

This PR brings you 🥁 🥁 🥁 🥁 🥁 🥁 toasts 🍞 !

Big changes:

1) Addition of a toast context/store along with hooks that enable usage of the store
1) Implementation of a "toast container" and toasts in Venia (html/css/etc)
1) Refactoring of error notifications to use toasts.
1) Refactoring of online indicator to use toasts.
1) Refactoring of online indicator to live in the header as just an offline icon (when offline).

Toasts should:

* Look delicious.
* Be removable by `timeout`, or some default (5 seconds).
* Have different states (info, error, etc)
* Be tested!

[Image from Gyazo](https://gyazo.com/2a6b7fc3fe20158690a8fa843c87574d)

[Image from Gyazo](https://gyazo.com/c8c60b14b454ac22abd77ba373fa4c50)

## Related Issue

Closes #783.

## Verification Steps

1. Open the app. Enable offline mode in dev tools. See offline toast and icon in header!
1. Disable offline mode. See online toast!
1. Add an error to `miniCart.js` and rerender. See an error toast!

## How Have YOU Tested this?
Manually, and with verification steps, and wrote unit tests.

## Checklist:

- [ ] I have updated the documentation accordingly, if necessary. paging @jcalcaben! Should we make changes to docs for toasts?


packages/venia-concept/src/components/App/app.js
```diff
@@ -43,6 +44,7 @@ class App extends Component {
                         errors={errors}
                         onDismissError={this.recoverFromRenderError}
                     />
+                    <ToastContainer />
```
sirugh: Should this live here or at the top level?

packages/peregrine/src/Toasts/useToastActions.js
```diff
@@ -0,0 +1,44 @@
...
+
+    const addToast = useCallback(
+        (type, message, dismissable, icon, actionText, actionCallback) => {
+            const id = Date.now();
+
+            // Queue to delete the toast after some time.
+            if (!dismissable) {
+                setTimeout(() => {
+                    removeToast(id);
+                }, 5000);
```
sirugh: Move this to a constant.

packages/peregrine/src/Toasts/useToastState.js
```diff
@@ -0,0 +1,34 @@
+import { useReducer } from 'react';
+const initialState = {};
+
+const reducer = (prevState = initialState, action = {}) => {
+    const { type, payload } = action;
+    console.log(payload);
```
sirugh: Remove

packages/venia-concept/src/components/Logo/logo.js
```diff
@@ -1,33 +1,62 @@
...
-        return (
-            <img
-                className={classes.logo}
-                src={logo}
-                height={height}
-                alt="Venia"
-                title="Venia"
-            />
+    // DELETE THIS BEFORE MERGE
+    const { addToast } = useToastActions();
```
sirugh: Delete all this

packages/venia-concept/src/components/OnlineIndicator/onlineIndicator.js
```diff
@@ -1,35 +1,38 @@
...
+                message: 'You are online.'
+            });
+        } else {
+            addToast({
+                type: 'error',
+                icon: CloudOffIcon,
+                message: 'You are offline. Some features may be unavailable.'
+            });
+        }
+    }, [isOnline]);
```
sirugh: @jimbo do I need to add `addToast` or the Icons here? I'm still conflicted about what to pass in this array.

packages/venia-concept/src/components/ToastContainer/toast.css
```diff
@@ -0,0 +1,129 @@
...
+    line-height: 1.25rem;
+    margin: 0 auto;
+    padding: 1rem;
+    width: 20rem;
+
+    border: 1px solid;
+    border-color: #d1d1d1;
+}
+
+/* TODO: Figure out how to pulse borders for "identical" react trees */
```
sirugh: @jimbo @supernova-at do ya'll have ideas about how to go about this? If a toast currently exists in the DOM and an identical toast is added (same props) Soumya asked that we reset the timeout and "pulse" the border.

This CSS pulses the border when a toast is first added but it does not pulse when I add a second duplicate toast. You can repro by checking the PR out at this state and running the app. You can see in `logo.js` that I add two identical toasts after a timeout, and though the removal timer is reset the pulse does not occur.

packages/peregrine/src/Toasts/useToastActions.js
```diff
@@ -0,0 +1,54 @@
...
+
+// If a toast _is not_ dismissable remove it in this many milliseconds.
+const DEFAULT_TIMEOUT = 5000;
+
+/**
+ * Generates an identifier for a toast in a pure manner. Should always return
+ * the same identifier for a given set of props.
+ */
+const getToastId = props => {
+    return Object.keys(props).reduce((acc, curr) => acc + props[curr]);
```
sirugh: There's gotta be a better way to do this efficiently and consistently over string concatenation...

packages/venia-concept/src/components/ToastContainer/toast.css
sirugh: Handled it by using a random key. 🤷‍♂

packages/venia-concept/src/components/ToastContainer/toastContainer.js
```diff
@@ -0,0 +1,41 @@
...
+import Toast from './toast';
+import { mergeClasses } from 'src/classify';
+import defaultClasses from './toastContainer.css';
+
+const ToastContainer = props => {
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const { toasts } = useToastStore();
+
+    // Given a map of toasts each with a property "timestamp", sort and display
+    // based on the timestamp.
```
sirugh: Should the responsibility of sorting be the responsibility of the display or the store? The store is a more opinionated location but in this case it seems like consumers of the store would probably like to just iterate over a sorted array for display purposes rather than have to write their own sorting function.

@supernova-at @jimbo @tjwiebell what do ya'll think?

packages/venia-concept/src/components/ToastContainer/toast.js
```diff
@@ -0,0 +1,54 @@
...
+import { useToastActions } from '@magento/peregrine/src/Toasts/useToastActions';
+import { mergeClasses } from 'src/classify';
+import Icon from 'src/components/Icon';
+
+const Toast = props => {
+    const { id, type, icon, message, actionText, dismissable } = props;
+    const classes = mergeClasses(defaultClasses, {});
+    const { removeToast } = useToastActions();
+
+    // TODO: Can these callbacks be defined within the addToast hook?
```
sirugh: If I can do this it won't be necessary for a toast to know its own `id`.

sirugh: But if we do move this to the toast action it becomes that much more opinionated about what happens when you click on an action or dismiss the toast.

Maybe there is a compromise that dismiss handler is made in the action but the action handler is defined within the toast, as that belongs to the consumer (in this case, Venia).

packages/venia-concept/src/components/App/errorNotifications.js
```diff
@@ -1,74 +1,50 @@
...
+            return dismissers.has(error)
+                ? dismissers.get(error)
+                : dismissers.set(error, () => onDismissError(error)).get(error);
+        },
+        [onDismissError]
+    );
+
+    return props.errors.map(({ error, id, loc }) => {
+        useEffect(() => {
+            // TODO: Without the enclosing useEffect this endless loops. Why?
```
sirugh: I don't understand why emitting the toast here with `addToast` within `<App />` could result in an endless loop. The only state/DOM that is updated should be in the top level `<ToastContainer />`. Is it because this all lives within `<ToastContextProvider>`? See [here](https://github.com/magento-research/pwa-studio/pull/1218/files#diff-1078ac6bf3201692a3f2202512f5b03dR41).

packages/venia-concept/src/components/Logo/logo.js
```diff
@@ -1,33 +1,79 @@
...
+import SmileIcon from 'react-feather/dist/icons/smile';
+import { useToastActions } from '@magento/peregrine';
+
+const Logo = props => {
+    const { height } = props;
+    const classes = mergeClasses({}, props.classes);
+
+    const { addToast } = useToastActions();
+
+    // TODO REMOVE THESE DEMO TOASTS
```
sirugh: I think I'll leave these in here so you guys can play around with them. before we merge I'll just go in and delete the code related to these demo toasts.

packages/venia-concept/src/components/ToastContainer/toast.js
```diff
@@ -0,0 +1,82 @@
...
+        onAction,
+        onDismiss,
+        type
+    } = props;
+    const classes = mergeClasses(defaultClasses, {});
+    const { removeToast } = useToastActions();
+
+    let handleDismiss, handleAction;
+    if (dismissable) {
+        handleDismiss = useCallback(() => {
```
sirugh: Should these be the responsibility of the display (this Venia component) or the data layer (within the hook/reducer)? 

Personally I think the dismiss callback can be defined within the data since it will always be the same. The question becomes whether or not we think people will want to have their toasts removed immediately when the action text is clicked or whether they want it to do something else (and then be removed? or maybe stay?).

soumya-ashok:
@sirugh They look great to start with, awesome work!
I'll work on desktop presentation of these. The difference in color between the orange for warning and the red for error won't meet accessibility standards. We have not previously discussed a warning state but if we would like to include one, we should try another color.
Is it possible to slow the opacity effect a little on the repeat toasts? It might happen too quickly to register that it is a re-appearance.

sirugh:
> I'll work on desktop presentation of these.

Awesome. Definitely need that cause they look a bit weird popping in the middle of the screen 👍 

> The difference in color between the orange for warning and the red for error won't meet accessibility standards. We have not previously discussed a warning state but if we would like to include one, we should try another color.

Let me know what you'd like. It's easy enough to remove or to try other colors.

> Is it possible to slow the opacity effect a little on the repeat toasts? It might be happening too quickly to register that it is a re-appearance.

Honestly the animation of the toasts has added some complexity to the code that I do not like. Is it possible to make it _slower_ on repeat toasts only? Maybe, but I think it'll get even more complex. Right now the display doesn't know whether a toast is a duplicate or not, it just flashes on _every_ addition. We just "trick" it by making it think duplicates are new.

packages/venia-concept/src/components/App/errorNotifications.js
sirugh: Yea, I'd rather not enclose the `addToast` call with the hook but if I call it without `useEffect` the app infinitely loops and I don't know why.

packages/peregrine/src/Toasts/useToastContext.js
```diff
@@ -0,0 +1,127 @@
+import React, { createContext, useContext, useReducer } from 'react';
+/**
+ * The state of the toast store.
+ * @typedef {Object} ToastState
+ * @property {Object} toasts - unordered map of id:toast
```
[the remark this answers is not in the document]
sirugh: I was under the impression that objects/immutables are preferred for state values. Using a map for `toasts` would prevent me from being able to simply spread the previous state or the new state. I'd probably have to copy the map from the previous state, modify the toast, set that value in the copy, and then assign the copy to the `toasts` object in state. Is there a simpler way that I'm missing?

packages/peregrine/src/Toasts/useToastActions.js
```diff
@@ -0,0 +1,107 @@
...
+import { useToastDispatch } from './useToastContext';
+
+// If a toast _is not_ dismissable remove it in this many milliseconds.
+const DEFAULT_TIMEOUT = 5000;
+
+/**
+ * Generates an identifier for a toast in a pure manner. Should always return
+ * the same identifier for a given set of props.
+ */
+const getToastId = props => {
```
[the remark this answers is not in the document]
sirugh: OK. I'll pull it into a `toast/util` module and export it for testing 🍞

packages/peregrine/src/Toasts/useToastContext.js
```diff
@@ -0,0 +1,127 @@
...
+
+            // For duplicate toasts, do not update the timestamp to maintain
+            // order of toast emission.
+            const timestamp = duplicate
+                ? prevState.toasts[payload.id].timestamp
+                : Date.now();
+
+            // Use a random key to trigger a recreation of this component if it
+            // is a duplicate so that we can re-trigger the blink animation.
+            const key = duplicate ? Math.random() : payload.id;
```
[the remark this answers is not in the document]
sirugh: Is using `Date.now()` to create the timestamp also bad form?

Another idea: 

1) Move `timestamp` creation to the action and use that as the ID for the toast store. 
2) In the reducer, if duplicate, reset the timestamp to the previous timestamp (otherwise leave it).
3) In the display, `toastContainer`, if we see `duplicate: true`, generate a random key.

This would still have a side-effect of modifying the incoming `timestamp` value in the event of a duplicate...

packages/peregrine/src/Toasts/useToastContext.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @example
+ *   <ToastContextProvider>
+ *     <ToastContainer /> // A component which would display based on state.
+ *     <AddToastComponent /> // A component which adds a toast using actions.
+ *   </ToastContextProvider>
+ */
+export const ToastContextProvider = ({ children }) => {
+    return (
+        <ToastContext.Provider value={useReducer(reducer, initialState)}>
```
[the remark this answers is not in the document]
sirugh: Maybe we should add `eslint-plugin-react-hooks` to our rules :D

packages/venia-concept/src/components/App/errorNotifications.js
```diff
@@ -1,74 +1,50 @@
...
+    const getErrorDismisser = useCallback(
+        error => {
+            return dismissers.has(error)
+                ? dismissers.get(error)
+                : dismissers.set(error, () => onDismissError(error)).get(error);
+        },
+        [onDismissError]
+    );
+
+    return props.errors.map(({ error, id, loc }) => {
```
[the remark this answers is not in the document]
sirugh: I'm totally on board *however*... `App` is an error boundary due to its use of `getDerivedStateFromError`, which [doesn't yet have a hook implementation](https://reactjs.org/docs/hooks-faq.html#how-do-lifecycle-methods-correspond-to-hooks). I could add another layer around `App` whose sole purpose is to be an `ErrorBoundary` and then I could pass `renderError` down as props to a functional `App` component.

packages/peregrine/src/Toasts/useToastContext.js
```diff
@@ -0,0 +1,127 @@
...
+            };
+
+            // It is possible to attempt to delete a non-existant toast so let's
+            // gracefully handle that.
+            if (newState.toasts[payload.id]) {
+                // Clear the old toast removal timeout incase an identical toast
+                // is added after this.
+                window.clearTimeout(
+                    newState.toasts[payload.id].removalTimeoutId
+                );
```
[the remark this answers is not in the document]
sirugh: That would require access to the `removalTimeoutId` from the toast in the store. We don't have access to that in the `removeToast` function.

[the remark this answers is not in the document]
jimbo: Perhaps we should, though. If we follow the pattern from `useQueryResult`, the following is possible:

```es6
export const useToasts = () => {
  const [state, dispatch] = useReducer(reducer, initialState)
  const { toasts } = state

  const addToast = useCallback(/* ... */)

  const removeToast = useCallback(
    () => {
      payload => {
        const timeoutId = toasts.get(payload)

        // `timeoutId` doesn't have to match a timeout
        // in fact, it can even be `null` or `undefined`
        window.clearTimeout(timeoutId)

        dispatch({
          payload,
          type: "remove"
        })
      }
    },
    [toasts]
  )

  const api = useMemo(
    () => ({
      addToast,
      dispatch,
      removeToast
    }),
    []
  )

  return [state, api]
}
```

packages/peregrine/src/Toasts/useToastContext.js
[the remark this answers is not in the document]
jimbo: Yes, that's my fault for promising to do it and then not doing it. 😓

packages/venia-concept/src/components/App/errorNotifications.js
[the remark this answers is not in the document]
jimbo: `App` should really be the top level of hierarchy, and it shouldn't have any siblings. I do think wrapping things in an `ErrorBoundary` is a good idea, though. Will this work?

```jsx
// App.js

return (
  <ErrorBoundary>
    <Main />
    <Mask />
    <Toasts />
  </ErrorBoundary>
)
```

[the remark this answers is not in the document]
sirugh: I refactored this but ended up not adding a new callback for now. I'm fine just doing `toasts.forEach(addToast)` for now, given that `addToast` returns the `id` and the app may want to do something with that in the future.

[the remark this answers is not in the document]
sirugh: > Will this work?

That's basically what we did, except that logic now lives in the `AppContainer` in `container.js`.

From [the docs](https://reactjs.org/docs/error-boundaries.html):
> A class component becomes an error boundary if it defines either (or both) of the lifecycle methods static getDerivedStateFromError() or componentDidCatch()

packages/peregrine/src/Toasts/useToastContext.js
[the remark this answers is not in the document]
sirugh: Got it. I'll give this a shot.

packages/peregrine/src/Toasts/useToastContext.js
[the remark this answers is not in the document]
jimbo: > Is using Date.now() to create the timestamp also bad form?

Yes. A reducer should be idempotent: call it with the same arguments (`payload` and `type`) and it should return the same result every time. That won't happen if something in the reducer calls `Math.random()` or `Date.now()`, since those return a different value on every call.

The solution is usually to handle randomness in the action creator.

packages/peregrine/src/Toasts/useToastContext.js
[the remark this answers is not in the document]
jimbo: > I was under the impression that objects/immutables are preferred for state values.

Redux prefers plain objects because they're serializable and compatible with its devtool extensions. Also, a Redux state tree tends to be very deep and gets sliced a lot. I don't think that strictness is quite as important for local state reducers, especially when maps are more suitable under the conditions I listed above.

To be clear, neither a plain object nor a `Map` are immutable, we just treat them as if they were.

```es6
const initialState = {
  toasts: new Map()
}

const reducer = (state, action = {}) => {
  const { payload, type } = action

  switch (type) {
    case "remove": {
      // a map is iterable, and each entry is a `[key, value]` array
      // which happens to be the shape expected by its constructor
      // so you can always call `new Map(currentMap)`
      const nextToasts = new Map(state.toasts)

      // the `delete` method returns a boolean
      // `true` if an entry was deleted, `false` if not
      // but it won't throw if it doesn't find an entry
      nextToasts.delete(payload)

      // we still spread the state object
      // since we only used a map for the toasts
      return {
        ...state,
        toasts: nextToasts
      }
    }
  }
}
```

packages/peregrine/src/Toasts/useToastContext.js
```diff
@@ -0,0 +1,125 @@
...
+
+                // Delete the toast from the store.
+                delete nextState.toasts[payload.id];
+            }
+
+            // Do not blink on removal.
+            Object.keys(nextState.toasts).forEach(key => {
+                nextState.toasts[key].duplicate = false;
+            });
```
jimbo: On further review, I noticed this is actually mutating the toast objects, which we can't do. 😞 

```es6
const toasts = {
  a: { duplicate: true },
  b: { duplicate: true },
  c: { duplicate: false },
  d: { duplicate: false },
}

const mutate = (id) => {
  const nextToasts = { ...toasts }

  delete nextToasts[id]

  Object.keys(nextToasts).forEach(key => {
    nextToasts[key].duplicate = false
  })

  return nextToasts
}

test("shouldn't mutate objects", () => {
  const { a: before } = toasts
  const { a: after } = mutate("d")

  // sets `duplicate` to `false`, as intended
  expect(before.duplicate).toBe(true)   // ✅
  expect(after.duplicate).toBe(false)   // ✅

  // but doesn't change object identity
  expect(before).not.toBe(after)        // ❌
})
```

sirugh: This is something I did in an attempt to get the "blink" animation to work properly. Without that side effect, all the toasts were blinking because the toast container is essentially re-rendering/recomputing the list on every mutation. I'd really like to make this less complex.

sirugh: Edit: Actually wait, I think I remedied that by using the key. Why did I leave this in?

packages/peregrine/src/Toasts/useToastContext.js
```diff
@@ -0,0 +1,99 @@
...
+        case 'add': {
+            const nextToasts = new Map(prevState.toasts);
+            const prevToast = nextToasts.get(payload.id);
+
+            const isDuplicate = !!prevToast;
+            let timestamp = payload.timestamp;
+            if (isDuplicate) {
+                // If this is a _new_ duplicate toast we need to clear the
+                // previous timeout to prevent premature removal.
+                window.clearTimeout(prevToast.removalTimeoutId);
```
sirugh: @jimbo as we discussed I moved the references to previous state out of the action and into the reducer. But this is also undesirable as we are now have a side effect.

packages/venia-concept/src/components/Logo/logo.js
```diff
@@ -1,33 +1,73 @@
...
-        }),
-        height: PropTypes.number
+import SmileIcon from 'react-feather/dist/icons/smile';
+import { useToasts } from '@magento/peregrine';
+
+const Logo = props => {
+    const { height } = props;
+    const classes = mergeClasses({}, props.classes);
+
+    const [, { addToast }] = useToasts();
```
jimbo: Don't forget to remove all of this stuff.

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -233,6 +233,7 @@ class MiniCart extends Component {
     }
 
     render() {
+        // throw new Error('oops');
```
jimbo: Remove this line as well.

packages/venia-concept/src/components/OnlineIndicator/onlineIndicator.css
```diff
@@ -1,67 +1,8 @@
...
-        height: 3rem;
-    }
-    100% {
-        height: 0;
-    }
+    display: flex;
+    grid-area: primary;
+    height: 3rem;
+    margin-left: 4.5rem;
+    width: 3rem;
```
jimbo: Is this no longer going to be a custom indicator? Are we just making it a toast now?

packages/venia-concept/src/index.js
```diff
@@ -37,7 +38,10 @@ ReactDOM.render(
         store={store}
     >
         <WindowSizeContextProvider>
-            <App />
+            <ToastContextProvider>
+                <ToastContainer />
```
jimbo: As mentioned before, I would really like this to be a child of `App`, rather than a sibling. Is that doable?

packages/venia-concept/src/components/ToastContainer/toastContainer.js
jimbo: In general, I think the store should keep records in a canonical order—in this case, the order in which the toasts are added. Since we're storing toasts in a map, that happens to be the natural order anyway. If a view wants to sort them differently, that should be the prerogative of the view.

packages/venia-concept/src/components/ToastContainer/toastContainer.js
```diff
@@ -0,0 +1,42 @@
...
+    const [{ toasts }] = useToasts();
+    // Given a map of toasts each with a property "timestamp", sort and display
+    // based on the timestamp.
+    const timestampMap = new Map();
+    toasts.forEach(toast => {
+        timestampMap[toast.timestamp] = toast;
+    });
+
+    const toastList = Object.keys(timestampMap)
+        .sort()
```
jimbo: I think we can cut out one pass here, and stick to iterables in the process. Maybe consider something along the lines of the following?

```jsx
const sortByTimestamp = ([, toastA], [, toastB]) => toastB.timestamp - toastA.timestamp

const toastElements = Array.from(toasts)
    .sort(sortByTimestamp)
    .map(([id, toast]) => {
        const key = toast.isDuplicate ? Math.random() : id

        return (
            <Toast key={key} {...toast} />
        )
    })
```

packages/venia-concept/src/components/ToastContainer/toastContainer.css
```diff
@@ -0,0 +1,13 @@
+.container {
+    position: fixed;
+    min-width: 100%;
+    display: grid;
+    grid-row-gap: 1rem;
+
+    /* Necessary to display over page contents */
+    z-index: 9999;
```
jimbo: `5` or `6` should be sufficient, I would think. 😂

packages/venia-concept/src/components/ToastContainer/toastContainer.css
```diff
@@ -0,0 +1,13 @@
...
+
+    /* Show toasts on bottom of screen */
+    bottom: 0px;
+    margin-bottom: 15px;
```
jimbo: Typically we stick to a `16px` grid for lengths. Try `1rem` maybe?

packages/venia-concept/src/components/ToastContainer/toast.js
```diff
@@ -0,0 +1,82 @@
...
+                />
+            ) : null}
+            <div className={classes.message}>{message}</div>
+            {dismissable ? (
+                <div className={classes.controls}>
+                    <button
+                        className={classes.dismissButton}
+                        onClick={handleDismiss}
+                    >
+                        {'×'}
```
jimbo: I think we should use an icon here instead.

packages/venia-concept/src/components/ToastContainer/toastContainer.js
zetlen: First of all, there should never be enough toasts that sorting is a significant cost. Definitely sort them by timestamp! They can re-sort if they want, and they should never want to.

Also: it seems like these next few lines are doing the same thing as `Array#sort(comparator)`. How about this?

```js
    const toastList = toasts.slice()
        .sort((a, b) => a.timestamp - b.timestamp)
        .map(toast => {
            // Use a random key to trigger a recreation of this component if it
            // is a duplicate so that we can re-trigger the blink animation.
            const key = toast.isDuplicate ? Math.random() : toast.id;
            return <Toast key={key} {...toast} />;
        });
```

Seems like it does the same thing unless I'm missing something. This should still go in the store, though 🙃

zetlen:
One other thing: can we leave the OnlineIndicator the way it was? Unless @soumya-ashok approved this visual change already, I think we should put that refactor in a separate PR.

packages/peregrine/src/Toasts/__tests__/useToasts.spec.js
```diff
@@ -0,0 +1,102 @@
...
+
+const log = jest.fn();
+const dispatchMock = jest.fn();
+useToastDispatch.mockReturnValue(dispatchMock);
+useToastState.mockReturnValue({ toasts: {} });
+
+beforeEach(() => {
+    log.mockReset();
+    dispatchMock.mockReset();
+});
```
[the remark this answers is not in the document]
sirugh: I did not know that! I will remove these additions, and we should probably go through and remove them from the codebase in general (but in another PR).

packages/peregrine/src/Toasts/useToastContext.js
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * A hook that provides access to the toast dispatch function.
+ * !Any component using this hook _must_ be a child of a `ToastContextProvider`.
+ *
+ * @return {Function} The action dispatch function.
+ * @example
+ *   const dispatch = useToastDispatch();
+ */
+export const useToastDispatch = () => useContext(ToastContext)[1];
```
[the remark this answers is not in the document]
sirugh: That makes sense. I'll reset this back to that pattern. Funny enough this will be the third time switching :D

packages/peregrine/src/Toasts/useToasts.js
```diff
@@ -0,0 +1,123 @@
...
+import { useToastState, useToastDispatch } from './useToastContext';
+
+// If a toast _is not_ dismissable remove it in this many milliseconds.
+const DEFAULT_TIMEOUT = 5000;
+
+/**
+ * Generates an identifier for a toast in a pure manner. Should always return
+ * the same identifier for a given set of props.
+ */
+export const getToastId = props => {
```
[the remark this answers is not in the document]
sirugh: Unfortunately I wanted to base it on all the possible configurable props. Before I added the hashing function I was simply concatenating the string values (see `combined` below) and using that as the ID but it made debugging a bit difficult.

All that to say that I would prefer to keep this as is but I also understand the issue with the speed of the function given that `combined.length` is pretty much always going to depend on how big the icon function is.

packages/venia-concept/src/components/OnlineIndicator/onlineIndicator.css
sirugh: The indicator still exists though it only displays when offline and it now appears as an icon in the header to the left of the logo. Try offline mode and you'll see.

packages/venia-concept/src/components/ToastContainer/toast.css
```diff
@@ -0,0 +1,125 @@
...
+    }
+    50% {
+        opacity: 0.5;
+    }
+    100% {
+        opacity: 1;
+    }
+}
+
+.toast--info {
```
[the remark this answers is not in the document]
sirugh: Yea, it's BEM. I can simplify this.

packages/venia-concept/src/components/ToastContainer/toastContainer.css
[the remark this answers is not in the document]
sirugh: Yea, it should appear on top of the mask, especially in the case where the application has thrown a render error.

sirugh: 3 years later...

> *[BUG]: Toasts are not displayed over my modal!

sirugh: But I can change it :P

sirugh:
> One other thing: can we leave the OnlineIndicator the way it was? Unless @soumya-ashok approved this visual change already, I think we should put that refactor in a separate PR.

That change was discussed and designed with @soumya-ashok's blessing :)

packages/venia-concept/src/components/Logo/logo.js
```diff
@@ -1,33 +1,73 @@
...
-class Logo extends Component {
-    static propTypes = {
-        classes: PropTypes.shape({
-            logo: PropTypes.string
-        }),
-        height: PropTypes.number
+import SmileIcon from 'react-feather/dist/icons/smile';
+import { useToasts } from '@magento/peregrine';
+
+const Logo = props => {
```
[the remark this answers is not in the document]
sirugh: I'll add this!

packages/venia-concept/src/components/ToastContainer/toast.js
sirugh: I like the idea of allowing for an async callback. I'll experiment with this.

sirugh:
@zetlen @jimbo all feedback has been addressed or I've responded. Should be ready for another look.

packages/peregrine/src/Toasts/useToasts.js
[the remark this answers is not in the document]
zetlen: The reason this code exists is to prevent confusing the shopper with identical _looking_ toasts, right? Right now, you're not doing referential comparison against the handler functions (and it wouldn't be safe to do so anyway, since they're often bound functions) so you're only comparing on visible properties. And that makes sense to me. Why not make this function just deduplicate _visually_ identical toasts?

packages/venia-concept/src/components/ToastContainer/toastContainer.css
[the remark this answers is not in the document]
zetlen: What if there is a modal that should appear above any state? I don't think we should zip right to z-index 9999, is all.

packages/peregrine/src/Toasts/useToasts.js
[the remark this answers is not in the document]
sirugh: Oh I see why we are confused :D This does indeed look at all the props passed to `addToast` and therefor at the handlers which it _should not_. However, one of those props is the `icon` which is actually a function component returned from `react-feather`. So I could use your suggestion but I would need to add `icon` to the joined values.

I get it, and I will do it. Thanks!

packages/venia-concept/src/components/ToastContainer/toastContainer.css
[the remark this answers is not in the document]
sirugh: What if there is a toast that should appear above any state? :D I changed this to 6.

dpatil-magento:
@sirugh - Observations - 
1. Toast disappears for a moment and appears.
2. Online Toast content is not same as mock-up https://magento.invisionapp.com/share/J8RYQ176SVB#/screens/330015009

sirugh:
> Toast disappears for a moment and appears.

This is intentional. When a toast is added it should "blink".

> Online Toast content is not same as mock-up https://magento.invisionapp.com/share/J8RYQ176SVB#/screens/330015009

Ah, the icon changed! I will update.

dpatil-magento:
@sirugh How about message content?  "You are online" vs "You are back online"

sirugh:
> How about message content? "You are online" vs "You are back online"

Fixed.

soumya-ashok:
@sirugh Sorry to ask for this to be changed again, but "You are online" would work better for most use-cases since it is more generic. Updated mockup -

dpatil-magento:
@sirugh QA Pass but need @supernova-at review approval to merge.

sirugh:
> Is it possible to add a user-dismissable toast that doesn't get auto-dismissed? If timeout is 0 or falsy at all, it should probably not set a timeout, right?

@soumya-ashok  and I discussed this, though not sure if we documented it anywhere. Basically all toasts should auto-dismiss for now and the answer for user-dismissable toasts would be to set some long timeout. I can add a workaround to this that does not set the timeout if `timeout === 0`, but I don't think `dismissable: true` should have any bearing on auto-dismissal.

> If a toast has a timeout, and a user tries to copy some text from it to their clipboard, won't the toast often disappear before they can do so? In most apps, I think a toast's timeout is cleared when the user interacts with that toast at all.

This is a good point. We would need some mechanic to pause the timeout. [react-toast-notifications](https://github.com/jossmac/react-toast-notifications/blob/0f07028a48f47caa6e72470e46f4c5fec14b21d3/src/ToastController.js#L81) does this and we could too. If you feel strongly about this I can do some work to get this functionality in, otherwise I would say that we hardcode some longer `timeout` values into the emissions of toasts we expect users to want to copy and then reduce them later. If we chose this option I will create an issue for pausing the timeout. Let me know.

> The icon prop of toasts is too tightly coupled to our visual presentation of toasts, I think. I'd prefer to use the type enum to determine what icons to render; if end users want to change the presentation, they'll override the Toast component or use a hook version of it from Peregrine. As is, icon would effectively be a render prop, which we're avoiding. Also, we probably don't want to store functions in state, even if this isn't Redux.

The way we create/use icons in our application is that we have the component who needs it import/use it. We can't know all the different icons that toast emitters want to put into their implementations, so we allow them to pass the whole icon. This is definitely a tight coupling to the display, but intentionally so that the display doesn't have to import every possible icon. We *could* have the toast component enumerate a subset of icons and basically ask anyone who reimplements Toasts to provide an icon enums as well. Again, we can talk about this before I make changes.

jimbo:
> I can add a workaround to this that does not set the timeout if timeout === 0, but I don't think dismissable: true should have any bearing on auto-dismissal.

I like this approach.

> I would say that we hardcode some longer timeout values into the emissions of toasts we expect users to want to copy and then reduce them later. If we chose this option I will create an issue for pausing the timeout.

I also like this approach.

> We *could* have the toast component enumerate a subset of icons and basically ask anyone who reimplements Toasts to provide an icon enums as well.

This is exactly what I'd like to do. We already have the enum with `type`, and the icon should be a reflection of `type`. More complicated functionality should be left to the re-implementation, I think. Thoughts?

sirugh:
> This is exactly what I'd like to do. We already have the enum with type, and the icon should be a reflection of type. More complicated functionality should be left to the re-implementation, I think. Thoughts?

Can you point me at the `type` you are talking about?

jimbo:
> Can you point me at the `type` you are talking about?

[Here](https://github.com/magento-research/pwa-studio/pull/1218/files#diff-9726897b28048b54e7680b2ae7dbc4bfR69).

| type |
| :--- |
| `info` |
| `warning` |
| `error` |

sirugh:
Oh so the type of the toast does not indicate the icon, unfortunately. Anyone could emit a sad face icon with an error toast to indicate some error or a cloud-off icon with an error toast to indicate offline.

sirugh:
Created #1284 for timeout pausing.

PR #1229, opened 2019-05-09
Product Out of Stock Message

### LINKED ISSUE #1159: [bug]: If user tries to access product page after it goes to out of stock then system enters a infinite loop.
Steps - 
1. Access a product details page say for Product A.
2. Login to backend and make Product A as out of stock.
3. Now either refresh the product details page on storefront venia or try accessing the product again.

Expected - System should throw error message indicating product not available.
Actual - System show error notification and if user click on in the same error re-appears. User has to clear browser cache to make it work.

Screen -

sudeep-cedcoss:
## Description

This PR fixes an issue where the application could enter an infinite loop of error notifications when a product becomes out of stock.

## Related Issue

Closes #1159 .

## Verification Steps

📝  Any product should work but these steps use the `Augusta Earrings`.

### Product In Stock
1. Go to the Product page for Augusta Earrings
1. Verify that it shows as expected

### Product Out of Stock after being In Stock
1. Log in to your Magento backend (`MAGENTO_SITE_URL` + `/admin` in your browser)
1. Click `Category` > `Products`
1. In the search field, enter `VA12-SI-NA` (the SKU for Augusta Earrings)
1. Click `Edit`
1. Change `Quantity` to `0` and `Stock Status` to `Out of Stock`
1. Click `Save`
1. Refresh the Product page in Venia
1. Verify that a "Product out of stock" message appears

📝 Sometimes it takes a second for the cache to update and the "out of stock" message to appear.

### Product Back in Stock
1. Log in to your Magento backend (`MAGENTO_SITE_URL` + `/admin` in your browser)
1. Click `Category` > `Products`
1. In the search field, enter `VA12-SI-NA` (the SKU for Augusta Earrings)
1. Click `Edit`
1. Change `Quantity` to `100` and `Stock Status` to `In Stock`
1. Click `Save`
1. Refresh Venia
1. See that the product page shows as expected

📝 For some reason it definitely takes longer for the cache to flush when the product comes back in stock as opposed to going out of stock. I am honestly unsure the cache timings and where the caching is taking place. If you want to shortcut the process, delete the `apollo-cache-persist` entry from local storage on your browser (Application > Local Storage) and refresh Venia using `Empty Cache and Hard Reload` (long press on the refresh button).

## How Have YOU Tested this?

1. `yarn test`
1. Verification Steps above

### Infinite Error Loop (before)

image

### Out of Stock Message (after)

Screen Shot 2019-06-04 at 10 29 36 AM

This PR adds a new `prop` to the `ErrorView` component. Additions are backwards-compatible.

## Checklist:


sirugh:
Hello! Is there a reason you're closing and opening the same changeset? You previously opened #1202 #1228 for this issue. Just wondering if there was a particular reason as I couldn't see any difference in the code.

sudeep-cedcoss:
Hello Stephen,

I have just started on Magento PWA a few weeks back and those pull request were among my starting pull requests. So I was not sure how to create a pull request properly and link them properly with original issue. So that's why I created some of the pull requests by mistake. Sorry for that. Now I am getting more habitual with these things, and I will make sure that this does not happen again.

Thank you.

sirugh:
No problem! Don't worry about the linking -- we can handle that on our end. I recommend a workflow like this:

1) Find an issue that sparks your interest. Comment in the issue saying that you'd like to work on it.
2) Branch off develop like `git checkout origin/develop && git checkout -b my_branch_name`.
3) Code away and when done, push the branch.
4) Fill out the PR template as descriptively as possible. The more detail the easier it will be for us to review and get your changes merged!

Good luck.

sudeep-cedcoss:
Ok Thanks. I'll follow these things while creating PRs in the future. :-)

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import { string, func } from 'prop-types';
 
+import { isUndefined } from 'util';
```
supernova-at: No need for this guy, we can check if an object is `undefined` easily without pulling this in.

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -32,12 +34,17 @@ class Product extends Component {
 
     // map Magento 2.3.1 schema changes to Venia 2.0.0 proptype shape to maintain backwards compatibility
     mapProduct(product) {
-        const { description } = product;
-        return {
-            ...product,
-            description:
-                typeof description === 'object' ? description.html : description
-        };
+        if (!isUndefined(product)) {
```
supernova-at: Let's do our check for `undefined` in the `render` method instead of here.

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -51,13 +58,17 @@ class Product extends Component {
...
 
                     const product = data.productDetail.items[0];
-
-                    return (
-                        <ProductFullDetail
-                            product={this.mapProduct(product)}
-                            addToCart={this.props.addItemToCart}
-                        />
-                    );
+                    const productData = this.mapProduct(product);
```
supernova-at: Here is the result of the query for a product that is out of stock:

```json
{
  "data": {
    "productDetail": {
      "items": []
    }
  }
}
```

Therefore `const product = data.productDetail.items[0]` will result in `product` being `undefined` here and your conditional can just be `if (product) { ...`

packages/venia-concept/src/RootComponents/Product/Product.js
supernova-at: Also, just an FYI that `isUndefined` is deprecated: https://nodejs.org/api/util.html#util_util_isundefined_object.

packages/venia-concept/src/RootComponents/Product/Product.js
```diff
@@ -51,13 +58,17 @@ class Product extends Component {
...
+                    if (productData) {
+                        return (
+                            <ProductFullDetail
+                                product={this.mapProduct(product)}
+                                addToCart={this.props.addItemToCart}
+                            />
+                        );
+                    } else {
+                        return <div>Product Out Of Stock</div>;
```
supernova-at: We have a very basic [ErrorView component](https://github.com/magento-research/pwa-studio/blob/develop/packages/venia-concept/src/components/ErrorView/errorView.js) that would work nicely here. For consistency and ease of update later, let's use that.

Feel free to add an `outOfStock` (boolean) prop to it that shows this message when the prop is `true`.

Then you can just render `<ErrorView outOfStock={true} />` here 👍

supernova-at:
It took us a while to give feedback on this so after waiting a week I went ahead and applied the feedback myself. Thanks for your contribution!

PR Updated:
  * Applies feedback to simplify logic in `Product.js`
  * Adds `outOfStock` to `ErrorView`
  * Updates PR description / verification steps

packages/venia-concept/src/components/ErrorView/errorView.js
```diff
@@ -5,15 +5,21 @@ import { loadingIndicator } from 'src/components/LoadingIndicator';
...
 class ErrorView extends Component {
     render() {
-        const { loading, notFound } = this.props;
+        const { loading, notFound, outOfStock } = this.props;
         const message = loading
             ? messages.get('loading')
             : notFound
             ? messages.get('notFound')
+            : outOfStock
+            ? messages.get('outOfStock')
```
sirugh: We may want to allow an error string rather than using this prop. This will allow us to separate the error messages and types from the `ErrorView` implementation.

```js
export const ERROR_TYPES = {
  'LOADING': 'loading',
  'OUT_OF_STOCK': 'outOfStock',
  ...
};
export const errorMap = new Map({
  'loading': loadingIndicator,
  'outOfStock': 'This Product is currently out of stock. Please try again later.',
  'default': 'Something went wrong. Please try again.'
  ...
});

class ErrorView extends Component {
  render() {
    const { type } = this.props;
    const key = ERROR_TYPES[type] || 'default';
    return <h1>{errorMap[key]}</h1>;
  }
} 
```

supernova-at: Yeah the `ErrorView` certainly could use a refactor. I already felt like adding `outOfStock` was a little scope creep here and refactoring `ErrorView` entirely would be out of scope entirely.

supernova-at: I can create a follow up ticket to refactor `ErrorView` for sure

supernova-at:
PR Updated:
  * Adds Unit Tests to `ErrorView` to complete the checklist

packages/venia-concept/src/components/ErrorView/errorView.js
sirugh: That's cool -- though the component is super small and I feel like the effort to refactor now is less than the effort to create, groom and work on it later. But w/e you want to do is cool, you're the one doing this work now :D

soumya-ashok:
@supernova-at @sirugh - I think this out-of-stock message should load as a toast, since that is how we are dealing with all of the other alert messages. Thoughts?

sirugh:
Yes! I think this is a perfect example for a toast. In order to do this properly we need to:

1) Decide on a UX for the toast ie. message, icon, etc.
2) Decide at what point we want to emit the toast ie within this product component.
3) Refactor the emitting/implementing component to be a functional component so that we can use the `useToasts` hook.

We will probably need to do something like navigate back if we emit a toast rather than show the "empty/no product" view.

The design/discussion should happen in the issue, #1159, and then we can go from there. Just to prevent all this extra work being done _after_ time has already been spent on coding I would recommend always talking to UX in the issue (use the `needs-ux` label) when you recognize a bug that is user-facing/front end.

soumya-ashok:
@sirugh After a discussion with @awilcoxa - I will open a new issue with a clearly defined user story for this because there is some complexity for us to think through from an experience perspective. For now, I've created a mockup that addresses this bug specifically.

brendanfalkowski:
@soumya-ashok + @sirugh + @supernova-at — A couple thoughts on inventory that vary across sites I've worked on:

1. If "in stock" status is treated as the default some sites don't present this state to customers (but they do show "out of stock").
1. Sometimes when "out of stock" then adding to cart is blocked, so usually this message is positioned close the CTA that's blocked.
1. Dismissing the inventory toast is weird, if that's the reason add-to-cart is blocked. As in-page content you'd always see it.
1. If showing quantity of "in stock" items is important it would be page-level content — so a toast for the opposite state might be weird.
1. If "out of stock" products can be purchased, then the warning looks a little aggressive.
1. For sites with multiple physical locations, we sometimes need to show "store X in stock, store Y out of stock, global available to ship".

packages/venia-concept/src/components/ErrorView/errorView.js
supernova-at: A fair point, lol.

I created #1308 just to keep things moving though.

soumya-ashok:
> @soumya-ashok + @sirugh + @supernova-at — A couple thoughts on inventory that vary across sites I've worked on: 

_This is a stop-gap solution for the moment, and I agree @brendanfalkowski that we need to define a strategy first on how we deal with out-of-stock items in the shopper flow._

> 1. If "in stock" status is treated as the default some sites don't present this state to customers (but they do show "out of stock").
> 2. Sometimes when "out of stock" then adding to cart is blocked, so usually this message is positioned close the CTA that's blocked.

_My preference would be to proactively notify the user that an item is out of stock, and give them the option to save it to a list or be notified when it becomes available._

> 3. Dismissing the inventory toast is weird, if that's the reason add-to-cart is blocked. As in-page content you'd always see it.

_Right now we have all toasts set to auto-dismiss after a certain amount of time. In addition to the toast, we should probably have on-page text that says the item is out of stock if we block add to cart. I have seen a pattern where "Add to Cart" is disabled but "Add to List" is enabled for an out-of-stock item. Thoughts on this?_

> 4. If showing quantity of "in stock" items is important it would be page-level content — so a toast for the opposite state might be weird.

_Noted._

> 5. If "out of stock" products can be purchased, then the warning looks a little aggressive.

_I think this would be dependent on how a particular merchant deals with out of stock items. For Venia, we haven't decided yet whether or not we would allow out of stock items to be added to the cart._

> 6. For sites with multiple physical locations, we sometimes need to show "store X in stock for pickup, store Y out of stock for pickup, ready to ship".

_Great point, will make a note.

As always, thanks for the thoughtful feedback!_

brendanfalkowski:
@soumya-ashok Cool cool, inventory strategy varies a lot between merchants so definitely not expecting a one-size-fits-all solution. The toast feels good for a B2C experience.

For B2B, we try to keep the UI actions in-place but disabled, because we may block adding to cart for different contexts (EPA/brand/geo restrictions, discontinued but in stock at other locations, cannot show pricing due to account). We prefer to always show the "add to cart" action in the UI, so it's clear that exists but isn't accessible for XYZ reason. Example:

Screen Shot 2019-06-07 at 12 17 50 PM

PR #1240, opened 2019-05-14
Product Details Component to Hooks

### LINKED ISSUE #974: Peregrine: Product Details Component
This story represents the effort to translate the product details component from Venia into Peregrine. Additional details to follow pending publish of @jimbo blog post via gist.

supernova-at:
## Description

As part of the effort to transition components to using Hooks and identifying pieces of functionality that would be good fits for Peregrine, this PR transitions the Product Details Component.

:memo: 
No functionality actually moved out of this component and into Peregrine though. When reviewing, please do indicate any pieces you think could be moved to Peregrine.

## Related Issue

Closes #974 .

### Non-Configurable Products

1. Go to the "Accessories" category and click an item
1. Verify that the product details page has no errors
1. Verify that you can add the item to your cart

### Configurable Products

1. Go to the "Bottoms" category and click an item
1. Verify that the product details page has no errors
1. Verify that you can select item options
1. Verify that you can add the item to your cart only once you have selected item options
1. Verify that the item variant in your cart matches the item selections you made

### Check Out

After adding both a configurable and non-configurable product to your cart, ensure that you can complete the checkout flow successfully.

## How Have YOU Tested this?

1. `yarn test`
1. The verification steps above

The product details page:

Screen Shot 2019-05-14 at 11 41 57 AM

## Checklist:


packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
```diff
@@ -95,11 +96,9 @@ test('Configurable Product has correct media gallery image count', async () => {
...
+    const productFullDetailComponent = root.children[0].children[0];
+    const carouselComponent =
+        productFullDetailComponent.children[0].children[1].children[0];
 
-    expect(instance.mediaGalleryEntries).toHaveLength(2);
-    instance.setState({
-        optionSelections: new Map([['1', 1]])
-    });
-    expect(instance.mediaGalleryEntries).toHaveLength(3);
+    expect(carouselComponent.props.images).toHaveLength(2);
```
supernova-at: We did lose an assertion here (the result of updating options) but since these components are functions as opposed to classes now, we don't have access to an `instance` on which to set `state`.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -1,254 +1,241 @@
...
+        }
 
-export default classify(defaultClasses)(ProductFullDetail);
+        addToCart(payload);
+    }, [addToCart, optionCodes, optionSelections, product, quantity]);
+
+    return (
+        <Form className={classes.root}>
+            <section className={classes.title}>
+                <h1 className={classes.productName}>{product.name}</h1>
```
supernova-at: Pointing out that heading tags no longer nest their contents inside of a `span`.  This used to look like:

```jsx
<h1 className={classes.productName}>
  <span>{product.name}</span>
</h1>
```

Let me know if those `span`s were there on purpose and we want them back.

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
sirugh: There is probably a way to do this but it'll involve mocking `useState` or somehow gaining access to the `setQuantity` callback and then using the that function within an `act`. Look at some of the useQueryResult.spec.js tests for inspiration :)

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
sirugh: Check that there is no style being applied somewhere to `productName > span` or `h1 > span`, etc. If not, then this is probably OK to remove!

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -1,254 +1,241 @@
...
+    return numProductSelections < numProductOptions;
+};
 
-        if (
-            !isConfigurable ||
-            (isConfigurable && optionSelections.size === 0)
-        ) {
-            return media_gallery_entries;
-        }
+const getMediaGalleryEntries = (product, optionCodes, optionSelections) => {
```
sirugh: Great job cleaning up this function! 👍

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -1,254 +1,241 @@
...
-        }
+    return { key, value };
+};
 
-        // Configurable products are missing options if we have fewer
-        // option selections than the product has options.
-        const { configurable_options } = product;
-        const numProductOptions = configurable_options.length;
-        const numProductSelections = this.state.optionSelections.size;
+const getProductOptions = (product, optionSelections, setOptionSelections) => {
```
sirugh: Any chance we can refactor this into a `ProductOptions` component and simplify the props? I think that'd look cleaner now that it basically is a function component.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -1,254 +1,241 @@
...
-                    <strong>{product.sku}</strong>
-                </section>
-            </Form>
-        );
-    }
-}
+    const handleSelectionChange = useCallback(
+        (optionId, selection) => {
+            // We must create a new Map here so that React knows that the value
+            // of optionSelections has changed.
```
sirugh: I don't follow -- is this some limitation of updating a Map within state? Why would you need to create a new map?

supernova-at: When I updated the existing `Map`, the component never re-rendered with the updated `optionSelections`.

I'm assuming that React is doing reference equality here and since the `Map` is the same map (just with different values in it), it thinks nothing changed and therefore doesn't update.

The same as how using `const` with objects works. 

```js
const a = {};
a.b = c;
```

The reference to `a` doesn't change, even though I've changed what `a` contains.

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
supernova-at: Resolved in https://github.com/magento-research/pwa-studio/pull/1240/commits/cf77f6f1f14c5164b902ad0ed10cdda1cba0face.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -1,254 +1,241 @@
...
-        // otherwise, cache attribute codes to avoid lookup cost later
-        for (const option of configurable_options) {
-            optionCodes.set(option.attribute_id, option.attribute_code);
-        }
-
-        return { optionCodes };
+const deriveOptionCodesFromProduct = product => {
+    // If this is a simple product it has no option codes.
+    if (!isProductConfigurable(product)) {
+        return INITIAL_OPTION_CODES;
```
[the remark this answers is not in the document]
supernova-at: I can't make this change because there are other downstream functions that expect `optionCodes` to be a `Map`, even if it is an empty one.

One example is [findMatchingVariant](https://github.com/magento-research/pwa-studio/blob/develop/packages/venia-concept/src/util/findMatchingProductVariant.js#L9).

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -1,254 +1,241 @@
...
-            return null;
-        }
-
-        // otherwise, cache attribute codes to avoid lookup cost later
-        for (const option of configurable_options) {
-            optionCodes.set(option.attribute_id, option.attribute_code);
-        }
-
-        return { optionCodes };
+const deriveOptionCodesFromProduct = product => {
```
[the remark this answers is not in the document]
supernova-at: I think this is really great feedback in the general case but I think the `ProductFullDetail` component is an exception to the rule. Here's why:

> See how initially state.optionCodes will be an empty map but any time props are provided we update the state with more and more option codes.

Great call but we don't actually want to do this. Playing that out, the list of option codes would just get bigger and bigger based on how many different `product`s got sent in as props, and the list would lose its meaning. We really only care about *this* product's option codes.

For an illustrative example, we determine if all options have been selected by comparing the number of selections (`optionSelections`) to the number of options available (`optionCodes`). We don't want the number of options available to include other products' options.

I think the reason this "bug" was never uncovered was because we never actually re-use this component and pass a new `product` prop to it. When users navigate away from the product detail page, `ProductFullDetail` is unmounted. When they navigate back, a new instance is mounted (with a different `product` prop).

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -1,254 +1,241 @@
...
-                Array.from(selection).pop()
-            )
-        }));
-    };
-
-    get fallback() {
-        return loadingIndicator;
+    // Initialize optionCodes based on the options of the product.
+    const initialOptionCodes = new Map();
+    for (const option of product.configurable_options) {
```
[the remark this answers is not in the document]
supernova-at: Resolved by https://github.com/magento-research/pwa-studio/pull/1240/commits/9502049fdd41c9b70653d21f5c91218a82a638ae.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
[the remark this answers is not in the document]
sirugh: There's never a time in which the PDP is re-used? Not even when something changes or a selection is chosen?

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
sirugh: I get it. From the docs:

> this is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders

packages/venia-concept/src/components/MiniCart/__tests__/cartOptions.spec.js
```diff
@@ -50,40 +44,6 @@ const configItem = {
...
-            configItem={{}}
-            updateCart={updateCart}
-            closeOptionsDrawer={closeOptionsDrawer}
-            classes={classes}
-        />
-    ).dive();
-    const options = wrapper.find('section.options');
-    expect(options.length).toBe(0);
-    expect(wrapper.hasClass(classes.root)).toBe(true);
-});
```
supernova-at: These tests are no longer needed because they test the internals of a different component (`options.js`), not anything in `cartOptions.js`.

packages/venia-concept/src/components/ProductOptions/options.js
```diff
@@ -1,16 +1,14 @@
...
 
 class Options extends Component {
     static propTypes = {
         onSelectionChange: func,
-        options: arrayOf(
-            shape({
-                attribute_id: string.isRequired
-            })
-        ).isRequired
+        product: object
```
supernova-at: Instead of taking `options`, this `Options` component now takes a `product` prop. It internally handles whether to show options based on whether the product is configurable or not.

This allowed us to remove a lot of repeated conditional logic from other components.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
supernova-at: Resolved by https://github.com/magento-research/pwa-studio/pull/1240/commits/130afe85aa585973b1670b6f26468b2f25402c86.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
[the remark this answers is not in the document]
supernova-at: Those are all state changes, not new props being passed in.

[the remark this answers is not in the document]
supernova-at: I am incorrect, `getDerivedStateFromProps` does get called every time a selection is chosen.

But the props never change and therefore the return value never changes. Check out these console logs:

Screen Shot 2019-05-22 at 12 09 50 PM

When I first load the page `optionCodes` is an empty `Map`. Every time I make a selection `getDerivedStateFromProps` runs but nothing is updated.

When I load up the PDP with a different product the `optionCodes` starts again as an empty `Map`.

Therefore I'm confident we aren't losing any functionality here.

Thanks for your patience on this one!

packages/venia-concept/src/components/MiniCart/cartOptions.js
```diff
@@ -136,7 +125,14 @@ class CartOptions extends Component {
                     </span>
                 </div>
                 <div className={classes.form}>
-                    {options}
+                    <section className={classes.options}>
```
sirugh: Now that `isProductConfigurable` is used within `Options` to return `null` this looks cleaner. The only difference would be that we now always have a `<section>` element with `class="options"`. That doesn't matter because none of the css does any sizing with `section` or the class but it could so just a warning. Not a big deal :)

sirugh: Man after digging into how these are styled its not very clear. Looks like there is a `section` class that is used to compose the sections in this form _but_ the options class does not use it for some reason.

You don't have to do it here but I just wanted to note a better way forward would be to add `composes: section` to the options class, remove the margins, and add a similar section for composing within the miniCart.css class :D

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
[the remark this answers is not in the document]
sirugh: > But the props never change and therefore the return value never changes. 

Sounds perfect for a memoized callback XD

packages/venia-concept/src/components/MiniCart/__tests__/cartOptions.spec.js
zetlen: I love it! Thanks for maintaining and streamlining the test suite, and for keeping it min mind not to test implementation details.

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
```diff
@@ -95,11 +101,45 @@ test('Configurable Product has correct media gallery image count', async () => {
...
+                outerWidth: 99
+            },
+            jest.fn()
+        ])
+        // quantity
+        .mockReturnValueOnce([1, jest.fn()])
+        // optionSelections
+        .mockReturnValueOnce([new Map([['1', 1]]), jest.fn()]);
+
+    debugger;
```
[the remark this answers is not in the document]
sirugh: OOPS

packages/venia-concept/src/components/ProductOptions/options.js
supernova-at: I shall double check this, good question 👍

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
```diff
@@ -1,10 +1,16 @@
...
     WindowSizeContextProvider,
     createTestInstance
 } from '@magento/peregrine';
 
+import ProductFullDetail from '../ProductFullDetail';
+
+jest.mock('react', () => {
+    const React = jest.requireActual('react');
+
+    return Object.assign(React, { useState: jest.fn(React.useState) });
```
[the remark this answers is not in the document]
supernova-at: Good catch, this is me being used to `sinon`'s spies and not remembering that Jest spies [return mock functions](https://jestjs.io/docs/en/jest-object.html#jestspyonobject-methodname).

packages/venia-concept/src/components/ProductFullDetail/__tests__/ProductFullDetail.spec.js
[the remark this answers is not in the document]
supernova-at: image

tjwiebell:
Ran through manual checkout flow and ran full automated test suite with no failures, QA passed 🌮 

Screen Shot 2019-05-24 at 3 59 44 PM

Coveralls says test coverage dipped a trivial amount, but not a blocker.

PR #1283, opened 2019-05-30
MiniCart Checkout Button

### LINKED ISSUE #1071: [BUG] Cropped shopping cart in Venia for larger phones

|Property | Value|
| Device | iPhone XR |
| Browser | Mobile_safari |
| Viewport | 414 x 896 dp |
| Aspect Ratio | 19.5 : 9 |

**Screenshot Attached** 
 Screenshot URL

supernova-at:
## Description

This PR updates the measurement the MiniCart uses to compute its `height` to use `window.innerHeight` instead of `100vh` as described by https://css-tricks.com/the-trick-to-viewport-units-on-mobile/. 

As described in the attached issue #1071, `100vh` often doesn't account for browser address bars or other frames.

This PR takes advantage of the new [useWindowSize](https://github.com/magento-research/pwa-studio/blob/develop/packages/peregrine/src/hooks/useWindowSize.js) hook to update this value even after resize events occur.

## Related Issue

Closes #1071 .

## Verification Steps

1. Add an item to your cart
1. Open the MiniCart (right drawer)
1. See that the Checkout button is fully visible
1. Resize the window (or change orientation of device) and ensure Checkout button is fully visible
1. The payments form depended a bit on these CSS variables, so ensure that looks good too

Run the verification steps above on a variety of phones.

## How Have YOU Tested this?

1. Tested using iPhone XR on iOS simulator
1. `yarn test`.

### Before Fix
image

### After Fix
Screen Shot 2019-05-30 at 3 35 28 PM

## Checklist:

## Note

This PR is based on #1266 so that it can take advantage of the `useWindowSize` hook.

As such, it contains commits from #1266. The only code change introduced by this PR is https://github.com/magento-research/pwa-studio/pull/1283/commits/b7cabd5ff27dcd74dae7103601a9a3a0da27c02c.

I have added the `hold` label because this PR should not be merged before #1266 .


packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -22,12 +24,18 @@ const MiniCart = props => {
...
     } = props;
 
+    // Hooks.
+    const { innerHeight: viewportHeight } = useWindowSize();
+
     // Members.
     const classes = mergeClasses(defaultClasses, props.classes);
     const rootClass = isOpen ? classes.root_open : classes.root;
+    const rootStyle = {
+        '--minicart-height-unit': `${viewportHeight * 0.01}px`
```
sirugh: Should we debounce this?

>⚠️ Updating the value of --vh will trigger a repaint of the page and the user may experience a jump as a result. Because of this, I’m not advising that this trick should be used for every project or to replace all usage of the vh unit but only when you may need your users to have an exact viewport unit value.

sirugh: Actually we can't -- @jimbo informed me that you must have the same number of hook calls per render. Looks like this is off your plate :D

sirugh:
Removing hold since miniCart refactor is now merged.

supernova-at:
Closing in favor of #1387 .

PR #1309, opened 2019-06-06
[dev]: Refactor Checkout

### LINKED ISSUE #1294: Peregrine: Checkout
Refactor the checkout components within the right drawer, to prepare for migration to Peregrine.

- [ ] Refactor class components to functional ones
- [ ] Add unit tests for new components

sirugh:
## Description
Refactoring of checkout components, including updating unit tests. Notably: 

* Removed `SubmitButton` in favor of just using the `Button` component directly.
* Removed the unused `ResetButton`.
* Address Form Refactor
  * fixed bug in checkout reducer where street address was not saved
* BraintreeDropin Refactor
  * The conversion to the mount/unmount logic required some...interesting code. Check out the `useEffect` in the component and see how I had to use a local variable to ensure we teardown/error properly.
* Payment Form Refactor
  * identified a "bug" where `informed` updates state if `initialValues` are used causing multiple renders.
* Form Refactor
  * The parent form is now split into multiple function components. Should we move these into a `Form/` directory and split them out into individual files?

## Related Issue

Closes #1294.

## Verification Steps

1. Go to the checkout form and step through it. Everything should work including all validation, etc.
2. Try doing the checkout form out of order.
3. Try quickly opening and closing sub-forms such as the payments form.
4. Make sure form state is used appropriately. Does it remember address? Does it remember last choices such as "use same as shipping address", etc?

## How have YOU tested this?

Lots of manual testing through different scenarios. See above.

## Checklist:


packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -22,228 +21,232 @@ const DEFAULT_FORM_VALUES = {
...
+        cancel,
+        classes,
+        countries,
+        formApi,
+        isRequestingPaymentNonce,
+        setIsRequestingPaymentNonce,
+        submit,
+        submitting
+    } = props;
+    const formState = useFormState();
```
sirugh: A wild hook appears!

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,329 +1,354 @@
...
+    } = props;
+
+    const handleCancel = useCallback(() => {
+        editOrder(null);
+    }, [editOrder]);
+
+    const handleSubmitAddressForm = useCallback(
+        formValues => {
+            submitShippingAddress({
+                type: 'shippingAddress',
```
sirugh: I don't think we need to forward these `type` props since we have specific functions for submission of each type. Maybe a remnant from some earlier code? I see two ways to handle this, and I'd like opinions:

1) Remove `type`. Easiest "fix".
2) Refactor into a single `submitForm` action that accepts a `type` and handles logic internally instead of having multiple different `submit` functions. Benefit of only having to thread one function prop down instead of three.

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,329 +1,354 @@
...
-                        valid={ready}
-                        submitOrder={submitOrder}
-                    />
-                </div>
-            </Fragment>
-        );
-    }
+const isSubmitDisabled = (busy, valid) => busy || !valid;
+
+const EditableForm = props => {
```
sirugh: The parent form is now split into multiple function components. Should we move these into a Form/ directory and split them out into individual files?

packages/venia-concept/src/components/Checkout/container.js
```diff
@@ -0,0 +1,169 @@
...
+    } = checkout;
+
+    return [billingAddress, paymentData, shippingAddress, shippingMethod].every(
+        hasData
+    );
+};
+
+// TODO: This is exported for testing, but I think we should split this into a
+// separate file entirely and just do the prop/action shaping here in this file.
+export const CheckoutContainer = props => {
```
sirugh: Yea, I'm not even sure we need this component. We do the prop shaping, split it into multiple groups, and then [pull them apart](https://github.com/magento-research/pwa-studio/pull/1309/files#diff-c54588d5f498544c2028e38a1865dc0aR48) inside flow anyways.

packages/venia-concept/src/components/Checkout/form.js
sirugh: @jimbo says he prefers flat structure. I like nested, for things that are specific to a parent and won't be shared. 🤷‍♂

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,329 +1,366 @@
...
+    const handleCancel = useCallback(() => {
+        setEditing(null);
+    }, [setEditing]);
+
+    const handleSubmitAddressForm = useCallback(
+        async formValues => {
+            await submitShippingAddress({
+                formValues
+            });
+            setEditing(null);
```
sirugh: If we move state out of redux we must at some point `await` some async action completion where the logic would have [otherwise been handled internally in the reducer](https://github.com/magento-research/pwa-studio/blob/develop/packages/venia-concept/src/reducers/checkout.js#L116). For example, we would have `setEditing(null)` in the `paymentMethod.accept` function instead. We will see the same sort of logic when/if we convert the checkout `step` to a local state.

Is there some best practice when it comes to doing this sort of thing? Personally I think this reads better but it has a consequence of "hiding" these state changes within the components.

sirugh: After discussion with @jimbo we decided that while this may be the right direction to take when it comes to refactoring out redux, we don't want to do it _here_. So I will be reverting this commit into a separate branch/PR so we have it around (since I spent time on updating tests). And we can deal with it later.

packages/venia-concept/src/components/Checkout/form.js
sirugh: I opted for option 1 right now since nothing is using `type`. And we can easily refactor the actions later.

packages/venia-concept/src/components/Checkout/form.js
sirugh: Moved to #1338

packages/venia-concept/src/components/Checkout/cart.js
```diff
@@ -1,35 +1,34 @@
...
-        return (
-            <div className={classes.root}>
-                <div className={classes.footer}>
-                    <CheckoutButton
-                        ready={ready}
-                        submitting={submitting}
-                        submit={beginCheckout}
-                    />
-                </div>
+    const disabled = isDisabled(submitting, ready);
```
supernova-at: Do we use `isDisabled` anywhere else? Want to just inline its logic here?

packages/venia-concept/src/components/Checkout/form.js
supernova-at: > The parent form is now split into multiple function components. Should we move these into a Form/ directory and split them out into individual files?

We should definitely split them out into individual files, as per existing convention. I do like the idea of a `Form/` directory personally - especially since we have a `Receipt` directory within `Checkout` already.

supernova-at: > @jimbo says he prefers flat structure. I like nested, for things that are specific to a parent and won't be shared. 🤷‍♂

fight

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,329 +1,364 @@
...
+                    submit={handleSubmitShippingForm}
+                    submitting={submitting}
+                />
             );
         }
+        default: {
+            return null;
+        }
+    }
+};
```
supernova-at: Would be nice to have `propTypes` defined for each of these subcomponents.

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,329 +1,364 @@
...
-                </span>
+                <AddressForm
+                    initialValues={shippingAddress}
+                    submitting={submitting}
+                    countries={countries}
+                    cancel={handleCancel}
+                    submit={handleSubmitAddressForm}
+                    isAddressIncorrect={isAddressIncorrect}
+                    incorrectAddressMessage={incorrectAddressMessage}
+                />
```
supernova-at: This wouldn't block a 👍 but just pointing out that alphabetizing really helps readability in cases where there are a lot of props like this.

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,329 +1,364 @@
...
+                    shippingMethod={shippingMethod}
+                    submit={handleSubmitShippingForm}
+                    submitting={submitting}
+                />
             );
         }
+        default: {
+            return null;
+        }
+    }
```
supernova-at: This looks great!

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,329 +1,364 @@
...
+    shippingMethod: string,
+    shippingTitle: string,
+    submitShippingAddress: func.isRequired,
+    submitOrder: func.isRequired,
+    submitPaymentMethodAndBillingAddress: func.isRequired,
+    submitShippingMethod: func.isRequired,
+    submitting: bool.isRequired
+};
+
+export default Form;
```
supernova-at: Man, all this is so much easier to reason about now. Great work!

packages/venia-concept/src/components/Checkout/flow.js
```diff
@@ -1,192 +1,157 @@
...
+    const {
+        billingAddress,
+        paymentData,
+        shippingAddress,
+        shippingMethod
+    } = checkout;
+
+    return [billingAddress, paymentData, shippingAddress, shippingMethod].every(
+        hasData
+    );
```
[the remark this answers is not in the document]
sirugh: Good point. I didn't update these util functions, just move them, but I will update the check.

packages/venia-concept/src/components/Checkout/flow.js
```diff
@@ -1,192 +1,157 @@
...
+        shippingAddress,
+        shippingMethod,
+        shippingTitle,
+        step,
+        submitting
+    } = checkout;
+
+    // ensure state slices are present
+    if (!(cart && checkout)) {
+        return null;
```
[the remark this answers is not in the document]
sirugh: TBH I don't know why this is necessary. We shouldn't ever be setting these to null as they are state objects that come from redux.

packages/venia-concept/src/components/Checkout/flow.js
```diff
@@ -1,192 +1,157 @@
...
-export default classify(defaultClasses)(Flow);
+    return <div className={classes.root}>{child}</div>;
+};
+
+Flow.propTypes = {
+    beginCheckout: func,
+    cancelCheckout: func,
+    cart: shape({
+        details: object.isRequired
+    }),
```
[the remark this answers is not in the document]
sirugh: Yea, I was moving props around when I removed the container. `Flow` didn't previously have the helper function as it was in `wrapper`, and that had proptypes which required `cart` and `details`. I think I ran into a problem where `cart` was an empty object and was failing the prop checking. I'll investigate.

packages/venia-concept/src/components/Checkout/cart.js
sirugh: Sure. This used to live in `CheckoutButton` but moving it here allowed me to simplify the button API, which IMO isn't super necessary anyways. I may end up replacing it entirely with one of the shared button/trigger components.

packages/venia-concept/src/components/Checkout/form.js
sirugh: WOULDN'T IT!?



Fine :D

packages/venia-concept/src/components/Checkout/form.js
sirugh: Definitely. I did it in some places but not others. I shall look for and alphabetize the props on component implementations :)

packages/venia-concept/src/components/Checkout/flow.js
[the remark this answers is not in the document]
jimbo: This is legacy code from when these slices were dynamically imported and added to the root reducer. We don't do that anymore, so the safeguard should no longer be necessary

packages/venia-concept/src/components/Checkout/form.js
jimbo: We should never have had a `Receipt` folder either—last year we had a few contributors who didn't like to follow conventions. 😉 

Honestly though, if you guys prefer to nest folders in places with a lot of files like this, I'm fine with it. Just don't go overboard.

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -22,228 +21,229 @@ const DEFAULT_FORM_VALUES = {
...
-    render() {
-        const { classes, initialValues } = this.props;
-        const { formChildren } = this;
+const BillingAddressFields = ({ classes, countries }) => {
+    return (
+        <Fragment>
+            <div className={classes.street0}>
+                <Field label="Street">
+                    <TextInput
+                        id={classes.street0}
```
jimbo: Not that this is your problem (in fact, this might be my code), but the `id` attribute may not contain spaces. The classnames output by CSS Modules may contain spaces, though, so I don't think we can use them like this.

packages/venia-concept/src/components/Checkout/paymentsForm.js
jimbo: Good comment. I'm not surprised, though; Informed implements change publishing manually, so they're not doing hooks + context in any sort of optimal way.

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -22,228 +21,229 @@ const DEFAULT_FORM_VALUES = {
...
-    };
-
-    state = {
-        isRequestingPaymentNonce: false
-    };
-
-    render() {
-        const { classes, initialValues } = this.props;
-        const { formChildren } = this;
+const BillingAddressFields = ({ classes, countries }) => {
```
jimbo: We definitely want to stick to one component definition per file. There's nothing wrong with having more files.

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -22,228 +21,229 @@ const DEFAULT_FORM_VALUES = {
...
+    }, [cancel]);
+
+    const handleError = useCallback(() => {
+        setIsSubmitting(false);
+    });
+
+    // The success callback. Unfortunately since form state is created first and
+    // then modified when using initialValues any component who uses this
+    // callback will be rendered multiple times on first render. See above
+    // comments for more info.
```
jimbo: Unfortunately, that's the reality of almost any convenient API where we set initial values, validators, etc., on individual components. Those things have to be sent up to the form's state object for actual initialization, meaning it takes a re-render before everything is fully set up.

I think that re-render is acceptable, and it'll be even better if we ever move away from Informed.

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -22,228 +21,229 @@ const DEFAULT_FORM_VALUES = {
...
+            submit({
+                billingAddress,
+                paymentMethod: {
+                    code: 'braintree',
+                    data: value
+                }
+            });
+        },
+        [formState.values]
+    );
```
jimbo: Good work on these callbacks. 👍

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -22,228 +21,229 @@ const DEFAULT_FORM_VALUES = {
...
+    return (
+        <Fragment>
+            <div className={classes.body}>
+                <h2 className={classes.heading}>Billing Information</h2>
+                <div className={classes.braintree}>
+                    <BraintreeDropin
+                        shouldRequestPaymentNonce={isSubmitting}
+                        onError={handleError}
+                        onSuccess={handleSuccess}
+                    />
```
jimbo: Is there any way for `BraintreeDropin` to be loaded with `React.lazy()` and `Suspense`?

packages/venia-concept/src/components/Checkout/label.js
```diff
@@ -1,25 +1,27 @@
...
-    static propTypes = {
-        children: node,
-        classes: shape({
-            root: string
-        }),
-        plain: bool
+const Label = props => {
+    const { children, plain, ...restProps } = props;
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const elementType = plain ? 'span' : 'label';
```
jimbo: Thanks for keeping this. There are no comments in the file, but it's important not to have nested `<label>` elements for accessibility.

packages/venia-concept/src/components/Checkout/form.js
jimbo: Agreed. Much cleaner. 👍

packages/venia-concept/src/components/Checkout/form.js
jimbo: And put them in new files! 💯

packages/venia-concept/src/components/Checkout/form.js
```diff
@@ -1,329 +1,364 @@
...
+    );
+};
+
+/**
+ * The Form component is similar to Flow in that it renders either the overview
+ * or the editable form based on the 'editing' state value.
+ */
+const Form = props => {
+    const classes = mergeClasses(defaultClasses, props.classes);
+    let child;
```
jimbo: Not that it's *super* important, but you should be able to use `const` even in places like this. We could tighten up the `prefer-const` rule if warnings would help.

```es6
const child = editing ? (
  <Foo />
) : (
  <Bar />
)

```

packages/venia-concept/src/components/Checkout/flow.js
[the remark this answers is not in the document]
jimbo: @supernova-at Good analysis.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -1,139 +1,147 @@
...
-        this.setState({ isError });
-    }
+        // Initialize the dropin with a reference to the container mounted in
+        // this component. We do this via onmount because we have to be sure
+        // the container id has mounted per braintree's API.
+        createDropinInstance();
+
+        // If we close the form quickly after opening we may end up in a
+        // semi-mounted state so we need a local variable to make sure we
+        // behave correctly.
```
jimbo: Great work. This is obviously important to handle but easy to overlook.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -1,139 +1,147 @@
...
-                `Unable to initialize Credit Card form (Braintree). \n${err}`
-            );
-            isError = true;
+/**
+ * This BraintreeDropin component has two purposes which lend to its
+ * implementation:
+ *
+ * 1) Mount and asynchronously create the dropin via the braintree api.
+ * 2) On submission (triggered by a parent), request the payment nonce.
+ */
```
jimbo: This component is a great example of what we're aiming for. 👍

packages/venia-concept/src/components/Checkout/addressForm.js
```diff
@@ -27,180 +26,175 @@ const fields = [
...
-    };
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const validationMessage = isAddressIncorrect
+        ? incorrectAddressMessage
+        : null;
 
-    static defaultProps = {
-        initialValues: {}
-    };
+    const values = useMemo(
```
jimbo: Good use of `useMemo()` here.

packages/venia-concept/src/components/Checkout/paymentsForm.js
sirugh: I can investigate. TBH I haven't used either yet, so I'm eager to try.

packages/venia-concept/src/components/Checkout/section.js
```diff
@@ -1,54 +1,48 @@
...
+Section.propTypes = {
+    classes: shape({
+        content: string,
+        icon: string,
+        label: string,
+        root: string,
+        summary: string
+    }),
+    label: node,
+    onClick: func,
```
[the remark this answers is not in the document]
sirugh: I don't remember 😢

packages/venia-concept/src/components/Checkout/flow.js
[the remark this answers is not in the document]
sirugh: So all of these props start as null. As they are filled they are added to the store. All are added as objects except for `shippingMethod` which is a string. I can update this check to be more explicit.

packages/venia-concept/src/middleware/log.js
```diff
@@ -10,13 +10,17 @@ const log = store => next => action => {
 
     console.groupCollapsed(action.type);
     console.group('payload');
+    // TODO: There is a bug where array properties of `action.payload` are
+    // logged as empty even though they aren't and the state is updated with
+    // incoming values (although that is dependent on the action). Probably
+    // related to Chrome logging a reference to the array and some code clearing
+    // the array at some point.
```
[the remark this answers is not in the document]
sirugh: Ok.

packages/venia-concept/src/components/Checkout/paymentsForm.js
sirugh: I don't think I follow. Can you clarify @jimbo?

packages/venia-concept/src/components/Checkout/form.js
sirugh: DONE. 💀

packages/venia-concept/src/components/Checkout/form.js
sirugh: I think I'm tired of staring at this stuff. I did it where I could. If we want to enforce this we should add some linter/plugin that alphabetizes this for us automatically. Ugh. lol

sirugh:
I addressed most of the feedback. There's a few things I need to look into such as lazy loading braintree and some leftover TODOs. I also split out `Form` into multiple component files. I chose not to create a `Form/` folder right now because I didn't want to lose some of the requested changes. I can do that later if we care, otherwise I'm going to leave it as is.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -1,139 +1,147 @@
...
+    classes: shape({
+        root: string,
+        error: string
+    }),
+    onError: func.isRequired,
+    onSuccess: func.isRequired,
+    shouldRequestPaymentNonce: bool
+};
+
+export default BraintreeDropin;
```
supernova-at: Looks great!

packages/venia-concept/src/components/Checkout/addressForm.js
```diff
@@ -27,180 +26,176 @@ const fields = [
...
-        const { isAddressIncorrect, incorrectAddressMessage } = this.props;
-        if (isAddressIncorrect) {
-            return incorrectAddressMessage;
-        } else {
-            return null;
-        }
-    };
+    const handleCancel = useCallback(() => {
+        cancel();
+    }, [cancel]);
```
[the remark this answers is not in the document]
sirugh: Yea I had this same feedback given on another PR. The only reason for renaming the callback would be so that it falls into some paradigm we use throughout the application. Of course, it's technically overkill and unnecessary. I am totally fine removing this sort of use of `useCallback` with a direct passing of the prop.

packages/venia-concept/src/components/Checkout/paymentsForm.js
sirugh: Do you still want me to look into this? Not sure how much work it'll be do we want to add to this PR?

packages/venia-concept/src/components/Checkout/paymentsForm.js
```diff
@@ -21,8 +21,26 @@ const DEFAULT_FORM_VALUES = {
     addresses_same: true
 };
 
-const BillingAddressFields = ({ classes, countries }) => {
-    return (
+const PaymentsFormChildren = props => {
```
jimbo: `PaymentsFormChildren` should also be extracted into its own file, no?

Also, consider `PaymentsFormItems`, if you'd like to follow the pattern we have in a couple other places.

sirugh: I wasn't sure about this -- the component wrapping it is literally just one layer:

```
return (
  <div>
    <PaymentsFormChildren {...props} />
   </div>
)
```
We can probably just get rid of this separated component.

sirugh: Oh I see -- it's pretty involved. Maybe I need to break out the `BillingAddressFields` too...

sirugh:
@jimbo feedback addressed! If this is good to go please put it in Ready for QA :)

PR #1316, opened 2019-06-07
Wrapped Peregrine Context Provider Component

### LINKED ISSUE #1312: [feature]: Unified Peregrine Provider

As we move state out of Redux and increase our usage of React Context, asking consumers to wrap `App` with a thousand context providers is going to become a problem.

In addition to all the individual providers it offers, Peregrine should export a single provider with which we can wrap `App`. It would be nice if it were configurable, but only if that doesn't introduce new problems.

```jsx
// peregrine/src/Peregrine

import React from "react"
import { ToastContextProvider } from "./Toasts"
import { WindowSizeContextProvider } from "./WindowSize"

const Peregrine = props => (
  <WindowSizeContextProvider>
    <ToastContextProvider>
      {props.children}
    </ToastContextProvider>
  </WindowSizeContextProvider>
)

export default Peregrine

// venia-concept/src/index

import React from "react"
import { render } from "react-dom"
import { Peregrine } from "@magento/peregrine"
import App from "./components/App"

render(
  <Peregrine>
    <App />
  </Peregrine>
)
```

While verbose, wrapping `App` with many individual providers would be explicit, so conflicts would be easy to resolve and unneeded code wouldn't be included in the build. Rolling up providers could also be considered the domain of the consumer, rather than Peregrine.

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

tjwiebell:
## Description
Add a Peregrine component that wraps the composition of context providers needed to use Peregrine hooks (currently just WindowSize and Toasts).

## Related Issue

Closes #1312

## Verification Steps

1. Go to Home page and use React DevTools to inspect source
2. Verify that `<App />` is wrapped by `WindowSizeContextProvider` and `ToastContextProvider`

## How have YOU tested this?

Ran verification steps and ran jest

## Checklist:


packages/venia-concept/src/index.js
```diff
@@ -37,11 +36,9 @@ ReactDOM.render(
         apollo={{ link: authLink.concat(Adapter.apolloLink(apiBase)) }}
         store={store}
     >
-        <WindowSizeContextProvider>
-            <ToastContextProvider>
-                <App />
-            </ToastContextProvider>
-        </WindowSizeContextProvider>
+        <Peregrine>
```
sirugh: How set are we on this name? I feel like this should be `PeregrineContextProvider`. Also, I know that you added `defaultProps` for toasts and window_size but I feel like we should just be explicit here. The less magic the better since people will be referencing this in the future and should be able to, at a glance, understand what exactly the top-level is doing.

tjwiebell: Naming is flexible. I'm not sure if there was an intention here that Peregrine might wrap more than just ContentProviders at some point, but if Providers is its only responsibility, this name change makes sense to me.

My concern with an explicit config prop is the reason we're creating this component now; I anticipate if we proceed with replacement of Redux with Peregrine hooks, our list of providers is going to be very large. I would be fine removing the defaultProp if we had a flag to use all (which is now just equivalent to not passing the prop).

jimbo: The only thing a root-level `Peregrine` component would ever do is provide context, and it enables all the features of the package, so I figure `Peregrine` is adequate here. But I don't mind renaming it; exports can always be renamed by the importer.

packages/peregrine/src/Peregrine/peregrine.js
```diff
@@ -0,0 +1,26 @@
...
+
+const Peregrine = ({ children, providers }) => {
+    return providers.reduceRight((memo, providerKey) => {
+        const ContextProvider = contextProviders[providerKey];
+        return <ContextProvider>{memo}</ContextProvider>;
+    }, children);
+};
+
+Peregrine.propTypes = {
+    providers: arrayOf(oneOf(Object.keys(contextProviders)))
```
jimbo: There a couple of problems with this approach.

1. Opting out of a provider is difficult, because you have to pass the rest of the entire array as a prop. Once we have a dozen providers, that won't really work.
1. Since all of the providers are statically imported, you're not saving anything by opting out of a provider. The same amount of code is bundled and served, even if you pass an empty array. If there's no bundle savings by opting out, no one should ever opt out of a provider.

packages/venia-concept/src/index.js
sirugh: I'd like to change the export/component at least. We can import as `Peregrine` here.

packages/peregrine/src/Peregrine/peregrine.js
sirugh: @jimbo are you suggesting that people pass in a list of the Providers they wish to use? So they'd import in `App` for example and pass the list?

Also can we use dynamic import/suspense here?

packages/peregrine/src/index.js
```diff
@@ -21,6 +21,7 @@ export { default as List, Items, Item } from './List';
 export { default as Page } from './Page';
 export { default as Price } from './Price';
 export { default as Router } from './Router';
+export { default as Peregrine } from './Peregrine';
```
sirugh: Yea so I'd want to alias this as `PeregrineContextProvider` and in venia `import { PeregrineContextProvider as Peregrine }`.

packages/peregrine/src/Peregrine/peregrine.js
tjwiebell: @jimbo - we can switch to using a `disabledProviders` array prop, but in the current state wouldn't solve the bundle size concern. Are you suggesting we explore lazy loading of these components, or just stripping config out entirely and not giving users the ability to configure providers?

sirugh: We definitely want users to be able to configure which providers are imported/used. Can you explore lazy loading the providers based on a flag?

jimbo: @sirugh I don't think we can state that quite so unequivocally.

### Case 0: no unified provider

The default use case here is that we don't have a unified context provider. In this case, an app imports all of the context providers it needs individually.

- 👍 Only the providers needed are imported
- 👍 Only the providers needed are rendered
- 👍 Providers are imported & rendered synchronously
- 👎 It's verbose

### Case 1: unified provider, no config

The first proposal, which @tjwiebell originally created in this PR, is a simple unified provider that imports all of the common providers and renders them.

- 👎 All providers are imported, including unneeded ones
- 👎 All providers are rendered, including unneeded ones
- 👍 Providers are imported & rendered synchronously
- 👍 It's concise

### Case 2: unified provider, granular rendering

The second proposal, which @tjwiebell has added to this PR, is a configurable unified provider that imports all of the common providers and *conditionally renders* them.

- 👎 All providers are imported, including unneeded ones
- 👍 Only the providers needed are rendered
- 👍 Providers are imported & rendered synchronously
- 👍 It's concise

### Case 3: unified provider, granular importing

A third proposal, which we're discussing, is a configurable unified provider that *conditionally imports* providers and renders them.

- 👍 Only the providers needed are imported
- 👍 Only the providers needed are rendered
- 👎 Providers are imported & rendered asynchronously
- 👍 It's concise

### Non-proposals

To be clear, there is no proposal in which an app would import providers and pass them to the unified provider; that workflow makes zero sense. In such a scenario, the app itself should just render its own providers, since rendering is all the unified provider does.

### Async importing

The point of conflict here is that Case 2 doesn't offer much value over Case 1 because the only cost savings that matter are import cost savings, and Case 2 doesn't offer those savings. Case 3 does, but **async importing providers is not necessarily something we want to do**.

We're envisioning a scenario in which Peregrine contains dozens of providers. If those are dynamically imported, every single consumer of those providers in the app will have to account for a case where the provider isn't present yet, and in some cases will have to duplicate the provider's initial state in order to render meaningfully.

We tried this early on with Redux slices. We figured out how to dynamically import individual reducers and add them to the store on demand, and we were doing it for cart and checkout. Unfortunately, it makes such a logical mess of the code that we stopped doing it, which is why today all reducers are present at first render.

### Other options

We could try categorizing providers as essential and non-essential. In this scenario, the unified provider would import and render the essential providers synchronously, so either Case 1 or Case 2. For non-essential providers, it would be left to the app to import and render them individually, as in Case 0—and it could probably do so asynchronously for certain providers. Thus:

### Case 4: unified provider, for essentials only

- 👍 Only the providers needed are imported, mostly
- 👍 Only the providers needed are rendered
- 👍 Providers are imported & rendered synchronously
- 👍 It's concise

tjwiebell:
Reviewed with the team and we're going to proceed with Option 4, and establish patterns for creating `ContextProvider` containers; an essential Peregrine container, and non-essential container in Venia. The Peregrine container will initially be empty, but is expected to contain Providers in the future.

Until we encounter performance bottlenecks related to container initialization, we will not be attempting asynchronous imports.

packages/peregrine/src/index.js
tjwiebell: Made these name and alias changes in bf94de4

PR #1321, opened 2019-06-10
Peregrine REST API Hook

### LINKED ISSUE #1295: Peregrine: Add a hook for making REST calls
## Background

When creating PWA Studio, we decided not to create an abstraction for REST calls, in order to favor and encourage GraphQL usage. This decision was made possible by relegating REST calls to Redux actions, which are already full of boilerplate and generally less friendly than components. As we move more of Redux state into local, hook-oriented state, we will inevitably need a new REST abstraction to avoid putting massive amounts of request plumbing in many components.

Specifically, we need a hook that does for REST calls what [`useQuery`](https://github.com/magento-research/pwa-studio/blob/develop/packages/peregrine/src/hooks/useQuery.js) does for GraphQL calls.

## Requirements

In `peregrine`, create React hooks for issuing REST requests and storing their results. Use `useQuery` as a guide for how these hooks should be constructed.

- [ ] Determine what to name these hooks (names below are pending)
- [ ] `useRestRequest` should accept a single argument, the string `endpoint`
- [ ] `useRestRequest` should return `[state, api]`, as `useQuery` does
- [ ] `useRestRequest` should call `useRestResponse` to get and set state
- [ ] State should consist of `data`, `error`, and `loading`

supernova-at:
## Description
This PR introduces a hook into Peregrine for using Magento's REST API.

## Related Issue

Closes #1295 .

## Verification Steps

There aren't any components that use this REST API hook yet.
Verify that the unit tests are written correctly.

## How have YOU tested this?

`yarn test`

## Checklist:


packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
```diff
@@ -191,15 +191,25 @@ class M2ApiRequest {
                             // includes the original context of the request,
                             // and formats the server response.
                             .then(bodyText => {
-                                throw new M2ApiResponseError({
```
supernova-at: @zetlen `M2ApiResponseError` builds up a nice pretty message out but it looks more formatted for the terminal than for the front end. The front end really just needs the error `message`.

Can we work together to find a solution here that works for both terminal and UI?

packages/peregrine/src/hooks/useRestResponse.js
```diff
@@ -0,0 +1,121 @@
...
+    error: null,
+    loading: false
+};
+
+const reducer = (state, { payload, type }) => {
+    switch (type) {
+        case 'set data': {
+            return { ...state, data: payload };
+        }
+        case 'set error': {
```
supernova-at: @jimbo should I just drop `set data` and `set error` from the API?  `useRestApi` uses `receive response` and `receive error`.

packages/venia-concept/src/RootComponents/CMS/CMS.js
```diff
@@ -1,8 +1,41 @@
...
+            setLoading(false);
+        };
+
+        signIn();
+    }, []);
+
+    if (error) return <h1>Error!</h1>;
+    if (loading || !data) return loadingIndicator;
+
+    return <span>data: {data}</span>;
```
supernova-at: @jimbo @sirugh Here's what using the REST API would look like from a component.

👍  / 👎  ?

packages/peregrine/src/hooks/useRestApi.js
```diff
@@ -0,0 +1,32 @@
...
+
+export const useRestApi = endpoint => {
+    const [restResponseState, restResponseApi] = useRestResponse();
+    const { receiveError, receiveResponse } = restResponseApi;
+
+    // Define a callback that performs a query
+    // either as an effect or in response to user interaction.
+    const sendRequest = useCallback(async ({ options }) => {
+        try {
+            const response = await request(endpoint, options);
```
supernova-at: This hook could call `setLoading` before and after the request so that calling components don't have to worry about it.

packages/peregrine/src/hooks/useRestResponse.js
```diff
@@ -0,0 +1,121 @@
...
+        () => ({
+            dispatch,
+            receiveError,
+            receiveResponse,
+            resetState,
+            setData,
+            setError,
+            setLoading
+        }),
+        [dispatch, receiveResponse, resetState, setData, setError, setLoading]
```
supernova-at: Add `receiveError` to this dependencies list

packages/venia-concept/src/RootComponents/CMS/CMS.js
```diff
@@ -1,8 +1,37 @@
...
+        const signIn = async () => {
+            await sendRequest({
+                options: {
+                    method: 'POST'
+                }
+            });
+        };
+
+        signIn();
+    }, []);
```
supernova-at: Ideally we'd just 

```
useEffect(async () => {
  await sendRequest(...)
});
```

but `async` results in the anonymous function passed to `useEffect` to return a `Promise` when the function passed to `useEffect` is supposed to return an "unsubscribe" callback.

The pattern above is recommended by [Dan Abramov](https://github.com/facebook/react/issues/14326#issuecomment-441680293) and is a temporary one until the React team releases the real solution (`Suspense`).

sirugh: I had to do this for the braintree stuff, and it gets even more complicated when you have potential side effects of your `await` call that require the component to be mounted :D

packages/peregrine/src/hooks/useRestResponse.js
supernova-at: `set data` and `set error` set those properties individually (ignoring `loading). Since we already had a `receive response` for successful responses (sets `data` to the payload, `error` to `null`, and `loading` to `false`, I wanted something similar for results that weren't `2xx`s.

So `receive error` sets `data` to null, `error` to the error message (payload), and `loading` to `false`.

Without this function we'd have to call each of `set data`, `set error`, and `set loading` individually.

I think having a `receive error` makes good sense, I'm just wondering if we need the individual setters mentioned above. I don't see them used anywhere, though that doesn't mean they aren't potentially useful. Ultimately I'm fine keeping them in.

supernova-at: > Infact these seem really similar to me, maybe we can just use the one and call it useApiResult?

They are totally similar right now but I don't know that we want to combine them. The `useQueryResult` hook is pretty dependent on how apollo returns data, which I wouldn't want the REST API hook to have to care about.

sirugh: > The useQueryResult hook is pretty dependent on how apollo returns data

Is it? They look super similar to me with minor differences in how we handle errors. Even `useRestApi` and `useQuery` are pretty similar. You could do the payload manipulation within the `runQuery`/`sendRequest` functions and combine into a single `useApiResponse` hook.

jcalcaben:
basic JSDocs would help. I can add more details if needed when I need to create the doc page for it

supernova-at:
Known Issues:
  * ESLint and tests are failing due to me changing `M2ApiRequest`. Still need some direction here and then these are easy fixes.

TODO: 
  - [ ] Add JSDoc comments to hooks

packages/peregrine/src/hooks/useRestResponse.js
jimbo: @sirugh We don't need to be overly aggressive about sharing abstractions. The priority here is ergonomics, so I'd like to leave some room for them to grow apart if necessary.

If they don't grow apart, and we find that they're doing the same thing, that will be a straightforward optimization in the future—especially since `useQueryResult` and `useRestResponse` are basically private.

jimbo: > should I just drop set data and set error from the API?
> [...] I don't see them used anywhere, though that doesn't mean they aren't potentially useful.

We could. I don't foresee using them, and it would save a few bytes. Perhaps we just don't define the methods, but we leave the reducer cases in?

packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
supernova-at: I ended up exposing the base error `message` as `baseMessage`. This keeps the current functionality and adds a simple property to access the original `message` easily if needed.

packages/peregrine/src/RestApi/Magento2/M2ApiRequest.js
```diff
@@ -191,15 +191,25 @@ class M2ApiRequest {
...
+                                //     method: this.opts.method,
+                                //     resourceUrl: this.resourceUrl,
+                                //     response,
+                                //     bodyText
+                                // });
+
+                                let message;
+                                try {
+                                    message = JSON.parse(bodyText).message;
+                                } catch {
```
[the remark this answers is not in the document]
supernova-at: Oh nice, thanks for the link. I ended up being able to do this much easier by modifying `M2ApiResponseError` instead.

packages/peregrine/src/hooks/useRestResponse.js
supernova-at: Could do that too, yeah. I'm fine with whatever honestly. Like you said, we can optimize later.

sirugh: @jimbo Yea, I talked to @supernova-at yesterday and we agreed that they may differ in the future so this is safer for the time being. If they merge later it won't affect the public API so it shouldn't be a big deal.

PR #1324, opened 2019-06-11
[feature] Added increment decrement stepper (+/-) instead of dropdown on PDP and Item Edit from Mini Cart.

### LINKED ISSUE #1281: [feature]: Implement quantity dropdown as increment decrements with text-box on PDP
**Description:**
As currently on PDP and on product-edit a dropdown is provided for "Quantity" in venia theme. That doesn't seems to be good from a user point of view. 

So instead of dropdown we may prefer going with an increment-decrement (+/-) with a text-box as well. So that user may simply increase/decrease value as per the need or may provide any quantity inside the text-box of quantity.

imageedit_2_6918117914

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## Tracked By

https://jira.corp.magento.com/browse/PWA-119

narendravyas24:
## Description

As quantity dropdown is not a perfect match from user point of view, implemented a Stepper with text-box.

Benefits of Stepper:
--------------------------
1. Dropdown blocks the input of any desired quantity by user that is not specified in the dropdown.

2. Number of clicks/touches are more than one in case of drop-down's than the stepper.

3. Only stepper can also be very annoying to user so I mentioned to use a text-box as well in addition to stepper. In which a user can provide any input as needed.

4. In the current implementation the options of drop-down are also mentioned in (src/components/ProductQuantity/mockData.js). So in case if a client needs to change the option that can done by changing the content of mockData.js, that is also not good.

5. Dropdown seems not feasible if quantities in decimal is allowed. A text-box can work in this case as well. 

Also applied various client side validations on the Stepper added, as:
1. Shown error if "0" or "negative" value is been input by user.
2. Shown error if no value is provided in quantity text-box

## Related Issue
Closes #1281.

## Verification Steps

1. Go to the Product details page.
2. Find increment-decrement stepper instead of quantity dropdown 

Screenshot from 2019-06-11 17-47-10
Screenshot from 2019-06-11 17-47-53
Screenshot from 2019-06-11 17-48-04
Screenshot from 2019-06-11 17-51-46

## Checklist:


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1324) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1324) before we can accept your contribution.<br/><hr/>**ronak@123** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

packages/venia-concept/src/components/MiniCart/cartOptions.js
```diff
@@ -116,48 +129,104 @@ class CartOptions extends Component {
...
+
+        const DecrementButtonFormApi = ({ formApi }) => {
+            return (
+                <button
+                    type="button"
+                    onClick={() => {
+                        this.decrementQty(formApi);
+                    }}
+                >
+                    -
```
belbiy: Please use the minus symbol "−" instead of hyphen "-".

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
```diff
@@ -131,54 +130,123 @@ const ProductFullDetail = props => {
...
+                        <Text
+                            type="number"
+                            field="quantity"
+                            initialValue={quantity}
+                            onValueChange={setQuantity}
+                            validateOnChange
+                            validate={validateQuantity}
+                            className={classes.qtyfield}
+                        />
+                        <IncrementButtonFormApi formApi={formApi} />
```
belbiy: Would it be better from the usability and accessibility standpoint to have `+` `−` buttons grouped together after the qty input field? So the first thing on the tab path is the input? 

stepper

packages/venia-concept/src/components/MiniCart/cartOptions.js
brendanfalkowski: The buttons should have readable content (increase quantity, decrease quantity) or `aria-label` attributes, and SVG icons so they're not affected by typeface variance.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
brendanfalkowski: I think it's clearest as (minus, input, plus). The source order can be controlled with CSS so the tab order is not a concern, but normally I don't do that because it's equally weird to tab in retrograde motion. If the initial button (minus) has a useful value it should make sense when tabbing in any order.

The arrangement like `type="number"` with two small buttons has too small targets. This isn't as usable across devices, and there isn't a need to constrain the space.

packages/venia-concept/src/components/MiniCart/cartOptions.css
```diff
@@ -61,6 +61,42 @@
...
+    font-size: 28px;
+    line-height: 30px;
+    padding: 0px 0 6px 0;
+    vertical-align: top;
+}
+.quantity button:hover {
+    background-color: #333;
+    color: #fff;
+}
+.qtyfield {
```
brendanfalkowski: Should not abbreviate here. Suggest `quantityInput` or simply `input`.

packages/venia-concept/src/components/MiniCart/cartOptions.js
```diff
@@ -116,48 +129,104 @@ class CartOptions extends Component {
...
+                                    <Options
+                                        onSelectionChange={
+                                            handleSelectionChange
+                                        }
+                                        product={configItem}
+                                    />
+                                </Suspense>
+                            </section>
+                            <section className={classes.quantity}>
+                                <h2 className={classes.quantityTitle}>
```
brendanfalkowski: Should be `<label htmlFor="">` for accessibility.

packages/venia-concept/src/components/MiniCart/cartOptions.js
```diff
@@ -116,48 +129,104 @@ class CartOptions extends Component {
...
+                                    <span>Quantity</span>
+                                </h2>
+                                <DecrementButtonFormApi formApi={formApi} />
+                                <Text
+                                    type="number"
```
brendanfalkowski: If decimals don't need to be supported, then `<input type="text" pattern="[0-9]*">` is a better option than `<input type="number">` because the virtual keyboard on touch devices is optimized for integer input.

packages/venia-concept/src/components/MiniCart/cartOptions.js
narendravyas24: Have replaced the "+" and "-" symbols with the corresponding svg's (Plus, Minus) imported from "react-feather".

packages/venia-concept/src/components/MiniCart/cartOptions.css
narendravyas24: @brendanfalkowski :
Have resolved the same in latest commit.

packages/venia-concept/src/components/MiniCart/cartOptions.js
narendravyas24: @brendanfalkowski 
Have resolved the same in latest commit.

packages/venia-concept/src/components/MiniCart/cartOptions.js
narendravyas24: @brendanfalkowski 
Have resolved the same in latest commit.

packages/venia-concept/src/components/ProductFullDetail/ProductFullDetail.js
narendravyas24: In my opinion (minus, input, plus) seems to be better enough. 
Because as in (minus, input, plus) minus and plus are separated by input so there can be less chances of confusion to user. 

Also I think as @brendanfalkowski suggested that two small buttons has too small targets, so we can prefer the (minus, input, plus) one.

Let me know your views, I can perform the changes as per the UI finalized.

Thanks.

soumya-ashok:
Thank you for the work on this.

1. I believe the recommendation when this was discussed on a separate thread was to have both increase and decrease buttons on the same side for greater ease of use.

This is the design I would prefer to see implemented, and there are 2 options.

*Option 1* - Start with the enabled button in the same state as the other selectors, and have the focus state while tapping be filled in with the dark gray and the disabled decrease button switching to the enabled state.



*Option 2* - Treat the increase/decrease the same as CTA buttons with the dark gray fill for enabled, and the focus state teal and the disabled decrease button switching to the enabled state.



2. It doesn't seem necessary to allow the user to make the quantity 0 or negative for more than one reason. The decrease button should be disabled when at the default quantity of 1.

3. We do need to consider the upper limit when the merchant does not allow more than a certain number of the product to be purchased, or if only a certain number are remaining.

4. The use case where the input is removed via the keyboard looks fine. I would change the message from "Please enter a valid number in this field" to "Enter a quantity of 1 or more"

brendanfalkowski:
> It doesn't seem necessary to allow the user to make the quantity 0 or negative for more than one reason. The decrease button should be disabled when at the default quantity of 1.

This is true for simple products, but for grouped or bundle items the user may choose (on the PDP) to set the quantity for specific items to 0 (if the admin has set the item quantities as customizable). So the stepper must be configurable to allow 0 so the user can add a bundle/grouped product without adding items from the set they don't want.

In our component, we allow passing a `minValue` and `maxValue` as props to override the default behavior and the stepping behavior adjusts to that.

I'm not sure if Venia is including support for bundle/grouped products, but that's a core Magento use case.

soumya-ashok:
@brendanfalkowski - In this implementation, all entering a negative or 0 value is doing is showing an error message that the input is invalid.

This iteration of Venia is for simple products, and considering that the default should be 1, my feedback still holds.

We will definitely consider other core Magento use cases such as grouped or bundled products in later iterations of Venia.

Thanks for the prompt feedback, as always!

brendanfalkowski:
@soumya-ashok — Definitely fair, the simple/configurable type is actually used 99% of the time. Bundle 1%. Grouped/Virtual/Downloadable I've never actually needed to implement more than 1-2 times.

narendravyas24:
Hi @soumya-ashok : looks good that we are considering the stepper, hope to listen feedback's from others as well so that final changes can be done.

Thanks.

soumya-ashok:
@narendravyas24 Great, if you have any questions regarding design feedback or need more clarification, please leave a comment and tag me.

narendravyas24:
> @narendravyas24 Great, if you have any questions regarding design feedback or need more clarification, please leave a comment and tag me.
Sure, thanks.

supernova-at:
> hope to listen feedback's from others as well so that final changes can be done.

Hi @narendravyas24 , do you have everything you need to continue work on this ticket?

narendravyas24:
> Hi @narendravyas24 , do you have everything you need to continue work on this ticket?

@supernova-at : I need some suggestion on one concern.

* In case when the product is "out of stock" or "merchant wants to restrict user to purchase more than a certain quantity of  product":
 I think these two cases should be managed from API end (as product quantity changes dynamically and an API response can give the accurate value).
I tried checking the response of add to cart API with the product "outofstock" and found that the response is not in a format to show validation message.

So apart from this everything that was needed is done. I have implemented the design changes and Stepper as per the shared mockups.

If the above mentioned validations are to be managed from API, you may review my till done work.

Please suggest.

Thanks.

packages/venia-concept/src/components/ProductQuantity/quantity.js
```diff
@@ -1,41 +1,86 @@
...
+                        decrementQty(formApi);
+                    }}
+                >
+                    <Minus />
+                </button>
+                <button
+                    aria-label="Quantity Increment"
+                    type="button"
+                    onClick={() => {
+                        incrementQty(formApi);
```
sirugh: Instead of passing an inline function here you could rewrite your `handleIncrement` callback. Since you are going to switch to `useFormApi` you can access the api within the callback to obtain the current value.

```js
const formApi = useFormApi();

const handleIncrement = useCallback(() => {
  const currentValue = formApi.getValue('quantity');
  ...
}, [formApi])

File: packages/venia-concept/src/components/ProductQuantity/quantity.js [hunk 7]
```diff
@@ -1,41 +1,86 @@
...
+Quantity.propTypes = {
+    classes: shape({
+        root: string,
+        quantityField: string
+    }),
+    items: arrayOf(
+        shape({
+            value: number
+        })
+    )
```
sirugh: Can you add the `onValueChange` prop here? It is type `func`, and it is required.
```suggestion
    ),
    onValueChange: func.isRequired
```

packages/venia-concept/src/components/ProductQuantity/quantity.js
```diff
@@ -1,41 +1,86 @@
...
-            })
-        )
+const Quantity = props => {
+    const { classes, initialValue, onValueChange, formApi, formState } = props;
+
+    const incrementQty = formApi => {
+        var currentQuantity = parseInt(formApi.getValue('quantity')) || 0;
+        const incrementedQty = currentQuantity + 1;
+        formApi.setValue('quantity', incrementedQty);
+        onValueChange(formApi.getValue('quantity'));
```
sirugh: You already have the new desired quantity in `incrementedQty`. There is no need to call `getValue` here or in the decrement callback.
```suggestion
        onValueChange(incrementedQty);
```

packages/venia-concept/src/components/ProductQuantity/quantity.js
```diff
@@ -1,41 +1,86 @@
...
+        const decrementedQty = currentQuantity - 1;
+        formApi.setValue('quantity', decrementedQty);
+        onValueChange(formApi.getValue('quantity'));
     };
 
-    render() {
-        const { classes, selectLabel, ...restProps } = this.props;
+    const validateQuantity = value => {
+        var validateRegex = /^[1-9]\d*(\.\d+)?$/;
+        return value <= 0
```
sirugh: Can you rewrite this to not use a nested ternary? As written it is much harder to read.

packages/venia-concept/src/components/ProductQuantity/quantity.js
```diff
@@ -1,41 +1,86 @@
...
+                    validateOnChange
+                    validate={validateQuantity}
+                    className={classes.quantityField}
                 />
-            </div>
-        );
-    }
-}
+                <button
+                    disabled={formApi.getValue('quantity') <= 1 ? true : false}
```
sirugh: Pull this out into a const before the return.

`const isDisabled = formApi.getValue('quantity') <= 1 ? true : false`

packages/venia-concept/src/components/ProductQuantity/quantity.js
```diff
@@ -1,41 +1,86 @@
...
+                    aria-label="Quantity Decrement"
+                    type="button"
+                    onClick={() => {
```
sirugh: Replace this inline function as well.

packages/venia-concept/src/components/ProductQuantity/quantity.js
```diff
@@ -1,41 +1,86 @@
...
+                        decrementQty(formApi);
+                    }}
+                >
+                    <Minus />
```
sirugh: We don't directly use the icons like this. Instead, use the `Icon` component and pass the desired sizing attributes:

`<Icon src={MinusIcon} size={16}`.

packages/venia-concept/src/components/ProductQuantity/quantity.js
```diff
@@ -1,41 +1,86 @@
...
+                </button>
+                <button
+                    aria-label="Quantity Increment"
+                    type="button"
+                    onClick={() => {
+                        incrementQty(formApi);
+                    }}
+                >
+                    <Plus />
```
sirugh: Please do the same here for the icon.

packages/venia-concept/src/components/ProductQuantity/quantity.js
```diff
@@ -1,41 +1,86 @@
...
+                </button>
+            </label>
+            <p className={classes.errors}>{formState.errors.quantity}</p>
```
sirugh: A nitpick, and not something you _must_ do, but it may be better to conditionally render this. You could replace it with:
```js
const errors = formState.errors ? 
  (<p className={classes.errors}>{formState.errors.quantity}</p>) :
  null;

...
return (
 ... 
 {errors}
);
```

packages/venia-concept/src/components/ProductQuantity/quantity.css
```diff
@@ -2,3 +2,41 @@
     display: flex;
     flex-direction: column;
 }
+.quantityFieldbox button {
+    margin: 0 2px;
```
sirugh: The spacing could use some work. Once you switch the icons over to the `Icon` component, can you try to make them centered appropriately? You may need to experiment with flexbox here.

[Image from Gyazo](https://gyazo.com/758e5a0830366079887c24307459ad60)

narendravyas24:
@sirugh : Thanks for review and sharing the feedbacks, I will work on the same and let you know at my soonest.

Thanks.

sirugh:
As it's been a few weeks since we've heard from the contributor I'm adding the `help-wanted` label to the issue.

If anyone wants to pick this work up to take it across the finish line it would be greatly appreciated!

For now I am closing this PR and moving the issue back.

PR #1337, opened 2019-06-13
Removes back arrow for first level in main menu.

### LINKED ISSUE #1334: Remove "back arrow" icon from first level of main menu
The current state is this - 

The back arrow performs the same thing as the close icon. It would be preferable to make the action clear by removing the back arrow and keeping just the close icon at the topmost level of the nav. menu.

narendravyas24:
## Description
Resolved bug #1334 . Removed back error to be shown for first level. 

## Related Issue
Closes #1334 .

## Verification Steps
1. Go to the Menu.
2. Back arrow is not shown for first level.

## Checklist:


packages/venia-concept/src/components/Navigation/navHeader.js
```diff
@@ -18,13 +18,17 @@ class NavHeader extends Component {
...
-                <Trigger key="backButton" action={onBack}>
-                    <Icon src={ArrowLeftIcon} />
-                </Trigger>
+                {!isCategoryRoot ? (
+                    <Trigger key="backButton" action={onBack}>
+                        <Icon src={ArrowLeftIcon} />
+                    </Trigger>
+                ) : (
+                    <div />
+                )}
```
supernova-at: I'd prefer defining a `backButton` or similar variable above the JSX and using that here.

For consistency, mainly, but also because with this pattern the reader has to mentally switch from JSX to JS back to JSX and that can be avoided easily.

packages/venia-concept/src/components/Navigation/navigation.js
```diff
@@ -307,11 +317,16 @@ class Navigation extends PureComponent {
 
         return (
             <aside className={className}>
-                <div className={classes.header}>
+                <div
+                    className={`${classes.header} ${
+                        isCategoryRoot ? classes.isRoot : null
+                    }`}
```
supernova-at: Same feedback as above. Please build up this value in a variable above the `return` statement and use it here.

packages/venia-concept/src/components/Navigation/navHeader.js
```diff
@@ -18,13 +18,17 @@ class NavHeader extends Component {
...
             <Fragment>
-                <Trigger key="backButton" action={onBack}>
-                    <Icon src={ArrowLeftIcon} />
-                </Trigger>
+                {!isCategoryRoot ? (
+                    <Trigger key="backButton" action={onBack}>
+                        <Icon src={ArrowLeftIcon} />
+                    </Trigger>
+                ) : (
+                    <div />
```
supernova-at: As far as styling goes, I was expecting this `div` to simply slot into the space that the `Trigger` normally takes up. Is the container not set up properly for that? Can it be?

supernova-at:
Hi @narendravyas24 ,  do you think you'll have any time to look at this this week? If not I can push some minor things to address the feedback.

Thanks!

narendravyas24:
> Hi @narendravyas24 , do you think you'll have any time to look at this this week? If not I can push some minor things to address the feedback.

Hi @supernova-at, Sorry for late reply. Yes I will work on the feedback's shared by you and will let you know asap.
Thanks.

packages/venia-concept/src/components/Navigation/navHeader.js
narendravyas24: @supernova-at : I have performed needed changes as per suggestion, please help to have a look into and share any other concern as well with me.
Thanks.

packages/venia-concept/src/components/Navigation/navigation.js
narendravyas24: ` const isRootClass = isCategoryRoot ? classes.isRoot : null;`

@supernova-at : I have performed suggested changes, please help to have a look into.

packages/venia-concept/src/components/Navigation/navHeader.js
narendravyas24: @supernova-at : Actually removing the back arrow in specific case was distorting the UI completely so in order to match my work according to the shared mockups, I thought it would be better to add a div.

> As far as styling goes, I was expecting this div to simply slot into the space that the Trigger normally takes up. Is the container not set up properly for that? Can it be?

Sorry but I am unable to get what you meant by div to simply slot into the space that the trigger normally takes up.

As I investigated on trigger there is a portion occupied by button that contains svg of back arrow, in my work I thought to replace that with an empty div in order to meet the design changes.

This was my approach but as said earlier that I found no way to adjust UI other than this, please suggest. 

Thanks.

narendravyas24:
@supernova-at : I have worked on feedbacks provided by you also I have updated the branch and performed all changes now in venia-ui.

Please help to have a look into.

packages/venia-ui/lib/components/Navigation/navHeader.js
```diff
@@ -17,13 +17,19 @@ class NavHeader extends Component {
     };
 
     render() {
-        const { classes, onBack, onClose, title } = this.props;
+        const { classes, onBack, onClose, title, isCategoryRoot } = this.props;
```
supernova-at: Please also add `isCategoryRoot` to this component's `propTypes`.

packages/venia-concept/src/components/Navigation/navHeader.js
supernova-at: Oh I just was hoping there was a way to do this without adding the extra `.isRoot` CSS class. No worries 👍

narendravyas24:
Hi @supernova-at ,

I have updated PR's code by taking latest pull from develop. 
I found that there were various recent changes deployed that were affecting the same files that I modified. 

Also components such as navHeader.js and navigation.js are converted to Functional one instead of Class one. 
So I have modified my code accordingly so that there may be no issue further in merging.

I may request you to have a look and share your feedback's if any. 

Thanks.

packages/venia-ui/lib/components/Navigation/navHeader.js
narendravyas24: @supernova-at , I am in no need to send this prop type anymore because after taking latest pull from develop and rework according to the latest deployed changes I found that there is already a variable i.e. isTopLevel been send in navHeader.

So I have used the same. Please have a look and let me know in case if you have any feedback in this.

Thanks.

packages/venia-ui/lib/components/Navigation/navHeader.js
```diff
@@ -26,11 +26,17 @@ const NavHeader = props => {
     const title = titles[view] || titles.MENU;
     const backIcon = isTopLevel && view === 'MENU' ? MenuIcon : ArrowLeftIcon;
 
+    const backButton = !isTopLevel ? (
+        <Trigger key="backButton" action={onBack}>
+            <Icon src={backIcon} />
+        </Trigger>
+    ) : (
+        <div />
```
sirugh: Do we need to render an empty `div`? Is this for spacing?

narendravyas24: Yes it is to adjust spacing. I was unable to find any other way.
If you have any other suggestion, please let me know.

narendravyas24:
> Because of the rewrite of the navigation menu this PR now prevents the back button from showing on any "top level" menu. Because of this, the UX for things like "sign in" or "my account" will no longer have a back arrow. Try it yourself on your PR. Open the menu and click "sign in". You won't be able to navigate back

Thanks for pointing that out @sirugh . I have modified my code and have checked that now back-arrow is available in Sign-in and Create-account as well.

Please help to have a look into and let me know your views.

narendravyas24:
> You'd also have to adjust the CSS accordingly for the other root views, though.
Yes there are some places where css need to be checked, I will check that.

sirugh:
Can you fix the CSS? Look at any of the non "main-menu" pages ie sign in, create account, etc.

[Image from Gyazo](https://gyazo.com/6f3db7c80edd36459cd3bc98cc344bfb)

narendravyas24:
> Can you fix the CSS? Look at any of the non "main-menu" pages ie sign in, create account, etc.
Yeah saw that, will do that.

narendravyas24:
> Can you fix the CSS? Look at any of the non "main-menu" pages ie sign in, create account, etc.

@sirugh : Have fixed that one, please have a look into and share your feedback if any.

> Also test snapshots are broken, you'll need to update those.

Sorry, but I am not aware about what is **test snapshots**, can you please help me in pointing towards right direction.
Thanks.

sirugh:
To update snapshots run `jest -u` and commit the updated snapshot _if_ they make sense. See [here](https://jestjs.io/docs/en/snapshot-testing) for info on snapshot testing :)

narendravyas24:
Hi @sirugh : I have updated the test snapshots and have also resolved all the concerns shared by you. 

Also tested on: https://pr-1337.pwa-venia-staging.com/ 

Please help to have a look into and let me know in case if you have any concern.

Thanks.

packages/venia-ui/lib/components/Navigation/navHeader.js
narendravyas24: @sirugh : Please help to have a look at the changes, I have worked on the concerns shared by you.

narendravyas24:
@sirugh , @revanth0212 : I have resolved the changes as requested please help to review.

packages/venia-ui/lib/components/Navigation/navigation.js
```diff
@@ -76,9 +76,11 @@ const Navigation = props => {
         setView('SIGN_IN');
     }, [setView]);
 
+    const isRootClass = isTopLevel && view === 'MENU' ? classes.isRoot : null;
```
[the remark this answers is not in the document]
narendravyas24: @revanth0212 : I have modified the name as "menuTopLevelClassName" instead of "rootClassName" as it was already given to "aside".
Have also removed the null and replaced with ' '.

Please review.

packages/venia-ui/lib/components/Navigation/navHeader.js
jimbo: > Yes it is to adjust spacing. I was unable to find any other way. If you have any suggestion regarding this, please let me know.

That's certainly one way to do it. 😅

Since you're already making the grid layout dynamic based on whether it's the root, you could just change the number of columns and add some left padding. Then you could render `null` instead of an empty `div`.

```css
.isRoot {
  grid-template-columns: 1fr 3.5rem; /* title, close icon */
  padding-left: 1rem;  /* so the title isn't up against the edge */
}
```

Also, for what it's worth, we don't do class concatenation in our components. Each element should have no more than one classname from the imported object; you can use `composes` to combine styles.

```jsx
const Foo = props => {
  const classes = mergeClasses(defaultClasses, props.classes)

  // `.root` might have some styles of its own
  // `.root_open` and `.root_closed` might compose `.root`
  // and add some extra styles
  const rootClass = props.isOpen ? classes.root_open : classes.root_closed

  return (
    <div className={rootClass} />
  )
}
```

jimbo:
@narendravyas24 Thanks for submitting this PR, and for keeping it updated even as the `Navigation` component changed after you made this branch. I've left some feedback, which you can address if you like. If not, we can merge as-is. 👍

narendravyas24:
@jimbo : Thanks for providing the valuable feedbacks. 

I have worked in order to resolve the feedbacks shared by you, please help to have a look into and let me know if in case anything else is been needed.

Thanks.

packages/venia-ui/lib/components/Navigation/navHeader.js
narendravyas24: @jimbo : Thanks for those suggestions, I have worked on the same. Please review.
Thanks.

PR #1338, opened 2019-06-13
[feature]: refactor edit steps out of redux and into local checkout state

### LINKED ISSUE #1405: [feature]: Refactor local state for checkout out of redux
Currently almost all state used within checkout is managed via redux.
Now that we have hooks and are using function components we can more easily manage local state without having to use something like redux.

I want to evaluate the usage of redux within `Checkout/` and refactor to local state things that are better managed locally such as the `editing` state.

- [ ] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
## Description

This is a followup to #1309. The original refactor left much of the original redux piping in place. This PR shows how we can remove parts of it relatively painlessly. Here we are migrating the state of what form is being edited, ie payment, address or shipping, into local form state. 

I also removed some unnecessary props and added some safety checks/state to make sure we only enable the submission of billing information when appropriate.

## Related Issue

Closes #1405 

## Verification Steps

1. Open checkout and work through the steps.
2. You should be able to open and close each step in no particular order, and no errors should occur.

## Checklist:


packages/venia-concept/src/components/Checkout/editableForm.js
```diff
@@ -23,34 +23,37 @@ const EditableForm = props => {
...
     const handleCancel = useCallback(() => {
-        editOrder(null);
-    }, [editOrder]);
+        setEditing(null);
+    }, [setEditing]);
 
     const handleSubmitAddressForm = useCallback(
-        formValues => {
-            submitShippingAddress({
+        async formValues => {
```
sirugh: We had to make some of the callbacks `async` so that they could appropriately handle an error response.

packages/venia-concept/src/components/Checkout/braintreeDropin.js
```diff
@@ -91,25 +89,16 @@ const BraintreeDropin = props => {
...
 
     useEffect(() => {
         async function requestPaymentNonce() {
             try {
                 const paymentNonce = await dropinInstance.requestPaymentMethod();
                 onSuccess(paymentNonce);
             } catch (e) {
-                // If payment details were missing or invalid but we have data
-                // from a previous successful submission, use the previous data.
-                const storedPayment = storage.getItem('paymentMethod');
```
sirugh: This was a weird UX. If storage has a valid payment and I enter a new card but _something_ is invalid, it'll just close the form and a user will assume the new info was accepted. I don't think we should do that automatically, so I removed this code. In the future we could display some UX that shows a "edit" button next to saved CC info that would re-mount the braintree dropin but for now we just have to deal with re-entering CC info.

supernova-at: Agreed it was a weird UX. If I remember correctly we were trying to avoid having to re-enter CC info if the only thing that was changing was the billing address.

I think you're right that even though re-entering CC info is cumbersome it is better than leaving the user uncertain or unknowledgeable about what CC info is really being used. I'd rather them have to be explicit.

packages/venia-concept/src/components/Checkout/__tests__/form.spec.js
```diff
@@ -9,9 +10,15 @@ import Section from '../section';
 import Button from 'src/components/Button';
 
 jest.mock('src/classify');
+jest.mock('../addressForm', () => 'AddressForm');
+jest.mock('../paymentsForm', () => 'PaymentsForm');
+jest.mock('../shippingForm', () => 'ShippingForm');
+jest.mock('react', () => {
+    const React = jest.requireActual('react');
+    return Object.assign(React, { useState: jest.fn(React.useState) });
+});
```
jimbo: There are a lot of ways to write tests, so I tend to let most things pass, but I think this is overkill. See the comments below.

packages/venia-concept/src/components/Checkout/__tests__/form.spec.js
```diff
@@ -37,114 +43,171 @@ const defaultProps = {
...
+test('clicking address form edit sets `editing` state value to "address"', () => {
+    const mockSetEditing = jest.fn();
+    useState.mockReturnValueOnce([null, mockSetEditing]);
 
-    component.root.findAllByType(Button)[0].props.onClick();
+    const component = createTestInstance(<Form {...defaultProps} />);
 
-    expect(mockCancelCheckout).toHaveBeenCalled();
+    act(() => {
+        component.root.findAllByType(Section)[0].props.onClick();
```
jimbo: Here, an implementation detail of a different component is being tested. `Form` passes `setEditing` to `EditableForm` without any knowledge that it'll end up being a click handler, yet this test drills down even further into `EditableForm` to find a particular child (again, unknown) and calls its click handler.

All this test should do is verify that `setEditing` is passed to its children. We don't need to mock `useState` or verify that `setEditing` updates state.

sirugh: Yea so what I ended up doing was not writing tests for each of the sub-components that were created in this PR. Instead I just mocked React and wrote tests that could apply to `EditableForm` and `Overview`.

I have a few options:
1) Split all these tests into individual files
2) Remove them from this test entirely.
3) Leave it as is.

What would you like me to do?

jimbo: Definitely split the tests into files that correspond to one component each. We're only writing unit tests.

Most of these are simple enough that a snapshot test should get you 100% coverage.

sirugh:
Uh idk why the tests are failing...

PR #1349, opened 2019-06-13
List component refactor.

revanth0212:
## Description
Moving List Component from Stateful Class Components to Stateful Functional Components using Hooks.

## Related Issue
Closes https://github.com/magento-research/pwa-studio/issues/1328


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1349) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1349) before we can accept your contribution.<br/><hr/>**Revanth Kumar Annavarapu** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

revanth0212:
According to the issue, we should be moving away from render props structure. Before making the change I wanted to know if we will be using this component anywhere else other than current usage? If yes, then I feel render props structure is probably one of the right ways to get this component up and running. If we bake the rendering logic into the component, it won't be extensible in the future, if another app wants to render it in some other way. Any thoughts on that?

packages/peregrine/src/List/items.js
```diff
@@ -1,104 +1,112 @@
...
+            <ListItem
+                key={key}
+                item={item}
+                itemIndex={index}
+                render={renderItem}
+                hasFocus={hasFocus && cursor === key}
+                isSelected={selection.has(key)}
+                onBlur={removeFocus}
+                onClick={updateSelection(key)}
+                onFocus={setFocus(key)}
```
jimbo: Without memoization, these callbacks will be recreated on each render. After creating the custom hook described earlier, we should work out what new props to pass to `ListItem` and have that component create the handlers itself.

packages/peregrine/src/List/item.js
```diff
@@ -1,54 +1,47 @@
...
-            <Root className={classes.root} {...customProps} {...restProps}>
-                {this.children}
-            </Root>
-        );
-    }
+const isString = item => typeof item === 'string';
+
+const getChild = item => (isString(item) ? item : null);
+
+function Item(props) {
```
jimbo: Like all the other functional components in the codebase, these ones should be defined as constants.

```es6
const Item = props => {
  // ...
}
```

packages/peregrine/src/List/item.js
```diff
@@ -1,54 +1,47 @@
...
+        classes,
+        hasFocus,
+        isSelected,
+        item,
+        itemIndex,
+        render,
+        ...restProps
+    } = props;
+    const customProps = { classes, hasFocus, isSelected, item, itemIndex };
+    const Root = fromRenderProp(render, Object.keys(customProps));
```
jimbo: This can probably also be memoized.

```es6
const Root = useCallback(
  () => fromRenderProp(render, Object.keys(customProps)),
  [customProps, render]
)
```

jimbo:
> If we bake the rendering logic into the component, it won't be extensible in the future if another app wants to render it in some other way. Any thoughts on that?

Right, so the goal would not be to bake the render logic into the component. Rather, the goal would be to make these custom hooks instead of components, since all they do is store some state and provide an API for changing that state. The only reason they're components at all is because that's how render props work; as hooks, they can accept arguments and return values without calling `createElement()` at all.

Of course, that will affect every component in Venia that depends on them, so we'll probably just add the hooks and deprecate the components.

packages/peregrine/src/List/list.js
```diff
@@ -9,89 +9,84 @@ import Items from './items';
...
-        );
-    }
-
-    handleSelectionChange = selection => {
-        const { onSelectionChange } = this.props;
-
-        if (onSelectionChange) {
-            onSelectionChange(selection);
-        }
+        },
```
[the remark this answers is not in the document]
revanth0212: My Bad. Changed the syntax to represent a function instead.

packages/peregrine/src/List/list.js
```diff
@@ -9,89 +9,84 @@ import Items from './items';
...
+        [classes, getItemKey, items, onSelectionChange, selectionModel]
+    );
+    const Root = useMemo(fromRenderProp(render, Object.keys(customProps)), [
+        render,
+        customProps
+    ]);
```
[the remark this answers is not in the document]
revanth0212: My bad. Ill change the syntax. Coming to useCallback, Isn't Root a value? It's a Component right. I understand it is a function because every react component is a function in its core, but I thought using a memo makes more sense since memo is a callback that is already called.

```
// psuedo code
const memoizedCallback = useCallback(fn)
const useMemo = memoizedCallback()
```

packages/peregrine/src/List/list.js
```diff
@@ -9,89 +9,84 @@ import Items from './items';
...
+    const handleSelectionChange = selection => {
```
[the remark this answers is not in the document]
revanth0212: Will `onSelectionChange` not get a new reference on every render?

packages/peregrine/src/List/items.js
```diff
@@ -1,104 +1,112 @@
...
+        selectionModel,
+        onSelectionChange
+    } = props;
+    const [{ cursor, hasFocus, selection }, dispatch] = useReducer(
+        reducer,
+        initialState
+    );
+    useEffect(() => {
+        onSelectionChange && onSelectionChange(selection);
+    }, [Array.from(selection).toString()]); // when ever the selection changes, make the call
```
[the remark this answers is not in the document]
revanth0212: Sure, will make the changes for comments accordingly. Regarding the `selection` object, will it not get a new reference on every render?

packages/peregrine/src/List/items.js
```diff
@@ -1,104 +1,112 @@
...
+            return {
+                ...state,
+                hasFocus: true,
+                cursor: action.key
+            };
+        case 'UPDATE_SELECTION': {
+            const { key, selectionModel } = action;
+            return {
+                ...state,
+                selection: updateSelection(key, state.selection, selectionModel)
```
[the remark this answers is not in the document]
revanth0212: Nice catch. My bad. But how did it work? 😕

packages/peregrine/src/List/items.js
```diff
@@ -1,104 +1,112 @@
...
+
+function Items(props) {
+    const {
+        getItemKey,
+        items,
+        renderItem,
+        selectionModel,
+        onSelectionChange
+    } = props;
+    const [{ cursor, hasFocus, selection }, dispatch] = useReducer(
```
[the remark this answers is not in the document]
revanth0212: Sure will do. Any particular reason behind it? Reusability or Separation or Concerns?

packages/peregrine/src/List/items.js
revanth0212: Will something like this solve our problem of recreating handlers upon re-render?

`Commit: 55df3e063fc8f78cacf3e94e727c48ffb9360575`

```
const key = getItemKey(item, index);
const updateSelection = useMemo(() => _updateSelection(key), [ key ]);
const setFocus = useMemo(() => _setFocus(key), [key]);
```

packages/peregrine/src/List/item.js
revanth0212: Did

```
const Root = useMemo(
  fromRenderProp(render, Object.keys(customProps)),
  [customProps, render]
)
```

which is the same since useMemo is a useCallback which is already executed.

packages/peregrine/src/List/list.js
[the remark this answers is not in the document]
sirugh: Maybe, but the benefit is that `handleSelectionChange` will be memoized based on the reference so you can optimize up the tree to wherever/whatever is making `onSelectionChange` different.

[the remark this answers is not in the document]
revanth0212: Sure. Will memoize it using `useCallback` but we should be doing this with an assumption that the parent whoever is using this component will memoize `onSelectionChange` before passing it down to the `List` component as prop. If not, we will be assigning more work to the garbage collector since now it has to clean up 2 functions instead of 1 on every render.

packages/peregrine/src/List/items.js
```diff
@@ -1,104 +1,137 @@
...
+        selectionModel,
+        onSelectionChange
+    } = props;
+    const [
+        { cursor, hasFocus, selection },
+        { setFocus: _setFocus, removeFocus, updateSelection: _updateSelection }
+    ] = useListState({ selectionModel, onSelectionChange });
+    return (
+        <Fragment>
+            {items.map((item, index) => {
```
jimbo: Avoid using multiline expressions in JSX; if you have to map some data to elements, try to define that as a `const` before `return`.

```jsx
const itemElements = items.map(item => <div />)

return (
  <Fragment>
    {itemElements}
  </Fragment>
)
```

packages/peregrine/src/List/items.js
```diff
@@ -1,104 +1,137 @@
...
+                const key = getItemKey(item, index);
+                const updateSelection = useMemo(() => _updateSelection(key), [
```
jimbo: Unfortunately, [hooks can't be called inside loops.](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level). In situations where you would call a hook once for each child, the recommended solution is to have the child (or perhaps a wrapper) call the hook instead.

Consider ways to allow `Item` to call `useMemo`, or to not need to call it.

revanth0212: You are right. My bad. Fixed it.

revanth0212:
@jimbo all the requested changes have been addressed.

packages/peregrine/lib/List/item.js
```diff
@@ -1,54 +1,63 @@
...
+        updateSelection,
+        setFocus,
+        ...restProps
+    } = props;
+    const onClick = useCallback(() => updateSelection(key), [
+        updateSelection,
+        key
+    ]);
+    const onFocus = useCallback(() => setFocus(key), [setFocus, key]);
+    const customProps = {
```
jimbo: I think you'll want to add `uniqueId` to this object so it doesn't get applied as a DOM attribute.

revanth0212: In this case, it wouldn't because the execution stops at line [19 of fromRenderProp](https://github.com/magento-research/pwa-studio/blob/develop/packages/peregrine/lib/util/fromRenderProp.js#L19) and not enter [36 of fromRenderProps](https://github.com/magento-research/pwa-studio/blob/develop/packages/peregrine/lib/util/fromRenderProp.js#L36) where the filtering actually happens. This is because the element to render is a composite element.

packages/peregrine/lib/List/useListState.js
```diff
@@ -0,0 +1,74 @@
...
+        if (selection.has(key)) {
+            selection.delete(key);
+        } else {
+            selection.add(key);
+        }
+    }
+    return selection;
+};
+
+const reducer = (state, action) => {
```
jimbo: Could we clean this reducer up to be more like the other reducers we have? Have a look at [`useQueryResult`](https://github.com/magento-research/pwa-studio/blob/develop/packages/peregrine/lib/hooks/useQueryResult.js) for an example.

- [ ] change the second argument, `action`, to have the shape `{ payload, type }`
- [ ] wrap the reducer with `withLogger`
- [ ] memoize an `api` object with `useMemo` and return it from `useListState`
- [ ] add docblock comments

```es6
// define some initial state
const initialState = {}

// define the reducer
const reducer = (state, { payload, type }) => {
  case "": {
    // ...
  }
  default: {
    return state
  }
}

// define the hook
const useListState = () => {
  const [state, dispatch] = useReducer()

  const setFocus = useCallback(
    // ...
  )

  // memoize the api, and include dispatch
  const api = useMemo(
    () => {
      dispatch,
      setFocus
    }
  )

  return [state, api]
}
```

jimbo:
Just make sure the tests are passing and the build works. This is ready for QA, I think. 👍

PR #1351, opened 2019-06-13
Header: Replace Suspense Fallback for SearchBar

### LINKED ISSUE #1273: [bug]: [RE-Opened ] - Second search icon breaks header bar during load
**Describe the bug**
For a few milliseconds during load, this happens:

doubled mag

And then it's back to this:

single mag

**To Reproduce**
1. Completely clear cache and launch incognito mode.
2. Start a dev server.
3. Use the Chrome dev tools mobile emulator to artificially limit network and CPU speed.
4. Load some page.

**Expected behavior**
No jank from header.

**Possible solutions**
It's probably something small.

 - iPhone SE and Nokia 2
 - iOS and Android
 - Safari and Chrome
 - PWA Studio Version: 2.1.0

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

tjwiebell:
## Description
Fix edge case scenario on slow devices where Suspense temporarily loads a fallback component while the SearchBar component is being loaded.

## Related Issue

Closes #1273 

## Verification Steps

1. Simulate Slow 3G in network tab and navigate to `/search.html?query=top`
2. Observe new placeholder where search icon used to be misaligned

## How have YOU tested this?

Ran `yarn test` and manually checked above steps

## Checklist:

### DIFF
--- a/packages/venia-concept/src/components/Header/header.css
+++ b/packages/venia-concept/src/components/Header/header.css
@@ -69,17 +69,20 @@
 .loader,
 .loader:before,
 .loader:after {
+    --dot-size: 2em;
+    --dot-font-size: 6px;
+    --dot-shadow-offset: calc(-1 * var(--dot-size) + var(--dot-font-size));
     border-radius: 50%;
-    width: 2em;
-    height: 2em;
+    width: var(--dot-size);
+    height: var(--dot-size);
     animation-fill-mode: both;
     animation: pulse 1.8s infinite ease-in-out;
 }
 
 .loader {
     color: rgb(var(--venia-grey-dark));
-    font-size: 6px;
-    margin: calc(-2em + 6px) auto 0;
+    font-size: var(--dot-font-size);
+    margin: var(--dot-shadow-offset) auto 0;
     position: relative;
     transform: translateZ(0);
     animation-delay: -0.16s;
@@ -99,7 +102,7 @@
 }
 
 .loader:after {
-    color: rgb(var(--venia-text-alt));
+    color: rgb(var(--venia-grey-darker));
     left: 3.5em;
 }
 
@@ -107,9 +110,9 @@
     0%,
     80%,
     100% {
-        box-shadow: 0 2em 0 -1.3em;
+        box-shadow: 0 var(--dot-size) 0 -1.3em;
     }
     40% {
-        box-shadow: 0 2em 0 0;
+        box-shadow: 0 var(--dot-size) 0 0;
     }
 }
--- a/packages/venia-concept/src/components/Header/header.css
+++ b/packages/venia-concept/src/components/Header/header.css
@@ -73,8 +73,9 @@
     width: 2em;
     height: 2em;
     animation-fill-mode: both;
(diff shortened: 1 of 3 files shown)


packages/venia-concept/src/components/Header/header.css
```diff
@@ -56,3 +56,60 @@
...
+    width: 2em;
+    height: 2em;
+    animation-fill-mode: both;
+    animation: pulse 1.8s infinite ease-in-out;
+}
+
+.loader {
+    color: rgb(var(--venia-grey-dark));
+    font-size: 6px;
+    margin: calc(-2em + 6px) auto 0;
```
sirugh: Why the calc?

packages/venia-concept/src/components/Header/header.css
```diff
@@ -56,3 +56,60 @@
...
+}
+
+.loader:before {
+    color: rgb(var(--venia-grey));
+    left: -3.5em;
+    animation-delay: -0.32s;
+}
+
+.loader:after {
+    color: rgb(var(--venia-text-alt));
```
sirugh: Going from `grey` and `grey-dark` to `text-alt` doesn't immediately make sense. Can you create a new variable to better represent that the `:after` loader is even darker than `grey-dark`?

packages/venia-concept/src/components/Header/header.css
tjwiebell: The animation is using box shadow, which needed to be offset by the height and font-size. Pulled this from an animation tutorial, so happy to use another effect if you have a suggestion.

sirugh: Oh I see. Maybe pull those into a `var` each to be clear about whats going on?

dpatil-magento:
QA Pass. @sirugh  If new changes are good, please go ahead and merge this one.

packages/venia-concept/src/components/Header/header.css
```diff
@@ -56,3 +56,66 @@
...
+.searchFallback {
+    display: flex;
+    justify-content: center;
+    margin-bottom: 0.5rem;
+    padding: 0 1rem;
+}
+
+.input {
+    composes: input from '../TextInput/textInput.css';
+    max-width: 24rem;
```
sirugh: This value comes from `searchBar.css` as we want the placeholder to look identical to the `searchBar` once it loads. I think this should be a constant somewhere shared between both css files so we only have to update in one place in the future.

packages/venia-concept/src/components/Header/header.css
```diff
@@ -56,3 +56,66 @@
...
+
+.loader:before,
+.loader:after {
+    content: '';
+    position: absolute;
+    top: 0;
+}
+
+.loader:before {
+    color: rgb(var(--venia-grey));
```
sirugh: Not to nitpick but on my screen this is super light and really difficult to see. If this value is from the mocks, fine, but maybe it needs to be darker.

PR #1358, opened 2019-06-17
[dev]: Add eslint plugin for react hooks

### LINKED ISSUE #1357: [dev]: Update eslint for react hooks
We started using hooks with the Peregrine refactors but we did not update to our latest [eslint config](https://github.com/magento-research/magento-eslint/commit/5a35555daba8dd9fbf401c04d28e021e9dd563db) which provides access to the `eslint-plugin-react-hooks` plugin. So let's upgrade!

sirugh:
## Description
We use hooks now. We don't lint the hooks. We should lint the hooks. I added released a new version of the `magento-eslint` repo, updated the version here, added the requisite plugin, and ran it once to fix errors (the only one was in `app.js`). In this PR I also fixed the warnings. There were two cases around `addToast` where we did not update the dependency according to the warning. These are potential bugs and should be addressed with this issue: #1377 

## Related Issue

Closes #1357.

## Verification Steps
Run `yarn run lint` and make sure there are no errors.
Also test:
* Offline/Online (useToasts)
* Error handling (useToasts), throw an error in miniCart.js#render
* searchBar (useDropdown)
* Thumbnail (useWindowSize)
* autocomplete (useQuery)
* category pages (useQuery)

Things tested inherently:
* useEventListener, tested by useWindowSize and useDropdown

## Checklist:


packages/venia-concept/src/components/App/app.js
```diff
@@ -69,17 +69,7 @@ const App = props => {
                 addToast(errorToastProps);
             }
         }
-    }, [errors, handleDismiss]);
-
-    if (renderError) {
```
sirugh: Because there was a `useEffect` hook below this condition we got an error which said that we should always call the same amount of hooks in the same order every render. A conditional breaks this requirement, so I moved it below all the hooks.

tjwiebell:
@sirugh - I wouldn't feel comfortable merging this PR as-is, it seems important to either fix all the `exhaustive-deps` warnings, or remove the rule. Using `--fix` didn't go smoothly for me, so would require some human interaction.

Let's poll the team, but I would propose suppress in scope of this PR, and create a new issue to add the rule back with fixes (if the team is on board).

sirugh:
I'll fix the errors.

packages/venia-concept/src/components/ToastContainer/__stories__/toasts.js
```diff
@@ -33,7 +33,7 @@ const ToastEmitter = ({
 
     useEffect(() => {
         addToast(toastProps);
-    }, []);
+    }, []); // eslint-disable-line
```
tjwiebell: You mentioned this would be an exception to the rule; a comment may be good here so the community knows of valid reasons to suppress this lint rule. I think I'm just going to approve though, IMO Stories aren't API, and really shouldn't be used as implementation references.

sirugh: I added a comment here and in the other place where we added an `eslint-disable-line`.

packages/venia-concept/src/components/App/app.js
```diff
@@ -69,17 +72,7 @@ const App = props => {
...
-        return (
-            <Fragment>
-                <Main isMasked={true} />
-                <Mask isActive={true} />
-                <ToastContainer />
-            </Fragment>
-        );
-    }
+    }, [errors]); // eslint-disable-line
```
sirugh: This and the other eslint-disable should be addressed by #1377

tjwiebell: I think you accidentally removed `handleDismiss` from the array; re-add and this is ready to QA.

packages/venia-concept/src/components/App/__tests__/__snapshots__/app.spec.js.snap
```diff
@@ -390,13 +391,145 @@ ReactTestInstance {
                       "loc": "1",
                     },
                   ],
-                  [Function],
                 ],
                 "destroy": undefined,
-                "next": [Circular],
+                "next": Object {
```
tjwiebell: No problem here, but just wanted to document that we suspect the diff in all these React internals is because of moving all effects above any renders in `app.js`. To be safe, I'm requested @dpatil-magento does a full sanity check to make sure this didn't introduce any unexpected behavior.

supernova-at:
Picking up QA on this.

supernova-at:
QA Pass 👍

PR #1361, opened 2019-06-17
[feature]: Add height to `makeUrl` and enforce crop in middleware if provided

### LINKED ISSUE #1359: [feature]: Image middleware should allow `height` and `crop`
The product images from sample data are 25:31 ratio.
The placeholder image we use is 4:5.

This is a difference of only a few pixels but it leads to a poor UX where we sometimes have a placeholder image that is too large (ie. on the category page) or we have a UI that "shifts" down/up when we render a placeholder and then replace it with the product image (ie. on the pdp).

I would like to allow our image resizer (sharp middleware) to crop/resize based on a height parameter. Currently we only pass the width and let it resize to the width based on the image's original ratio.

Ideally we would have sample data and a placeholder that have identical ratios but apparently it is difficult to regenerate/remake sample data and the smallest ratio that would match our sample data is `370:379` 🤣 

**Additional context**
[Image from Gyazo](https://gyazo.com/cf52b9b14184d77d5b6f8043db4922f4)

- [x] `venia-concept`
- [x] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
## Description
The product images from sample data are 25:31 ratio.
The placeholder image we use is 4:5.

This is a difference of only a few pixels but it leads to a poor UX where we sometimes have a placeholder image that is too large (ie. on the category page) or we have a UI that "shifts" down/up when we render a placeholder and then replace it with the product image (ie. on the pdp).

I made a change here to the `makeUrl` to accept a `height` param. I also added some logic to our middleware to now enforce `crop` if both height and width are provided.

To understand these changes consider the following. Say we have an image with ratio 25:31 (as is our sample data) and we want to size it. The following now can happen:

a) Using `width=640`, sharp and fastly will scale to `640x796` because they retained the aspect ratio.
b) Using `width=640&height=800`, fastly will scale to `640x800` but sharp will scale to `645x800` because of the `fit: outside` [prop](https://github.com/magento-research/express-sharp/blob/develop/lib/transform.js#L23) used. Now, with these changes sharp will scale and then crop to `cover` the provided height and width.

If a user wants to retain the aspect ratio of the image using either fastly or our express-sharp, they can still just pass `width`.

Read more here:
[sharp resize](https://sharp.pixelplumbing.com/en/stable/api-resize/)
[fastly](https://docs.fastly.com/api/imageopto/width)

## Related Issue

Closes #1359 .

## Verification Steps

1. Go to the category page and notice that the images are sized correctly and there is no overflow from the placeholder!
2. Update your backend url to one of our cloud instances with fastly enabled. I can give you a URL if you need one. Then set `IMAGE_OPTIMIZING_ORIGIN="backend"` in your `.env` and re-run your server.
3. Go back to the category page and images should still be sized correctly.

## How have YOU tested this?

I made a change to the category page `GalleryItem` which is only _one_ implementation of `makeUrl`. Ideally we would just us these changes elsewhere when we identify a problem.

In the BEFORE screenshot notice the grey rectangle below the product images.
BEFORE
[Image from Gyazo](https://gyazo.com/cf52b9b14184d77d5b6f8043db4922f4)
AFTER
[Image from Gyazo](https://gyazo.com/ff8a21b3f4c7a55a3f7d04ce62782d9c)

## Checklist:


zetlen:
The URLs made by `makeUrl` need to adhere to the [Fastly image opto](https://docs.fastly.com/api/imageopto/crop) query parameter API. That's why we have the rewrite function in `addImgOptMiddleware` that transforms `makeUrl` URLs into the kind that `express-sharp` can understand.

The reason we follow Fastly's image opto API is that we want our customers to be on Enterprise Cloud and to have Fastly IO enabled; that should be the "happy path" and the simplest thing. In that simplest case, Fastly takes over all the image manipulation (and it's super fast at it). Since Fastly is an external service and our express-sharp implementation is an internal one, we can freely mess with the `addImgOptMiddleware` public API.

So you just gotta change so `makeUrl` outputs a `crop` parameter adhering to the Fastly spec linked above. It can be internal to `makeUrl` if you want.

Test it by connecting your local instance to one of our [Fastly-enabled Enterprise Cloud shops](https://wiki.corp.magento.com/display/FF/PWA+Cloud+Instances). Then set the env var `IMAGE_OPTIMIZING_ORIGIN=backend` and it'll render media URLs at the attached store base domain, instead of your local. Then you should see Fastly in operation (and you can confirm it by looking at response headers).

packages/pwa-buildpack/src/Utilities/addImgOptMiddleware.js
```diff
@@ -90,6 +100,10 @@ https://github.com/nodejs/node-gyp#installation`
             if (incomingQuery.auto === 'webp') {
                 params.set('format', 'webp');
             }
+            if (incomingQuery.crop === 'true') {
+                params.set('true', 'crop');
```
sirugh: o.o How did this get switched...

sirugh:
Looks like Fastly is actually a good citizen when it comes to resizing. When you give a height and width that don't match the aspect ratio they must resize the image automatically whereas express-sharp retains the aspect ratio and matches the latter param (height). So what I need to do is figure out if there is a way to get express-sharp to resize in the same way that fastly does (looks like crop is the answer right now).

packages/pwa-buildpack/src/Utilities/addImgOptMiddleware.js
```diff
@@ -70,12 +70,22 @@ https://github.com/nodejs/node-gyp#installation`
...
 
-            const rewritten = new URL(
-                `https://0.0.0.0/resize/${incomingQuery.width}`
-            );
+            const { width, height } = incomingQuery;
+            let rewrittenUrl = 'https://0.0.0.0/resize/';
 
+            if (width) {
+                rewrittenUrl += `${width}/`;
+            } else {
```
tjwiebell: This section seems unreachable given the check on 132. Can simplify and remove.

sirugh: What check are you referencing? If you request with `height=300` and no width this will error.

tjwiebell: https://github.com/magento-research/pwa-studio/pull/1361/files#diff-4038ce758492d17b6add3daa61c02f5eR132

It won't enter this function w/o width.

sirugh: Oh haha nice catch.

sirugh:
@tjwiebell addressed feedback. Please move to Ready for QA if it looks good!

dpatil-magento:
@sirugh Other than below observations rest all looks good.
1. Image size is now 300x375  vs 300x372   on `develop` branch.
2. Horizontal border between image and the product name is missing on this branch compared to `develop`

sirugh:
> Image size is now 300x375 vs 300x372 on develop branch.

Expected, we are now requesting `height` and `width`. Glad you caught it though ;) 

> Horizontal border between image and the product name is missing on this branch compared to develop

So that's not actually an intentional border. It is the overlap of the placeholder that is sized slightly too large and ends up showing a bit below the full image.

PR #1366, opened 2019-06-18
(#1332) Add timeout on tooltip to hide automatically

### LINKED ISSUE #1332: Color tooltip needed on PD page
Immediately upon color selection on the product detail page, this is the current state -

The desired state is -

realchriswells:
## Description
Added timeout on the Tooltips so that they automatically disappear after 2.5s

## Related Issue

Closes #1332 .

## Verification Steps

1. Go to the PDP.
2. Click on a swatch (e.g. colour or size).
3. Tooltip should appear. Tooltip will disappear after 2.5s

## How have YOU tested this?

Checked the timeout in Chrome 75 on MacOS.

## Checklist:


packages/venia-concept/src/components/ProductOptions/toolTip.js
```diff
@@ -62,6 +62,11 @@ class Tooltip extends Component {
         clearTimeout(this.timeoutId);
 
         this.setState({ isShowing: true });
+
+        // Creates a timer to hide the tooltip after 2.5s
+        this.timeoutId = setTimeout(() => {
+            this.setState({ isShowing: false });
+        }, 2500);
```
supernova-at: I think we need to assign this `setTimeout` to its own class variable (in other words, don't reuse `timeoutId`) because 

* This could lead to the tooltip "blinking" while hovering
* This could lead to a situation where we don't cleanup properly

Both of which occur because this instance of `setTimeout` is not cleared (`window.clearTimeout`).


To reproduce:
1. The tooltip target receives focus. `this.timeoutId` is set to a timer going off in 2.5 seconds.
1. The target loses focus (`onBlur` triggered). `this.timeoutId` is re-assigned and on the next tick of the event loop, the tooltip is hidden.
1. The target gains focus again before the original 2.5 seconds have expired. When that timer expires, the tooltip will change its state to `isShowing: false`, even though we're actively hovering the target and it should be shown.

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1366) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1366) before we can accept your contribution.<br/>

supernova-at:
✅ Verification steps complete

realchriswells:
@dpatil-magento @supernova-at What is the expected behaviour for this?
Do you want the tooltip to show every time it is clicked, or only once?

The issue seems to be when it is already `onFocus`, once the tooltip disappears the element stays `onFocus` and does not trigger it again, which is why it is missing the timeout.

supernova-at:
> @dpatil-magento @supernova-at What is the expected behaviour for this?
Do you want the tooltip to show every time it is clicked, or only once?

Checking with @soumya-ashok , our UX guru 😃

supernova-at:
On hold until we get a response from UX

soumya-ashok:
@realchriswells Would it be possible to change the focus state to this treatment instead of a tooltip?

Screen Shot 2019-06-25 at 9 37 21 AM

If that is outside the scope of this issue, then the tooltip behavior can be as follows for now -
Shopper taps the swatch, tooltip appears on focus, and self destructs in 3-4 seconds. If the shopper changes the color selection, tooltip then appears on the selected color swatch in the same way. For any colors that are unavailable, there is an X through the swatch visually, and the tooltip could say "Unavailable".
Thoughts @supernova-at ?

supernova-at:
> the tooltip behavior can be as follows for now -
Shopper taps the swatch, tooltip appears on focus, and self destructs in 3-4 seconds. If the shopper changes the color selection, tooltip then appears on the selected color swatch in the same way. For any colors that are unavailable, there is an X through the swatch visually, and the tooltip could say "Unavailable". Thoughts @supernova-at ?

I really like the new treatment! As mentioned in stand up, it will be a bonus for our accessibility as well.

I agree with what you've said here about the tooltip but I really do think we should go with the new treatment instead.

@realchriswells does this seem like something you'd want to take on? I'm fine repurposing the issue, just let me know 👍

soumya-ashok:
@supernova-at Awesome, let me know if there's anything else needed from me.

supernova-at:
@realchriswells I'm going to close this issue (and this PR with it). I opened #1390 for the new treatment and I have assigned it for you. If you don't want to pick it up, kindly leave a comment in #1390 saying as much.

Thanks!

realchriswells:
I'll give it a go :)

Issue #1284, opened 2019-05-31
[feature]: When a user hovers over or presses into a toast the timeout should pause


sirugh:
**Is your feature request related to a problem? Please describe.**
When a user hovers on a toast or clicks in a toast (not the dismiss/cta) or presses on it as in a mobile press to copy the text we should **pause** the timeout. Imagine a short-living toast that has some text a user wants to copy so they click into it to copy but it disappears. That would be frustrating!

**Describe the solution you'd like**
[react-toast-notifications](https://github.com/jossmac/react-toast-notifications) has an interesting take, basically they wrap their toasts with a controller that manages the state. We can do something like that but it's probably something that should live in the presentation layer which may add complexity. **Specifically** the removal timeout is created/removed within the action/reducer but pausing a timeout would live in the display, which means we probably are going to need to refactor how timeouts are managed.

**Describe alternatives you've considered**
Current alternative is that a toast with information that a user may want to copy (like an error) should have a long `timeout` value for now.

**Please let us know what packages this feature is in regards to:**
- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

supernova-at:
Desired UX likely includes pausing **all** toast timeouts on hover of any individual toast.

If a user is attempting to copy the text of the third toast in a list of seven, and other toasts below it start disappearing because their timeout expires, that would be frustrating as well.

jflanaganuk:
#mmuk2019

awilcoxa:
Moving to backlog, toasts feature is working as designed. This enhancement adds value but currently not a high priority. Will re-evaluate for future release.

PR #1368, opened 2019-06-18
Feature/1284 user hovering toasts

jflanaganuk:
## Description
Toasts now do not timeout when the user hovers over the container of the toasts, allowing them to copy text if needed etc.
The timeout then resumes if the user moves out of the toast container

## Related Issue
Closes #[1284](https://github.com/magento-research/pwa-studio/issues/1284).

## Verification Steps
1. Run venia storybook and go to existing story "Non-dismissable toasts that timeout"
2. Hover over the toast and see it should remain
3. Hover outside of all the toasts and see it should fade as expected

## How have YOU tested this?
1. Ran venia storybook with existing story "Non-dismissable toasts that timeout"
2. Hovered over a toast and saw it remain
3. Hovered outside of all the toasts and saw it fade as expected

## Checklist:

### DIFF
--- a/packages/peregrine/src/Toasts/useToasts.js
+++ b/packages/peregrine/src/Toasts/useToasts.js
@@ -78,6 +78,7 @@ export const useToasts = () => {
      *
      * @returns {Number} id - the key referencing the toast in the store
      */
+    let allowToastToDie = true;
     const addToast = useCallback(
         toastProps => {
             const {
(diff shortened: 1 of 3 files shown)


packages/peregrine/src/Toasts/useToasts.js
```diff
@@ -116,16 +117,28 @@ export const useToasts = () => {
             const handleAction = () =>
                 onAction ? onAction(() => removeToast(id)) : () => {};
 
-            // A timeout of 0 means no auto-dismiss.
+            const pauseToast = () => {
+                allowToastToDie = false;
```
sirugh: This value will be lost on a re-render. We shouldn't use variables outside of the callback's scope.

packages/peregrine/src/Toasts/useToasts.js
```diff
@@ -169,9 +197,11 @@ export const useToasts = () => {
         () => ({
             addToast,
             dispatch,
-            removeToast
+            removeToast,
+            pauseAllToasts,
```
sirugh: If you are going to add functions to the toast api you need to add to the jsdoc above.

sirugh:
I'm going to close this, it's referenced in the issue. #1284, but whomever picks this up we need to probably design this out before writing any code :)

PR #1381, opened 2019-06-24
bug: fix pagination double fetch

### LINKED ISSUE #1297: [bug]: Duplicate Image request calls on Category page
**Describe the bug**
On Venia Category page every image has 2 image request calls instead of just one.

**Note**
Issue does not occur when Fastly enabled backend is used with `IMAGE_OPTIMIZING_ORIGIN=backend`

**To Reproduce**
1. Navigate to Venia Home page on develop branch https://magento-venia-develop.now.sh/
2. Open dev tools on chrome and click on any Category say - Tops.

**Expected behavior**
Category page should load with all images and on dev tool network tab image requests should have one call each image.

**Actual behavior**
Category page should load with all images and on dev tool network tab image requests have two calls each image.

**Screenshots**
If applicable, add screenshots to help explain your problem.

 - Browser Version [e.g. 22]:
 - Magento Version:
 - PWA Studio Version:
 - NPM version `npm -v`:
 - Node Version `node -v`:

- `venia-concept`

sirugh:
## Description

Pagination had some bugs that caused a double fetch. Specifically the `usePagination` hook initialized with a current page of `0`, which is just wrong -- it should default to 1, if anything. So I updated the hook to accept an `initialPage` param which if omitted will default to 1. Now, when we first load, we check the `page` query param and use it if defined.

I ran into another issue, which is that the apollo client will actually just throw the GQL errors, and we never caught them. So, for example, if you request `page=5` and there are only 4 pages, ApolloClient throws an error and we just break. To fix this I added a try-catch and then some retry logic to the category page.

Additional "bonus" fixes:
* Also fixed default page size in gallery items that was causing a visual discrepancy between the "loading" state and the final loaded category page (12 items vs 6 default).
* Also removed some unnecessary cleanup from pagination component and moved it to the category page.

## Related Issue

Closes #1297.

## Verification Steps
1) Load your local and click on a category. It should only fetch the 6 images for the category once.
2) Click on page 2. It should still only fetch the page 2 images only once.
3) CTRL+R to refresh the page. It should still only fetch the images once.
4) In the url, make the query param `page=65` and press enter. It should update to `page=1` and it should only fetch one set of images.

`usePagination` now defaults to `1` instead of `0` so this is technically a breaking change.

## Checklist:


packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -1,22 +1,43 @@
...
+    // hook. Without this we may load `page=4` but `usePagination` will init to
+    // 1, so it'll fetch page 1 _and_ page 4 (bad!).
+    const initialPage = useMemo(
+        () =>
+            Math.max(
+                1,
+                Math.floor(
+                    ~~getQueryParameterValue({
+                        location,
+                        queryParameter: 'page'
```
sirugh: One thing that stands out to me in all of this is that I don't know who the `page` parameter "belongs" to anymore. We need to read it to get the `initialPage` value for the `usePagination` hook we initialize here in `category.js`, but the only place in which we update the query parameter is within `pagination.js`, a child component.

packages/venia-concept/src/RootComponents/Category/category.js
```diff
@@ -65,9 +85,19 @@ const Category = props => {
 
     useEffect(() => {
         setTotalPages(totalPagesFromData);
+        return () => {
+            setTotalPages(null);
+        };
     }, [totalPagesFromData]);
 
-    if (error) return <div>Data Fetch Error</div>;
+    // Retry 3 times and then display the fallback error.
```
sirugh: 3 is an arbitrary retry count. If ya'll want I can move it to a constant and change it. 🤷‍♂

packages/venia-concept/src/RootComponents/Category/category.js
zetlen: That's a really good observation, related to the question in https://github.com/magento-research/pwa-studio/issues/1305. I think we need to encapsulate the query parameter handling entirely inside `usePagination`; it should use [useSearchParam](https://github.com/magento-research/pwa-studio/blob/develop/packages/peregrine/src/hooks/useSearchParam.js) under the hood.

packages/venia-concept/src/RootComponents/Category/category.js
zetlen: Well, my rule is, "if it shows up three times, abstract it". For this type of slightly nonintuitive logic, I think maybe we abstract it after the _second_ time it crops up. So this is fine for now.

packages/venia-concept/src/components/Gallery/items.js
```diff
@@ -2,7 +2,8 @@ import React, { Component } from 'react';
 import { arrayOf, number, shape } from 'prop-types';
 import GalleryItem from './item';
 
-const pageSize = 12;
+// Default of 6 to match the default category page size.
+const pageSize = 6;
```
zetlen: We can pull this from the `storeConfig` query when https://github.com/magento/graphql-ce/pull/650 is released! Can you add a TODO reflecting that?

packages/peregrine/src/hooks/usePagination.js
```diff
@@ -1,10 +1,29 @@
...
 
+    /**
+     * @typedef PaginationState
+     * @property {Number} currentPage the current page
+     * @property {Number} totalPages the total pages
+     */
     const paginationState = { currentPage, totalPages };
+
+    /**
+     * @typedef PaginationApi
```
zetlen: Thanks for the typedefs :+1:

packages/peregrine/src/hooks/usePagination.js
```diff
@@ -1,14 +1,92 @@
...
+ * TODO update with defaults
+ *
+ * @returns {[PaginationState, PaginationApi]}
+ */
+export const usePagination = ({
+    location = window.location,
+    history = window.history,
+    namespace = '',
+    parameter = 'page',
+    initialPage = 1
```
tjwiebell: If we're going to do a re-factor, I think we should go ahead and support a `totalPages` parameter as well. Right now we assume that pagination data is loaded async, but it would be nice to support components that already know the totalPages, so they don't have to init and immediately setTotalPages.

sirugh: Cool, will do.

packages/venia-concept/src/RootComponents/Category/category.js
sirugh: I made these changes. I'll wait for feedback before fixing the tests.

dpatil-magento:
@sirughLooks like there are unit test failures, can you plesae take a look?  
Summary of all failing tests 
FAIL packages/venia-concept/src/components/Pagination/__tests__/pagination.spec.js

dpatil-magento:
@sirugh When I directly access https://venia-81a6om2eg.now.sh/venia-tops.html?_page=3.  or reload any page other than first then  first page image requests (Total 12 image request instead of 6) are also made.

sirugh:
Dang. I'll take a look.

sirugh:
Ok @dpatil-magento fixed.

dpatil-magento:
@sirugh When we use backend **without** fastly >>  Click on any category image requests are made. Navigate to second page of category and come back to first page using arrow navigator.  Image requests are made again instead of loading from the cache.   This exists on develop branch too. If this is something not related to this fix then I can create a separate ticket.

sirugh:
> 1: When we use backend without fastly >> Click on any category image requests are made. Navigate to second page of category and come back to first page using arrow navigator. Image requests are made again instead of loading from the cache. This exists on develop branch too. If this is something not related to this fix then I can create a separate ticket.

I'm going to need you to show me how you reproduced this. Can we do a call or pair on this later?

> 2: When we access https://venia-mz6wig31s.now.sh/shop-the-look.html?_page=23 directly (when first page images are not loaded previously, clear cache to achieve this). then for a fraction of second UI shows data fetch error and then images are loaded.

I will look into this. I think I know what the problem is.

sirugh:
After working with @dpatil-magento we found that regardless of the backend url we are disregarding cache and requesting images again. To repro:

1) Make sure "Disable Cache" in network dev tools is unchecked.
2) Clear application cache and navigate to home page.
3) Click on category, see images requested for 1st page.
4) Click on second page, see images requested for 2nd page.
5) Click back to first page, images are requested _again_.

Noteably, on `develop`, https://magento-venia-develop.now.sh, following the steps above will result in the images not being requested again on step 5 (as they come from the cache).

dpatil-magento:
@sirugh  QA pass, good to merge if review is approved.

packages/venia-concept/src/components/Pagination/pagination.js
```diff
@@ -117,34 +97,24 @@ class Pagination extends Component {
                 <NavButton
                     name={navButtons.lastPage.name}
                     active={isActiveRight}
-                    onClick={() => this.setPage(rightSkip)}
+                    onClick={() => setPage(rightSkip)}
```
jimbo: Inline callbacks are recreated and reattached on every render. Can we avoid these?

sirugh: Good catch!

sirugh: Gotta be a lint rule somewhere...

packages/peregrine/lib/hooks/usePagination.js
```diff
@@ -43,12 +43,12 @@ export const usePagination = ({
     initialPage,
     initialTotalPages = 1
 } = {}) => {
+    const searchParam = namespace ? `${namespace}_${parameter}` : parameter;
```
sirugh: This only includes  an `_` if namespace is provided. It was incorrectly using a param of `_page`.

PR #1382, opened 2019-06-24
useDOMQuery and usePageTitle custom hooks.

revanth0212:
## Description
This change adds 2 new hooks, 1 for DOM Manipulation and the other that uses the DOM Manipulation hook to change the page title.

## Related Issue
Closes https://github.com/magento-research/pwa-studio/issues/422


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1382) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1382) before we can accept your contribution.<br/><hr/>**Revanth Kumar Annavarapu** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

packages/peregrine/src/hooks/useDOMQuery.js
```diff
@@ -0,0 +1,48 @@
...
+ * to find all elements.
+ *
+ * @param {string} queryString
+ * @return Array<HTMLElement> and {
+ *  setInnerHTML: (string) => void,
+ *  setInnerText: (string) => void,
+ *  setAttribute: (string, any) => void
+ * }
+ */
+export const useDOMQuery = queryString => {
```
supernova-at: Since the term "query string" is very much tied to the URL, can we rename this variable `selector` or `querySelector`?

packages/peregrine/src/hooks/useDOMQuery.js
```diff
@@ -0,0 +1,48 @@
...
+    const [elements, setElements] = useState([]);
+    useEffect(() => {
+        setElements(findDOMElements(queryString));
+    }, []);
+    const setInnerHTML = useCallback(
+        content => {
+            elements.forEach(element => {
+                element.innerHTML = content || element.innerHTML;
```
supernova-at: Do we need this `||` here? Why would anyone pass in falsy `content`?

If we do need it, consider 

```js
if (content) {
 // forEach here
}
```

It will save some time if we have a lot of `elements` - we can just skip the entire iteration since we won't be updating anything anyway.

packages/peregrine/src/hooks/useDOMQuery.js
```diff
@@ -0,0 +1,48 @@
...
+            });
+        },
+        [elements]
+    );
+    const setInnerText = useCallback(
+        content => {
+            elements.forEach(element => {
+                element.innerText = content || element.innerText;
```
supernova-at: Can apply the same optimization as above with this one too.

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
```diff
@@ -0,0 +1,58 @@
...
+                newHTML={newHTML}
+                newName={newName}
+            />,
+            { attachTo: window.domNode }
+        );
+        return new Promise(resolve => {
+            setTimeout(() => {
+                resolve();
+            }, 0);
+        }).then(() => {
```
supernova-at: There has to be a better way to do this 😆

packages/peregrine/src/hooks/useDOMQuery.js
```diff
@@ -0,0 +1,48 @@
...
+ *  setInnerText: (string) => void,
+ *  setAttribute: (string, any) => void
+ * }
+ */
+export const useDOMQuery = queryString => {
+    const [elements, setElements] = useState([]);
+    useEffect(() => {
+        setElements(findDOMElements(queryString));
+    }, []);
+    const setInnerHTML = useCallback(
```
zetlen: What if the caller wants to delete the contents of the element?

packages/peregrine/src/hooks/__mocks__/useDOMQuery.js
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect } from 'react';
+
+import { useDOMQuery } from '../useDOMQuery';
+
+export const TestValidComponent = ({ newText, newHTML, newName }) => {
+    const [, { setInnerText }] = useDOMQuery('#div1');
```
supernova-at: Files in the `__mocks__` folder are really to mock functionality. In other words, if `useDOMQuery` did some expensive operation and you wanted to provide a mock implementation that didn't do that expensive operation for testing purposes, you would put that code in this file.
See https://jestjs.io/docs/en/manual-mocks. 

What you have here are more like `fixtures` or mock data, which we usually put into a `__fixtures__` directory instead.

packages/peregrine/src/hooks/useDOMQuery.js
```diff
@@ -0,0 +1,48 @@
...
+        content => {
+            elements.forEach(element => {
+                element.innerHTML = content || element.innerHTML;
+            });
+        },
+        [elements]
+    );
+    const setInnerText = useCallback(
```
zetlen: You may understand hooks better than me. The collection of elements will likely change identity every time. Why do we need `useCallback` here?

packages/peregrine/src/hooks/__mocks__/useDOMQuery.js
supernova-at: Ideally all the logic for the test resides in the actual `.spec.js` test file, not the fixture. The fixture would just return the JSX you're going to test against:

```jsx
    return (
        <React.Fragment>
            <div id="div1">{'I am Test Div 1'}</div>
            <div id="div2" name="oldName">
                {'I am Test Div 2'}
            </div>
        </React.Fragment>
    );
```

If it's small enough (like this one), you can inline it directly into the `.spec.js` file as well.

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
```diff
@@ -0,0 +1,58 @@
...
+            const div1 = wrapper.find('#div1').getDOMNode();
+            const div2 = wrapper.find('#div2').getDOMNode();
+            expect(div1.innerText).toBe(newText);
+            expect(div2.innerHTML).toBe(newHTML);
+            expect(div2.getAttribute('name')).toBe(newName);
```
supernova-at: These assertions are hard to reason about because the logic is spread between two files.

Each unit test should test one distinct thing. In this case, it would be something like:

```js
// 1. Setup / prepare the test if needed

// 2. Call the function under test
const [, { setInnerText }] = useDOMQuery('#div1');
setInnerText(newText);

// 3. Make your assertions
expect(div1.innerText).toBe(newText);
```

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
supernova-at: I'm assuming this will go away after removing `Enzyme` anyway

packages/peregrine/src/hooks/usePageTitle.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+export const usePageTitle = () => {
+    const [, { setInnerText }] = useDOMQuery('title');
+    const updateTitle = useCallback(
+        newTitle => {
+            setInnerText(newTitle);
+        },
+        [setInnerText]
+    );
+    return updateTitle;
```
supernova-at: In an effort to keep some consistency between hooks, can we have this return an array of the current value and an updater?

Ex: 
```js
const [elements, { setInnerText }] = useDOMQuery('title');
const titleElement = elements[0];

// other stuff

return [titleElement.innerText, updateTitle];
```

packages/peregrine/src/hooks/__tests__/usePageTitle.spec.js
```diff
@@ -0,0 +1,39 @@
...
+
+import { usePageTitle } from '../usePageTitle';
+
+const TestComponent = ({ newTitle }) => {
+    const updateTitle = usePageTitle();
+    useEffect(() => {
+        updateTitle(newTitle);
+    }, [updateTitle]);
+    return <React.Fragment />;
+};
```
supernova-at: This is much better, yes :)

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import {
+    TestValidComponent,
+    TestInvalidComponent
+} from '../__mocks__/useDOMQuery';
+
+describe('Unit testing useDOMQuery hook.', () => {
```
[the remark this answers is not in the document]
supernova-at: > We try to avoid using describe

This is news to me, lol. For some files (like our `asyncActions`) that have a ton of functions that they're testing (and a handful of tests for each function), I've been wrapping each function's tests in a `describe` block. It also helps for having variables that are scoped only to that `describe` function.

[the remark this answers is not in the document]
supernova-at: But for test files that only have a handful of test cases like this, I agree that the `describe` block is overkill.

packages/peregrine/src/hooks/__mocks__/useDOMQuery.js
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect } from 'react';
```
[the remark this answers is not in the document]
revanth0212: Actually my bad. Shouldn't have put them in mocks folder. They are not snapshots/mocks. They are actually the examples that I will be using to test the hooks.

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { mount } from 'enzyme';
```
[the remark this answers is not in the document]
revanth0212: I have tried using `react-test-renderer` but unfortunately, it is not working as enzyme's `mount` which is a full dom render. It is only rendering components and leaving them in the react world but since we are directly manipulating DOM and bypassing react middle layer, the changes are not showing up in the output. For instance, when I did this:

```
act(() => {
  ReactDOM.render(
    // or TestRenderer.create
    <TestValidComponent
      newText={newText}
      newHTML={newHTML}
      newName={newName}
    />,
    container
  );
});
```

The output looked like this:

```
[
  { type: "div", props: { id: "div1" }, children: ["I am Test Div 1"] },
  {
    type: "div",
    props: { id: "div2", name: "oldName" },
    children: ["I am Test Div 2"]
  }
];
```

I was not able to get full dom output from the renderer.

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
```diff
@@ -0,0 +1,58 @@
...
+
+import {
+    TestValidComponent,
+    TestInvalidComponent
+} from '../__mocks__/useDOMQuery';
```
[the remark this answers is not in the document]
revanth0212: My bad. I am not mocking the implementation here. `__mocks__` are just examples or mock data that I am using for testing the hooks.

packages/peregrine/src/hooks/useDOMQuery.js
revanth0212: Collection of elements will only change on the component creation and component mount but component itself might be re-rendered multiple times depending on the parent's implementation. I realized we don't need to recreate all the helper functions on every render. This is based on the assumption that state variables references are memoized internally when we use `useState` and only change when its setter is used. Correct me if I am wrong.

revanth0212: When I said component in the above context, I meant the hook.

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
revanth0212: Unfortunately no 😭

packages/peregrine/src/hooks/useDOMQuery.js
supernova-at: > User can set it to the empty string '' to delete all content.

Careful, in JavaScript the empty string is falsy 😉

packages/peregrine/src/hooks/usePageTitle.js
jimbo: @supernova-at Good suggestion.

packages/peregrine/src/hooks/useDOMQuery.js
```diff
@@ -0,0 +1,48 @@
...
+ *  setInnerHTML: (string) => void,
+ *  setInnerText: (string) => void,
+ *  setAttribute: (string, any) => void
+ * }
+ */
+export const useDOMQuery = querySelector => {
+    const [elements, setElements] = useState([]);
+    useEffect(() => {
+        setElements(findDOMElements(querySelector));
+    }, []);
```
jimbo: The array of dependencies is missing `querySelector` here, so the effect would never run again, even if `querySelector` changes.

packages/peregrine/src/hooks/useDOMQuery.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param {string} querySelector
+ * @return Array<HTMLElement> and {
+ *  setInnerHTML: (string) => void,
+ *  setInnerText: (string) => void,
+ *  setAttribute: (string, any) => void
+ * }
+ */
+export const useDOMQuery = querySelector => {
+    const [elements, setElements] = useState([]);
```
jimbo: Unfortunately, I believe this is a memory leak. Storing references to DOM nodes in state will [prevent them from being collected](https://developers.google.com/web/tools/chrome-devtools/memory-problems/#discover_detached_dom_tree_memory_leaks_with_heap_snapshots), even if React removes those nodes from the DOM.

packages/peregrine/src/hooks/useDOMQuery.js
jimbo: @zetlen is asking the right question here. Re-implementing a subset of the `Element` API as callbacks is a mistake; this is an incomplete abstraction, and to complete it we'd have to add more and more callbacks until we covered 100% of the API, at which point we would have a large amount of code adding no value.

revanth0212: My bad. Fixed it. But you can still send an empty string to clear.

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
revanth0212: Cleaned it up. If in the future we can find a cleaner solution to this, it will be easy to delete unnecessary code.

revanth0212:
> Unfortunately, I think `useDOMQuery` is a step in the wrong direction here, as written.

@zetlen had a bigger picture for this hook. He should be able to explain it much better than me.

> It seems like useDOMQuery was written specifically for usePageTitle.

In this PR it looks like this is only for page title but page title hook acts as an example of how to use the dom query hook as well as a custom hook.

> findDOMElements doesn't add anything over document.querySelectorAll()

Ideally speaking this hook is more of a way to query a DOM element once and perform manipulations/actions on it in the future. The component that uses this hook can pass it down to its children and they don't need to worry about querying or knowing the query string/selector. All they need to know is that an action will be provided to them which they can use to manipulate the DOM. I am totally against the idea of directly manipulating the DOM but this should be used in cases where there is no other option.

Obviously, we will not be able to provide all the actions that can be performed over DOM elements, hence we are exposing the elements array so the user can use it as an escape hatch. But you are right, it not in our best interest to expose the DOM. But at the end of the day, any developer can use the querySelector functionality to get access to the DOM anyway.

packages/peregrine/src/hooks/useDOMQuery.js
revanth0212: That is by design I feel. User should not be using the same hook for different queries. User can create different hooks for different query selectors. Let me know if that design needs to change.

jimbo:
> @zetlen had a bigger picture for this hook. He should be able to explain it much better than me.

Sure. This PR illustrates how one might be able to perform DOM queries with a custom hook, but also why it's not a good idea to do so. It's possible to do it, but better to not do it. We don't *need* this hook.

> Ideally speaking this hook is more of a way to query a DOM element once and perform manipulations/actions on it in the future.

Yes, that's exactly what this hook is, but no, that's not ideal. We don't want to hold references to DOM elements in order to later perform actions on them. And in truth, we don't want to query and manipulate the DOM at all.

> I am totally against the idea of directly manipulating the DOM but this should be used in cases where there is no other option.

Excellent. I do think there are other options, and we should look for one one instead of manipulating the DOM here.

> But you are right, it not in our best interest to expose the DOM. But at the end of the day, any developer can use the querySelector functionality to get access to the DOM anyway.

On the contrary, I do think it's in our best interest to expose the DOM, if that's what we're working with. DOM elements have a [powerful API](https://developer.mozilla.org/en-US/docs/Web/API/Element), so in the extremely rare situations where we operate on them, we shouldn't try to constrain that API to some subset of "most common" methods, even as a convenience.

packages/peregrine/src/hooks/__tests__/useDOMQuery.spec.js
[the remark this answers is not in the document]
zetlen: See, I told you we didn't make those conventions clear enough.

The reason we don't like to use them is that they're harder to find in the codebase using global text search.

```js
describe('the wheels on the bus', () => {
  it('go round and round', () => {
  })
})
```

The tree of describe/describe/it turns into a concatenated test description in the reporter. So this test appears in the log as `the wheels on the bus go round and round`, but if I search for that exact string I come up empty.

And the only thing we seem to get in return is indentation! So that's why. I'm open to counterargument as always.

zetlen:
@jimbo is making good points as always. For posterity, here's what my thought process was:

1. The document head may be rendered by React in the future
2. Long-running references to DOM elements are unsafe already, doubly so within the React root
3. Future integrations may want an "escape hatch" that gives them DOM elements
4. It should re-run the DOM query every time it is called in `useEffect`, so the element refs are always fresh
5. To discourage keeping stale elements around, we shouldn't directly expose them but expose a limited set of getters and setters

My thought process might not hold up under scrutiny. Plus, in an outside channel you just linked to [this article about `react-head`](https://jeremygayed.com/making-head-tag-management-thread-safe-with-react-head-323654170b45), which makes a strong case for using that library--it's designed with modern React in line, unlike Helmet, and it works with SSR.

@revanth0212 I think this is still wise to implement as a hook, but I think we should see how it feels using `react-head` and exposing it through a hook API if possible/necessary.

Sorry I led you astray on this contribution!

jimbo:
@revanth0212 I'm going to close this PR. If people want to query the DOM, they should do so with standard web APIs, not a hook.

As for the goal here—writing to the document head—I recommend creating a spike that uses `react-head`, which renders a portal and uses a context provider to allow elements to update it. Each page should be able to use that library's components to affect the document title and metadata as necessary.

PR #1387, opened 2019-06-25
MiniCart Height Fix

### LINKED ISSUE #1071: [BUG] Cropped shopping cart in Venia for larger phones

supernova-at:
## Description

This PR updates the measurement the MiniCart uses to compute its `height` to use `window.innerHeight` instead of `100vh` as described by https://css-tricks.com/the-trick-to-viewport-units-on-mobile/. 

As described in the attached issue #1071, `100vh` often doesn't account for browser address bars or other frames.

This PR takes advantage of the new [useWindowSize](https://github.com/magento-research/pwa-studio/blob/develop/packages/peregrine/src/hooks/useWindowSize.js) hook to update this value even after resize events occur.

## Related Issue

Closes #1071 .

## Verification Steps

1. Add an item to your cart
1. Open the MiniCart (right drawer)
1. See that the Checkout button is fully visible
1. Resize the window (or change orientation of device) and ensure Checkout button is fully visible
1. The payments form depended a bit on these CSS variables, so ensure that looks good too

Run the verification steps above on a variety of phones.

## How Have YOU Tested this?

1. Tested using iPhone XR on iOS simulator
1. `yarn test`.

### Before Fix
image

### After Fix
Screen Shot 2019-05-30 at 3 35 28 PM

## Checklist:


packages/venia-concept/src/components/MiniCart/miniCart.css
```diff
@@ -1,7 +1,9 @@
 .root {
     --base-z-index: 4;
-    --minicart-height: 100vh;
```
sirugh: Having trouble finding our supported browser matrix but for browsers that cannot support custom props we should still have `height: 100vh`. See [the example](https://css-tricks.com/the-trick-to-viewport-units-on-mobile/#article-header-id-1).

supernova-at:
> it would be more ideal and much faster for CSS to handle the sizing on its own.

Totally agree.

> Is there literally any other way to set a fixed element to be full height that does respect browser chrome? I can be the person to experiment with that, if necessary.

Based on the [CSS Tricks](https://css-tricks.com/the-trick-to-viewport-units-on-mobile/) article the problem is that the mobile browsers have calculated `100vh` to be based on the maximum height of the screen (not including any browser chrome). If we can find a way to `calc` that value minus the browser chrome height or something, that would definitely be ideal.

I'm not aware of any property like that in CSS though, so I went with JS `window.innerHeight` instead.

packages/venia-concept/src/components/MiniCart/miniCart.css
supernova-at: Resolved in https://github.com/magento-research/pwa-studio/pull/1387/commits/fd9fa4d5173a2660786c410dcebab0f4ce27f095.

supernova-at:
> My only concern is that we are using resize events to update a css property which means repaint. This may cause a resize to stutter. Perhaps this is a good time to try to debounce so that we update the size on the last value after of a bunch of resize events.

Great idea but I don't actually know where we'd do that.

Here are the relevant bits:

```jsx
    const { innerHeight: viewportHeight } = useWindowSize();
    const rootStyle = {
        '--minicart-height-unit': `${viewportHeight * 0.01}px`
    };

    // then...
    <aside className={rootClass} style={rootStyle}>
```

I don't think we can wrap `debounce` around `useWindowSize` because of the "consistent number of hook calls" restriction, and I thought we decided against adding it inside of `useWindowSize` itself.

Where were you thinking?

sirugh:
> I don't think we can wrap debounce around useWindowSize because of the "consistent number of hook calls" restriction, and I thought we decided against adding it inside of useWindowSize itself.

We can't debounce the hook but we can debounce the update of `rootStyle`:

```js
const { innerHeight } = useWindowSize();
const [debouncedHeight, setDebouncedHeight] = useState(window.innerHeight);

useEffect(() => {
  _.debounce(() => {
    setDebouncedHeight(innerHeight)
  }, 50);
}, [innerHeight]);
```

const rootStyle = {
  '--minicart-height-unit: `${debouncedHeight * 0.01}px`
}
return (
  <aside className={rootClass} style={rootStyle}>
    //...
  </aside>
)
```

That's just rough code but the idea applies I think.

**Comment by sirugh** [comment]
> I'm not aware of any property like that in CSS though, so I went with JS window.innerHeight instead.

`fill-available` may do the trick but it is experimental.

**Comment by supernova-at** [comment]
PR Updated:
  * Debounces repaint operations (https://github.com/magento-research/pwa-studio/pull/1387/commits/2e6719623d081596fd5e55979400565d48b093d2).

This one was interesting! Turns out timing events (used by `lodash.debounce`) have unintuitive interactions with React Hooks.

Here's some great reading: https://overreacted.io/making-setinterval-declarative-with-react-hooks/.

The code is therefore not quite as straightforward as we'd like but I added some comments to mitigate confusion.

It works great!

File: packages/peregrine/src/hooks/useRestApi.js [hunk 2]
```diff
@@ -32,7 +32,7 @@ export const useRestApi = endpoint => {
                 receiveError(error.baseMessage);
             }
         },
-        [receiveError, receiveResponse, setLoading]
+        [endpoint, receiveError, receiveResponse, setLoading]
```
supernova-at: This change has nothing to do with this PR but it failed linting and I couldn't push without it.

packages/venia-concept/src/components/MiniCart/miniCart.js
```diff
@@ -26,12 +29,35 @@ const MiniCart = props => {
...
+    innerHeightMutableRef.current = innerHeight;
+
+    const setViewportHeightDebounced = debounce(() => {
+        // We're really just doing setViewportHeight(innerHeight) here
+        // but because we're within a debounced function, innerHeight
+        // gets captured (never updates).
+        // So we use this ref instead to ensure we always have the latest value.
+        setViewportHeight(innerHeightMutableRef.current);
+    }, 1000);
+    const setViewportHeightDebouncedRef = useRef(setViewportHeightDebounced);
```
supernova-at: I didn't add a comment in the code here but just FYI we also had to make the debounced function a ref because otherwise a new debounced function would get created each time through this function and they would ALL fire after the debounce delay.

packages/peregrine/src/hooks/useRestApi.js
sirugh: *cough* `--no-verify` *cough* I mean what? Good job thanks!

packages/venia-concept/src/components/MiniCart/miniCart.js
sirugh: Can we not just `useMemo` here?

supernova-at: Good question - I had that thought too but for some reason tried `useRef` first. Since it worked I didn't try `useMemo`.

If you think it's better I can switch it.

jimbo:
@supernova-at That's a great article from Dan. I would have expected us to debounce things inside `useWindowSize`, though, since that's where the changes are published. Updating the context provider on every resize event is probably too noisy, so if we debounce there, we won't have to do a bunch of ref work in all of the consumers.

Also, I think `height: 100%` will work, no JS sizing necessary. Try it out in XCode's simulator, it works.

packages/venia-concept/src/components/MiniCart/miniCart.js
sirugh: Isn't this also going to multiple references to `setViewportHeightDebounced` and only invoke one? As in whichever is attached to the `ref` last? It just feels overly confusing currently but I can see now how my original example would have queued a bunch of `useEffect` callbacks too...

sirugh:
> I would have expected us to debounce things inside useWindowSize

Didn't we talk about this and decide that we didn't want to enforce this? I don't know how much time each additional use of `useWIndowSize` takes but I feel like it is the downstream use of the values as in this PR that may have a bigger impact on performance.

jimbo:
> Didn't we talk about this and decide that we didn't want to enforce this?

I just looked through #1193 and didn't see any discussion of it. The context provider publishes an updated value on each resize event, which causes every consumer to determine whether to re-render. To optimize this, we'd want to debounce in the provider so we were doing less of that determining.

(In other words, subscribers don't control the "stream" of updates. Context updates are distributed via push, not pull, and consumers can't conditionally call `useContext()`.)

The only reason not to debounce in the provider is to protect the edge case in which a hypothetical consumer **legitimately needs** maximum granularity—literally every resize event. I don't anticipate seeing that edge case in our application.

sirugh:
@jimbo @supernova-at using `height: 100%` seems to "fix" the problem but it does add a weird UX when scrolling, which our current scroll lock solution does not account for.

[Image from Gyazo](https://gyazo.com/c2a1cfb73e8b269e5e7008897b17aecd)

It also breaks the `paymentsForm` in iOS, which for some reason is nested within the cart footer:

[Image from Gyazo](https://gyazo.com/a0a2f0c44004bee1a462baa7b4b36dde)

supernova-at:
> It also breaks the paymentsForm in iOS, which for some reason is nested within the cart footer

Yes, this is where `height: 100%` breaks. I shied away from refactoring the payments form placement out of scope concerns but if we think that'd be an easier fix than all this, I can try that instead?

dpatil-magento:
Addressed as part of #1449

PR #1392, opened 2019-06-26
Merge master to develop

jcalcaben:
## Description

Merges the 3.0 release from `master` into the `develop` branch. The only significant merge conflicts found were in `package.json` files and they were resolved by accepting the higher version numbers in each change set.

## Related Issue

Closes #1389

## Verification Steps

1. Run regression testing
2. Verify no errors

## Checklist:

- [ ] I have updated the documentation accordingly, if necessary.
- [ ] I have added tests to cover my changes, if necessary.

PR #1393, opened 2019-06-26
Document useQueryResult

jimbo:


jcalcaben:
This may conflict with the incoming changes in #1392

sirugh:
Calcaben already wrote docs for it but against the release branch. Seems like you guys doubled up on accident. We should merge the release branch in to develop and then see if there's any gaps you can fill with docs in this PR.

jimbo:
@jcalcaben Wow, this won't be necessary then! Thanks for taking care of this already.

PR #1395, opened 2019-06-27
bug Add name for search trigger button

### LINKED ISSUE #1386: [bug]: SearchTrigger button does not have an accessible name
**Describe the bug**
The search trigger in the header contains a `<button>` element which violates accessibility guidelines by not having an accessible name.

**To Reproduce**
1. Run Lighthouse on https://magento-venia-develop.now.sh
2. Scroll down to the Accessibility section.
3. Expand the "Buttons do not have an accessible name" warning.

**Expected behavior**
All elements required to have an accessible name (that's all form elements) have one.

**Screenshots**
image

**Additional context**
This _alone_ is docking us 15 points from the Accessibility score in Lighthouse. And it's an easy fix!

**Possible solutions**
Add an `aria-label` field to the `<button>` element in the `SearchTrigger` component. It's that easy!

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

khoa-le:
## Description

Add name property for search button on searchTrigger component  to increase Accessibility score in Lighthouse.

## Related Issue

Closes #1386

## Verification Steps

1. Run Lighthouse on https://magento-venia-develop.now.sh
2. Scroll down to the Accessibility section.
3. Don't see any warning about like ""Buttons do not have an accessible name""

## Checklist:
Verification steps above!

### DIFF
--- a/packages/venia-concept/src/components/Header/searchTrigger.js
+++ b/packages/venia-concept/src/components/Header/searchTrigger.js
@@ -21,7 +21,7 @@ class SearchTrigger extends Component {
 
         return (
             <Fragment>
-                <button className={searchClass} onClick={toggleSearch}>
+                <button className={searchClass} name="button" onClick={toggleSearch}>
                     {children}
                 </button>
             </Fragment>
--- a/packages/venia-concept/src/components/Header/searchTrigger.js
+++ b/packages/venia-concept/src/components/Header/searchTrigger.js
@@ -23,7 +23,7 @@ class SearchTrigger extends Component {
             <Fragment>
                 <button
                     className={searchClass}
-                    name="button"
+                    role="button"
                     onClick={toggleSearch}
                 >
                     {children}
--- a/packages/venia-concept/src/components/Header/searchTrigger.js
+++ b/packages/venia-concept/src/components/Header/searchTrigger.js
@@ -21,7 +21,11 @@ class SearchTrigger extends Component {
 
         return (
             <Fragment>
-                <button className={searchClass} name="button" onClick={toggleSearch}>
+                <button
+                    className={searchClass}
+                    name="button"
+                    onClick={toggleSearch}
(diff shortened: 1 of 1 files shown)


khoa-le:
I'm waiting to clarify on issue #1386. The issue still happen, please keep it.

packages/venia-concept/src/components/Header/searchTrigger.js
```diff
@@ -21,7 +21,11 @@ class SearchTrigger extends Component {
 
         return (
             <Fragment>
-                <button className={searchClass} onClick={toggleSearch}>
+                <button
+                    className={searchClass}
+                    name="button"
```
sirugh:
```suggestion
                    aria-label={'Search Button'}
```

khoa-le: Thank you! I changed.

brendanfalkowski: I could be wrong about this but as I understand it:
`arial-label={'Search Button'}`

Should probably be:
`aria-label={'Search'}`

When an accessibility aid encounters `aria-label` it announces the label AND the element type for clarity. Including a description of the element in the label is an anti-pattern because it clutters the announcement with duplicate info.
> search button, button

The rewrite would be announced as intended:
> search, button

Aside: I'd love to be proven wrong if I've misunderstood this, but this article makes a clear case for not qualifying the element as part of the label:
https://knowbility.org/blog/2019/ARIA-labels/

zetlen: You're absolutely right, Brendan. I'm amending the commit to leave @khoa-le as author but slightly modify the wording.

PR #1404, opened 2019-06-28
feat: remove esm mid-compilation and add babel preset

### LINKED ISSUE #1403: [feature]: Shareable Babel "preset" package to simplify build config
Peregrine and Venia modules should be easy to consume in a third-party app, like any PWA created with the future scaffolding CLI. But in order to consume it, you need to be able to parse and transpile it.

Right now, we have no guarantees about transpiler configuration outside of this repository itself. So we have to keep Peregrine code, Venia code, and any other code that Webpack may bundle, aligned with the features supported in the root `babel.config.js`...and people consuming our code don't even have access to that.

### Remove our workaround

We came up with an intermediate solution for portability in #788: we added an intermediate compilation step which transpiles Peregrine code into slightly more "generic" ES Modules. This comes at a cost though:

- Slower build times
- More complex scripting
- Serious bugs with hot reloading
- No clear definition of the intermediate compilation target

And even this "generic" transpile is still hard to anticipate for third party consumers.

### Use a Babel Preset

The right way to do this is a [Babel preset](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/user-handbook.md#making-your-own-preset). Instead of maintaining complex Babel configuration in our root repository, we should maintain a separate package that other projects can consume and use for its base Babel configuration. Any project which builds Peregrine code would simply have to include the `@magento/babel-preset-peregrine` module in their devDependencies.

### Decide what really needs Babel transpilation

"Build" is a broad category. Lots of our code will require build steps, and some of them are as simple as copying files from one directory to another. Those operations are usually very fast, so this issue is focused on *code transpilation via Babel* and where/when it should take place.

We use Babel to transpile source code so it will run on a target platform, like Node or a particular browser. Babel is used in the following packages:

- `venia-concept`
  - Package scripts run `webpack`, Webpack `babel-loader` transpiles for browser compatibility
- `peregrine`
  - Package scripts run `babel` CLI to build CommonJS modules for third parties who aren't using ESM
  - Package scripts run `babel` CLI to build "generic" ES modules for `venia-concept` to consume via `@magento/peregrine/esm/*` imports
- `pwa-buildpack`
  - Package scripts run `babel` CLI to transpile for Node 8-10 compatibility

None of the other packages use Babel at all.  (Our test suite transpiles _all code_ on the fly, whether it needs it or not, because of the presence of a root `babel.config.js`. This is maintenance-free, though it probably slows down our test runs unnecessarily.)

**Only Venia really _needs_ a build step.** Peregrine code will transpile as a result of being included in modules processed by Webpack. It just needs to use the same set of Babel features. A Babel preset will help us to keep track of that. _(We happen to have a build step for Buildpack, whose source code uses a few Babel features. I recommend that we get rid of this eventually.)_

### Formalize a directory structure for library code, source code, and build artifacts

Most of the time, third-party projects can consume our source modules directly:

```js
import useQuery from `@magento/peregrine/src/hooks/useQuery`
```

If we keep all of our modules re-exported in `index.js` files, then third-party projects can use the cleaner API

```js
import { useQuery } from `@magento/peregrine`;
```

This is nice, because it means the directory structure of the published module is not part of the public API. _However, it risks including extra dead code in the bundle unless tree-shaking is set up perfectly._ For that reason alone, I think we should continue to expose our directory structure as public API for import.

But this means specifying a full path to the imported module, including its base directory. Since Peregrine was set up with its own build step, the base directory is `src`: therefore you have to import `@magento/peregrine/src/hooks/useQuery` instead of `@magento/peregrine/hooks/useQuery`. This is a minor thing, but it needs to be consistent.

I propose the following rules:

- **All code that expects to be compatible with PWA Studio code should use a new Babel preset `@magento/babel-preset-peregrine`
- **Design all packages to require as little individual build as possible**
- **Use Babel only when compiling the final app Venia**
- **Any code meant to be imported, not directly transpiled, should be in a `/lib` directory relative to package root. This indicates that the files are library code that will be included in a final bundle.**
- **Any code meant to be directly transpiled, like the source code of Venia which Venia runs Webpack against, should be in a `/src` directory relative to package root.
- **Build artifacts should go to a `/dist` directory relative to package root.

If we create and maintain a Babel preset, and then stick to this convention, we will have faster and simpler repository management and easier-to-maintain APIs.

- [x] `venia-concept`
- [x] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

zetlen:
## Description

- Add a package defining [Babel preset](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/user-handbook.md#making-your-own-preset) for all PWA Studio frontend code to use. (#1403) Move all Babel config into that preset and use the preset in project packages.
- Remove intermediate `esm` compilations, which are no longer necessary.
- Remove all compilation from `peregrine` package, which is only for use in compiled apps anyway and should lazily recompile as a dependency.
- Since Peregrine no longer has a direct compilation target, rename its `src` directory to `lib`, per the rules laid out in #1403.

## Related Issue
Closes #1403.

## Verification Steps

1. Clone the repository.
2. Run an install and a full build.
3. Observe that it's faster!
4. Check unit test results for preset package.

This is totally for 4.0.0.

## Checklist:


babel.config.js
```diff
@@ -1,96 +1,11 @@
...
-    ['@babel/plugin-transform-react-jsx'],
-    ['babel-plugin-graphql-tag']
-];
-
-const targets = {
-    dev: 'last 2 Chrome versions',
-    prod: browsers,
-    test: 'node 10'
+module.exports = api => {
+    // const env = api.env() || 'development';
```
sirugh:
```suggestion
```

babel.config.js
```diff
@@ -1,96 +1,11 @@
...
+    const config = { presets: ['@magento/peregrine'] };
+    if (process.env.CODEBUILD_BUILD_ID) {
```
sirugh: What does this signify?

packages/babel-preset-peregrine/lib/index.js
```diff
@@ -0,0 +1,73 @@
...
+        {},
+        {
+            // For maximum recompile speed:
+            dev: 'last 2 Chrome versions',
+            // A consuming package can optionally provide a list of browsers;
+            // otherwise, use the defaults.
+            prod: require('./browserslist'),
+            // The Jest test runner provides a synthetic DOM, but  not a real
+            // browser environment; instead, it's Node.
+            test: 'node 8'
```
sirugh: Why the switch from `node 10`?

packages/babel-preset-peregrine/lib/index.js
```diff
@@ -0,0 +1,73 @@
...
+        },
+        opts.targets
+    );
+
+    const envConfigs = {
+        development: {
+            plugins,
```
sirugh: We may need to add a rule to ignore node_modules somewhere in this chain. Not sure if that belongs here inside the config or in `babel.config.js` above.
```
Error:  Error: Reentrant preset detected trying to load "/codebuild/output/src213/src/github.com/magento-research/pwa-studio/node_modules/@magento/babel-preset-peregrine/index.js". This module is not ignored and is trying to load itself while compiling itself, leading to a dependency cycle. We recommend adding it to your "ignore" list in your babelrc, or to a .babelignore.
```

packages/babel-preset-peregrine/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@magento/babel-preset-peregrine",
+  "version": "2.1.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "description": "Babel transpile configuration common to progressive web projects build with Peregrine",
```
sirugh:
```suggestion
  "description": "Babel configuration common to progressive web projects built with Peregrine",
```

packages/venia-concept/webpack.config.js
```diff
@@ -72,7 +72,7 @@ module.exports = async function(env = {}) {
                     ]
                 },
                 {
-                    include: [themePaths.src, /peregrine\/src\//],
+                    include: [themePaths.src, /peregrine/],
```
sirugh: We don't need to run all of peregrine through babel, right?

```suggestion
                    include: [themePaths.src, /peregrine\/lib\//],
```

pwa-devdocs/scripts/create-reference-docs/config/peregrine/index.js
```diff
@@ -1,10 +1,10 @@
 module.exports = [
     {
-        target: 'peregrine/src/Price/Price.js',
+        target: 'peregrine/lib/Price/price.js'
         type: 'class'
     },
     {
-        target: 'peregrine/src/List/list.js',
+        target: 'peregrine/lib/List/list.js',
```
sirugh: There are other lines in here that reference `peregrine/src`. Please update :)

pwa-devdocs/src/peregrine/reference/list/index.md
```diff
@@ -2,11 +2,11 @@
 title: List
 ---
 
-<!-- 
+<!--
 The reference doc content is generated automatically from the source code.
 To update this section, update the doc blocks in the source code
 -->
-{% include auto-generated/peregrine/src/List/list.md %}
+{% include auto-generated/peregrine/lib/List/list.md %}
```
sirugh: I also think you missed some `hooks` doc refs in other files. Just do a search `peregrine/src`.

pwa-devdocs/scripts/create-reference-docs/config/peregrine/index.js
```diff
@@ -1,10 +1,10 @@
 module.exports = [
     {
-        target: 'peregrine/src/Price/Price.js',
+        target: 'peregrine/lib/Price/price.js'
```
jcalcaben:
```suggestion
        target: 'peregrine/lib/Price/price.js',
```
add comma back in

sirugh:
Also please resolve conflicts :D

babel.config.js
zetlen: That env var is present in AWS CodeBuild. I'm trying to make sure we don't leave stale Babel cache in the build, which has happened before.

packages/babel-preset-peregrine/lib/index.js
zetlen: Now.sh lambda only supports Node 8 still.

packages/babel-preset-peregrine/lib/index.js
zetlen: Working on this now!

packages/venia-concept/webpack.config.js
zetlen: That's a filter expression for files that get required during Webpack compilation. It's not a command to include that whole directory, it's just to say "if we load a module whose path matches this regex, then run it through this rule chain".

zetlen:
@sirugh @jcalcaben Ready for re-review!

babel.config.js
```diff
@@ -1,13 +1,11 @@
 module.exports = api => {
-    const config = { presets: ['@magento/peregrine'] };
-    if (process.env.CODEBUILD_BUILD_ID) {
-        // In AWS CodeBuild CI, don't cache
-        api.cache.never();
-    } else if (api.env() === 'development') {
+    const config = {
+        presets: ['@magento/peregrine'],
+        exclude: [/packages\/babel\-preset\-peregrine\//]
```
sirugh: This is odd -- do we do this because we co-locate the preset with other packages that use the preset? Why not just directly reference the preset by `presets: ['packages/babel-preset-peregrine']` and just have the `create-pwa-app` command use the `@magento/peregrine` tag?

packages/peregrine/lib/Price/__docs__/price.md
```diff
@@ -7,7 +7,7 @@ Formatting of prices and currency symbol selection is handled entirely by the EC
 ## Usage
 
 ```jsx
-import Price from '@peregrine/Price';
+import Price from '@magento/peregrine/lib/components/Price';
```
sirugh: Should we encourage named exports so that people can do `import { Price } from '@magento/peregrine'`

packages/babel-preset-peregrine/lib/index.js
sirugh: Did you resolve this with your recent commits?

packages/babel-preset-peregrine/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@magento/babel-preset-peregrine",
+  "version": "2.1.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "description": "Babel configuration common to progressive web projects build with Peregrine",
+  "main": "index.js",
+  "scripts": {
+    "clean": " "
```
sirugh: We can remove this right?

package.json
```diff
@@ -67,21 +63,19 @@
...
-    "stream-snitch": "~0.0.3",
+    "rimraf": "~2.6.3",
     "wait-for-expect": "~1.1.0"
   },
   "optionalDependencies": {
     "bundlesize": "~0.17.1",
     "sharp": "~0.22.1"
   },
   "engines": {
-    "node": "8.10.0 || >=10.7.0",
```
sirugh: Is node 8 no longer necessary for now.sh?

packages/peregrine/package.json
```diff
@@ -5,21 +5,11 @@
...
   "scripts": {
-    "build": "concurrently --raw yarn:build:cjs yarn:build:esm",
-    "build:ci": "yarn run -s build:esm",
-    "build:cjs": "BABEL_ENV=production babel src --out-dir dist --root-mode 'upward' --source-maps",
-    "build:dev": "yarn run -s build",
-    "build:esm": "BABEL_ENV=development babel src --out-dir esm --root-mode 'upward' --source-maps",
-    "clean": "rimraf dist esm",
-    "prebuild": "yarn run -s clean",
-    "prepublishOnly": "yarn run build",
+    "clean": " ",
```
sirugh: Can remove?

packages/pwa-buildpack/bin/buildpack
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 require('yargs')
-    .commandDir('../dist/cli')
+    .commandDir('../lib/cli')
```
sirugh: Looks like this will resolve #1425 

@jcalcaben we won't need to add `yarn build` to the instructions.
cc @ericerway

package.json
zetlen: It is not! :tada:

packages/peregrine/package.json
zetlen: Unfortunately a limitation of `yarn workspaces run` is that there is no flag to make it ignore workspaces lacking the command. Instead, it errors on all packages where the script is not present. So for any command we want to run on every workspace, we need to have dummy scripts implemented.

packages/peregrine/lib/Price/__docs__/price.md
zetlen: Eventually. That relies on Webpack tree shaking, which as of right now is less reliable than just direct path imports.

babel.config.js
zetlen: @sirugh I hadn't thought of that! If you wanna try it, please be my guest. Otherwise, we can call that a later refactor. It'll simplify the build and maybe even speed it up.

sirugh: We can do it later :P

packages/peregrine/lib/Price/__docs__/price.md
jimbo: We literally just upgraded `react-feather` (in #1416) to a version that allows us to import from the root module rather than digging into its directory structure, and I was glad to merge it. Couldn't we do the same thing?

zetlen:
@jimbo Yes, all those things are preserved! Like I mentioned to @sirugh I recommend that we maintain the direct-path imports as public API, which means we shouldn't mess around with directory structure too much in the future. But _we still must support named imports and tree-shaking_, and this PR doesn't interfere with that at all.

PR #1430, opened 2019-07-10
Refactor create-account route to use appShell to get provided requisite properties

### LINKED ISSUE #1429: [bug]: create-account link doesn't work
**Describe the bug**
Navigating to the `create-account` page is broken.

**To Reproduce**
1. Go to https://magento-venia-develop.now.sh/create-account

**Expected behavior**
The create account page should open

**Screenshots**
[Image from Gyazo](https://gyazo.com/d19f1f4cc94e94b83c6c69f8d4d842e4)

**Additional context**

**Possible solutions**

 - Browser Version [e.g. 22]:
 - Magento Version:
 - PWA Studio Version:
 - NPM version `npm -v`:
 - Node Version `node -v`:

- [ ] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
## Description

The create-account page (and probably other pages) was broken. Upward used `index` when serving `/create-account`. `index` uses generic-shell.mst, which uses default-initial-contents.st, which references `assets.svg.logo`. This means that anything using default-initial-contents.mst must also include the asset manifest.

## Related Issue

Closes #1429.

## Verification Steps

1. Run this PR locally and go to `<url>/create-account`. It should show the create account page.

[Image from Gyazo](https://gyazo.com/7ef3f450a48660adaa2f42780a7fe762)

## Checklist:

### DIFF
--- a/packages/venia-concept/src/util/makeUrl.js
+++ b/packages/venia-concept/src/util/makeUrl.js
@@ -1,8 +1,9 @@
 // If the root template supplies the backend URL at runtime, use it directly
-const {
-    imageOptimizingOrigin,
-    mediaBackend = 'https://backend.test/media/'
-} = document.querySelector('html').dataset;
+const htmlDataset = document.querySelector('html').dataset;
+const { imageOptimizingOrigin } = htmlDataset;
+// Protect against potential falsy values for `mediaBackend`.
+const mediaBackend = htmlDataset.mediaBackend || 'https://backend.test/media/';
+
 const useBackendForImgs = imageOptimizingOrigin === 'backend';
 
 // Tests if a URL begins with `http:` or `https:` or `data:`
--- a/packages/venia-concept/src/util/makeUrl.js
+++ b/packages/venia-concept/src/util/makeUrl.js
@@ -2,7 +2,11 @@
 const htmlDataset = document.querySelector('html').dataset;
 const { imageOptimizingOrigin } = htmlDataset;
 // Protect against potential falsy values for `mediaBackend`.
-const mediaBackend = htmlDataset.mediaBackend || 'https://backend.test/media/';
+let mediaBackend = htmlDataset.mediaBackend;
+if (!mediaBackend) {
+    console.warn('A media backend URL should be defined in your config.');
+    mediaBackend = 'https://backend.test/media/';
+}
 
 const useBackendForImgs = imageOptimizingOrigin === 'backend';
 
--- a/packages/venia-concept/upward.yml
+++ b/packages/venia-concept/upward.yml
@@ -356,6 +356,7 @@ index:
       engine: mustache
       template: './templates/generic-shell.mst'
       provide:
+        assets: assetManifest
         isDevelopment: isDevelopment
 
 # The static object is a DirectoryResolver that allows access to the files
--- a/packages/venia-concept/upward.yml
+++ b/packages/venia-concept/upward.yml
@@ -33,11 +33,10 @@ response:
     - matches: request.url.pathname
       pattern: '^/(graphql|rest)(/|$)'
       use: proxy
-    # Requests to create an account are handled by the top-level 'index'
-    # object, which returns a generic app shell
+    # Requests to create an account are handled by the 'appShell' object.
     - matches: request.url.pathname
       pattern: '^/create-account'
-      use: index
+      use: appShell
     # Requests for static resources provided by the application project, such
     # as icons, manifests, and static images, are handled by the top-level
     # 'static' object, which defines a DirectoryResolver that serves assets
@@ -62,18 +61,6 @@ response:
   # request is the top-level 'bundles' object, which is a DirectoryResolver that
   # serves the Webpack output assets from the './dist' folder.
   default: bundles
-
-# An object that returns true if the NODE_ENV environment variable is set
-# to 'development'. Used as a test in other configurations in this file.
-# Returns false otherwise.
-isDevelopment:
-  when:
-    - matches: env.NODE_ENV
-      pattern: 'development'
-      use:
-        inline: true
-  default:
-    inline: false
 
 # A ProxyResolver object that passes a request to the backend Magento
 # server defined in the MAGENTO_BACKEND_URL environment variable.
@@ -198,6 +185,15 @@ resource:
         inline:
           entityTypeName:
             inline: "Search Results"
+          template: './templates/generic-shell.mst'
+    - matches: request.url.pathname
+      pattern: '/create-account'
+      use:
+        inline:
+          model: createAccountModel
+          name: createAccountModel.title
+          entityTypeName:
+            inline: "Create Account"
           template: './templates/generic-shell.mst'
   default:
     # If none of the resolver conditions pass, no 'model' property is set and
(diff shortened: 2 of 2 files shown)


sirugh:
@dpatil-magento not sure why this is failing to build -- I ran the failing command locally and it worked. :(

dpatil-magento:
@sirugh Looks like the backing instance is down. Checking this now - https://venia-cicd-lrov2hi-mfwmkrjfqvbjk.us-4.magentosite.cloud/

dpatil-magento:
@sirugh After creating account on https://venia-qbl9lrsc0.now.sh/create-account, I see js error.

image

sirugh:
I'll look into the bug @dpatil-magento.

Also it looks like we have two paths to `/create-account`, one where the user is on the sign-in page and clicks the create account button, and the other where the user navigates directly to the `/create-account` url. Now that I realize this, I'm not sure what the intended UX was. Do we even want a `/create-account` page in venia or do we just want users to create via the sidebar/login page? Adding needs-ux label.

dpatil-magento:
@sirugh the /create-account page already exists when guest user place an order and then click on Create Account button.

image

soumya-ashok:
@dpatil-magento and @sirugh 

First off, I see this as being such an edge case where someone takes the trouble to save the create account URL and then tries to visit it via the URL especially on shopping sites where they really don't want to create an account in the first place unless there's a compelling reason.

However, if we do want to cover this scenario, we could have the following cases for the sign-in/create account experience

If the user has created an account and is signed in, the footer within the main nav. shows that, like in this mockup - https://magento.invisionapp.com/share/KMT00QQH79C
In this case, unless the user signs out of their account, they would not see the sign-in/create account page. If they do try to visit the create account page via URL while signed in, we should probably allow them to create an account with a different email, but if they try to enter the same email as the one they're currently using, then just show the error we already have for that use-case.

If the user has an account but is signed out and visits the create account page, then they should be able to create an account with an email other than the one they're using, and if they try to use the existing email, display the appropriate error.

Does that answer the question?

sirugh:
@soumya-ashok as @dpatil-magento showed there are two routes to create an account.

1) Guest makes purchase and clicks `Create Account` and is redirected to `/create-account`. This opens the form as the whole page.

2) User clicks menu hamburger and `Sign In` and then `Create Account`. This opens the form on the side menu.

I don't think we're talking about a situation where a user is navigating directly to `/create-account` via an external link or anything.

soumya-ashok:
@sirugh I see. One of the comments above said "navigates directly to the create account URL", so I thought that meant they were going to it outside of the path provided.
Is there a specific reason why the two routes @dpatil-magento specified are different? I had assumed that clicking create account would lead to the form in the side menu.

packages/venia-concept/upward.yml
```diff
@@ -63,18 +62,6 @@ response:
   # serves the Webpack output assets from the './dist' folder.
   default: bundles
 
-# An object that returns true if the NODE_ENV environment variable is set
-# to 'development'. Used as a test in other configurations in this file.
-# Returns false otherwise.
-isDevelopment:
```
sirugh: Unused properties, AFAICT.

packages/venia-concept/upward.yml
```diff
@@ -33,11 +33,10 @@ response:
     - matches: request.url.pathname
       pattern: '^/(graphql|rest)(/|$)'
       use: proxy
-    # Requests to create an account are handled by the top-level 'index'
-    # object, which returns a generic app shell
+    # Requests to create an account are handled by the 'appShell' object.
     - matches: request.url.pathname
       pattern: '^/create-account'
-      use: index
+      use: appShell
```
sirugh: Instead of having two routes, `index` vs `appShell` or `hardcodedAppShell` vs `urlResolvedAppShell` as @zetlen and I discussed, I went with this method in which I re-used the single `appShell`. The benefit being that I did not have to rewrite much of the foundation that appShell already provides.

packages/venia-concept/src/util/makeUrl.js
```diff
@@ -1,8 +1,9 @@
 // If the root template supplies the backend URL at runtime, use it directly
-const {
-    imageOptimizingOrigin,
-    mediaBackend = 'https://backend.test/media/'
-} = document.querySelector('html').dataset;
+const htmlDataset = document.querySelector('html').dataset;
+const { imageOptimizingOrigin } = htmlDataset;
+// Protect against potential falsy values for `mediaBackend`.
+const mediaBackend = htmlDataset.mediaBackend || 'https://backend.test/media/';
```
sirugh: This fixes a bug identified where `mediaBackend` was an empty string which destructures as `""` instead of being falsy.

TBH I'm not sure why we have a default value here anyways -- seems like it just gets rewritten in the makeUrl code.

packages/venia-concept/upward.yml
```diff
@@ -33,11 +33,10 @@ response:
     - matches: request.url.pathname
       pattern: '^/(graphql|rest)(/|$)'
       use: proxy
-    # Requests to create an account are handled by the top-level 'index'
-    # object, which returns a generic app shell
+    # Requests to create an account are handled by the 'appShell' object.
     - matches: request.url.pathname
```
sirugh: Also, it's not clear to me why lines 37-39 are required. I would have assumed we could just resolve similar to how we resolve `/search.html`, but if I remove these lines I get failures like `unable to load "<url>/dist/create-account"`

dpatil-magento:
@sirugh  QA Pass, can be merged once review is approved.

packages/venia-concept/src/util/makeUrl.js
zetlen: Yeah, there shouldn't be a default here.

packages/venia-concept/upward.yml
zetlen: Actually, you've uncovered an optimization opportunity in the `resource` definition. Matches are evaluated top to bottom, so we want the hardcoded matches to be _above_ the matches that check `urlResolver`. When we know ahead of time that the hardcoded URL will not be resolved with urlResolver, we want to avoid evaluating urlResolver, since it's actually placing a graphQL query. There's gotta be a simpler way to do that in the future.

For now, to prevent merge conflicts, let's just move all the match clauses against literal URL paths to the _top_ of the conditional stack, and all the matches against `urlResolver` under them. That helps us avoid an additional request.

packages/venia-concept/src/util/makeUrl.js
sirugh: Ok -- rather than remove the default, @zetlen and I discussed and it is out of scope. That being said we should probably rewrite `makeUrl` or at least fix this to not need a fallback (ie. console.error and THEN use a default.

packages/venia-concept/upward.yml
sirugh: Ok, moved.

packages/venia-concept/src/util/makeUrl.js
sirugh: I just added a warning if it's falsy to at least indicate a problem.

PR #1286, opened 2019-06-01
[TASK] [issue-1154] Added firstname, lastname, email, telephone field…

lewisvoncken:
…s to the Billing Address and updated styling so is shown as shipping

## Description

Show additional fields in the Billing Address Form. The following fields are added:

- Firstname
- Lastname
- Email
- Telephone

## Related Issue

Closes #1154.
Closes #918.

## Verification Steps

1. Add a product to the cart
2. Click on the right sidebar on the Checkout Button
3. Fill in the Shipping Address Form
4. Now select a payment method
5. Uncheck the checkbox for `Billing address same as shipping address`
6. Now the same fields will be shown as the Shipping Address Form
7. Fill in the fields
8. Select a payment method
9. Place the order.

### Shipping Address Form
image

### Billing Address Form
image

## Proposed Labels for Change Type/Package

- [ ] major (e.g x.0.0 - a breaking change)
- [ ] minor (e.g 0.x.0 - a backwards compatible addition)
- [ ] patch (e.g 0.0.x - a bug fix)

## Checklist:

- [ ] I have updated the documentation accordingly, if necessary.
- [ ] I have added tests to cover my changes, if necessary.

PR #1449, opened 2019-07-18
Improve scroll locking

### LINKED ISSUE #1071: [BUG] Cropped shopping cart in Venia for larger phones

jimbo:
## Description

When the app is masked, such as when a drawer or other modal is open, we intend to prevent scrolling at the window level. Currently, we apply styles to the `<Main>`  element to achieve this, but it doesn't work properly in mobile Safari. This PR would shift scroll locking to the `html` and `body` elements as necessary for mobile Safari.

This also resolves a number of visual quirks and inconsistencies in the cart/checkout flow.

- [x] Lock window scrolling and overscrolling when mask is active
- [x] Fit minicart height to viewport in mobile Safari (replace `100vh`)
- [x] Adjust padding in cart footers (decrease)
- [x] Adjust payments form height (decrease)
- [x] Adjust braintree form styles (slightly)
- [x] Adjust checkout subform header padding (increase)
- [x] Adjust padding in cart item editor (increase)
- [x] Fit cart item editor to viewport in mobile Safari (fix auto sizing)

## Related Issue
Closes #1071.

## Verification Steps
1. Open the mobile simulator in Xcode.
2. Go to the deployment of this branch and open the cart.
3. Verify that the window can't be scrolled (or overscrolled) at all.
4. Verify that the cart is not obscured by Safari's chrome.

### DIFF
--- a/packages/venia-concept/src/components/Checkout/paymentsForm.css
+++ b/packages/venia-concept/src/components/Checkout/paymentsForm.css
@@ -1,31 +1,17 @@
 .root {
 }
 
+.heading {
+    composes: heading from './flow.css';
+}
+
 .body {
-    composes: body from './form.css';
-    align-content: start;
-    background-color: white;
-    grid-gap: 0.5rem;
+    composes: body from './flow.css';
     height: 30rem;
-    overflow: auto;
-    padding: 0 1.5rem;
 }
 
 .footer {
-    composes: footer from './form.css';
-}
-
-.heading {
-    background-color: white;
-    font-size: 0.875rem;
-    font-weight: 600;
-    grid-column-end: span 2;
-    line-height: 1rem;
-    padding: 1.5rem 0 1.25rem;
-    text-align: center;
-    text-transform: uppercase;
-    top: 0;
-    z-index: 1;
+    composes: footer from './flow.css';
 }
 
 .button {
--- a/packages/venia-concept/src/components/Checkout/paymentsForm.css
+++ b/packages/venia-concept/src/components/Checkout/paymentsForm.css
@@ -3,14 +3,12 @@
 
 .body {
     composes: body from './form.css';
+    align-content: start;
+    background-color: white;
     grid-gap: 0.5rem;
-    height: calc(
-        var(--minicart-height) - var(--flow-footer-height) -
-            var(--minicart-footer-padding-bottom)
-    );
-    overflow-y: scroll;
+    height: 30rem;
+    overflow: auto;
     padding: 0 1.5rem;
-    grid-auto-rows: max-content;
 }
 
 .footer {
@@ -18,13 +16,17 @@
 }
 
 .heading {
+    background-color: white;
     font-size: 0.875rem;
     font-weight: 600;
     grid-column-end: span 2;
     line-height: 1rem;
-    padding: 0.75rem 0;
+    padding: 1.5rem 0 1.25rem;
+    position: sticky;
     text-align: center;
     text-transform: uppercase;
+    top: 0;
+    z-index: 1;
 }
 
 .button {
@@ -84,4 +86,8 @@
  */
 [data-braintree-id='upper-container'] {
     z-index: unset;
+}
+
+:global .braintree-placeholder {
+    display: none;
 }
--- a/packages/venia-concept/src/components/Checkout/paymentsForm.css
+++ b/packages/venia-concept/src/components/Checkout/paymentsForm.css
@@ -22,8 +22,6 @@
     grid-column-end: span 2;
     line-height: 1rem;
     padding: 1.5rem 0 1.25rem;
-    position: -webkit-sticky;
-    position: sticky;
     text-align: center;
     text-transform: uppercase;
     top: 0;
--- a/packages/venia-concept/src/components/Checkout/paymentsForm.css
+++ b/packages/venia-concept/src/components/Checkout/paymentsForm.css
@@ -22,6 +22,7 @@
     grid-column-end: span 2;
     line-height: 1rem;
     padding: 1.5rem 0 1.25rem;
+    position: -webkit-sticky;
     position: sticky;
     text-align: center;
     text-transform: uppercase;
@@ -64,6 +65,10 @@
 .braintree,
 .street0 {
     grid-column-end: span 2;
+}
+
+.braintree {
+    min-height: 356px;
 }
 
 .validation {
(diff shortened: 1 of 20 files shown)


packages/peregrine/src/hooks/useScrollLock.js
```diff
@@ -0,0 +1,16 @@
...
+ * A [React Hook]{@link https://reactjs.org/docs/hooks-intro.html} that sets
+ * an attribute on the document element indicating that scrolling should be
+ * locked. This is performed with a layout effect (before paint).
+ *
+ * @kind function
+ *
+ * @param {Boolean} locked Whether scrolling should be locked.
+ */
+export const useScrollLock = locked => {
+    useLayoutEffect(() => {
```
sirugh: Just to clarify, we _must_ use `layoutEffect` here instead of `effect` because we need to do this before paint?

sirugh:
Did you also check the `paymentsForm` which relies on the cart height but also some calculations based on footers/etc?

sirugh:
Also check this screenshot out. In chrome, I set the height of the `paymentsForm` body to:

```
    height: calc(
        100vh - var(--flow-footer-height) -
            var(--minicart-footer-padding-bottom)
    );
```

In iOS I have to add an additional height calculation for the iOS footer to achieve the same effect, note the `- 7rem`:

```
    height: calc(
        100vh - 7rem - var(--flow-footer-height) -
            var(--minicart-footer-padding-bottom)
    );
```
[Image from Gyazo](https://gyazo.com/bb5f219beb0dc0262955caf2be775057)

sirugh:
Going to need to move files to `lib`.

packages/venia-concept/src/components/Checkout/paymentsForm.css
```diff
@@ -3,28 +3,30 @@
...
     composes: body from './form.css';
+    align-content: start;
+    background-color: white;
     grid-gap: 0.5rem;
-    height: calc(
-        var(--minicart-height) - var(--flow-footer-height) -
-            var(--minicart-footer-padding-bottom)
-    );
-    overflow-y: scroll;
+    height: 30rem;
```
sirugh: Are you making a conscious decision to not have the billing form span the height of the viewport? If so, I'm fine with this change, but we are definitely changing the way the payment form looks now.

[Image from Gyazo](https://gyazo.com/783ab9be485c555f1e056cef407c1b57)

sirugh: Also, why not just use `vh` units?

packages/peregrine/src/hooks/useScrollLock.js
jimbo: Yeah, we know it's going to cause a reflow so I think we want to have it take effect before the next paint.

packages/venia-concept/src/components/Checkout/paymentsForm.css
jimbo: The whole problem here is that mobile Safari doesn't understand that the "viewport" shouldn't include browser chrome. So `100vh` works fine in every other browser, but in Safari you end up with buttons sitting underneath the browser UI at the bottom of the screen.

jimbo:
> Did you also check the paymentsForm which relies on the cart height but also some calculations based on footers/etc?

This has been addressed.

> Are you making a conscious decision to not have the billing form span the height of the viewport? If so, I'm fine with this change, but we are definitely changing the way the payment form looks now.

Yes, I'm making that conscious decision. None of the other forms do that—you're supposed to be able to see some of the mask and the cart items underneath. The payments form already has to handle overflow (read: scroll) because it can exceed the height of the viewport, so we might as well reduce it to an intermediate size and let it overflow.

packages/venia-concept/src/components/Checkout/paymentsForm.css
sirugh: Ok -- can we maybe increase it a tiny bit? As is unchecking the box to add a distinct address doesn't cause the UI to change or move so it's a little bit of a weird UX. Maybe we should add an scroll/focus to the address form when the button is unchecked?

jimbo: Haha, I was just demoing this to @soumya-ashok and she had the same feedback. I've added an effect to scroll the new fields into view when added.

packages/venia-concept/src/components/Checkout/paymentsFormItems.js
```diff
@@ -79,6 +81,7 @@ const PaymentsFormItems = props => {
                     />
                 </Field>
             </div>
+            <span ref={anchorRef} />
```
sirugh: I like this. I just wanted to point out that we are expanding this to be 5 lines of form inputs in #1286. I checked and this code will function fine, even if the form is 6 lines long. Beyond that, however, the "Billing Information" banner hides the top of the form. If you could make this resilient to that, such as if the top field/label were scrolled just under the header banner, then that would be best, however I don't think it's worth holding up the PR since, as I said, it'll work as is (3 lines of form inputs) and will work even up to 6 lines.

dpatil-magento:
@jimbo @sirugh Please check below observations -
1.  Desktop chrome browser > Home page, search results, category page shifts left on accessing Left/Right drawer.



2. On Android - When user access shipping form and tap on Name field - Form shifts up and user cant see whats being typed. (Tried on actual device and also on browser stack virtual device.  )

sirugh:
@dpatil-magento is the second one a problem on `develop`?

dpatil-magento:
@sirugh Yes, I do see the second issue on `develop` too but thought that needs to be fixed as part of this ticket as its related to scrolling/focus ?

packages/venia-concept/src/components/MiniCart/cartOptions.css
```diff
@@ -1,11 +1,11 @@
 .root {
-    position: inherit;
     display: grid;
-    grid-template-rows: min-content 1fr auto auto;
-    width: 100%;
-    bottom: 0;
-    right: 0;
+    grid-template-rows: min-content 1fr;
+    height: calc(100% - var(--minicart-header-height));
```
supernova-at: I'm so happy to get rid of this ⭐️

jimbo:
@dpatil-magento I've fixed the second issue for now. In general, presentation is not great in the Android scenario because the keyboard reduces the viewport height rather than overlaying it, but the checkout forms have now been adjusted to constrain themselves to the viewport.

dpatil-magento:
Thanks @jimbo , looks good now. 
QA Pass.

PR #1450, opened 2019-07-18
Use Google Tag manager for analytics

belbiy:
## Description

This PR changes the way we track users with Google Analytics on the PWA docs site. It adds the Google Tag manager script to deliver the tracking and the Hotjar code. 

Same Google Tag manager container is used as we use on devdocs.magento.com


jcalcaben:
@dpatil-magento I don't think there's anything here to QA. There is a dangerfile error but I don't think its related to this PR.

dpatil-magento:
@jcalcaben After latest develop merge error gone. Current errors are related to missing details. Regression scripts pass.

PR #1467, opened 2019-07-25
Fix ApolloContext

### LINKED ISSUE #1465: [bug]: ApolloContext not available
**Describe the bug**
The `react-apollo` module appears to no longer export `ApolloContext`. This causes an error in the build and at runtime; since the export is missing, the Apollo client can no longer be retrieved from context and queries can't be performed.

```
ERROR in ./node_modules/react-apollo/ApolloContext.js
Module build failed: Error: ENOENT: no such file or directory, open '/Users/sanford/code/pwa-studio/packages/venia-concept/node_modules/react-apollo/ApolloContext.js'
 @ ./src/drivers/adapter.js 7:0-59 76:31-44
 @ ./src/drivers/index.js
 @ ./src/index.js
 @ multi (webpack)-dev-server/client?https://magento-venia-concept-e1r4b.local.pwadev:8048 (webpack)/hot/dev-server.js ./src/index.js

ERROR in ./node_modules/react-apollo/react-apollo.esm.js
Module build failed: Error: ENOENT: no such file or directory, open '/Users/sanford/code/pwa-studio/packages/venia-concept/node_modules/react-apollo/react-apollo.esm.js'
 @ ./src/drivers/index.js 1:0-37 1:0-37
 @ ./src/index.js
 @ multi (webpack)-dev-server/client?https://magento-venia-concept-e1r4b.local.pwadev:8048 (webpack)/hot/dev-server.js ./src/index.js
ℹ ｢wdm｣: Failed to compile.
```

**To Reproduce**
1. `yarn watch:venia`
2. Open the app

**Expected behavior**

There should be errors in the build or the browser, and queries should work.

**Additional context**

Apollo was already using a faulty custom implementation of context to "provide" their client, which we were working around in `Adapter`. Since we were effectively relying on a private API, they may have broken it.

**Possible solutions**

Provide the Apollo client via vanilla context.

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

jimbo:
## Description

The latest `react-apollo` update broke our usage of `ApolloContext`.

- Now that it's a top-level export, we need to import it directly from `react-apollo`.
- Since the context value has changed, we need to destructure it to obtain `client`.
- Peregrine's version of `react-apollo` needs to match Venia's.

## Related Issue

Closes #1465.

## Verification Steps

1. `yarn install`
2. `yarn watch:venia`
3. Search for a product.

There should be no errors from the build or in the browser console.

## Checklist:


packages/peregrine/lib/hooks/useApolloContext.js
```diff
@@ -1,10 +1,14 @@
...
 
 /**
  * A function that provides access to the [Apollo client]{@link https://www.apollographql.com/docs/react/api/apollo-client} API.
  *
  * @return {ApolloClient} The Apollo client for this application
  * @kind function
  */
-export const useApolloContext = () => useContext(ApolloContext);
+export const useApolloContext = () => {
+    const { client } = useContext(ApolloContext) || {};
```
jimbo: Apollo's context value seems to have changed in order to accommodate some private query data (maybe cached results?). We could pass the new raw context object, but we'd have to change the consumers, but I think that'd be a breaking change, and they should only need the client anyway.

If a consumer *really* needs the other private members, it can consume `ApolloContext` directly with `useContext`.

sirugh: I agree with your reasoning, but I think we should include that as a note in the comment block.

packages/peregrine/package.json
```diff
@@ -22,7 +22,7 @@
     "@storybook/react": "~4.1.13",
     "intl": "~1.2.5",
     "react": "~16.8.3",
-    "react-apollo": "~2.5.3",
+    "react-apollo": "~2.5.5",
```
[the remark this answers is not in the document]
jimbo: I'm sure this qualified as private API.

packages/venia-concept/src/components/SearchBar/__tests__/autocomplete.spec.js
```diff
@@ -8,7 +8,7 @@ import Autocomplete from '../autocomplete';
 jest.mock('src/classify');
 jest.mock('@magento/peregrine');
 jest.mock('../suggestions', () => () => null);
-jest.doMock('react-apollo/ApolloContext', () => React.createContext());
+jest.doMock('react-apollo', () => React.createContext());
```
supernova-at: Since the import is now 

```js
import { ApolloContext } from 'react-apollo';
```

doesn't this mock have to return an object with an `ApolloContext` property?

```js
jest.doMock('react-apollo', () => ({ ApolloContext: React.createContext() }));
```

jimbo: Actually, that mock isn't even need anymore. Will remove.

PR #1478, opened 2019-07-26
bug: fix pdp image jutter

### LINKED ISSUE #1476: [bug]: pdp image jutter with transparent placeholder
**Describe the bug**
The transparent placeholder is 4x5. Our sample data images are not 4:5 so manually enforce a crop where transparent placeholder is used.

**To Reproduce**
1. Go to pdp in desktop
2. Click on thumbnails and see image jutter.

**Expected behavior**
No page jutter since the placeholder and image are same size.

**Screenshots**

**Additional context**

**Possible solutions**

 - Browser Version [e.g. 22]:
 - Magento Version:
 - PWA Studio Version:
 - NPM version `npm -v`:
 - Node Version `node -v`:

- [ ] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
## Description

See issue

## Related Issue

Closes #1476 .

## Verification Steps

For minicart:

1. Sign in, add an item to cart and open minicart. The image should appear.
2. Clear app cache and disable cache. Refresh the page.
3. Open the cart, you should see the item in cart but the image should be a grey box.

For PDP:
1. Check that images do not cause jutter on the PDP when changing image via clicking thumbnails.

## Checklist:


packages/venia-concept/src/actions/cart/asyncActions.js
```diff
@@ -350,6 +350,9 @@ export const getCartDetails = (payload = {}) => {
             // for each item in the cart, look up its image in the cache
             // and merge it into the item object
             // then assign its options from the totals subResource
+            // TODO: If we don't have the image in cache we should probably try
+            // to find it some other way otherwise we have no image to display
+            // in the cart and will have to fall back to a placeholder.
```
sirugh: If a user signs in on a different device after adding an item we have no image to display in the cache. I fixed it by adding some conditional usage in this PR but I also added this TODO because we should probably find a way to re-request a non-cached image for an item added from another device or before a cache clear.

jimbo:
How does adding height parameters affect the issue?

sirugh:
> How does adding height parameters affect the issue?

The jutter? Because the placeholder is 4:5 and our images are not.

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -22,14 +26,26 @@ const Product = props => {
...
+
+    const productImage = useMemo(() => {
+        const src =
+            image && image.file
+                ? resourceUrl(image.file, {
+                      type: 'image-product',
+                      width: imageWidth,
+                      height: imageHeight
+                  })
+                : transparentPlaceholder;
```
supernova-at: Out of curiosity, would we ever need a transparent placeholder that isn't 4x5?

packages/venia-concept/src/components/MiniCart/product.js
```diff
@@ -22,14 +26,26 @@ const Product = props => {
...
+
+        return (
+            <Image
+                alt={name}
+                classes={{ root: classes.image }}
+                placeholder={transparentPlaceholder}
+                src={src}
```
supernova-at: I'm wondering if we shouldn't just have the `Image` component render the `placeholder` when the `src` image fails to load. We could get rid of the conditional logic above (L32) if the component would take care of that for us.

packages/venia-concept/src/components/MiniCart/product.js
supernova-at: Related, would an alternate fix be to change our transparent placeholder to match the sample data ratio?

packages/venia-concept/src/components/ProductImageCarousel/carousel.css
```diff
@@ -14,7 +14,7 @@
 }
 
 .currentImage {
-    background-color: rgb(var(--venia-grey));
+    background-color: white;
```
supernova-at: Do we want to create a named CSS variable for this?

packages/venia-concept/src/components/MiniCart/product.js
sirugh: > Out of curiosity, would we ever need a transparent placeholder that isn't 4x5?

Sure, and we can create/update those when we get to that point. All of the images (except for the landing page category "circle" images) _should_ be 4x5.

> Related, would an alternate fix be to change our transparent placeholder to match the sample data ratio?

I don't think it's worth it right now. @revanth0212 is going to investigate better ways to move forward with images/lazy-loading/etc and I think this may all change anyways.

supernova-at:
✅ Verification steps complete

That said,

> Open the cart, you should see the item in cart but the image should be a grey box.

What is actually happening here? If the image is not cached, shouldn't we be trying to fetch it so we can show it?

packages/venia-concept/src/components/MiniCart/__tests__/__snapshots__/product.spec.js.snap
```diff
@@ -2,10 +2,7 @@
 
 exports[`it renders a mask div before the kebab if it is loading 1`] = `
 <li>
-  <img
-    alt="Unit Test Product"
-    src="path/to/image/source"
-  />
+  path/to/image/source
```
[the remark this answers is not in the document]
sirugh: This whole test is unnecessary anymore.

packages/venia-concept/src/components/MiniCart/product.js
sirugh: It's already doing that. It'll render a placeholder _until_ the `src` loads, which means if it fails it'll just stay as the placeholder. What the conditional is doing is saying "if there is no source, just render a placeholder AS the source`.

packages/venia-concept/src/components/ProductImageCarousel/carousel.css
sirugh: Sure.

sirugh:
> What is actually happening here? If the image is not cached, shouldn't we be trying to fetch it so we can show it?

If a user signs in on a different device after adding an item we have no image to display in the cache. I fixed it by adding some conditional usage in this PR but I also added this TODO because we should probably find a way to re-request a non-cached image for an item added from another device or before a cache clear.

supernova-at:
> So yes, we should try, but probably not in the scope of this PR :D

Okay, got it. At any rate, it's an improvement over what currently exists on `develop`, which is an error toast saying "Sorry! An unexpected error occurred. Debug: TypeError36T at joinUrls (/js/client.js)"  😬

packages/venia-concept/src/components/ProductImageCarousel/carousel.css
jimbo: (Obviously not for `white`, which is a global constant, but perhaps for how this view uses it.)

PR #1480, opened 2019-07-29
[dev]: use less restrictive version

### LINKED ISSUE #1479: Node js version fixed to 10.x. bug or expected?
(some) Packages in pwa-studio have a node engine version fixed to 10.x... some require at least 10.x. Found this while trying to deploy on 12.6.0. Is this an expected behavior? Manually adjusting packages to use node ">=10.x" seems to work just fine, however, every deployment would require manual adjustment.

sirugh:
## Description

We specifiy our allowed node version to `10.x` which is perhaps a bit too restricting.

## Related Issue

Closes #1479 .

## Verification Steps

1. `yarn install` should work

## Checklist:


sirugh:
@zetlen I'm totally fine restricting to `10.x` as it is LTS. LMK if you want me to delete the PR.

supernova-at:
Do we know what breaks on `12.x`? Can we fix it?

packages/venia-concept/package.json
```diff
@@ -98,7 +98,7 @@
     "sharp": "~0.21.2"
   },
   "engines": {
-    "node": "10.x",
+    "node": ">=10.x",
```
zetlen: Can you leave this one more restrictive? We don't want to support the actual consuming app for infinite forward versions. That's too much that can go wrong.

The library code, on the other hand, can totally use the version string you suggest.

sirugh:
> Do we know what breaks on 12.x? Can we fix it?

I'm not sure, maybe @revanth0212 has some info? By making this change we will at least start getting reports of when it's broken for people on 12, and then we can debug from there.

packages/venia-concept/package.json
sirugh:
```suggestion
    "node": "10.x",
```

PR #1482, opened 2019-07-29
Remove storybook from peregrine and update references

### LINKED ISSUE #1380: Remove Storybook from Peregrine
Given the definition and direction of Peregrine, at this time we don't recommend storybook integration with the Peregrine library (which is not meant to contain visual components). 

As such, please remove Storybook from Peregrine.

Note that we may add this functionality back later if it becomes relevant.

sirugh:
## Description

Removes storybook from peregrine and all references from the repo.

## Related Issue

Closes #1380.

## Verification Steps

1. Check that I did, in fact, remove storybook from the peregrine repo and that all references to peregrine specific storybook are gone.

## Checklist:


pwa-devdocs/src/_data/peregrine.yml
```diff
@@ -6,9 +6,6 @@ entries:
     - label: Routing in PWA Studio
       url: /peregrine/routing/
 
-    - label: Storybook UI testing
```
sirugh: @jcalcaben I removed the docs references for storybook but I think they aren't specific to just peregrine. Maybe we can migrate them to the `venia-concept` docs? LMK if you want me to do that and I can do it here, or you are welcome to push to my branch.

jcalcaben: If we're still going to use Storybook, we should move it somewhere else. If possible, can you move the `/peregrine/testing/storybook` directory to `pwa-devdocs/drafts` instead of deleting it so it doesn't get published. 

Then we can create another issue to revise the content and move it to a more appropriate place. That way that unrelated task won't hold up this issue.

packages/peregrine/.npmignore
```diff
@@ -1,4 +1,3 @@
 .eslintrc.js
-.storybook
 /lib/**/__{docs,helpers,mocks,stories,tests}__/**
```
supernova-at: You could remove `stories` from this line too. It won't break anything keeping it in, so not a huge deal.

sirugh:
```suggestion
/lib/**/__{docs,helpers,mocks,tests}__/**
```

dpatil-magento:
QA pass, as Regression on venia looks good, ,  `yarn run storybook` does not work anymore, reference has been removed in devdocs.

PR #1483, opened 2019-07-29
[SPIKE] Migrate some app state

jimbo:

### DIFF
--- a/packages/peregrine/lib/PeregrineContextProvider/peregrineContextProvider.js
+++ b/packages/peregrine/lib/PeregrineContextProvider/peregrineContextProvider.js
@@ -1,11 +1,13 @@
 import React from 'react';
 
+import { UserContextProvider } from '../state/User';
+
 /**
  * List of essential context providers that are required to run Peregrine
  *
  * @property {React.Component[]} contextProviders
  */
-const contextProviders = [];
+const contextProviders = [UserContextProvider];
 
 const PeregrineContextProvider = ({ children }) => {
     return contextProviders.reduceRight((memo, ContextProvider) => {
--- a/packages/peregrine/lib/hooks/useRestApi.js
+++ b/packages/peregrine/lib/hooks/useRestApi.js
@@ -32,7 +32,7 @@ export const useRestApi = endpoint => {
                 receiveError(error.baseMessage);
             }
         },
-        [receiveError, receiveResponse, setLoading]
+        [endpoint, receiveError, receiveResponse, setLoading]
     );
 
     const api = useMemo(
--- a/packages/peregrine/lib/state/User/context.js
+++ b/packages/peregrine/lib/state/User/context.js
@@ -1,0 +1,15 @@
+import React, { createContext, useContext } from 'react';
+
+import { useUserState } from './state';
+
+const UserContext = createContext();
+
+export const UserContextProvider = ({ children }) => {
+    const store = useUserState();
+
+    return (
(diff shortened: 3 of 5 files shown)


packages/peregrine/lib/state/Checkout/state.js
```diff
@@ -0,0 +1,229 @@
+import { useCallback, useMemo, useReducer } from 'react';
```
sirugh: You'll note in here that I didn't copy over _everything_. For example I left out much of the reducer functions that were "enclosing" actions like `begin` and `reject`. If we want to store "error" or "active" state such as `editing` or `step` or `submitting` we can, but IMO those belong in the UI.

packages/venia-concept/src/components/Checkout/cart.js
```diff
@@ -1,19 +1,133 @@
...
+
+    useEffect(() => {
+        if (userState.isSignedIn) {
+            const { loading, data } = authedShippingResponseState;
+            if (!loading && !data) {
+                authedShippingRestApi.sendRequest({
+                    options: SHIPPING_METHOD_OPTIONS
+                });
+            }
+        } else if (cartState.cartId) {
```
sirugh: We can't fetch the guest shipping cart options until we have a cart id.

packages/venia-ui/lib/components/Checkout/cart.js
```diff
@@ -1,19 +1,152 @@
...
+
+const isCheckoutReady = (cartState, checkoutState, directoryState) => {
+    const hasCountries = !!directoryState.countries;
+    const hasShippingMethods = checkoutState.availableShippingMethods.length;
+    return hasCountries && hasShippingMethods && isCartReady(cartState);
+};
 const Cart = props => {
-    const { beginCheckout, ready, submitting } = props;
+    const { cart: cartState, user: userState } = props;
+    // TODO: Use new state when we migrate cart and user over.
```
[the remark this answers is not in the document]
sirugh: Not entirely. The state and some initial set/update reducers have been written but the code that creates/queries cart info is not created. This is using the connected redux cart for now. One way around this would be to have the redux actions update the new cart context as well as the redux context. Then we could move each action over separately.

packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
```diff
@@ -269,7 +269,7 @@ async function configureWebpack({ context, vendor = [], special = {}, env }) {
         });
 
         if (isDevServer()) {
-            config.devtool = 'cheap-source-map';
+            config.devtool = 'eval-source-map';
```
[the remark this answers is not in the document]
sirugh: Yea, `cheap-source-map` strips empty lines and comments. I want both for readability which is why we have source maps in the first place :D

packages/peregrine/lib/state/Cart/state.js
```diff
@@ -0,0 +1,165 @@
...
+
+const wrappedReducer = withLogger(reducer);
+export const useCartState = () => {
+    const [state, dispatch] = useReducer(wrappedReducer, initialState);
+    const { cartId } = state;
+
+    const createCart = useCallback(
+        async isSignedIn => {
+            if (cartId) {
+                return cartId;
```
sirugh: Will this actually update the `cartId` across components? ie If componentA calls `createCart` which should dispatch to the state, what happens when componentB calls `createCart` later?

PR #1494, opened 2019-07-30
Updates GitHub Pages to redirect to pwastudio.io

### LINKED ISSUE #1492: Update GitHub Pages to Redirect to new location
Our GitHub Pages site will not automatically redirect when the repository moves to the `magento` organization, it may serve stale content.

Update the GitHub Pages site before the move to redirect to http://pwastudio.io/ so that if stale content is served it will redirect to the new content.

Also update the 404 page to redirect.

See more details in https://github.com/magento-research/pwa-studio/issues/1472#issuecomment-516498022.

supernova-at:
## Description

This PR replaces the `index` and `404` pages with redirects to [pwastudio.io](https://pwastudio.io).
Proper redirect code taken directly from [WCAG recommendations](https://www.w3.org/TR/WCAG20-TECHS/H76.html).

We should publish these changes right before we move the repository, ensure they work, and then move it.

## Related Issue

Closes #1492 , created when Columbus sailed the ocean blue 🌊 .

## Verification Steps

In a private repo, @jcalcaben is ensuring that modifying these files with redirects does what we expect it to.

We could test these changes and then change them back before we move the repo if we want to ensure they work.

Any commit to this `gh-pages` branch will update the GitHub Pages site.

## Checklist:

## Note / Risk

In an effort to not index this old stale content we remove Google Tag Manager code. 
The downside of this is that we won't be able to track when / how many folks are still hitting the old site.


jcalcaben:
We don't need this PR to update gh-pages in this case. We just need to force push the contents of [my test repo](https://github.com/jcalcaben/migration-test/tree/gh-pages) to this repo's `gh-pages` branch before the migration.

supernova-at:
Oh okay, however you want to do it.  I'll update this with the clickable link "if you're not redirected", I think that's a good addition.

supernova-at:
> contents of my test repo

Do you not need `head` and `body` tags there?

I was going to mention that you may want to put the `script` tag that redirects as the last element in `head` but then I didn't see a `head` at all!

image

There's a joke in here somewhere about going `headless` too...

jcalcaben:
I guess it works without it? 🤷‍♂ But I'll add it just in case.

PR #1495, opened 2019-07-30
MagentoRouteHandler: correct handling of unknown routes

### LINKED ISSUE #1454: [bug]: MagentoRouteHandler gives InternalError instead of NotFound (404)
When the urlResolver-Query can not resolve a URL (let's say '/asdf') it gives back NULL.
`{
  "data": {
    "urlResolver": null
  }
}`

Because of that `resolvedRoute` can not be destructered.
https://github.com/magento-research/pwa-studio/blob/70700cf7d9a745623cc9a44119520ff278d9c20e/packages/peregrine/lib/Router/magentoRouteHandler.js#L97 

Which gets caught in the catch block, but because the `message` isn't 400, it gives back `InternalError` instead of `NotFound`, what would be the more accurate usecase in my opinion.
https://github.com/magento-research/pwa-studio/blob/70700cf7d9a745623cc9a44119520ff278d9c20e/packages/peregrine/lib/Router/magentoRouteHandler.js#L112

Because of this whole thing, you can't render a nice 404-page in the venia-app, because the false errorstate gets returned to the errorview (internalError instead of notFound).
https://github.com/magento-research/pwa-studio/blob/70700cf7d9a745623cc9a44119520ff278d9c20e/packages/peregrine/lib/Router/magentoRouteHandler.js#L137

- [ ] `venia-concept`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

niklaswolf:
## Description
MagentoRouteHandler now correctly returns NotFound if the urlResolver-Query returns null.
Further description see issue #1454 

## Related Issue
Closes #1454 .

## Verification Steps
1. Edit `RootComponents/CMS/cms.js`:

```jsx
import React, {Component, Fragment} from 'react';
import CategoryList from '../../components/CategoryList';
import {Link} from "react-router-dom";

export default class CMS extends Component {
    render() {
        return <Fragment>
            <CategoryList title="Shop by category" id={2} />
            <Link to={'/junk.html'}>Click me</Link>
        </Fragment>;
    }
}
```

2. Click the link in Venia
3. See that the error message reads "That page could not be found. Please try again." instead of "Something went wrong. Please try again."

## Checklist:


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1495) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento-research/pwa-studio?pullRequest=1495) before we can accept your contribution.<br/>

niklaswolf:
@supernova-at You don't see this error in these cases, because you do initial requests. The initial requests get resolved directly in the upward server, so you get the "not found" shell.

Only non-initial route changes get processed by the magento route handler (see linked issue #1454 ). So to force this error, you'd have to include a broken route link somewhere into the app, e.g. I included a broken link on the home page
RootComponents/CMS/cms.js
```
import React, {Component, Fragment} from 'react';
import CategoryList from '../../components/CategoryList';
import {Link} from "react-router-dom";

export default class CMS extends Component {
    render() {
        return <Fragment>
            <CategoryList title="Shop by category" id={2} />
            <Link to={'/junk.html'}>Click me</Link>
        </Fragment>;
    }
}
```

When you click the link, you'll now see a 404 "That page could not be found. Please try again." from the ErrorView component. Without the changes, it's a 500 "Something went wrong. Please try again."

Hit me up when you have further issues testing (I'm also available in the community slack ;))

packages/peregrine/lib/Router/magentoRouteHandler.js
```diff
@@ -94,8 +94,12 @@ export default class MagentoRouteHandler extends Component {
                 route: pathname
             });
 
-            const { type, id } = resolvedRoute;
+            // urlResolver query returns null if a route can't be found
+            if (resolvedRoute === null) {
```
zetlen: This seems to rely on a literal null, and I'm not sure that's guaranteed. Any reason not to just test it for truthiness here?

niklaswolf: In all of my tests null was returned, if there was no match, that's why I decided to make it rely on literal null. And I wanted to have these changes as unobstrusive as possible. But I'm also fine with just checking for truthiness. Should I change this?

zetlen: Yeah, would you mind?
```suggestion
            if (!resolvedRoute) {
```

niklaswolf: @zetlen Applied your suggested change

PR #1497, opened 2019-07-31
V4P1: chore: remove root-relative imports from venia code

### LINKED ISSUE #1502: [feature]: Make Venia components less dependent on specific Webpack config
Venia's root-relative imports and complex Webpack config file make it too easy to tightly couple its React components to the specific Venia app. If our code is to be portable this needs to change.

- An abstraction of Webpack config into Buildpack
- Simplification and rigor applied to Venia

**Additional context**
Ultimately this will lead to modularity and a scaffolding story :)

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

zetlen:
## Description

We configured Venia's build process to allow [root-relative imports](https://spin.atomicobject.com/2017/10/07/absolute-paths-javascript/), because a common pain of CommonJS and ES6 development is tracking the relative path strings between your modules. We should stop, because it makes the Venia code less portable--and portability is the goal of Venia modules.

Root-relative imports allow us to replace code like this:

```js
import actions from '../../../actions/checkoutReceipt';
import { createAccount } from '../../../actions/checkout';
import Receipt from './receipt';
import { getOrderInformation } from '../../../selectors/checkoutReceipt';
```

with code like this:

```js
import actions from 'src/actions/checkoutReceipt';
import { createAccount } from 'src/actions/checkout';
import Receipt from './receipt';
import { getOrderInformation } from 'src/selectors/checkoutReceipt';
```

That's a big improvement to DX, but it only works in _local modules_; a root-relative import in a module from the node_modules directory won't work, because Webpack won't resolve it in the same way. And if Webpack _did_ resolve it that way, that would create circular dependencies where dependency code was requiring application code.

So, we should preserve the ability to do root-relative imports in apps created with PWA Studio (see upcoming PR TODO), but our Venia code must use regular relative imports for maximum reusability.

This PR simply replaces them all.

This and #1498 closes #1502.

## Verification Steps

1. From a newly cloned repo, run `yarn && yarn build && yarn watch:venia`
2. Make sure the app still works in dev mode
3. Kill that process and run `yarn stage:venia`
4. Make sure the app still works in staging mode

This is a patch change because it doesn't affect public API.

## Checklist:


sirugh:
> a root-relative import in a module from the node_modules directory won't work, because Webpack won't resolve it in the same way. And if Webpack did resolve it that way, that would create circular dependencies where dependency code was requiring application code.

Isn't venia-concept still "built"? Is the plan to make it so that one can `yarn install @magento/venia-modules` and then use them directly as a library? I think I need a refresher on "the plan" :D

PR #1512, opened 2019-08-01
Swatch Treatments

### LINKED ISSUE #1390: [feature]: Update Swatch Treatments
Immediately upon color selection on the product detail page, this is the current state -

The desired state is -

Screen Shot 2019-06-25 at 9 37 21 AM

Acceptance Criteria:
  * Remove the `Tooltip` component and its usages (on the PDP and in the MiniCart)
  * Replace with the treatment from the image above
  * New component should be fully unit tested

dani97:
## Description

Update Swatch Treatments.

## Related Issue

Closes #1390 .

## Verification Steps

1. Go to the configurable PDP.
2. Verify the tool-tips not shown when selecting the options and new treatment comes.
3. Make sure the same working in cart.

image

- [X] minor (e.g 0.x.0 - a backwards compatible addition)

## Checklist:


packages/venia-concept/src/components/ProductOptions/tile.js
```diff
@@ -40,11 +39,11 @@ class Tile extends Component {
         const { label } = item;
 
         return (
-            <Tooltip text={label}>
-                <button {...restProps} className={className}>
-                    <span>{label}</span>
-                </button>
-            </Tooltip>
+
```
sirugh: Can remove whitespace.

packages/venia-concept/src/components/ProductOptions/tile.js
```diff
@@ -2,7 +2,6 @@ import React, { Component } from 'react';
 import { bool, number, shape, string } from 'prop-types';
 
 import classify from '../../classify';
-import Tooltip from './toolTip';
```
sirugh: You can also delete the entire Tooltip component and related tests, I believe.

packages/venia-concept/src/components/ProductOptions/swatch.js
```diff
@@ -68,16 +67,13 @@ class Swatch extends Component {
         });
 
         return (
-            <Tooltip text={label}>
-                <button
-                    {...restProps}
-                    className={className}
-                    style={finalStyle}
-                    title={label}
```
sirugh: Any reason why you removed the `title` prop?

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -54,6 +69,9 @@ class Option extends Component {
                     items={values}
                     onSelectionChange={handleSelectionChange}
                 />
+                <p className={classes.selection}>
+                    {selectedValue?`Selected ${label} : ${this.valuesMap.get(Array.from(selectedValue).pop())}`:''}
```
sirugh: I think this would be more readable if you pulled it out above the `return`.

For example:
```js
const selectedLabel = selectedValue ? ... : '';

return (
  ...
   <p className={classes.selection>
      {selectedLabel}
  </p>
  ...
);
```

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -10,6 +10,20 @@ import defaultClasses from './option.css';
 const getItemKey = ({ value_index }) => value_index;
 
 class Option extends Component {
+
+    constructor(props) {
```
sirugh: How do you feel about rewriting this component as a function component? Look at some of the rewrites we've been doing lately for reference but basically we will convert things like state to `useState` and the creation of the values map to a `useMemo` generator.

packages/venia-concept/src/components/ProductOptions/option.js
sirugh: I'm also a little worried about human readable label values. Will it be `Selected fashion_color` or is the values label already human readable?

dani97: I have kept the map  as { attribute_code : store_label } So the text will be `Selected Fashion Color<attribute_label>: Burgundy<store_label>`

packages/venia-concept/src/components/ProductOptions/option.js
dani97: Will refactor to a function components @sirugh

packages/venia-concept/src/components/ProductOptions/option.js
dani97: sure will make that change. :+1:

packages/venia-concept/src/components/ProductOptions/swatch.js
dani97: Just to avoid default tooltip generated created by the title attribute.
I will disable with css if it affects a11y

dani97:
@jimbo As per @sirugh  comment I refactored options component into function component but I could not fix  tests
1 ) does not call onSelectionChange if not provided
2 ) calls onSelectionChange function with optionId and selection
These two tests get options instance and call `handleSelectionChange` method.
Since I refactored it into function component, I cannot access the instance. Please advise on this.

jimbo:
@dani97 Thanks for taking this on. Those tooltips were bothering me for a long time! 😅 

I've updated the tests for you; have a look and let me know if you have any questions. Make sure to use `createTestInstance` from Peregrine, and use the [Test Renderer docs](https://reactjs.org/docs/test-renderer.html) to work with the returned instance.

I also made some minor changes to the `Option` component. In general, it's better to not keep the `selection` set in multiple components' state, so I swapped `selection` to be the derived state (the label of the selected option). Should simplify things a bit.

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -1,62 +1,85 @@
...
-                    onSelectionChange={handleSelectionChange}
-                />
-            </div>
+// TODO: get an explicit field from the API
+// that identifies an attribute as a swatch
+const getListComponent = (attribute_code, values) => {
+    const optionType = getOptionType({ attribute_code, values });
+
+    return optionType === 'swatch' ? SwatchList : TileList;
+};
```
jimbo: Note that I extracted this function from the component body. If it can exist statically, it should.

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -1,62 +1,85 @@
...
+        onSelectionChange,
+        values
+    } = props;
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const [selection, setSelection] = useState(null);
+
+    const ValueList = useMemo(() => getListComponent(attribute_code, values), [
+        attribute_code,
+        values
+    ]);
```
jimbo: This is also a good place for `useMemo`.

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -1,62 +1,85 @@
...
+
+    const valuesMap = useMemo(() => {
+        return new Map(
```
jimbo: Good use of `useMemo` here. 👍

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -1,62 +1,85 @@
...
+    }, [values]);
+
+    const selectedValueLabel =
+        selection != null ? `Selected ${label} : ${selection}` : '';
+
+    const handleSelectionChange = useCallback(
+        selection => {
+            const [selectedValue] = Array.from(selection);
+
+            setSelection(valuesMap.get(selectedValue));
```
jimbo: Note that we're now storing the selected label in state, rather than the set of indices.

packages/venia-concept/src/components/ProductOptions/option.js
```diff
@@ -1,62 +1,85 @@
...
+        <div className={classes.root}>
+            <h3 className={classes.title}>
+                <span>{label}</span>
+            </h3>
+            <ValueList
+                getItemKey={getItemKey}
+                items={values}
+                onSelectionChange={handleSelectionChange}
+            />
+            <p className={classes.selection}>{selectedValueLabel}</p>
```
jimbo: Consider making this paragraph (`<p>`) a description list (`<dl>`) and splitting the label and value into term (`<dt>`) and description (`<dd>`).

packages/venia-concept/src/components/ProductOptions/swatch.js
jimbo: That default tooltip from the `title` attribute is very important for accessibility, even for users without vision impairment. I do suggest restoring it (and not hiding it with CSS).

brendanfalkowski: It doesn't affect this context (button with only an icon as child content), but I read somewhere that this was an accessibility anti-pattern:

`<button title="Thing">Thing</button>`

Which a screen reader would announce as: 

`BUTTON Thing TITLE Thing`

Has anyone encountered that, or has the practice changed? Years ago a screenreader user was begging people to stop putting redundant title tags on everything because it clogged the audio.

dani97: sure will restore @jimbo @brendanfalkowski

packages/venia-concept/src/components/ProductOptions/option.js
dani97: Yes, Thanks for pointing that out.

packages/venia-concept/src/components/ProductOptions/option.js
dani97: Thanks. Credits to @sirugh. Now I have got understanding of useMemo hook  :)

sirugh:
@dani97 I am going to take over this PR since we haven't heard from you in a while. I'll be resolving the conflict files and addressing any remaining feedback.

dani97:
Thanks @sirugh . I couldn't contribute for over a month. I hope I will join you soon :)

PR #1531, opened 2019-08-06
Coveralls Test PR

supernova-at:
Just to see if Coveralls is working.


supernova-at:
The badge on the main `README.md` still says "coverage unknown" but that may be a data propagation issue that will resolve itself. The coverage data is being sent to and received successfully by Coveralls.

Issue #422, opened 2018-10-25
[Bug] Handle title and meta tags for navigation on client side


Starotitorov:
## This issue is for the following packages:

[x] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-devdocs`
[ ] `upward-js`
[ ] `upward-spec`

## This issue is a:

[x] Bug
[ ] Feature suggestion
[ ] Documentation issue
[ ] Other (Please Specify)

### Environment

| Question                    |    Answer     |
| --------------------------- | :-----------: |
| Magento version             | 2.3  |
| Operating System + version  | Ubuntu 18.04  |
| node.js version (`node -v`) | 8.12.0  |
| npm version (`npm -v`)      | 6.4.1  |

## Description

For example, open any product page _/augusta-earrings.html_ in the application and go to home page using logo link. You can see that title of the page is still _Augusta Earrings -_ and HTML head tag still contains following:

```
<html>
  <head>
    ...
    <title>Augusta Earrings - </title>
    <meta name="title" content="Augusta Earrings">
    <meta name="keywords" content="">
    <meta name="description" content="<p>Silver &amp;amp; aqua chandelier earrings.</p><ul><li>2.5&quot; length</li><li>Sterling silver</li><li>Fish hook style</li></ul>">
    ...
  </head>
  <body>
  </body>
</html>
```

This issue is not related only to this pages, it is a general issue of the application.

@ericerway, @zetlen, please, verify this.

## Expected result:

Title of the home page should be _Venia_. In html head tag we expect to have just:

```
<html>
  <head>
    ...
    <title>Venia</title>
    ...
  </head>
  <body>
  </body>
</html>
```

## Possible solutions:

Right now we do not update page title and meta tags inside head tags when we handle navigation on client side using react-router. We can use react-helmet to manage title and meta tags to the document head. https://github.com/nfl/react-helmet#readme

zetlen:
Thanks, @Starotitorov! We de-prioritized this because the title and metadata are most important on initial load, for SEO and SMO, but it's still confusing when the user sees the browser title. Definitely worth looking at.

vasilii-b:
Hi! I'd like to have a look on the issue. Can someone assign it to me, please? Thank you!

awilcoxa:
Moving ticket forward as the issues is still relevant. The previously proposed solution (below) was closed, but the team will re-investigate additional solutions. 

~~Possible solutions:
[...]

PR #1537, opened 2019-08-07
Head and meta tags handling

revanth0212:
Added wrappers around the `HeadProvider` and `Title` components of `react-head`. Also exporting `Meta`, `Link` and `Style` tags directly from `react-head`.

These 5 components will be used to handle the document's `head` tag.

To be able to use the components anywhere in the app, the app has to be wrapped inside `HeadProvider` which provides the context needed to maintain all the rendered tags.

Closes [422](https://github.com/magento/pwa-studio/issues/422)


packages/venia-concept/src/components/Head/headProvider.js
```diff
@@ -0,0 +1,19 @@
+import React, { useEffect } from 'react';
+import { HeadProvider as _HeadProvider } from 'react-head';
+
+const filterOutTitleTags = titleTags => {
+    titleTags
+        .filter(titleNode => !titleNode.dataset.synteticTag)
```
brendanfalkowski: synteticTag or syntheticTag?

packages/venia-concept/src/components/Head/title.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Title as _Title } from 'react-head';
+
+export default props => (
+    <_Title {...props} data-syntetictag>
```
brendanfalkowski: syntetictag or synthetictag?

packages/venia-concept/src/components/Head/headProvider.js
```diff
@@ -0,0 +1,19 @@
...
+        .forEach(titleTag => {
+            if (titleTag) titleTag.parentNode.removeChild(titleTag);
```
supernova-at: Did this pass `prettier` formatting?

`yarn run prettier`

packages/venia-concept/src/components/Head/headProvider.js
```diff
@@ -0,0 +1,19 @@
...
+        });
+};
+
+export default props => {
```
supernova-at: For consistency in the codebase let's give this function a semantic name and export it separately.

```js
const HeadProvider = props => { ... }

export default HeadProvider;
```

packages/venia-concept/src/components/Head/headProvider.js
revanth0212: My bad. Spelling mistake.

supernova-at: Can you drop a comment here to explain why we're removing some and not all `title` tags, and how `syntheticTag` plays into that?

packages/venia-concept/src/components/Head/headProvider.js
```diff
@@ -0,0 +1,19 @@
+import React, { useEffect } from 'react';
+import { HeadProvider as _HeadProvider } from 'react-head';
```
supernova-at: I'm not a huge fan of the underscore as the first character of a React component 😬

packages/venia-concept/package.json
```diff
@@ -73,6 +74,7 @@
     "@magento/pwa-buildpack": "~3.0.0",
     "@magento/upward-js": "~2.2.0",
     "@storybook/react": "~4.1.13",
+    "@testing-library/react": "~8.0.8",
```
revanth0212: Using `@testing-library/react` for DOM testing since the test renderer that comes with react renders components all in react as JS objects. Needed something like JSDom to assert head tags. Let me know if we can achieve this without `@testing-library/react`, will be happy to change.

FYI, the API `@testing-library/react` provides is very intuitive and can be used for other tests in the future.

packages/venia-concept/src/components/Head/title.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Title as _Title } from 'react-head';
+
+export default props => (
+    <_Title {...props} data-synthetictag>
```
supernova-at: Oh, so we're adding the synthetic tag? 

Why does `HeadProvider` need to filter them out then? How could they exist before `HeadProvider` is mounted?

packages/venia-concept/src/components/Head/title.js
```diff
@@ -0,0 +1,8 @@
...
+        {props.children}
+    </_Title>
```
supernova-at: Still not a fan of the underscore 😝

packages/venia-concept/src/components/Head/title.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Title as _Title } from 'react-head';
+
+export default props => (
```
supernova-at: For consistency in the codebase please also define and export this function separately.

packages/venia-concept/package.json
supernova-at: Oh, I assumed @jimbo steered you in this direction. We probably want his sign off on this one.

supernova-at:
https://github.com/magento/pwa-studio/pull/1537#issuecomment-519243093. Please fill out the PR template :)

Looks like there are some conflicts to resolve as well 👍

supernova-at:
I also have a small concern about stale meta tags hanging around.

If Page2 provides the same meta tag as Page1 with a different value, it'll get overwritten properly. But what if Page1 has a specific meta tag that Page2 doesn't care / know about? Won't that just stick around forever?

packages/venia-concept/src/components/Head/headProvider.js
revanth0212: So the problem with `react-head` is, it won't render a `title` tag if there is one already present in the markup. So according to their docs, it is required to remove the existing tags and then render the needed ones using `react-head`. Now we should not be removing the tags that were created using `react-head` in case of a re-render. That is where the `synthetictag` flag comes in. If the `syntheticTag` flag is set, it means that title tag was created using `react-head` and should not be removed.

packages/venia-concept/src/components/Head/headProvider.js
revanth0212: Yeah, it did. Does it mean I am running a different prettier config than yours?

packages/venia-concept/src/components/Head/title.js
revanth0212: Explained in the above comment.

packages/venia-concept/src/components/Head/headProvider.js
revanth0212: What shall I change it to? I wanted to signify vanila `HeadProvider` as` _HeadProvider`.

revanth0212:
> I also have a small concern about stale meta tags hanging around.
> 
> If Page2 provides the same meta tag as Page1 with a different value, it'll get overwritten properly. But what if Page1 has a specific meta tag that Page2 doesn't care / know about? Won't that just stick around forever?

If the new page is a new HTML it shouldn't unless the new markup has that meta tag. If it a page transition using react-router, then it is on us to clear those tags. As of today, `react-head` does not support the manipulations of tags (like removing). I have checked their codebase, it is possible for us to go ahead and add manipulations API as submit a PR, or the simplest would be to write a simple wrapper around `Meta` (just like how we did for `Title`) that does these kinds of manipulations.

The ticket dint mention those use cases but you are right, they will be hanging around and there has to be a way to delete the existing ones. Shall I go ahead and write that wrapper? What do you say @jimbo @supernova-at?

packages/venia-concept/src/components/Head/__tests__/headProvider.spec.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import TestRenderer from 'react-test-renderer';
+
+import { HeadProvider, Title, Link, Meta, Style } from '../';
+
+afterEach(cleanup);
+
+test('snapshot testing HeadProvider', () => {
```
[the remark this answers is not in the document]
revanth0212: Not sure what to call this. It is a casual snapshot test.

packages/venia-concept/package.json
revanth0212: This is on me. Jimmy dint ask me to use this package. But if you guys have any other solution to render components on JSDOM or similar DOM implementations without a new library, let me know.

packages/venia-concept/src/components/Head/__tests__/headProvider.spec.js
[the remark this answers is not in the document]
supernova-at: "HeadProvider by itself doesn't render anything" ?

Basically you just want someone who is only looking at the snapshot to be able to ensure it is correct.

packages/venia-concept/src/components/Head/headProvider.js
supernova-at: No, you're right - it worked for me too. Just haven't seen very many of these I guess 🤷‍♂

supernova-at:
> If it a page transition using react-router, then it is on us to clear those tags.

That is the case when navigating within the app, yes.

> Shall I go ahead and write that wrapper?

I think we do need it.

packages/venia-concept/src/components/Head/headProvider.js
supernova-at: > the problem with react-head is, it won't render a title tag if there is one already present in the markup.

Isn't that what [this test](https://github.com/magento/pwa-studio/pull/1537/files#diff-6ad3f434d87e531558aa3b745f33c3ecR40) does though? Or does that test only work because you added the synthetic tag handling?

supernova-at: I meant drop a comment in the code itself so that in the future no one removes this line thinking that it's not needed 😛

packages/venia-concept/package.json
supernova-at: I think this is the recommended approach, I was just surprised we didn't already have it. I'm 👍

packages/venia-concept/src/components/Head/headProvider.js
```diff
@@ -0,0 +1,21 @@
...
+        .forEach(titleTag => {
+            if (titleTag) titleTag.parentNode.removeChild(titleTag);
+        });
+};
+
+const HeadProvider = props => {
+    useEffect(() => {
+        const titleTagsToRemove = document.getElementsByTagName('title');
+        filterOutTitleTags([...titleTagsToRemove]);
+    }, []);
```
supernova-at: I'm still hoping we can simplify this 😄 

Because we're passing an empty array as the second argument to `useEffect`, it will run on mount and unmount only:

> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the dependencies array always works.

Since our `Title` component doesn't do anything unless it's wrapped in this `HeadProvider` component, there can never be a case where the `Title` component has inserted a `<title>` that the `HeadProvider` wants to skip removing, right?

If I'm understanding all that correctly, then there is no need for the synthetic tag data attribute, and therefore no need to filter for them here - just removing all `<title>` tags should be sufficient 👍

revanth0212: You are right. But that was added in place to handle the server-rendered output. If the server-rendered output already has a `synthetictag` on it, we don't want to delete it.

But after talking to you, I have realized that is a bleak use-case. We can probably get rid of the filtering for now and add this logic when needed.

packages/venia-concept/src/components/Head/headProvider.js
revanth0212: Interesting. Let me get this done.

brendanfalkowski: @supernova-at The pattern of naming things `VeniaHeadProvider` after the package is kind of frustrating in practice. These components get used in implementations that aren't identified as Venia anymore and that naming doesn't feel good to see, so we rewrite and maintain that rewrite. It makes upstream integration more complex.

revanth0212:
> Looks straightforward. Remove `@testing-library/react` and fix `yarn.lock` and this is good to go. 👍

Done. Refactored tests to not use that library.

dpatil-magento:
@revanth0212  I don't see original bug has fixed. 
1. Click on any category say - "Tops"
expected - Title should say "Tops - Venia"
actual - Title does not change , its still - "Home Page - Venia"

2. Go to ant PDP say /valeria-two-layer-tank.html
expected - title should say "Valeria Two-Layer Tank - Venia"
actual - "Home Page - Venia"
Note - After page reload - "Valeria Two-Layer Tank - "  `venia` text still missing.

revanth0212:
Dev is right. We are able to reproduce this issue using his steps.

image

Will look into the issue.

packages/venia-ui/upward.yml
```diff
@@ -100,6 +100,10 @@ veniaAppShell:
         urlResolver: veniaUrlResolver
         env: env
         mediaBackendURL: mediaBackendURL
+        site:
+          inline:
+            name:
+              inline: Venia
```
supernova-at: Why are these changes needed? I'm assuming this is just changing the title to "Venia" whenever it is supplied by Upward?

supernova-at: Did you not also have to change any part of the templates themselves?

revanth0212: This is needed for the product template that needs `site.name`. Others don't as of now. I can create a new PR that changes other templates as well to use `site.name`.

supernova-at:
This looks great! One thing I noticed while navigating:

Sometimes while it is loading a new route, the title of the tab changes to `Home Page - Venia`. If we can't get that out of there, might we consider changing the title in `app.js` to just be "Venia"?

revanth0212:
> This looks great! One thing I noticed while navigating:
> 
> Sometimes while it is loading a new route, the title of the tab changes to `Home Page - Venia`. If we can't get that out of there, might we consider changing the title in `app.js` to just be "Venia"?

Sounds good to me, but there was a reason why I went with `Home Page - Venia`. The UPWARD server outputs HTML that has `Home page - Venia` (small P), so if we don't have the same title, for a brief second you'll see `Home page - Venia` from UPWARD and then once our app loads it will change to `Venia` and then to the product's title. 

We will also need to change the UPWARD server's output for the home page to have `P` not `p` to be consistent.

dpatil-magento:
@revanth0212 Please check unit test failures. Able to reproduce on my local as well.
    packages/venia-ui/lib/components/App/__tests__/app.spec.js
    packages/venia-ui/lib/RootComponents/Category/__tests__/categoryContent.spec.js

jimbo:
This is ready to go pending QA approval.

PR #1191, opened 2019-05-02
Refactor left drawer

jimbo:
## Description

Refactor the left drawer to use hooks, especially the navigation.

## Related Issue

- Closes #972
- Closes #977

## Proposed Labels for Change Type/Package
- [x] major (e.g x.0.0 - a breaking change)
- [ ] minor (e.g 0.x.0 - a backwards compatible addition)
- [ ] patch (e.g 0.0.x - a bug fix)

## Checklist:
- [ ] I have updated the documentation accordingly, if necessary.
- [x] I have added tests to cover my changes, if necessary.

PR #1552, opened 2019-08-12
Refactor left drawer

### LINKED ISSUE #972: Peregrine: Menu Component
Refactor the menu within `Navigation`.

- [ ] Refactor class components to functional ones
- [ ] Use the `useQuery` hook to fetch data
- [ ] Fix queries to fetch *all* categories
- [ ] Treat each leaf as a potential branch, since `children_count` is unsupported

Ideally we'd get the full tree of categories in one batch query, then translate that data into a tree of elements within the component. Unfortunately, that isn't supported yet, and the nested queries have serious performance issues. Further, the current implementation doesn't fetch some categories.

jimbo:
## Description
Rewrite the workflows within the left drawer to fix a variety of issues and clean up the UI.

## Related Issue
Replaces #1191.
Closes #972 

## Verification Steps
1. Open the left drawer.
2. Navigate to a category.
3. Sign in.
4. Create an account.
5. Submit the forgot password form.

## Checklist:


packages/venia-ui/lib/components/CreateAccount/createAccount.js
```diff
@@ -82,10 +84,7 @@ class CreateAccount extends Component {
                 initialValues={initialValues}
                 onSubmit={handleSubmit}
             >
-                <h3 className={classes.lead}>
-                    {`Check out faster, use multiple addresses, track
-                         orders and more by creating an account!`}
-                </h3>
+                <h3 className={classes.lead}>{LEAD}</h3>
```
sirugh: We probably don't need this to be a header anymore especially since we set line and font size. I would reserve it for a11y purposes.

packages/venia-ui/lib/components/ForgotPassword/forgotPassword.js
```diff
@@ -1,61 +1,64 @@
...
+const INSTRUCTIONS = 'Enter your email below to receive a password reset link.';
 
+const ForgotPassword = props => {
+    const {
+        completePasswordReset,
+        email,
+        initialValues,
+        isInProgress,
+        resetPassword,
+        onClose
```
sirugh: This is a `noop` now. Can we just remove it entirely?

packages/venia-ui/lib/components/ForgotPassword/forgotPassword.js
```diff
@@ -1,61 +1,64 @@
...
-    handleContinue = () => {
-        const { completePasswordReset, email, onClose } = this.props;
+const INSTRUCTIONS = 'Enter your email below to receive a password reset link.';
 
+const ForgotPassword = props => {
+    const {
+        completePasswordReset,
+        email,
+        initialValues,
+        isInProgress,
```
sirugh: I think there is room to improve this workflow. Right now we set `isInProgress` to true when you submit the form and then set it to false when you click the "Continue Shopping" button. I think this UX is strange. We should probably close the menu entirely (since that's what "Continue Shopping" would do, I imagine) or we should change the button to say `Back` and then present the user with the navigation menu, sign in page, or something else. I think we could probably pass one of the `show` callbacks from navigation.js down to the forgotPassword form and use that when the user submits/clicks back.

[Image from Gyazo](https://gyazo.com/edaa580ebd26cdf1a5cb68955d79cf62)

packages/venia-ui/lib/components/MyAccount/accountLink.js
```diff
@@ -0,0 +1,42 @@
+import React, { useCallback } from 'react';
+import { arrayOf, func, node, shape, string } from 'prop-types';
+
+import { mergeClasses } from '../../classify';
+import Button from '../Button';
+
+import defaultClasses from './accountLink.css';
+
+const AccountLink = props => {
```
sirugh: Nice 👍

packages/venia-ui/lib/components/MyAccount/myAccount.js
```diff
@@ -0,0 +1,57 @@
...
+import { func, shape, string } from 'prop-types';
+
+import { mergeClasses } from '../../classify';
+import AccountLink from './accountLink';
+import defaultClasses from './myAccount.css';
+
+const MyAccount = props => {
+    const { signOut, user } = props;
+    const { email, firstname, lastname } = user;
+    const name = `${firstname} ${lastname}`.trim() || 'My Account';
```
sirugh: These string defaults could be made constants.

packages/venia-ui/lib/components/ForgotPassword/forgotPassword.js
jimbo: Good catch. You're right, `Continue Shopping` should reset the view state (back to `MENU`) and close the drawer. The wording of the button is design's preference, so I don't think I'll change it.

packages/venia-ui/lib/components/ForgotPassword/forgotPassword.js
jimbo: Yes, I'll either use it or remove it.

packages/venia-ui/lib/components/CreateAccount/createAccount.js
jimbo: Heading. 😉 

I agree, though; semantically this should just be a paragraph.

packages/venia-ui/lib/components/Navigation/container.js
```diff
@@ -1,40 +1,84 @@
...
-    };
+/**
+ * This file does more than connect to the Redux store.
+ *
+ * It also bridges the gap between the current architecture and
+ * the upcoming one in which app state will be available via
+ * context.
+ */
+
+// TODO: get these contexts from peregrine instead
```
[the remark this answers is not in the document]
jimbo: Yes, I'll do that. These are obviously the app state contexts that will be global.

packages/venia-ui/lib/reducers/catalog.js
```diff
@@ -1,29 +1,83 @@
...
+        for (const child of children) {
+            childMap.set(child.id, {
+                ...child,
+                ...(state.categories[child.id] || {}),
+                parentId: id
+            });
+        }
+
+        // merge in the fetched child last
+        // TODO: use `children_count` from graphql once it's accurate
```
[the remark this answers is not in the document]
jimbo: I don't know of one. We could look, though.

packages/venia-ui/lib/components/Navigation/navigation.js
```diff
@@ -1,327 +1,129 @@
...
+import { shape, string } from 'prop-types';
+
+import { mergeClasses } from '../../classify';
+import AuthBar from '../AuthBar';
+import AuthModal from '../AuthModal';
+import CategoryTree from '../CategoryTree';
+import { AppContext, CatalogContext, UserContext } from './container';
 import NavHeader from './navHeader';
 import defaultClasses from './navigation.css';
-import { MyAccountMenuTrigger } from '../MyAccountMenuPage';
```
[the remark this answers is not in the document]
jimbo: Yes, I need to do that.

packages/venia-ui/lib/components/AuthBar/userChip.js
```diff
@@ -0,0 +1,54 @@
...
+            <span className={classes.content}>
+                <span className={classes.avatar}>
+                    <Icon src={UserIcon} />
+                </span>
+                <span className={classes.user}>
+                    <span className={classes.fullName}>{display}</span>
+                    <span className={classes.email}>{email}</span>
+                </span>
+                <span className={classes.icon}>
+                    <Icon src={ChevronUpIcon} />
```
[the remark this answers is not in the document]
jimbo: I had the same thought.

Originally, the user chip slid up and we faked its persistence by rendering another copy of it in the `MyAccount` view. That added a lot of complexity to both the code and the styles for an experience that ended up looking tacky.

Sliding new views in from the side and ditching the user chip in `MyAccount` avoids all of that, but it does make the `ChevronUp` kind of inaccurate. I'll change it.

packages/venia-ui/lib/components/Navigation/navHeader.js
```diff
@@ -1,38 +1,54 @@
...
-                </h2>
-                <Trigger key="closeButton" action={onClose}>
-                    <Icon src={CloseIcon} />
-                </Trigger>
-            </Fragment>
-        );
-    }
-}
+    const title = titles[view] || titles.MENU;
+    const backIcon = isTopLevel && view === 'MENU' ? MenuIcon : ArrowLeftIcon;
```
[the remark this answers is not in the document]
jimbo: This is a tricky problem.

- If we have a back arrow at the root, then while you're clicking back from nested categories, you're bound to click it and close the drawer inadvertently before you realize you're at the root.
- If we change it to a hamburger icon, it becomes clear that you're reached the root and the only thing it could reasonably do is close the drawer. Least surprise, for sure, but redundant with the close icon.
- If we remove the button and preserve the space, the blank space looks wrong.
- If we remove the button and don't preserve the space, the title moves as you navigate.

Other options:

- When the drawer is open and at the root, change the back arrow to a close icon and hide the righthand close icon.
- When the drawer is open and at the root, show a small, noninteractive Venia logo in place of the back arrow.

packages/venia-ui/lib/components/Navigation/navigation.css
```diff
@@ -1,22 +1,50 @@
+/* animation helpers */
+
+.enter {
+    transition-duration: 384ms;
```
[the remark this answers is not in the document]
jimbo: From Material's animation guidelines, which differentiate between `enter` and `exit` animations. In a nutshell: enter softly, leave briskly.

In general, the duration for a view entering the screen should default to `384ms`, and the duration for that same view exiting the screen should be `192ms`. They also use different timing functions.

packages/peregrine/package.json
```diff
@@ -18,18 +18,18 @@
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/peregrine#readme",
   "devDependencies": {
     "intl": "~1.2.5",
-    "react": "~16.8.3",
+    "react": "~16.9.0",
```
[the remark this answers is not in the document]
jimbo: Yeah, lots of version differences going on in these `package.json` files. Need to unify them.

If you want to bump them to `16.9.0` outside of this PR (and merge it beforehand) I'd be fine with it.

packages/venia-ui/lib/components/CreateAccount/createAccount.js
sirugh: You're heading for trouble!

Cool thanks.

packages/venia-ui/lib/components/AuthBar/userChip.js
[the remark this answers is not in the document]
sirugh: Thanks for the explanation! I think your code makes it much cleaner.

packages/venia-ui/lib/components/Navigation/navHeader.js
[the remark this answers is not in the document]
sirugh: I think we need @soumya-ashok's input but she is traveling. I say leave it as is for now and open an issue with `needs-ux`.

packages/peregrine/package.json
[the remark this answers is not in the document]
sirugh: https://github.com/magento/pwa-studio/pull/1564

sirugh:
A minor CSS issue I just noticed but on desktop when you click the menu button the page shifts slightly to the right.

packages/venia-ui/lib/components/ForgotPassword/forgotPassword.js
jimbo: This has been addressed.

packages/venia-ui/lib/components/MyAccount/myAccount.js
jimbo: This has been addressed.

packages/venia-ui/lib/components/ForgotPassword/forgotPassword.js
jimbo: This has been addressed; it's no longer a noop.

packages/venia-ui/lib/components/AuthBar/userChip.js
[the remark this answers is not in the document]
jimbo: This has been addressed.

packages/venia-ui/lib/components/CreateAccount/createAccount.js
jimbo: This has been addressed.

dpatil-magento:
@jimbo Please have look on these observations :
1. Create account flow does not work from https://pr-1552.pwa-venia-staging.com/create-account.
2. On smaller screen devices/browsers, left drawer scrolling is broken. Does not scroll at all.  Try on left drawer Create account form.
3. On `develop` branch after sign-in "Category" menu loads but on `jimbo/left-drawer` its "My Account" . Is this a intentional change?

sirugh:
> Create account flow does not work from https://pr-1552.pwa-venia-staging.com/create-account.

Weird. I didn't check this, didn't think it would even be related as this PR related to the left drawer.

> On smaller screen devices/browsers, left drawer scrolling is broken. Does not scroll at all. Try on left drawer Create account form.

Oops. Good catch. I was using tabbing on my simulator :/

> On develop branch after sign-in "Category" menu loads but on jimbo/left-drawer its "My Account" . Is this a intentional change?

Yes, this was updated intentionally.

jimbo:
@dpatil-magento Thanks for the thorough QA on this one. I definitely overlooked the `/create-account` edge case.

1. Create account flow should work in both cases now, including showing a static error message if necessary.
2. The nav modal should now handle overflow (scroll) if necessary.
3. Yes, the drawer title should accurately reflect the current drawer view. Note that it doesn't attempt to reflect the current category, though; it should still just say "Main Menu" whenever the category tree is the active view.

dpatil-magento:
Thanks @jimbo @sirugh  Looks good now. 

Last observation - Category “Dresses” does not have any child categories but still it goes a level down and show “All Dresses” sub category.
Is it because GraphQL does not know Parent category has Child or not?

dpatil-magento:
Fix looks good. QA Pass, good to merge.

PR #1556, opened 2019-08-13
Cart Counter

### LINKED ISSUE #1412: [bug]: CartCounter is not persistent after page reload
**Describe the bug**
When a users adds a product to the cart the item counter gets updated in the header. But when the user refreshes the page (cntrl+r) the counter is not displayed anymore. 

**To Reproduce**
1. Add product to cart
2. Refresh page with cntrl + r
3. See cart icon there should be a item count

**Expected behavior**
We expect the counter to be persistent even after a hard page reload

**Additional context**
It seems to be broken after the 3.0.0 release. When i check  https://veniapwa.com that i believe to still run on 2.1.0 the cart is persistent.

 - Device [Mac]:
 - OS [IOS 10.14.5]:
 - Browser [Chrome]:
 - Browser Version [76.0.3809.46]:
 - Magento Version: 2.3.2
 - PWA Studio Version: 3.0.0
 - NPM version `6.5.0`:
 - Node Version `v11.9.0`:

- [x] `venia-concept`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

supernova-at:
## Description

The problem was that we were never requesting the cart's details on page load. Only when adding an item to the cart, updating an item in the cart, removing an item from the cart, or toggling the cart open did we request the cart's details.

Now we do.

This PR also updates components in the app's header:
  * `CartTrigger` calls `getCartDetails` when it gets mounted
  * `CartTrigger` is no longer a connected component
  * `Header/container.js` is the connected component, passing info down to `CartTrigger` via `header.js`
  * `CartTrigger` and `CartCounter` rewritten to be functional components
  * `CartTrigger` and `CartCounter` tests updated away from using `enzyme`

## Related Issue

Closes #1412.

## Verification Steps

1. Load up Venia's homepage
1. If you have an item in your cart, see that the shopping cart shows the number of items correctly
1. If you do not have an item in your cart, add some and go back to step 1 😄 

Screen Shot 2019-08-13 at 4 30 29 PM

## Checklist:


packages/venia-ui/lib/components/Header/cartTrigger.js
```diff
@@ -1,75 +1,67 @@
...
-                onClick={toggleCart}
-            >
-                {cartIcon}
-                <CartCounter counter={itemsQty ? itemsQty : 0} />
-            </button>
-        );
-    }
-}
+    useEffect(() => {
+        if (getCartDetails) {
```
sirugh: Is this a necessary conditional? Seems like the function should always be present.

packages/venia-ui/lib/components/Header/cartTrigger.js
```diff
@@ -1,75 +1,67 @@
...
+            getCartDetails();
+        }
+    }, [getCartDetails]);
+
+    const cartIcon = useMemo(() => {
+        return getCartIcon(cartDetails);
```
sirugh: Minor improvement would be to pull off the `items_qty` from `cartDetails` as it seems like the only value the function cares about.

Beyond that we could also just hardcode the two possible icon states, one that has color and one that is filled and one that isn't.

packages/venia-ui/lib/components/Header/container.js
```diff
@@ -1,17 +1,20 @@
 import { connect } from '@magento/venia-drivers';
 
-import Header from './header';
 import { toggleSearch } from '../../actions/app';
+import { getCartDetails, toggleCart } from '../../actions/cart';
```
sirugh: How much work would it be to write a new query/action that only requests the item count? I know the efficiency of our gql calls has been under the microscope lately so maybe a smaller query would help, especially for something that shows up on the header/initial load.

packages/venia-ui/lib/components/Header/cartTrigger.js
```diff
@@ -1,75 +1,67 @@
...
 
-        if (itemsQty > 0) {
-            svgAttributes.fill = iconColor;
-        }
+    return <Icon src={ShoppingCartIcon} attrs={svgAttributes} />;
+};
 
-        return <Icon src={ShoppingCartIcon} attrs={svgAttributes} />;
-    }
+const Trigger = props => {
```
sirugh: Can we match this with the name of the file? There's already another `Trigger` or two in this repo :D

packages/venia-ui/lib/components/Header/cartCounter.js
```diff
@@ -1,23 +1,27 @@
...
+    const { numItems } = props;
 
-    render() {
-        const { counter, classes } = this.props;
-        return counter > 0 ? (
-            <span className={classes.root}>{counter}</span>
-        ) : null;
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    // Don't show anything both when we don't have data to show, and when numItems is zero (0).
```
sirugh: It appears that we used to _always_ show either 0 or the number of items. The conditional where we created this component passed the `item_qty` or `0`. This changes that behavior so I just want to call that out to ensure it was intentional.

packages/venia-ui/lib/components/Header/cartTrigger.js
supernova-at: Yeah I wasn't really sure about this. It will always be present because I connected it that way, but I was thinking if someone else uses the `Header` component they may or may not connect it.

Marking this prop type as `.isRequired` is along these same lines.

I didn't mark it as `isRequired`, so I included this conditional.

supernova-at: `Header` is [available as an export](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/Header/index.js#L1) from `Header/index.js`.

packages/venia-ui/lib/components/Header/cartCounter.js
supernova-at: I don't think that has changed, `cartTrigger` had a [conditional in its render function](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/Header/cartCounter.js#L17-L19).

It shouldn't have ever displayed a zero.

packages/venia-ui/lib/components/Header/cartTrigger.js
sirugh: But `Header` passes the function now, so it _should_ always be defined.

I'd probably just mark it as `isRequired` and be done with.

packages/venia-ui/lib/components/Header/cartCounter.js
sirugh: Oh you're right! My bad :)

packages/venia-ui/lib/components/Header/cartTrigger.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1556/commits/ced668878ed9b61219559142c28ecbfc2bff75d4.

packages/venia-ui/lib/components/Header/cartTrigger.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1556/commits/ced668878ed9b61219559142c28ecbfc2bff75d4.

packages/venia-ui/lib/components/Header/container.js
supernova-at: I thought this was a great suggestion but unfortunately implementing it uncovered [a bug in our GraphQL](https://github.com/magento/graphql-ce/issues/830) where the item quantity is not being returned correctly.

sirugh: 😢

packages/venia-ui/lib/components/Header/cartTrigger.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1556/commits/8bb2e7e3ff3feadfb28dbde4400f2a98d372536d.

supernova-at:
I changed the recommended version to be `Minor` because according to our [Public API](https://github.com/magento/pwa-studio/blob/37cd2c18247eeb773677c960fa9620d2f98161fa/pwa-devdocs/src/technologies/versioning/index.md#public-api) policy, `Header/header.js` is part of the public API (it gets exported by [Header/index.js](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/Header/index.js#L1) ), and I added additional props to it.

dpatil-magento:
@supernova-at There is a unit test failure, can you please take a look?
packages/venia-ui/lib/components/Header/__tests__/cartTrigger.spec.js

supernova-at:
> @supernova-at There is a unit test failure, can you please take a look?
packages/venia-ui/lib/components/Header/tests/cartTrigger.spec.js

@dpatil-magento oops, forgot to update a snapshot. All set now, thanks!

dpatil-magento:
@supernova-at Fix looks good. Just curious to know why it takes ~10 seconds to show the count on page reload?

supernova-at:
> Just curious to know why it takes ~10 seconds to show the count on page reload?

It definitely should not!

PR #1582, opened 2019-08-20
[WIP] List Initial Selections

### LINKED ISSUE #900: Cart Edit screen should allow just to edit and update single parameter ( ex: qty) also for configurable product.
Description:
After adding configurable product to cart > Edit Item >  if user updates only qty and hit update cart button then qty is not getting updated.  Edit item feature only works if user updates all configurable parameters like color, size and qty.

1. Go to Venia storefront and a add product to cart.
2. Now click on edit item option present under more options (3 vertical dots).
3. Update qty and click on Update Cart button.

Expected behavior:
Shopping Cart right drawer should be displayed with updated qty.

Actual:
Shopping cart doesn't get updated with new qty.

Note - If user updates all configurable parameters like color, size and qty and then clicks on Update Cart button then cart is updated.

Coalescing all the acceptance criteria from the comments below:
- [ ] Existing configured options are pre-selected in the edit item panel
- [ ] A toast that displays as soon as the user hit "update cart" reading "Cart item updated" or similar

supernova-at:
## Description

TBD

### Discovery

When we get the cart's details, we assign `options` to each `item` based on the information contained in the `totals` resource [here in cart actions](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/actions/cart/asyncActions.js#L350-L372).

An individual `item`'s `options` is an array that looks like
```js
[
  {value: "Lilac", label: "Fashion Color"}
  {value: "2", label: "Fashion Size"}
]
```

Unfortunately, when we make a selection we are provided only with an [optionId and selection](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/MiniCart/cartOptions.js#L55), which looks like:
```js
{
  optionId: "176",
  selection: "Rain"
}
```

There is a missing piece that informs the system that `optionId` "176" (in this case) is "Fashion Color".

It should be noted that we do get this "missing piece" when we [query for the product's details](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/MiniCart/editItem.js#L26) to populate the edit cart screen.

## Related Issue

Closes #900.

## Verification Steps

TBD

## Checklist

## TODO

- [x] Turn all the product / variant matching stuff into utility functions so no one else has to feel this pain
- [x] `List/Item.js` focus handling
- [ ] Does passing in an array as an `initialSelection` to `List` work?
- [ ] Should the `List` control the `selection` state instead of `List/Items` ?
- [x] Remove `useListState`
- [x] Update unit tests


supernova-at:
This initial WIP PR has been broken up into and replaced by #1589 and #1592.

PR #1586, opened 2019-08-21
Changed prop type of uniqueID for Item component.

revanth0212:
## Description

uniqueID prop check was failing since it was mentioned that uniqueID will be a number but in case of `category` component which internally uses `item`, it can be a string as well.

So no more errors like this:

image

## Related Issue
None. No issue for this. It is just a dev fix.
Closes nothing.

## Verification Steps
No more console errors related to prop types.

## Checklist
Check the console for prop types errors in category and product pages.

### DIFF
--- a/packages/peregrine/lib/List/item.js
+++ b/packages/peregrine/lib/List/item.js
@@ -54,7 +54,8 @@ Item.propTypes = {
     item: PropTypes.any.isRequired,
     itemIndex: PropTypes.number.isRequired,
     render: PropTypes.oneOfType([PropTypes.func, PropTypes.string]).isRequired,
-    uniqueID: PropTypes.number
+    uniqueID: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
+        .isRequired
 };
 
 Item.defaultProps = {


packages/peregrine/lib/List/item.js
```diff
@@ -54,7 +54,8 @@ Item.propTypes = {
     item: PropTypes.any.isRequired,
     itemIndex: PropTypes.number.isRequired,
     render: PropTypes.oneOfType([PropTypes.func, PropTypes.string]).isRequired,
-    uniqueID: PropTypes.number
+    uniqueID: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
```
sirugh: Are the uses of `uniqueID` within this component capable of handling either?

sirugh: I think this uses a set downstream which means we need to make sure to use a single type.

revanth0212: Yes, we just use it as key for a component. Since react does not allow users to send a prop called key to its children, I am just passing it down as uniqueID. Once it reaches the `Item` component, it is used as `key` which can be `number` or `string`.

sirugh:
Thanks for fixing this :)

PR #1592, opened 2019-08-22
Edit Item uses Peregrine List initialSelection

### LINKED ISSUE #1634: Use List initial selections in Venia Edit Item
Now that Peregrine's `List` component can support a set of initial selections, have Venia take advantage of that functionality in Minicart > Edit Item.

Configurable items should have their options pre-selected when the Edit Item screen is opened.

supernova-at:
## Description

This PR is the second step in solving #900 : After [Peregrine's List component is updated](https://github.com/magento/pwa-studio/pull/1589) to support an `initialSelection` prop, this PR takes advantage of that in the `Edit Item` screen to pre-populate the color and sizes lists with the cart item's values:

Screen Shot 2019-08-22 at 4 13 17 PM

## Related Issue

Closes #1634 .

### Configurable Items
1. Add a configurable item to the cart
1. Take note of the option selections you made (color, size)
1. Open the MiniCart
1. Edit the item
1. See that the Edit Item screen options are pre-selected (and are correct)
1. Change some options
1. Ensure you can save successfully and the item is updated in the cart

### Simple Items
1. Add a simple item to the cart
1. Open the MiniCart
1. Edit the Item
1. See that no errors occur
1. Ensure you can save successfully and the item is updated in the cart

## Checklist

## Note

This PR is based on #1589 and should be rebased once that one gets merged.
I have applied the `hold` label to indicate this.


packages/venia-ui/lib/components/MiniCart/cartOptions.js
```diff
@@ -46,19 +47,50 @@ const CartOptions = props => {
     } = props;
     const { name, price, qty } = cartItem;
 
+    // cartItem.options[0] = { value: "Lilac", label: "Fashion Color" }
+    // configItem.configurable_options[0] = attribute_code ("fashion_color"), attribute_id (176), id, label ("Fashion Color"), values[]
+    //   values[0] = default_label, label, store_label, use_default_value, value_index
```
supernova-at: These comments can be removed.

packages/venia-ui/lib/components/MiniCart/editItem.js
```diff
@@ -38,7 +38,10 @@ const EditItem = props => {
             return;
         }
 
-        const fetchItemOptions = async () => {
+        // We know the item has options, but we don't know the
+        // other possible values those options could take on.
+        // Fetch them now.
+        const fetchItemVariants = async () => {
```
supernova-at: I renamed this function because it was a frequent point of confusion "wait, doesn't the item already have options?". This makes it more clear what we're actually fetching.

packages/venia-ui/lib/components/ProductOptions/option.js
```diff
@@ -1,62 +1,87 @@
...
         return optionType === 'swatch' ? SwatchList : TileList;
-    }
+    }, [attribute_code, values]);
+
+    let initialSelection = {};
+    if (selectedVariant) {
+        // If we have a selected variant, pass it through to List.
+        // Attempt to find an option of the variant that's one of the product's configurable options.
+        // If found, set that configurable option as the initial selection of the List.
+        selectedVariant.options.find(option => {
```
supernova-at: We can use the new `productVariants` utility for this instead.

packages/venia-ui/lib/util/__tests__/productVariants.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    test('it returns the correct object when no match is found due to no matching option', () => {
+        const variantOption = Object.assign({}, variant.options[0], {
+            label: 'NO_MATCH '
+        });
+
+        const result = findMatch({ product, variantOption });
+
+        expect(result).toBeInstanceOf(Object);
+        expect(result.option).toBeUndefined();
+        expect(result.value).toBeUndefined();
```
supernova-at: If no matching option can be found, it would be better API design to have this match `findMatchingProductOption` and return `undefined` instead of an object with two properties that are `undefined`.

packages/venia-ui/lib/components/ProductOptions/option.js
supernova-at: Turns out we don't want to do this: we want this match to be restricted only to the `values` that this list is passed (as opposed to ALL `configurable_options` values).

supernova-at:
* We may want to remove the logic that enables / disables the "Update Cart" button based on how many selections have been made. We now always have all the selections, so the check always passes and the button is always enabled.
* We can save ourselves a call to the server by checking if any of the item's options actually changed. If not, there's no need to make the `update` call.

packages/peregrine/lib/List/list.js
```diff
@@ -1,5 +1,13 @@
 import React, { useMemo, useCallback } from 'react';
-import PropTypes from 'prop-types';
+import {
```
sirugh: Not sure where to comment this, but does `items` need to exist or can it be folded into the list component?

packages/venia-ui/lib/components/ProductOptions/option.js
supernova-at: We may be able to add a function to the `productVariants` utility for this ^ use case.

packages/venia-ui/lib/components/ProductOptions/option.js
```diff
@@ -1,62 +1,87 @@
-import React, { Component } from 'react';
+import React, { useCallback, useMemo } from 'react';
```
supernova-at: Lots of changes in this file but mostly due to converting this to a functional component.

packages/venia-ui/lib/components/ProductOptions/swatchList.js
```diff
@@ -1,22 +1,38 @@
-import React, { Component } from 'react';
-import { arrayOf, object, shape, string } from 'prop-types';
+import React from 'react';
```
supernova-at: Re-written to a functional component + adds `initialSelection` prop.

packages/venia-ui/lib/components/ProductOptions/tileList.js
```diff
@@ -1,22 +1,38 @@
-import React, { Component } from 'react';
-import { arrayOf, object, shape, string } from 'prop-types';
+import React from 'react';
```
supernova-at: Re-written to be a functional component + adds `initialSelection` prop.

packages/venia-ui/lib/components/ProductOptions/option.js
supernova-at: Eh, we'll add it if/when we need it later.

packages/peregrine/lib/List/list.js
supernova-at: These changes are no longer part of this PR

packages/venia-ui/lib/components/ProductOptions/option.js
```diff
@@ -1,62 +1,87 @@
...
+Option.propTypes = {
+    attribute_code: string.isRequired,
+    attribute_id: string,
+    classes: shape({
+        root: string,
+        title: string
+    }),
+    label: string.isRequired,
+    onSelectionChange: func,
+    selectedVariant: object,
```
sirugh: Can you add a `shape` to this object?

packages/venia-ui/lib/components/ProductOptions/options.js
```diff
@@ -34,6 +35,7 @@ class Options extends Component {
                 {...option}
                 key={option.attribute_id}
                 onSelectionChange={handleSelectionChange}
+                selectedVariant={selectedVariant}
```
sirugh: It's excessive to pass the entire `selectedVariant` object down to every single option block.

`selectedVariant` is a big object containing lots of props but the important one is `selectedVariant.options` an array of objects where the each object has a `label` and `value`. You could probably do a mapping here to pull off the specific selection that each `Option` cares about rather than asking the component to [figure it out](https://github.com/magento/pwa-studio/pull/1592/files#diff-5eee7b86d923aa851bd87db4f9e8b296R44).

[Image from Gyazo](https://gyazo.com/423f59d6325249c41b10da7009a29fe0)

packages/venia-ui/lib/components/ProductOptions/option.js
```diff
@@ -1,62 +1,87 @@
...
         // TODO: get an explicit field from the API
         // that identifies an attribute as a swatch
         const optionType = getOptionType({ attribute_code, values });
 
         return optionType === 'swatch' ? SwatchList : TileList;
-    }
+    }, [attribute_code, values]);
+
+    let initialSelection = {};
+    if (selectedVariant) {
```
sirugh: This could be replaced by a simple memo:

```js
const initialSelection = useMemo(() => {
  let selection = {};
  if (selectedVariant) {
    selection = values.find(value => value.default_label === selectedVariant.value) || {};
  }
  return selection;
}, [selectedVariant, values]);
```

This assumes you make the suggested changes to pre-select the variant in `Options`.

packages/venia-ui/lib/components/ProductOptions/option.js
```diff
@@ -80,7 +78,7 @@ Option.propTypes = {
     }),
     label: string.isRequired,
     onSelectionChange: func,
-    selectedVariant: object,
+    selectedValue: oneOfType([number, string]), // TODO: can these be objects?
```
sirugh: @supernova-at off the cuff do you recall whether these values are always
```js
{
  value: <string or number>
  label: <string>
}
```

supernova-at: That looks correct to me, yeah

sirugh:
```suggestion
    selectedValue: oneOfType([number, string]),
```

dpatil-magento:
QA pass for issue scope #1634 .

@supernova-at Should `update cart` button be disabled for default/pre-selected set of attributes?  We can create separate issue, if needed!

PR #1595, opened 2019-08-23
Integrated Webpack HTML Plugin.

### LINKED ISSUE #1529: [feature]: Add static, compiled SSR to Venia
Venia renders all pages dynamically and all requests through UPWARD. This is our current unified system for doing SEO and preload. We should change it. We should start delivering a fully statically served, client-rendered app to real users (mobile and desktop), as well as to fully rendering crawlers like Googlebot. We should only do SSR for the few relevant things (Link expanders for social media objects) that absolutely require it.

### Why tho

SSR is not necessary for SEO according to the PWA standards team. Their preferred approach is a [quick static render](https://developers.google.com/web/fundamentals/performance/prpl-pattern/), then progressive load. Plus, _**they endorse rendering different pages based on different browser capabilities**_ using https://github.com/Polymer/prpl-server. This is a big change from the conventional wisdom in eCommerce, and because SEO is so important to eCommerce, this will have to be a thoroughly proven approach before it reaches wide adoption. Therefore, we must adopt it as soon as possible!

#### What is wrong

Our SSR is very slow, which is partially the responsibility of the Magento 2.3 GraphQL API in its current state. But we are also not consuming that GraphQL API very efficiently. In Venia, we do _partial_ server-side render in an attempt to send content to all browsers that works for all browsers. But in practice, the backend GQL requests take [seconds before beginning to send an app shell, according to this WebPageTest performed by Alex Russell at Google.](https://www.webpagetest.org/result/190718_RD_89e9f4079dbcf0131a94ac08fc1db1f8/) All we use this for is pre-checking the route. Not worth it.

Most of the `upward.yml` file in Venia concerns strategies for rendering various app shells. Since this logic is so generic and UPWARD must handle requests dynamically, this adds a lot of time and complexity for all users, when the vast majority of the users are not going to need SSR at all. Instead, if they can get the app downloaded and loaded faster, they can use client-side SSR and ServiceWorker caching to achieve the same experience.

Google's best-practices showcase site https://shop.polymer.org is [fully client-side rendered](https://github.com/Polymer/prpl-server#differential-serving), and beats almost all other storefronts.

1. Refactor Venia to deliver from a statically generated HTML file. Measure performance gaoins
2. Add an HTML asset generator, like [the standard and popular one](https://github.com/jantimon/html-webpack-plugin), to Venia's Webpack build
3. Generalize this HTMLWebpackPlugin into the Buildpack pipeline
4. Add SSR back through extensions, to use only when requests come from a user agent known not to have client side rendering. [Outsource the tracking of those user agents.])https://github.com/Polymer/tools/tree/master/packages/browser-capabilities)

Full React SSR render, which is non-portable (it essentially requires NodeJS at runtime), slower, and more brittle.

- [x] `venia-concept`
- [x] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

revanth0212:
## Description

Added Webpack HTML Plugin to generate HTML at compile instead of runtime with UPWARD. HTML will be auto-generated using a template file that carries the boilerplate. In this case `template.html`.

## Related Issue
Closes #1529.

## Verification Steps
1. Compilation should not fail and should generate `index.html` file in the `dist` folder inside `venia-concept`.
2. When you go to any page on the UI and refresh, it should download the exact same `index.html` file irrespective of which page you are in. The previous implementation used to download a different file for different page.

Home Page: 
image
Category Page: 
image
Product Page: 
image

## Checklist
* Functionality of Venia should not change. This is a compiler modification. A sanity test on the app should be fine.


revanth0212:
Metrics coming soon.

packages/venia-concept/template.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8" />
+        <title><%= htmlWebpackPlugin.options.title %></title>
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
```
revanth0212: Note to my self: need to check the meta tags in prev templates and try getting them into either here or the app logic.

PR #1605, opened 2019-08-26
[feature]: Peregrine Implementation

### LINKED ISSUE #944: EPIC: Modular Component Library
Within the PWA project, the concept of "modularity" should exist as a system through the Peregrine component library. Today, we have deployed these concepts through Venia but the long term vision is to pull universal data and functional components into Peregrine for reuse in OOB or Customized storefronts. Peregrine, therefore, will become a universal library of react components that contains all of the functionality/features necessary for creating a PWA storefront.

## Definitive List (see [roadmap](https://gist.github.com/jimbo/f99b5adc4a5bcf959234e2206b798c26))
* Investigation of Redux replacement #1599
* Extraction of view state from reducers #1600
* Making global state available via context #1601
* Moving global state context providers to Peregrine #1602
* Rebinding components to consume Peregrine context #1603
* Extracting Venia component logic into Peregrine Hooks (ongoing) #1604

## Maybe List
* Peregrine Vision and Definition Documentation
* Developer Documentation
* Accessibility

## Deferred List
* Internationalization/Localization

sirugh:
## Description
This PR tracks the feature work for the #944 milestone. Opened as a `draft` pull request and based off `develop`. Feature work will be based on this branch unless it is applicable to the codebase as a whole in which case we will open pull requests against `develop`.

## Related Issue
Closes #944. 

## Verification Steps
1) Check everything.
2) ...
3) Profit!


dpatil-magento:
@sirugh When payment is disabled and I try to place order, I dont see error Toast anymore.

`develop` branch - 

image

This branch - 

image

packages/venia-ui/lib/components/App/container.js
```diff
@@ -1,39 +1,23 @@
...
-const { markErrorHandled } = appActions;
-const mapDispatchToProps = { closeDrawer, markErrorHandled };
-
-const ConnectedApp = connect(
-    mapStateToProps,
-    mapDispatchToProps
-)(App);
+import { useErrorBoundary } from './useErrorBoundary';
+
+const AppContainer = () => {
```
supernova-at: Just curious about why `ErrorBoundary` is special and gets its own file. We also may want to not name this file `container.js` because that's traditionally been for connecting components to Redux and may cause confusion.

packages/venia-ui/lib/components/App/contextProvider.js
```diff
@@ -17,10 +16,10 @@ const contextProviders = [
     ToastContextProvider
 ];
 
-const AppContextProvider = ({ children }) => {
+const ContextProvider = ({ children }) => {
```
supernova-at: Does the `PeregrineContextProvider` include all of the Redux state slice contexts? I'm having trouble piecing together all the things that wrap `App`.

packages/venia-ui/lib/components/Checkout/Receipt/receipt.js
```diff
@@ -60,6 +61,7 @@ Receipt.propTypes = {
         footer: string,
         root: string
     }),
+    onClose: func,
```
supernova-at: `.isRequired` because we call it directly

packages/venia-ui/lib/components/Checkout/addressForm.js
```diff
@@ -140,18 +128,10 @@ const AddressForm = props => {
...
-                    Cancel
-                </Button>
-                <Button
-                    className={classes.button}
-                    type="submit"
-                    priority="high"
-                    disabled={submitting}
-                >
+                <Button onClick={cancel}>Cancel</Button>
+                <Button type="submit" priority="high" disabled={isSubmitting}>
```
supernova-at: Lost the `className={classes.button}` on these two `Button`s, not sure if that was intentional

supernova-at: Nevermind, I see it removed from the CSS too 👍

packages/venia-ui/lib/components/CreateAccountPage/createAccountPage.js
```diff
@@ -1,40 +1,43 @@
...
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const { createAccount, history } = props;
 
-    createAccount = accountInfo => {
-        const { createAccount, history } = this.props;
-        createAccount({ accountInfo, history });
-    };
+    const handleCreateAccount = useCallback(
+        async accountInfo => {
+            await createAccount(accountInfo);
```
supernova-at: Okay so this pattern is different - we wait for actions / asyncActions directly in the component.

packages/venia-ui/lib/components/MiniCart/body.js
```diff
@@ -34,13 +34,17 @@ const Body = props => {
     const classes = mergeClasses(defaultClasses, props.classes);
 
     // Callbacks.
-    const handleEditItem = useCallback(
+    const handleBeginEditItem = useCallback(
         item => {
-            beginEditItem(item);
+            beginEditItem();
+            setEditItem(item);
```
supernova-at: [beginEditItem](https://github.com/magento/pwa-studio/blob/fa0a4c9b45757c272f6b745c546e8b61b1da40db/packages/venia-ui/lib/reducers/cart.js#L122) is entirely view state, we may be able to get rid of it here entirely (looks like it also sets `isEditingItem`).

packages/peregrine/lib/store/reducers/checkout.js
```diff
@@ -0,0 +1,157 @@
...
+    billingAddressError: null,
+    isSubmitting: false,
+    orderError: null,
+    paymentMethodError: null,
+    paymentCode: '',
+    paymentData: null,
+    shippingAddress: null,
+    shippingAddressError: null,
+    shippingMethod: '',
+    shippingMethodError: null,
```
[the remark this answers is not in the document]
sirugh: We tried but the tangle of actions invoking other actions mean that sometimes it's not a one-for-one.

There may be multiple error states created by invoking one action but you can only throw a single error from a stack.

```js
const handleSubmit = useCallback(() => {
  try {
    // invokes action 1 and action 2
    await doMultipleActions();
  }
  catch(err) {
    // What if action 1 and 2 throw?
  }
});
```

packages/peregrine/lib/store/reducers/user.js
```diff
@@ -17,23 +17,23 @@ const initialState = {
         firstname: '',
         lastname: ''
     },
-    getDetailsError: {},
-    isGettingDetails: false,
+    createAccountError: null,
+    getDetailsError: null,
+    isCreatingAccount: null,
+    isGettingDetails: null,
+    isResettingPassword: null,
```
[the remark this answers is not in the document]
sirugh: They are _sort of_ view states. What if you have multiple unrelated views that need to know if you are resetting the password? We started pulling out loading/wait and error states but decided to leave them in the global state because of all the coupled views.

packages/venia-concept/src/store.js
```diff
@@ -1,8 +1,12 @@
...
+import { combineReducers, createStore } from 'redux';
 
-import middleware from '@magento/venia-ui/lib/middleware';
-import errorHandler from '@magento/venia-ui/lib/middleware/errorHandler';
-import reducer from '@magento/venia-ui/lib/reducers';
-import composeEnhancers from '@magento/venia-ui/lib/util/composeEnhancers';
+// import middleware from '@magento/venia-ui/lib/middleware';
+// import errorHandler from '@magento/venia-ui/lib/middleware/errorHandler';
+// import reducer from '@magento/venia-ui/lib/reducers';
+// import composeEnhancers from '@magento/venia-ui/lib/util/composeEnhancers';
```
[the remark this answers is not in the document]
sirugh: @jimbo did you have plans for these? We talked about allowing the themes to compose/create their store but I recall you mentioning something wasn't quite working with that approach.

packages/peregrine/lib/store/enhancers/index.js
```diff
@@ -0,0 +1,7 @@
+import errorHandler from './errorHandler';
+import middleware from './middleware';
+import composeEnhancers from '../../util/composeEnhancers';
+
+const enhancer = composeEnhancers(middleware, errorHandler);
+
+export default enhancer;
```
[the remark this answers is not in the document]
jimbo: Enhancers are a core Redux concept. They're composable functions that add features to a store. Middleware, such as `redux-thunk`, can be applied to a store using the only enhancer Redux provides, `applyMiddleware`. [Read more in their docs.](https://redux.js.org/api/applymiddleware)

Our error handler is implemented with the enhancer pattern, so here we're composing our two enhancers (middleware and error handling) into a single enhancer that can be passed to `createStore`.

packages/venia-concept/src/store.js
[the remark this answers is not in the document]
jimbo: Yes, I need to delete these lines. Note that this file

- exists in the app ("theme"), `venia-concept`
- imports `reducers` from `peregrine`
- imports `enhancer` from `peregrine`
- combines reducers using the standard `combineReducers` export
- creates the store using the standard `createStore` export

A user can

- add their own reducers here
- ensure they get combined correctly with ours
- add their own enhancers here
- ensure they get combined correctly with ours

packages/peregrine/lib/store/actions/purchaseDetails/__tests__/asyncActions.spec.js
```diff
@@ -1,11 +1,10 @@
...
 import { fetchOrderDetails } from '../asyncActions';
 
-//TODO: write the rest part of the test when fetching async action will be in working condition(currently it's mock)
-jest.mock('../../../store');
-
+const dispatch = jest.fn();
+const getState = jest.fn();
 const thunkArgs = [dispatch, getState];
 
+// TODO: write the rest part of the test when fetching async action will be in working condition(currently it's mock)
```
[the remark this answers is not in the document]
sirugh: No I don't think we should make tickets for updating tests, especially for things that are likely to be deprecated (REST v GQL).

What we can do is probably remove the TODO and add a `test.skip`

packages/venia-ui/lib/components/App/contextProvider.js
sirugh: Yea, we had this discussion back when we made the global reduced provider and it was confusing then too :P The `PeregrineContextProvider` is meant to export a single wrapper that contains all the other context providers that we felt *essential* to Peregrine.

I think now that we're shifting those lines of what is "owned" by a developer (the storefront is where customization is made) we should reconsider where we define this reducer. IMO we should move it to `venia-concept` just like we have `createStore` in the storefront.

@jimbo does what I'm saying make sense? If so we should open an issue for this.

packages/venia-ui/lib/components/CreateAccountPage/createAccountPage.js
sirugh: Yea, if there's a UI related update that should happen in the view not in the action.

Most awaited function calls in the actions are just dispatches or non-ui related things like updating storage.

packages/venia-ui/lib/components/MiniCart/body.js
sirugh: The only thing it does right now is `setIsEditingItem` to `false` in the root miniCart component. Instead of threading setters/state down I just modified the callback.

packages/venia-ui/lib/components/CreateAccountPage/createAccountPage.js
supernova-at: Ok, good to know, thanks 👍

packages/venia-ui/lib/components/App/contextProvider.js
jimbo: @supernova-at Yes. `PeregrineContextProvider` is simply all of Peregrine's Redux slice context providers, composed into a single component.

This is how a storefront app is intended to import and expose core app state, so we wanted it to involve a minimum of boilerplate.

jimbo: > IMO we should move it to venia-concept just like we have createStore in the storefront.

It's good to bring that up.

The way we're exporting reducers and letting `venia-concept` combine them is actually counter to my intent. I wanted to eliminate Redux so that Venia wouldn't have to create a store at all; since we couldn't eliminate Redux yet, though, I had to let Venia create the store because it would've been too hard for Venia to customize a store created by Peregrine.

With context providers, there's no need to make Venia compose them. Peregrine needs some certainty over the structure of app state, and Venia doesn't need more opportunities to be misconfigured. In the future, we may export individual providers _in addition to_ this pre-composed one, but I don't foresee it being necessary.

jimbo:
> One thing in general that can be a bit confusing is that the store slices are a vestige of the REST API (I think), which we hope to be rid of soon.

> When that day comes and there's no clear mapping between a GraphQL query and these store slices, we may run into confusion. We may instead want to base our slices on types defined in GraphQL or more view-centric slices.

Interesting take. To be sure, slices end up mapping to REST because

1. the store acts as a client-side cache for REST response data
2. we use async actions to orchestrate REST requests & process responses

Our GraphQL client has its own cache for responses, but we still don't have an alternative to async actions (thunks, specifically) for orchestrating our async work. Once we do, we'll eliminate Redux, meaning these slices will just contain view state that's global in nature (cart, checkout, errors, etc).

dpatil-magento:
QA pass, good to merge.

PR #1606, opened 2019-08-26
fix test config

sirugh:
# Description

OK here's why it's broken.

`jest.config.js` uses `inPackage` to create a path to things in configs. It's a fancy helper for resolving paths. The specific problem is that `__mocks__/fileMock.js` does not exist in `venia-concept`. I _think_ that jest caches some things after running which is why you may or may not see this issue but I assure you it is a problem. If you do a clean checkout into a new directory you should see it happen.

So, we can solve this one of two ways.

1) Any test config using `testVenia` hardcodes a path to `__mocks__/fileMock.js` in `venia-ui`.

2) Create the `__mocks__/fileMock.js` in `venia-concept`.

Personally I don't care what we chose but it _is_ broken and we should fix it.

# Verification

1) `yarn test:dev`/`jest --watch` doesn't break.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -45,8 +45,7 @@ const testVenia = inPackage => ({
     browser: true,
     moduleNameMapper: {
         // Mock binary files to avoid excess RAM usage.
-        '\\.(jpg|jpeg|png)$':
-            '<rootDir>/packages/venia-ui/__mocks__/fileMock.js',
+        '\\.(jpg|jpeg|png)$': inPackage('__mocks__/fileMock.js'),
         // CSS module classes are dynamically generated, but that makes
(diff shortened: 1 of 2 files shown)


supernova-at:
Man, I'm still not getting it.

Your console output includes "Determining test suites to run..." but mine never does 🤔

revanth0212:
Happened to me till yesterday when I tried to use jest in watch mode. Unfortunately, I am not able to reproduce it now. Weird.

sirugh:
Checked out the repo anew and ran `yarn test:dev` caused it to fail again for me. lol

supernova-at:
Verification steps complete ✅

PR #1608, opened 2019-08-26
Provide app state via context

### LINKED ISSUE #1601: Make global state available via context
Regardless of whether global state is kept by a Redux store or React hooks, it should be exposed to the app by context providers so that components can consume it with a hook rather than wrapping themselves with an HOC.

There should be one provider per slice of global state.

jimbo:
## Description

As demonstrated in #1552, provide slices of app state via context. Each provider should provide state and bound action creators for a single slice, following our usual signature.

```es6
const [state, api] = useContext(SliceContext)
```

Consuming these contexts is not in scope for this issue, so it will be handled in a subsequent PR.

## Related Issue

Closes #1601.

## Verification Steps
1. Smoke test the app.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/App/useErrorBoundary.js
+++ b/packages/venia-ui/lib/components/App/useErrorBoundary.js
@@ -1,0 +1,30 @@
+import React, { Component, useMemo } from 'react';
+
+export const useErrorBoundary = WrappedComponent =>
+    useMemo(
+        () =>
+            class ErrorBoundary extends Component {
+                constructor(props) {
+                    super(props);
+
+                    this.state = { renderError: null };
+                }
+
+                static getDerivedStateFromError(renderError) {
+                    return { renderError };
+                }
+
+                render() {
+                    const { props, state } = this;
+                    const { renderError } = state;
+
+                    return (
+                        <WrappedComponent
+                            {...props}
+                            renderError={renderError}
+                        />
+                    );
+                }
+            },
+        []
+    );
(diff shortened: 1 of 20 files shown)


packages/venia-ui/lib/context/catalog.js
```diff
@@ -0,0 +1,39 @@
...
+import { connect } from '../drivers';
+
+export const CatalogContext = createContext();
+
+const CatalogContextProvider = props => {
+    const { catalog: catalogState, children, updateCategories } = props;
+
+    const catalogApi = useMemo(
+        () => ({
+            updateCategories
```
sirugh: Do we want to add the other actions like `addFilter` and `removeFilter`? I think we should make sure that each of these new contexts exports every action available and then in a subsequent PR we can clean up unused ones.

sirugh:
```js
import * as actions from '../actions/catalog';
const catalogApi = useMemo(
  () => ({
    ...actions
  })
);
```

or something...

packages/venia-ui/lib/components/App/useErrorBoundary.js
```diff
@@ -0,0 +1,30 @@
...
+
+                    return (
+                        <WrappedComponent
+                            {...props}
+                            renderError={renderError}
+                        />
+                    );
+                }
+            },
+        []
```
sirugh: What's the point in a `useMemo` that recomputes every time?

PR #1619, opened 2019-08-29
Security Dependency Upgrades

### LINKED ISSUE #1613: [bug]: Resolve GitHub Security Alerts
**Describe the bug**
GitHub's automated security fixes has identified issues with two dependencies: `mixin-deep` and `set-value`.

https://github.com/magento/pwa-studio/network/alerts

**Possible solutions**
- Upgrade `mixin-deep` to version **1.3.2** or later.
- Upgrade `set-value` to version **2.0.1** or later

supernova-at:
## Description

The `mixin-deep` and `set-value` dependencies had vulnerabilities that we wanted to fix.

Unfortunately, we don't depend on these packages directly:

### mixin-deep

`mixin-deep` is a transitive dependency of `chokidar` and `danger`:

```
npm ls mixin-deep

@magento/pwa-studio@3.0.0
└─┬ chokidar@2.1.6
  └─┬ braces@2.3.2
    └─┬ snapdragon@0.8.2
      └─┬ base@0.11.2
        └── mixin-deep@1.3.2 
```

✅  This PR upgrades `chokidar` to the latest version (`3.0.2`), which no longer pulls in `mixin-deep@1.3.2`.

```
npm ls mixin-deep

@magento/pwa-studio@3.0.0
└─┬ danger@7.1.4
  └─┬ micromatch@3.1.10
    └─┬ snapdragon@0.8.2
      └─┬ base@0.11.2
        └── mixin-deep@1.3.2
```

❌ This PR does not update `danger` to the latest version (`9.1.8`), because unfortunately this still doesn't fix the problem, as `danger@9.1.8` still relies on the same version of `micromatch` (`3.1.10`):

```
npm ls mixin-deep

@magento/pwa-studio@3.0.0
└─┬ danger@9.1.8
  └─┬ micromatch@3.1.10
    └─┬ snapdragon@0.8.2
      └─┬ base@0.11.2
        └── mixin-deep@1.3.2 
```

### set-value

`micromatch@3.1.10` is the culprit for `set-value` as well:

```
npm ls set-value

@magento/pwa-studio@3.0.0
└─┬ danger@9.1.8
  └─┬ micromatch@3.1.10
    └─┬ snapdragon@0.8.2
      └─┬ base@0.11.2
        └─┬ cache-base@1.0.1
          ├── set-value@2.0.1 
          └─┬ union-value@1.0.1
            └── set-value@2.0.1  deduped
```

### Yarn resolutions

`micromatch` has released a [major version 4](https://github.com/micromatch/micromatch/releases) which no longer depends on `snapdragon` (which leads to `mixin-deep` and `set-value`.

✅ This PR uses [yarn resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) to force `micromatch@4.0.2` instead of the problematic `3.1.10`.

## Related Issue

Closes #1613.

### We no longer depend on `mixin-deep` or `set-value`

1. Run `npm ls mixin-deep` and verify that it is no longer a dependency (empty result)
1. Run `npm ls set-value` and verify that it is no longer a dependency (empty result)

### chokidar

The only place `chokidar` is used is in [watch-all.js](https://github.com/magento/pwa-studio/blob/develop/scripts/watch-all.js), run by `yarn watch:all`.

1. `yarn clean:all`
1. `yarn install`
1. `yarn build` 
1. Ensure `yarn watch:all` succeeds

For an extra sense of security, review [the changes](https://github.com/paulmillr/chokidar/releases) since `chokidar@2.1.6`.

## Checklist

## Risk

Sadly this PR was failing unit test checks. I tried locally and there aren't any tests failing, it's the coverage report because it is attempting to collect coverage from `.json` files. Here's some output:

```
STACK: SyntaxError: /Users/aterrano/code/pwa-studio/packages/venia-concept/package.json: Unexpected token, expected ";" (2:8)

  1 | {
> 2 |   "name": "@magento/venia-concept",
    |         ^
  3 |   "version": "3.0.0",
  4 |   "publishConfig": {
  5 |     "access": "public"
    at Parser.raise (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:6325:17)
    at Parser.unexpected (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:7642:16)
    at Parser.semicolon (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:7624:40)
    at Parser.parseExpressionStatement (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:10364:10)
    at Parser.parseStatementContent (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:9963:19)
    at Parser.parseStatement (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:9829:17)
    at Parser.parseBlockOrModuleBlockBody (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:10405:25)
    at Parser.parseBlockBody (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:10392:10)
    at Parser.parseBlock (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:10376:10)
    at Parser.parseStatementContent (/Users/aterrano/code/pwa-studio/node_modules/@babel/parser/lib/index.js:9905:21)
```

This error occurs for every `.json` file in the codebase.

Unfortunately, `jest.config.js` wasn't respecting any combination of `collectCoverageFrom` or `coveragePathIgnorePatterns` to stop these `.json` files from being instrumented.

The only solution that stopped the errors was turning off coverage completely.

I created #1620 to dig deeper into the problem and turn coverage back on.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -157,7 +157,7 @@ const jestConfig = {
     ],
     // Include files with zero tests in overall coverage analysis by specifying
     // coverage paths manually.
-    collectCoverage: true,
+    collectCoverage: false,
     collectCoverageFrom: [
         // Code directories
         'packages/*/{src,lib}/**/*.js',
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "chalk": "~2.4.2",
     "chokidar": "~3.0.2",
     "coveralls": "~3.0.3",
-    "danger": "^9.1.8",
+    "danger": "^7.1.4",
     "eslint": "~5.14.1",
     "eslint-plugin-babel": "~5.3.0",
     "eslint-plugin-graphql": "~3.0.3",
--- a/package.json
+++ b/package.json
@@ -46,9 +46,9 @@
   "devDependencies": {
     "@magento/eslint-config": "~1.4.1",
     "chalk": "~2.4.2",
-    "chokidar": "~2.1.2",
+    "chokidar": "~3.0.2",
     "coveralls": "~3.0.3",
-    "danger": "^7.1.4",
+    "danger": "^9.1.8",
     "eslint": "~5.14.1",
     "eslint-plugin-babel": "~5.3.0",
     "eslint-plugin-graphql": "~3.0.3",
@@ -78,7 +78,8 @@
     "sharp": "~0.22.1"
   },
   "resolutions": {
-    "graphql": "~14.3.1"
+    "graphql": "~14.3.1",
+    "micromatch":"~4.0.2"
   },
   "engines": {
     "node": ">=10.x",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3000,13 +3000,6 @@ abbrev@1:
   resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.1.tgz#f8f2c887ad10bf67f634f005b6987fed3179aac8"
   integrity sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==
 
-abort-controller@^3.0.0:
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/abort-controller/-/abort-controller-3.0.0.tgz#eaf54d53b62bae4138e809ca225c8439a6efb392"
-  integrity sha512-h8lQ8tacZYnR3vNQTgibj+tODHI5/+l06Au2Pcriv/Gmet0eaj4TwWH41sO9wnHDiQsEj19q0drzdWdeAHtweg==
-  dependencies:
-    event-target-shim "^5.0.0"
-
 accepts@^1.3.0, accepts@^1.3.5, accepts@~1.3.4, accepts@~1.3.5, accepts@~1.3.7:
   version "1.3.7"
   resolved "https://registry.yarnpkg.com/accepts/-/accepts-1.3.7.tgz#531bc726517a3b2b41f850021c6cc15eaab507cd"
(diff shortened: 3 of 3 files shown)


sirugh:
Nice write up! IMO if `danger`s latest doesn't fix it then we shouldn't touch it until they upgrade

sirugh:
The reason why coverage is broken on this PR is because we are forcing micromatch to v4 from v3. `jest` uses v3 and there is an [open issue](https://github.com/micromatch/micromatch/issues/179) about broken matching in v4.

That said, I looked into the security vulnerabilities reported by github and I'm not sure we even need to do anything. Infact if you look at the output of the `npm ls <package>` from the root level you'll see that we only include the fixed versions of these dependencies. `mixin-deep@1.3.2` and `set-value@2.0.1` are both "fixed" according to the reports. Additionally, the report from github is about problems within `pwa-buildpack` and `pwa-devdocs` specifically.

For both `pwa-buildpack` and `pwa-devdocs`, go into the directory and `npm install` and run the same `npm ls <package>` commands. You should see the patched versions of the vulnerable dependencies in both cases.

All of this is to say I'm just not sure we need this PR. I _think_ we can dismiss these security reports. Please double check me on this, of course!

supernova-at:
> All of this is to say I'm just not sure we need this PR. I think we can dismiss these security reports. Please double check me on this, of course!

I think you're totally right.

> The report from github is about problems within pwa-buildpack and pwa-devdocs specifically.

I had this thought too, these are "non-production" packages / devDependencies and therefore the severity of these issues is decreased.

---

It was a little disconcerting to see we were multiple major versions behind some of these dependencies though - I created an issue here to upgrade: #1621 .

Issue #1617, opened 2019-08-29
Fetch store name from the GraphQL server at compile time.


revanth0212:
**Is your feature request related to a problem? Please describe.**
 This is related to #1595. As part of that PR, we have created a global variable called `STORE_NAME` that can be added to the bundle at compile time using the Webpack Define plugin. Instead of hardcoding it in Define plugin, can we request that from the GraphQL server and then insert that into the `global` object just like the Define plugin?

**Describe the solution you'd like**
* Modify the [getStoreConfigData.graphql](https://github.com/magento/pwa-studio/blob/develop/packages/pwa-buildpack/lib/queries/getStoreConfigData.graphql) file to include `store_name`, as per the `StoreConfig` schema.
* Modify the [getAvailableStoresConfigData.graphql](https://github.com/magento/pwa-studio/blob/develop/packages/pwa-buildpack/lib/queries/getAvailableStoresConfigData.graphql) to also include the `store_name`.
* Using the data from these queries, if `STORE*VIEW_CODE` environment variable is used, use the store*name value from the corresponding `availableStore`. If not, use the default value returned by the getStoreConfigData query.
* Update jest config with whatever the sample data store name is from venia.

**Describe alternatives you've considered**
 We can expect the partner developers to simply change the hardcoded string in webpack's Define plugin.

**Please let us know what packages this feature is in regards to:**
 - <x] `venia-concept`
 - [ ] `venia-ui`
 - [x] `pwa-buildpack`
 - [ ] `peregrine`
 - [ ] `pwa-devdocs`
 - [ ] `upward-js`
 - [ > `upward-spec`

**Test Plan:**
* On first load (clear cache) of home page route, url will be displayed initially as title in browser.
* After a short time, title will update to contain view name.
* Refresh page. Title should contain correct view name (from cache). You should not see url in title.
* Repeat above steps for:
  * CMS page
  * Category page
  * Product page
  * Account Information page
  * Address Book page
  * Communications page
  * Reset Password page
  * Order History page
  * Cart page
  * Checkout page
  * Order confirmation page (refresh will go back to cart page)
* Run MFTF

supernova-at:
Check the Risks & Caveats section of #1739 for issues we may run into when fetching GraphQL data at build time.

That said, "store name" should theoretically not change all that often 😄

zetlen:
We already have a relatively new plugin which gets the media URL from GraphQL at build time. Who wants to extend it?

You'll need:

[...]

PR #1628, opened 2019-08-30
Replacing async webpack plugin.

revanth0212:
Generalized MediaBackendURLPlugin to work for any Async workload at compile time.

## Related Issue
Doesn't really close any PR. But it sets up the code for https://github.com/magento/pwa-studio/issues/1617.

## Verification Steps
Doesn't change anything. It is just a code refactor.


packages/venia-concept/src/AsyncWebpackPlugin.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * A Webpack plugin that takes an Async function to execute and
+ * halts webpacks execution till the Async function is not completed.
+ */
+class AsyncWebpackPlugin {
```
supernova-at: This feels like it should go in `buildpack` instead of `venia-concept`.

packages/venia-concept/src/AsyncWebpackPlugin.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * A Webpack plugin that takes an Async function to execute and
+ * halts webpacks execution till the Async function is not completed.
```
supernova-at:
```suggestion
 * halts webpacks execution till the Async function is completed.
```

I'm assuming? 😛

packages/venia-concept/src/AsyncWebpackPlugin.js
```diff
@@ -0,0 +1,33 @@
...
+            try {
+                return this.callback();
+            } catch (err) {
+                console.error(
+                    `Error in AsyncWebpackPlugin while executing ${
+                        this.callback ? this.callback.name : 'the callback'
+                    }`,
+                    err
+                );
+                return Promise.resolve();
```
supernova-at: Any reason not to return `Promise.reject` here?

packages/venia-concept/src/__tests__/AsyncWebpackPlugin.spec.js
```diff
@@ -0,0 +1,43 @@
...
+        emit: {
+            tapPromise
+        }
+    }
+};
+
+beforeEach(() => {
+    process.env.MAGENTO_BACKEND_URL =
+        'https://venia-cicd-lrov2hi-mfwmkrjfqvbjk.us-4.magentosite.cloud/';
+    fetch.resetMocks();
```
supernova-at: Can you add a comment in the code about what `fetch` is and where it comes from?

Who is actually mocking it?

packages/venia-concept/src/__tests__/AsyncWebpackPlugin.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    asyncPluginInstance.apply(compiler);
+    tapPromise.mock.calls[0][1].call();
+
+    expect(callback).toHaveBeenCalledTimes(1);
+});
+
+test('If an invalid callback is provided while creating instance, instance creation should fail', () => {
+    expect(() => new AsyncWebpackPlugin()).toThrow();
+    expect(() => new AsyncWebpackPlugin(null)).toThrow();
+});
```
supernova-at: A couple other tests I can think of might be testing that the promise resolves or rejects as appropriate when the `callback` succeeds or fails.

packages/venia-concept/webpack.config.js
```diff
@@ -56,7 +58,7 @@ module.exports = async env => {
              */
             STORE_NAME: JSON.stringify('Venia')
         }),
-        new MediaBackendUrlFetcherPlugin(),
+        new AsyncWebpackPlugin(getMediaURL),
```
supernova-at: Since the export of this module is already an `async` function, another approach I've seen is to just do the async work before returning your webpack config object return value:

```js
module.exports = async env => {
    const mediaUrl = await getMediaURL();

    const config = await configureWebpack({ ... });

    return config;
};
```

packages/venia-concept/src/webpackUtils.js
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * An Async function that will asynchronously fetch the
+ * media backend Url from magento graphql server. Also
+ * will set the MAGENTO_MEDIA_BACKEND_URL value on global
+ * object to the resulting url.
+ *
+ * @returns Promise that will resolve to the media backend url.
+ */
+const getMediaURL = () =>
```
supernova-at: I like this better as a utility function than a webpack plugin for sure. Seems also like something that could / should live in `buildpack` instead of `venia-concept` though.

packages/venia-concept/src/AsyncWebpackPlugin.js
revanth0212: 💩 how did I miss that?

packages/venia-concept/src/AsyncWebpackPlugin.js
revanth0212: If we reject inside one of the webpack plugins, it will cancel the execution of all the next plugins to follow.

packages/venia-concept/src/__tests__/AsyncWebpackPlugin.spec.js
revanth0212: Got it. Done.

packages/venia-concept/src/__tests__/AsyncWebpackPlugin.spec.js
revanth0212: `resolves` test-case is already present in the suite. I have added one for `reject`.

packages/venia-concept/webpack.config.js
revanth0212: Doing so will move the async logic from the plugin to the exports function which might become clumsy when we start adding more async functionalities like fetching store name for instance in the future. Also, this will be defeating the purpose of the plugin.

I have been reading on plugins and apparently we can use webpack plugins outside `plugins` object, since they are just JS classes at the end of the day. Leaving logic for async functionality inside the plugin gives us more grip over the usage as well.

This is a long shot but if there is a new and more optimized way to do async stuff in JS in the future, having such things in 1 place will make it easy to perform modifications.

packages/venia-concept/src/AsyncWebpackPlugin.js
revanth0212: Makes sense. Will do that.

packages/venia-concept/webpack.config.js
supernova-at: > this will be defeating the purpose of the plugin.

Yes 😄 
I'd rather we create and maintain a bunch of utility functions than webpack plugins though.

> Doing so will move the async logic from the plugin to the exports function

Looking at this line for example:

```js
const mediaUrl = await getMediaURL();
```

We move the async logic to `getMediaURL`, with a simple call to that function in the exports function.

> which might become clumsy

It's good to keep an eye on this 👍 . I think straightforward utility functions are less clumsy than webpack plugins in the end though.

revanth0212: My only problem with performing such async stuff as part of configureWebpack is code maintainability.

```es6
module.exports = async env => {
    const mediaUrl = await getMediaURL();
    const storeName = await getStoreName();
    const timeZone = await getTimeZone();
    ....
    const config = await configureWebpack({
      // use mediaUrl somewhere in here
    });

    return config;
};
```

packages/pwa-buildpack/lib/Utilities/fetcherUtils.js
```diff
@@ -0,0 +1,39 @@
+const fetch = require('node-fetch');
+
+const fetchQuery = query =>
```
supernova-at: I suppose this is okay for now, we may want to use Apollo client with `buildpack` eventually since it bakes in a bunch of best practices. We won't really need it for the cache though, so it may be overkill.

packages/pwa-buildpack/lib/Utilities/fetcherUtils.js
```diff
@@ -0,0 +1,39 @@
...
+    fetch(new URL('graphql', process.env.MAGENTO_BACKEND_URL).toString(), {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
```
supernova-at: We should also send a `Accept-Encoding: gzip` header in accordance with the [GraphQL best practices](https://graphql.org/learn/best-practices/#json-with-gzip).

packages/venia-concept/webpack.config.js
```diff
@@ -1,9 +1,12 @@
...
+const {
+    configureWebpack,
+    fetcherUtils: { getMediaURL }
+} = require('@magento/pwa-buildpack');
 const { DefinePlugin } = require('webpack');
 const HTMLWebpackPlugin = require('html-webpack-plugin');
-const MediaBackendUrlFetcherPlugin = require('./src/MediaBackendURLFetcherPlugin');
 
 module.exports = async env => {
+    await getMediaURL();
```
supernova-at: Do we not need to do anything with the result of this?

I see the promise also sets `global.MAGENTO_MEDIA_BACKEND_URL` ... we probably want to move that out of being a side effect of the "fetcher" and have some code intentionally set that value elsewhere (here?).

supernova-at: This is the main thing I'd like to see changed, then 👍

revanth0212: The idea was to not clutter exports object with these. But Ill refactor this to have a getter that is pure and a setter that has the global object setting logic.

packages/pwa-buildpack/lib/Utilities/fetcherUtils.js
revanth0212: We could if we have more GQL calls to make at build time which is what buildpack does. If not, we will be having a dependency just to make a call.

zetlen: It's not an additional dependency. We already know that apollo-client will be available in the project because it's a peer dependency. However, it's an optimization to replace this with apollo-client and I'm happy to merge this as-is.

PR #1630, opened 2019-09-03
[WIP] Migrating store and redux to peregrine

### LINKED ISSUE #1602: Move global state context provider to Peregrine
Once global state is available from context providers, both the state and context providers should migrate from Venia to Peregrine.

If the Redux store is still in use, it should also move to Peregrine, which will require dependencies to be reorganized.

sirugh:
## Description

This PR moves the store and all dependencies to Peregrine.
See the issue for more information.

## Related Issue

Closes #1602.

## Verification Steps

This touches a _lot_ of code. A regression test should cover our bases on top of unit tests, but I encourage you to try various actions/ buttons within the app that utilize redux to ensure I didn't break anything.

## Checklist


packages/peregrine/lib/util/createErrorRecord.js
```diff
@@ -0,0 +1,64 @@
+// Cache the generated error records, but let them be garbage collected.
```
sirugh: The original file should be deleted from `venia-ui/lib/util`.

packages/peregrine/lib/util/getFilterParamsFromUrl.js
```diff
@@ -0,0 +1,32 @@
+import { persistentQueries } from '../shared/persistentQueries';
```
sirugh: The original file should be deleted from `venia-ui/lib/util`.

sirugh: Wait, did you leave some behind intentionally?

packages/peregrine/lib/util/preserveQueryParams.js
```diff
@@ -1,10 +1,11 @@
+import { persistentQueries } from './persistentQueries';
 /**
  * Function takes location object, filters through it and creates
  * new query params with preserved values from the array.
  * @param {object} location
  * @param {array} queries
  */
-export const preserveQueryParams = (location, queries) => {
+export const preserveQueryParams = (location, queries = persistentQueries) => {
```
sirugh: Not sure if this is correct but it was the easiest way to move this to peregrine, for now.

We use this in both `peregrine` (in asyncAction for catalog) and `venia-ui` (in filterFooter). I guess, if someone wanted to use this utility, they would need to know _what_ the defaults were at least. 

TODO: Update the jsdoc comment.

packages/peregrine/lib/store/actions/checkout/asyncActions.js
```diff
@@ -3,7 +3,7 @@ import { RestApi, Util } from '@magento/peregrine';
 import { closeDrawer } from '../app';
 import { clearCartId, createCart } from '../cart';
 import { getCountries } from '../directory';
-import { getAccountInformation } from '../../selectors/checkoutReceipt';
+import { getAccountInformation } from '@magento/peregrine/lib/store/selectors/checkoutReceipt';
```
sirugh: We could clean up these later to be a little more specific like:

```js
import { selectors } from '@magento/peregrine'
const { checkoutReceipt } = selectors;
```

jimbo: I don't really want to have selectors anyway; these were added without approval in the first place.

Especially now that app state is being provided by context, if we really need to have derived state, we can append derived properties directly to the context value.

jimbo:
Merging this into the feature branch. 👍

PR #1636, opened 2019-09-05
Convert components to function components

sirugh:
As we work towards utilizing peregrine hooks for context, state and logic we need to also make sure components are written in a way that can handle hooks. One of the core requirements to use hooks is that a component utilizing it be a "function component" ie.

```js
   ...do hook thangs
}
```

This PR converts components into function components.

I am intentionally avoiding some specific components as they aren't being used so there's no point in diving into that right now.

* Anything within `PurchaseHistoryPage`
* Anything within `PurchaseDetailsPage`


packages/peregrine/lib/hooks/useQuery.js
```diff
@@ -27,7 +27,7 @@ export const useQuery = query => {
      * @param {DocumentNode} query A GraphQL document
      */
     const runQuery = useCallback(
-        async ({ variables }) => {
+        async ({ variables } = {}) => {
```
sirugh: This is a QOL fix. Right now if you call `runQuery` for a query that doesn't accept variables you have to pass an empty object. This makes that unnecessary so you can just call `runQuery()` instead of `runQuery({})`.

packages/venia-ui/lib/RootComponents/CMS/cms.js
```diff
@@ -1,8 +1,8 @@
...
-export default class CMS extends Component {
-    render() {
-        return <CategoryList title="Shop by category" id={2} />;
-    }
-}
+const CMS = () => {
+    return <CategoryList title="Shop by category" id={2} />;
+};
+
+export default CMS;
```
sirugh: Some of these are pretty simple but also perhaps unnecessary. This is one of those examples. It may be preferable not to do _all_ components and instead focus on ones that use some context/connected logic or functions provided by a container. These will eventually move to hooks in the scope of the peregrine migration milestone.

packages/venia-ui/lib/RootComponents/Product/product.js
```diff
@@ -17,53 +14,56 @@ import productQuery from '../../queries/getProductDetail.graphql';
  * https://github.com/magento/graphql-ce/issues/86
  * TODO: Replace with a single product query when possible.
  */
-class Product extends Component {
-    static propTypes = {
-        cartId: string
-    };
+const Product = () => {
```
sirugh: Double check that Product views are functionally identical as previous.

packages/venia-ui/lib/components/Footer/footer.js
```diff
@@ -1,93 +1,87 @@
...
-    static propTypes = {
-        classes: PropTypes.shape({
-            copyright: PropTypes.string,
-            root: PropTypes.string,
-            tile: PropTypes.string,
-            tileBody: PropTypes.string,
-            tileTitle: PropTypes.string
-        })
-    };
+const Footer = props => {
```
sirugh: Double check that this is functionally the same as previous.

packages/venia-ui/lib/RootComponents/Product/product.js
sirugh: Works!

packages/venia-ui/lib/components/Footer/footer.js
sirugh: Works!

sirugh: Works!

PR #1637, opened 2019-09-05
Zetlen/inline logo svg

### LINKED ISSUE #1622: [bug]: venia logo does not load on storefront for upward-php.
**Describe the bug**
venia logo does not load on storefront for upward-php.

**To Reproduce**
1. Deploy pwa on cloud starter 
2. Access storefront

**Expected behavior**
Storefront should load with all images.

**Actual**
Venia logo does not load.

**Screenshots**
image

dpatil-magento:
## Description
Fixes #1622. Also Magento treats the path "/static" as special, so changed to "/venia-static"

## Related Issue

Closes #1622 

## Verification Steps

1. Verify #1622 works fine. 
2. For upward-php and js make sure no network call ends up in errors like 404, 500 etc.


zetlen:
This looks amazing. Well done, whoever.

zetlen:
This appears to be an issue with `upward-php` and not with `upward-js`, or maybe with `module-upward-connector` which connects `upward-php` with the Magento routing table. In any case, we just needed to rename that directory.

The right way to hammer out this confusing file path stuff is to add proper routing concepts to UPWARD. Opening a separate issue to handle that ASAP.

dpatil-magento:
QA Pass, good to merge once review approved.

packages/venia-concept/template.html
```diff
@@ -21,8 +21,8 @@
             Apple Icons for Homescreen
             Sizes: https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/app-icon/
         -->
-        <link rel="apple-touch-icon" href="/icons/venia_square_57.png">
-        <link rel="apple-touch-icon" sizes="180x180" href="/icons/apple-touch-icon.png">
+        <link rel="apple-touch-icon" href="/venia-static/icons/venia_square_57.png">
```
sirugh: So was the problem that we were using `/icons` instead of `/static/icons`?

PR #1644, opened 2019-09-09
fix UPWARD yml path

blackpr:
fix the  UPWARD yml file name:
https://github.com/magento/pwa-studio/blob/master/packages/venia-concept/venia-upward.yml


sirugh:
Isn't this getting renamed in `4.0.0`? @jcalcaben 

https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/upward.yml
https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/upward.yml

jcalcaben:
Yeah you're right @sirugh 
I'll need to audit and update this topic after we release 4.0.0

PR #1664, opened 2019-09-10
Replace connected components with context hooks

### LINKED ISSUE #1603: Rebind components to consume context
Once global state is available from Peregrine context providers, components should consume it from context rather than wrapping themselves with an HOC.

sirugh:
## Description

We no longer need `mapStateToProps` or `mapDispatchToProps` as we've migrated the redux store to Peregrine hooks/context. This is just one more step in removing venia's dependency on redux.

Also contained in this PR is a refactor of filter/search components as it was not a function component and we needed it to be in order to use context.

## Related Issue 
Closes #1603.

## Verification Steps

Another general regression will be necessary but no functional changes should occur.


packages/venia-ui/lib/RootComponents/Search/categoryFilters.js
```diff
@@ -0,0 +1,68 @@
...
+import { X as CloseIcon } from 'react-feather';
+import gql from 'graphql-tag';
+import { Query } from '@magento/venia-drivers';
+
+import { mergeClasses } from '../../classify';
+import Icon from '../../components/Icon';
+import getQueryParameterValue from '../../util/getQueryParameterValue';
+import defaultClasses from './categoryFilters.css';
+
+const GET_CATEGORY_NAME = gql`
```
sirugh: I know this is a small bit of code but should we store these with the other queries in the query dir?

packages/venia-ui/lib/RootComponents/Search/categoryFilters.js
```diff
@@ -0,0 +1,68 @@
...
+            <button
+                className={classes.filter}
+                onClick={handleClearCategoryFilter}
+            >
+                <small className={classes.text}>
+                    <Query
+                        query={GET_CATEGORY_NAME}
+                        variables={{ id: categoryId }}
+                    >
+                        {renderResult}
```
sirugh: Would it be more appropriate to put all the html as the query child instead of just the name? In the event of an error from GQL we don't want to render a blank div anyways.

packages/venia-ui/lib/RootComponents/Search/search.js
```diff
@@ -1,181 +1,154 @@
...
-        if (!inputText) {
-            return <Redirect to="/" />;
+    }, []);
+    /* eslint-enable react-hooks/exhaustive-deps */
+
+    // clear filters whenever the `query` search param changes
+    // but don't clear them on mount
+    useEffect(() => {
+        if (shouldClear.current) {
+            filterClear();
```
sirugh: We also shouldn't need to clear if it's already clear right? ie I start typing in a new search input, filter gets cleared after one character and re-cleared every subsequent character because of this effect.

packages/venia-ui/lib/RootComponents/Search/search.js
```diff
@@ -1,181 +1,154 @@
...
+const Search = props => {
+    const {
+        executeSearch,
+        filterClear,
+        history,
+        location,
+        openDrawer,
+        searchOpen,
+        toggleSearch
+    } = props;
```
sirugh: Can you pull any of these off the context?

sirugh:
Dropping a `hold` label on here until the slices are merged in #1686 and #1684

packages/venia-ui/lib/RootComponents/Search/search.js
jimbo: No, this effect doesn't fire when the text field value changes; it fires when the search param value changes. [Compare to the original.](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/RootComponents/Search/search.js#L42-L71)

sirugh: `inputText` and `queryText` confuses me :D

packages/venia-ui/lib/RootComponents/Search/categoryFilters.js
jimbo: Rather than a blank container, it's actually a button with functionality (and a close icon). If the query were to fail, we'd still want the button and the close icon to be there.

I think we should preserve the original work here.

packages/venia-ui/lib/RootComponents/Search/search.js
```diff
@@ -1,181 +1,156 @@
...
+    const { history, location } = props;
+    const shouldClear = useRef(false);
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    // retrieve app state and action creators
+    const [appState, appApi] = useAppContext();
+    const { searchOpen } = appState;
+    const { executeSearch, toggleDrawer, toggleSearch } = appApi;
+    const [, catalogApi] = useCatalogContext();
+    const { clear: clearFilters } = catalogApi.actions.filterOption;
```
sirugh: Do you want to use the async action `filterClear` instead of this? I know you fixed the `actions` export already. Just LMK, I don't feel strongly either way.

packages/venia-ui/lib/RootComponents/Search/search.js
```diff
@@ -1,181 +1,156 @@
...
+
+    // retrieve search params
+    const inputText = getQueryParameterValue({
```
sirugh: Renaming this to `queryText` would make it a little clearer IMO.

packages/peregrine/lib/store/actions/catalog/asyncActions.js
```diff
@@ -83,3 +85,13 @@ export const setPrevPageTotal = payload =>
     async function thunk(dispatch) {
         dispatch(actions.setPrevPageTotal.receive(payload));
     };
+
+export const setToApplied = payload =>
+    async function thunk(dispatch) {
+        dispatch(actions.filterOption.setToApplied(payload));
+    };
+
+export const filterClear = payload =>
```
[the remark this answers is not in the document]
sirugh: Now that we fixed `actions` I can delete this and the `setToApplied` asyncActions that I created. Thoughts @jimbo ?

packages/peregrine/lib/util/bindActionCreators.js
```diff
@@ -0,0 +1,15 @@
+import { bindActionCreators } from 'redux';
+
+const getBindFunction = value =>
+    typeof value === 'function'
+        ? bindActionCreators
+        : bindActionCreatorsRecursively;
+
+const bindActionCreatorsRecursively = (actions, dispatch) =>
```
sirugh: JSdoc here would go a long way, IMO.

packages/venia-ui/lib/components/Checkout/flow.js
```diff
@@ -157,31 +159,9 @@ const Flow = props => {
...
-    }).isRequired,
     classes: shape({
         root: string
-    }),
-    submitOrder: func,
-    submitPaymentMethodAndBillingAddress: func,
-    submitShippingAddress: func,
-    submitShippingMethod: func,
-    user: object
+    })
```
sirugh: Add `step` and `setStep` to propTypes.

packages/venia-ui/lib/components/MiniCart/miniCart.js
```diff
@@ -9,31 +9,36 @@ import defaultClasses from './miniCart.css';
...
 const MiniCart = props => {
+    const [, { closeDrawer }] = useAppContext();
+    const [
+        cartState,
+        { updateItemInCart, removeItemFromCart }
+    ] = useCartContext();
+    const [, { cancelCheckout }] = useCheckoutContext();
     const [step, setStep] = useState('cart');
+
+    const isCartEmpty =
```
sirugh: We can replace this with the derived state value when #1703 is merged.

sirugh: Not actually a merge conflict either -- I'll add a TODO here and if we remember, we remember, and if we forget, we forget :D

packages/venia-ui/lib/RootComponents/Search/search.js
sirugh: Decided to stick with actions like this.

packages/venia-ui/lib/RootComponents/Search/search.js
```diff
@@ -37,8 +37,8 @@ const Search = props => {
         queryParameter: 'category'
     });
     const queryVariable = categoryId
-        ? { inputText, categoryId }
```
sirugh: Ah, this does need to be `inputText` though since it is the variable name in the gql query.

packages/venia-ui/lib/components/MiniCart/miniCart.js
```diff
@@ -9,31 +9,37 @@ import defaultClasses from './miniCart.css';
...
 import { mergeClasses } from '../../classify';
 import getCurrencyCode from '../../util/getCurrencyCode';
+import { useAppContext } from '@magento/peregrine/lib/context/app';
+import { useCartContext } from '@magento/peregrine/lib/context/cart';
+import { useCheckoutContext } from '@magento/peregrine/lib/context/checkout';
 
 const MiniCart = props => {
+    const [, { closeDrawer }] = useAppContext();
+    const [
+        cartState,
```
jimbo: Thanks for calling this `cartState` rather than `cart`, and updating the references throughout. I know that's minor, but I think it will help people trying to follow what's going on in these components.

jcalcaben:
There's some JSDocs that could be added to create reference docs, but I'm okay with it being merged without as long as we create a separate doc issue for adding them.

packages/venia-ui/lib/components/Checkout/__tests__/flow.spec.js
```diff
@@ -22,6 +23,48 @@ jest.mock('@magento/peregrine', () => {
...
+    const state = {
+        // availableShippingMethods,
+        // billingAddress,
+        // isSubmitting,
+        // paymentData,
+        // shippingAddress,
+        // shippingAddressError,
+        // shippingMethod,
+        // shippingTitle
+    };
```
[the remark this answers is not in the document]
sirugh: Nooooo

packages/venia-ui/lib/components/FilterModal/FilterFooter/filterFooter.js
```diff
@@ -1,113 +1,78 @@
...
 
-export default compose(
-    withRouter,
-    classify(defaultClasses),
-    connect(
-        mapStateToProps,
-        mapDispatchToProps
-    )
-)(FilterFooter);
+export default compose(withRouter)(FilterFooter);
```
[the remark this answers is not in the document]
sirugh: We do. Probably something we can do later.

packages/venia-ui/lib/components/FilterModal/FilterFooter/filterFooter.js
```diff
@@ -1,113 +1,78 @@
...
-            footer: PropTypes.string
-        }),
-        history: PropTypes.object,
-        filterClear: PropTypes.func,
-        chosenFilterOptions: PropTypes.object,
-        closeDrawer: PropTypes.func
-    };
+import { useCatalogContext } from '@magento/peregrine/lib/context/catalog';
+import { useAppContext } from '@magento/peregrine/lib/context/app';
+import { serialize } from '@magento/peregrine/lib/store/actions/catalog';
```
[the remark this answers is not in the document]
sirugh: Should we do that here?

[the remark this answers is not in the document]
sirugh: Ok I'll open an issue.

sirugh:
@jcalcaben can you open those issues? Not sure which specific things you want created since we are really just moving things around.

packages/venia-ui/lib/components/FilterModal/FilterFooter/filterFooter.js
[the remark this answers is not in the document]
sirugh: https://github.com/magento/pwa-studio/issues/1715

PR #1674, opened 2019-09-11
Pins dependency for npm-run to fix security vulnerability in sync-exec

### LINKED ISSUE #1645: [bug]: GitHub Security Alerts
Describe the bug
GitHub's automated security fixes has identified issues with four dependencies: `eslint-utils`, `open`, `diff`, `sync-exec`

https://github.com/magento/pwa-studio/network/alerts

### LINKED ISSUE #1656: [bug]: GitHub Security Alerts - `sync-exec`
GitHub's automated security fixes has identified issues with four dependencies:  `sync-exec`

https://github.com/magento/pwa-studio/network/alerts

supernova-at:
## Description

Pins our transitive dependency on `npm-run` to a version that does not depend on `sync-exec@0.6.2`, which [has a security vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2017-16024).

## Related Issue

Closes #1656 .

## Verification Steps

1. `yarn clean:all`
1. `yarn install`
1. Verify that `yarn list --pattern sync-exec` does not return any matches
1. Verify that our query validation plugin still works by running `yarn run validate-queries` and ensuring it is successful
1. Verify that `yarn build` succeeds

## Checklist


packages/graphql-cli-validate-magento-pwa-queries/package.json
```diff
@@ -32,5 +32,8 @@
   },
   "peerDependencies": {
     "graphql-cli": "^3"
+  },
+  "resolutions": {
+    "npm-run": "~5.0.0"
```
sirugh: A few questions:

* Why not pin to the "safe" version of `sync-exec` instead?
* I'm still seeing this in `yarn-lock`. Is that OK?
```sync-exec@^0.6.2:
  version "0.6.2"
  resolved "https://registry.yarnpkg.com/sync-exec/-/sync-exec-0.6.2.tgz#717d22cc53f0ce1def5594362f3a89a2ebb91105"
  integrity sha1-cX0izFPwzh3vVZQ2LzqJouu5EQU=
```

sirugh: Oh I'm dumb -- read the notice that there is _no_ safe version for sync-exec :P Still, if that's the case it shouldn't be in `yarn.lock` right?

sirugh:
I'm still seeing the package in list. Should we be updating lockfile?

```
~/code/pwa-studio remotes/origin/pr/1674
❯ yarn list --pattern sync-exec
yarn list v1.17.3
└─ sync-exec@0.6.2
✨  Done in 1.11s.
```

supernova-at:
> Should we be updating lockfile?

Hmm yeah, good point. Not sure why that didn't update, I'll dig into it.

packages/graphql-cli-validate-magento-pwa-queries/package.json
supernova-at: Looking into this now...

supernova-at:
Okay, I have no idea why this isn't working.

[According to proposed new documentation on the "resolutions" field](https://github.com/yarnpkg/website/pull/970/files) there are a few ways to pin the `npm-run` dependency:

<hr />

```
"resolutions": {
  "npm-run": "~5.0.0"
}
```

This would pin **our** direct depedency of `npm-run` to `~5.0.0`.
This option doesn't work for us because we don't depend on `npm-run` directly.

<hr />

```
"resolutions": {
  "**/npm-run": "~5.0.0"
}
```

This pins **all** depedencies of `npm-run` to `~5.0.0`, no matter where the dependency occurs in the tree.
This could work for us: it ensures that we'll never pull in a vulnerable version of `npm-run`.
It comes with the caveat that some future dependency that depends on `npm-run@4` could be incompatible with `npm-run@5` though.

<hr />

```
"resolutions": {
  "**/graphql-cli/npm-run": "~5.0.0"
}
```

This pins `graphql-cli`'s depedency of `npm-run` to `~5.0.0`.
This option could work for us too because `graphql-cli` is currently the only thing depending on `npm-run@4.1.2`.
I think this is our best bet (this PR takes this approach).

sirugh:
FWIW: [Old PR](https://github.com/Urigo/graphql-cli/pull/441) in gql-cli repo that should fix this -- looks like it's pretty stale. Not sure what's going on in that repo :P

sirugh:
@supernova-at after reading your comment I also agree that the third option is probably our best bet  as globally fixing `npm-run` could introduce a difficult-to-track bug later on. Let's try the direct `graphql-cli/npm-run` fix here. Maybe we should also consider some documentation somewhere in our codebase that explains why specific `resolutions` exist (since we can't comment in json).

supernova-at:
Aaaaaand it was just because `resolutions` needed to be added to the root `package.json`, not the sub-packages.

Works great now! 🎉

supernova-at:
@sirugh documentation added ✅

PR #1687, opened 2019-09-15
Fix allowing "addToCart" on configurable product, with unselected options

### LINKED ISSUE #1658: [bug]: Product Full Detail allows adding to cart without all required options selected
**Describe the bug**
A product should only be allowed to be added to cart if all the required options are selected. For example we require a color _AND_ a size to be selected, however currently if you select one of them then the `add to cart` button is enabled and you can click it. This will throw an error because the API requires both options.

**To Reproduce**
1. Go to a product page
2. Click a color or size, but not both.
3. Click `Add to cart`
4. See error in console.

**Expected behavior**
The button should not be enabled until _both_ options are selected.

**Screenshots**
[Image from Gyazo](https://gyazo.com/1d92498f8da673cca0a9205c554eb436)

**Possible solutions**
In `getIsMissingOptions` the `optionSelections.size` check is invalid. If only one option is picked it will show size 2 and one of the keys will have a value of `undefined`. While this check is valid (if the map is smaller than the possible options we are missing options) we should also check to make sure no _required_ values are `undefined`.

MKostiantyn:
## Description

Prevent set state on "handleSelectionChange" action in ProductOption component, when selection argument is empty.

## Related Issue

Closes #1658 

## Verification Steps

1. Go to the ProductConfigurable page with two options.
2. Choose only one option, button "AddToCart" should be disabled. 

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/ProductOptions/__tests__/option.spec.js
+++ b/packages/venia-ui/lib/components/ProductOptions/__tests__/option.spec.js
@@ -65,11 +65,12 @@ test('calls onSelectionChange function with attribute_id and selection', () => {
         onSelectionChange: onSelectionChangeMock
     };
     const component = testRenderer.create(<Option {...props} />);
+    const set = new Set(['test']);
 
-    component.root.children[0].instance.handleSelectionChange('test');
+    component.root.children[0].instance.handleSelectionChange(set);
     expect(onSelectionChangeMock).toHaveBeenCalledWith(
         defaultProps.attribute_id,
-        'test'
+        set
     );
 
     onSelectionChangeMock.mockReset();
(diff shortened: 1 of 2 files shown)


sirugh:
@MKostiantyn are you interested in continuing this work? I know MM Poland has just finished so you may be traveling. We will keep this open for 2 weeks and if we haven't heard from you will close it so that someone else can give it a shot.

sirugh:
Putting on `hold`. If we don't hear from the contributor by next Monday, 9/30/19, I'll close this.

sirugh:
Closing as we haven't heard back from you in a while. If you'd like to pick this up and work on it again please feel free!

PR #1691, opened 2019-09-16
Webpack client side caching.

### LINKED ISSUE #1623: [Research Story]: Audit Build Settings for Performance Improvements
**What?**
Audit the Webpack build configuration to ensure that it builds the smallest and fastest possible app code.

**Why?**
We need to understand performance issues in relation to our build process. Webpack is very configurable and it's possible we could achieve great performance gains here.

**How?**
We will play with build settings, rebuild, measure, experiment, and repeat, logging our results along the way.

We'll determine the best parameters for:
  - code splitting
  - optimization and minification settings
  - deduplication plugins
  - browser and device support without unnecessary polyfills
  - reducing built times for development and CI
  - tree shaking and dead code elimination

**Acceptance Criteria**
- Write-up of audit results and proposed changes
- Actions and next steps created in the form of GitHub issues

revanth0212:
## Description
This PR addresses concerns and solutions raised in #1623. 

## Related Issue
Closes #1623.

## Verification Steps
1. Build the venia code base using `yarn build`.
2. Stage the newly created build using `yarn stage:venia`.
3. Open a browser tab with cache cleared or try using an incognito tab.
4. Open network dev tools tab.
5. Now go to the venia instance using the URL that was provided in the terminal during stage process.
6. The first time you should see `client.[hash].js`, `runtime.[hash].js`, `vendors.[hash].js`, `registerSW.[hash].js` and `sw.[hash].js` downloaded apart from other files that are not of significance in this PR.
7. When you refresh the UI again, no files have to redownload. Meaning they have been cached the first time.
8. Now stop the staging server and run steps 1, 2 and 5 again. This time again since the code has not changed the build should not change and nothing should be downloaded from the server.
9. Again stop the staging server, try changing couple lines of code and redo steps 1, 2 and 5. This time only the client.[hash].js should be downloaded because the app build has changed.

Check out #1623 for screenshots and metrics.

## Checklist
* App logic should not have changed. These are code build changes.


sirugh:
Does this address all of the outcome of #1623? As 1623 was a research story I think it appropriate to spawn some sub-task/issues per specific item if not.

Also can you check why AWS is failing? You'll need to log in to the AWS console since we can't log the output in danger for security reasons.

revanth0212:
> Does this address all of the outcome of #1623? As 1623 was a research story I think it appropriate to spawn some sub-task/issues per specific item if not.

Yes, it does. But all of them are related to each other. That's why I created a PR with all the changes.

> Also can you check why AWS is failing? You'll need to log in to the AWS console since we can't log the output in danger for security reasons.

Sure will do 👍

revanth0212:
Develop:
https://www.webpagetest.org/result/190918_6P_680f0e368fcb85ce22158aca82d38d42/

3G Network Simulation:
https://www.webpagetest.org/result/190918_VM_d2674f44e6416aa4b96ef9541a6fa3cf/

This PR:
https://www.webpagetest.org/result/190918_NP_76bd0fdbbd99b06d7524a195bb013858/

3G Network Simulation:
https://www.webpagetest.org/result/190918_ZF_da34345fed22481ff9353661935fe569/

sirugh:
Does WPT test refreshes 😄

revanth0212:
> Does WPT test refreshes 😄

It takes forever sometimes 🤣

revanth0212:
Here is a breakdown of first render and multiple refreshes with the changes.

image

packages/pwa-buildpack/lib/Utilities/__tests__/createFileHash.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    'lib',
+    'Utilities',
+    '__tests__',
+    'sampleFile.txt'
+);
+
+test('Should return a promise that resolves to a string', async () => {
+    const returnValue = createFileHash(DEFAULT_SAMPLE_FILE_PATH);
+
+    expect(returnValue instanceof Promise).toBeTruthy();
```
supernova-at: Jest's [toBeInstanceOf](https://jestjs.io/docs/en/expect#tobeinstanceofclass) assertion reads nicer here (it uses `instanceOf` underneath).

packages/pwa-buildpack/lib/Utilities/__tests__/createFileHash.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    const hash = await returnValue;
+
+    expect(typeof hash).toBe('string');
+});
+
+test('Should return same hash if the contents of the file has not changed', async () => {
+    const oldValue = await createFileHash(DEFAULT_SAMPLE_FILE_PATH);
+    const newValue = await createFileHash(DEFAULT_SAMPLE_FILE_PATH);
+
+    expect(oldValue === newValue).toBeTruthy();
```
supernova-at: Check out the [toBe assertion](https://jestjs.io/docs/en/expect#tobevalue) for this one.

packages/pwa-buildpack/lib/Utilities/__tests__/createFileHash.spec.js
```diff
@@ -0,0 +1,54 @@
...
+});
+
+test('Should return different value if the contents of the same file have changed', async () => {
+    const oldValue = await createFileHash(DEFAULT_SAMPLE_FILE_PATH);
+
+    fs.appendFileSync(`${__dirname}/sampleFile.txt`, ' Venia here.');
+
+    const newValue = await createFileHash(DEFAULT_SAMPLE_FILE_PATH);
+
+    fs.truncateSync(`${__dirname}/sampleFile.txt`, 11);
```
supernova-at: We should mock the relevant methods of the `fs` and `crypto` modules instead of actually changing the file on disk here.

packages/pwa-buildpack/lib/Utilities/__tests__/createFileHash.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+    expect(oldValue === newValue).toBeFalsy();
+});
+
+test('Should not throw error if given file path is invalid. Instead should print error in console and resolve to empty string as hash', async () => {
+    const errorLogger = jest.fn();
+
+    console.error = errorLogger;
```
supernova-at: ⚠️ This will leak out into other tests.

```js
test('it leaks', () => {
    // This assertion fails, console.error is still a mocked function.
    expect(jest.isMockFunction(console.error)).toBe(false);
});
```

You can either clean this up manually or use [spyOn](https://jestjs.io/docs/en/jest-object#jestspyonobject-methodname) instead:

```js
const spy = jest.spyOn(console, 'error');

// do your test & assertions

spy.mockRestore();
```

See https://jestjs.io/docs/en/mock-function-api#mockfnmockrestore.

packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
```diff
@@ -291,26 +306,45 @@ async function configureWebpack({ context, vendor = [], special = {}, env }) {
...
+                            semicolons: false
+                        },
+                        keep_fnames: true
+                    }
+                })
+            ]
+        };
+        config.plugins.push(
+            new webpack.optimize.LimitChunkCountPlugin({
+                maxChunks: 5
```
supernova-at: Does this mean we will only ever split into five chunks?

If so, can you explain why we would want this? Seems pretty arbitrary at the moment.

packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
```diff
@@ -291,26 +306,45 @@ async function configureWebpack({ context, vendor = [], special = {}, env }) {
...
+            ...config.optimization,
+            moduleIds: 'hashed',
+            runtimeChunk: 'single',
+            splitChunks: {
+                cacheGroups: {
+                    vendor: {
+                        test: new RegExp(vendorTest),
+                        name: 'vendors',
+                        chunks: 'all'
+                    }
```
supernova-at: Since webpack configurations are pretty cryptic to begin with and we have the benefit of being in a JavaScript file here, would you mind adding some comments to this file indicating why we made these particular choices?

packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
```diff
@@ -90,18 +91,32 @@ async function configureWebpack({ context, vendor = [], special = {}, env }) {
...
+        SERVICE_WORKER_HASH: serviceWorkerFileHash
+    };
+
     const config = {
         mode,
         context, // Node global for the running script's directory
         entry: {
-            client: path.resolve(paths.src, 'index.js')
+            client: path.resolve(paths.src, 'index.js'),
+            registerSW: path.resolve(paths.src, 'registerSW.js')
```
supernova-at: What benefits do we get from splitting this out into its own entry point?

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
```diff
@@ -1,34 +1,54 @@
...
 
 // No longer modifiable since v4
-const SW_FILENAME = 'sw.js';
+let SERVICE_WORKER_FILENAME = 'sw.js';
+let SERVICE_WORKER_HASH = '';
+
+const DEFAULT_SW_FILE_PATH = path.join(
+    path.resolve(__dirname).split('/packages')[0],
+    'packages',
+    'venia-concept',
```
supernova-at: What about projects that aren't `venia-concept`? Like if someone uses the new CLI to start their own project - are they intended to use this plugin too?

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
```diff
@@ -1,34 +1,54 @@
...
+    'src',
+    'sw.js'
+);
+
+createFileHash(DEFAULT_SW_FILE_PATH).then(hash => {
```
supernova-at: Are we creating this hash here and in `configureWebpack.js`?

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
```diff
@@ -1,34 +1,54 @@
...
+    SERVICE_WORKER_HASH = hash;
+    SERVICE_WORKER_FILENAME = `sw.${hash}.js`;
+});
 
 class ServiceWorkerPlugin {
     constructor(config) {
         ServiceWorkerPlugin.validateOptions('ServiceWorkerPlugin', config);
         this.config = config;
     }
+
```
supernova-at: Thank you! No sarcasm, this makes it so much more readable 😝

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
```diff
@@ -1,34 +1,54 @@
...
             // TODO: (feature) autogenerate glob patterns from asset manifest
             globPatterns: ['**/*.{gif,jpg,png,svg}'],
 
             // activate the worker as soon as it reaches the waiting phase
             skipWaiting: true,
 
             // the max scope of a worker is its location
-            swDest: SW_FILENAME
+            swDest: SERVICE_WORKER_FILENAME
```
supernova-at: What does the comment above this line mean? What is "the max scope"?

packages/pwa-buildpack/lib/Utilities/createFileHash.js
```diff
@@ -0,0 +1,36 @@
...
+                    resolve('');
+                });
+
+                fd.pipe(hash);
+            } else {
+                throw new Error(`Error reading file ${filePath}`);
+            }
+        } catch (err) {
+            console.error(`Unable to create hash of ${filePath}`, err);
+            resolve('');
```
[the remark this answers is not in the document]
revanth0212: We shouldn't because this is not mandatory and should not stop the build process by throwing an error. If we are unable to create a hash, we have to create a random string and insert because the file name structure of `filename.[hash].fileextention` should be maintained for consistency.

packages/pwa-buildpack/lib/Utilities/__tests__/createFileHash.spec.js
revanth0212: Not sure how it missed that 👍

packages/pwa-buildpack/lib/Utilities/__tests__/createFileHash.spec.js
revanth0212: We are relying on Crypto and FS to create a hash and if we are mocking them, then we are not testing hash creation logic. We need to test if the file has changed, which will often happen while development, the hash changes.

supernova-at: I'm bringing it up mostly to highlight unit test best practices. As much as possible we don't want to be making actual network calls or reading from actual files on disk.

Might we be able to simplify and use [digest](https://nodejs.org/api/crypto.html#crypto_hash_digest_encoding) instead?

```
// createFileHash.js
function createFileHash(filePath) {
  const contents = fs.readFileSync(filePath, { encoding: 'utf8' });

  const hash = crypto.createHash('md5');
  hash.update(contents);
  return hash.digest('hex');
}
```

This way in our test we can mock `readFileSync` to just return a string - we trust that the `fs.readFileSync` works correctly, we don't need to test it - and then assert on the result of the function?

```
// createFileHash.spec.js

// mock fs.readFileSync to return 'abc' the first time and 'def' the second time.

const output1 = await createFileHash(DEFAULT_SAMPLE_FILE_PATH);
const output2 = await createFileHash(DEFAULT_SAMPLE_FILE_PATH);

expect(output1).not.toBe(output2);
```

This way we're still testing the "file contents changed" case, but we don't actually have to read from a file on disk :)

packages/pwa-buildpack/lib/Utilities/createFileHash.js
```diff
@@ -1,32 +1,34 @@
 const fs = require('fs');
 const crypto = require('crypto');
 
+const DEFAULT_RANDOM_STRING = crypto.randomBytes(8).toString('hex');
```
supernova-at: Ooh interesting, I've never seen `randomBytes` before 👍

supernova-at: Also an important comment keeps getting buried: https://github.com/magento/pwa-studio/pull/1691#discussion_r327638144

supernova-at:
An important comment keeps getting buried 😞 : https://github.com/magento/pwa-studio/pull/1691#discussion_r327638144

packages/pwa-buildpack/lib/Utilities/createFileHash.js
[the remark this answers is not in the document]
revanth0212: Yeah, I have changed the implementation recently to reflect the changes. Instead of 'something_else' it will be a random string for each call. This fallback will be just like the previous implementation because previously as well, we used to get a new file on every compile.

packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
revanth0212: `registerSW.js` has the registration part for the SW which will never change. So if we separate that out, it will only be requested once in the app's lifetime (assuming cache is not cleared).

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
revanth0212: If someone does that, will the folder name change to something else?

I was in an assumption that `venia-concept` will remain as its own folder because it is being used in multiple places in `pwa-buildpack`.

If that's not the case, first we will have to save the name of the project in a global variable somewhere are start using that in the places where we are using `venia-concept` right now.

Correct me if I am wrong on this.

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
revanth0212: Yes, this is in accordance with the implementation of the `ServierWorkerPlugin`. According to its implementation, you can use that with or without`injectManifestConfig`. If that is not provided while building, the plugin tries to use the default SW file, so we need the hash for the default SW file.

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
revanth0212: It is the prettier extension not me 😲

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
revanth0212: That was written by Zetlen a while ago but here is my take on that. According to the SW docs:

[One subtlety with the register() method is the location of the service worker file. You'll notice in this case that the service worker file is at the root of the domain. This means that the service worker's scope will be the entire origin. In other words, this service worker will receive fetch events for everything on this domain. If we register the service worker file at /example/sw.js, then the service worker would only see fetch events for pages whose URL starts with /example/ (i.e. /example/page1/, /example/page2/).](https://developers.google.com/web/fundamentals/primers/service-workers#register_a_service_worker)

So if we decode the above paragraph and apply to our code, we are setting the domain of the service worker as the location of the default service worker file which in our case is `venia-concept/src/sw.js`.

packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
revanth0212: Yeah, creating too many chunks can create HTTP overhead issues at the same time creating less number of them might create issues related to bundle size and caching. That is why the online community feels 5 is an ideal number. Also if we decide on doing the route based bundling, we will have to change this. This is a filler for now. If I remove this, webpack is creating 8-10 chunks on avg and that will hamper the app performance.

packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
revanth0212: Done 👍

packages/pwa-buildpack/lib/Utilities/createFileHash.js
```diff
@@ -0,0 +1,36 @@
...
+
+                fd.on('end', function() {
+                    hash.end();
+                    resolve(hash.read());
+                });
+
+                fd.on('error', function() {
+                    hash.end();
+                    console.error(`Unable to read file ${filePath}`);
+                    resolve('');
```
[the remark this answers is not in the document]
revanth0212: I tried doing that but had a hard time getting it done the first time when I was writing this function. But on reading an article about how errors are propagated in node which is diff than normal JS does to async and streams in node, I changed the code a bit and it is done.

packages/pwa-buildpack/lib/Utilities/__tests__/createFileHash.spec.js
revanth0212: The problem with readFileSync is first it eats a huge chunk of memory because it tries to load the whole file at a time and the second is it is sync, meaning it will stop the main thread execution till the file is read completely. On the other hand, if we are using streams like in createReadStream the memory footprint is very small.

Sample comparison:

image

On the other hand, mocking the whole stream object to be returned by the mocked createFileHash is too verbose and diminishes the benefit of using this function.

But I get your point, in the case of the 3rd test, I was creating a new file, editing it to check the hash change effect. We can achieve that with mocking without new files. I'll try changing that test if possible.

supernova-at: > mocking the whole stream object just to prove a point is not economical

Totally agree, I don't want to do that either.

packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
```diff
@@ -309,9 +326,24 @@ async function configureWebpack({ context, vendor = [], special = {}, env }) {
...
             ...config.optimization,
             moduleIds: 'hashed',
+            /**
+             * This will move the runtime configuration to
+             * its own bundle. Since runtime config tends to
+             * change on each compile even though the app logic
+             * doesn't, if not separated the whole client bundle
+             * needs to be downloaded. Separating them will only
+             * download runtime bundle and use the cached client code.
+             */
```
supernova-at: ❤️

packages/pwa-buildpack/lib/WebpackTools/plugins/ServiceWorkerPlugin.js
supernova-at: I think this is fine for now but I unresolved the conversation in case we want to continue discussing it.

PR #1694, opened 2019-09-17
Merge directory state and actions into checkout

### LINKED ISSUE #1684: Merge directory actions/state/etc into checkout slice
The directory state slice should be merged into the checkout slice in Peregrine

sirugh:
## Description

Merges directory state, actions, and reducers into checkout. This gets us closer to being able to pull off `countries` data from the checkout slice.

## Related Issue

Closes #1684.

## Verification Steps

1. Open checkout, you should notice a `GET_COUNTRIES` action in console.
2. Complete checkout, you should get no errors.
3. Open another checkout. You should _not_ see a `GET_COUNTRIES` action (because we preserve the `countries` state.)

## Checklist


packages/peregrine/lib/store/reducers/checkout.js
```diff
@@ -166,7 +177,21 @@ const reducerMap = {
...
+        // Avoid refetching countries.
+        if (state.countries) {
+            preservedState.countries = {
+                ...state.countries
+            };
+        }
+
+        return {
+            ...initialState,
+            preservedState
```
jimbo: Having a `preservedState` property strikes me as odd. Did you mean to spread `preservedState` onto the return value instead?

sirugh: OOps

sirugh:
```suggestion
            ...preservedState
```

Issue #1632, opened 2019-09-04
Replace remaining redux selectors with derived state values


jimbo:
**Is your feature request related to a problem? Please describe.**

Putting selector functions for getting derived state from a redux store is not an uncommon pattern, but it's not one we ever endorsed or used widely. Now that we're providing redux state from context, with a single connector per slice, we shouldn't need them.

**Describe the solution you'd like**

- For derived state that's used widely, append it directly to the context value.
- For derived state that's only used in a few places, let the component calculate it.

**Describe alternatives you've considered**

This is the alternative to the current pattern.

**Please let us know what packages this feature is in regards to:**
- [ ] `venia-concept`
- [ ] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
Added to Modular Component Library milestone as a follow-on/nice-to-have.

sirugh:
>append it directly to the context value.
@jimbo when you say this do you mean do this work in the context creator like [here](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/context/user.js#L21) or in the reducer's initialState?

jimbo:
> > For derived state that's used widely, append it directly to the context value.
> 
> @jimbo when you say this do you mean do this work in the context creator like [here](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/context/user.js#L21) or in the reducer's initialState?

The former.

```es6
// selectors/cart.js
export const isCartEmpty = cartState =>
    !cartState.details.items || cartState.details.items.length === 0

// context/cart.js
const cartApi = useMemo(/* ... */)

const derivedCartState = useMemo(
  () => ({
    ...cartState,
    isEmpty: isCartEmpty(cartState)
  }),
  [cartState]
)

const contextValue = useMemo(
  () => [derivedCartState, cartApi],
  [cartApi, derivedCartState]
)
```

PR #1700, opened 2019-09-18
Remove unused purchase history and purchase details pages

### LINKED ISSUE #1699: Remove purchase history/details
These are unused components. You cannot even navigate to them right now. It's more work to maintain them as is given all the change we've gone through with Peregrine. If we want to reimplement these in the future we can certainly look back at the code but IMO we should delete it so we don't have to maintain it anymore.

sirugh:
## Description

The Purchase History and Purchase Details pages are unused. The async actions are mocked and the there is no way to view the UI. If we want to ever re-implement these functionalities or features we can revert this PR or pull from it where applicable but

## Related Issue

Closes #1699

## Verification Steps

Make sure I deleted everything :D

## Checklist


sirugh:
Closing this cause I'll do it with #1632

PR #1702, opened 2019-09-18
[PB] Block component

cspruiell:
## Description
Introduce Block component for Page Builder Block content type

## Related Issue
https://jira.corp.magento.com/browse/PB-20

## Performance Results
https://wiki.corp.magento.com/display/COREENG/Block

## Verification Steps
- [ ] Check that each property for each content type appearance renders correctly on PWA Venia storefront in desktop and mobile.
- [ ] Compare Venia PWA rendering of content type with Luma to ensure the experience is the same or better for each property of each appearance in desktop and mobile.
- [ ] Check each supported browser and on mobile that is supported according to PWA documentation.
- [ ] Acceptable performance has been validated


davemacaulay:
@cspruiell please see part of description outlining how we should update `<CmsBlock />` venia-ui component and use that within our `<Block />` component.

> Venia-ui already contains a <CmsBlock /> component which should be updated and used instead of creating a new component (the PoC contains a new component, where as the final implementation should not). It maybe necessary to maintain the <Block /> component to wrap <CmsBlock /> when used within Page Builder due to our advanced configuration.

PR #1703, opened 2019-09-18
Peregrine implementation/selectors be gone

### LINKED ISSUE #1632: Replace remaining redux selectors with derived state values

### LINKED ISSUE #1699: Remove purchase history/details

sirugh:
## Description

Removing selector code from our repo.

`cart` selector is merged with state.
`checkout` selector is now part of the action which used it directly.
`user` selector wasn't used (part of `MyAccountMenuPage` which was replaced by `MyAccount`)
`purchaseHistory` selector wasn't used (part of `PurchaseHistory` and that is going bye bye)

## Related Issue

Closes #1632 

## Verification Steps

Make sure I deleted everything properly. 
Make sure cart and checkout still function.

## Checklist


packages/peregrine/lib/store/actions/checkout/asyncActions.js
```diff
@@ -301,7 +300,18 @@ export const submitOrder = () =>
 
 export const createAccount = history => async (dispatch, getState) => {
     const { checkout } = getState();
-    const accountInfo = getAccountInformation(checkout);
+
+    const {
+        email,
+        firstname: firstName,
+        lastname: lastName
+    } = checkout.receipt.order.billingAddress;
```
jimbo:
```suggestion
    } = checkout.receipt.order.billing_address;
```

Caught a bug here.

Issue #1662, opened 2019-09-10
Fix Simple Product Heuristic Fragment Matcher


supernova-at:
By default, `apollo-cache-inmemory` uses a heuristic fragment matcher. Since PWA Studio is using fragments on unions and interfaces, we will need to use an `IntrospectionFragmentMatcher`.

Apollo provides a guide on how to accomplish this here:
https://www.apollographql.com/docs/react/advanced/fragments/#fragments-on-unions-and-interfaces

**Acceptance Crtieria**
1. The console warning that appears in development when viewing a simple product PDP no longer appears

Jordaneisenburger:
@supernova-at I've just implemented this solution in our project, I can make a PR if you're interested ?

sirugh:
@Jordaneisenburger definitely, or we can talk about it in the public channel first, or in here! Whatever works for you.

Jordaneisenburger:
> @Jordaneisenburger definitely, or we can talk about it in the public channel first, or in here! Whatever works for you.

@sirugh I'll make a PR so that we have something tangible to talk about

Jordaneisenburger:
@sirugh PR open at #1707

supernova-at:
I took this one over and have updated #1707.

PR #1707, opened 2019-09-19
Bugfix/heuristic fragment matcher

### LINKED ISSUE #1474: [bug]: WARNING: heuristic fragment matching going on!
**Describe the bug**
When you visit a simple product detail page you will see this error in the chrome console
`WARNING: heuristic fragment matching going on!`

Not 100% verified but when you visit a category page with multiple simple products the error gets thrown multiple times preventing the page from loading 

 - Device [Mac]:
 - Browser [Chrome]:
 - Browser Version [76]:
 - Magento Version: 2.3.2
 - PWA Studio Version: 3.0.0
 - NPM version `6.4.1`:
 - Node Version `10.15.1`:

- [x] `venia-concept`
- [x] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

Jordaneisenburger:
## Description

This PR adds the `IntrospectionFragmentMatcher` to resolve the heuristic fragment matcher warning.

It:
  * Issues an introspection query to the GraphQL schema at build time to get the union and interface types
  * It passes those to the bundle via webpack's `DefinePlugin`
  * The app uses them to initialize the `IntrospectionFragmentMatcher` at run time

As such, there is also some refactoring of buildpack to make issuing GraphQL requests easier 🎉 .

## Related Issue
Related issues #1474 & #1662 

Closes #1474

## Verification Steps
1. Go to /augusta-earrings.html
2. Open chrome console
3. See console warning
4. update you code base with this PR and run  `yarn run build` in project root
5. Visit /augusta-earrings.html again and, the warning should be gone.


packages/venia-concept/generateGraphQlFragmentTypes.js
```diff
@@ -0,0 +1,43 @@
+const fetch = require('node-fetch');
+const fs = require('fs');
+
+// Right now I've hardcoded this url because we only have process.env.MAGENTO_BACKEND_URL on run time
```
sirugh: Really? I thought when you run `yarn run build` it generates the `.env`. Can you not access that here?

jimbo:
@Jordaneisenburger Wow, this is so valuable! Thanks for tackling this issue; it's been bothering us lately.

Let's see if we can't find a way to get the backend URL from `.env`.

packages/venia-concept/generateGraphQlFragmentTypes.js
supernova-at: If this is done at build time I'd like for it to live in `buildpack`.
We have some basic utilities there already for communicating with the GraphQL backend.

packages/venia-concept/generateGraphQlFragmentTypes.js
```diff
@@ -0,0 +1,43 @@
...
+    })
+})
+    .then(result => result.json())
+    .then(result => {
+        // here we're filtering out any type information unrelated to unions or interfaces
+        const filteredData = result.data.__schema.types.filter(
+            type => type.possibleTypes !== null
+        );
+        result.data.__schema.types = filteredData;
+        fs.writeFile(
```
supernova-at: If we move the logic to buildpack we could inline the fragment types into the bundle using the `DefinePlugin` or similar to avoid having to read a file at run time.

packages/venia-concept/generateGraphQlFragmentTypes.js
Jordaneisenburger: @sirugh nevermind managed to get the env config like so:
```
const config = require('dotenv').config();
const MAGENTO_BACKEND_URL = config.parsed.MAGENTO_BACKEND_URL;
```

Jordaneisenburger: Yeah i guess that makes sense, I'm not too experienced with this part of Studio but i'll try to make the change.

supernova-at:
I moved this build-time logic to `buildpack`, this is ready for review.

packages/pwa-buildpack/lib/Utilities/graphQL.js
```diff
@@ -0,0 +1,63 @@
+const fetch = require('node-fetch');
```
supernova-at: So GitHub didn't pick it up, but I renamed `fetcherUtils` to `graphQL` (and their associated test files).

This is instrumentation to make GraphQL calls so I think this name is more indicative of what this module's job is.

packages/pwa-buildpack/lib/Utilities/graphQL.js
```diff
@@ -0,0 +1,63 @@
...
+                'Accept-Encoding': 'gzip'
+            },
+            body: JSON.stringify({ query })
+        }
+    )
+        .then(result => result.json())
+        .then(json => json.data)
+        .catch(err => {
+            console.error(err);
+            throw err;
```
supernova-at: Note that instead of returning an empty string here we are now re-throwing the caught error.

This allows the caller to determine how they want to handle the error case in a more idiomatic way `try / catch` instead of having to check success cases against an empty string.

packages/pwa-buildpack/lib/queries/getSchemaTypes.graphql
```diff
@@ -0,0 +1,12 @@
+# Introspect the GraphQL schema to obtain information about its types.
```
supernova-at: Note that `buildpack` now has its own `queries` folder with `.graphql` files in it.

This allows us to not have to inline large GraphQL strings 🎉

packages/pwa-buildpack/lib/queries/index.js
```diff
@@ -0,0 +1,12 @@
...
+
+// Import all the build-time queries.
+const getMediaUrl = requireGraphQL('../queries/getStoreMediaUrl');
+const getSchemaTypes = requireGraphQL('../queries/getSchemaTypes');
+
+// Export the queries for use by the rest of buildpack.
+module.exports = {
+    getMediaUrl,
+    getSchemaTypes
+};
```
supernova-at: This file isn't _strictly_ necessary, and we'll have to add all new `.graphql` files here to make them available for the rest of `buildpack`, but we do need to `require` the `.graphql` files in a different way (using `requireGraphQL`) so this consolidates that logic.

packages/pwa-buildpack/package.json
```diff
@@ -52,6 +52,7 @@
     "pkg-dir": "~4.1.0",
     "portscanner": "~2.2.0",
     "redis": "~2.8.0",
+    "require-graphql-file": "~1.0.1",
```
supernova-at: Needed to be able to `require` `.graphql` files at build time.

sirugh: Seems like it just lets you leave off the `.graphql` which doesn't seem that beneficial.
If we need to use this utility we should probably just [roll our own](https://github.com/Ikos9/require-graphql-file/blob/master/src/index.js) as this was last published 2 years ago.

supernova-at: I definitely thought about rolling our own, yeah. It's just reading the contents of the file and returning back a string.

I can do that if you want.

packages/pwa-buildpack/lib/queries/index.js
revanth0212: I would prefer that. Having a simple index.js makes things simpler for the future.

packages/pwa-buildpack/package.json
revanth0212: We can do it ourselves if all it does is the conversion. We can use `fs` for this.

sirugh: It turns `myFileName` into `myFileName.graphql` and then reads that file from the path. I wonder if `require('myFileName.graphql')` would just work.

supernova-at: Unfortunately `require('myFileName.graphql')` doesn't work.

I'm not an expert on `require` but I just need the contents of the file as a string, and I think `require` is trying to turn it into an object or something that can be understood by other code.

Anyway, I removed this dependency and wrote our own 👍

packages/pwa-buildpack/lib/queries/index.js
```diff
@@ -1,9 +1,22 @@
...
+
+/**
+ * Normal `require` doesn't know what to do with .graphql files, so this helper function
+ * simply imports their contents as a string.
+ * @see https://github.com/apollographql/apollo-server/issues/1175#issuecomment-397257339.
+ *
+ * @param   {String} filepath - A relative path to a .graphql file to read.
+ * @returns {String} - The contents of the file as a string.
+ */
+const requireGraphQL = filePath => {
```
revanth0212: Can we call this `requireFile` instead of `requireGraphQL`? Also, move it to its own file and have some tests around it. This can be useful even for non GraphQL files.

supernova-at: Honestly I considered that but couldn't think of another use case - do you have one in mind?

I left it here for now thinking that if we want to extract / generalize it later we can.

supernova-at:
> But I remember @zetlen wanted them to be implemented as part of a webpack plugins rather than utility functions.

Maybe we can all chat about this, I'd like to understand more about why Webpack plugins are preferred.

> `configureWebpack` should not worry about fetching data. Someone should do that and let `configureWebpack` deal with using them the expected way. Separation of concerns.

I'm not sure I agree here, `configureWebpack`'s job is to construct a webpack configuration file. In order to do that it has to fetch some data first.

>  Instead, we can move them to a separate file, make the calls as part of a Promise.all and return the combined result, mainly for maintenance reasons.

A function whose job it is to construct an entire webpack configuration object is going to get large, yes. I'm open to ideas here too. 

But I think 

```js
const [
  result1,
  result2,
  result3,
  // ...
  resultN
] =  await Promise.all([
  asyncCall1,
  asyncCall2,
  asyncCall3,
  // ...
  asyncCallN
]);
```

is no different than and actually less readable than

```js
const result1 = await asyncCall1();
const result2 = await asyncCall2();
const result3 = await asyncCall3();
// ...
const resultN = await asyncCallN();
```

packages/pwa-buildpack/lib/queries/index.js
sirugh: Optimize later, IMO :)

jest.config.js
```diff
@@ -79,7 +79,115 @@ const testVenia = inPackage => ({
     // @magento namespaced packages like Peregrine and Venia UI as well, when
     // it's testing Venia. That way, changes in sibling packages don't require a
     // full compile.
-    transformIgnorePatterns: ['node_modules/(?!@magento/)']
+    transformIgnorePatterns: ['node_modules/(?!@magento/)'],
+    globals: {
+        TEST_GLOBAL: 'true',
+        UNION_AND_INTERFACE_TYPES: {
```
sirugh: This PR introduced a dependency on globals in a way that uncovered a bug in our config. The previous location of the `globals` configuration did not actually set globals and the test for `STORE_NAME` hardcoded it in the test spec rather than relying on the config (probably because it was broken).

This now properly creates serializable globals for Jest.

PR #1727, opened 2019-09-20
Create myAccount mixin

sirugh:
## Description

Create and use a myAccount mixin `useMyAccount`.

## Related Issue

Closes #ISSUE_NUMBER.

## Verification Steps

1. Make sure myAccount page is still functional (in the left nav)

## Checklist


packages/peregrine/lib/mixins/MyAccount/useMyAccount.js
```diff
@@ -0,0 +1,27 @@
...
+const useMyAccount = props => {
+    const { signOut } = props;
+    const [{ currentUser }] = useUserContext();
+    const { email, firstname, lastname } = currentUser;
+    const name = `${firstname} ${lastname}`.trim() || DEFAULT_TITLE;
+    const title = email ? name : UNAUTHED_TITLE;
+    const subtitle = email ? email : UNAUTHED_SUBTITLE;
+
+    const handleSignOut = useCallback(() => {
+        signOut({ history: window.history });
```
jimbo: We'll want to use the router history, which we can hopefully get from context. There's not much point in passing around window properties anyway.

packages/peregrine/lib/mixins/MyAccount/useMyAccount.js
```diff
@@ -0,0 +1,27 @@
...
+    }, [signOut]);
+
+    return {
+        handleSignOut,
+        subtitle,
+        title
+    };
+};
+
+export default useMyAccount;
```
jimbo: At least within PWA Studio, I've been making sure to export hooks as named exports, rather than `default`. Since a hook's name must start with `use` in order to work correctly, I want to encourage users to retain the name we give it, and so I put a tiny obstacle in the way of a user renaming it on import.

packages/venia-ui/lib/components/MyAccount/myAccount.js
```diff
@@ -1,29 +1,21 @@
...
-    const { email, firstname, lastname } = user;
-    const name = `${firstname} ${lastname}`.trim() || DEFAULT_TITLE;
-    const title = email ? name : UNAUTHED_TITLE;
-    const subtitle = email ? email : UNAUTHED_SUBTITLE;
     const classes = mergeClasses(defaultClasses, props.classes);
 
-    const handleSignOut = useCallback(() => {
-        signOut({ history: window.history });
-    }, [signOut]);
+    const { handleSignOut, subtitle, title } = useMyAccount({
```
jimbo: Some mixins are going to return a lot of props, and like this one, they may take a lot of arguments as well. Destructuring is kind of tough to read in those scenarios, so I recommend splitting it into two statements.

```es6
const mixinProps = useMyAccount({
  multiline: "really",
  long: "arguments",
  make: "it hard to read"
})
const {
  handleSignOut,
  subtitle,
  title: renamedTitle // avoid clash with variable in closure
} = mixinProps
```

packages/peregrine/lib/mixins/MyAccount/useMyAccount.js
sirugh: Sure -- I just didn't change the api for it. This is a zero-functional change.

packages/peregrine/lib/mixins/MyAccount/useMyAccount.js
sirugh: Oh I noticed that. I will do that.

packages/venia-ui/lib/components/MyAccount/myAccount.js
sirugh: Sure. I like this idea.

PR #1745, opened 2019-09-23
Creates signIn mixin

sirugh:


packages/peregrine/lib/mixins/SignIn/useSignIn.js
```diff
@@ -0,0 +1,57 @@
+import { useCallback, useRef } from 'react';
+import { useUserContext } from '../../context/user';
+
+// Note: we can't access the actual message that comes back from the server
+// without doing some fragile string manipulation. Hardcoded for now.
+const ERROR_MESSAGE =
+    'The account sign-in was incorrect or your account is disabled temporarily. Please wait and try again later.';
+
+export const useSignIn = props => {
+    const { setDefaultUsername, showCreateAccount, showForgotPassword } = props;
```
jimbo: Custom hooks such as `useSignIn` don't support prop types, so we won't be able to port them directly. I'm concerned about losing that type information, though; maybe we should just bring them over and put them in a comment, then circle back eventually and convert them to JSDoc?

packages/peregrine/lib/mixins/SignIn/useSignIn.js
```diff
@@ -0,0 +1,57 @@
...
+        showCreateAccount();
+    }, [setDefaultUsername, showCreateAccount]);
+
+    return {
+        errorMessage,
+        formRef,
+        handleCreateAccount,
+        handleForgotPassword,
+        handleSubmit,
+        isBusy: isGettingDetails || isSigningIn
```
jimbo: This is good. 👍

packages/peregrine/lib/mixins/SignIn/useSignIn.js
sirugh: Yea -- this is something I was concerned about as well. I haven't been touching the `propTypes` on the components since the validation will still function. IMO we should not document in multiple places.

PR #1754, opened 2019-09-23
Add useCategoryTree talon

jimbo:
## Description

Add a `useCategoryTree` mixin to Peregrine, and use it in Venia's `CategoryTree` component.


packages/venia-ui/lib/components/CategoryTree/categoryTree.js
```diff
@@ -17,33 +17,20 @@ const Tree = props => {
...
-    }, [categoryId, runQuery]);
-
-    // update redux with fetched categories
-    useEffect(() => {
-        if (data && data.category) {
-            updateCategories(data.category);
-        }
-    }, [data, updateCategories]);
+    const mixinProps = useCategoryTree({
+        categories,
```
sirugh: I like that you're being explicit here about what you pass to the hook.

packages/venia-ui/lib/components/CategoryTree/categoryTree.js
```diff
@@ -17,33 +17,20 @@ const Tree = props => {
...
+        categoryId,
+        query: MENU_QUERY,
```
sirugh: I think this is a good choice -- provide query via argument rather than importing within hook.

packages/peregrine/lib/talons/CategoryTree/index.js
```diff
@@ -0,0 +1,3 @@
+export { useCategoryBranch } from './useCategoryBranch';
```
sirugh: Once we're done with the peregrine refactors I'd like to have one PR where we go through and create these indexes and exports. Right now we're importing with full paths in `venia-ui` but we can clean all that up with proper exports.

packages/peregrine/lib/talons/CategoryTree/useCategoryBranch.js
```diff
@@ -0,0 +1,25 @@
+import { useCallback } from 'react';
+
+/**
+ * Returns props necessary to render a CategoryBranch component.
```
sirugh: I love these comments. I'm going to go through and update my existing PRs with doc comments as well.

packages/venia-ui/lib/components/CategoryTree/categoryTree.js
jimbo: Yeah, I think we have to let people customize the query or else these hooks will quickly lose value. That said, we'll have to find a way to set expectations for what kind of query should be passed in.

packages/peregrine/lib/talons/CategoryTree/useCategoryLeaf.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @param {object} props
+ * @param {function} props.onNavigate - callback to fire on link click
+ * @return {{ handleClick: function }}
+ */
+export const useCategoryLeaf = props => {
+    const { onNavigate } = props;
+
+    const handleClick = useCallback(() => {
+        onNavigate();
```
sirugh: This isn't doing anything but wrapping/renaming the function. I think we don't need this talon.

packages/venia-ui/lib/components/CategoryTree/categoryTree.js
```diff
@@ -17,33 +17,20 @@ const Tree = props => {
...
-    const { children } = rootCategory || {};
+    const { childCategories } = talonProps;
+    const classes = mergeClasses(defaultClasses, props.classes);
 
     // for each child category, render a direct link if it has no children
     // otherwise render a branch
-    const branches = Array.from((rootCategory && children) || [], id => {
-        const category = categories[id];
-        const isLeaf = category.children_count === '0';
+    const branches = Array.from(childCategories, childCategory => {
```
sirugh: I like that you made this more clear -- we iterate over children categories rather than the `rootCategory && children`. That said, can we just use `forEach` instead of `Array.from`?

packages/peregrine/lib/talons/CategoryTree/useCategoryTree.js
```diff
@@ -0,0 +1,62 @@
...
+            const category = categories[id];
+            const isLeaf = category.children_count === '0';
+
+            childCategories.set(id, { category, isLeaf });
+        }
+
+        return childCategories;
+    }, [categories, children]);
+
+    return { childCategories };
```
sirugh: I'm a big fan of not returning the pure `data` object of these query calls as I think I've done in some of my other talon PRs. This is good and I'll be using this going forward.

packages/peregrine/lib/talons/CategoryTree/useCategoryLeaf.js
jimbo: In general, a callback received as a prop shouldn't get directly applied as an event handler. I think it's important to wrap the event handler so that `event` doesn't get passed to `onNavigate`; `onNavigate` should be called with no arguments.

packages/venia-ui/lib/components/CategoryTree/categoryTree.js
jimbo: Hm? To be clear, this code is mapping one iterable (a `Map` of objects) to another (an array of React elements). `Array.from` is the right choice here because it's a mapping function that works on any iterable, which is what we need when working with a `Map`.

sirugh: Oh, yes of course -- you need to return the iterable. ignore me :D

packages/peregrine/lib/talons/CategoryTree/useCategoryLeaf.js
sirugh: OK -- in that case keep your eye out on my PRs cause I think I did not wrap these callbacks in some of my talons.

PR #1755, opened 2019-09-23
Create categoryTile/categoryList talon

sirugh:
Refactor of `categoryList` and `categoryTile` to function component along with their...talons.

Requires a full review plz.


packages/peregrine/lib/mixins/CategoryList/useCategoryTile.js
```diff
@@ -0,0 +1,39 @@
+import { useMemo } from 'react';
+
+// TODO: get categoryUrlSuffix from graphql storeOptions when it is ready
+const categoryUrlSuffix = '.html';
+const previewImageSize = 480;
+
+export const useCategoryTile = props => {
+    const { item, resourceUrl } = props;
```
sirugh: Big question: If we want to utilize a dynamic `resourceUrl` function that can be overwritten by users of the library we will have to either pass it in as a callback/prop or figure out a mechanism by which to provide this in `peregrine` via drivers.

Thoughts?

packages/venia-ui/lib/components/CategoryList/categoryList.js
```diff
@@ -1,96 +1,90 @@
...
 
-        return title ? (
-            <div className={classes.header}>
-                <h2 className={classes.title}>
-                    <span>{title}</span>
-                </h2>
-            </div>
-        ) : null;
-    }
+    const { data, error, loading } = mixinProps;
```
sirugh: I could probably make this simpler by just returning the error and the normalized data, `categoryChildren`.

packages/peregrine/lib/talons/CategoryList/useCategoryList.js
```diff
@@ -0,0 +1,39 @@
+import { useEffect } from 'react';
+import { useQuery } from '../../hooks/useQuery';
+
+/**
+ * Returns props necessary to render a CategoryList component.
+ *
+ * @param {object} props
+ * @param {object} props.query - category data
+ * @param {string} props.id - category id
+ * @return {{ data: object, error: boolean, loading: boolean }}
```
jimbo: Is `error` a boolean or an `Error` object?

packages/venia-ui/lib/components/CategoryList/categoryList.js
```diff
@@ -1,96 +1,90 @@
...
+            })
+        }
     };
+};
 
-    get header() {
-        const { title, classes } = this.props;
+const CategoryList = props => {
+    const { id, title } = props;
+    const mixinProps = useCategoryList({
```
jimbo:
```suggestion
    const talonProps = useCategoryList({
```

packages/peregrine/lib/talons/CategoryList/useCategoryList.js
sirugh: I'll look and see how we're using it in `categoryList` and update the output accordingly.

packages/peregrine/lib/talons/CategoryList/useCategoryList.js
```diff
@@ -32,8 +32,8 @@ export const useCategoryList = props => {
     }, [runQuery, setLoading, id]);
 
     return {
-        data,
         error,
-        loading
+        childCategories:
```
jimbo: This is a good change, in line with your other comment. Stylistically, I do think you should do calculations outside of the return block, but it's not a big deal.

Also I think we should still return `loading`, since a well-designed component would use it. Just because ours doesn't use it doesn't mean we should exclude it.

PR #1764, opened 2019-09-24
Lazy Load SearchBar correctly

### LINKED ISSUE #1742: Fix Lazy Loading SearchBar
# Overview

On initial load of the application, one specific instance of the `Route` component takes an order of magnitude more time than other components to load:

When we drill down into this 

We can see that it is the [call to Route from Header](https://github.com/magento/pwa-studio/blob/release/4.0/packages/venia-ui/lib/components/Header/header.js#L68):

```jsx
<Suspense fallback={searchOpen ? suspenseFallback : null}>
    <Route
        render={({ history, location }) => (
            <SearchBar
                isOpen={searchOpen}
                history={history}
                location={location}
            />
        )}
    />
</Suspense>
```

# The Ask

## Initial Page Load

One thing that immediately jumps out is that this component is needed on the initial page load but it is being lazy loaded via `React.lazy` and `Suspense`. To decrease application load time, it should be available immediately for rendering.

## Subsequent Renders

The other thing that is concerning is why this code is being executed in the first place. On initial load, the `searchOpen` boolean should be `false` and the `SearchBar` should only be loaded in when it turns `true`.

But when profiling, after the initial `mount` of the application we see a `Suspense` update:

And in fact here's where we can realize the largest benefit: this code is actually executed multiple times as you interact with the application - here's an example after opening the MiniCart drawer:

_There's our guy! Notice the `SearchBar` update occurs._

supernova-at:
## Description

When the search icon

search_icon

is clicked, the `SearchBar` component

Screen Shot 2019-09-24 at 1 12 14 PM

is intended to be lazy loaded into the application.

Unfortunately, while we were properly code splitting the `SearchBar` out into its own chunk, we didn't have our presentation logic set up quite right and we ended up always downloading the `SearchBar` chunk on initial load (read: not lazy loading).

This PR makes it so that the `SearchBar` is actually lazy loaded in and downloaded only after the search icon is clicked and search is open.

## Related Issue

Closes #1742 .

#### SearchBar is Code Split

1. `yarn stats:venia`
2. Open the search panel and search for `SearchBar`

Screen Shot 2019-09-24 at 1 24 40 PM

3. Verify that it is split out into its own chunk

Screen Shot 2019-09-24 at 11 28 30 AM

#### SearchBar is lazy loaded

1. `yarn watch:venia`
1. Open the `Network` tab
1. Verify that no chunks are included in the initial homepage load (chunks have filenames `1-[hash].js`, `2-[hash].js`, etc.
1. Click the search icon to open search

search_icon

5. See that the `Network` tab shows that a chunk was downloaded

Screen Shot 2019-09-24 at 1 28 52 PM

6. Verify that the chunk is in fact the `SearchBar` chunk by searching the downloaded file's contents for "SearchBar". It should appear many times.

## Checklist


packages/venia-ui/lib/components/Header/header.js
```diff
@@ -29,13 +29,26 @@ const Header = props => {
...
-    const suspenseFallback = (
+    const searchBarFallback = (
         <div className={classes.searchFallback}>
             <div className={classes.input}>
                 <div className={classes.loader} />
             </div>
         </div>
     );
+    const searchBar = searchOpen ? (
+        <Suspense fallback={searchBarFallback}>
```
supernova-at: 1. L40: The `searchBarFallback` is not conditional - it's always the fallback when the `SearchBar` hasn't been downloaded yet.

2. L39: We only ever include `<SearchBar>` in the rendered output when `searchOpen` is `true`. This will actually make it lazy load.

PR #1768, opened 2019-09-24
[PB] WYSIWYG Image optimization in PWA

davemacaulay:
## Description
WYSIWYG Image optimization in PWA

This PR enables WYSIWYG components to use the same sharp express back-end image optimization system as catalog images currently uses. It also opens the API for optimization options to the individual components 

## Related Issue
https://jira.corp.magento.com/browse/PB-5

## Related PR
https://github.com/magento-research/express-sharp/pull/4

## Verification Steps
1. Validate all existing Product images are still optimized
2. Verify row & column background images are resized and optimized
3. Verify optimization works for parallax images


packages/pwa-buildpack/lib/Utilities/addImgOptMiddleware.js
```diff
@@ -92,11 +103,21 @@ https://github.com/nodejs/node-gyp#installation`
             //   https://github.com/magento-research/express-sharp/blob/develop/lib/transform.js#L23
             //   https://sharp.pixelplumbing.com/en/stable/api-resize/
             if (height) {
-                rewrittenUrl += `/${height}`;
-                params.set('crop', true);
+                params.set('height', height);
+                if (!crop) {
```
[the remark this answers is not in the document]
danmooney2: `crop = true` as replacement for 107-109 comes to mind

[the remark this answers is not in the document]
davemacaulay: I've updated the logic to handle this in an easier to understand way:
```
            if (crop) {
                params.set('crop', crop);
            } else if (width && height) {
                // If we received height and width we should force crop since our
                // implementation of express sharp defaults fit to "outside" if crop
                // is falsy. `outside` sizes the image, retaining the aspect ratio
                // but may fall "outside" the desired height or width. `cover`
                // retains the aspect ratio like `outside` but clips to fit desired
                // height and width.
                //   https://github.com/magento-research/express-sharp/blob/develop/lib/transform.js#L23
                //   https://sharp.pixelplumbing.com/en/stable/api-resize/
                params.set('crop', true);
            }
```

packages/pwa-buildpack/lib/Utilities/addImgOptMiddleware.js
```diff
@@ -22,7 +22,9 @@ try {
  * UPWARD, and then used in various places: here, the UPWARD resolution
  * path itself, the `makeURL` function in the client, etc.
  */
-const wantsResizing = req => !!req.query.width;
+const wantsResizing = req =>
+    req.path.indexOf('/media') === 0 &&
```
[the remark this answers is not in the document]
davemacaulay: I've removed the path check, the TODO dictates this should one day be handled by UPWARD

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Column/column.js
```diff
@@ -116,9 +114,36 @@ const Column = props => {
...
+    // Determine the containers width and optimize the image
+    useEffect(() => {
+        if (image && columnElement.current) {
+            setBgImageStyle(
+                `url(${resourceUrl(image, {
+                    type: 'image-wysiwyg',
+                    width: columnElement.current.offsetWidth,
+                    height: columnElement.current.offsetHeight,
+                    quality: 85,
+                    crop: false,
```
danmooney2: Since width and height are present this will be converted to true

packages/venia-ui/lib/util/makeUrl.js
```diff
@@ -44,8 +44,14 @@ const mediaBases = new Map()
  * @param {string} props.type - "image-product" or "image-category"
  * @param {number} props.width - the desired resize width of the image
  * @param {number} props.height - the desired resize height of the image
+ * @param {number} props.quality - the desired quality of the image
+ * @param {bool} props.crop - should the image be cropped
+ * @param {string} props.fit - how should the image be fit with the dimensions: bounds, cover, crop
```
danmooney2: express-sharp only seems to assign values of `cover` or `outside` to `fit`, and assigns that value solely based on the value of `crop`.  https://github.com/magento-research/express-sharp/blob/develop/lib/transform.js#L23

davemacaulay: This is due to Fastly: https://docs.fastly.com/api/imageopto/fit

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Column/column.js
davemacaulay: I've fixed the logic within addImgOptMiddleware to handle this correctly

PR #1775, opened 2019-09-24
Checkout These Talons!

sirugh:
## Description

Talon-ized the checkout components. I did not add much documentation because IMO we should refactor checkout to use context. At that time the props being passed around will be greatly reduced in number making docs _not_ a pain in the rear to write.

Components not being "talonized". I didn't do these for reasons. Most of them are just display logic but I wasn't sure about `BraintreeDropin`.

```
  Form
  Label
  Section
  BraintreeDropin (crazy?) I think this is something very custom/confusing -- people will probably roll their own or just use ours wholesale w/ no customization
  *Summary (too simple?) These don't have much functionality beyond display
```
### Verification Stakeholders
@jimbo 

### Verification Steps
Regression test checkout. Authed and unauthed.

## Checklist


packages/peregrine/lib/talons/Checkout/useFlow.js
```diff
@@ -0,0 +1,98 @@
...
+        shippingTitle,
+        submitPaymentMethodAndBillingAddress,
+        submitShippingAddress,
+        submitShippingMethod,
+        handleBeginCheckout,
+        handleCancelCheckout,
+        handleCloseReceipt,
+        handleSubmitOrder,
+        step
+    };
```
jimbo: It looks to me like we're destructuring `checkoutState`, then including both `checkoutState` *and* its properties in the return value. I don't think that's necessary; `checkoutState` alone is probably enough, don't you think?

packages/peregrine/lib/talons/Checkout/useFlow.js
```diff
@@ -0,0 +1,98 @@
...
+        setStep('cart');
+    }, [setStep]);
+
+    return {
+        availableShippingMethods,
+        billingAddress,
+        cartState,
+        checkoutDisabled: isSubmitting || cartState.isEmpty,
+        checkoutState,
+        isReady: isCheckoutReady(checkoutState),
```
jimbo: Perfect. 👍

packages/peregrine/lib/talons/Checkout/useAddressForm.js
```diff
@@ -0,0 +1,37 @@
...
+    );
+
+    return {
+        countries,
+        error,
+        handleCancel: onCancel,
+        handleSubmit: onSubmit,
+        initialValues: values,
+        isSubmitting
+    };
```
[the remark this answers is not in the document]
sirugh: Right, good question. My gut says I should rewrite this as `isDisabled` like I did in the other form talons. This is one I missed.

Basically my process has been to pass everything back but in a "standard" way so that it's a little uniform among the talons. They all return an `isDisabled`, a `handleSubmit`, and a `handleCancel`. They probably should all take an `onSubmitError`, too. I didn't want to get too caught up in *new* API though, I would rather make this a copy-paste of most functionality and we can rework the API afterwards. Ideally we wouldn't release the unfinished API but deadlines...

packages/peregrine/lib/talons/Checkout/useFlow.js
sirugh: Good catch -- I think I can clean this up.

packages/peregrine/lib/talons/Checkout/useAddressForm.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Returns values used to render an AddressForm component.
+ * @param {Object} props
+ * @param {Object} props.fields an array of fields to reduce over for initial values
+ * @param {Object} props... TODO complete these
+ */
+export const useAddressForm = props => {
+    const {
+        onCancel,
+        countries,
```
sirugh: @jimbo `countries` would normally come from context so I wanted to make sure to export it from the talon even though nothing happens to the incoming prop here. Thoughts?

packages/peregrine/lib/talons/Checkout/useAddressForm.js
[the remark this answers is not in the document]
sirugh: I cleaned these up. Resolving this convo but we can talk more about whether the new approach is correct when you are online.

packages/peregrine/lib/talons/Checkout/usePaymentsFormItems.js
```diff
@@ -0,0 +1,88 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+// TODO install informed?
```
sirugh: @jimbo I want your thoughts on this talon.

jimbo: Yes, please add `informed` as a peer dep to `peregrine`. We're going to need its hooks.

sirugh: Done.

packages/venia-ui/lib/components/Checkout/addressForm.js
```diff
@@ -130,7 +125,7 @@ const AddressForm = props => {
                 </div>
             </div>
             <div className={classes.footer}>
-                <Button onClick={cancel}>Cancel</Button>
+                <Button onClick={onCancel}>Cancel</Button>
```
sirugh: @jimbo see how I'm passing the function directly here? Based on our other conversation I should change these to be wrapped/returned by the talon.

packages/peregrine/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "peerDependencies": {
     "@babel/runtime": "~7.4.2",
+    "informed": "~2.1.13",
```
sirugh: This seems to have updated the yarn lock. 🤷‍♂

PR #1779, opened 2019-09-25
[PB] Heading Component

eug123:
## Description
Introduce Heading component for Page Builder Heading content type

## Related Issue
https://jira.corp.magento.com/browse/PB-11

## Performance Results
https://wiki.corp.magento.com/display/COREENG/Heading

## Verification Steps

### DIFF
--- a/pwa-devdocs/scripts/create-reference-docs/config/venia-concept/index.js
+++ b/pwa-devdocs/scripts/create-reference-docs/config/venia-concept/index.js
@@ -1,5 +1,9 @@
 module.exports = [
   {
+    target: 'venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Heading/heading.js',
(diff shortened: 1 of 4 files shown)


pwa-devdocs/scripts/create-reference-docs/config/venia-concept/index.js
```diff
@@ -1,4 +1,8 @@
 module.exports = [
+  {
+    target: 'venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Heading/heading.js',
+    type: 'function'
+  },
```
davemacaulay: You need to also add `richContent.js` to this file

davemacaulay: Resolved.

PR #1786, opened 2019-09-25
Add SearchBar talons

jimbo:
## Description

Adds Peregrine talons for the components within Venia's `SearchBar` folder.

## Acceptance 

No net change.

### Verification Steps
1. Search for `dresses` using the search bar
2. Verify that the prompt or result message is correct in all states
3. Verify that there are no errors in the console
4. Verify that the results and links work

## Checklist


packages/venia-ui/lib/components/SearchBar/autocomplete.js
```diff
@@ -1,51 +1,33 @@
...
-
-    const [queryResult, queryApi] = useQuery(PRODUCT_SEARCH);
-    const { data, error, loading } = queryResult;
-    const { resetState, runQuery, setLoading } = queryApi;
-
-    const { value } = useFieldState('search_query');
-    const valid = value && value.length > 2;
+    const talonProps = useAutocomplete({ query: PRODUCT_SEARCH, visible });
+    const { messageType, resultCount, queryResult, value } = talonProps;
+    const { data } = queryResult;
```
sirugh: I was thinking about this -- do we want to return the `queryResult` or do we want to normalize within the talon?

sirugh: Since you already inspect `data` inside the talon you may as well return `products: data.products || {}` here.

packages/venia-ui/lib/components/SearchBar/suggestions.js
```diff
@@ -9,19 +10,16 @@ import defaultClasses from './suggestions.css';
...
 
-    const onNavigate = useCallback(() => {
-        setVisible(false);
-    }, [setVisible]);
+    const talonProps = useSuggestions({ filters, items, setVisible, visible });
+    const { categories, onNavigate, shouldRender } = talonProps;
 
-    if (!visible || !filters || !items || !items.length) {
+    // render null without data
+    if (!shouldRender) {
```
sirugh: 👍

packages/venia-ui/lib/components/SearchBar/autocomplete.js
```diff
@@ -1,51 +1,33 @@
...
-            runQuery({ variables: { inputText: value } });
-        } else if (!value) {
-            resetState();
-        }
-    }, [resetState, runQuery, setLoading, valid, value, visible]);
+    const messageTpl = MESSAGES.get(messageType);
+    const message =
+        typeof messageTpl === 'function'
+            ? messageTpl`${resultCount}`
+            : messageTpl;
```
sirugh: This is odd but only so much that I'm calling it out and not necessarily asking you to change anything :D

jimbo:
> I was thinking about this -- do we want to return the queryResult or do we want to normalize within the talon?

It's a tough call.

- One the one hand, it would be nice to return `queryResult` because we allow people to call the talon with a custom query, which is pointless if we complete obscure the result from them.
- On the other hand, talons are the right place to take what we need from the data and transform it as necessary, so you're right, we could normalize it here.

So I'm doing both now. The talon now returns `products` in addition to the raw `queryResult`.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -0,0 +1,71 @@
...
+    } else if (!resultCount) {
+        messageType = 'EMPTY_RESULT';
+    } else {
+        messageType = 'RESULT_SUMMARY';
+    }
+
+    // log any errors to the console
+    useEffect(() => {
+        if (error && process.env.NODE_ENV !== 'production') {
+            console.error(error);
```
sirugh: Maybe we should do this in `useQuery`? I think it'd be beneficial and we probably don't want to add this effect to every component that uses it.

packages/venia-ui/lib/components/SearchBar/suggestedCategory.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { func, shape, string } from 'prop-types';
+import { Link } from '@magento/venia-drivers';
+import { useSuggestedCategory } from '@magento/peregrine/lib/talons/SearchBar';
+
+import { mergeClasses } from '../../classify';
+import defaultClasses from './suggestedCategory.css';
```
sirugh: You added the file but there is no css. I think you need to, at a minimum, move over the `value` class but looks like you're also missing `root` and `label` (which are net new)

packages/peregrine/lib/talons/SearchBar/useSuggestedCategory.js
```diff
@@ -0,0 +1,40 @@
+import { useCallback } from 'react';
+
+// TODO: derive from store config when available
```
sirugh: What exactly are we supposed to derive from the config? We're currently querying the config when we first render.

Can you provide specifics in this TODO and/or link to an issue containing specifics :D

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }, [error]);
+
+    // run the query once on mount, and again whenever state changes
+    useEffect(() => {
+        if (visible && valid) {
+            setLoading(true);
+            runQuery({ variables: { inputText: value } });
```
sirugh: I think you should move this to an async function and await it so we can `setLoading(false)` after.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
jimbo: Yeah, we could do that.

packages/venia-ui/lib/components/SearchBar/suggestedCategory.js
jimbo: That's correct; as it happens, there are no styles associated with those classnames at the moment, so we don't need to define rulesets for them in the `.css` file. (No style changes were made to any elements, but yes, new classnames were added.)

Since `defaultClasses` happens to be `{}`, I could just define it as a `const` instead of importing it, but that would make the file less predictable for both tests and future changes.

packages/peregrine/lib/talons/SearchBar/useSuggestedCategory.js
jimbo: I don't know; this comment was already there and I preserved it.

packages/venia-ui/lib/components/SearchBar/suggestedCategory.js
sirugh: Oh, I didn't realize there weren't any old styles :D

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
jimbo: Explicitly calling `setLoading(false)` after `await runQuery()` is [redundant](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/hooks/useQueryResult.js#L27).

PR #1791, opened 2019-09-26
Gallery Refactor

sirugh:
I was going to create a talon for the `Gallery` component and I started doing a little refactoring. Turns out we can pretty much remove the `GalleryItems` component and we can simplify how Gallery works. In doing this I discovered that there's pretty much no talon to be written for `Gallery`.

 It is important to note that `Gallery` is used in `Category` _and_ in `Search`. In both cases there may be a different number of incoming or expected items being rendered which was counterintuitive when you looked at the hardcoded `pageSize` value.

So what I did was remove the `pageSize` prop and instead put the onus on the parent component. In this case, `Category` knows that it expects 6 items so it generates an empty placeholder array.


sirugh:
I disagree -- `GalleryItems` wasn't doing anything important and I'd rather simplify the codebase than keep old things around just because.

PR #1793, opened 2019-09-26
Refactor header components and add talons

sirugh:
This PR is largely a refactor of header with some moving of code to talons.

* OnlineIndicator has been moved under the `Header/` directory as it didn't make sense as a standalone component.
* I've added context to some components such as `cartTrigger` so that they can more easily be used independently of the `Header` at a later time.
* I removed `cartCounter` as it was a <span> with minimal styling.
* I inlined icons rather than passing them as children of some of the triggers.


packages/venia-ui/lib/components/Header/header.css
```diff
@@ -32,13 +32,6 @@
     padding: 0 1rem;
 }
 
-.navTrigger,
-.cartTrigger {
```
sirugh: Unused classes.

packages/venia-ui/lib/components/Header/header.js
```diff
@@ -54,12 +48,10 @@ const Header = props => {
...
                 <div className={classes.secondaryActions}>
                     <SearchTrigger
-                        searchOpen={searchOpen}
-                        toggleSearch={toggleSearch}
-                    >
-                        {searchIcon}
-                    </SearchTrigger>
-                    <CartTrigger {...cartTriggerProps} />
+                        active={searchOpen}
+                        onClick={handleSearchTriggerClick}
```
sirugh: Renamed props.

PR #1798, opened 2019-09-27
[cleanup] Delete unused home component

sirugh:
Unused component.


sirugh:
Technically a major change if anyone was utilizing this component (though I doubt they were)

PR #1803, opened 2019-09-27
Imagine image talons. 

sirugh:


packages/venia-ui/lib/components/Image/image.js
```diff
@@ -14,39 +15,47 @@ import defaultClasses from './image.css';
  * @param {string} props.fileSrc the raw source of the image without width and height added
  */
 const Image = props => {
-    const { alt, onError, onLoad, placeholder, src, fileSrc, ...rest } = props;
-    const classes = mergeClasses(defaultClasses, props.classes);
+    const {
+        alt,
+        classes: propsClasses,
```
sirugh: I pulled this off so that we weren't trying to spread `classes` into the `img` attributes.

jimbo: Good catch, that was a bug in waiting.

PR #1807, opened 2019-09-27
MiniCart Talons

sirugh:
Notable changes:
* moved a bunch of utilities to `peregrine`. Intentionally not refactoring them right now.


packages/venia-ui/lib/components/MiniCart/footer.js
```diff
@@ -37,7 +36,7 @@ const Footer = props => {
                 subtotal={subtotal}
             />
             <Suspense fallback={placeholderButton}>
-                <Checkout cart={cart} setStep={setStep} step={step} />
+                <Checkout setStep={setStep} step={step} />
```
jimbo: This is because `Checkout` aka `Flow` gets `cart` from context now, correct?

packages/peregrine/lib/talons/MiniCart/useProduct.js
```diff
@@ -0,0 +1,45 @@
...
+        beginEditItem(item);
+    }, [beginEditItem, item]);
+
+    const handleRemoveItem = useCallback(() => {
+        setIsLoading(true);
+        removeItemFromCart({ item });
+    }, [item, removeItemFromCart]);
+
+    const imageObj = {
+        url: image.file,
```
jimbo: Interesting—any reason to change from `file` to `url`? Not that I mind necessarily.

packages/peregrine/lib/talons/MiniCart/useKebab.js
```diff
@@ -0,0 +1,27 @@
+import { useCallback, useRef, useState } from 'react';
+import { useEventListener } from '@magento/peregrine';
+
+export const useKebab = () => {
+    const kebabRef = useRef(null);
```
jimbo: This is great. Coincidentally, it's also literally a full reimplementation of `useDropdown`, in which the only differences are the inclusion of a callback to expand (`handleKebabClick`) and attachment of an extra listener (`touchend`).

As I've mentioned before, creating talons is going to reveal real opportunities for code reuse. 😄 

We can probably keep `useKebab` for now, but I would add a to-do comment.

packages/peregrine/lib/talons/MiniCart/useEditItem.js
```diff
@@ -0,0 +1,43 @@
...
+            setLoading(true);
+
+            await runQuery({
+                variables: {
+                    name: item.name,
+                    onServer: false
+                }
+            });
+
+            setLoading(false);
```
jimbo: As mentioned in another PR, `setLoading(false)` is redundant here, since `runQuery` will call `receiveResponse`, prompting the reducer to set `loading` to `false`.

packages/peregrine/lib/talons/MiniCart/useProduct.js
sirugh: No, it's just the syntax we used in `useCategoryTile` so I'm trying to be consistent.

packages/venia-ui/lib/components/MiniCart/footer.js
sirugh: Yes.

packages/peregrine/lib/talons/MiniCart/useEditItem.js
sirugh: I've done this elsewhere in other talons but I'll remove it here. Hopefully we can start using apollo hooks soon anyways.

packages/peregrine/lib/talons/MiniCart/useKebab.js
```diff
@@ -0,0 +1,27 @@
+import { useCallback, useRef, useState } from 'react';
+import { useEventListener } from '@magento/peregrine';
+
```
sirugh:
```suggestion

// TODO: Compare with `useDropdown` and consolidate if possible.
```

packages/peregrine/lib/talons/MiniCart/useKebab.js
sirugh: Added a comment.

packages/peregrine/lib/talons/MiniCart/useEditItem.js
sirugh: I still think it's weird we have to call `setLoading(true)` :D

dpatil-magento:
QA Pass, @jimbo if recent commits looks good, merge it!

packages/peregrine/lib/talons/MiniCart/useEditItem.js
jimbo: > I still think it's weird we have to call `setLoading(true)` :D

Yeah, it is weird. We only have to do it because the Apollo client is bugged, such that `runQuery()` never returns a response with `loading: true`. When we switch to official hooks, we shouldn't have to call `setLoading()` at all.

PR #1810, opened 2019-09-30
Navigation talon

sirugh:
Disregard branch name, this is actually `Navigation/` component folder talons.


packages/peregrine/lib/talons/Navigation/useNavigationHeader.js
```diff
@@ -0,0 +1,31 @@
+import { useCallback } from 'react';
+
+const titles = {
+    CREATE_ACCOUNT: 'Create Account',
+    FORGOT_PASSWORD: 'Forgot Password',
+    MY_ACCOUNT: 'My Account',
+    SIGN_IN: 'Sign In',
+    MENU: 'Main Menu'
+};
```
jimbo: I think this mapping should happen in the view. See how `useAutocomplete` returns the constant while [`Autocomplete` maps it to the final string](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/SearchBar/autocomplete.js#L10-L15).

sirugh: OK I'll do this.

PR #680, opened 2018-12-19
Select variant media gallery entries after option selections are made

mtbottens:
Closes #844 .

## This PR is a:

- [ ] New feature
- [ ] Enhancement/Optimization
- [ ] Refactor
- [x] Bugfix
- [ ] Test for existing code
- [ ] Documentation

## Summary

When this pull request is merged, it will...

Swap out media gallery entries after option selections are performed by the user. #607 
c89a7e96-5717-441f-914d-89a7842e61fe

PR #1814, opened 2019-10-01
ProductFullDetail talon

sirugh:
NOTE: 

This PR refactors `useCarousel` to have a local state for `activeIndex` instead of using a reducer. This is because of a bug found in using the hook with a variable length input array of images.

This PR also changes the PDP to show the correct amount of images when the PDP first renders and to show the correct image after a color is selected.


packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.js
```diff
@@ -12,143 +11,42 @@ import Carousel from '../ProductImageCarousel';
...
-                        value={productPrice.value}
+                        currencyCode={productDetails.price.currency}
+                        value={productDetails.price.value}
                     />
                 </p>
             </section>
             <section className={classes.imageCarousel}>
-                <Carousel
-                    images={mediaGalleryEntries.value}
-                    key={mediaGalleryEntries.key}
```
sirugh: `key` didn't seem used so I deleted the code generating it and just returned the computed entries.

packages/venia-ui/lib/components/ProductOptions/options.js
```diff
@@ -24,6 +24,8 @@ class Options extends Component {
         const { handleSelectionChange, props } = this;
         const { product, selectedValues = [] } = props;
 
+        // TODO: Do this check in parent and only pass `configurable_options`
+        // instead of the entire `product` as a prop.
```
sirugh: Fun note, we already do this in the mini cart product options page but we don't do it in the PFD. :shrug:

sirugh:
I have to do some more work on this -- @dpatil-magento identified a discrepancy in that we no longer reset the selected image/thumbnail to the first when we select a color. So I have to do some work to figure out how to do that.

Additionally, when clicking an image for a new color we no longer re-render/fetch _every_ thumbnail, instead we just request the changed one (the first one, in this case).

So good and bad unintentional bugs :D

dpatil-magento:
@sirugh As discussed in mobile view now last image does not load, it skips to first image.

packages/peregrine/lib/List/useListState.js
```diff
@@ -218,7 +225,13 @@ const getInitiallySelectedKeys = ({
...
         }
 
-        return [getItemKey(initialSelection)];
+        const itemKey = getItemKey(initialSelection);
+
+        if (itemKey) {
+            return [itemKey];
+        }
+
+        return [];
```
sirugh: This didn't actually manifest in a bug that I could tell but it was initializing with an array of `[undefined]` which seems wrong. If there is no `itemKey` that should be handled here as a case rather than the former method.

packages/peregrine/lib/hooks/useCarousel.js
```diff
@@ -32,10 +32,11 @@ const wrappedReducer = withLogger(reducer);
  * @param {number} startIndex the index at which to start the carousel
  */
 export const useCarousel = (images = [], startIndex = 0) => {
-    const [{ index }, dispatch] = useReducer(wrappedReducer, {
+    const [{ index, length }, dispatch] = useReducer(wrappedReducer, {
         index: startIndex,
         length: images.length
     });
+    console.log(`useCarousel length`, length);
```
sirugh: Oops, will remove this.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -0,0 +1,167 @@
...
+        return INITIAL_OPTION_SELECTIONS;
+    }
+
+    const initialOptionSelections = new Map();
+    for (const { attribute_id } of product.configurable_options) {
+        initialOptionSelections.set(attribute_id, undefined);
+    }
+
+    return initialOptionSelections;
+};
```
sirugh: This is a new function.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -0,0 +1,167 @@
...
+const getMediaGalleryEntries = (product, optionCodes, optionSelections) => {
+    let value = [];
+
+    const { media_gallery_entries, variants } = product;
+    const isConfigurable = isProductConfigurable(product);
+
+    // Selections are initialized to "code => undefined". Once we select a value, like color, the selections change. This filters out unselected options.
+    const optionsSelected =
+        Array.from(optionSelections.values()).filter(value => !!value).length >
+        0;
```
sirugh: This is new code.

dpatil-magento:
QA Pass, good to merge once review is approved.

packages/peregrine/lib/hooks/useCarousel.js
```diff
@@ -1,62 +1,40 @@
...
+    const sortedImages = useMemo(() => sortImages(images), [images]);
 
-    const handleNext = useCallback(() => dispatch({ type: 'next' }), []);
+    const handlePrevious = useCallback(() => {
+        if (activeItemIndex > 0) {
+            setActiveItemIndex(activeItemIndex - 1);
+        } else {
+            setActiveItemIndex(images.length - 1);
+        }
+    }, [activeItemIndex, images]);
```
jimbo: When using `useCallback`, look out for scenarios like this one. As written, `handlePrevious` will be recreated whenever `activeItemIndex` changes, and since `handlePrevious` itself changes `activeItemIndex`, this memoization is pointless.

Before, we were keeping `index` in a reducer so that `handlePrevious` could be memoized. If you're going to remove the reducer in favor of `useState`, you'll want to use the [functional update](https://reactjs.org/docs/hooks-reference.html#functional-updates) form.

```es6
const handlePrevious = useCallback(() => {
  setActiveItemIndex(prev => prev > 0 ? prev - 1 : images.length - 1)
}, [images])
```

packages/peregrine/lib/hooks/useCarousel.js
```diff
@@ -1,62 +1,40 @@
...
 
-    const setActiveItemIndex = useCallback(
-        index => dispatch({ type: 'set', value: index }),
-        []
-    );
+    const handleNext = useCallback(() => {
+        setActiveItemIndex((activeItemIndex + 1) % images.length);
+    }, [activeItemIndex, images]);
```
jimbo: This memoization also fails, for the same reason.

```es6
const handleNext = useCallback(() => {
  setActiveItemIndex(prev => (prev + 1) % images.length)
}, [images])
```

packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.js
jimbo: Just from [looking quickly](https://github.com/magento/pwa-studio/pull/680), I think it's safe to say there was a valid reason to calculate that key before. Did something change so that it's no longer necessary?

packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.js
```diff
@@ -12,143 +11,42 @@ import Carousel from '../ProductImageCarousel';
...
     return (
         <Form className={classes.root}>
             <section className={classes.title}>
-                <h1 className={classes.productName}>{product.name}</h1>
+                <h1 className={classes.productName}>{productDetails.name}</h1>
                 <p className={classes.productPrice}>
                     <Price
-                        currencyCode={productPrice.currency}
-                        value={productPrice.value}
+                        currencyCode={productDetails.price.currency}
```
jimbo: Is there any way this could result in an error being thrown?

sirugh: https://github.com/magento/pwa-studio/pull/1814/files#diff-dd082926e212e0be673b35eb01025b3aL101

`productPrice` was previously just `amount` which is an object containing `currency` and `value` props. The normalization done in the talon is equivalent with only a rename of `productPrice` to `productDetails.price`.

Technically the old way (and the new method) can both throw if any of the values are undefined in the chain. Since we don't yet have option chaining I can leave these to some default.

dpatil-magento:
@sirugh  If user is on 2nd or 3rd image and click on Add to Cart,  automatically first image gets loaded.  On `develop` branch it stays on selected images only.

dpatil-magento:
@sirugh Simple Product page does not load anymore https://pr-1814.pwa-venia-staging.com/carmina-earrings.html   :(

dpatil-magento:
QA Pass, Need reviewer approval to get merged.

PR #1816, opened 2019-10-01
Refactors VeniaAdapter to improve initial page load

### LINKED ISSUE #1741: Venia Adapter Optimization
# Overview

On initial application load, the [VeniaAdapter](https://github.com/magento/pwa-studio/blob/release/4.0/packages/venia-ui/lib/drivers/adapter.js) component takes order(s) of magnitude more time than other components:

# The Ask

Investigate the `VeniaAdapter` component and determine if any optimizations can be made. The delay is likely in creating the `ApolloClient`. See if we can't do any of that work ahead of time.

supernova-at:
## Description

This PR refactors `VeniaAdapter` to improve initial page load. It:

  * Updates `VeniaAdapter` to a functional component
  * Pre-instantiates an Apollo cache object (`InMemoryCache`) at `require` time so that it doesn't have to when the component is instantiated

## Related Issue

Closes #1741 .

### Verification Steps

1. Ensure the Venia home page loads correctly
1. Issue some GraphQL requests (navigate around the app) and ensure everything still works
1. Open Chrome Dev Tools > Application > Local Storage and clear `apollo-cache-persist` if it exists. Reload the page and ensure that `apollo-cache-persist` is recreated. This tests that we didn't break cache persistence.

**Before this PR**

**After this PR**

Screen Shot 2019-10-01 at 10 00 50 AM

See that the "Self Time" of `VeniaAdapter` has decreased.
This measures how long a component takes to render itself (not including its child components).

## Checklist


packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -8,6 +8,24 @@ import { InMemoryCache } from 'apollo-cache-inmemory';
...
+/**
+ * We intentionally do not wait for the async function persistCache to complete
+ * because that would negatively affect the initial page load.
+ *
+ * The tradeoff is that any queries that run before the cache is persisted may not be persisted.
+ */
+persistCache({
+    cache: preInstantiatedCache,
+    storage: window.localStorage
+});
```
jimbo: Very informative comments. I agree with the direction of the changes here; I think *not awaiting* `persistCache()` will work toward your goal of not blocking initial page load.

Have you considered moving the cache instantiation and this `persistCache()` call into a `useEffect()`, though? If the goal is to free up the main thread to render as quickly as possible, and if we're already okay with allowing that render to occur before the Apollo cache gets prepared, why not defer this work until [after first paint](https://reactjs.org/docs/hooks-reference.html#useeffect) by doing it as an effect?

supernova-at: I'll try that - thanks!

supernova-at: Actually, I don't think I can do this.

>  if we're already okay with allowing that render to occur before the Apollo cache gets prepared

We're okay with the render occurring before the cache is _persisted_ (via `persistCache`) but we can't render before the cache is _created_ - `cache` is a required argument to the `ApolloClient` constructor and the `ApolloContext.Provider` and `ApolloProvider` below need an Apollo client instance on initial render.

PR #1817, opened 2019-10-01
ProductImagecarousel and Thumbnail talons

sirugh:


packages/peregrine/lib/talons/ProductImageCarousel/useThumbnail.js
```diff
@@ -0,0 +1,13 @@
+import { useCallback } from 'react';
+
+export const useThumbnail = props => {
+    const { itemIndex, onClickHandler } = props;
```
jimbo: I think this kind of talon is going to prove to be a good candidate for a utility hook in the future. 👍

PR #1819, opened 2019-10-01
Product Options Talons

sirugh:
Notes:

* Big refactor of `ProductOptions`, `List` is no longer used. I ran into a problem where we were hitting the max update depth after only converting `Options` to a function component. Rather than debug list, a known buggy component, I just switched it out for a map. Some modifications were necessary in the handlers. I rushed this conversion to get it functional so I may need to revisit/change things. Feedback welcome.
* Major version label as this changes API around props.


packages/peregrine/lib/talons/MiniCart/useCartOptions.js
```diff
@@ -59,12 +59,11 @@ export const useCartOptions = props => {
...
-            const lastSelection = Array.from(selection).pop();
-
-            newSelections.set(optionId, lastSelection);
-
-            setOptionSelections(newSelections);
+            // We must create a new Map here so that React knows that the value
+            // of optionSelections has changed.
+            const newOptionSelections = new Map([...optionSelections]);
+            newOptionSelections.set(optionId, selection);
+            setOptionSelections(newOptionSelections);
```
sirugh: This is exactly how `ProductFullDetail`'s selection change callback works. Mirrored that. only change was the spread of the array into the new map.

packages/venia-ui/lib/components/ProductOptions/swatch.js
```diff
@@ -38,6 +38,11 @@ class Swatch extends Component {
         return isSelected ? <Icon src={CheckIcon} /> : null;
     }
 
+    handleClick = () => {
+        const props = this.props;
+        props.onClick(props.item.value_index);
```
sirugh: Should I return the `item` or the `value`?

packages/venia-ui/lib/components/ProductOptions/swatchList.js
```diff
@@ -1,26 +1,38 @@
...
 import Swatch from './swatch';
 
 import { mergeClasses } from '../../classify';
 import defaultClasses from './swatchList.css';
 
 const SwatchList = props => {
-    const { getItemKey, initialSelection, items, onSelectionChange } = props;
+    const {
+        getItemKey,
+        initialSelection = {},
```
sirugh: I want to change this prop name to `selectedValue`. The time at which the value is selected doesn't matter.

packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.js
```diff
@@ -127,7 +127,7 @@ const ProductFullDetail = props => {
             // We must create a new Map here so that React knows that the value
             // of optionSelections has changed.
             const newOptionSelections = new Map([...optionSelections]);
-            newOptionSelections.set(optionId, Array.from(selection).pop());
+            newOptionSelections.set(optionId, selection);
```
sirugh: This feels better. I don't know why we were popping the return value but it felt too coupled to whatever was invoking the callback.

packages/venia-ui/lib/components/ProductOptions/option.js
```diff
@@ -38,14 +38,14 @@ const Option = props => {
...
-        let selection = {};
-        if (selectedValue) {
-            selection =
-                values.find(value => value.default_label === selectedValue) ||
-                {};
+        let initialSelection = {};
+        const searchValue = selection || selectedValue;
+        if (searchValue) {
+            initialSelection =
+                values.find(value => value.default_label === searchValue) || {};
```
sirugh: This is the coupling that basically enforces we use `label` as the indicator of selection. Not a big deal, but wanted to point this out.

jimbo: But only for product options, right? I don't see anything wrong with it; is there something you would like to use instead?

sirugh: I just think the normalization should occur elsewhere.

fooman:
While this is being refactored might also be worthwhile considering renaming it to ConfigurableProductOptionsTalon.
(this naming would match the GraphQl naming)

This comment `// Non-configurable products don't have options.` is not quite accurate as options can be added to other product types as well - see below screenshots:

image

image

image

packages/peregrine/lib/talons/ProductOptions/useOption.js
```diff
@@ -0,0 +1,57 @@
...
+        );
+    }, [values]);
+
+    const selectedValueLabel = `Selected ${label}:`;
+    const selectedValueDescription =
+        selection || initialSelection.default_label || 'None';
+
+    const handleSelectionClick = useCallback(
+        selection => {
+            setSelection(valuesMap.get(selection));
```
sirugh: This no longer pulls the value off the array, it just assumes whomever invoked the callback (in this case the lowest `handleClick` has properly formed the value, unlike the old method: `const [selectedValue] = Array.from(selection);`

packages/peregrine/lib/talons/ProductOptions/useSwatch.js
```diff
@@ -0,0 +1,13 @@
+import { useCallback } from 'react';
+
+export const useSwatch = props => {
+    const { onClick, value_index } = props;
+
+    const handleClick = useCallback(() => {
+        onClick(value_index);
```
sirugh: Previously this was being handled by a click handler applied by the `List` component which has been refactored out of `Swatch` and `Tile`.

sirugh:
@fooman good points. I decided to put the onus of deciding whether to render the options to the parent component. So in this case, `ProductFullDetail` and `CartOptions`. Those components can inspect the context/data for whatever they need and make that decision. The `Options` component should be a dumb component that does some mapping of type-to-value in order to know some rendering specifics like `Swatch` vs `Tile` but beyond that it doesn't need to know anything about the `product`.

packages/venia-ui/lib/components/MiniCart/cartOptions.js
```diff
@@ -56,8 +56,8 @@ const CartOptions = props => {
         <Suspense fallback={loadingIndicator}>
             <section className={classes.options}>
                 <Options
-                    onSelectionChange={handleSelectionChange}
-                    product={configItem}
+                    onSelectionClick={handleSelectionClick}
+                    options={configItem.configurable_options}
```
sirugh: `Options` now takes an array `options` instead of the product.

sirugh:
On hold pending #1814

PR #1828, opened 2019-10-02
adds app talon

sirugh:
Updates `App` to use the `useApp` talon.


zetlen:
Looks great! I am pretty excited for the App component to get littler.

A hook with an API this large might need to be documented more in comments. I know we don't use JSDoc frequently, but I'm wondering if we ought to start, to aid in editor autocompletion.

sirugh:
> A hook with an API this large might need to be documented more in comments.

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
```diff
@@ -29,6 +29,7 @@ export const useMiniCart = () => {
         !((cartState.isEmpty && step === 'cart') || isLoading || isEditingItem);
 
     const isMiniCartMaskOpen = step === 'form';
+    const isOpen = drawer === 'cart';
```
sirugh: This was removed as a prop being passed by `App` since we already reference `useAppContext` here anyways. Might as well just fetch the state we need in `useMiniCart`.

packages/venia-ui/lib/components/App/container.js
```diff
@@ -1,23 +1,14 @@
...
 
-    return (
-        <ErrorBoundary
-            unhandledErrors={unhandledErrors}
-            app={appState}
-            {...appApi}
-            {...errorApi}
-        />
-    );
+    return <ErrorBoundary unhandledErrors={unhandledErrors} {...errorApi} />;
```
sirugh: State and api are accessible within the hook. No longer necessary within the container.

packages/venia-ui/lib/components/App/app.js
```diff
@@ -83,64 +68,45 @@ const App = props => {
...
     if (renderError) {
         return (
-            <Fragment>
+            <HeadProvider>
+                <Title>{`Home Page - ${STORE_NAME}`}</Title>
                 <Main isMasked={true} />
                 <Mask isActive={true} />
                 <ToastContainer />
-            </Fragment>
+            </HeadProvider>
```
sirugh: I think we forgot to add this when we added `HeadProvider` and `Title` so I did so here.

packages/venia-ui/lib/components/App/app.js
```diff
@@ -83,64 +68,45 @@ const App = props => {
...
+    );
+
+    const talonProps = useApp({
+        handleError,
+        handleIsOffline,
+        handleIsOnline,
+        markErrorHandled,
+        renderError,
+        unhandledErrors
+    });
```
sirugh: Most of these, aside from the online/offline handlers could probably be moved to `useErrorBoundary` or something other than `app`. That said, I chose not to do that in this PR to keep the changeset minimal.

PR #1830, opened 2019-10-03
feat: Simplify image middleware with express-sharp update

### LINKED ISSUE #1808: [bug]: When express-sharp is down, the media files don't get served.
When express-sharp is down, the media files don't get served.
```
Emitting no ServiceWorker in development mode. To enable development mode for ServiceWorkers, pass `enableServiceWorkerDebugging: true` to the ServiceWorkerPlugin configuration.
Cannot add image optimization middleware due to dependencies that are not installed or are not compatible with this environment:
- @magento/express-sharp: Reason: Found vendored libvips v8.8.1 but require v8.7.4. Please remove the 'node_modules/sharp/vendor' directory and run 'npm install'.
- @magento/express-sharp: Reason: expressSharp is not a function
```
@zetlen found a possible solution.
https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/upward.yml#L15
needs to be updated to: `pattern: '^/(graphql|rest|media)(/|$)'`

zetlen:
## Description

The express-sharp library has updated to v3.0.0, which includes a [breaking change](https://github.com/magento-research/express-sharp/pull/5) to the URL scheme, and therefore the public API of the library.

This helps us simplify the code in addImgOptMiddleware, because we don't have to rewrite the URL as aggressively--it's closer to the Fastly paradigm.

## Related Issue

Closes #1808 and #1842 

### Verification Steps

1. Visit #1808 and follow those verification steps.
2. Check out this branch and reproduce the error conditions.
3. Should work!
4. Bug #1842 should also be resolved.

## Checklist


sirugh:
Test failing, but otherwise the code looks like it should work. I don't know exactly how to test this either. Can you fix the test and fill out some verification steps?

packages/pwa-buildpack/lib/Utilities/addImgOptMiddleware.js
```diff
@@ -59,30 +63,18 @@ If possible, install additional tools to build NodeJS native dependencies:
...
+            // Rewrite them just a little bit.
+            if (query.format === 'pjpg') {
+                query.progressive = true;
+                query.format = 'jpeg';
             }
-            if (incomingQuery.auto === 'webp') {
-                params.set('format', 'webp');
+            if (query.auto === 'webp') {
+                query.format = 'webp';
             }
```
Jordaneisenburger: I think we should support .png as well. We know that .png doesn't perform well on the web but in production shops there is a very likely possibility that merchants will use this format (just like we do in our current project). Best we can do is set .png to progressive just like we do for .jpeg

zetlen: I haven't tested that! I believe this supports PNG; all you have to do is use URL parameters `format=png&progressive=true`. By default, it'll optimize PNGs into WebP files, which I think is a good default, but that query string should override it.

Jordaneisenburger: That's correct, but ios & safari don't support WebP. So till they do i believe we should make a .png case

sirugh: Yea, I think we need to fix this on Safari.

packages/pwa-buildpack/lib/Utilities/addImgOptMiddleware.js
```diff
@@ -91,35 +85,28 @@ https://github.com/nodejs/node-gyp#installation`
...
-            const url = rewritten.href.slice(rewritten.origin.length);
-
-            // make new query object for express-sharp to use
-            const query = {};
-            for (const [key, value] of params) {
-                query[key] = value;
-            }
-            debug({ url, query });
-            return { url, query };
+            debug('rewrote query to %o', query);
```
sirugh: I noticed we are now leaving `auto: 'webp'` on the `query` object. Is that intended
```
2019-10-07T20:37:31.473Z pwa-buildpack:Utilities:addImgOptMiddleware.js rewrote query to { auto: 'webp', format: 'jpeg', width: '640', height: '800', progressive: true, crop: true }
```

dpatil-magento:
Images looks good in iphone and safari now.

dpatil-magento:
Verified both linked bugs, looks good.

PR #1835, opened 2019-10-03
Remove unused imports and ignore some parameters

DanielRuf:
## Description

This PR removes unused imports and variables and marks some parameters as ignored.

## Related Issue

Closes #<issue_number>

## Checklist

* [ ] I have updated the documentation accordingly, if necessary.
* [ ] I have added tests to cover my changes, if necessary.

### DIFF
--- a/pwa-devdocs/scripts/create-env-variables-docs/handleVariable.js
+++ b/pwa-devdocs/scripts/create-env-variables-docs/handleVariable.js
@@ -1,10 +1,8 @@
 const {
-  root,
   heading,
   text,
   paragraph,
   strong,
-  brk,
   inlineCode,
 } = require('mdast-builder');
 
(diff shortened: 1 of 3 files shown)


pwa-devdocs/scripts/create-env-variables-docs/handleVariable.js
```diff
@@ -17,7 +15,7 @@ const {
  * @returns {array} An array of mdast nodes containing the environment variable data
  */
 const handleVariable = variable => {
-  const {name, type, desc, example, default: defaultValue} = variable;
+  const {type, desc, example, default: defaultValue} = variable;
```
supernova-at: I'd prefer this the other way (keep the destructure and just use `name` on L21) but that's super minor.

PR #1855, opened 2019-10-06
feat(webpack): strip all comments for PROD build

### LINKED ISSUE #1074: Strip JS Comments from Venia Production bundle
Because Venia will soon be both a reference storefront and an "as is / quick deploy" option for clients, we should outfit it with security best practices.

Comments left in JavaScript source files and delivered to end users can clue attackers in to potential weaknesses of the system. This is especially true for `// TODO` comments or similar.

Since these comments are still helpful to developers we don't want to lose them at the source code level. One potential solution that satisfies both conditions is to strip them from the final production bundle.

Venia uses a plugin called [terser](https://github.com/magento-research/pwa-studio/blob/develop/packages/venia-concept/webpack.config.js#L15) to [configure its output webpack bundle](https://github.com/magento-research/pwa-studio/blob/develop/packages/venia-concept/webpack.config.js#L224-L249).

According to [this comment](https://github.com/webpack-contrib/terser-webpack-plugin/issues/27#issuecomment-439964816) in `terser-webpack-plugin`, we should be able to set `terserOption.output.comments` to `false` when building for production to strip the comments from the bundle.

Note: Our product is still Open Source and attackers could just go view the source code. This is just a small step towards our goal.

RanadeepPolavarapu:
Resolves #1074.

## Description

Remove comments in webpack prod builds.

## Related Issue

Closes #1074 .

### Verification Steps

1. `yarn build`
2. Verify commentless output `dist` content.

## Checklist

### DIFF
--- a/packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
+++ b/packages/pwa-buildpack/lib/WebpackTools/configureWebpack.js
@@ -340,7 +340,8 @@ async function configureWebpack({ context, vendor = [], special = {}, env }) {
                         },
                         output: {
                             ecma: 7,
-                            semicolons: false
+                            semicolons: false,
+                            comments: false
                         },
                         keep_fnames: true
                     }


sirugh:
Sorry, this is already done by default as we discovered in the original issue. Thanks for your contribution but we will close this.

Issue #374, opened 2018-10-16
Implement basic offline experience for Venia (PWA)


ericerway:
## This issue is for the following packages:

[ ] `venia-concept`
[ ] `pwa-buildpack`
[X] `peregrine`
[ ] `pwa-module`
[ ] `pwa-devdocs`

## Description

As a shopper, I want basic feedback on the Venia storefront when my device is having connectivity issues or offline so that I know that I can have an uninterrupted shopping experience.

This is a requirement for PWA per Google Lighthouse and applies to all of the Venia storefront.

#### Assumptions
- Dependent on examples of the Venia storefront
- Dependent on 200 response type to respond
- Browser messaging for this scenario is not sufficient (must use service worker)
- Utilizes Workbox for basic behaviors/needs (service worker debugging in DevDocs; ref.)
 
## Tasks
- [ ] Discovery for offline response types (200) by Venia storefront and UPWARD
- [ ] Implement offline response (fallback); verify in Lighthouse
- [ ] Testing and documentation (https://github.com/magento-research/pwa-studio/wiki/Documentation-Guidelines)

ericerway:
Marking as in progress from @pcvonz and team -- preview here: http://upward.bargreen.io

pcvonz:
Some notes on the obstacles I've encountered so far. Let me know if anything sounds incorrect: 

**Problem:** Workbox will only cache `GET` requests. On the first load this works since Upward returns an html file. However, every subsequent load doesn't return an html file.

**Solution:** Cache requests in the `MagentoRouteHandler`. 

Recipe for how to access the cache in the [Workbox docs](https://developers.google.com/web/tools/workbox/guides/common-recipes#access_caches_from_your_web_apps_code).

<hr />

**Problem:** When components are retrieved through the `fetchRootComponent` method of `MagentoRouteHandler` and Venia has no network connection, this results in a failed network request. 

**Solution:** (slap dash, probably not what we want)

Bypass `MagentoRouteHandler` when offline and request the actual html file.

<hr />

[Related issue in the Workbox repo](https://github.com/GoogleChrome/workbox/issues/1620)

rowan-m:
Great progress so far! Adding a few comments here on things I've seen. They may not all be relevant to this issue, but I wanted to make sure it's down.

1. As per the Lighthouse audit, the manifest is missing the 192px and 512px icons.
2. As per the Lighthouse audit, the manifest is missing a "`theme_color`"
[...]

PR #1867, opened 2019-10-08
getNavigationMenu Query Efficiency

### LINKED ISSUE #1677: GetNavigationMenu Query Inefficiency
In our `getNavigationMenu` query, [there is a comment](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/queries/getNavigationMenu.graphql#L11-L12) that explains an inefficiency in the query:

>      # Temporarily include products field in query to get around GraphQL bug
>      # affecting Magento Cloud Pro instances. Remove once magento\graphql-ce\#374 is fixed.

Track down `graphql-ce/374` and see if we can resolve this now.

supernova-at:
## Description

In our `getNavigationMenu` query, [there is a comment](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/queries/getNavigationMenu.graphql#L11-L12) that explains an inefficiency in the query:

>      # Temporarily include products field in query to get around GraphQL bug
>      # affecting Magento Cloud Pro instances. Remove once magento\graphql-ce\#374 is fixed.

**tl;dr:**

There is a new `products` field but it just issues a `products` query and it doesn't have any way to limit the size of the response (the default is 20 items).

We can improve the efficiency of this default behavior if we keep our `products` query with the `pageSize` set to `1`. 

**More Details:**

I tracked down https://github.com/magento/graphql-ce/issues/374 and found it was closed and a fix supposedly went in with https://github.com/magento/graphql-ce/issues/567.

There is now a `products` field in `category` that gives us access to the products of a category without having to issue a `products` query as mentioned in the comment.

The new query would be:

```graphql
query navigationMenu($id: Int!) {
  category(id: $id) {
    id
    name
    children {
      children_count
      id
      include_in_menu
      name
      position
      productImagePreview: products {
        items {
          small_image{
            url
          }
        }
      }
      url_path
    }
  }
}
```

But unfortunately there is no way to limit the size of the `products.items` array that is returned. So if we update our query, we will get the `small_image.url` of the first 20 items (the default limit) in the category. 

All we ever use is [the first one](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/CategoryList/useCategoryTile.js#L19), so we'd actually be drastically overfetching data if we made this change.

Since it appears the `CategoryInterface` includes a `products` sub-query anyway, we're also not extremely deviating from the functionality already in place:

```graphql
interface CategoryInterface {
  # other fields
  products(
    pageSize: Int = 20
    currentPage: Int = 1
    sort: ProductSortInput
  ): CategoryProducts
}
```

So long story short: There is a new `products` field but it just issues a `products` query and it doesn't have any way to limit the size of the response (the default is 20 items).

We can improve the efficiency of this default behavior if we keep our `products` query with the `pageSize` set to `1`. 

## Related Issue

Closes #1677 .

### Verification Steps

Check my reasoning on why we can safely remove this comment and no further changes are needed.
There are no steps to verify anything, as the resolution of this ticket is simply removing a `TODO` comment from the codebase.

## Checklist


sirugh:
Wow. I mean you killed it in terms of discovery for removal of a comment :o I applaud you for your diligence.

PR #1871, opened 2019-10-09
Lazy Load Images

### LINKED ISSUE #1689: [feature]: Enable native image lazy-loading in browsers
Like all other assets in a PWA, images should not load until they're needed. In a traditional webpage (and an older browser), images load when the page loads--even if they're way down the screen! The "lazy loading" technique developed to correct this.

In a lazy load, the code assumes an image is only needed when it appears in the viewport, or when it's about to. JavaScript libraries use scroll height measurement and the IntersectionObserver API to observe this and begin loading images when necessary. We could add one of these libraries to our codebase and to Venia.

Or, we could use the [native lazy loading features](https://www.machmetrics.com/speed-blog/browser-native-lazy-loading-coming-in-2019/) that browsers are starting to support. Since browsers are doing the work to support this natively, we should let them do it and not add a polyfill or additional library.

Add the `loading="lazy"` attribute to our images, as described in [this article](https://addyosmani.com/blog/lazy-loading/) and in other places. **That's it.**

Many lazy loading libraries exist, and using them as polyfills would support the browsers that don't do it natively yet.

However, they would add complexity and payload size; they would have an overall negative impact on customer experience.

Some of these libraries support a custom experience for lazy loading. We can replicate this using a combination of native lazy loading and the transparent placeholder images we already use.

**Additional context**
Between this and #1584, this would fix #668.

- [x] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

supernova-at:
## Description

* Updates all direct usages of `` to `<Image>`
* Adds support for `loading="lazy"` attribute on `` elements created by our `<Image>` component
* Adds support for `srcset` and `sizes` on `<Image>`s

Well that got out of hand fast 😬 . The original scope of the ticket was to support `loading="lazy"`, but since I was already consolidating logic inside the `Image` component this PR also includes `srcset` and `sizes` support app-wide 😅 🎉 .

## tl;dr

Desktop savings: **160 KB** (not counting category page)
Mobile savings: **226 KB**

Not counting the category page where we were previously requesting images that were too small, this PR saves us **160 KB** in image data on the primary checkout flow (homepage > category page > product detail page > minicart) on desktop and **226 KB** on mobile.

The biggest savings on the desktop are on the homepage and the PDP, and the biggest savings on mobile are on the homepage and the category page.

## Related Issue

Closes #1689.

### Verification Stakeholders

Front-end team

#### loading="lazy"

1. Load up the homepage, category page or PDP page
1. Inspect element on an image
1. See that the image has its `loading` attribute set to "lazy"

#### No Direct Uses of `img`

All `img` elements should now be rendered via the `Image` component. The `Image` component consolidates logic for lazy loading and responsive sizing in one place.

1. Search the codebase for " 640px.
1. Verify that the larger images are requested when the layout switches from two to three columns:

category_page

See above.

## Checklist

## In Progress

- [x] The `Carousel` thumbnails start very small before loading to their proper size
- [x] The `Carousel` main image issues multiple network requests for the same image
- [x] The `MiniCart/Product` image is requesting too large of an image (`1280`)
- [x] Unit Tests
- [ ] Documentation

## Note!

This PR does not address the fact that once you select a color on the PDP, that item image gets added to the thumbnails section. Out of scope for this PR.


sirugh:
I noticed that we have some weird image load/cancels happening on the product page. Do you think it'd be in scope to investigate this issue? It happens on `develop` so if you think we should just open an issue that'd be fine:

sirugh:
I also noticed a weird thing now where instead of hiding the image until displayed we are rendering both the placeholder and something else:

[Image from Gyazo](https://gyazo.com/3e8168acec63f49df023f5e3e75adf02)

sirugh:
I pushed a few commits to fix the category page placeholder and I updated some css but I wasn't able to fix the thumbnail issue on the product page. I _think_ it's something to do with using a `button` vs `a` but I'm not sure and I gotta run for the day. Will try to work with you tomorrow on this  if I can!

packages/venia-ui/lib/components/Gallery/__tests__/__snapshots__/gallery.spec.js.snap
```diff
@@ -6,17 +6,14 @@ exports[`renders if \`items\` is an array of objects 1`] = `
...
     <div>
-      <img
-        alt=""
-        height="375"
-        src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAQAAAAFCAQAAADIpIVQAAAADklEQVR42mNkgAJGIhgAALQABsHyMOcAAAAASUVORK5CYII="
-        width="300"
-      />
       <img
         alt="Test Product 1"
+        className="undefined undefined"
```
supernova-at: TODO

packages/venia-ui/lib/components/Gallery/__tests__/__snapshots__/item.spec.js.snap
```diff
@@ -0,0 +1,67 @@
...
+
+exports[`renders a placeholder item while awaiting item 1`] = `
+<div
+  className="l"
+>
+  <div
+    className="f"
+  >
+    <img
+      alt=""
```
supernova-at: Is there a good `alt` value we can use for a placeholder?

packages/venia-ui/lib/components/Gallery/__tests__/gallery.spec.js
```diff
@@ -2,17 +2,30 @@ import React from 'react';
...
+});
+
 const classes = { root: 'foo' };
 const items = [
     {
         id: 1,
         name: 'Test Product 1',
         small_image: {
-            url: '/test/product/1.png'
+            url: 'https://hackernoon.com/hn-images/1*ELZJStuoVnqb-Y3duIrH0Q.png'
```
supernova-at: Revert these changes

packages/venia-ui/lib/components/Gallery/item.css
```diff
@@ -38,18 +34,13 @@
...
     composes: image;
-    opacity: 0;
-    visibility: hidden;
+    background-color: rgb(var(--venia-grey));
 }
 
-.imagePlaceholder_pending {
-    composes: imagePlaceholder;
+.images_pending {
+    composes: images;
```
supernova-at: I did not make this change....

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,21 +1,29 @@
...
 const imageWidth = '300';
 const imageHeight = '375';
 
-const ItemPlaceholder = ({ children, classes }) => (
+const ItemPlaceholder = ({ classes }) => (
     <div className={classes.root_pending}>
-        <div className={classes.images_pending}>{children}</div>
+        <div className={classes.images_pending}>
+            <Image
+                alt=""
```
supernova-at: Missing `alt` value

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -24,114 +32,71 @@ const ItemPlaceholder = ({ children, classes }) => (
...
+        <div className={classes.root}>
+            <Link to={resourceUrl(productLink)} className={classes.images}>
+                <Image
+                    alt={name}
+                    classes={{ root: classes.image }}
+                    fileSrc={small_image}
+                    height={imageHeight}
+                    placeholder={transparentPlaceholder}
+                    sizes={`${imageWidth}px`}
+                    width={imageWidth}
```
supernova-at: Can we drop `width` and `height` in favor of `sizes`?

packages/venia-ui/lib/components/Gallery/item.css
sirugh: T'was me!

packages/venia-ui/lib/components/Logo/logo.js
```diff
@@ -1,18 +1,20 @@
...
     return (
-        <img
-            className={classes.logo}
-            src={logo}
-            height={height}
+        <Image
             alt="Venia"
+            classes={{ root: classes.logo }}
+            height={height}
+            loading="eager"
```
supernova-at: This was more for proof of concept than anything - we don't need `eager` here.

packages/venia-ui/lib/components/Gallery/item.css
supernova-at: I just read your comments 😆

packages/venia-ui/lib/components/Gallery/__tests__/__snapshots__/item.spec.js.snap
supernova-at: It is a temporary placeholder but the `alt` of the image is the `item.name`, and this only gets rendered when `item` doesn't exist (yet).

I went with "Placeholder for gallery item image" 👍 .

supernova-at: https://github.com/magento/pwa-studio/pull/1871/commits/70dc5fd9cc7db80d41403353df8037299713898b

packages/venia-ui/lib/components/Gallery/item.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1871/commits/70dc5fd9cc7db80d41403353df8037299713898b.

packages/venia-ui/lib/components/Logo/logo.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1871/commits/a12d03275ea133bea7521ffd077cefb6ca68cbca.

packages/venia-ui/lib/components/Gallery/__tests__/__snapshots__/gallery.spec.js.snap
supernova-at: This occurs because `mergeClasses` and `defaultClasses` are not available when running unit tests.

Stopping to debug on this line:
```js
const classes = mergeClasses(defaultClasses, propsClasses);
```

shows

Screen Shot 2019-10-10 at 1 36 36 PM

Therefore on a later line,

```js
const imageClass =
        classes.root + ' ' + (isLoaded ? classes.loaded : classes.notLoaded);
```

we're left with `undefined undefined`.

packages/venia-ui/lib/components/MiniCart/product.js
```diff
@@ -39,27 +37,27 @@ const Product = props => {
...
-                fileSrc={image.url}
-                sizes={`${image.width}px`}
-            />
-        );
-    }, [hasImage, image, productName, classes.image]);
+        // Sizes
+        const imageProps = {
+            alt: productName,
+            classes: { root: classes.image },
+            sizes: `80px`
```
sirugh: Move this const to the top. Can merge with the other width setting ie `sizes: ${IMAGE_WIDTH}px`

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,21 +1,27 @@
...
 
-const ItemPlaceholder = ({ children, classes }) => (
+const ItemPlaceholder = ({ classes }) => (
     <div className={classes.root_pending}>
-        <div className={classes.images_pending}>{children}</div>
+        <div className={classes.images_pending}>
+            <Image
+                alt="Placeholder for gallery item image"
+                classes={{ root: classes.image_pending }}
+                src={transparentPlaceholder}
```
sirugh: Something is still wrong with placeholder images in gallery: 

[Image from Gyazo](https://gyazo.com/f286c69debc68ce5dfed2c2bdab8c05f)

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -1,18 +1,26 @@
...
- * @param {string} props.src the source of the image
- * @param {string} props.fileSrc the raw source of the image without width and height added
+ *
+ * @param {string}   props.alt the alt text for the image
+ * @param {string}   props.classes any classes to apply to this component
+ * @param {number}   props.height the desired height of the image
+ * @param {function} props.onError callback for error loading image
+ * @param {function} props.onLoad callback for when image loads successfully
+ * @param {string}   props.placeholder the placeholder source to display while the image loads or if it errors on load
+ * @param {string}   props.resource the Magento path to the image ex: /v/d/vd12-rn_main_2.jpg
```
sirugh: If you keep the props, add `resourceWidth` and `resourceHeight` to these docs.

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -75,18 +113,48 @@ const Image = props => {
...
+    resourceHeight: oneOfType([number, string]),
+    resourceWidth: oneOfType([number, string]),
+    sizes: string,
+    src: conditionallyRequiredString,
+    type: string,
+    width: oneOfType([number, string])
+};
+
+Image.defaultProps = {
+    placeholder: transparentPlaceholder,
```
sirugh: The problem is that this placeholder is very specific to venia sample data. IMO the components should provide this or it should just not have one.

sirugh:
As a follow-on, why can't we use `max-width` within the `sizes` attribute instead of/in addition to CSS?

packages/peregrine/lib/talons/Image/useImage.js
```diff
@@ -33,6 +33,6 @@ export const useImage = props => {
         handleImageLoad,
         hasError,
         isLoaded,
-        shouldRenderPlaceholder: !!placeholder && !isLoaded
+        shouldRenderPlaceholder: !isLoaded || hasError
```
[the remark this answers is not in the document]
supernova-at: There's a defaultProp in `Image` that provides it if one isn't passed

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -75,18 +113,48 @@ const Image = props => {
     );
 };
 
+const conditionallyRequiredString = (props, propName, componentName) => {
+    // This component needs one of src or resource to be provided.
+    if (!props.src && !props.resource) {
```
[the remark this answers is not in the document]
supernova-at: Yeah, there are really two different "use cases" which I'm working on the documentation for:

1. I have a `src`

Pass it to `<Image>` as `src` and it will be used directly.
`srcset` and `sizes` do not come into play here.

2. I have a Magento "resource" - some string returned from GraphQL like "vt-13-main.jpg".

Pass that to `<Image>` as `resource`.
Optionally pass a `resourceWidth` and `resourceHeight` too, that will affect the fallback `src` used by older browsers that don't support `srcset` / `sizes`.
It will create a `srcset` for you.
You should pass in `sizes` depending on your usage.
The browser will pick the best image in `srcset` based on `sizes`, the viewport, and `window.devicePixelRatio`.

So here I'm trying to say "you need to pass at least one of `src` or `resource`".

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -21,49 +29,79 @@ const Image = props => {
...
+            loading="lazy"
             {...rest}
             alt={alt}
             className={imageClass}
             onError={handleError}
             onLoad={handleImageLoad}
-            src={src}
+            src={source}
             srcSet={imageSrcset}
+            sizes={sizes}
```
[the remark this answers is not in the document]
supernova-at: Good feedback, I did the same for `width` and `height` too. I'm not really a fan of the spread props for things like this though because then the API isn't clear.

I suppose I could add them to `propTypes` but not destructure them.

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -75,18 +113,48 @@ const Image = props => {
...
     }),
+    height: oneOfType([number, string]),
     onError: func,
     onLoad: func,
     placeholder: string,
-    src: string,
-    fileSrc: string
+    resource: conditionallyRequiredString,
+    resourceHeight: oneOfType([number, string]),
+    resourceWidth: oneOfType([number, string]),
```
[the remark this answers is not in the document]
supernova-at: Sadly not, I had to separate them.

`resourceWidth` and `resourceHeight` are used for the fallback `src` for browsers that don't support `srcset` / `sizes`.

```jsx
<Image resource="resource.jpg" resourceWidth="80" resourceHeight="100" />
```

will result in something like ``.

But that `img` can be styled differently. For example, it may have `width: 45px`. In English, "request the width 80 image, and put it into this 45px space".

Often the CSS widths are dependent on the viewport / parent elements, so these two values don't always match up.

packages/venia-ui/lib/components/Gallery/item.js
supernova-at: 👍 , I'll take a look.

packages/venia-ui/lib/components/Image/image.js
supernova-at: I'm glad you brought this up, I thought about that too.

1. Every component I ran into that was passing in `placeholder` was passing in the same one
2. Some of them weren't passing in any placeholders (we lost this functionality. Do we want it back?)

I do see the downside that non-Venia consumers of our `Image` component may want a different size placeholder though.

supernova-at:
> As a follow-on, why can't we use max-width within the sizes attribute instead of/in addition to CSS?

We do that in `Gallery/item.js`:

```js
    const galleryBreakpoint = '640px';
    const sizeWhenTwoColumns = '50vw';
    const sizeWhenThreeColumns = '33vw';
    const sizes = `(max-width: ${galleryBreakpoint}) ${sizeWhenTwoColumns},
                   ${sizeWhenThreeColumns}`;
```

packages/peregrine/lib/talons/Image/useImage.js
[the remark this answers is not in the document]
sirugh: Yea I realized that later and commented too -- I don't think we should include or bundle `transparentPlaceholder` with `Image`. We don't know what form factors will want to use the `Image` component. In our case we happen to only have a single placeholder but it's specifically a "product image" placeholder.

packages/venia-ui/lib/components/Image/image.js
sirugh: Yea! As I said in the other comment (playing ping pong now!) we are just coincidentally only using `Image` for product images but it's supposed to be more generic.

packages/venia-ui/lib/components/MiniCart/product.js
```diff
@@ -39,27 +37,27 @@ const Product = props => {
...
-                alt={productName}
-                classes={{ root: classes.image }}
-                placeholder={transparentPlaceholder}
-                src={src}
-                fileSrc={image.url}
-                sizes={`${image.width}px`}
-            />
-        );
-    }, [hasImage, image, productName, classes.image]);
+        // Sizes
```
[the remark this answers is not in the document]
supernova-at: https://github.com/magento/pwa-studio/pull/1871/commits/3302c9b6f29dd756c1d1bb9c41a92c9d8c1c2073

packages/venia-ui/lib/components/MiniCart/product.js
```diff
@@ -39,27 +37,27 @@ const Product = props => {
...
+            sizes: `80px`
+        };
+
+        if (!hasImage) {
+            imageProps.src = transparentPlaceholder;
+        } else {
+            imageProps.resource = image.file;
+
+            // Should be kept in sync with product.css .image max-width.
+            imageProps.resourceWidth = 80;
```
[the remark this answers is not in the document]
supernova-at: https://github.com/magento/pwa-studio/pull/1871/commits/3302c9b6f29dd756c1d1bb9c41a92c9d8c1c2073

packages/venia-ui/lib/components/MiniCart/product.js
supernova-at: https://github.com/magento/pwa-studio/pull/1871/commits/3302c9b6f29dd756c1d1bb9c41a92c9d8c1c2073

packages/venia-ui/lib/components/ProductImageCarousel/carousel.js
```diff
@@ -72,6 +63,30 @@ const ProductImageCarousel = props => {
...
+                resourceWidth={DEFAULT_IMAGE_WIDTH}
+                sizes={sizes}
+            />
+        );
+    } else {
+        image = (
+            <Image
+                alt={altText}
+                classes={{ root: classes.currentImage }}
+                src={transparentPlaceholder}
```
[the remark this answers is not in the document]
supernova-at: Without a `srcset`, `sizes` won't do anything. There's only one size of the `transparentPlaceholder` asset.

packages/venia-ui/lib/components/CategoryList/categoryTile.js
```diff
@@ -12,28 +13,25 @@ const CategoryTile = props => {
...
-    const imageWrapperStyle = { '--venia-image': imageUrl };
-
     const classes = mergeClasses(defaultClasses, props.classes);
 
-    // render an actual image element for accessibility
-    const imagePreview = imagePath ? (
-        <img className={classes.image} src={imagePath} alt={item.name} />
-    ) : null;
+    // Preview sizes is derived from classes.image.
+    const previewSizes = '5rem';
```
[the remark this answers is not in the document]
supernova-at: https://github.com/magento/pwa-studio/pull/1871/commits/d79d8b84b04e2bec9fe4a226b9edfd2b5b22e393

packages/venia-ui/lib/components/MiniCart/product.js
```diff
@@ -40,20 +43,17 @@ const Product = props => {
...
+            classes: { root: classes.image }
         };
 
         if (!hasImage) {
             imageProps.src = transparentPlaceholder;
         } else {
             imageProps.resource = image.file;
-
-            // Should be kept in sync with product.css .image max-width.
             imageProps.resourceWidth = 80;
```
sirugh: I think this is still related to the const :D You could make this:

```
const PRODUCT_IMAGE_SIZE = 80;
...
imageProps.resourceWidth = PRODUCT_IMAGE_SIZE;
imageProps.sizes = ${PRODUCT_IMAGE_SIZE}px;
```

packages/peregrine/lib/talons/Image/useImage.js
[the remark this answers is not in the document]
supernova-at: Okay, by default now the `Image` component is going to render a placeholder while the image hasn't loaded yet or it errored. Callers can remove this functionality by setting a `usePlaceholder` prop to `false`.

https://github.com/magento/pwa-studio/pull/1871/commits/50761a370eaf396a6eec1e489e3766b2f430e096

packages/venia-ui/lib/components/Image/image.js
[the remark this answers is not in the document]
supernova-at: Ok I went to do this but only the actual image needs `sizes` (not the placeholder), so I kept the destructure so that I could spread `rest` out onto both the actual image and the placeholder.

packages/venia-ui/lib/components/Gallery/item.js
supernova-at: I added the concept of "layout only" placeholder styles to `Image` for this. See `classes.placeholder_layoutOnly`.

These styles _hold the place_ in the page while the image loads. If `usePlaceholder` is true (the default), it will also draw a gray box there.

https://github.com/magento/pwa-studio/pull/1871/commits/50761a370eaf396a6eec1e489e3766b2f430e096

packages/venia-ui/lib/components/MiniCart/product.js
```diff
@@ -39,27 +37,27 @@ const Product = props => {
...
+        if (!hasImage) {
+            imageProps.src = transparentPlaceholder;
+        } else {
+            imageProps.resource = image.file;
+
+            // Should be kept in sync with product.css .image max-width.
+            imageProps.resourceWidth = 80;
+        }
+
+        return <Image {...imageProps} />;
```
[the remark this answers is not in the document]
supernova-at: Fixed by https://github.com/magento/pwa-studio/pull/1871/commits/50761a370eaf396a6eec1e489e3766b2f430e096.

packages/venia-ui/lib/components/ProductImageCarousel/carousel.js
```diff
@@ -72,6 +63,30 @@ const ProductImageCarousel = props => {
 
     const classes = mergeClasses(defaultClasses, props.classes);
 
+    let image;
+    if (currentImage.file) {
+        // See .root grid-template-columns. 640px is 80% of 800px.
+        const sizes = '640px';
```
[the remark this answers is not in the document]
supernova-at: Done https://github.com/magento/pwa-studio/pull/1871/commits/50761a370eaf396a6eec1e489e3766b2f430e096.

packages/venia-ui/lib/components/Image/image.js
supernova-at: Yes, thank you: https://github.com/magento/pwa-studio/pull/1871/commits/50761a370eaf396a6eec1e489e3766b2f430e096.

packages/venia-ui/lib/components/Image/image.js
supernova-at: Okay, `Image` should be able to handle anything now. We still bundle & use the 4x5 transparent image but we can size it however we want now.

https://github.com/magento/pwa-studio/pull/1871/commits/50761a370eaf396a6eec1e489e3766b2f430e096

packages/venia-ui/lib/components/MiniCart/product.js
supernova-at: Moved these to `const`s. They just happen to have the same value so I didn't define them in terms of each other, but I left a comment in the code explaining it.

https://github.com/magento/pwa-studio/pull/1871/commits/0e5fd9af4eb39b3d92cb8d33643059176245d5ff.

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -21,72 +38,142 @@ const Image = props => {
...
+            className={placeholderImageClass}
+            loading="eager"
+            src={placeholder}
+            {...rest}
+        />
     );
 
-    const actualImage = !hasError && (
+    /*
+     * These don't live in the talon because they depend on @magento/venia-drivers.
```
sirugh: Can you clarify this comment? What do you mean by "these"?

packages/venia-ui/lib/components/Gallery/__tests__/__snapshots__/gallery.spec.js.snap
```diff
@@ -6,21 +6,33 @@ exports[`renders if \`items\` is an array of objects 1`] = `
...
-        alt="Test Product 1"
-        height="375"
-        loading="lazy"
-        sizes="300px"
-        src="a.url"
-        srcSet="a.url 40w,a.url 80w,a.url 160w,a.url 320w,a.url 640w,a.url 1280w,a.url 1600w,a.url 2560w"
-        width="300"
-      />
+      <div
+        className="undefined undefined"
```
sirugh: I think you can correct this in snapshots by mocking classify ie `jest.mock('../../classify');`

packages/venia-ui/lib/components/ProductImageCarousel/carousel.css
```diff
@@ -7,17 +7,23 @@
...
 .currentImage {
-    background-color: rgb(var(--venia-background-color));
-    border-radius: 2px;
     width: 100%;
-    grid-area: 1 / 1 / 2 / 4;
+}
+
+.currentImage_placeholder {
+    composes: currentImage;
+    background-color: rgb(var(--venia-grey));
```
sirugh: Can we just use the background color on `currentImage` and do away with this class entirely?

packages/venia-ui/lib/components/CategoryList/categoryTile.js
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
 import { arrayOf, string, shape } from 'prop-types';
 import { mergeClasses } from '../../classify';
-import { Link, resourceUrl } from '@magento/venia-drivers';
+import { Link } from '@magento/venia-drivers';
 import defaultClasses from './categoryTile.css';
 import { useCategoryTile } from '@magento/peregrine/lib/talons/CategoryList/useCategoryTile';
+import Image from '../Image';
+
+// This value mirrors classes.image.
```
jimbo: 😅 

The `CategoryTile` component has no idea what style properties are associated with `.image`. If there's a reasonable belief that `.image` will want to use this value to derive a style, it should provide this value as a custom property. We don't manually synchronize values between a component and its default stylesheet.

```jsx
const IMAGE_SIZES = ["5rem"]

const Image = props => {
  // sizes should be an array
  const { sizes } = props
  const sizesAttribute = sizes.join(",")

  // final size must be a width, not a media condition
  const finalSize = sizes[sizes.length - 1]
  const style = { "--size": finalSize }

  return (
    <img sizes={sizesAttribute} style={style} />
  )
}
```

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,138 +1,113 @@
...
 
-const ItemPlaceholder = ({ children, classes }) => (
-    <div className={classes.root_pending}>
-        <div className={classes.images_pending}>{children}</div>
-        <div className={classes.name_pending} />
-        <div className={classes.price_pending} />
-    </div>
-);
+// See gallery.css for breakpoints and number of columns.
+const galleryBreakpoint = '640px';
```
revanth0212:
```suggestion
const GALLERY_BREAK_POINT = '640px';
```

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,138 +1,113 @@
...
+const sizeWhenTwoColumns = '300px';
```
revanth0212:
```suggestion
const SIZE_FOR_2_COLUMNS = '300px';
```

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,138 +1,113 @@
...
+const sizeWhenThreeColumns = '840px';
```
revanth0212:
```suggestion
const SIZE_FOR_3_COLUMNS = '840px';
```

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,138 +1,113 @@
...
+const IMAGE_SIZES = `(max-width: ${galleryBreakpoint}) ${sizeWhenTwoColumns}, ${sizeWhenThreeColumns}`;
```
revanth0212:
```suggestion
const IMAGE_SIZES = `(max-width: ${GALLERY_BREAK_POINT}) ${SIZE_FOR_2_COLUMNS}, ${SIZE_FOR_3_COLUMNS}`;
```

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,138 +1,113 @@
...
 import { transparentPlaceholder } from '@magento/peregrine/lib/util/images';
-import { generateSrcset } from '../../util/images';
+
+import { mergeClasses } from '../../classify';
+import Image from '../Image';
 import defaultClasses from './item.css';
 
 // The placeholder image is 4:5, so we should make sure to size our product
 // images appropriately.
 const imageWidth = '300';
```
revanth0212:
```suggestion
const IMAGE_WIDTH = '300';
```

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,138 +1,113 @@
...
 const imageHeight = '375';
```
revanth0212:
```suggestion
const IMAGE_HEIGHT = '375';
```

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,138 +1,113 @@
...
+        <div className={classes.root}>
+            <Link to={productLink} className={classes.images}>
+                <Image
+                    alt={name}
+                    classes={{
+                        image: classes.image,
+                        root: classes.imageContainer
+                    }}
+                    resource={small_image}
+                    resourceHeight={imageHeight}
```
revanth0212:
```suggestion
                    resourceHeight={IMAGE_HEIGHT}
```

packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -1,138 +1,113 @@
...
+                    resourceWidth={imageWidth}
```
revanth0212:
```suggestion
                    resourceWidth={IMAGE_WIDTH}
```

packages/venia-ui/lib/components/Gallery/item.js
revanth0212: Also, are these values always constant? I am trying to see if they can be moved to a separate file of their own, something like `constants/image.js` and fetched from there so it will be easy to maintain and use for potential vendors.

revanth0212:
Also, can you check what is this warning about?

image

packages/venia-ui/lib/components/Image/image.js
supernova-at: Clarified in https://github.com/magento/pwa-studio/pull/1871/commits/8e815492172cac77557241c4e7dc51a15a634220.

packages/venia-ui/lib/components/Gallery/__tests__/__snapshots__/gallery.spec.js.snap
supernova-at: Thank you! Resolved in https://github.com/magento/pwa-studio/pull/1871/commits/8e815492172cac77557241c4e7dc51a15a634220.

packages/venia-ui/lib/components/ProductImageCarousel/carousel.css
supernova-at: I tried this and got the dreaded gray bar 😨 

Screen Shot 2019-10-25 at 9 32 51 AM

But I have no idea why, honestly. The `img` element boundary definitely does not include that area:

Screen Shot 2019-10-25 at 9 40 56 AM

And it is coming from this property - toggling it on and off in dev tools made it appear / disappear 🤷‍♂

packages/venia-ui/lib/components/CategoryList/categoryTile.js
supernova-at: > We can't declare sizes in CSS
`sizes` is an attribute on `img`. Therefore I had to "mirror" some CSS values in JS. If these values ever changed, they'd need to be updated in both places. While I added comments where appropriate to try and fend this off, it feels gross 🤮.

Thank you for addressing this, lol. I can change these to custom properties.

packages/venia-ui/lib/components/ProductImageCarousel/carousel.css
sirugh: Well in the past that was happening because the placeholder image, which is rendered just behind the actual product image, was *slightly* bigger. I had fixed this by cropping/stretching the product images to cover the placeholder but it seems like that got lost somewhere.

I think with some investigation we could figure it out but probably not really worth it in this PR anymore.

packages/venia-ui/lib/components/CategoryList/categoryTile.js
supernova-at: I think this will work for single value `sizes` like this, but I'm struggling with the multiple sizes use case, specifically this line:

```js
// final size must be a width, not a media condition
const finalSize = sizes[sizes.length - 1]
```

We don't want to use that width if we have a media condition that is satisfied. 

Consider `IMAGE_SIZES = ["(max-width: 640px) 2rem", "5rem"];` and CSS that looks like:

```css
.image {
  width: 5rem;
}

@media (max-width: 640px) {
    .image {
        width: 2rem;
    }
}
```

If both of these values were re-written to use a `--size` custom property, one of them would be wrong:

```css
.image {
  width: var(--size); /* 5rem */
}

@media (max-width: 640px) {
    .image {
        width: var(--size); /* 5rem, but should be 2rem! */
    }
}
```

So now I'm thinking we need a custom property for each media condition. And as long as we're there, we also need one for the breakpoint(s) (`640px`) to keep that in sync.

So our CSS becomes:

```css
.image {
  width: var(--size-large); /* 5rem */
}

@media (max-width: var(--size-breakpoint)) {
    .image {
        width: var(--size-small); /* 2rem */
    }
}
```

Which already feels like way too much presentational logic in the JS, but that aside: who names these things?

It makes sense that a `component.js` would want to define the names the custom properties used by `component.css`, but how does it pass the `Image` component those names?

The `Image` component itself could also do it but then the custom property names couldn't have much meaning and would have to be things like `--size0`, `--size1`, `--size-breakpoint0` or similar.

Given `IMAGE_SIZES = ["(max-width: 640px) 2rem", "5rem"];`, `Image` could generate and set values for:
`--size0`: '5rem'
`--size-breakpoint1`: '640px'
`--size1`: '2rem'

But then `component.css` would look like:

```css
.image {
  width: var(--size0); /* 5rem */
}

@media (max-width: var(--size-breakpoint1)) {
    .image {
        width: var(--size1); /* 2rem */
    }
}
```

which is clunky.

I'd love some advice / feedback here.

jimbo: > Which already feels like way too much presentational logic in the JS, but that aside: who names these things?
> 
> It makes sense that a `component.js` would want to define the names the custom properties used by `component.css`, but how does it pass the `Image` component those names?
> 
> The `Image` component itself could also do it but then the custom property names couldn't have much meaning and would have to be things like `--size0`, `--size1`, `--size-breakpoint0` or similar.

Good questions.

**Who names things?**

`Image` would name them, not the component that rendered `Image`. Ordinarily, `Image` wouldn't know much about its elements' presentation, including how that presentation would change in response to media queries. The CSS would declare presentation, and `Image` would blindly apply that CSS.

In this case, size information is actually **data** , since it determines which image resource is requested. Since `Image` can presume that its presentation will depend on this data, in the same way an accordion's presentation depends on an `expanded` state, we need to relay this data to the stylesheet. The mechanics of how that happens are the responsibility of `Image`.

`Image` decides the custom property names in the same way it decides what classname keys it will apply to its elements. If my custom component, `Foo`, happens to render an `<Image />` element, it may supply overrides for the classname keys `Image` has defined, such as `classes.root`, but even if it were to supply new custom keys such as `classes.bar`, `Image` would not (*could not*) apply them to one of its internal elements.

Custom properties work the same way: if `Image` has logic to apply certain properties via `style`, with values derived from `props.sizes`, then `Foo` may adjust those property values via `props.sizes`. In other words, that's `Image`'s interface.

**Isn't this clunky?**

Less clunky than hoping things stay manually synced. 😅 

It's a bit verbose because you can't supply a collection of values via a single custom property—and even if you could, you couldn't have a rule *select* a single value from that collection. So we do have to supply multiple properties.

brendanfalkowski:
@supernova-at If you know how to make this work, let me have it, but I don't think CSS custom properties can be used in a media query like this. You have to hardcode it, or bring on a pre-processor's variables.

```
@media (max-width: var(--size-breakpoint)) {
    // Nope
}
```

A thousand curses on the CSS spec writers for deliberately making that not work, and being pedantic about properties vs variables.

supernova-at:
>  I don't think CSS custom properties can be used in a media query like this.

@brendanfalkowski yarrr, thanks - I was afraid of that :/

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -1,92 +1,162 @@
...
+    string
+} from 'prop-types';
 import { useImage } from '@magento/peregrine/lib/talons/Image/useImage';
 
-import { generateSrcset } from '../../util/images';
 import { mergeClasses } from '../../classify';
 import defaultClasses from './image.css';
+import PlaceholderImage from './placeholderImage';
+import ResourceImage from './resourceImage';
+import SimpleImage from './simpleImage';
```
jimbo: I know we still do this in a couple places (tiles come to mind), but when we're conditionally rendering child components, it's not ideal to synchronously import all of them up front. Not only are we loading more code than we need sooner than we need it, we're losing the ability to tree-shake these components.

Consider letting components choose the image variant they need without going through `Image`.

packages/venia-ui/lib/components/Image/resourceImage.js
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @param {string}   props.className the class to apply to this image.
+ * @param {string}   props.resource the Magento path to the image ex: /v/d/vd12-rn_main_2.jpg
+ * @param {number}   props.resourceHeight the height to request for the fallback image for browsers that don't support srcset / sizes.
+ * @param {object}   props.resourceSizeBreakpoints breakpoints related to resourceSizes.
+ * @param {object}   props.resourceSizes image sizes used by the browser to select the image source.
+ * @param {number}   props.resourceWidth the width to request for the fallback image for browsers that don't support srcset / sizes.
+ * @param {string}   props.type the Magento image type ("image-category" / "image-product"). Used to build the resource URL.
+ */
+const ResourceImage = props => {
```
jimbo: Great doc comments throughout these components. 💯

packages/venia-ui/lib/components/Image/simpleImage.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { string, func } from 'prop-types';
+
+const SimpleImage = props => {
+    const { className, handleError, handleLoad, src, ...rest } = props;
+
+    return (
+        // eslint-disable-next-line jsx-a11y/alt-text
```
jimbo: This lint rule exists for a good reason. Let's make sure `alt` is a required prop. 😅

packages/venia-ui/lib/components/ProductImageCarousel/thumbnail.js
```diff
@@ -44,24 +47,26 @@ const Thumbnail = props => {
...
-                  width: DEFAULT_THUMBNAIL_WIDTH,
-                  height: DEFAULT_THUMBNAIL_HEIGHT
-              })
-            : transparentPlaceholder;
+        if (!isDesktop) {
+            return null;
+        }
 
-        return isDesktop ? (
+        return Boolean(file) ? (
```
jimbo:
```suggestion
        return file ? (
```

Pretty sure the ternary makes this redundant.

packages/venia-ui/lib/components/CategoryList/categoryTile.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1871/commits/ab27aa394627cf9292a4b133f4359a39f5a587fc.

packages/venia-ui/lib/components/CategoryList/categoryTile.js
```diff
@@ -12,28 +16,23 @@ const CategoryTile = props => {
...
-    const imageUrl = imagePath ? `url(${imagePath})` : 'none';
-    const imageWrapperStyle = { '--venia-image': imageUrl };
-
     const classes = mergeClasses(defaultClasses, props.classes);
 
-    // render an actual image element for accessibility
-    const imagePreview = imagePath ? (
-        <img className={classes.image} src={imagePath} alt={item.name} />
-    ) : null;
+    const imagePreview = image.url ? (
```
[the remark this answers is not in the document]
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1871/commits/5615a0a718b9d81d5cd4ca87f9e811ee18e0b11b.

packages/venia-ui/lib/components/Gallery/item.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1871/commits/97b24bc8d601f52d93d26756bdb9154d02ba6170.

packages/venia-ui/lib/components/Gallery/item.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1871/commits/97b24bc8d601f52d93d26756bdb9154d02ba6170.

packages/venia-ui/lib/components/Gallery/item.js
supernova-at: Good idea. They've since been altered slightly but these are coupled to this specific component so I don't think a separate file would help much.

supernova-at:
> Also, can you check what is this warning about?

```
[Intervention] An <img> element was lazyloaded with loading=lazy, but had no dimensions specified. Specifying dimensions improves performance.
```

It appears this message is not relevant per https://bugs.chromium.org/p/chromium/issues/detail?id=1002455 (we use fluid layouts).

Since this comes from the browser itself, I'm not sure there's much we can do here.

packages/peregrine/lib/talons/Image/useImage.js
```diff
@@ -28,11 +34,29 @@ export const useImage = props => {
...
+            const value = resourceSizes[size];
+
+            result[styleKey] = value;
+        }
+
+        return result;
+    }, [resourceSizes]);
+
     return {
+        customCSSProperties,
```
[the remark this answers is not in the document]
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1871/commits/0de2fec7cd19b07c021719911e9ab477aff76505.

packages/peregrine/lib/talons/Image/useImage.js
```diff
@@ -28,11 +34,29 @@ export const useImage = props => {
...
 
+    // Example: { '--image-size-small': '5rem' }
+    const customCSSProperties = useMemo(() => {
+        const result = {};
+
+        if (!resourceSizes) {
+            return result;
+        }
+
+        for (const size in resourceSizes) {
```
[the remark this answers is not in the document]
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1871/commits/bbf8a1ba47dcd0cae04c1beee5c2824ac88bd8c0.

packages/venia-ui/lib/components/Image/image.js
supernova-at: Good points. I think I may create a separate issue for this, as this would be a non-trivial refactor at this point. The `Image` component takes care of all the placeholder handling, for example.

packages/venia-ui/lib/components/Image/simpleImage.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1871/commits/1c31db4905410ce8f09baa55066c636086f2cd85.

packages/venia-ui/lib/components/Image/___tests__/image.spec.js
```diff
@@ -72,22 +73,22 @@ test('supports overriding the loading attribute', () => {
...
-        // Act.
-        const instance = createTestInstance(<Image {...resourceProps} />).root;
-
-        // Assert.
-        const container = instance.children[0];
-        const image = container.children[1];
-        expect(image.props.srcSet).toBeTruthy();
-    });
-});
+// describe('resource handling', () => {
```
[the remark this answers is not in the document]
supernova-at: :D I really just have to move this test to the right place. Working on updating tests now!

packages/venia-ui/lib/components/ProductImageCarousel/thumbnail.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1871/commits/051efa8201892481f3d044336efcd92705b38c75

packages/venia-ui/lib/components/CategoryList/categoryTile.js
```diff
@@ -23,7 +24,7 @@ const CategoryTile = props => {
             alt={item.name}
             classes={{ image: classes.image, root: classes.imageContainer }}
             resource={image.url}
-            sizes={IMAGE_SIZES}
+            resourceSizes={IMAGE_SIZES}
```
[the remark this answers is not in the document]
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1871/commits/34771c1b98e425ee56b31ba60d2460f5d41fc7e0

packages/venia-ui/lib/components/Image/image.js
supernova-at: Ok I created [JIRA 85](https://jira.corp.magento.com/browse/PWA-85) for this.

packages/venia-ui/lib/components/Gallery/__tests__/__snapshots__/item.spec.js.snap
```diff
@@ -0,0 +1,93 @@
...
+      />
+      <img
+        alt="Test Product"
+        className="image loaded"
+        height={375}
+        loading="lazy"
+        onError={[MockFunction]}
+        onLoad={[MockFunction]}
+        sizes="(max-width: 640px) 300px, 840px"
+        src="/media/catalog/product/foo/bar/pic.png?auto=webp&format=pjpg&height=375"
```
supernova-at: Shouldn't this have a `width` parameter on it?

packages/venia-ui/lib/components/Image/___tests__/image.spec.js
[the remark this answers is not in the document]
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1871/commits/70b42fcb3d8db17091dcdaea81850a96d830f56c and https://github.com/magento/pwa-studio/pull/1871/commits/33b04472ef5cbebbea1dea311e46b3aa3d1a2148.

supernova-at:
The following comments detail all of Venia's uses of the `Image` component and compare the currently deployed version of `develop` (https://develop.pwa-venia-staging.com/) to this PR.

The versions are compared on "Desktop" (1040px wide) and "Mobile" (iPhone 6/7/8).

supernova-at:
### The Homepage

https://develop.pwa-venia-staging.com/

Screen Shot 2019-10-14 at 1 38 16 PM

#### Desktop

Current:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-14 at 1 46 59 PM | 159.3 KB |

This PR:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-14 at 1 54 04 PM | 24.8 KB |

#### Mobile

Current:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-15 at 9 49 46 AM | 160.5 KB |

This PR:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-15 at 9 50 00 AM | 85.2 KB |

These changes can be found in `categoryTile.js`.

supernova-at:
### Category Page

https://develop.pwa-venia-staging.com/venia-bottoms.html

Screen Shot 2019-10-14 at 1 59 01 PM

#### Desktop

Current:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-14 at 2 00 03 PM | 66.3 KB |

This PR:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-14 at 2 01 38 PM| 221.3 KB<sup>1</sup> |

#### Mobile

Current:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-15 at 1 25 55 PM | 220 KB |

This PR:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-15 at 1 26 05 PM | 69 KB |

These changes can be found in `gallery/Item.js`.

<sup>1</sup>: While this one is an increase in size, we were previously requesting images that were too small, which is especially evident on wide screen widths (ex: 2180px):

**Current**:
Screen Shot 2019-10-14 at 2 09 05 PM

**This PR**:
Screen Shot 2019-10-14 at 2 09 12 PM

Additionally, we previously were not requesting alternate image sizes as the page size adjusted and now we are.

supernova-at:
### Produce Detail Page (PDP)

https://develop.pwa-venia-staging.com/isadora-skirt.html

Screen Shot 2019-10-14 at 2 17 13 PM

#### Desktop

Current: 

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-14 at 2 18 05 PM |  70.8 KB |

This PR:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-15 at 9 25 54 AM | 45.3 KB |

#### Mobile

Current:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-15 at 1 19 21 PM | 89.9 KB |

This PR:

| Image Network Requests | Total Image Data Requested | 
| --- | --- |
| Screen Shot 2019-10-15 at 1 23 45 PM |  90.1 KB |

These changes can be found in `ProductImageCarousel`'s `carousel.js` and `thumbnail.js`.

supernova-at:
### Search: Recommended Product

Screen Shot 2019-10-14 at 2 24 40 PM

No net change here, we're still requesting a 1.5 KB image.

These changes can be found in `SearchBar/suggestedProduct.js`.

supernova-at:
### MiniCart Product

Screen Shot 2019-10-14 at 2 30 02 PM

No net change here, we're still requesting a 1.5 KB image.

These changes can be found in `MiniCart/product.js`.

supernova-at:
Updated the verification steps ✅

packages/peregrine/lib/talons/Image/useResourceImage.js
```diff
@@ -0,0 +1,81 @@
...
+        return resourceUrl(resource, {
+            type,
+            height: resourceHeight,
+            width: resourceWidth
+        });
+    }, [resource, resourceHeight, resourceUrl, resourceWidth, type]);
+
+    const srcSet = useMemo(() => {
+        return generateSrcset(resource, type);
+    }, [generateSrcset, resource, type]);
```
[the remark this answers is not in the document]
jimbo: It's an interesting rabbit hole, since defining inline/unstable functions in components is a known antipattern. So `generateSrcSet` should never be inline. That said, there are two reasons to put it in the deps list:

1. If `generateSrcSet` changes across renders, it should result in a de-opt, not a bug. Including it in the deps list results in a de-opt (`useMemo` creating a new object). Excluding it from the deps list would result in a bug.
2. The lint rule, `exhaustive-deps`, is not smart enough to know which functions should be stable, and we can't really hint at it (afaik). When it sees a function from props that's missing from the deps list, it complains. Disabling the rule is more destructive than including a superfluous dependency.

packages/peregrine/lib/talons/Image/useResourceImage.js
```diff
@@ -0,0 +1,81 @@
...
+        // Helper function for prepending sizes media constraints.
+        const constrain = sizeName => {
+            const breakpoint = resourceSizeBreakpoints.get(sizeName) + 'px';
+            const size = getPixelSize(sizeName);
+
+            return `(max-width: ${breakpoint}) ${size}`;
+        };
+
+        // Note: it is assumed sizes will be filled from small, to medium, to large.
+        // In other words, having values for small and large but not medium is not supported.
```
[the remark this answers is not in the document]
jimbo: I'm fine with having a static set of named widths. That said, I think the main reason for it here has maybe been eliminated now that we're not sharing these widths between JS and CSS (right?).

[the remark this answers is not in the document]
zetlen: Also yes I just noticed that I'm putting the `useRef()` inside a conditional, but...Webpack will factor the conditional out, so it should be okay at runtime.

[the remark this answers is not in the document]
supernova-at: > the main reason for it here has maybe been eliminated now that we're not sharing these widths between JS and CSS

Yes, actually this is exactly right. I can change these to arrays 👍 .

packages/venia-ui/lib/components/CategoryList/categoryTile.js
```diff
@@ -1,9 +1,14 @@
...
 import { mergeClasses } from '../../classify';
-import { Link, resourceUrl } from '@magento/venia-drivers';
+import { Link } from '@magento/venia-drivers';
 import defaultClasses from './categoryTile.css';
 import { useCategoryTile } from '@magento/peregrine/lib/talons/CategoryList/useCategoryTile';
+import Image from '../Image';
+
+const IMAGE_WIDTH = 80;
+const IMAGE_SIZES = new Map();
+IMAGE_SIZES.set('small', IMAGE_WIDTH);
```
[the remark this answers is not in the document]
supernova-at: They don't have to be, no. Since it was new API I suppose I was trying to use / show it in as many places as possible. With these `Map`s changing to arrays, I can change this too.

packages/peregrine/lib/talons/Image/useResourceImage.js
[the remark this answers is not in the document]
supernova-at: I created JIRA-90 to tackle this.

PR #1848, opened 2019-10-04
[PB] Buttons component

eug123:
## Description
Introduce Buttons and ButtonItem components for Page Builder respective content types

## Related Issue
https://jira.corp.magento.com/browse/PB-12

## Verification Steps
1. Check that each property for each content type appearance renders correctly on PWA Venia storefront in desktop and mobile.
2. Compare Venia PWA rendering of content type with Luma to ensure the experience is the same or better for each property of each appearance in desktop and mobile.
3. Check each supported browser and on mobile that is supported according to PWA documentation.
4. Acceptable performance has been validated

PR #1872, opened 2019-10-09
[PB] Page Builder

davemacaulay:
## Description
Introduce Page Builder support to PWA Studio

## Related Issue
https://jira.corp.magento.com/browse/PB-1

## Verification Steps


packages/venia-ui/lib/components/Button/button.css
```diff
@@ -56,7 +56,6 @@
 .root_lowPriority {
     composes: root;
     border-width: 0;
-    color: rgb(var(--venia-error));
```
sirugh: This changes the forgot password button text from red to black. Luckily it's the only place, but this is a public API. Anyone using `Button` with `low` priority will see this change.

https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/SignIn/signIn.js#L66

packages/venia-ui/lib/RootComponents/CMS/cms.js
```diff
@@ -1,8 +1,42 @@
...
+import React, { useEffect } from 'react';
+import { useLazyQuery } from '@apollo/react-hooks';
+import cmsPageQuery from '../../queries/getCmsPage.graphql';
+import { fullPageLoadingIndicator } from '../../components/LoadingIndicator';
+import RichContent from '../../components/RichContent';
 
-const CMS = () => {
-    return <CategoryList title="Shop by category" id={2} />;
+const CMSPage = props => {
+    const { id } = props;
```
sirugh: Add a `CMS.propTypes` please :)

pwa-devdocs/src/_data/top-nav.yml
```diff
@@ -7,6 +7,9 @@
 - label: Peregrine Components
   url: /peregrine/
 
+- label: PageBuilder
```
jcalcaben: This should not be published yet, so this entry needs to be removed.

docker/.env.docker.dev
```diff
@@ -14,3 +14,4 @@ UPWARD_JS_UPWARD_PATH=upward.yml
 UPWARD_JS_BIND_LOCAL=1
 UPWARD_JS_LOG_URL=1
 CHECKOUT_BRAINTREE_TOKEN=sandbox_8yrzsvtm_s2bg8fs563crhqzk
+GOOGLE_MAPS_API_KEY=redacted
```
sirugh: Do you have a sandbox dev token?

packages/venia-ui/lib/RootComponents/CMS/cms.js
```diff
@@ -1,8 +1,42 @@
...
+    const [runQuery, queryResponse] = useLazyQuery(cmsPageQuery);
```
sirugh: Why `lazyQuery`?

pwa-devdocs/src/_data/top-nav.yml
davemacaulay: @bdenham can you resolve this please.

packages/venia-ui/lib/components/RichContent/richContent.js
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @param {Object} props React component props
+ *
+ * @returns {React.Element} A React component that renders Heading with optional styling properties.
+ */
+const RichContent = props => {
+    const { html } = props;
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    if (detectPageBuilder(html)) {
```
sirugh: It occurs to me that not everyone will have PageBuilder but everyone using PWA Studio will get _all_ the PageBuilder code bundled, right? Or do you do do some lazy load? I think you must...

davemacaulay: You're right, if someone isn't using Page Builder but does invoke `<RichContent />` they'll pull down some of Page Builder (well in fact it'll be in the bundle).

This was _somewhat intentional_ as if Page Builder is used on the store we want to ensure important content is loaded _as quickly as possible_. We want to avoid huge jumps in content loads. Including `<PageBuilder />` with a `React.lazy` could solve this, but I believe that'd make the rendering of the actual pages content delayed (even if just by a few ms).

Scenario: a homepage with content on it. The content is the most important part of said page, so it should have priority for the render. 

@zetlen we discussed around this topic a little bit, would love to hear your views here. Should we just `React.lazy` the whole `<PageBuilder />` component? or is the current situation okay?

docker/.env.docker.dev
davemacaulay: Google Dev Console requires you to add a payment method to the API key now. So while we do maintain a few within the team we recommend each team to maintain their own, otherwise we'd end up with a large bill on someones card. 

_This is annoying and we contacted Google about getting a sandbox key, but they said no._

pwa-devdocs/src/_data/top-nav.yml
bdenham: Yep. Got it. I will remove top level nav and move all page builder docs into _drafts immediately after the PR for PB-36 branch is finished (to make reviewing preview easier on reviewers).

docker/.env.docker.dev
sirugh: Understandable lol

packages/venia-ui/lib/components/Button/button.css
davemacaulay: I've updated the signin component to set the color for this specific instance of the button. We worked with UX and determined all low priority buttons shouldn't have the error color.

packages/venia-ui/lib/RootComponents/CMS/cms.js
davemacaulay: Updated to use `useQuery`: https://github.com/magento/pwa-studio/pull/1872/commits/c11d5364191e687cfc394f354c3a12740dea1aef

packages/venia-ui/lib/RootComponents/CMS/cms.js
davemacaulay: Done! https://github.com/magento/pwa-studio/pull/1872/commits/c11d5364191e687cfc394f354c3a12740dea1aef

packages/venia-ui/lib/components/RichContent/richContent.js
davemacaulay: Spoke about this with @zetlen and using `stats:venia` we identified that PageBuilder will only be included in the CMS Page chunk. We agreed that we shouldn't implement something now to resolve this and instead wait until we have a better extensibility model which will allow us to disable / enable Page Builder.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Banner/banner.js
```diff
@@ -0,0 +1,312 @@
...
+        const linkProps = resolveLinkProps(link, linkType);
+        const LinkComponent = linkProps.to ? Link : 'a';
+        BannerFragment = (
+            <LinkComponent
+                className={classes.link}
+                {...linkProps}
+                {...(openInNewTab ? { target: '_blank' } : '')}
+                onDragStart={event => {
+                    event.preventDefault();
+                }}
```
jimbo: Avoid recreating event handlers on render, since they have to be reattached. Create and memoize the handler, then attach it (or don't) as necessary.

```jsx
const Banner = props => {
  // define it with `useCallback`
  const handleDragStart = useCallback(event => {
    event.preventDefault()
  }, [])

  return <LinkComponent onDragStart={handleDragStart} />
}

// or, since there are no dependencies, just define it outside of the component
const handleDragStart = event => event.preventDefault

const Banner = props => {
  return <LinkComponent onDragStart={handleDragStart} />
}
```

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Block/block.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import CmsBlock from '../../../../CmsBlock/block';
```
jimbo: Given that `CmsBlock` does nothing, why doesn't this component use `RichContent` directly?

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/ButtonItem/buttonItem.js
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        if (openInNewTab) {
+            window.open(url, '_blank');
+        } else if (linkProps.to) {
+            props.history.push(url);
+        } else {
+            window.location.assign(url);
+        }
+    }, [openInNewTab, props.history, url, linkProps.to]);
```
jimbo: I left a [comment](https://github.com/magento/pwa-studio/pull/1848/files#r333615434) on the individual PR for this component, and it went unanswered. Does `props.history` change?

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Buttons/buttons.js
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        const { current: el } = ref;
+        let min = 0;
+        for (const button of el.querySelectorAll('button')) {
+            const { offsetWidth } = button;
+            if (offsetWidth > min) min = offsetWidth;
+        }
+        setMinWidth(min);
+    }, [isSameWidth]);
```
jimbo: Nice to see this at work. 👍

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Column/__tests__/configAggregator.spec.js
```diff
@@ -0,0 +1,79 @@
...
+            background-position: left top;
+            background-size: cover;
+            background-repeat: no-repeat;
+            background-attachment: scroll;
+            text-align: center;
+            border-style: solid;
+            border-color: rgb(0, 0, 0);
+            border-width: 1px;
+            border-radius: 10px;
+            width: calc(33.3333% - 20px);
```
jimbo:
```suggestion
            width: calc((100% / 3) - 20px);
```

If you like, you could make this change in the other places this kind of precalculation appears.

jest.config.js
```diff
@@ -79,7 +79,9 @@ const testVenia = inPackage => ({
     // @magento namespaced packages like Peregrine and Venia UI as well, when
     // it's testing Venia. That way, changes in sibling packages don't require a
     // full compile.
-    transformIgnorePatterns: ['node_modules/(?!@magento/)'],
+    transformIgnorePatterns: [
+        'node_modules/(?!@magento|jarallax|video-worker/)'
```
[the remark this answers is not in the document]
davemacaulay: Yup, we include a library called Jarallax (ironically for Parallax). We opted to use this library rather than a dedicated React library so we could easily achieve feature parity.

packages/venia-ui/lib/components/CmsBlock/block.js
```diff
@@ -1,27 +1,28 @@
...
+ * CMS Block component.
+ *
+ * @typedef Block
+ * @kind functional component
+ *
+ * @param {props} props React component props
+ *
+ * @returns {React.Element} A React component that displays a CMS Block.
+ */
+const Block = ({ content }) => <RichContent html={content} />;
```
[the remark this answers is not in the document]
davemacaulay: We did this due to the semantics of how Magento developers understand building pages. I'd personally expect there to be a CMS Block component that handled this for me. However, I can appreciate this component isn't doing anything more useful than just being named something related in Magento.

Should we remove this or leave it for Magento developers ease of understanding how to render a CMS Block?

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Block/block.js
davemacaulay: See https://github.com/magento/pwa-studio/pull/1872#discussion_r336003557

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Column/__tests__/configAggregator.spec.js
davemacaulay: Updated all occurrences: https://github.com/magento/pwa-studio/pull/1872/commits/51e840ee0742ea494f1d6b57c0b062c8fe83d1c1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Banner/banner.js
davemacaulay: Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/51e840ee0742ea494f1d6b57c0b062c8fe83d1c1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Banner/banner.js
```diff
@@ -0,0 +1,312 @@
+import React, { useState } from 'react';
+import defaultClasses from './banner.css';
+import { mergeClasses } from '../../../../../classify';
+import { arrayOf, bool, object, oneOf, shape, string } from 'prop-types';
+import Button from '../../../../Button/button';
+import resolveLinkProps from '../../resolveLinkProps';
+import { Link, withRouter } from '@magento/venia-drivers';
```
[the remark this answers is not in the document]
davemacaulay: Good catch, maybe a good example of an ESLint rule? 
Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/51e840ee0742ea494f1d6b57c0b062c8fe83d1c1

jest.config.js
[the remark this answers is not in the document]
jimbo: Sure. Does Jarallax need to be transpiled by `babel-jest` though? Ordinarily any third-party code in `node_modules` doesn't need to be, so I'm not sure why these two libraries would.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Banner/banner.js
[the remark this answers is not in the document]
jimbo: Yes, if there's a lint rule for that, it'd be nice to have in our config. 👍

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Banner/banner.js
jimbo: Sorry, there was a small bug in that suggestion (forgot to call `preventDefault`).

```es6
const handleDragStart = event => event.preventDefault()
```

davemacaulay: @jimbo and I'm the silly one for blindly copying it without checking 🤦‍♂ 
Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/e982a8cb3d3237c94df8329d000fdb57f24b3613

packages/venia-ui/lib/components/CmsBlock/block.js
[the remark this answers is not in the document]
davemacaulay: @jimbo did some quick investigation and the default export of `CmsBlock` is `cmsBlock.js`, not `block.js`. `block.js` is consumed by `cmsBlock.js`. 

CMS Blocks by default provide `content` from GraphQL where as `<RichContent />` expects a `html` prop. `block.js` is currently handling this prop translation, so I think it provides some value.

I don't feel like we should refactor `cmsBlock.js` this late in an epics lifecycle for the sake of a very small component. Thoughts?

jest.config.js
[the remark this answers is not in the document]
davemacaulay: `video-worker` is a dep of Jarallax, so they came in through the same feature. 

This is because Jarallax is published _very weirdly_, their index file points out their `src` which has ES6 code within. Without this entry into `transformIgnorePatterns` it errors on the `import` statements within the library.

```
const jarallax = require('./src/jarallax.esm').default;
const jarallaxVideo = require('./src/jarallax-video.esm').default;
const jarallaxElement = require('./src/jarallax-element.esm').default;

module.exports = {
    jarallax,
    jarallaxElement() {
        return jarallaxElement(jarallax);
    },
    jarallaxVideo() {
        return jarallaxVideo(jarallax);
    },
};

```

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Banner/banner.js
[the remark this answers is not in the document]
davemacaulay: Opened a PR and resolved all issues in current PR: https://github.com/magento-research/magento-eslint/pull/9
https://github.com/magento/pwa-studio/pull/1872/commits/564920846f2d12368021f4a70677a75dce6ece90

packages/venia-ui/lib/components/CmsBlock/block.js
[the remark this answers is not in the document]
jimbo: That's fine. I'll do it in a separate PR after yours is merged, and label it with `Major`.

jest.config.js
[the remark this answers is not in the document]
jimbo: That's the answer I was looking for, thanks. 👍

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Map/map.js
```diff
@@ -0,0 +1,283 @@
...
+                        position,
+                        title: location.name
+                    });
+
+                    const infoWindow = new googleMaps.InfoWindow({
+                        content: getLocationFormattedAsHtml(location),
+                        maxWidth: 350
+                    });
+
+                    marker.addListener('click', () => {
```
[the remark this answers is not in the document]
davemacaulay: `addListener` is a Google Maps API, `addEventListener` doesn't exist on a marker, gross.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Map/map.js
```diff
@@ -0,0 +1,283 @@
...
+ * This component is part of the Page Builder / PWA integration. It can be consumed without Page Builder.
+ *
+ * @typedef Map
+ * @kind functional component
+ *
+ * @param {props} props React component props
+ *
+ * @returns {React.Element} A React component that displays a Map.
+ */
+const Map = props => {
```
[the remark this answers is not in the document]
davemacaulay: Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/66a2a448c1162e1a52fb69e63d5ddbcde589e386
Retaining folder name `Map` due to the folder names relating to the content types.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
```diff
@@ -0,0 +1,297 @@
...
+        setIsScrolling(false);
+    }, [setIsScrolling]);
+
+    const onMouseMove = useCallback(
+        event => {
+            if (isScrolling && scrollElement) {
+                scrollElement.scrollLeft = scrollX + (clientX - event.clientX);
+                setScrollX(scrollElement.scrollLeft);
+                setClientX(event.clientX);
+            }
```
jimbo: Setting state and triggering a render on `mousemove` is generally unwise. This event fires rapidly and reconciliation is not that fast, especially for nontrivial components such as this one, and even less so if you have side effects on each render (which this one does).

Consider throttling this event, or perhaps constraining its updates to a ref and then using an effect to sync them with the DOM on an interval.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
```diff
@@ -0,0 +1,297 @@
...
+ * @returns {React.Element} A React component that displays a set of Tabs.
+ */
+const Tabs = props => {
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const navigationRef = useRef(null);
+    const [isScrolling, setIsScrolling] = useState(false);
+    const [clientX, setClientX] = useState(0);
+    const [scrollX, setScrollX] = useState(0);
+    const [scrollElement, setScrollElement] = useState(null);
+    const [navWrapperClass, setNavWrapperClass] = useState();
```
jimbo: Rather than keeping classnames in state, you should keep local strings in state and retrieve the associated classname only at the very end as you're creating elements.

```jsx
const [gradient, setGradient] = useState("")

// note that this effect no longer depends on `classes`
useEffect(() => {
  // here, "BOTH" is the *actual state*
  // classname isn't state, it's a derivative of state
  setGradient("BOTH")
}, [setGradient])

// consider making a mapping outside of the component definition
// similar to how you look up flex properties
return (
  <div className={gradientClasses.get(gradient)}>
)
```

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
```diff
@@ -0,0 +1,297 @@
...
+                setClientX(event.clientX);
+            }
+        },
+        [isScrolling, scrollElement, scrollX, clientX]
+    );
+
+    useEffect(() => {
+        let navScrollElement;
+        const navigationWrapper = navigationRef.current;
+        const handleScroll = () => {
```
jimbo: I believe this handler can be created with `useCallback()` outside of this effect.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
```diff
@@ -0,0 +1,297 @@
...
+        >
+            <div className={navWrapperClass} ref={navigationRef}>
+                <TabList
+                    onMouseDown={onMouseDown}
+                    onMouseUp={onMouseUp}
+                    onMouseMove={onMouseMove}
+                    onMouseLeave={onMouseUp}
+                    className={navigationClass}
+                >
+                    {headers.map((header, i) => (
```
jimbo: Please avoid multiline expressions such as this in JSX, especially mapping. The way you defined `tabPanels` is ideal. 👍

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Video/video.css
```diff
@@ -0,0 +1,28 @@
+.root {
+    font-size: 0;
+}
+
+.inner {
+    box-sizing: border-box;
```
jimbo: Likewise, I believe this inherits from our `index.css`.

packages/venia-ui/lib/components/RichContent/PageBuilder/factory.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @param data
+ * @returns {*}
+ * @constructor
+ */
+const ContentTypeFactory = ({ data }) => {
+    if (data.isHidden) {
+        return null;
+    }
+    delete data.isHidden;
```
jimbo: Minor suggestion: avoid mutating arguments.

```es6
const { isHidden, ...rest } = data

if (isHidden) {
  return null
}

const contentType = renderContentType(config.component, rest)
```

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
davemacaulay: This is a really neat solution, thanks for suggesting it :)

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Products/products.js
```diff
@@ -0,0 +1,153 @@
...
+const restoreSortOrder = (skus, products) => {
+    const sortedProducts = [];
+    skus.forEach(sku => {
+        const productBySku = products.find(product => product.sku === sku);
+        if (productBySku) {
+            sortedProducts.push(productBySku);
+        }
+    });
+    return sortedProducts;
+};
```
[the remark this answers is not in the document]
davemacaulay: Great suggestion, resolved: https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Row/row.js
```diff
@@ -0,0 +1,223 @@
+import React, { useEffect, useRef, useState } from 'react';
+import defaultClasses from './row.css';
+import { verticalAlignmentToFlex } from '../../utils';
+import { mergeClasses } from '../../../../../classify';
+import { arrayOf, oneOf, shape, bool, string, number } from 'prop-types';
+import { jarallax } from 'jarallax';
```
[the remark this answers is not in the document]
davemacaulay: Added it as a dynamic `require` once we determine we actually want the library: https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Slider/slider.css
```diff
@@ -0,0 +1,213 @@
...
+    height: auto;
+}
+
+.root :global .slick-slider .slick-track,
+.root :global .slick-slider .slick-list {
+    -moz-transform: translate3d(0, 0, 0);
+    -ms-transform: translate3d(0, 0, 0);
+    -o-transform: translate3d(0, 0, 0);
+    -webkit-transform: translate3d(0, 0, 0);
+    transform: translate3d(0, 0, 0);
```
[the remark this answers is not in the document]
davemacaulay: Oops, fixed! https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Slider/slider.js
```diff
@@ -0,0 +1,159 @@
...
+        dots: showDots,
+        arrows: showArrows,
+        infinite,
+        autoplay,
+        autoplaySpeed,
+        fade
+    };
+    cssClasses.push(classes.root);
+
+    // Override classes on banner to ensure min height is respected
```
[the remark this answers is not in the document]
davemacaulay: 🤷‍♂ I didn't want to include this in the Banner itself just in case it had any unwanted side effects.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/TabItem/tabItem.css
```diff
@@ -0,0 +1,4 @@
+.root {
+    box-sizing: border-box;
```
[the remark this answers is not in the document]
davemacaulay: Removed all instances of border-box: https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.css
```diff
@@ -0,0 +1,180 @@
...
+    --tabs-border: solid;
+    --tabs-border-color: #ccc;
+    --tabs-border-width: 1px;
+    --tabs-border-radius: 0px;
+}
+
+ul li.header.selected {
+    background: #fff;
+    z-index: 3;
+}
```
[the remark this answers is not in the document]
davemacaulay: Scoped all these inside of `.root` https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
davemacaulay: Oh daaamnn, yeah I did a test and this was causing the component to get rendered a lot. I've updated this to use `useRef` and now it doesn't re-render 5000 times.
https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
davemacaulay: Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
davemacaulay: Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
davemacaulay: Look at us, being inconsistent. Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Video/video.css
davemacaulay: All instances have been removed 😄 https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

packages/venia-ui/lib/components/RichContent/PageBuilder/factory.js
davemacaulay: Nice suggestion, resolved: https://github.com/magento/pwa-studio/pull/1872/commits/dc8fb48bc7c2edba5bd3573da6e9fa28e48adfc1

awilcoxa:
Appears homepage content is broken on develop if this is merged, we should resolve any issue that breaks current app functionality prior to merging.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
jimbo: Thanks for implementing it. 😄

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
```diff
@@ -0,0 +1,308 @@
...
+            }
+            navScrollElement.addEventListener('scroll', handleScroll);
+        }
+
+        return () => {
+            if (navScrollElement) {
+                navScrollElement.removeEventListener('scroll', handleScroll);
+            }
+        };
+    }, [navigationRef, handleScroll]);
```
jimbo: The linter should have warned you about this, but refs don't belong in `dependencies`.

- Refs are objects that never change (instead, their private value property changes), so they would never cause a re-render.
- Even `ref.current` doesn't belong as a dependency, since it's a getter.
- React recognizes refs, so it knows not to flag a ref as a missing dependency.

```suggestion
    }, [handleScroll]);
```

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
```diff
@@ -0,0 +1,308 @@
...
+    const onMouseDown = useCallback(event => {
+        isScrolling.current = true;
+        clientX.current = event.clientX;
+    }, []);
+
+    const onMouseUp = useCallback(() => {
+        isScrolling.current = false;
+    }, []);
+
+    const onMouseMove = useCallback(
```
jimbo: The convention for naming handlers is `handleEvent`, not `onEvent`. This way, handlers received from props (`onEvent`) can be distinguished from locally defined handlers (`handleEvent`). 

```suggestion
    const handleMouseMove = useCallback(
```

Recommend the same change for `onMouseDown` and `onMouseUp`, and don't forget to update other references in this file.

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
davemacaulay: I got a little confused by `useRef` so thanks for the detailed explanation. Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/50cbc1f9d2749c2067d71b61e71393336208f5b1

packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/Tabs/tabs.js
davemacaulay: Ah yeah, that makes more sense!
Resolved: https://github.com/magento/pwa-studio/pull/1872/commits/50cbc1f9d2749c2067d71b61e71393336208f5b1

davemacaulay:
@awilcoxa @jimbo 
> Appears homepage content is broken on develop if this is merged, we should resolve any issue that breaks current app functionality prior to merging.

We've added some logic to show the category list if the page is empty or it has default content. 

https://github.com/magento/pwa-studio/pull/1872/commits/e35b4635ef31f6ba8f4bb52b7410d4085244a511#diff-68c0ddcfb13cfa542ddc32f37857f438R30-R39

davemacaulay:
> Approving this from my end. Nice work all around. 👍
> 
> @dpatil-magento I think we can move this into your court now.

@dhaecker will QA this from our perspective before merging , @dpatil-magento there's no need for you to QA this (unless you wish too).

zetlen:
HONK HONK

PR #1874, opened 2019-10-09
Service Worker HTML Handling

### LINKED ISSUE #1781: [feature]: Cache only 1 version of HTML and return that for all routes.
Related to #1673 

As of today, all routes get the same HTML file even if the request originated as `/`, `/search.html` or `/venia-tops.html` or `/valeria-two-layer-tank.html` but with a different name. The problem arises when the user directly goes to a certain products page during the offline mode, whose HTML is not cached. Even though it is the same HTML since the route is different the browser will redirect to Network Unreachable error. This is because we give different names to routes that bear the same content and the service worker caches that HTML with the key as the route. We should change the HTML handler in service worker to save only 1 version of HTML and return it even if the request originated from a different route.

By not requesting HTML every time and utilizing a cached version till something changes we can save 855 Bytes of data over the network and also reduce our cache memory footprint.

out_temp

image

- [x] `venia-concept`
- [x] `venia-ui`
- [x] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

revanth0212:
## Description

Modified the HTML handling capabilities of the Service Worker to fetch and return the same HTML file irrespective of the page the client is in. This is because HTML for `/` is the same `venia-bottoms.html` or `athena-tank-dress.html`. Also since the same HTML is served for all the routes, offline capabilities are improved compared to `develop` as demonstrated in the demo gif.

## Related Issue
Closes #1781 

### Verification Steps
1. Start the server using `yarn build && yarn stage:venia`.
2. Open the app once the server is ready.
3. Navigate between multiple pages like category, product, etc. The server should only get requests for `/` unline `develop` in which each page will request its own HTML file.

html caching demo-low-res

## Checklist
* `venia:stage and venia:watch` should both work as `develop`. No business logic changed.
* Merge only after #1832 and #1859 are merged.


packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -0,0 +1,17 @@
...
+              headers: request.headers
+          })
+        : request;
+
+export const cacheHTMLPlugin = {
+    cacheKeyWillBeUsed: async () => {
+        return '/';
+    },
+    requestWillFetch: async ({ request }) => {
+        const newRequest = cloneRequestWithDiffURL(request, '/');
```
supernova-at: Just to check my understanding, if an "html route" comes in (`/` or ends in `.html`), we treat it as if it was `/` and just return the cache results from that. Or more specifically, issue a new request and let workbox handle returning the correct value.

Do I have that right?

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -0,0 +1,17 @@
+const cloneRequestWithDiffURL = (request, url) =>
```
supernova-at: Do you see potential uses for this outside of what we have in this PR?
If not, since the only place that calls this always passes in `/` for the `url`, you could just hardcode that here and make this easier to read.

packages/venia-concept/src/ServiceWorker/registerRoutes.js
```diff
@@ -2,24 +2,16 @@ import {
     isResizedCatalogImage,
     findSameOrLargerImage
 } from './Utilities/imageCacheHandler';
+import { isHTMLRoute } from './Utilities/routeHandler';
 import { createCatalogCacheHandler } from './Utilities/catalogCacheHandler';
 import { THIRTY_DAYS, MAX_NUM_OF_IMAGES_TO_CACHE } from './defaults';
+import { cacheHTMLPlugin } from './Utilities/htmlHandler';
```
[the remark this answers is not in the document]
revanth0212: There are more functions coming to this file in my next PR.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: I wanted to separate out concerns and have a function just deal with what it needs to and let other utils like clone to take care of cloning a request. But coming to your question, I wasn't sure if this will be useful elsewhere. Took the benefit of the doubt and separated it into its own function but leaving it in the same file. It can be moved and exported later if needed by others.

revanth0212: This functionality might be changed a bit in my next PR.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: Nailed it.

packages/venia-concept/src/ServiceWorker/Utilities/routeHandler.js
```diff
@@ -0,0 +1,4 @@
+export const isHomeRoute = url => new URL(url).pathname === '/';
+
+export const isHTMLRoute = url =>
+    isHomeRoute(url) || new RegExp('.html$').test(new URL(url).pathname);
```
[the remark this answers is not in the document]
revanth0212: Awesome. Done 👍

PR #1879, opened 2019-10-10
[PB] Image bugs

davemacaulay:
## Description
Resolve bugs in Image component for Page Builder

## Related Issue
https://jira.corp.magento.com/browse/PB-161

## Verification Steps
1. Image in PWA is completely broken (add background & mobile image, empty figure on storefront)
2. Image w/ only caption breaks on PWA
3. Mobile image doesn't work on image
4. SEO not appearing on PWA storefront for image
5. Links not appearing in PWA for image


davemacaulay:
> Do ya'll do lazy loading or render placeholders to avoid DOM jutter when the image finally loads?

Nope, we just use `<figure />` with an image. Fancy things like lazy loading / rendering placeholders weren't part of the AC.

PR #1881, opened 2019-10-11
commit hash at bundle banner

### LINKED ISSUE #1227: Display Current Version
It is often confusing to know which version of PWA you're using / testing against - especially with our automatic deployments to demo sites (https://magento-venia-develop.now.sh/, https://veniapwa.com/). 

A couple options are:

1. Add a webpack [Banner Plugin](https://webpack.js.org/plugins/banner-plugin/) that contains the current version number. Then we can just inspect the network request for the bundle (`client.js`) and verify the version number.
1. Add an endpoint that displays git commit status etc. (disable in production)
1. ~~Bake the version number into the app shell mustache template~~

dani97:
## Description

Adding current version in bundle generated
## Related Issue

Closes #1227 

### Verification Stakeholders

@supernova-at 

### Verification Steps

1. Run `yarn build`
2. Open `packages/venia-concept/dist/client.[hash].js`
3. Verify that it contains a comment at the top that starts with `@version` (ex: `@version v4.0.0-106-g2eceeb7b2`).

Screenshot 2019-10-11 at 11 01 50 AM

## Checklist


dani97:
@supernova-at Can you check what tests are failing, AWS tests asks lAM creds

supernova-at:
> Can you check what tests are failing, AWS tests asks lAM creds

We're messing with AWS at the moment, may not be anything. I'll post back here if there are indeed failures 👍

supernova-at:
Hmm, this is failing in Docker because `git` doesn't exist there (even after we explicitly tell it to install `git`).

@dpatil-magento tested and thinks that the way we copy our projects over is the culprit.

We'll have to do some more Docker-specific digging on this one. You shouldn't have to update this anymore, but if we can't figure it out we may have to come up with some alternate solutions to accomplish this task.

zetlen:
This is a great contribution, @dani97 ,  but yeah, we need it to be more general. Here's what I would recommend.

1. Add an env var definition to `packages/pwa-buildpack/envVarDefinitions.json` called `STAGING_BUILD_ID`. Describe it as a variable that Buildpack will put in a comment above all generated bundle files and the index.html file. Say that if it is empty at production build time, Buildpack will try to use Git, and if it will fall back to a unique build hash.

2. Change your code so it uses the env var first (use `projectConfig.section('staging').buildId`) if it is populated.

3. If the env var is empty, then you can shell out to Git! I recommend using `git describe --long --always --dirty=-dev`, because that's a human-readable format that includes some useful stuff:
    - The most recently tagged version
    - The number of commits since that tag
    - An abbreviation of the current commit hash
    - The suffix `-dev` if there are current outstanding changes in the work tree (like if you're working in a local environment)

4. If Git fails, catch the exception and just use the following:
    ```js
    `@version ${require(path.resolve(context, './package.json')).version}-[hash]`
    ```
  which will at least have the most recently published version, plus a unique Webpack hash.

That kind of fall-through strategy supports basically all scenarios. A controlled deployment process can populate the `STAGING_BUILD_ID` env var with whatever it wants. A deploy from a Git repo will use a sensible marker of the repo state. And a deploy with neither of those things will still use a Webpack hash so you can quickly compare files for identity.

How's that sound?

dpatil-magento:
Thanks @dani97 , looks great now.  Verified following :
`watch:all`
`watch:venia`
`stage:venia`
`bash docker/run-docker`
steps from verification section.

PR #1884, opened 2019-10-11
FilterModal talon

### LINKED ISSUE #1930: [bug]: Filters: unable to select fashion color filters
**Describe the bug**
User is unable to select Fashion colors filters. 

**To reproduce**
1. Go to any category and click on Filters button on `develop` https://magento.pwa-venia.com/
2. On Filters drawer expand Fashion Color section.
3. Click on any Fashion color. 

**Expected behavior**
Color should get selected as Filter.

**Expected behavior**
Colors are no more getting selected.

Note - Works fine in `4.0.0` release https://magento.pwa-venia.com/.  Broken in `develop` https://develop.pwa-venia.com/

**Screenshots**
**`develop`**

image

**`master` (4.0.0)**

image

sirugh:
## Description

Create Peregrine talons for the `FilterModal` component. As part of this work, refactor the component to use local state and hooks throughout, and improve its performance.

## Related Issue
Closes #1930 .

## Acceptance 
@sirugh 

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Navigate to a category page
2. Click `Filter` to open the filter modal (drawer)
3. Select some filters
4. Click `Apply Filters` to close the drawer and update the URL
5. Refresh the page
6. Reopen the filter modal and verify that the state remains correct
7. Click `Reset Filters` to clear all of the filters
8. Click the mask or the `X` to close the drawer and update the URL
9. Repeat this process on the search page

## Checklist


packages/venia-ui/lib/components/FilterModal/FilterList/filterList.js
```diff
@@ -59,38 +60,43 @@ const FilterList = props => {
...
 
     return (
-        <List
-            items={items}
-            getItemKey={({ value_string }) => `item-${id}-${value_string}`}
-            render={props => <ul className={layoutClass}>{props.children}</ul>}
-            renderItem={({ item }) => {
-                const isActive = isFilterSelected(item);
+        <Form>
+            <FilterSearch name="foo" />
```
sirugh:
```suggestion
            <FilterSearch name={props.name} />
```

packages/peregrine/lib/talons/FilterModal/useFilterModal.js
```diff
@@ -0,0 +1,176 @@
...
+            const nextSearch = getSearchFromState(
+                search,
+                filterKeys,
+                filterState
+            );
+
+            // write filter state to history
+            history.push({ pathname, search: nextSearch });
+            // close the drawer
+            closeDrawer();
```
sirugh: Why is this effect a thing and not just what the `handleApply` callback does?

packages/peregrine/lib/talons/FilterModal/CurrentFilters/useCurrentFilters.js
```diff
@@ -0,0 +1,24 @@
+import { useCallback } from 'react';
+import { useCatalogContext } from '@magento/peregrine/lib/context/catalog';
+
+export const useCurrentFilters = props => {
+    // TODO: Replace with useRouter.
+    const { history, location } = props;
+
+    const [{ chosenFilterOptions }, { removeFilter }] = useCatalogContext();
```
sirugh: I noticed we don't hook up the `addFilter` from the catalog actions. Can the filter related actions be removed from catalog? I think we can use your new `removeItem` filter action.

jimbo: Yes, that's correct. Filter state is no longer connected to Redux at all. I need to remove that stuff.

packages/peregrine/lib/talons/FilterModal/useFilterModal.js
jimbo: Well, a callback that did all of those things would end up being recreated whenever `filterState` changed. This way, the effect has the latest dependencies, but the callback never changes.

I do wonder about this pattern though. Do you think it's too indirect?

packages/peregrine/lib/talons/FilterModal/useFilterState.js
```diff
@@ -0,0 +1,115 @@
...
+        case 'set items': {
+            return init(payload);
+        }
+    }
+};
+
+export const useFilterState = () => {
+    const [state, dispatch] = useReducer(reducer, null, init);
+
+    const addItem = useCallback(
```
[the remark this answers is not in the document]
jimbo: No, but it's an important part of the API. Our UI just happens to prefer `toggleItem`.

packages/peregrine/lib/talons/FilterModal/useFilterState.js
```diff
@@ -0,0 +1,115 @@
...
+            return nextState;
+        }
+        case 'set items': {
+            return init(payload);
+        }
+    }
+};
+
+export const useFilterState = () => {
+    const [state, dispatch] = useReducer(reducer, null, init);
```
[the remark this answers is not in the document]
jimbo: Yes, I'll add it.

packages/venia-ui/lib/components/FilterModal/FilterFooter/filterFooter.js
```diff
@@ -1,81 +1,45 @@
...
+    const handleApplyClick = useCallback(() => {
+        applyFilters();
+    }, [applyFilters]);
 
-    const applyButtonClass = areOptionsPristine
-        ? classes.applyButtonDisabled
-        : classes.applyButton;
+    const handleResetClick = useCallback(() => {
+        clear();
+        applyFilters();
```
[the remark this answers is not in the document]
jimbo: Yes, I meant to bring this up. This happens because `history.push()` and `closeDrawer()` are part of the same callback/effect now. We have to persist to the URL after resetting the filters or else the reset will be lost if they click the mask, and since I'm invoking the callback/effect to persist to the URL, it's also closing the drawer.

I'll see what happens if I rework the callback/effect.

packages/peregrine/lib/talons/FilterModal/useFilterModal.js
sirugh: Aside from readability I'm not sure there's a drawback to this method. It's just tough to parse at first but I guess the alternative would be to pass state directly to the callback which means _something_ has to be aware of state.

packages/venia-ui/lib/components/FilterModal/FilterFooter/filterFooter.js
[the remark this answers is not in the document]
jimbo: This has been fixed.

packages/peregrine/lib/talons/FilterModal/useFilterModal.js
jimbo: This pattern really is necessary. There are alternatives, such as passing state around or accessing it through a ref, but I think this is the cleanest one so far. We can always revisit, since this isn't a totally unique case.

packages/peregrine/lib/talons/FilterModal/CurrentFilters/useCurrentFilters.js
jimbo: All references to Redux state have been removed. I also removed all of the effects that formerly had to check for filter params.

jimbo:
For now, I've decided not to make talons for the last child components here, `CurrentFilters` and `FilterList`, at least for now. Right now, making talons for them would make them less efficient, and in the latter case that might actually matter. Making components for them will have to be a separate story that has performance as part of the AC.

packages/venia-ui/lib/components/FilterModal/filterFooter.js
```diff
@@ -0,0 +1,37 @@
...
+    const { applyFilters, hasFilters, resetFilters } = props;
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    return (
+        <div className={classes.root}>
+            <Button disabled={!hasFilters} onClick={resetFilters}>
+                {'Reset Filters'}
+            </Button>
+            <Button
+                disabled={!hasFilters}
```
sirugh: If I toggle a filter and click `Apply Filters`, but then want to clear it (by toggling it again or clicking the `x`) without clicking "reset", how do I do that? Right now if you cannot "apply" an empty set of filters but I would assume it would allow it. This behavior is present on develop so we don't have to change it but it jumped out as a weird UX when I did it.

packages/venia-ui/lib/components/ProductOptions/tile.css
```diff
@@ -3,8 +3,6 @@
     border: 1px solid rgb(var(--venia-text));
     border-radius: 2px;
     height: 3rem;
```
sirugh: Make sure to fix the media query that is busted with the 2rem height :o

jimbo: Also fixed.

packages/venia-ui/lib/components/FilterModal/FilterList/filterList.js
```diff
@@ -1,111 +1,86 @@
...
-import { compose } from 'redux';
-import { array, bool, object, shape, string } from 'prop-types';
+import React, { Fragment, Suspense, lazy, useMemo } from 'react';
+import { array, bool, shape, string } from 'prop-types';
+import { useFieldState } from 'informed';
+import setValidator from '@magento/peregrine/lib/validators/set';
+
 import { mergeClasses } from '../../../classify';
-import { withRouter } from 'react-router-dom';
+import FilterItem from './FilterItem';
```
sirugh:
```suggestion
import FilterItem from './filterItem';
```

packages/venia-ui/lib/components/FilterModal/filterBlock.js
```diff
@@ -1,118 +1,55 @@
...
-export default classify(defaultClasses)(FilterBlock);
+const FilterBlock = props => {
+    const { filterApi, filterState, group, items, name } = props;
+    const talonProps = useFilterBlock({ group });
+    const { handleClick, isExpanded, isSwatch } = talonProps;
+    const iconSrc = isExpanded ? ArrowUp : ArrowDown;
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    const listProps = { filterApi, filterState, group, isSwatch, items, name };
+    const listElement = isExpanded ? <FilterList {...listProps} /> : null;
```
sirugh: Previously we were using CSS to hide the list. I did a little performance check and it takes 0.2ms to render the list on my mac.

It's _just_ enough that I noticed it. Not sure if you want to address that in the PR but it's something of note.

jimbo: Fixed it anyway. 👍

dpatil-magento:
@jimbo Two UX things :
1. `Apply Filters` button is enabled by default.  It should be disabled. (Works in `develop` and `master`)
2. When you apply any filter and open filter overlay again > the filter sub section is expanded.  It should be collapsed. (Works in `master`  broken already in `develop`)

This PR - https://pr-1884.pwa-venia.com/
`develop` - https://develop.pwa-venia.com/
`master` - https://magento.pwa-venia.com/

jimbo:
> `Apply Filters` button is enabled by default.  It should be disabled. (Works in `develop` and `master`)

This isn't quite true, though it does differ from live. To be precise, `Apply Filters` is disabled at first, but it becomes enabled and stays enabled once the form has been touched (i.e., once any filter has been changed). While this may be different from live, it's an intentional, necessary improvement here.

If `Apply Filters` were simply disabled whenever there were no filters selected in the form, the user would be unable to click `Apply Filters` after clicking `Reset Filters`.

> When you apply any filter and open filter overlay again, the filter sub section is expanded.  It should be collapsed. (Works in `master`  broken already in `develop`)

This is also different from live but not quite true. To be precise, the state of the filter modal is preserved while closed (like all other drawers in the app, actually), so any filter blocks that are expanded when the modal is closed will remain expanded when the modal is reopened. (If the user were to collapse a filter block before clicking `Apply Filters`, it would still be collapsed when reopening the modal.)

This is also an intentional, necessary improvement.

jimbo:
> To be precise, `Apply Filters` is disabled at first, but it becomes enabled and stays enabled once the form has been touched (i.e., once any filter has been changed).

I did notice, though, that the `touched` state was not being reset to `false` when the modal was closed. This has been fixed. When reopening the modal, `Apply Filters` will now be disabled if there are no selected filters, or enabled if there are selected filters.

sirugh:
Nice fix. 👍 from me

abelyanvahe:
@sirugh  Hi. So I see that you moved the state to local state. But if you remember, the getCategory query happens in the category component, which is outside the FilterModal component. So How you are going to get the filters there in order to send it with the getCategory query? 
In the previous version, I could easily get it from Redux via context, but now I don't see any ways. 
Thanks in advance.

sirugh:
@abelyanvahe good point. Can you open an issue with your concern and ping @jimbo as he may have some input :) We can discuss there.

abelyanvahe:
OK Sure. Will do now.

PR #1886, opened 2019-10-12
[PB] Tabs & Tab item component

davemacaulay:
## Description
Introduce Tabs and TabItem components for Page Builder respective content types

## Related Issue
https://jira.corp.magento.com/browse/PB-9

## Verification Steps


packages/venia-ui/lib/components/RichContent/PageBuilder/ContentTypes/TabItem/tabItem.js
```diff
@@ -1,3 +1,166 @@
...
+        borderRadius,
+        marginTop,
+        marginRight,
+        marginBottom,
+        marginLeft,
+        paddingTop,
+        paddingRight,
+        paddingBottom,
+        paddingLeft,
+        children
```
eug123: cssClasses prop missed

PR #1893, opened 2019-10-15
Removed duplicate words from sentences.

sanganinamrata:
## Description
This pull request is to remove unnecessary duplicate words in a couple of pages.


packages/pwa-buildpack/lib/WebpackTools/PWADevServer.js
```diff
@@ -52,7 +52,7 @@ const PWADevServer = {
                 server.middleware.waitUntilValid(() => {
                     // We can try to set the hostname and port for the dev
                     // server all we want, but the only reliable way to know
-                    // what it is is to detect it once it's mounted and
+                    // what it is to detect it once it's mounted and
```
supernova-at:
```suggestion
                    // what it is is to detect it once it's mounted and
```

supernova-at: This one actually doesn't need to be changed.

sanganinamrata: @supernova-at Done :)

PR #1898, opened 2019-10-16
[chore?]: Use createCustomer gql mutation instead of create-account REST endpoint

### LINKED ISSUE #1210: [bug]: If registered user tries to create account with email already in use then system does not show proper error message.
Pre-condition - 
1. Customer should be created 

Steps - 
1. Go to venia storefront and try to create account with customer created in pre-condition.

Expected - User should be shown with proper error message like "Email not available. Please try different email". Please check with Business on message content. Or check what error API returns in this case.
Actual - User is shown with message "An error occurred. Please try again" which is confusing.

sirugh:
## Description

This change replaces the old, unused email validation function *and* the `createAccount` redux/rest action with usage of the `createCustomer` mutation.

## Related Issue

Closes #1210.

### Verification Steps

1. Go to `/create-account` and try to create an account with an email that is already used. You should see the new error.
2. Go to the create account view in left nav and do the same. You should see the error.
3. Now, in both views, try to create accounts with valid emails and it should succeed.

## Checklist


packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
```diff
@@ -1,24 +1,59 @@
...
+    const errors = [];
+    if (createAccountError) {
+        errors.push({
+            type: 'CREATE_ACCOUNT_ERROR'
+        });
+    }
+
+    if (data && !data.isEmailAvailable.is_email_available) {
+        errors.push({
+            type: 'EMAIL_UNAVAILABLE'
```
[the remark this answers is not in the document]
sirugh: Yea I was going to document them at least.

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
```diff
@@ -1,24 +1,59 @@
...
+
+    // Mapping of message to type is done in the UI component.
+    const errors = [];
+    if (createAccountError) {
+        errors.push({
+            type: 'CREATE_ACCOUNT_ERROR'
+        });
+    }
+
+    if (data && !data.isEmailAvailable.is_email_available) {
```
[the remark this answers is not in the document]
sirugh: Would be! I might be able to rewrite the query.

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
```diff
@@ -1,24 +1,59 @@
...
+                }
+            });
+        },
+        [runQuery]
+    );
+
+    useEffect(() => {
+        if (called && !loading && !error && data) {
+            if (data.isEmailAvailable.is_email_available) {
+                onSubmit();
```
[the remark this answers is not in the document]
sirugh: This is really saying "when we have done pre-submission validation and everything is OK, submit the form to the creation endpoint.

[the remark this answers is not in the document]
supernova-at: That makes sense. I guess I was expecting to see this logic in the `handleSubmit` callback initially. 

And that's what you are doing essentially, it's just a quirk of hooks that makes us have to split it up into two hooks like this.

Would be nice if we could do something like `const data = await runQuery` and have it all in one place.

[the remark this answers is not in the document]
sirugh: I'll go through this today and add some comments. Will try to clean up where possible to make it more clear about whats happening.

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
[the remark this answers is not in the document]
sirugh: So I couldn't rewrite the query _but_ there is a `createCustomer` mutation which I may be able to use instead of the create account endpoint. I'm going to go down a rabbit hole of rewriting this again :D

packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -78,31 +78,6 @@ export const getUserDetails = () =>
         }
     };
 
-export const createAccount = accountInfo => async dispatch => {
```
sirugh: Bye Felicia!

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -49,13 +49,9 @@ export const useAuthModal = props => {
...
 
-    const handleCreateAccount = useCallback(
-        async values => {
-            await createAccount(values);
-            showMyAccount();
-        },
-        [createAccount, showMyAccount]
-    );
+    const handleCreateAccount = useCallback(() => {
+        showMyAccount();
```
sirugh: See how this callback no longer cares about the `createAccount` action. So clean!

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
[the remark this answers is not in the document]
sirugh: Resolving this conversation. I converted to a mutation and this became _much_ cleaner.

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
```diff
@@ -1,24 +1,62 @@
...
+                        lastname: formValues.customer.lastname,
+                        password: formValues.password
+                    }
+                });
+
+                // Then sign the user in.
+                await signIn({
+                    username: formValues.customer.email,
+                    password: formValues.password
+                });
```
sirugh: This logic used to exist within the redux action but now lives in the talon!

packages/venia-ui/lib/queries/createAccount.graphql
```diff
@@ -0,0 +1,19 @@
...
+  createCustomer(
+    input: {
+      email: $email
+      firstname: $firstname
+      lastname: $lastname
+      password: $password
+    }
+  ) {
+    customer {
+      email
```
sirugh: I tried removing the `email` and `customer` props but you are required to request _something_ from the response. So even though the UI doesn't do anything with this data, we still have to ask for it. Not a big deal, IMO.

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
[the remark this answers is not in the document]
sirugh: Removed this stuff.

packages/venia-ui/lib/queries/createAccount.graphql
supernova-at: Good to know, and good that we're requesting a very small piece of data. Mind dropping a comment in the file saying as much?

packages/venia-ui/lib/queries/createAccount.graphql
```diff
@@ -0,0 +1,19 @@
+mutation createAccount(
```
supernova-at: Since this is the first mutation, I'm wondering if we shouldn't create a `lib/mutations/` folder. It's a bit weird to read the import:

```js
import CREATE_ACCOUNT_MUTATION from '../../queries/createAccount.graphql';
```

packages/venia-ui/lib/components/CreateAccount/createAccount.js
```diff
@@ -18,19 +18,31 @@ import {
...
+import CREATE_ACCOUNT_MUTATION from '../../queries/createAccount.graphql';
 
 const LEAD =
     'Check out faster, use multiple addresses, track orders and more by creating an account!';
 
 const CreateAccount = props => {
     const talonProps = useCreateAccount({
-        initialValues: props.initialValues
+        initialValues: props.initialValues,
+        query: CREATE_ACCOUNT_MUTATION,
```
supernova-at: Only because in GraphQL parlance "queries" and "mutations" are two distinctly separate things, this reads funny.

packages/venia-ui/lib/queries/createAccount.graphql
sirugh: Actually I don't think I should. This is just how GraphQL works: https://github.com/apollographql/graphql-tools/issues/277

packages/venia-ui/lib/queries/createAccount.graphql
sirugh: A mutation is just a type of query. I don't think we should do this.

packages/venia-ui/lib/components/CreateAccount/createAccount.js
sirugh: Yea -- I thought about changing the import name. I don't really think the UI needs to care about whether the query is a mutation/query/subscription.

I'd like to get @jimbo's take on naming here.

packages/venia-ui/lib/queries/createAccount.graphql
sirugh: Left more detail below.

packages/venia-ui/lib/components/CreateAccount/createAccount.js
```diff
@@ -18,19 +18,31 @@ import {
...
+        initialValues
+    } = talonProps;
 
-    const errorMessage = hasError
-        ? 'An error occurred. Please try again.'
+    // Map over any errors we get and display an appropriate error.
+    const errorMessage = errors.length
+        ? errors
+              .map(({ message }) => message)
+              .reduce((acc, msg) => acc + '\n' + msg, '')
```
sirugh: I need to flip this around :o

packages/venia-ui/lib/queries/createAccount.graphql
brendanfalkowski: Maybe "a mutation is just a type of request"?

packages/venia-ui/lib/queries/createAccount.graphql
supernova-at: Yeah, the comment request was just to help folks learn ;)

No big deal

packages/venia-ui/lib/queries/createAccount.graphql
supernova-at: [Queries and Mutations are separate, special types](https://graphql.org/learn/schema/#the-query-and-mutation-types) that serve as entry points for the respective operations.

The Magento GraphQL schema separates Queries from Mutations, here's GraphiQL:

Screen Shot 2019-10-21 at 3 44 58 PM

Whether or not the front end splits these things up into `queries` and `mutations` directories is a discussion worth having, but GraphQL and our Magento GraphQL schema does separate them, so I assumed we would want to too.

Is it the worst thing in the world if a `.graphql` file in the `queries/` directory contained `addConfigurableProductsToCart`? Probably not. 🤷‍♂

packages/venia-ui/lib/components/CreateAccount/createAccount.js
supernova-at: 👍 pending @jimbo 's verdict.

packages/venia-ui/lib/queries/createAccount.graphql
sirugh: Added!

packages/venia-ui/lib/components/CreateAccount/createAccount.js
jimbo: > Whether or not the front end splits these things up into `queries` and `mutations` directories is a discussion worth having, but GraphQL and our Magento GraphQL schema does separate them, so I assumed we would want to too, if only to reduce confusion when going back and forth.

I'm pretty convinced by @supernova-at's argument here. As long as we have a `queries` folder, it makes sense have a `mutations` folder, since they're separate entities.

> Since this is the first mutation, I'm wondering if we shouldn't create a lib/mutations/ folder.

That said, this is a reminder that there is no good reason to have a `queries` folder in the first place. Queries (and mutations) are supposed to be highly tailored to the component using them; it's not considered an optimization to reuse a query, so there's no value in keeping them in a shared folder.

Ideally, queries and mutations would reside in the component folder alongside `.js` and `.css` component files. If we're not going to do that, a `mutations` directory is fine.

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
```diff
@@ -1,24 +1,61 @@
...
-    const { initialValues = {} } = props;
+    const { initialValues = {}, onSubmit, query } = props;
+    const [isSubmitting, setIsSubmitting] = useState(false);
 
-    const [
-        { createAccountError, isCreatingAccount, isSignedIn }
-    ] = useUserContext();
-    const hasError = !!createAccountError;
+    const [{ isSignedIn }, { signIn }] = useUserContext();
+    const [createAccount, { error }] = useMutation(query);
```
jimbo: I'm so glad we didn't have mutations prior to official hooks. 😅

packages/venia-ui/lib/components/CreateAccount/createAccount.js
sirugh: > That said, this is a reminder that there is no good reason to have a queries folder in the first place.

I agree. I think we moved them into the `queries/` folder maybe to make it easier on some tooling? I'm not sure.

How about this compromise - we merge this as is and I open an issue to move all the queries/mutations to the component folders where they are used.

sirugh: Made https://github.com/magento/pwa-studio/issues/1928 :D

PR #1899, opened 2019-10-16
[SPIKE] Extension via portals (a different take)

sirugh:
This branches off the initial spike but is a sort of inverse approach in terms of rendering.

While the initial spike attached `div`s with ids and rendered extensions post-mount, this approach allows components to render `Portal` components with identifiers. Client code doesn't change much, they still use the `Extension` component which registers the JSX within a map and when the UI component calls `<Portal id="some-id" />` it just renders whatever is stored in the map for that identifier.

So in the first spike we have the portals/component code living in the tree at the level where the `Extension` is defined:

[Image from Gyazo](https://gyazo.com/d1f6250d9d799dfab469ed84c6040e58)

In this spike, the components are more representative of the DOM:

[Image from Gyazo](https://gyazo.com/c7063449147aa40f5b568da13b4c455f)

### DIFF
--- a/packages/venia-concept/src/index.js
+++ b/packages/venia-concept/src/index.js
@@ -10,6 +10,10 @@ import app from '@magento/peregrine/lib/store/actions/app';
 import App, { AppContextProvider } from '@magento/venia-ui/lib/components/App';
 import './index.css';
 
+import { Extension } from '@magento/venia-ui/lib/components/Extension';
+import { WelcomeToast } from './extensions/WelcomeToast/WelcomeToast';
+import { ProductRecommendations } from './extensions/ProductRecommendations/ProductRecommendations';
+
 const { BrowserPersistence } = Util;
 const apiBase = new URL('/graphql', location.origin).toString();
 
@@ -46,6 +50,10 @@ ReactDOM.render(
     <Adapter apiBase={apiBase} apollo={{ link: apolloLink }} store={store}>
         <AppContextProvider>
             <App />
+            <Extension targetId="extension-point-1">
+                <WelcomeToast />
+                <ProductRecommendations />
+            </Extension>
         </AppContextProvider>
     </Adapter>,
     document.getElementById('root')
--- a/packages/venia-concept/src/index.js
+++ b/packages/venia-concept/src/index.js
@@ -50,9 +50,13 @@ ReactDOM.render(
     <Adapter apiBase={apiBase} apollo={{ link: apolloLink }} store={store}>
         <AppContextProvider>
             <App />
+            {/* Some extensions require a target */}
             <Extension targetId="extension-point-1">
-                <WelcomeToast />
                 <ProductRecommendations />
+            </Extension>
+            {/* Other extensions don't need target (assumes #root at least) */}
+            <Extension>
+                <WelcomeToast />
             </Extension>
         </AppContextProvider>
     </Adapter>,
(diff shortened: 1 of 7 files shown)


packages/venia-ui/lib/components/Extension/extension.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param {String} [props.targetId=root] id of injection target for extension
+ * @param {ReactNode} props.children
+ */
+export const Extension = props => {
+    const { children, targetId = 'root' } = props;
+
+    // Add new extensions.
+    const existing = extensionMap.get(targetId) || [];
+    existing.push(children);
```
sirugh: How safe is this?

revanth0212: Safe from a security standpoint?

I feel yes, it bears the same security risks as the previous implementation.

revanth0212:
I like this implementation of extensions for its way of handling Extension definitions by having them all defined and registered first (`<Extension>`) and then rendering them (`<Portal>`). This will make sure the Extensions render after the App. Sweet 👍 

On the other hand, I have a couple of concerns:

1. You can not have granular control over each extension. You can only have control over the `targetId`. A developer can not control the rendering of each extension. If the code comes across `<Portal id="someID" />`, it will render all the extensions in the map for that ID.

2. It is not a Portal any more since it is rendering extensions on the spot. We might as well call it `<RenderExtensions id="someID" />`. We either have to change the title or implementation.

-----

I would prefer an API where the developer does not have to go into the code deep down to render an extension. All he needs to know is the `key or address` of the DOM element where he can render his extension and at the same time have granular control over each extension.

------

I hope I am making some valid points here. This is based on my understanding of how UI extensions work in Magento.

packages/venia-concept/src/index.js
```diff
@@ -45,7 +49,28 @@ const apolloLink = ApolloLink.from([
...
+                </div>
+            </Extension>
+            <Extension targetId="main-ep-after-children">
+                <ProductRecommendations />
+            </Extension>
+            {/* Other extensions don't need target (assumes #root at least) */}
+            <Extension>
+                <WelcomeToast />
+            </Extension>
+            {/* App/Portals must be rendered _after_ extensions. */}
```
supernova-at: Out of curiosity, why?

And I'm guessing none of the extensions above could have `targetId='root'`?

This begs the question of whether or not extensions will depend on the order of components loading. If so, we'll likely want some code that identifies the cases that won't work and warn about them in development mode. Otherwise we're signing everybody up for lots of debugging nightmares.

packages/venia-ui/lib/components/Extension/extension.js
sirugh: What content are you referring to? Can you give me an example?

jimbo: The children of `Extension`, which could be dynamic.

```jsx
const Dynamic = props => {
  const [clicked, setClicked] = useState(false)
  const text = clicked ? "yep" : "nope"

  const handleClick = useCallback(() => {
      setClicked(true)
  }, [setClicked])

  return (
    <div>
        <button onClick={handleClick}>
            click me
        </button>
        <Extension targetId="extension-point-1">
            {text}
        </Extension>
    </div>
  )
}
```

sirugh: In this PR we would only register extensions at the top level, at least in the examples I had so far. App/Portal mounting iterates through the extensions registered and then renders them, so it is entirely possible that if a component itself, within App, registers an `Extension` the portal for it may have already rendered.

Still I think to your question, if you had some Extension that was the output of some changing thing, would it update? I think so, as you see with the toast that uses user context to display if a user signs in after the fact.

cherdman:
@revanth0212 and @zetlen , I am closing out the stale spike.

PR #1912, opened 2019-10-18
#1911: fix Button is exported twice

### LINKED ISSUE #1911: [bug]: Button component is exported twice
**Describe the bug**
The `Button` component is exported twice.

**Expected behavior**
The named component is exported only as the default export.

**Additional context**
I don't think this actually matters for execution, I just haven't seen this done anywhere else.

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

brendanfalkowski:
## Description

The `Button` component is exported from the function and as a default export below. This doesn't seem to break anything, but I haven't seen it anywhere else so it's a little weird.

## Related Issue

Closes #1911.

### Verification Steps

1. Build should pass.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/Button/button.js
+++ b/packages/venia-ui/lib/components/Button/button.js
@@ -16,7 +16,7 @@ const getRootClassName = priority => `root_${priority}Priority`;
  *
  * @returns {React.Element} A React component that displays a single button.
  */
-export const Button = props => {
+const Button = props => {
     const {
         children,
         classes: propClasses,


packages/venia-ui/lib/components/Button/button.js
```diff
@@ -16,7 +16,7 @@ const getRootClassName = priority => `root_${priority}Priority`;
  *
  * @returns {React.Element} A React component that displays a single button.
  */
-export const Button = props => {
```
sirugh: Be careful here -- this was a named export vs the default below so _technically_ this is different.

```
import { Button } from './button'; // named
import  Button from './button'; //default
```

brendanfalkowski: Yeah, that's understandable (re: breaking change). I checked the repo and didn't see this named import being used anywhere, so it seemed ok.

This was also the only instance of something like `export const ComponentName` in a file where only the default export existed otherwise. There were other cases with multiple named exports but most of them didn't have a default export.

PR #1918, opened 2019-10-19
1917 avoid js string on text nodes in jsx

### LINKED ISSUE #1917: [bug]: avoid JS-wrapped text nodes in JSX
**Describe the bug**
In JSX, replaces simple text nodes like:
`<div>{'The text node'}</div>`

With the simpler:
`<div>The text node</div>`

The extra verbosity to treat a text node as a JS string seems like clutter.

Note: no idea if this does something for somebody, but I'm not a fan of the verbosity and it seems to be the exception. I heard somewhere that VS Code may auto-insert this but IDK.

**Expected behavior**
Plain text nodes don't need to be explicitly wrapped to parse as a JS string.

**Additional context**
I didn't make any changes to JS-wrapped text for attributes, though I think these are overly verbose/weird too compared to a double-quote that's normal in HTML.

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

brendanfalkowski:
## Description

In JSX, some plain text nodes in components are wrapped in JS curlies to parse a string like `{'thing'}`.

This looks like clutter, and I'm not sure if it serves a purpose.

Note: I did not touch attribute values using this technique because I'm guessing React escapes that.

## Related Issue

Closes #1917

### Verification Steps

Should be none. All strings will render as strings without being parsed as JS.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/RootComponents/Category/categoryContent.js
+++ b/packages/venia-ui/lib/RootComponents/Category/categoryContent.js
@@ -35,7 +35,7 @@ const CategoryContent = props => {
                 onClick={handleOpenFilters}
                 type="button"
             >
-                {'Filter'}
+                Filter
             </button>
         </div>
     ) : null;
--- a/packages/venia-ui/lib/components/AuthBar/authBar.js
+++ b/packages/venia-ui/lib/components/AuthBar/authBar.js
@@ -26,7 +26,7 @@ const AuthBar = props => {
             priority="high"
             onClick={handleSignIn}
         >
-            {'Sign In'}
+            Sign In
         </Button>
     );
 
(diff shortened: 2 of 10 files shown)


sirugh:
This was something I saw done in the repo before so I took it as the standard. I assume it'll make replacement with translated strings/variables easier in the future to just keep the curlies but I'll let @jimbo speak to this practice.

jimbo:
@brendanfalkowski Thanks for submitting a PR to go with the issue you created. 😄 

Unfortunately, this is intentional and just not covered by our lint rules (yet), so I'm going to close it for now. I've left a comment with more details on #1917; we can use that issue to discuss and figure out what else we should do on this front.

PR #1929, opened 2019-10-22
Fix hook dependencies for app toast error handling

### LINKED ISSUE #1377: [bug]: Fix useEffect dependency warnings in toasts
**Describe the bug**
After adding the eslint plugin for hooks we identified that there is a potential to introduce bugs around some apis such as `addToast`.

Currently, a component may import and use `addToast` within a `useEffect` hook like so:

```js
  const [, { addToast } ] = useToasts();
  useEffect(() => {
    addToast({
      // some locally scoped toast props
    })
  }, []); // warning! `addToast` should be added as a dependency!
  return (<div></div>);
}
```

The above code would throw an eslint warning because `addToast` should be provided as a dependency to `useEffect`. However, if you add the `addToast` function as a dependency the effect will actually re-render endlessly adding toast after toast after toast. I believe this happens because `addToast` is re-created, but I'm not positive -- you'll need to look into this.

**To Reproduce**
1. Add `addToast` and `toasts` as dependencies to the `useEffect` [here](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/App/app.js#L72).
2. Throw an error in `miniCart.js#render`.
3. Try to close the error toast. It should keep reappearing.
4. Also add `addToast` to the dependency array in the toasts storybook file. It should endlessly loop.

**Expected behavior**
No lint errors, and the app should behave "correctly" as in when you close the toast it should invoke the `onDismiss` and it definitely should not infinitely render.

**Possible solutions**
https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies
https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often
https://github.com/facebook/create-react-app/issues/6880

 - Browser Version [e.g. 22]:
 - Magento Version:
 - PWA Studio Version:
 - NPM version `npm -v`:
 - Node Version `node -v`:

- [ ] `venia-concept`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
## Description

Back when we added toasts there was a bug where we would recursively add toasts when the error handlers in app were invoked. There was an effect that was being re-run based on the dependencies. Instead of fixing it at the time, we papered over it by removing the problematic dependency and adding a `disable-line` rule.

Funny enough, after the Talon refactor it seems like we can fix the dependency arrays without having to do much else. We can even remove the extra check within the `handleError`.

## Related Issue

Closes #1377 .

### Verification Steps

1. Add an error to `miniCart.js` before the return and load the page. You should see the error toast.
2. Click the 'x' to clear the toast. It should not be replaced by a duplicate.
3. Refresh the page and this time let the error clear itself. It should disappear after some timeout and it should not be replaced by a duplicate.

## Checklist


packages/venia-ui/lib/components/App/app.js
```diff
@@ -61,15 +61,10 @@ const App = props => {
                 timeout: 15000,
                 type: 'error'
             };
-            // Only add a toast for new errors. Without this condition we would
-            // re-add toasts when one error is removed even if there were two
-            // added at the same time.
-            const errorToastId = getToastId(errorToastProps);
```
sirugh: I wrote this function specifically because of this bug. Now that the bug is no longer present we don't really have a need to export it aside from it just being part of the component API.

PR #1935, opened 2019-10-24
PDP Cache Hits

### LINKED ISSUE #1661: Improve Cache Hits for PDPs
Apollo's [InMemoryCache](https://www.apollographql.com/docs/react/v2.5/advanced/caching/#inmemorycache) is a normalized data store that serves as our Apollo cache.

It has a [configuration option](https://www.apollographql.com/docs/react/v2.5/advanced/caching/#configuration) called `cacheRedirects` which gives us an opportunity to serve PDP data from the cache instead of having to re-fetch product data.

Here is a [guide](https://www.apollographql.com/docs/react/v2.5/advanced/caching/#cache-redirects-with-cacheredirects) from Apollo on how to accomplish this.

The tradeoff is that the initial "category list" query that displays products must include all the details the PDP needs. This ticket should leverage `fragments` as a way to ensure those fields stay in sync, otherwise Apollo won't find it in the cache.

The tradeoff is between requesting too much data on the category pages (PDP data needs for every product) vs the server communication overhead required for a single PDP (additional request, response parse, latency, loading, etc.).

📝 This work is closely related to https://github.com/magento/pwa-studio/issues/1666.

supernova-at:
image

# Description

This PR manually instructs Apollo how to store and where to look in its own cache for product entries we may already have from previous queries (such as [getCategory.graphql](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/queries/getCategory.graphql)).

It:

## Updates the cache key used to store products to use a product's `url_key` instead of its `id`

Example previous key: `ConfigurableProduct:1143`.
Example new key: `ProductInterface:claudia-crochet-dress`.

This is needed because when we land on a product detail page (PDP), the only information we know about the product is its `url_key`, parsed directly from the URL.

Apollo was never able to find products in the cache when they were stored by `id` because it didn't know that `id` until after a roundtrip to the server to fetch it (via [getProductDetail.graphql](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/queries/getProductDetail.graphql)). It's obviously too late at that point, we already made the page wait for the server response.

By storing products by their `url_key`, the PDP can find product entries in the cache using the only piece of information it has: the `url_key`. If it finds a matching product in the cache, it returns it immediately and users no longer have to wait for the server response 🎉 .

I should note that we do still kick off a network request to fetch that product, and the cache entry gets updated when the server responds. This mimics the `stale-while-revalidate` caching strategy.

## Introduces our first GraphQL `fragment`: `productDetails.graphql`

A requirement for being able to use a cached product for display on the PDP is that the cached product must have all the fields required by the PDP. (There is such a thing as returning / using partial results, but I deemed that out of scope here.)

Therefore, this PR updates the [getCategory.graphql](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/queries/getCategory.graphql) query to ensure that its `products` query requests all the fields the [getProductDetail.graphql](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/queries/getProductDetail.graphql) query does.

A best practice is to use reusable [fragments](https://www.apollographql.com/docs/react/data/fragments/) to keep these things in sync.

Unfortunately this fragment can't be used by queries yet due to https://github.com/magento/graphql-ce/issues/1027 and https://github.com/magento/graphql-ce/issues/147. I have copied this fragment's contents (along with a comment explaining as much) into the `getCategory` and `getProductDetail` queries for now.

But it can be used by the Apollo client itself when we call `readFragment`.

## Does not use `cacheRedirects`

The best practice / simplest way to accomplish this task would have been to use Apollo's [cacheRedirects](https://www.apollographql.com/docs/react/caching/cache-interaction/#cache-redirects-with-cacheredirects) to tell Apollo's `InMemoryCache` where to look in the cache before issuing a network request.

Unfortunately, a Magento GraphQL `products` query returns a `Products` type.
What we have in the cache are concrete implementations (such as `SimpleProduct`) of the `ProductInterface` type. So even if we know exactly which product that `products` query is going to return, we can't point Apollo to it because the shape of the return object doesn't match.

```graphql
type Products {
  filters: [LayerFilter]
  items: [ProductInterface] # <-- We only have these guys
  page_info: SearchResultPageInfo
  sort_fields: SortFields
  total_count: Int
}
```

See https://github.com/magento/graphql-ce/issues/86#issuecomment-545935174 for more details.

The code in this PR can mostly disappear once Magento GraphQL supports individual `product` queries - we would just use `cacheRedirects` instead.

## Related Issue

Closes #1661 .

#### Product Already In Cache

1. Go to the `/venia-dresses.html` page
2. Chrome Dev Tools > Application > Clear Storage > Clear Site Data
3. Empty Cache and Hard Reload
4. Chrome Dev Tools > Application > Local Storage > (your site) > `apollo-cache-persist`
5. Verify that there is an entry for `ProductInterface:claudia-crochet-dress` in the cache
6. Chrome Dev Tools > Performance > ☑️ Screenshots, then ⚫️ Record
7. Click on the Claudia Crochet Dress product to go to its PDP
8. Verify that the PDP loaded before the `productDetail` GraphQL query returned by comparing the screenshots with the Network section:

Screen Shot 2019-10-24 at 1 05 54 PM

Note that there are two GraphQL queries sent by the PDP: `urlResolver` and `productDetail`. We must wait for the `urlResolver` one, but this PR skips waiting for the `productDetail` one.

#### Product Not In Cache / Direct Entry to PDP

0. Navigate to `/claudia-crochet-dress.html` if you're not already there
1. Chrome Dev Tools > Application > Clear Storage > Clear Site Data
2. Empty Cache and Hard Reload
3. Verify that the page loads as normal

## Checklist

## Questions

* Does using the `url_key` introduce any attack vectors into `localStorage`?


packages/venia-ui/lib/queries/getCategory.graphql
```diff
@@ -29,17 +29,64 @@ query category(
             }
         }
         items {
+            # Once graphql-ce/1027 is resolved, use a ProductDetails fragment here instead.
```
sirugh: I have concerns about requesting the entire product details fragment here in this request. This would be a form of "pre-fetching" wherein we would fetch _all_ the product details for the visible products on the category page before a user has navigated to the products. I don't think we should do this unless we can make a strong argument for it.

I'll detail the pros/cons. Please feel free to add!

Pros:

* Any product a user clicks on from the category view will have been cached so it should render faster.
* This query is more composable in that `items` will just use the Fragment. 

Cons:

* More data is fetched on the category view so the category view may render more slowly.
* Extra data is fetched that isn't ever used on the category page.
* If we remove something the category requires for rendering from the pdp fragment we may introduce a bug without realizing it.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -0,0 +1,111 @@
...
+            });
+        } catch (e) {
+            // The product is in the cache but it is missing some fields the fragment needs.
+            // We don't have to do anything here.
+            return null;
+        }
+    }, [apolloClient, fragment, fragmentVariables, urlKey]);
+
+    const product = useMemo(() => {
+        const mapProduct = product => {
```
sirugh: This can/should exist out of the product talon so we aren't recreating it (even though it's semi protected by memoization).

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -0,0 +1,111 @@
...
+         *
+         * If the object with the specified `id` is not in the cache, we get `null`.
+         * If the object is in the cache but doesn't have all the fields we need, an error is thrown.
+         *
+         * @see https://www.apollographql.com/docs/react/caching/cache-interaction/#readfragment.
+         */
+        try {
+            return apolloClient.readFragment({
+                // This `id` must match the result of `cacheKeyFromType` in `venia-ui/lib/util/apolloCache.js`.
+                id: `ProductInterface:${urlKey}`,
```
sirugh: This is coupled tightly to that utility then. I would expect that such information is provided to the talon as an argument (like the query) or that peregrine owns the utility. In this case I'm not sure where it should go.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -0,0 +1,111 @@
...
+
+    const [runQuery, { loading, error, data }] = useLazyQuery(query, {
+        // We're manually checking the cache here, so useLazyQuery can skip that.
+        fetchPolicy: 'network-only',
+        variables: queryVariables
+    });
+
+    const apolloClient = useApolloClient();
+
+    const productFromCache = useMemo(() => {
```
sirugh: After checking out the docs I'm not sure we need to do this fetch ourselves. Apollo has a [cache-and-network](https://www.apollographql.com/docs/react/api/react-apollo/) `fetch policy` which basically does the same thing that we have hand written here:

>This fetch policy will have Apollo first trying to read data from your cache. If all the data needed to fulfill your query is in the cache then that data will be returned. However, regardless of whether or not the full data is in your cache this `fetchPolicy` will always execute query with the network interface unlike `cache-first` which will only execute your query if the query data is not in your cache. This fetch policy optimizes for users getting a quick response while also trying to keep cached data consistent with your server data at the cost of extra network requests.

Can we just use the `cache-and-network`  policy in lieu of all this code?

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -0,0 +1,111 @@
...
+export const useProduct = props => {
+    const {
+        fragment,
+        query,
+        urlKey // TODO: move `getUrlKey` into Peregrine?
+    } = props;
+
+    const onServer = false;
+
+    const fragmentVariables = useMemo(() => {
```
sirugh: IMO this is overuse of `useMemo` here and for `queryVariables` below. We should save it for when we have some expensive recalculation that could occur on every render. I'd just move these inline to where they are used. I don't think they gain us much by pre-defining, even outside the talon.

packages/venia-ui/lib/RootComponents/Product/product.js
```diff
@@ -14,49 +15,35 @@ import getUrlKey from '../../util/getUrlKey';
...
-    }
-
-    // Note: STORE_NAME is injected by Webpack at build time.
-    return (
-        <Fragment>
-            <Title>{`${product.name} - ${STORE_NAME}`}</Title>
-            <ProductFullDetail product={product} />
-        </Fragment>
-    );
+    return <ErrorView outOfStock={true} />;
```
sirugh: I think the last `return` should be the "happy path" render. In this case we could leave it as it was right?

packages/venia-ui/lib/queries/getCategory.graphql
revanth0212: Regarding cons 1 and 2, we can have the SW do this for us. It can request in the back, so no lag/load on the main thread and we can write logic into it to only request if the client is running on WiFi.

supernova-at: Thanks for bringing this up! Yes, this is the tradeoff.

I made some inline edits to your pros & cons 😄 , we'll definitely be over-fetching on the Category page in order to make the PDP load faster.

Fortunately, this is gzipped JSON (string data) so - at least currently - we're talking about adding an extra ~2.5KB. The last `graphql` request shown is the one that requests category data:

This PR:

Screen Shot 2019-10-25 at 8 16 22 AM

Develop (upward-js):

Screen Shot 2019-10-25 at 8 17 07 AM

To your point though, the timing does get blown out a bit. I'd say that one is my local machine vs. a deployed instance but all the other requests took about the same time so this may in fact be an issue.

> * If we remove something the category requires for rendering from the pdp fragment we may introduce a bug without realizing it.

This is a good point and yeah we'll have to keep in mind that this fragment is definitely not PDP-specific.

supernova-at: There is a "middle ground" that I should point out too: the idea of "partial data". From [the useQuery docs](https://www.apollographql.com/docs/react/data/queries/#usequery-api):

> returnPartialData: Opt into receiving partial results from the cache for queries that are not fully satisfied by the cache. false by default.

Because this is `false` currently, we have to ensure that all the fields of a product are present in the cache for a cache hit to take place and avoid going to the network. This is why the category query adds all of the product fields.

If we change this to `true`, we could start rendering the PDP with whatever data we have in the cache for that product while we query the network for the rest. Currently the Category page requests `name`, `price`, image, etc. This will lead to the PDP initially rendering with only that data. When the server response comes back, then we'd fill in the rest of the details.

This would obviously take some refactoring of the PDP to accomplish as well. Mainly, it wouldn't be able to trust that all of its data will be present.

packages/venia-ui/lib/util/apolloCache.js
```diff
@@ -0,0 +1,187 @@
...
+        // Store all implementations of ProductInterface with the same prefix,
+        // and because we can't filter / query by id, use their url_key.
+        case MagentoGraphQLTypes.BundleProduct:
+        case MagentoGraphQLTypes.ConfigurableProduct:
+        case MagentoGraphQLTypes.DownloadableProduct:
+        case MagentoGraphQLTypes.GiftCardProduct:
+        case MagentoGraphQLTypes.GroupedProduct:
+        case MagentoGraphQLTypes.SimpleProduct:
+        case MagentoGraphQLTypes.VirtualProduct:
+            return `${MagentoGraphQLTypes.ProductInterface}:${object.url_key}`;
```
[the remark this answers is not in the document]
supernova-at: Good question!

Remember that the PDP only has a `url_key` to use to find its product. Without grouping these, entries in the cache would look like:

`SimpleProduct:augusta-earrings`
`ConfigurableProduct:claudia-crochet-dress`

Even though the PDP knows `claudia-crochet-dress`, it doesn't know the product's `type` and would therefore be unable to generate the correct key for the cache.

Grouping them all to:

`ProductInterface:augusta-earrings`
`ProductInterface:claudia-crochet-dress`

makes it so that the PDP doesn't have to know the product type.

It works only because the `url_key`s are unique even across product types. If we had simple products and configurable products with the same `url_key`, they would overwrite each other and this wouldn't work.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
supernova-at: Good point, I'll make this better.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
supernova-at: image

DO YOU KNOW HOW LONG I LOOKED FOR THAT DOCUMENTATION?!!!!

I saw it as an option but I couldn't find exactly what it did - thank you!
I'll see if this works.

packages/venia-ui/lib/RootComponents/Product/product.js
supernova-at: Oh yeah, thanks for pointing this out too. This is a definite departure from how we usually render, but `useLazyQuery` sets `loading` to `true` while the "background refresh" request is in flight.

So even though it had a cached product it could use, it'd show the full page loading indicator until the background refresh was complete, which defeated the entire purpose 😜

sirugh: Could we just change that line to `if (loading && !data) return fullPageLoadingIndicator;`?

packages/venia-ui/lib/queries/getCategory.graphql
jimbo: > If we change this to true, we could start rendering the PDP with whatever data we have in the cache for that product while we query the network for the rest.

This is *exactly* what the upcoming [React Concurrent Mode](https://reactjs.org/docs/concurrent-mode-suspense.html) enables and prescribes. They call this strategy `render-as-you-fetch`, as opposed to `fetch-on-render` or `fetch-then-render`.

> This would obviously take some refactoring of the PDP to accomplish as well. Mainly, it wouldn't be able to trust that all of its data will be present.

And since React will soon provide us with tools to do this canonically, I'll recommend that we hold off on restructuring things until then. Should come together nicely very soon.

jimbo: > I have concerns about requesting the entire product details fragment here in this request.

As you noted, this is just another form of prefetching, and as Revanth & Andy noted, it's a small amount of extra bytes that we can request in the background. Given how slow GraphQL calls are right now, a cache hit when a product is clicked would probably be well worth the potential overfetch.

sirugh: > Given how slow GraphQL response times are right now, a cache hit when a product is clicked would probably be well worth the potential overfetch.

I still have concerns but I guess this would be a pro :)

packages/venia-ui/lib/util/apolloCache.js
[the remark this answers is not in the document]
sirugh: A question was raised this evening and I think it warrants a look: https://magento.slack.com/archives/C7LRPD5FT/p1572299867001800

Basically this caching might not work for multi-store/locale clients since product names are localized.

packages/venia-ui/lib/RootComponents/Product/product.js
zetlen: I demand that you not use ErrorView until it throws errors in developer mode.

packages/venia-ui/lib/util/apolloCache.js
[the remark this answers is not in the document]
zetlen: We can't count on any aspect of the product data except the SKU being the same when you switch stores or locales, believe it or not. So breaking the cache when the url_key changes is actually what we want.

packages/venia-ui/lib/util/apolloCache.js
```diff
@@ -0,0 +1,187 @@
+import { defaultDataIdFromObject } from 'apollo-cache-inmemory';
+
+/**
+ * A non-exhaustive list of Magento Types defined by the GraphQL schema.
```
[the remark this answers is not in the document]
zetlen: We do plan to do this at build-time. You can pick which schema types you want inlined into client config with the tools available (and if you can't, _we'll make them)_.

packages/venia-ui/lib/queries/getCategory.graphql
zetlen: Yeah, I'm in favor of this. I understand and agree with @sirugh 's concerns as things to watch. Thank you (and you Andy) for enumerating them so clearly. Still, this is exactly the right use case for a fragment. Especially because Apollo understands fragments discretely, and can cache them by concrete type. So this should be really good.

Maybe we can get better opt in the future by splitting this fragment into a few smaller fragments that you can optionally compose together, but for now I don't think it's a problem. Partial fetch and React concurrent render should cover for the additional fetching, like @jimbo and @supernova-at say.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
jimbo: Yeah, this should be the same as the service worker's [`stale-while-revalidate`](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#stale-while-revalidate).

packages/venia-ui/lib/RootComponents/Product/product.js
jimbo: Yes, this is also a good time to remind everyone that `ErrorView`

- was designed to handle error types coming from `MagentoRouteHandler` only
- should not have had an `outOfStock` prop added to it
- should not be rendered here

If the custom error view for this component happens to have the same presentation as the one we use for routing failure, its CSS can use `composes` to share styles. 👍

packages/venia-ui/lib/util/apolloCache.js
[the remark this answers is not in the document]
supernova-at: Created JIRA-100 for this.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1935/commits/28de28a26c2ed22153a2c25666b73bcdedc5b9f7.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1935/commits/25a049cf373d4b6689e3030687b65ad6d4f235f7.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
supernova-at: I should have realized this before, but no dice on this one. The whole reason we have to do this PR is because there's no single `product` query in Magento GraphQL - it will never find a cache hit unless we manually override it like this and tell it exactly where to look.

supernova-at: I am simulating `cache-and-network` / `stale-while-revalidate` manually.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1935/commits/cb6e9acf5fac29f6ecebfc34faabc63fc7e7854b.

packages/venia-ui/lib/RootComponents/Product/product.js
supernova-at: Moved happy path to last return for consistency in https://github.com/magento/pwa-studio/pull/1935/commits/55253609a051a9e2e393ff5a94ca3fc489250a4b.

supernova-at: I did not add `ErrorView` here, do y'all want me to refactor it out as part of this PR?

supernova-at: Nevermind, it was easy: https://github.com/magento/pwa-studio/pull/1935/commits/2413375d7c51ac7004a22f699a6c4cfa1a8987e4.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
sirugh: Isn't `productDetails` filtered by `url_key` a "single product query"?

supernova-at: To clarify, I mean a query that returns a `ProductInterface`. A `products` query (we rename it to `productDetails`) returns a `Products` object that nests `ProductInterface`s under `items`.

See https://github.com/magento/graphql-ce/issues/86#issuecomment-545935174 for a more detailed explanation.

I also used different words to describe it in the "does not use `cacheRedirects`" part of this PR description. 👍

jimbo:
@supernova-at Regarding that last commit, see [my comment](https://github.com/magento/pwa-studio/issues/1917#issuecomment-544604537).

supernova-at:
PR Updated:

Ran into an issue where some cache entries had `undefined` in their keys: `ProductInterface:undefined`.

I added a null guard to ensure we have a `url_key` before using it as part of the cache key, and updated a few places where we were making `products` queries but not asking for the `url_key` as part of the response.

Unfortunately we can't do that in `products` queries that ask for `items.variants.product` because all of the variants have the same `url_key` and they'd collide.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -0,0 +1,94 @@
...
+    }, [data, mapProduct, productFromCache]);
+
+    useEffect(() => {
+        // Fetch the product from the network.
+        // Note that this always fires, regardless of whether we have a cached product or not:
+        // If we do not have a cached product, we need to go fetch it from the network.
+        // If we do have a cached product, we want to ensure its cache entry doesn't get stale.
+        // In both cases, Apollo will update the cache with the latest data when this returns.
+        runQuery({
+            variables: {
```
sirugh: You already provided these variables to the initial `useLazyQuery` call. Why do we have to provide them again?

Also I think we can just use `useQuery` with `fetchPolicy: 'cache-and-network'` which is exactly the same as what we're doing here except we don't have to manually invoke the query.

supernova-at: Great suggestion, looks way better now: https://github.com/magento/pwa-studio/pull/1935/commits/0512dacbe0ef01ed1021f60babe987f7fe64d00d.

packages/venia-ui/lib/queries/getNavigationMenu.graphql
```diff
@@ -13,6 +13,7 @@ query navigationMenu($id: Int!) {
                     small_image {
                         url
                     }
+                    url_key
```
sirugh: AFAICT this whole sub-query isn't used anywhere so we can just remove it entirely.

packages/venia-ui/lib/queries/getCategoryList.graphql
```diff
@@ -15,6 +15,7 @@ query categoryList($id: Int!) {
                     small_image {
                         url
                     }
+                    url_key
```
sirugh: I'd rather we not fetch the key here I think. We aren't actually using that piece of data where we use this query and that feels counter to the intent of the granularity of graphql.

packages/venia-ui/lib/queries/getNavigationMenu.graphql
supernova-at: Done: https://github.com/magento/pwa-studio/pull/1935/commits/bf4b0dd741eddc8ed586fffbea6245814fd7b94d

packages/venia-ui/lib/queries/getCategoryList.graphql
supernova-at: https://github.com/magento/pwa-studio/pull/1935/commits/bf4b0dd741eddc8ed586fffbea6245814fd7b94d.

PR #1937, opened 2019-10-27
Replace withRouter with react router hooks

### LINKED ISSUE #1715: [feature]: Make a `withRouter` hook and use it
We use the `withRouter` HOC in several places.

```js
const ComponentThatUsesRouterProps = props => {
  const { history } = props;
  // do something with provided history
};
export default withRouter(ComponentThatUsesRouterProps);
```

If there is a hook equivalent we should use it _or_ we should roll our own.

```js
const ComponentThatUsesRouterProps = props => {
  const { history } = useRouter();
  // do something with history
};
export default ComponentThatUsesRouterProps;
```

The alternative is that we leave it as is but I'd rather not ;)

dani97:
## Description

Replace WithRouter HOC with React router hooks

## Related Issue

Closes #1715 

### Verification Stakeholders

@sirugh 
### Verification Steps

1. Check all affected components.

## Checklist


dani97:
@sirugh  two tests are failing can you help me fix it.

dani97:
> @sirugh two tests are failing can you help me fix it.

Fixed Them  :)

jimbo:
@dani97 We're on the same page! The good news is that I think this PR is right on track and we shouldn't have much trouble getting it merged. The bad news is that I already upgraded `react-router-dom` to `~5.1.0` in #1884 and made some of these same changes, so I kind of stepped on your toes there. Sorry I didn't communicate that change sooner.

I believe there's still additional value in this PR, since you've accounted for the Pagebuilder components where I did not. Let's resolve conflicts here and try to get this one merged in. 👍

dani97:
@jimbo  Resolved conflicts  👍

packages/venia-ui/lib/drivers/index.js
```diff
@@ -1,4 +1,14 @@
...
+export {
+    Link,
+    Redirect,
+    Route,
+    Switch,
+    withRouter,
+    useLocation,
+    useHistory,
+    useParams,
+    useRouteMatch
```
sirugh: I don't think we need to define any of the hooks as `drivers` and we can probably remove `withRouter` too.

packages/venia-concept/src/drivers.js
```diff
@@ -4,7 +4,11 @@ export {
     Redirect,
     Route,
     Switch,
-    withRouter
+    withRouter,
+    useHistory,
+    useLocation,
+    useRouteMatch,
+    useParams
```
sirugh: Some comment here -- can remove these hooks and `withRouter`.

PR #1905, opened 2019-10-18
HTML Caching and SW Communication API Implementation.

revanth0212:
## Description

HTML Caching and Service Worker Communication API Implementation. The service worker will be following the `StaleWhileRevalidate` strategy for all HTML files. It will return the readily available file from the cache and try to update the file in the background. If the file happens to be different from the one already in the cache, it will send a message to the UI about an update available and a button in the message to refresh the page.

## Related Issue
Closes #1772 #1781 

### Verification Steps
1. Build and start the server using `yarn build && yarn stage:venia`.
2. Refresh the page once after it has loaded up. This will make sure that the Service Worker is installed successfully.
3. Open `venia-concept/dist/index.html` and make a change like adding a new script tag
`<script>console.log('Hey this is a new HTML')</script>`
4. Now navigate to a different page like Category or Product by clicking on the appropriate links/pictures. Note: Do not refresh the page.
5. Once you navigate to a new page, a toast should show up in the bottom letting you know that an update is available.

## Screenshots / Screen Captures (if appropriate)
HTML Update Toast Demo

## Checklist
* No functional change. Make sure both `yarn stage:venia` and `yarn watch:venia` should work like previous.
* Should be merged only after #1874 is merged.

PR #1938, opened 2019-10-28
Product Images Pre Fetching

### LINKED ISSUE #1766: [feature]: Implement image caching at Service Worker Level.
Related to #1673 

Always cache the best resolution version of an image.

1. If a higher resolution version is requested, fetch the higher resolution image, bust the cache, insert this image and return the image to the client. If fetcher failed, return the available image to the client. If the fetch is taking longer than a THRESHOLD value, return the available image and replace that with the fetch's result when resolved.
2. If a lower resolution image is requested, return the available image and do not make any network call.

- [x] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

### LINKED ISSUE #1769: [feature]: Implement product image pre fetching.
Related to #1673 

As of today, we do not prefetch images on the products page. For instance if the user is looking at image number X, we can make the service worker request images X-1 and X+1 for a smoother and faster experience. This can be achieved by the react component as well. We can make the call about where to implement this logic.

This might be a far fetch but we can prefetch the corresponding image of a different color. For instance if the user is viewing the green version of image X, prefetch the other versions of X as well. The reason I call this a far fetch is that we will be prefetching `number_of_colors * 3` every time (X-1, X, X+1). Again there will be incremental benefits to this but the initial investment is high due to too many prefetches.

Ideally, this only needs to be implemented on devices connected to WiFi.

Implementing this logic in https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/ProductImageCarousel/carousel.js

- [x] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

revanth0212:
## Description

Implemented Pre Fetching and Caching product images in the background by the SW.

This change speeds up the product page image retrieval by 7000% (500ms -> 2-7ms) on average by delegating the image fetch functionality to the Service Worker beforehand so the image is available when the component needs it. Also, this functionality is only implemented only when the network type is `4g`.

Note: Ideally I would only do this if the type is `WiFi` but the API for network details is not out yet. [It is still a draft.](https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API). Once it is out, we can change the `isFastNetwork` function of `networkUtils` to only work if `WiFi`.

## Related Issue
Closes #1769.
Closes #1766.

### Verification Steps
1. Build and Run the server using `yarn build && yarn stage:venia` command.
2. Go to the server URL in the browser.
3. Open the dev tools and go to the `Service Workers` sub-tab in the `Application` tab. Make sure the Service Worker status is `Activated and Running` if not refresh the page again.
4. Open any category and click on any product that has multiple images.
5. Open the `Network` tab in the dev tools and filter using `Img`. Clear the list for ease.
6. Click on any image thumbnail other than the first image. You should see that the time it takes to get the image should be significantly less than what the same steps in the develop branch take.

1. To verify Offline Capabilities, do steps 2 to 5 again in a new incognito tab.
2. Make the tab offline by changing from `Online` to `Offline` in the network tab.
3. Click on any image thumbnail other than the first image. Images should load unlike `develop` where they fail since network is not available.

Speed Without Changes:

Image Pre Fetching DEMO witPre Fetching DEMO without Changes

Speed With Changes:

Image Pre Fetching DEMO with Changes

Offline Experience Without Changes:

Image Pre Fetching Offline Demo without Changes

Offline Experience With Changes:

Image Pre Fetching Offline Demo with Changes

## Checklist

* Verify that nothing has changed in the Development mode. These changes are for only `stage` mode.
* Merge only after #1871 and #1905 are merged.


zetlen:
This is _extremely_ promising, but it looks like it'll conflict a lot with #1871 . Have you diffed those two branches to see how much rework it would require?

revanth0212:
> This is _extremely_ promising, but it looks like it'll conflict a lot with #1871 . Have you diffed those two branches to see how much rework it would require?

Yeah it was based out of #1871. Updated the branch just now. Should only show the changes.

package.json
```diff
@@ -34,9 +34,9 @@
     "stage:venia": "yarn venia run start",
     "stats:venia": "yarn venia run build:analyze && yarn venia run stats",
     "storybook:venia": "yarn venia run storybook",
-    "test": "jest",
-    "test:ci": "jest --no-cache -i --json --outputFile=test-results.json",
-    "test:debug": "node --inspect-brk node_modules/.bin/jest -i --no-cache",
+    "test": "node --experimental-worker node_modules/.bin/jest",
```
revanth0212: `--experimental-worker` is needed because jest runs on `node` and `MessageChannel` is only available if `--experimental-worker` flag is passed while running jest.

jest.config.js
```diff
@@ -187,7 +191,8 @@ const testVenia = inPackage => ({
                 ]
             }
         },
-        STORE_NAME: 'Venia'
+        STORE_NAME: 'Venia',
+        MessageChannel
```
revanth0212: `MessageChannel` by default is not available on node.

packages/venia-ui/lib/components/ProductImageCarousel/carousel.js
```diff
@@ -35,7 +35,8 @@ const ProductImageCarousel = props => {
     const { images } = props;
 
     const talonProps = useProductImageCarousel({
-        images
+        images,
+        containerWidth: DEFAULT_IMAGE_WIDTH
```
supernova-at: I have concerns about `containerWidth` - what if there are other elements in the container?

packages/venia-concept/src/ServiceWorker/Utilities/networkUtils.js
```diff
@@ -0,0 +1,12 @@
...
+ * is fast or not. It returns true if the connection is 4g which
+ * is same for 4g, LTE and WiFi.
+ *
+ * The new navigator API will have a more granular distribution
+ * for different types of connections once it is published.
+ *
+ * @returns {boolean}
+ */
+export const isFastNetwork = () =>
+    navigator.connection && navigator.connection.effectiveType === '4g';
```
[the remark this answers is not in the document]
revanth0212: Unfortunately not as of today.

packages/venia-concept/src/ServiceWorker/registerMessageHandlers.js
```diff
@@ -0,0 +1,5 @@
+import { registerImagePreFetchHandler } from './Utilities/imageCacheHandler';
+
+export default () => {
+    registerImagePreFetchHandler();
+};
```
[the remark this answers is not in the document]
revanth0212: They can, but I just wanted to consolidate all handler registrations in a single place for maintenance purposes. I am cool with what the team thinks should be the standard.

packages/peregrine/lib/talons/ProductImageCarousel/useProductImageCarousel.js
```diff
@@ -19,6 +24,22 @@ export const useProductImageCarousel = props => {
...
+    useEffect(() => {
+        const urls = images.map(
+            ({ file }) =>
+                `${location.origin}${generateURLFromContainerWidth(
+                    file,
+                    type,
+                    containerWidth
+                )}`
+        );
+        sendMessageToSW(PREFETCH_IMAGES, {
```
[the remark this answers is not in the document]
revanth0212: Yes. If the download is successful we'll have the images ready for future use, if not, it will just continue to be as it is today, which is to fetch when requested.

packages/venia-ui/lib/util/images.js
```diff
@@ -17,23 +17,48 @@ export const imageWidths = {
...
         type: mediaBase,
         width,
         height
     });
 
-    return `${url} ${width}w`;
+export const generateURLFromContainerWidth = (
+    imageURL,
+    type = 'image-product',
+    containerWidth
```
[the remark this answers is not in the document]
revanth0212: > Consider making the signature of this function take an "options" object with each of these properties to improve the readability of the call site and so that callers don't have to care about what order they're in.

Makes sense. Will do.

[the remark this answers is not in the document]
revanth0212: > How do images know their `containerWidth`? Are we expecting them to do DOM calculations?

No, this is expected to be present with the component. For instance, the product details page already knows where the image will be placed and how big the image container will be. 

Come to think of it, I am doubting if the component will know the width in all cases.

I am thinking of DOM calculations but I am sure there might be a better way to do this that I am missing. Any idea how should the component get the width if not provided?

packages/venia-concept/src/ServiceWorker/Utilities/imageCacheHandler.js
```diff
@@ -50,13 +73,84 @@ export const findSameOrLargerImage = async (url, request) => {
...
+                event.ports[0].postMessage({
+                    status: 'error',
+                    message: JSON.stringify(err)
+                });
+                return null;
+            });
+    } else {
+        event.ports[0].postMessage({
+            status: 'error',
+            message: `Slow Network detected. Not pre-fetching images. ${
```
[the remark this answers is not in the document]
revanth0212: You are absolutely right. Slow networks benefit from this but there is a cost barrier to this. That is actually the reason why I created the utils folder so we can keep adding such functions and simply change the `if` case to cache only if `download on mobile networks` is set to true. Assuming we will have a settings page in the future. Again it is debatable and scalable.

Good point though.

packages/venia-concept/src/ServiceWorker/registerMessageHandlers.js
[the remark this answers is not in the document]
supernova-at: That makes sense 👍

packages/venia-ui/lib/components/ProductImageCarousel/carousel.js
supernova-at: In this particular example we're setting `containerWidth` to the image's "default" width. I suppose I was thinking of it like

```html
<div id="container">
  <img />
</div>
```

If there are other things in `#container`, then `containerWidth` is not going to match `DEFAULT_IMAGE_WIDTH`.

Am I thinking about `containerWidth` incorrectly?

packages/venia-concept/src/ServiceWorker/Utilities/messageHandler.js
```diff
@@ -109,7 +109,11 @@ export const sendMessageToClient = (client, type, payload) =>
             }
         };
 
-        client.postMessage({ type, payload }, [channel.port2]);
+        if (client) {
+            client.postMessage({ type, payload }, [channel.port2]);
```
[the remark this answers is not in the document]
supernova-at: Could we do something like `const portToClient = channel.port2`  and then just use `portToClient` from there on out?

Not a huge deal, I'm just thinking this is an easy place for bugs to creep in.

supernova-at:
New changes look good!

Just two "architectural" type questions about `containerWidth` and then this is 👍 .

Looping @sirugh and @jimbo in on https://github.com/magento/pwa-studio/pull/1938#discussion_r342652832 and https://github.com/magento/pwa-studio/pull/1938#discussion_r342618476

packages/venia-ui/lib/components/ProductImageCarousel/carousel.js
revanth0212: My bad, I meant the image container width, the width of the `` tag. I will be using this width to calculate the probable `srcset` URL that might be downloaded in the future.

packages/venia-concept/src/ServiceWorker/Utilities/messageHandler.js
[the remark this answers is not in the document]
revanth0212: We only use `channel.port2` to send it over a message. If we have to rename it, it will only be for a single-use. Creating a variable for a single-use is usually discouraged.

packages/venia-ui/lib/components/ProductImageCarousel/carousel.js
supernova-at: Oh dang, ok - can we call it `imageWidth` or something then?

packages/venia-concept/src/ServiceWorker/registerMessageHandlers.js
```diff
@@ -0,0 +1,5 @@
+import { registerImagePreFetchHandler } from './Utilities/imageCacheHandler';
```
jimbo: You should be able delete everything after this line.

```suggestion
export { registerImagePreFetchHandler as default } from './Utilities/imageCacheHandler';
```

packages/venia-concept/src/ServiceWorker/setupWorkbox.js
```diff
@@ -1,10 +1,28 @@
...
+     * browser handle delete and update of the service worker.
+     */
     workbox.core.skipWaiting();
 
+    /**
+     * This will claim/control all clients once the service worker is
+     * active. If this is not desired, remove this line and the
+     * browser will let the service worker control the clients
+     * after a page reresh.
+     */
```
jimbo: Good docs. 💯

packages/venia-ui/lib/components/ProductImageCarousel/carousel.js
jimbo: Agree with @supernova-at here. Consider `imageWidth` or `defaultImageWidth`.

packages/venia-ui/lib/util/images.js
```diff
@@ -17,23 +17,48 @@ export const imageWidths = {
     XXXLARGE: 2560
 };
 
-const generateURL = (imageURL, mediaBase) => (width, height) => {
-    const url = resourceUrl(imageURL, {
+const generateURL = (imageURL, mediaBase) => (width, height) =>
+    resourceUrl(imageURL, {
```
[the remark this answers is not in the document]
jimbo: Good suggestion.

On another note, do these functions return a full URL or just a pathname? I noticed that after you call them in `useProductImageCarousel`, you're prepending `origin` to their return value. If these happen to return pathnames rather than URLs, consider naming them accordingly.

packages/venia-ui/lib/util/images.js
```diff
@@ -17,23 +17,52 @@ export const imageWidths = {
...
 
-    return `${url} ${width}w`;
+export const generateUrlFromContainerWidth = (
+    imageURL,
+    containerWidth,
+    type = 'image-product'
+) => {
+    const intrinsicWidth = window.devicePixelRatio * containerWidth;
+    /**
+     * Using the reduce on imageWidths to find the best width that is
```
jimbo:
```suggestion
     * Find the best width that is
```

packages/venia-ui/lib/util/images.js
```diff
@@ -17,23 +17,52 @@ export const imageWidths = {
...
+export const generateSrcset = (imageURL, type) => {
+    if (!imageURL || !type) return '';
+
+    const generateSrcsetUrl = generateUrl(imageURL, type);
 
-    const generateSrcsetURL = generateURL(imageURL, mediaBase);
     return Object.values(imageWidths)
-        .map(width =>
-            generateSrcsetURL(width, width / DEFAULT_WIDTH_TO_HEIGHT_RATIO)
+        .map(
```
jimbo: I strongly suggest making `imageWidths` a `Map` rather than an object if we're going to iterate over it, especially since we're doing so in multiple places.

packages/peregrine/lib/talons/ProductImageCarousel/useProductImageCarousel.js
```diff
@@ -19,6 +24,22 @@ export const useProductImageCarousel = props => {
         setActiveItemIndex(0);
     }, [images, setActiveItemIndex]);
 
+    useEffect(() => {
+        const urls = images.map(
+            ({ file }) =>
+                `${location.origin}${generateUrlFromContainerWidth(
```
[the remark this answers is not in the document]
revanth0212: I was not able to get `origin` from the react-router [`location`](https://reacttraining.com/react-router/web/api/location) object.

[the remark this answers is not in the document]
jimbo: Ah, my mistake. Looks like `window` it is.

packages/venia-ui/lib/util/images.js
```diff
@@ -55,7 +60,7 @@ export const generateSrcset = (imageURL, type) => {
 
     const generateSrcsetUrl = generateUrl(imageURL, type);
 
-    return Object.values(imageWidths)
+    return Array.from(imageWidths.values())
```
jimbo: The second argument to `Array.from()` is a mapping function, which will let you optimize by avoiding the intermediate array.

revanth0212: Nice point. I didn't know that. Done.

jimbo:
Approved.

On a side note, why does Github say there are 163 commits?

revanth0212:
> Approved.
> 
> On a side note, why does Github say there are 163 commits?

Because it was based out of 2 PRs which again were based out of other PRs. But now that all the PRs that this guy is based on have been merged, we only see the code changes but commits still persist.

package.json
```diff
@@ -34,9 +34,9 @@
...
+    "test:ci": "node --experimental-worker node_modules/.bin/jest --no-cache -i --json --outputFile=test-results.json",
```
revanth0212: image


```suggestion
    "test:ci": "node --experimental-worker node_modules/.bin/jest --no-cache --json --outputFile=test-results.json",
```

If we remove the `-i` flag from `test:ci` jest does not complain about `MessagePort` open handles. @zetlen @jimbo let me know how to solve this?

revanth0212:
Man, this is irritating. After removing `-i` flag from `test:ci` script, it works fast and closes promptly in local:

image

But again is stuck when the same job is run in AWS:

image

What shall I do?

It took 126 Seconds in local compared to 329 Seconds in AWS. Do you think AWS is running an old version of the script, meaning is it still running with the '-i` flag?

jest.config.js
```diff
@@ -6,6 +6,10 @@
  */
 const path = require('path');
 
+const worker = require('worker_threads');
```
sirugh: This breaks `yarn test:dev`, which I use regularly. Please add the experimental flag!

package.json
sirugh: Let's keep this open until you really resolve the whole issue. Seems like removing the flag still causes problems. I suggested in slack -- I think we just use `-i` and let it warn (only if we trash the instances after running tests).

jest.config.js
revanth0212: The experimental flag is present. Check out the package.json file.

Also tests pass in my local. Can you check if you are running an older verison?

revanth0212: Done

image

dpatil-magento:
@revanth0212 QA Pass, good to merge if there are no more changes.

package.json
```diff
@@ -35,7 +35,7 @@
         "stats:venia": "yarn venia run build:analyze && yarn venia run stats",
         "storybook:venia": "yarn venia run storybook",
         "test": "node --experimental-worker node_modules/.bin/jest",
-        "test:ci": "node --experimental-worker node_modules/.bin/jest --json --outputFile=test-results.json",
+        "test:ci": "node --experimental-worker node_modules/.bin/jest --max-workers=3 --json --outputFile=test-results.json",
```
revanth0212: Setting `--max-workers=3` for `test:ci`. This will limit threads spaw to 3. 3 because these tests are run on machine with 4 CPU cores, 1 for the main thread and 3 for testing workers. This is configuration is recommended by [Jest docs](https://jestjs.io/docs/en/cli#maxworkers-num-string).

After making the change, I have noticed the number of threads spawned was reduced by half (previously 7 on my local machine because it was quad-core dual threaded machine giving it 8 virtual cores, 1 for main thread and 7 workers). Tests take longer but less footprint at the same time. Hopefully, this will fix our AWS resource allocation issue.

revanth0212: I don't have access to change the AWS machine settings. Can someone change the settings to use `3 GB memory, 2 vCPUs`? As of now, it is `7 GB memory, 4 vCPUs`.

https://console.aws.amazon.com/codesuite/codebuild/projects/pwa-pr-test/edit/environment?region=us-east-1

image

revanth0212:
AWS has been updated to use the old configuration of 3GB and 2 vCPUs. Working like expected. @zetlen was right, changing the script to specify the number of workers to use was the key. Ready to be merged.

package.json
```diff
@@ -1,108 +1,108 @@
...
-  "husky": {
-    "hooks": {
-      "pre-push": "yarn run prettier:check && yarn run lint"
-    }
-  },
-  "bundlesize": [
-    {
-      "path": "./packages/venia-concept/dist/{client,vendors,sw}?(.)*.js",
-      "maxSize": "250 kB"
+    "name": "@magento/pwa-studio",
```
jimbo: Why is it changing indentation from two spaces to four? Something appears to be misconfigured on this branch.

revanth0212: Handled.

PR #1941, opened 2019-10-29
[chore]: Fixing Prettier use

sirugh:
## Description

1) Removes unnecessary `prettier-check` dependency and modifies pre-push to use `prettier:check` instead of `prettier:validate`.

2) Creates a `prettier.config.js` file and uses the config version of `--single-quote`. I think this is a better experience for developers for future configuration on top of venia-concept.

## Related Issue

Closes [PWA-89](https://jira.corp.magento.com/browse/PWA-89).

### Verification Stakeholders
@zetlen - I'd like to know if there are any additional changes necessary in `venia-concept` or in the scaffolding scripts.

### Verification Steps

1. Make some change in `venia-concept` that would fail prettier like adding double quotes to an import. Make sure you have "format on save" off so that it doesn't auto fix. 
2. Run `yarn prettier:check` from the root directory. It should complain.
3. Go to `packages/venia-concept` and run `yarn prettier`. It should also complain.

## Checklist


packages/venia-concept/prettier.config.js
```diff
@@ -0,0 +1,5 @@
+const config = {
+    singleQuote: true
+};
+
+module.exports = config;
```
jimbo: Should this file be located in the `pwa-studio` root?

sirugh:
Good question. My intent was to apply the prettier config that would be copied with venia concept and one that is simple and easy to override and modify. We don't make changes often within venia-concept. This just another template file, basically.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

jimbo:
I'll defer to @zetlen on this one. Looks fine to me.

packages/venia-concept/prettier.config.js
sirugh: Just to put context in this convo for later -- my intent was that `venia-concept`, when used as the root "theme" from the scaffolding script, would provide this config. I wasn't sure if the scaffolding script already did that work and copied the one in the root folder. If that is the case, then I can delete this file.

zetlen: This is the right move. Keeping these two files in sync shouldn't be a big problem, because they should VERY rarely change.

zetlen:
Oh, @sirugh please open a dummy issue against this and link it so it passes Danger.

PR #1945, opened 2019-10-30
Bugfix/fixing storybook venia ui

### LINKED ISSUE #1577: Venia UI Storybook is broken
`yarn run storybook:venia` is currently broken with the following error:

```
UnhandledPromiseRejectionWarning: Error: Cannot find module '../webpack.config.js'
```

There were some major updates to `webpack.config.js` in https://github.com/magento/pwa-studio/pull/1498.

LucasCalazans:
## Description

Venia UI was not able to execute Storybook (Probably since the separation with `venia-concept`) because it was depending on a lot of the `venia-concept` files.

My solution was to split the main part of the `configureWebpack` data in the `base.webpack.config.js` file, with that we can import the same file/config into the `venia-concept` and `venia-ui` without duplicate the code.

To do that I had to include `@magento/pwa-buildpack` as a `peerDependency` of `venia-ui`. Without that, we, probably, would have to import the webpack config manually.

## Related Issue
Closes #1577.

### Verification Steps
1. Go to the root project folder.
2. Run `yarn storybook:venia`

OR
1. Go to the `venia-ui` package folder
2. RUN `yarn storybook`

## Checklist

## Additional Information
I fixed some problems related to Context API (Some components are using) adding a storybook decorator. With that, we will no longer need to add the `Adapter` neither the `AppContextProvider`.

## Problems
1. We don't have the `index.css` anymore so some parts are missing styles (Like the MiniCart component)
2. The `cartTrigger` component is calling REST APIs without using the BACKEND_ENDPOINT, so the calls are been succeded.


zetlen:
@LucasCalazans Thank you for working on this abstraction. It's very well thought out; however, I'm not sure we want to add another layer of default Webpack config beyond `configureWebpack` at this stage. The `configureWebpack` interface is subject to future changes to make it compose together the needs of various dependent libraries, so in the future we shouldn't need this.

I don't really have a problem with implicit `dev` dependencies between `venia-ui` and `venia-concept`. Because `venia-ui` and `venia-concept` are developed only in this repo, the `venia-ui` storybook should only run if you have the full repo checked out (or you have deployed it in a custom way, as in our documentation). So I'm wondering if we could fix #1577 simply by changing the imported Webpack path to import from the correct package.

Maybe I'm forgetting something that @supernova-at could remind me of; this bug is pretty old, so maybe it's more complicated than that.

supernova-at:
Yes!!! Thank you so much for this. Taking a look now.

supernova-at:
> we, probably, would have to import the webpack config manually

> I don't really have a problem with implicit dev dependencies between venia-ui and venia-concept. Because venia-ui and venia-concept are developed only in this repo, the venia-ui storybook should only run if you have the full repo checked out (or you have deployed it in a custom way, as in our documentation). So I'm wondering if we could fix #1577 simply by changing the imported Webpack path to import from the correct package.

Agreed, let's try importing it manually.

I'd love to get `index.css` back, and some new `Image` stories just landed that it looks like GitHub is saying have some conflicts.

LucasCalazans:
@zetlen I see. I was thinking about an application that just uses the `venia-ui` without the `venia-concept`.

But yeah, sorry about the delay in the answer, I'm working in a new version based on the venia-concept dev dependency. I'll try to push a commit today

LucasCalazans:
@supernova-at Just to be clear, do you think that we should import the webpack dependencies in the venia-ui or import the venia-concept's `webpack.config`?

LucasCalazans:
Hey guys, I've tried importing the `venia-concept`, do you agree with this? Can you provide a new review please?

It's now getting some errors in the new Image stories. I've debugged and it seems to be a problem with some params that were not passed into the Image component like the `resource` that is required but in some stories (like `'An Image without a placeholder (throttle me)'`) it is not passed.

packages/venia-ui/.storybook/config.js
```diff
@@ -1,8 +1,26 @@
...
-    const context = require.context('../src', true, /__stories__\/.+\.js$/);
+    const context = require.context('../lib', true, /__stories__\/.+\.js$/);
     context.keys().forEach(context);
 }
 
+const apiBase = new URL('/graphql', backendUrl).toString();
+
+addDecorator(storyFn => (
+    <Adapter
+        apiBase={apiBase}
```
supernova-at: Note to team: this may need to be updated alongside #1966 .

supernova-at:
I think this is great - definitely love moving the `Adapter` into the setup area and out of individual stories 💯 

`yarn storybook:venia` isn't working for me, though. For some reason it's building the `storybook-dist` directory but it doesn't actually launch storybook.

I fixed it by updating the root `package.json` scripts:
```"storybook:venia": "pushd ./packages/venia-ui && yarn storybook && popd",```

Not sure if that's the best way to do it but at least it works.

I can verify that Storybook starts up without error, even though many of the stories are broken.
I'm okay with creating follow up tickets to address the actual contents of our Storybook.

**Comment by supernova-at** [comment]
Also I see ` A new version (5.2.6) is available! `, do we want to upgrade Storybook as part of this PR?

**Comment by LucasCalazans** [comment]
Hey @supernova-at, sorry about the delay.

Thank you for your help here. Do you think that there is something I could do yet?

**Comment by supernova-at** [comment]
No worries @LucasCalazans I'm excited to get this in :)

I'm just going through and fixing up the broken stories now.

Would love a review from you and I'll request one from the team when I'm done.

Thanks for your work on this!

**Comment by supernova-at** [comment]
I had to remove the `MiniCart` stories because it is much harder to mock states with the talons.

We may want to revisit Storybook as a whole, but this PR at least gets it working and the stories in it operational 💯

File: packages/venia-ui/.storybook/getEnvironmentVariable.js [hunk 2]
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Returns the value of the first variable in any section whose name matches the input.
+ * @param {string} targetVariable - The name of the environment variable to find.
+ */
+const getEnvironmentVariable = targetVariable => {
+    let match;
+    envVarDefinitions.sections.forEach(section => {
+        section.variables.forEach(variable => {
+            if (variable.name === targetVariable) {
+                match = variable;
```
revanth0212: What if there are multiple variables with `name === targetVariable`?

If there is only 1 that is expected or only the first one that is needed, consider using a `for of` loop so you can `break` the loop once you are in the `if` case.

packages/venia-ui/lib/components/SearchBar/__stories__/searchBar.js
```diff
@@ -3,8 +3,13 @@ import { storiesOf } from '@storybook/react';
...
-const stories = storiesOf('SearchBar', module);
+const stories = storiesOf('Components/SearchBar', module);
 
 stories.add('Search Bar', () => (
-    <SearchBar classes={defaultClasses} isOpen={true} />
+    <SearchBar
+        classes={defaultClasses}
+        isOpen={true}
+        location={{}}
+        history={{ push: () => {} }}
```
revanth0212: Would discourage from doing this. This will create a new function on every re-render.

revanth0212:
@LucasCalazans can you also check on the merge conflicts?

LucasCalazans:
Hey @revanth0212, thanks for your review. I've just pushed the changes and merge fixes

package.json
```diff
@@ -34,7 +34,7 @@
     "prettier:check": "prettier --list-different '@(packages|scripts)/**/*.@(css|graphql|js)' '*.js'",
     "stage:venia": "yarn venia run start",
     "stats:venia": "yarn venia run build:analyze && yarn venia run stats",
-    "storybook:venia": "yarn venia run storybook",
+    "storybook:venia": "pushd ./packages/venia-ui && yarn storybook && popd",
```
sirugh: Haha tricky. How about we just use the workspace command to run the command? `yarn workspace @magento/venia-ui run storybook`. I think you made this change because `yarn venia` correlates to `yarn workspace @magento/venia-concept` instead of `venia-ui`.

packages/venia-ui/.storybook/config.js
sirugh: Does this still need to get updated or can we resolve the comment? That PR has been merged.

packages/venia-ui/.storybook/getEnvironmentVariable.js
```diff
@@ -0,0 +1,30 @@
+const envVarDefinitions = require('@magento/pwa-buildpack/envVarDefinitions.json');
+
+/**
+ * Returns the value of the first variable in any section whose name matches the input.
+ * @param {string} targetVariable - The name of the environment variable to find.
+ */
+const getEnvironmentVariable = targetVariable => {
```
sirugh: Can you just use `loadEnvironment` instead of this custom function? `loadEnvironment` should be able to do exactly what this function is doing.

packages/venia-ui/package.json
```diff
@@ -75,6 +75,8 @@
   "peerDependencies": {
     "@magento/babel-preset-peregrine": "~1.0.0",
     "@magento/peregrine": "~4.0.0",
+    "@magento/pwa-buildpack": "~4.0.0",
+    "@magento/venia-concept": "~4.0.0",
```
sirugh: The idea is that `peregrine` is used by `venia-ui` is used by `venia-concept`. This seems like a red flag to me? But maybe I'm misunderstanding peer deps :P

I think this, along with the imports of `venia-concept` files in `venia-ui` just seem like red flags to me in terms of where we are supposed to be keeping these things. Maybe in the future we will move the majority of the base config to the top-level and `venia-concept` and `venia-ui/storybook` can just import/use it.

TL;DR - I think I'm fine with this, just rambling.

package.json
LucasCalazans: Great idea haha, I'll update with this.

packages/venia-ui/.storybook/getEnvironmentVariable.js
LucasCalazans: Hmmm, actually I'm not sure

I tried to use it but it seems to have a dependency exactly where I need, in the MAGENTO_BACKEND_URL, when I use loadEnvironment to get this variable it throws a exception:

```
  ⓧ  Missing required environment variables:
     MAGENTO_BACKEND_URL: Connect to an instance of Magento 2.3 by specifying its public domain name. (eg. "https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/")
```

I tried using this code: 
```
const loadEnvironment = require('@magento/pwa-buildpack/lib/Utilities/loadEnvironment');
loadEnvironment('MAGENTO_BACKEND_URL')
```

Do you know why is this happening? Maybe I'm using it in a wrong way

packages/venia-ui/.storybook/config.js
supernova-at: Looks like we're good 👍

packages/venia-ui/package.json
supernova-at: Yeah I think for the storefront app that's correct. For Storybook we relaxed the rules a bit so we could run it easily.

packages/venia-ui/.storybook/getEnvironmentVariable.js
supernova-at: I'm not sure about this either. The function signature of [loadEnvironment](https://github.com/magento/pwa-studio/blob/develop/packages/pwa-buildpack/lib/Utilities/loadEnvironment.js#L128) is

```js
function loadEnvironment(dirOrEnv, logger = prettyLogger) {
```

so it doesn't look like it wants the name of the environment variable.

sirugh: @LucasCalazans check out other places where we use it. For example in `create-custom-origin`:

```js
const projectConfig = loadEnvironment(projectRoot);
        if (projectConfig.error) {
            failExpected(projectConfig.error);
        }
        const config = projectConfig.section('customOrigin');
        // etc
```

You pass it a directory and it returns an object with props on it. I'm not sure how well it is documented unfortunately. It may take some experimenting.

sirugh:
@LucasCalazans Are you working on the last changes I requested? Just checking :)

sirugh:
I have some ideas on how to improve this. I'll go ahead and make them, should be small changes.

packages/venia-ui/package.json
jimbo: Yeah, it's a tough call. I do think @sirugh has the right of it here.

It's true that `venia-ui` components are hard to render without the glue that `venia-concept` adds, so we need to get that glue from somewhere. But `venia-ui` is built to be a dependency of `venia-concept`, so I'd prefer that it didn't list `venia-concept` as a dependency or import anything from it. Thinking about our various options here:

Option | Evaluation
:--- | :---:
Duplicate `store.js`, `webpack.config.js`, and `index.css` | bad
Move these files up to some shared package | bad
Import these files from `venia-concept` | not ideal
Move `venia-ui` storybook stories into `venia-concept` | okay?

Overall, I would accept the approach in this PR because the work's been put in, it has value, and only the Storybook files are importing from ` venia-concept`. But eventually I want to remove the circular dependency.

packages/venia-ui/.storybook/webpack.config.js
```diff
@@ -6,7 +6,7 @@ const {
 const baseWebpackConfig = require('@magento/venia-concept/webpack.config');
 const { DefinePlugin, EnvironmentPlugin } = require('webpack');
 
-module.exports = async storybookBaseConfig => {
+module.exports = async ({ config: storybookBaseConfig }) => {
```
sirugh: o.o

PR #1949, opened 2019-10-30
[feature] category breadcrumbs

### LINKED ISSUE #1489: [FEATURE]: Breadcrumbs
As a user I'd like to see breadcrumbs on most pages to give me a better idea of where I am while shopping. 

I'd like to see a new peregrine component that provides breadcrumbs

### JIRA issue for design
https://jira.corp.magento.com/browse/UXD-182

- [X] `venia-concept`
- [ ] `pwa-buildpack`
- [X] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

sirugh:
## Description

* **Adds breadcrumbs to Category pages.**
    Breadcrumbs endpoint, after 2.3.4, will return `category_url_path` which we can use to generate the link for the breadcrumb. For now, this is just rendering Links and dividers.

## Related Issue

Closes [PWA-98](https://jira.corp.magento.com/browse/PWA-98).

### Verification Stakeholders
@schensley 
### Verification Steps

1. From the landing page, click a category. You should see `Home`.
2. From left nav, click all the way into a category (like `Bottoms -> Skirts`). It should render and you should see breadcrumbs for it.
3. Do these things on desktop and mobile :)

[Image from Gyazo](https://gyazo.com/a986c4912e9765db6696c9fd8f1f6271)
## Checklist


schensley:
@sirugh  I think the implementation and behavior is good based on the fact that GraphQL cannot determine the specific path a shopper took to arrive at the PDP.

1. I think the font size of the breadcrumbs should be reduced, while maintaining an adequate "tappable area".
2. Add top padding to the breadcrumbs to provide more separation / clear space between the header and the breadcrumbs.  Same for "Back" button on PDP.
3. Could you explore center justification of the breadcrumbs, since the page title is centered?  But keep the "Back" button (on PDPD) left-justified?

brendanfalkowski:
Consider dropping the back button? Users have a physical back button in their device, browser UI, or gestures.

Also probably a good idea to keep the breadcrumbs left aligned, and provide wrapping or overflow because IRL category names get long and they will overflow or wrap (sometimes so badly to the point that you decide to drop breadcrumbs).

sirugh:
@schensley updates:

[Image from Gyazo](https://gyazo.com/0a54641f8b746738e0d99ac861798b6b)

zetlen:
Hey @schensley, do you have an opinion on the @brendanfalkowski suggestion?

On the one hand, It's a best practice for us to keep the navigation obvious. Not all interactions are routable, so not everything works with a browser-based "Back" swipe.

But this one does! We don't have Back buttons everywhere. Should we omit this one, or put them everywhere?

One argument for explicit Back buttons is that a PWA can be installed to the Home screen. When that happens, the browser navigation goes away! The Back swipe still works (unless something has captured the gesture) but it's still a different runtime, so we shouldn't rely too much on that.

sirugh:
@zetlen I have since removed the `Back` button from the PDP as it was out of scope anyways. This PR is now just for breadcrumbs in the category page :)

schensley:
A few notes on the UI/UX as we work through this design.
- Breadcrumbs should be left justified and extend from L to R in the viewport.
- Breadcrumb list should terminate with the current page (non-link)
- Need to better understand if/how GraphQL might allow for the structural category path to be determined and rendered on the PDP.  This becomes valuable especially in cases where a shopper has arrived on the PDP from an external search.

sirugh:
`hold` as I need to address UX feedback. I will implement some form of breadcrumbs on the PDP as well -- it just may be a different look than whats on the category pages.

sirugh:
@schensley added non-functional text for current page to end of breadcrumbs.

[Image from Gyazo](https://gyazo.com/ea72ddce534e1bff72f0eb00c49562aa)

schensley:
@sirugh The latest screens look good to me.

packages/venia-ui/lib/components/Breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,73 @@
+import React, { Fragment, useMemo } from 'react';
+import { Link, resourceUrl } from '@magento/venia-drivers';
+import { mergeClasses } from '../../classify';
+import defaultClasses from './breadcrumbs.css';
+
+const URL_SUFFIX = '.html';
```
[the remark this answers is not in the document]
sirugh: We use `.html` hardcoded everywhere else.

I could, for this PR, add `storeConfig > category_url_suffix` to the query but we'll just have to remember to switch this over when we grab it at build.

packages/venia-ui/lib/components/Breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,73 @@
...
+ * Breadcrumbs! Sorts and generates links for an array of breadcrumb data.
+ *
+ * @param {String} props.currentCategory name of the current category
+ * @param {Array} props.data Breadcrumb data
+ */
+const Breadcrumbs = props => {
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const data = props.data || [];
+
+    const sortedData = useMemo(() => data.sort(sortCrumbs), [data]);
```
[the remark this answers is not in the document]
sirugh: Anything unrelated to display. I completely forgot to do this talon!!

packages/venia-ui/lib/components/Breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,73 @@
...
+
+    const normalized = useMemo(
+        () =>
+            sortedData.map(category => ({
+                text: category.category_name,
+                path: getPath(category)
+            })),
+        [sortedData]
+    );
+    const divider = useMemo(() => <span className={classes.divider}>/</span>, [
```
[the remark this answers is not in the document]
sirugh: I'm not sure. If I could create this outside the component I would but it needs the `divider` class. I think this'll be a good lesson for me here :D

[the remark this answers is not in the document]
jimbo: I agree, this is overzealous. I'd rather just render this in the `return` statements.

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
```diff
@@ -0,0 +1,51 @@
...
+const sortCrumbs = (a, b) => a.category_level > b.category_level;
+
+// Generates the path for the category.
+const getPath = (path, suffix) => {
+    if (path) {
+        return `/${category_url_path}${suffix}`;
+    }
+
+    // If there is no path this is just a dead link.
+    return '#';
```
sirugh: Is this a more frustrating UX than breadcrumbs that don't actually do anything? Right now we display clickable links like `Home / Shop The Look / Minimalist` where "Shop The Look" is clickable but is a link to the current page (until `category_url_path` is returned from graphql and we start using it.

Should we return `null` or `undefined` instead and have the UI render plain text?

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
```diff
@@ -0,0 +1,51 @@
...
+                breadcrumbData.sort(sortCrumbs).map(category => ({
+                    text: category.category_name,
+                    path: getPath(category.category_url_path, categoryUrlSuffix)
+                }))
+            );
+        }
+    }, [data, loading]);
+
+    return {
+        currentCategory: (data && data.category.name) || '',
```
sirugh: Can `data` be truthy but not have `category`? I think we do this throughout the repo...

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
```diff
@@ -43,6 +45,7 @@ const CategoryContent = props => {
     const modal = filters ? <FilterModal filters={filters} /> : null;
     return (
         <Fragment>
+            <Breadcrumbs categoryId={categoryId} />
```
sirugh: Display breadcrumbs with this one simple prop!

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
tjwiebell: No, strict typed responses should guard this. Since you're not handling the error state, it would be worth checking that `data` doesn't get updated to something truthy when an error is thrown.

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
tjwiebell: Dead links for the sake of the UX looking closer to final state seems ideal here, I'm fine with this.

brendanfalkowski:
I'd prefer rendering a skeleton until the data returns, so it doesn't shift.

Aside — we started developing breadcrumbs before this feature started, and made a couple different UX choices.

1. Drop the "home" link. It's redundant because of the logo link, the obvious origin for all links, and takes up valuable space in the breadcrumb line.

2. Don't show breadcrumbs when there's only one crumb. This means no breadcrumbs on top-level categories (because home is gone). So on sub-categories you'd see "Vaccuums > Robotic".

This ends up working a lot better on small screens, and doesn't get overloaded until three categories down.

sirugh:
@brendanfalkowski interesting. I ended up just implementing the design @schensley and I discussed but I think we could iterate on it. I don't think there's any rush to get this feature just yet so we may as well try to get it right the first time!

sirugh:
> With moving the data fetch to the component, the re-render after data returns is very noticeable.

@tjwiebell Ah. I can return an empty, styled div so it's height is sized appropriately. It will still cause long (wrapping) crumbs to cause a pop-in though. We can't know the number of links or the length of the text so this is a difficult one. We could _always_ show the last two crumbs in a link and just put some indicator of such before? `... / sub category / last category`? Just throwing ideas around.

brendanfalkowski:
Another thing you can do to shorten the breadcrumb line is omitting the last crumb or putting a non-link (e.g. "here"). Whether it's a category, product, or CMS page that last crumb is usually redundant because the very next thing on screen is an H1 with the same text.

This makes a major difference on the product page because unlike categories product names can easily be 50+ characters alone if they have to distinguish variants.

sirugh:
Ok I have updated this PR once more with some suggestions. Namely:

1) No more Home or "current category" text.
2) Static height and no wrap/overflow.

Here's what it looks like:

**Mobile:**
[Image from Gyazo](https://gyazo.com/7076f3256f5c9ddd1c166e64f62eb205)

**Desktop:**
[Image from Gyazo](https://gyazo.com/7fe01a1cecfe5bfb0a276b0fa7344300)

PR #1952, opened 2019-10-31
[bug]: debounce search input

### LINKED ISSUE #1225: [bug]: Search suggestion box - Previous pending GraphQl query should be canclled if new GraphQl query is initiated
Steps - 
1. Go to Venia storefront and Enter search string Top    (GraphQl query should be initiated)
2. Now before search suggestion shows results ...append S to string "Top" to make it "Tops"

Expected - Query for string Top should be cancelled and results for new query Tops should be returned.
Actual - Both queries run together and updates results.

sirugh:
## Description

This is bad:

[Image from Gyazo](https://gyazo.com/6a29283668cd2f6b8c11255050815b35)

We had hoped to be able to cancel the query on new input but that isn't something the Apollo Client can do. Instead of cancelling, we will debounce the query function so we only run it 500ms after the last change.

## Related Issue

Closes #1225 .

### Verification Steps

1. Open the search bar and type some long word really quickly. The query should only run after you complete typing.

[Image from Gyazo](https://gyazo.com/b86083fcf54d0d3b2b17adbe377ca9b6)

## Checklist


packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -41,15 +59,7 @@ export const useAutocomplete = props => {
...
 
-    // run the query once on mount, and again whenever state changes
-    useEffect(() => {
-        if (visible && valid) {
-            runQuery({ variables: { inputText: value } });
-        }
-    }, [runQuery, valid, value, visible]);
-
     return {
-        hasResult,
```
sirugh: Unused in `autocomplete` component.

brendanfalkowski:
@sirugh — You might be able to improve the race condition handling due to lack of canceling requests. I did this in an old jQuery typeahead library I wrote where the debounce rate was low (150ms) for a faster response, but the incoming responses always compared the input's value to the returning query value. If they didn't match the response is discarded. Then it didn't matter what order the responses stream back. Only the current query's response can render.

packages/peregrine/lib/talons/SearchBar/__tests__/useAutocomplete.spec.js
```diff
@@ -19,6 +19,11 @@ jest.mock('@apollo/react-hooks', () => {
     return { runQuery, queryResult, useLazyQuery };
 });
 
+// Could not figure out fakeTimers. Just mock debounce and call callback.
```
zetlen: Lodash debounce hates fakeTimers. We can switch to a smaller, simpler implementation, since we don't need all the Lodash options.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -23,7 +24,24 @@ export const useAutocomplete = props => {
     const { value } = useFieldState('search_query');
     const valid = value && value.length > 2;
 
-    // determine message type
+    // Create a debounced function so we only search some delay after the last
+    // keypress.
+    const debouncedRunQuery = useMemo(
+        () =>
```
zetlen: Why not just debounce the callback to useMemo? Did you try it and it was weird? It seems like it should work out of the box.

```jsx
  debounce(runQuery, 500),
  [runQuery]
);
```

sirugh: I tried this and it didn't work :/ I can try again, maybe I borked the syntax or something.

packages/peregrine/lib/talons/SearchBar/__tests__/useAutocomplete.spec.js
sirugh: Is `lodash.debounce` that big? I don't mind writing a smaller implementation but we already `_` as a dependency to the project.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
sirugh: Yea I can't do this -- I can simplify and not forward `opts` in the internal callback to `debounce` but `useMemo` requires a ["create" function](https://reactjs.org/docs/hooks-reference.html#usememo).

zetlen: Oh, of course. My bad.

packages/peregrine/lib/talons/SearchBar/__tests__/useAutocomplete.spec.js
zetlen: Lodash is surprisingly big in places. I don't think the debounce is big, necessarily, but it is overly complex.

jimbo:
Would have liked a shot to review this one before it got merged.

PR #1953, opened 2019-11-01
Search RootComponent -> SearchPage Component + Talon

tjwiebell:
## Description

We've added talons for most of the components in `venia-ui`, but not for some of the root component. Create a talon for `Search`:

Create a `SearchPage` directory and `useSearchPage.js` in `peregrine/lib/talons`
Create a `SearchPage` component in `venia-ui/lib/components` and use the talon
Rewrite the `Search` root component to be a mere shell around `SearchPage`

## Related Issue

Closes [PWA-87](https://jira.corp.magento.com/browse/PWA-87) Create a talon for the Search root component

### Verification Steps

1. Go to Search page (`/search.html?query=shirt`)
2. Verify Jillian Top is displayed
3. Go wild, verify that no regressions were introduced

## Checklist


packages/venia-ui/lib/RootComponents/Search/search.js
```diff
@@ -1,119 +1,9 @@
...
-                {maybeCategoryFilters}
-                {maybeFilterButtons}
-            </div>
-            {maybeFilterModal}
-            <section className={classes.gallery}>
-                <Gallery items={items} />
-            </section>
-        </article>
-    );
+    return <SearchPage {...props} />;
```
sirugh: Why did we need to make this a thin wrapper around the `SearchPage` component?

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -0,0 +1,90 @@
...
+
+    if (items.length === 0) {
+        return <div className={classes.noResult}>No results found!</div>;
+    }
+
+    const maybeCategoryFilters = categoryId ? (
+        <CategoryFilters
+            categoryId={categoryId}
+            executeSearch={executeSearch}
+            history={history}
```
sirugh: The talon for `CategoryFilters` should be the one importing/using `useHistory`. Same with `location` if possible.

packages/venia-ui/lib/RootComponents/Search/search.js
jimbo: > Why did we need to make this a thin wrapper around the `SearchPage` component?

I put that in the AC. Talons don't need to know about the `RootComponent` concept; ideally we'd just have all our components in `venia-ui/lib/components` and all our talons in `peregrine/lib/talons`.

This way, `Search` does the following:

- is a React component
- maps to a page type that Magento recognizes
- exists in a place the router & bundler know to look
- doesn't need a talon

And `SearchPage` does the following:

- is a React component
- renders elements
- has default styles
- has a corresponding talon with logic

packages/venia-ui/lib/components/SearchPage/searchPage.js
tjwiebell: 👍 Should have mentioned I was waiting to refactor `categoryFilters` until this early feedback round, but I agree with you. Safe to assume that would live in a `CategoryFilter` talon, correct?

packages/venia-ui/lib/RootComponents/Search/search.js
brendanfalkowski: Sliding in a footnote: early on I found the naming of components, root components, and ThingPage confusing.

I tried replicating this when creating my own components/Thang, RootComponents/Thang, and components/ThangPage components — but then it looked ridiculous. I ended up reducing it all to RootComponents/Thang for each "page".

Is there an overall plan for these distinctions? I couldn't find anything around the ThingPage convention in PWA Studio docs or general React reading.

packages/venia-ui/lib/RootComponents/Search/search.js
```diff
@@ -1,119 +1,9 @@
...
 
-import { mergeClasses } from '../../classify';
-import Gallery from '../../components/Gallery';
-import FilterModal from '../../components/FilterModal';
-import { fullPageLoadingIndicator } from '../../components/LoadingIndicator';
-import PRODUCT_SEARCH from '../../queries/productSearch.graphql';
-import getQueryParameterValue from '../../util/getQueryParameterValue';
-import CategoryFilters from './categoryFilters';
-import defaultClasses from './search.css';
+import SearchPage from '../../components/SearchPage';
```
jimbo: Instead of defining a new component, you may be able to just re-export `SearchPage`.

```suggestion
export { default } from '../../components/SearchPage';
```

packages/venia-ui/lib/RootComponents/Search/search.js
jimbo: > Is there an overall plan for these distinctions? I couldn't find anything around the ThingPage convention in PWA Studio docs or general React reading.

We're only just now getting to these root components; essentially we're starting to define that plan in this PR.

I'd like `SearchPage`, etc., to live in `components` since they're just normal components, and we can just re-export them in the `RootComponent` directory since that's how our bundler associates them with Magento routes.

You'd be hard-pressed to find similar implementations in React world since a React Router app almost always allows the client to know about the routes. The way we consult Magento for our routes and then use a bundler to map them to components from a directory is fairly unusual.

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -0,0 +1,68 @@
+import { useCallback, useEffect } from 'react';
+import { useQuery } from '@apollo/react-hooks';
+import { useAppContext } from '@magento/peregrine/lib/context/app';
+import { useHistory, useLocation } from 'react-router-dom';
+
+import getQueryParameterValue from '../../util/getQueryParameterValue';
```
jimbo: Since we're moving these utilities, I think this is a good opportunity to rename them. I think the we should avoid the term "query parameter":

- the modern name for them appears to be [`SearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
- `query` is extremely overloaded thanks to GraphQL terminology

Would love to see this renamed to `getSearchParamValue`, and similar renames for other utils and local variables.

packages/venia-ui/lib/components/SearchPage/searchPage.js
jimbo: Cool. I agree with @rugh, but I also think this is fine for now. Your call, @tjwiebell.

packages/venia-ui/lib/RootComponents/Search/search.js
jimbo: > but then it looked ridiculous

I agree that if we have two components for one page it gets ridiculous. The `Product` and `ProductFullDetail` split is not ideal right now; logic is just arbitrarily split between the two of them, which is unfortunate.

I think with simple re-exports it won't be an issue. They're also default exports, so easily renamed on import.

packages/venia-ui/lib/components/SearchPage/searchPage.js
tjwiebell: Moved both to `CategoryFilter`.

packages/venia-ui/lib/RootComponents/Search/search.js
tjwiebell: Done

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
tjwiebell: Discovered that you created this utility in Peregrine already, so I went ahead and just deleted this one 💣

PR #1955, opened 2019-11-03
Lazy Load AuthModal

### LINKED ISSUE #1733: [bug]: Search and Create Account should load asynchronously on request instead of in main bundle
**Describe the bug**
The Search and CreateAccount RootComponents should load asynchronously, like all other RootComponents such as Product and Category. Instead, [the router statically imports them](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/App/renderRoutes.js) to manually specify them in the bundle.

These should change to dynamic imports, to reduce the size of the initial `client.js` bundle and improve load time.

**To Reproduce**
1. Run `yarn run stats:venia`.
2. When the Webpack Bundle Analyzer opens up, examine the `client.js` bundle.
3. Notice that the RootComponents for Search and CreateAccount are compiled into the `client.js` entry file.

**Expected behavior**
A separate async chunk file for each RootComponent.

**Additional context**

The Search and CreateAccount RootComponents are different from the catalog RootComponents, because the app does not route them with the `urlResolver` query via `resolveUnknownRoute()`. Instead, they are hardcoded routes in `renderRoutes.js`. That makes sense.

However, they don't need to be statically imported into `renderRoutes.js`. We can use dynamic `import()` calls to make separate chunks from those files, just as the router automatically does with Product, Category, and CMSPage RootComponents.

**Possible solutions**
Modify `renderRoutes.js` to import the hardcoded RootComponents using `import()` and `React.lazy()`. Use `Suspense` and a fallback spinner to asynchronously load them.

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

dani97:
## Description

The `AuthModal` component can be lazy loaded, as guest users may never use it.

This saves us ~6.5 KB on initial load of the application.

## Related Issue

Closes PWA-113.

## Acceptance 

@zetlen 
### Verification Steps

1. `yarn watch:venia`
1. Open Venia in the browser
1. Open the left (navigation) drawer
1. View Browser's "Network" tab
1. Click the "Sign In" button
1. Verify that new entries appeared in the Network tab (downloading `AuthModal` chunk)
1. Verify that the `AuthModal` chunk loads and the app works as expected

You can also verify that the loading spinner appears by artificially slowing the network speed in the Network tab before clicking the Sign In button.

Screen Shot 2019-12-04 at 10 42 13 AM

Screen Shot 2019-12-04 at 10 32 17 AM
Screen Shot 2019-12-04 at 10 33 00 AM

## Checklist


zetlen:
@dani97 This seems like too granular of a place for dynamic imports. Can you post a short screen capture session showing us how quickly this responds to user input? It might be that we want to combine all those possible AuthModal screens into a single dynamic chunk, rather than separate ones.

dani97:
I can see a bit of lag on 3g connections but user experience is still good. I visited Sign in form, Forgot Password form and Create Account form

**At no throttling, with cache**
Screenshot 2019-11-04 at 11 56 41 PM
**At 3g connection, without cache**
Screenshot 2019-11-04 at 11 57 39 PM

zetlen:
Can you confirm that all three of these new dynamic imports are combining into a single chunk? If chunks get too small, then the overhead of Webpack boilerplate starts to be too much.

Don't get me wrong, I appreciate the contribution! I just need to be sure that we benefit from asynchronous imports at this level of granularity.

dani97:
Screenshot 2019-11-05 at 12 12 27 AM
Lazy loaded AuthModal and got this result. I thought of this approach but this bundle is fetched before user opens authModal as AuthModal is mounted initially, regardless not present in client bundle

dani97:
> Can you confirm that all three of these new dynamic imports are combining into a single chunk? If chunks get too small, then the overhead of Webpack boilerplate starts to be too much.
> 
> Don't get me wrong, I appreciate the contribution! I just need to be sure that we benefit from asynchronous imports at this level of granularity.
- I agree, when I lazy loaded all of these individually I get components as separate chunks and not a single bundle. In a ideal case a registered customer won't go for forgot password or create account or login component. The person will stick with customer account component.
- If a registered user installs the PWA , will login and  may use sign in and forgot password. He may never use create account component.
- A person who wants to perform a guest checkout may never use any of these components 
Your thoughts?

zetlen:
Seems like login -> forgot password -> create account is a common enough use case. I'd worry about bad network latency in those very bounce-prone workflows.

Did you try changing the place where AuthModal is important to be a dynamic import, rather than changing all of its downstream dependencies?

dani97:
> Seems like login -> forgot password -> create account is a common enough use case. I'd worry about bad network latency in those very bounce-prone workflows.
> 
> Did you try changing the place where AuthModal is important to be a dynamic import, rather than changing all of its downstream dependencies?
https://github.com/magento/pwa-studio/pull/1955#issuecomment-549492462 added the results here, just a new bundle that is called at initial page load

jimbo:
@zetlen Do you recall how we had to change `renderRoutes.js` so that `<Suspense>` was *above* the `<Switch>` element instead of below it in order to fix some strange behavior? Do you think we're likely to encounter that same behavior with these `<Suspense>` elements being conditionally rendered?

zetlen:
@jimbo I don't know; we'd need a test case to be sure. That bug with `Suspense` and `Switch` was easily reproducible because it was completely broken. THis isn't.

supernova-at:
> Did you try changing the place where AuthModal is important to be a dynamic import, rather than changing all of its downstream dependencies?

PR updated to show what this would look like.

`AuthModal` and all of its dependencies are not loaded until the "Sign In" button in the left drawer is clicked.

supernova-at:
Screen Shot 2019-12-04 at 10 32 17 AM
Screen Shot 2019-12-04 at 10 33 00 AM

packages/venia-ui/lib/components/Navigation/navigation.js
```diff
@@ -37,6 +39,21 @@ const Navigation = props => {
     const rootHeaderClassName =
         isTopLevel && view === 'MENU' ? classes.isRoot : classes.header;
 
+    // Lazy load the auth modal because it may not be needed.
+    const authModal = hasModal ? (
```
dani97: I didn't think of this conditional rendering of suspense. Cool @supernova-at 💯

dpatil-magento:
@dani97 @supernova-at Can you please resolve the conflict?

PR #1956, opened 2019-11-04
Improved Image API

supernova-at:
## Description

This PR improves the `Image` component API. Specifically, it:
  * No longer restricts breakpoints / image widths to three ("small", "medium", "large") sizes only
  * Renames `resourceSizes` / `resourceSizeBreakpoints` to `widths` and makes it a `Map`
  * Removes `resourceWidth` in favor of `width` (or the default entry in `widths`, if not present)
  * Removes `resourceHeight` in favor of `height`

## Related Issue

Closes JIRA-90.

### Verification Stakeholders

@zetlen @jimbo 

### Verification Steps

📝 This is just a code refactor. There should be no impact on the UI.

1. Ensure images work correctly throughout the application

## Checklist


packages/peregrine/lib/talons/Image/useResourceImage.js
```diff
@@ -40,38 +40,44 @@ export const useResourceImage = props => {
...
         };
 
-        // Note: it is assumed sizes will be filled from small, to medium, to large.
-        // In other words, having values for small and large but not medium is not supported.
-        switch (resourceSizeBreakpoints.size) {
-            case 2: {
-                const small = constrain('small');
-                const medium = constrain('medium');
-                const large = getPixelSize('large');
+        // The number of breakpoints must be one less than the number of sizes.
```
supernova-at: There are a few "hidden" assumptions within this PR:

1. The number of breakpoints will be one less than the number of widths

There needs to be one width that is unconstrained. In other words, one that doesn't have a media query.

**Bad**:
```js
const breakpoints = [100, 200];
const widths = [50, 150];

// sizes = "(max-width: 100px) 50px, (max-width: 200px) 150px"
```

This is not a valid value for `sizes`.

**Good**:
```js
const breakpoints = [100];
const widths = [50, 150];

// sizes = "(max-width: 100px) 50px, 150px"
```

❓ Should I add logic to enforce this constraint? Basically, only use `widths.length - 1` `widthBreakpoints`?

2. The values in `widths` and `widthBreakpoints` will be sorted from small to large

This is because we hardcode `max-width` for the `sizes` media queries.

**Bad**:
```js
const breakpoints = [200, 100];
const widths = [250, 350, 150];

// sizes = "(max-width: 200px) 250px, (max-width: 100px) 350px, 150px".
```

This is a valid value for `sizes` but isn't going to be what the developer expects. In English this reads: "up to 100px, use the 250px-wide image. Between 100 and 200px, use the 350px-wide image. Beyond 200px, use the 150px image". 

**Good**:
```js
const breakpoints = [100, 200];
const widths = [150, 250, 350];

// sizes = "(max-width: 100px) 150px, (max-width: 200px) 250px, 350px".
```

❓  Should we sort them if they're not already sorted?

3. The entries in widths and breakpoints "match up"

That is, `breakpoint[0]` is for `widths[0]`.

```js
const breakpoints = [100];
const widths = [150, 250];

// sizes = "(max-width: 100px) 150px, 250px"
```

sirugh: > Should I add logic to enforce this constraint?

What about an `unconstrainedWidth` or some other name representing it? 

> The values in widths and widthBreakpoints will be sorted from small to large

They *should* be sorted right? I'm starting to think an array might be the wrong structure. Either that or we would have to add some prop validation for it.

revanth0212:
@supernova-at there are prettier fails. Check on that once. I am not sure how did it miss the git commit hook when you pushed them.

packages/peregrine/lib/talons/Image/useImage.js
```diff
@@ -34,18 +32,14 @@ export const useImage = props => {
         }
     }, [onError]);
 
-    // If we don't have a resourceWidth, use the smallest resource size.
+    // Use the smallest entry in widths.
     const resourceWidth = useMemo(() => {
-        if (propResourceWidth) {
-            return propResourceWidth;
+        if (!widths) {
+            return undefined;
```
revanth0212: Can this not be any default value returned if widths is not provided?

packages/peregrine/lib/talons/Image/usePlaceholderImage.js
```diff
@@ -1,3 +1,11 @@
+/**
```
revanth0212: Nice job.

packages/peregrine/lib/talons/Image/useResourceImage.js
revanth0212: > This is a valid value for `sizes` but isn't going to be what the developer expects. In English this reads: "up to 100px, use the 250px-wide image. Between 100 and 200px, use the 350px-wide image. Beyond 200px, use the 150px image".
> 
> ❓ Should we sort them if they're not already sorted?

I don't think we should sort them. We should leave this to the developer because 

```es6
const breakpoints = [200, 100];
const widths = [250, 350, 150];

// sizes = "(max-width: 200px) 250px, (max-width: 100px) 350px, 150px".
```

is a valid use case.

Potential Usecase:

In a desktop view, the product image should be shown on the left with details on right. Which means 50% of the width.

In a mobile view, the product image should take up the whole width. Which means 100% of the width.

This means, as the width of the viewport goes down, the size of the image should go up, hence it should download a high res image.

If we sort the `widths` array, this won't be possible.

This is replicable in Venia's product page as well:

Desktop View:
image


iPhone View:
image

packages/peregrine/lib/talons/Image/useImage.js
sirugh: It is odd that this would be `undefined`. I would assume that the components that care about this require it, right? We definitely don't want `resourceImage` throwing when it attempts to get `widths[index]` [here](https://github.com/magento/pwa-studio/blob/44965d25147980b97308a267b4539b0c7ba42015/packages/peregrine/lib/talons/Image/useResourceImage.js#L44).

Maybe instead of having the `useImage` talon do this work you could just pass the `widths` array to the components that use it and have them do this work. We're already talking about potentially splitting them anyways so that components can directly import resource or simple image rather than `Image`.

jimbo:
> However, I don't think an array is the right structure for maintaining the links between widths and breakpoints. I think a map fits this use case perfectly and would really simplify the logic in the `sizes` memo in `useResourceImage`.

Good suggestion. 👍

supernova-at:
On it! 🖖

packages/peregrine/lib/talons/Image/useResourceImage.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1956/commits/fa254e23d670ce5fa1f3e2d1aac25ccdfaff2562.

supernova-at:
PR updated with new `Map` structure.

I named the prop `widths` instead of `sizes` to avoid confusion with the actual `sizes` attribute of an image (which we generate here but might be passed in by a caller).

packages/peregrine/lib/talons/Image/useImage.js
supernova-at: This is a little shortcut if the caller doesn't pass a `width` but does pass `widths`.

If they don't pass either one, we literally don't have any information to go off of - `undefined` seemed the best choice here. This will result in the `img` element rendered to the DOM to not have a `width` attribute and the network request to fetch this image won't include resize parameters.

supernova-at: I realized it wasn't preferring the passed-in `width` so I updated that in https://github.com/magento/pwa-studio/pull/1956/commits/4e798f71e9c0329caae579550104cb0ae400ea88.

packages/peregrine/lib/talons/Image/useImage.js
```diff
@@ -5,16 +5,12 @@ import { useCallback, useMemo, useState } from 'react';
  *
  * @param {function} props.onError callback for error of loading image
  * @param {function} props.onLoad callback for load of image
- * @param {Map}      props.resourceSizes image sizes used by the browser to select the image source. Supported keys are 'small', 'medium', and 'large'.
- * @param {number}   props.resourceWidth the intrinsic width of the image & the width to request for the fallback image for browsers that don't support srcset / sizes.
+ * @param {string}   props.unconstrainedSizeKey the key in props.widths for the unconstrained / default width.
```
jimbo: This is pretty weird. Is there a reason why we can't just let `"default"` be the key here?

supernova-at: It is always "default", I just didn't want every file to have to know that.

I agree it is weird though.

I could go either way on it.

supernova-at: The other way I thought of doing it was `import`ing the key but then I ran into circular dependencies.

supernova-at: Yep, way better: https://github.com/magento/pwa-studio/pull/1956/commits/14fca612673a17a2ec3cd795505697de2a6fbb2d.

supernova-at:
PR Updated: 
  * No longer threads the `UNCONSTRAINED_SIZE_KEY`
  * Simplifies most common use case: have a `width` but no `widths`. `sizes` will just use the given `width` as the unconstrained size. Ex: `width=100 => sizes="100px".

packages/peregrine/lib/talons/Image/useImage.js
jimbo: > The other way I thought of doing it was `import`ing the key but then I ran into circular dependencies and Peregrine talons needing to `import` Venia UI files.

Right, the dependency should flow in one direction; we shouldn't have any situations where `peregrine` imports from `venia-ui`.

jimbo:
> PR Updated:
> 
>     * No longer threads the `UNCONSTRAINED_SIZE_KEY`
> 
>     * Simplifies most common use case: have a `width` but no `widths`. `sizes` will just use the given `width` as the unconstrained size. Ex: `width=100 => sizes="100px"`.

Excellent. 💯

PR #1960, opened 2019-11-05
[feature]: product page breadcrumbs

sirugh:
## Description

Based on #1949. This PR adds breadcrumbs to the product page. Main differences:
* Unlike the category page breadcrumbs, the last category text in the list is a link and will be underlined/clickable.
* If a product has only one related category it will display a single breadcrumb to that category
* ~~If a product has multiple categories it will display a dropdown with multiple options.~~ 

## Related Issue

Closes [PWA-99](https://jira.corp.magento.com/browse/PWA-99).

### Verification Stakeholders
@jimbo @schensley 

- Should we even display PDP breadcrumbs on mobile?
- Is a dropdown the right UX?

### Verification Steps

1. Go to a product with only one related category (isadora skirt). It should show regular breadcrumbs like on the category page.
2. Go to a product with multiple categories (valeria two-layer tank). It should also show regular breadcrumbs.

[Image from Gyazo](https://gyazo.com/dccb657882ff33b822e6de254062811b)

[Image from Gyazo](https://gyazo.com/e16a7a636f68abc4bb5c764f8d8b8881)

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
+++ b/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
@@ -34,7 +34,9 @@ export const useBreadcrumbs = props => {
         variables: { category_id: categoryId }
     });
 
-    const categoryUrlSuffix = data.storeConfig.category_url_suffix;
+    // Default to .html for when the query has not yet returned.
+    const categoryUrlSuffix =
+        (data && data.storeConfig.category_url_suffix) || '.html';
 
     // When we have breadcrumb data sort and normalize it for easy rendering.
     const normalizedData = useMemo(() => {
--- a/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
+++ b/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
@@ -34,8 +34,7 @@ export const useBreadcrumbs = props => {
         variables: { category_id: categoryId }
     });
 
-    const categoryUrlSuffix =
-        (data && data.storeConfig.category_url_suffix) || '.html';
+    const categoryUrlSuffix = data.storeConfig.category_url_suffix;
 
     // When we have breadcrumb data sort and normalize it for easy rendering.
     const normalizedData = useMemo(() => {
@@ -55,7 +54,7 @@ export const useBreadcrumbs = props => {
     return {
         currentCategory: (data && data.category.name) || '',
         currentCategoryPath:
-            data && `${data.category.url_path}${categoryUrlSuffix}`,
+            (data && `${data.category.url_path}${categoryUrlSuffix}`) || '#',
         isLoading: loading,
         normalizedData: normalizedData || []
     };
--- a/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
+++ b/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
@@ -30,7 +30,7 @@ const getPath = (path, suffix) => {
 export const useBreadcrumbs = props => {
     const { categoryId, query } = props;
 
-    const { data, loading } = useQuery(query, {
+    const { data, loading, error } = useQuery(query, {
         variables: { category_id: categoryId }
     });
 
@@ -58,6 +58,7 @@ export const useBreadcrumbs = props => {
         currentCategoryPath:
             (data && `${data.category.url_path}${categoryUrlSuffix}`) || '#',
         isLoading: loading,
+        hasError: !!error,
         normalizedData: normalizedData || []
     };
 };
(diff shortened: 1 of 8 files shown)


sirugh:
I just pushed some big changes to this PR. Rather than have a dropdown for breadcrumbs when there were multiple possible categories on a PDP I now:

1) Request categories for the product
2) Filter down the category breadcrumbs to a list of _only_ leaf categories since the API gives us all the intermediate categories.
**3) Select the first category id rather than the last.**

Note that the method of selection of which category breadcrumbs to display can cause a jarring display if a user navigated to the PDP by one route and are then shown a different breadcrumb set. See the gifs in the description.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -90,6 +90,26 @@ const getMediaGalleryEntries = (product, optionCodes, optionSelections) => {
...
+    });
+
+    // Filter out category data for categories that exist as breadcrumbs.
+    const leafCategories = categories.filter(
+        category => !breadcrumbSet.has(category.id)
+    );
+
+    // Return the first category id of the potential leaf categories.
+    // This can potentially be jarring if the user navigated to a product and
+    // the breadcrumbs don't represent that navigation path.
```
jimbo: I don't think it'll be jarring at all. Breadcrumbs should be a canonical path to the product as understood by the server (or in this case, estimated by the client using a simple heuristic: first category wins). We don't need to be in the business of showing the user their navigation history. That's what browser history and browser chrome are for.

At any rate, I like this a lot more than the dropdown.

sirugh: Well I certainly don't want to lead opinion through this comment -- I can remove it and simply say "Until we can get the single canonical breadcrumb path to a product we will just select the first leaf category"

packages/venia-ui/lib/fragments/productDetails.graphql
```diff
@@ -1,5 +1,11 @@
 fragment ProductDetails on ProductInterface {
     __typename
+    categories {
```
sirugh: As an FYI we should add a comment in here at the top saying that updates to this fragment should be mirrored in `getProductDetail` and vice versa.

packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.js
```diff
@@ -45,47 +47,55 @@ const ProductFullDetail = props => {
...
-                <h1 className={classes.productName}>{productDetails.name}</h1>
-                <p className={classes.productPrice}>
-                    <Price
-                        currencyCode={productDetails.price.currency}
-                        value={productDetails.price.value}
+        <Fragment>
+            <Breadcrumbs
+                categoryId={breadcrumbCategoryId}
+                currentProduct={productDetails.name}
+            />
```
sirugh: This and the returned prop are the only changes. Everything just shifted cause of the `<Fragment>`.

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
```diff
@@ -29,11 +29,14 @@ export const useBreadcrumbs = props => {
         variables: { category_id: categoryId }
     });
 
+    const categoryUrlSuffix =
```
tjwiebell: This default value seems unreachable/unused, do we need it?

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
```diff
@@ -20,7 +20,7 @@ const getPath = (path, suffix) => {
  * @param {object} props
  * @param {object} props.query - the breadcrumb query
  * @param {string} props.categoryId - the id of the category for which to generate breadcrumbs
- * @return {{ currentCategory: string, isLoading: boolean, normalizedData: array }}
+ * @return {{ currentCategory: string, currentCategoryPath: string, isLoading: boolean, normalizedData: array }}
```
tjwiebell: Pushing the line length limit, should probably break out for readability.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -90,13 +90,40 @@ const getMediaGalleryEntries = (product, optionCodes, optionSelections) => {
...
+    categories.forEach(({ breadcrumbs }) => {
+        // breadcrumbs can be `null`...
+        (breadcrumbs || []).forEach(({ category_id }) =>
+            breadcrumbSet.add(category_id)
+        );
+    });
+
+    // Filter out category data for categories that exist as breadcrumbs.
+    const leafCategories = categories.filter(
+        category => !breadcrumbSet.has(category.id)
```
tjwiebell: Given we only care about the first item returned, we should use `categories.find` so we can bail early.

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
sirugh: If `category_url_suffix` is undefined wouldn't we want a fallback?

tjwiebell: The return type is `String`, so that shouldn't be possible. I don't think we need to guard against GraphQL returning partial results, since we're using the default error handling policy of `none`.

> none: This is the default policy to match how Apollo Client 1.0 worked. Any GraphQL Errors are treated the same as network errors and any data is ignored from the response.

PR #1965, opened 2019-11-07
Tests for Service Worker

### LINKED ISSUE #1789: [misc]: Tests for sw.js
Related to #1673 

Write tests for sw.js. Now that a lot of things will be changing relative to current sw.js, having tests will help.

- [x] `venia-concept`
- [ ] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

revanth0212:
## Description

Added tests for the Service Worker.

## Related Issue
Closes #1789 

### Verification Steps
1. Run `yarn test` and `yarn test:ci`. You should not see any errors.

image

## Checklist
None


packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    global.Request = function(url, { method, headers } = {}) {
+        this.url = url;
+        this.method = method;
+        this.headers = headers;
+    };
+    global.Response = function() {};
+});
+
+beforeEach(() => {
+    sendMessageToWindow.mockClear();
```
supernova-at: Is `sendMessageToWindow` a mocked function here? Where do we mock it?

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -0,0 +1,141 @@
...
+});
+
+test('cacheHTMLPlugin should have cacheKeyWillBeUsed, requestWillFetch, fetchDidSucceed async functions implemented', () => {
+    matchFn.mockReturnValue(Promise.resolve(null));
+
+    expect(cacheHTMLPlugin).toHaveProperty('cacheKeyWillBeUsed');
+    expect(cacheHTMLPlugin).toHaveProperty('requestWillFetch');
+    expect(cacheHTMLPlugin).toHaveProperty('fetchDidSucceed');
```
supernova-at: Just to save yourself some time in the future, these aren't _strictly_ necessary because your next set of assertions will fail if these properties aren't present.

But it reads better like this, so it's likely a personal preference thing.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    const response = await cacheHTMLPlugin.cacheKeyWillBeUsed();
+
+    expect(response).toBe('/');
+});
+
+test('requestWillFetch should return a new Request with url set to /', async () => {
+    const request = new Request('https://develop.pwa-venia.com/');
+    const response = await cacheHTMLPlugin.requestWillFetch({ request });
+
+    expect(response.url).not.toBe(request.url);
```
supernova-at: Assertion doesn't match test description.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    global.Response.prototype.text = function() {
+        return Promise.resolve(
+            `Hey I am random text ${Math.random()
+                .toString()
+                .slice(2)}`
+        );
+    };
+    global.Response.prototype.clone = function() {
+        return new Response();
+    };
```
supernova-at: Don't forget to return these to their original states in case other tests rely on them.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    matchFn.mockReturnValue(Promise.resolve(new Response()));
+
+    const url = 'https://develop.pwa-venia.com/';
+    const request = new Request(url);
+    const response = new Response();
+
+    await cacheHTMLPlugin.fetchDidSucceed({ request, response });
+
+    expect(sendMessageToWindow).not.toHaveBeenCalled();
+});
```
supernova-at: Great mocks & tests!

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
```diff
@@ -0,0 +1,323 @@
+import {
+    __handlers__,
```
supernova-at: It's usually a ⚠️ to see private variables like this imported in a test.

Sometimes the solution is as easy as "don't test that internal piece" / the tests are too granular, but other times you may have to refactor the code to be more easily unit-testable.

supernova-at: In this case, you may want to add a `clearHandlers` (or similar) function on `messageHandler.js` itself and just call that from here? Depends if you think anyone else could benefit from such a function.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
```diff
@@ -0,0 +1,323 @@
...
+    });
+
+    test('registerMessageHandler should add a message handler for a given type in the handlers object', () => {
+        registerMessageHandler(messageType, handler);
+
+        expect(__handlers__).toHaveProperty(messageType);
+        expect(__handlers__[messageType]).toBeInstanceOf(Array);
+        expect(__handlers__[messageType]).toHaveLength(1);
+        expect(__handlers__[messageType]).toContain(handler);
+    });
```
supernova-at: This may be too granular. The unit tests should test the input / output of the module and its functions. If someone comes along tomorrow and optimizes how `__handlers__` works (or any internals of the module), that shouldn't cause unit tests to fail.

I've found that approach to be the most scalable.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
```diff
@@ -0,0 +1,323 @@
...
+            messageType,
+            handler
+        );
+
+        unregisterHelloWorld();
+
+        expect(__handlers__[messageType]).not.toContain(handler);
+    });
+
+    test("registerMessageHandler's return function should only unregister the function it registered", () => {
```
supernova-at: Great test

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
supernova-at: See if there's any other way you can test that the rest of these test cases succeed without having to reach into `__handlers__` directly.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
```diff
@@ -0,0 +1,323 @@
...
+        const handler1 = jest.fn();
+        const handler2 = jest.fn();
+
+        registerMessageHandler(messageType, handler1);
+        registerMessageHandler(messageType, handler2);
+
+        handleMessageFromClient(messageType, payload, event);
+
+        expect(handler1).toHaveBeenCalledWith(payload, event);
+        expect(handler2).toHaveBeenCalledWith(payload, event);
```
supernova-at: 💯

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/routeHandler.spec.js
```diff
@@ -0,0 +1,34 @@
...
+test("isHTMLRoute should return true if route's pathname has .html", () => {
+    expect(isHTMLRoute(new URL(urlWithHTML))).toBeTruthy();
+});
+
+test("isHTMLRoute should return false if route's pathname does not have .html", () => {
+    expect(isHTMLRoute(new URL(nonHTMLRoute))).toBeFalsy();
+});
+
+test('isHTMLRoute should return true for home route', () => {
+    expect(isHTMLRoute(new URL(homePageUrl))).toBeTruthy();
```
supernova-at: This test duplicates the previous "isHomeRoute should return true if route's pathname is /" test.

packages/venia-concept/src/ServiceWorker/Utilities/messageHandler.js
```diff
@@ -9,6 +9,11 @@
  */
 const handlers = {};
 
+/**
+ * Exporting for testing purposes
+ */
+export const __handlers__ = handlers;
```
supernova-at: Good comment. I'd love to be able to drop this, per other feedbacks.

packages/venia-concept/src/ServiceWorker/Utilities/messageHandler.js
```diff
@@ -109,7 +114,11 @@ export const sendMessageToClient = (client, type, payload) =>
             }
         };
 
-        client.postMessage({ type, payload }, [channel.port2]);
+        if (client && client.postMessage) {
+            client.postMessage({ type, payload }, [channel.port2]);
+        } else {
+            reject(`Unable to send message to ${client.type}`);
```
supernova-at: What happens if `client` is falsy?

packages/venia-concept/src/ServiceWorker/__tests__/registerRoutes.spec.js
```diff
@@ -0,0 +1,123 @@
...
+        expiration: {
+            Plugin: expirationPlugin
+        },
+        routing: {
+            registerRoute: function() {}
+        }
+    };
+});
+
+test("A total of 5 routes need to be registered using workbox's registerRoute API", () => {
```
supernova-at: I think we can probably drop this test.

We're testing the five routes in more depth below, and any time we add a new route we'd have to come in and update this test. It just doesn't add much value.

packages/venia-concept/src/ServiceWorker/__tests__/registerRoutes.spec.js
```diff
@@ -0,0 +1,123 @@
...
+    registerRoutes();
+
+    /**
+     * This is a crude way to find the route by searching for
+     * isHTMLRoute function in the .toString() output, but I am
+     * out of options at this point.
+     *
+     * Obviously there might be a better way to do it just that I
+     * am not aware of it at this point. Feel free to change it.
+     */
```
supernova-at: I think this is fine for now. What is `call[0]` itself?

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
revanth0212: In the first line, I mocked the whole `messageHandler.js` module.

revanth0212: On the other hand, we don't need that line anymore. Removing it.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
revanth0212: I needed tests that will fail if in the future someone accidentally changes them without them pondering about what happened. Also, I feel it should be a norm usually to have tests around testing the shape of an entity like snapshots do for react components.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
revanth0212: You are right. Modified it.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
revanth0212: Nice catch. Done.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
revanth0212: I modified the code to be able to test. Ideally `__handlers__` is supposed to be an `internal use only` variable but I had to change to make the tests possible. Without exposing `__handlers__` there is no way to test them easily. We will have to go around and mock a lot of stuff. I feel, as a developer, if someone sees `__VARIABLE_NAME__` they need to use caution.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
revanth0212: I don't think that is possible. As far as I have researched online, the only way to test such `internal only/private` stuff it to either expose it with `__` so others know the risk or use [rewire](https://www.npmjs.com/package/rewire). I am not sure if we want to add a new npm dependency to test this. That is the reason why I settled with the first option.

Also, I feel, tests need to be granular enough to fail when their shape changes, in this case, they need to fail if the handlers are not stored as an `Array` because other functions expect it to be an `Array`.

packages/venia-concept/src/ServiceWorker/Utilities/messageHandler.js
revanth0212: Had a chat with @zetlen a couple weeks ago where he was talking about modifying function behavior to make testing possible. Took his advice and made this change but I needed to let future developers know that `__handlers__` is strictly for testing only. But yeah I am up for suggestions.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/routeHandler.spec.js
revanth0212: Well, they are different functions. First is `isHomeRoute` second is `isHTMLRoute` which internally uses `isHomeRoute` because `/` is a home route and also a HTML route.

packages/venia-concept/src/ServiceWorker/Utilities/messageHandler.js
revanth0212: Nice catch. Changed it.

packages/venia-concept/src/ServiceWorker/__tests__/registerRoutes.spec.js
revanth0212: `call[0]` is the first argument, in this case the first argument is a function that uses `isHTMLRoute` function. I am trying to find that using a filter on all the `calls` in `registerRoute.mock`.

packages/venia-concept/src/ServiceWorker/__tests__/registerRoutes.spec.js
revanth0212: You are right but I added it to deliberately fail when a new route is added. @zetlen had a plan for a future feature that will auto-create these routes based on the app routes from `venia` in which case this test can be modified to suit that change and also fail if someone added a route signifying that they are changing and needs caution. But, I feel we should remove this test for not as it doesn't serve a lot of purpose.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
supernova-at: Yeah, I've used `rewire` before ... please, no 😅 

> Also, I feel, tests need to be granular enough to fail when their shape changes, in this case, they need to fail if the handlers are not stored as an Array because other functions expect it to be an Array.

If those other functions are internal to the module we don't have to explicitly care 😃 . The module will fail some test if has internal errors / inconsistencies.

My larger point is just that tests should rarely care about _how_ a module does its job, just that it does it. 

Tests for this module should be along the lines of "I registered multiple handlers, do they all get called when the event happens?". I think you already have these tests. If they work, we can assume the internal logic related to `__handlers__` is working properly.

Does that make sense?

packages/venia-concept/src/ServiceWorker/__tests__/registerRoutes.spec.js
revanth0212: I meant this be a snapshot test equivalent.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
supernova-at: Yeah I'm just saying the assertions immediately below this serve the same purpose.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/messageHandler.spec.js
revanth0212: Or I could do something like this:

Assume there is a test mode like `process.env.TEST_MODE` set to true. In which case we can expose `__handlers__` by setting it in the `global` scope. This way we won't be leaking `__handlers__` in non-test environments.

What do you think about this option?

revanth0212: > Does that make sense?

Yeah, it does. Will keep that in mind for future tests. Thanks.

PR #1966, opened 2019-11-08
Supplant MagentoRouteHandler

jimbo:
## Description

Now that we're using React Router `5.1.0`, which includes hooks and real context, we no longer need our customized router. Not only can we replace custom code with canonical implementations (similar to our adoption of Apollo hooks), we can eliminate render props at the base of the app, which should improve rendering performance across the board.

**Added**

- The new `Routes` component replaces `renderRoutes()`
- The new `MagentoRoute` component replaces `MagentoRouteHandler`
- The new `useMagentoRoute` talon replaces `MagentoRouteHandler`

**Changed**

- `VeniaAdapter` now renders `BrowserRouter` directly
- The Apollo `client` object now exposes `apiBase`

## Related Issue

[PWA-137](https://jira.corp.magento.com/browse/PWA-137)

## Acceptance 
This is a one-for-one refactor; just ensure there no regressions.

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Navigate to the home page
2. Navigate to a category page and product page
3. Navigate to a known nonexistent page (e.g., `/foo.html`)
4. Navigate to the search page

## Checklist
* I have **not** updated the documentation accordingly, if necessary.
* I have **not** added tests to cover my changes, if necessary.


packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -0,0 +1,45 @@
...
+
+    const routeData = componentMap.get(pathname);
+
+    // ask Magento for a RootComponent that matches the current pathname
+    useEffect(() => {
+        const isKnown = componentMap.has(pathname);
+        const isClientOnly = CLIENT_PATHS.has(pathname);
+
+        // avoid asking if we already know the answer
+        if (!isKnown && !isClientOnly) {
```
jimbo: In the current implementation, there's one more condition here, which is when [the known component has a certain `id`](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/Router/magentoRouteHandler.js#L61). Is it still relevant?

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -0,0 +1,45 @@
...
+            getRouteComponent(apiBase, pathname).then(
+                ({ component, id, pathname, routeError }) => {
+                    // add the pathname to the browser cache
+                    addToCache(pathname);
+
+                    // then add the pathname and result to local state
+                    setComponentMap(prevMap => {
```
jimbo: Probably need to avoid setting state here if the component has been unmounted.

packages/peregrine/lib/talons/MagentoRoute/addToCache.js
```diff
@@ -0,0 +1,15 @@
+const addToCache = async (...urls) => {
+    if (!window.caches) {
```
revanth0212: It is good to see caching but can we delegate this work to the SW? 2 reasons:

1. No main thread cycles wasted
2. `cacheName` can change in the future and having it hardcoded here can be a potential cause for bugs. We can leave this to the SW and if the `cacheName` changes in the future workbox will take care of it.

You can use [sendMessageToSw](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/util/swUtils.js#L90) to send a message along with the list of URL you want to cache and have a [handler](https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/src/ServiceWorker/Utilities/messageHandler.js#L27) on the SW do the caching part.

For instance:

On the UI:

```es6
sendMessageToSW('PREFETCH_URLS', { urls: ['https://develop.pwa-venia.com/main.css', 'https://develop.pwa-venia.com/font.woff'] })
```

In SW:

```es6
registerMessageHandler('PREFETCH_URLS', prefetchUrls)
```

Check out this example:

https://github.com/magento/pwa-studio/blob/21aadbc054e9d874f0bdaf86ee346e03801c5e09/packages/peregrine/lib/talons/ProductImageCarousel/useProductImageCarousel.js#L36

https://github.com/magento/pwa-studio/blob/21aadbc054e9d874f0bdaf86ee346e03801c5e09/packages/venia-concept/src/ServiceWorker/Utilities/imageCacheHandler.js#L144

packages/peregrine/lib/talons/MagentoRoute/getRouteComponent.js
```diff
@@ -0,0 +1,56 @@
...
+
+        // associate the matching RootComponent with this location
+        return {
+            component,
+            id,
+            pathname,
+            type
+        };
+    } catch (e) {
+        if (process.env.NODE_ENV === 'development') {
```
revanth0212: I don't think we need this anymore. In prod mode all `console.*` statements are omitted.

packages/venia-ui/lib/components/Routes/routes.js
```diff
@@ -0,0 +1,30 @@
+import React, { Suspense, lazy } from 'react';
+import { Route, Switch } from 'react-router-dom';
+
+import MagentoRoute from '../MagentoRoute';
+
+const CreateAccountPage = lazy(() => import('../CreateAccountPage'));
+const Search = lazy(() => import('../../RootComponents/Search'));
+
+const LOADING_MESSAGE = 'Just a moment...';
+const fallback = <h1>{LOADING_MESSAGE}</h1>;
```
revanth0212: Can we use the [LoadingIndicator](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/LoadingIndicator/indicator.js#L9) here? I might be wrong.

packages/venia-ui/lib/components/App/app.js
```diff
@@ -9,7 +9,8 @@ import Main from '../Main';
 import Mask from '../Mask';
 import MiniCart from '../MiniCart';
 import Navigation from '../Navigation';
-import renderRoutes from './renderRoutes';
+import Routes from '../Routes';
+// import renderRoutes from './renderRoutes';
```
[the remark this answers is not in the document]
jimbo: Fixed in [7a7ef14](https://github.com/magento/pwa-studio/pull/1966/commits/7a7ef14063db684a9103e63de7e826a78e08f1c3).

packages/venia-ui/lib/components/Routes/routes.js
jimbo: Fixed in [7a7ef14](https://github.com/magento/pwa-studio/pull/1966/commits/7a7ef14063db684a9103e63de7e826a78e08f1c3).

packages/peregrine/lib/talons/MagentoRoute/getRouteComponent.js
jimbo: Fixed in [7a7ef14](https://github.com/magento/pwa-studio/pull/1966/commits/7a7ef14063db684a9103e63de7e826a78e08f1c3).

packages/peregrine/lib/talons/MagentoRoute/addToCache.js
jimbo: Good feedback, I agree with you on all points. I'd like to handle the change in a separate PR, though. This one is just a 1:1 refactor—nothing about `addToCache` has changed here.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
jimbo: Fixed in [5882424](https://github.com/magento/pwa-studio/pull/1966/commits/5882424c9c17d2d66f1b8158e12405a26ab2a29a).

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
jimbo: Fixed in [5882424](https://github.com/magento/pwa-studio/pull/1966/commits/5882424c9c17d2d66f1b8158e12405a26ab2a29a).

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -0,0 +1,45 @@
...
+    useEffect(() => {
+        const isKnown = componentMap.has(pathname);
+        const isClientOnly = CLIENT_PATHS.has(pathname);
+
+        // avoid asking if we already know the answer
+        if (!isKnown && !isClientOnly) {
+            getRouteComponent(apiBase, pathname).then(
+                ({ component, id, pathname, routeError }) => {
+                    // add the pathname to the browser cache
+                    addToCache(pathname);
```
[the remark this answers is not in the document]
jimbo: I don't know. @revanth0212 pointed out that this appears to be an outdated way of getting the service worker to participate, and that we have a newer approach we should use. All I did here was move the original approach out of lifecycle methods and into `useEffect` calls.

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -0,0 +1,25 @@
...
+
+const MagentoRoute = () => {
+    const talonProps = useMagentoRoute();
+    const { component: RootComponent, id, isLoading, routeError } = talonProps;
+
+    if (isLoading) {
+        return fullPageLoadingIndicator;
+    } else if (RootComponent) {
+        return <RootComponent id={id} />;
+    } else if (routeError === 'NOT_FOUND') {
```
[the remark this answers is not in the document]
jimbo: Fixed in [701f7bf](https://github.com/magento/pwa-studio/pull/1966/commits/701f7bf96a3d3575fe55716c9e539d7fd7deb3d9).

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -0,0 +1,63 @@
+import { useEffect, useRef, useState } from 'react';
+import { useLocation } from 'react-router-dom';
+import { useApolloClient } from '@apollo/react-hooks';
+
+import addToCache from './addToCache';
+import getRouteComponent from './getRouteComponent';
+
+const CLIENT_PATHS = new Set().add('/search.html');
```
supernova-at: One last thing, it feels weird that the list of client-only routes is kept here and in `Routes`.

```jsx
<Suspense fallback={fullPageLoadingIndicator}>
    <Switch>
        <Route exact path="/search.html">
            <Search />
        </Route>
        <Route exact path="/create-account">
            <CreateAccountPage />
        </Route>
        <Route>
            <MagentoRoute />
        </Route>
    </Switch>
</Suspense>
```

Don't we know by the fact we made it down to `MagentoRoute` that we're not a client-only route?

supernova-at: Also, if `/search.html` is here, should `/create-account` be too?

jimbo: I agree. Of course, I'm not the one who carved out the exception for `search.html` here, so I don't know why that clause is here when it appears to be unreachable. Worth investigating.

packages/venia-ui/lib/RootComponents/Search/search.js
```diff
@@ -1 +0,0 @@
-export { default } from '../../components/SearchPage';
```
tjwiebell: I thought this felt redundant 👍

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
jimbo: @supernova-at Looks like you were the one who added the clause for `/search.html` to `MagentoRouteHandler` in #639, almost a year ago! In fact, [here's what I had to say about it](https://github.com/magento/pwa-studio/pull/639/files#r241900060) at the time: 😅 

> This pattern doesn't seem like it will scale well. Are we going to add a new condition for every client-side route?

I also found another [fun comment I left](https://github.com/magento/pwa-studio/pull/639#issuecomment-449088676) on that PR.

> Now that React Router uses the new context API, our context work in Router is probably redundant. The only additional property being passed by our context is `apiBase`, which could probably be an environment variable or something.

Fixed!

jimbo: Also, fixed in [cabd8ec](https://github.com/magento/pwa-studio/pull/1966/commits/cabd8ecc94dab18501fde1ea7d722e6355b54c1c).

packages/peregrine/lib/talons/MagentoRoute/addToCache.js
zetlen: The `addToCache` implementation is different here, though. The original implementation had some big problems, so I'm not about bringing it back, but I think this isn't populating the cache correctly. Can you confirm that it's actually caching and reusing the `urlResolver` responses? I think this is just caching an app shell instead of a GraphQL response.

packages/peregrine/lib/talons/MagentoRoute/getRouteComponent.js
```diff
@@ -0,0 +1,57 @@
+import resolveUnknownRoute from '../../Router/resolveUnknownRoute';
+
+export const INTERNAL_ERROR = 'INTERNAL_ERROR';
+export const NOT_FOUND = 'NOT_FOUND';
+
+const getRouteComponent = async (apiBase, pathname) => {
+    // At build time, `fetchRootComponent` is injected as a global.
+    // Depending on the environment, this global will be either an
+    // ES module with a `default` property, or a plain CJS module.
```
zetlen: Very clear, thank you.

packages/peregrine/lib/talons/MagentoRoute/getRouteComponent.js
```diff
@@ -0,0 +1,57 @@
...
+        return {
+            component,
+            id,
+            pathname,
+            type
+        };
+    } catch (e) {
+        const routeError = e.message === '404' ? NOT_FOUND : INTERNAL_ERROR;
+
+        console.error(e);
```
zetlen: Thank you for this, too.

sirugh:
> Expected: It reuses the cached route resolution and doesn't call /graphql again with urlResolver.

I'm getting the expected outcome. Can someone else check this as well?

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -60,13 +60,15 @@ const VeniaAdapter = props => {
...
 
+        client.apiBase = apiBase;
+
         return client;
     }, [apiBase, apollo]);
 
     return (
         <ApolloProvider client={apolloClient}>
             <ReduxProvider store={store}>
-                <Router apiBase={apiBase}>{children}</Router>
```
sirugh: Just to clarify we are intentionally not removing the unused `Router` implementation right? We also have some old devdocs for it so I want to make sure we're not leaving outdated references lying around.

packages/venia-ui/lib/components/App/app.js
```diff
@@ -9,7 +9,7 @@ import Main from '../Main';
 import Mask from '../Mask';
 import MiniCart from '../MiniCart';
 import Navigation from '../Navigation';
-import renderRoutes from './renderRoutes';
```
sirugh: Same question here -- we're not deleting the old stuff on purpose, right?

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -0,0 +1,61 @@
...
+        if (!isKnown || shouldReload) {
+            getRouteComponent(apiBase, pathname).then(
+                ({ component, id, pathname, routeError }) => {
+                    // avoid setting state if unmounted
+                    if (!isMountedRef.current) {
+                        return;
+                    }
+
+                    // add the pathname to the browser cache
+                    addToCache(pathname);
```
sirugh: In the old implementation we had a `catch` that logged errors in development. Can we add that here? I know we're probably changing this soon so nbd either way.

jimbo: I believe that is [here](https://github.com/magento/pwa-studio/pull/1966/files#diff-062ef3510d7860a80e69ce915563639aR49) still. One of the other reviewers suggested that I remove the `process.env` check since it was redundant, so I did that.

packages/venia-ui/lib/drivers/adapter.js
jimbo: Correct. I didn't delete any of the old implementations since it wasn't strictly necessary to achieve the goal here. So we have a choice:

- Remove unused files, such as `Router` and `MagentoRouteHandler`
- Deprecate unused files, but remove them later on

packages/peregrine/lib/talons/MagentoRoute/addToCache.js
jimbo: Call me crazy, but did I do something other than just move [this code](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/Router/magentoRouteHandler.js#L32-L42) into its own file?

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
sirugh: Cool, didn't realize we eliminated consoles in prod.

Issue #1885, opened 2019-10-12
[bug]: 4.0.0 update, FetchError: request to graphql failed, reason: unable to verify the first certificate


brendanfalkowski:
**Describe the bug**
After updating to PWA Studio 4.0.0, `yarn run build` fails when verifying the certificate for the GraphQL endpoint.

**To reproduce**
1. Updated existing project to PWA Studio 4.0.0
2. `yarn install`
3. `yarn run build`

```
yarn run v1.17.3
$ yarn ce run build
$ yarn workspace @watscomagento/ce run build
$ yarn run build:prod
$ yarn run clean && yarn run load-env && webpack --no-progress --env.mode production
$ rimraf dist
$ node ../pwa-buildpack/bin/buildpack load-env --core-dev-mode .
{ FetchError: request to https://ce.test/graphql failed, reason: unable to verify the first certificate
    at ClientRequest.<anonymous> (/Users/brendan/work/clients/ce/carrierenterprise.com_m2/repo-pwa/packages/pwa-buildpack/node_modules/node-fetch/lib/index.js:1444:11)
    at ClientRequest.emit (events.js:198:13)
    at TLSSocket.socketErrorListener (_http_client.js:392:9)
    at TLSSocket.emit (events.js:198:13)
    at emitErrorNT (internal/streams/destroy.js:91:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:59:3)
    at process._tickCallback (internal/process/next_tick.js:63:19)
  message:
   'request to https://ce.test/graphql failed, reason: unable to verify the first certificate',
  type: 'system',
  errno: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE',
  code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE' }
FetchError: request to https://ce.test/graphql failed, reason: unable to verify the first certificate
    at ClientRequest.<anonymous> (/Users/brendan/work/clients/ce/carrierenterprise.com_m2/repo-pwa/packages/pwa-buildpack/node_modules/node-fetch/lib/index.js:1444:11)
    at ClientRequest.emit (events.js:198:13)
    at TLSSocket.socketErrorListener (_http_client.js:392:9)
    at TLSSocket.emit (events.js:198:13)
    at emitErrorNT (internal/streams/destroy.js:91:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:59:3)
    at process._tickCallback (internal/process/next_tick.js:63:19)
[...]

PR #1969, opened 2019-11-12
PWA-127: Suppress GraphQL Certificate Errors

tjwiebell:
## Description

Allow self-signed certificates when running GraphQL query validation scripts at build time.

## Related Issue

[PWA-127](https://jira.corp.magento.com/browse/PWA-127) 4.0.0 update, FetchError: request to graphql failed, reason: unable to verify the first certificate

### Verification Steps

1. Spin up Magento backend with self-signed or invalid SSL certificate
2. Update `.env` with this backend
3. `yarn build`
4. Verify certificate is no longer output and build completes
## Checklist

### DIFF
--- a/packages/pwa-buildpack/lib/Utilities/graphQL.js
+++ b/packages/pwa-buildpack/lib/Utilities/graphQL.js
@@ -1,5 +1,9 @@
 const fetch = require('node-fetch');
 const graphQLQueries = require('../queries');
+const { Agent: HTTPSAgent } = require('https');
+
+// To be used with `node-fetch` in order to allow self-signed certificates.
+const fetchAgent = new HTTPSAgent({ rejectUnauthorized: false });
 
 const fetchQuery = query => {
     return fetch(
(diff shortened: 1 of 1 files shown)


packages/pwa-buildpack/lib/Utilities/graphQL.js
```diff
@@ -10,7 +14,8 @@ const fetchQuery = query => {
                 'Content-Type': 'application/json',
                 'Accept-Encoding': 'gzip'
             },
-            body: JSON.stringify({ query })
+            body: JSON.stringify({ query }),
+            agent: fetchAgent
```
sirugh: This [SO post](https://stackoverflow.com/questions/31861109/tls-what-exactly-does-rejectunauthorized-mean-for-me) suggests that the better solution is to use a custom CA. This may not be as easy for people who might want to use PWA Studio directly out of the box without much customization. What do you think?

tjwiebell: There's a comment thread in the original issue to research/implement that solution as a followup, I fully agree. Woes of splitting off to JIRA.

https://github.com/magento/pwa-studio/issues/1885#issuecomment-552658990

sirugh: Seems like we should make a ticket for that in JIRA. Regardless I'll approve it :)

tjwiebell: 😎 
https://jira.corp.magento.com/browse/PWA-158

PR #1971, opened 2019-11-13
Lazy Load Category Filters

supernova-at:
## Description

The category page's load performance looks like:

Screen Shot 2019-11-13 at 9 18 58 AM

As you can see, over half the time spent rendering `Category` is spent rendering the `FilterModal`. There weren't any major optimizations to make therein (it just takes time), so instead this PR lazy loads the `FilterModal` ~~when the user hovers, focuses, or clicks the "Filter" button~~:

Screen Shot 2019-11-13 at 11 19 59 AM

The new performance profile looks like this:

Screen Shot 2019-11-13 at 9 43 29 AM

Notice that `Category` is finished rendering much quicker now, and `FilterModal` is lazy loaded in after the fact.

~~Bonus: if the user never interacts with the "Filters" button, we never download the `FilterModal` chunk 🎉~~
Note: the above is still supported, but Venia defaults to loading the chunk immediately instead of waiting for hover / focus / click interactions. Developers who want this functionality can easily add it with a one-line change (it is documented how to do so in the code itself).

## Related Issue

PWA-55.

### Verification Stakeholders

@jimbo @sirugh @tjwiebell @revanth0212 

### Verification Steps

1. Go to a category page in Venia
2. Open your browser Network tab
3. Hover over, focus, or click the "Filter" button
4. See that then and only then are the chunks for `FilterModal` downloaded (you can inspect their contents to see what components are loaded).

## Checklist


packages/venia-ui/lib/RootComponents/Category/categoryContent.js
```diff
@@ -58,7 +71,7 @@ const CategoryContent = props => {
                 <div className={classes.pagination}>
                     <Pagination pageControl={pageControl} />
                 </div>
-                {modal}
+                <Suspense fallback={null}>{modal}</Suspense>
```
sirugh: On a slow network this can cause a weird UX where we display the overlay until the modal component is fully loaded and then we render it in. In the below gif you can see I hover/click quickly and but the chunks are slow to load and so it looks off.

I would suggest that we might create some sort of skeleton drawer/modal that we can render until then?


[Image from Gyazo](https://gyazo.com/fb22c6a706bcca509c7027fabd32172d)

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
```diff
@@ -1,28 +1,37 @@
...
 
     const handleOpenFilters = useCallback(() => {
+        setLoadFilters(true);
         toggleDrawer('filter');
-    }, [toggleDrawer]);
+    }, [setLoadFilters, toggleDrawer]);
+
+    const handleLoadFilters = useCallback(() => {
+        setLoadFilters(true);
+    }, [setLoadFilters]);
```
sirugh: At some point we should add a talon for this component... I wonder why we didn't in the initial work :o

If you want to do that in this PR, that'd be awesome, but I wouldn't hold up this PR on it.

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
supernova-at: Great call. I actually did go down this path but found it to be much more more than anticipated because the `FilterModal` itself contains all the styles for rendering the drawer / modal (white background, animation in, etc.).

We could duplicate them but that could cause maintenance headaches keeping them in sync.
I also tried lazy loading the _contents_ of the modal, but by that time we've already eaten most of the cost and I didn't think it was worth it.

It would be much easier to do if we had `LeftDrawer` / `RightDrawer` components that we completely replaced the children of.

Right now we're injecting into new modals / drawers each time:

**Before "Filter" Button Click**:
Screen Shot 2019-11-14 at 3 02 44 PM

**After "Filter" Button Click**:
Screen Shot 2019-11-14 at 3 02 57 PM

I agree the experience is not good - I'm in the middle of duplicating the minimum set of "modal" styles now, as re-writing the drawer system is too large / out of scope for this imo.

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
supernova-at: I can do it here, np 👍

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
sirugh: A drawer placeholder that can slide in from left or right would be a good first step!

brendanfalkowski:
Kool-Aid_Man_Entrance.gif ➔ I know Venia is a fashion site simulation, but just chiming in because I mainly work on sites where filtering is the primary interaction on product listings (after scrolling, scrolling is always #1 regardless). It would be nicer if the plumbing existed to artfully defer loading filters but do it, rather than wait till hover/select. 

I'd prefer wasting some bandwidth to have the page's UI areas ready if it meant there was no lag when the user interacts with them. I think there's more patience to see skeletons filled in on first render than every tap resulting in a new skeleton.

This is a great case of the way Venia is designed today can incur cost to implementors needing to maintain a standard customization request. This is something I'll need to undo for spec-driven listings where filters are always visible on wide screens (no offscreen drawer), and then maybe support again when the screen is narrow.

I guess I'm saying "design is always baked into application logic" and I feel that happening here. I don't have a smart default that's not more complicated with conditions.

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1971/commits/a3ffd929180686bb97cc14f4e3f67a131ebc83fe.

sirugh:
> Disclaimer: we get product data and filters from an API that serves both in one payload, so we don't really have the option of splitting up delivery.

Good point, though it doesn't necessarily _have_ to be like that. With graphql we should be able to request filters separately and within the `Filter` component rather than have the Product fetch the filters and pass them. This would be a similar workflow to how Breadcrumbs now work where we pass only the category Id to the component and it does the work of fetching the data for rendering.

supernova-at:
> I think there's more patience to see skeletons filled in on first render than every tap resulting in a new skeleton.

Totally, thanks for chiming in!

> It would be nicer if the plumbing existed to artfully defer loading filters but do it, rather than wait till hover/select.

Talked it over with the team, I'll do this 👍 .  The plumbing will still be there for developers who do want to defer the loading until user interactions.

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
supernova-at: Okay, https://github.com/magento/pwa-studio/pull/1971/commits/a0228590afc1452be253e23f3681dfbb362f38ed loads the filters immediately so supporting the drawer placeholder / skeleton becomes less of an issue right now.

I'll create a separate ticket for re-working the drawers.

supernova-at:
As per recent commit https://github.com/magento/pwa-studio/pull/1971/commits/a0228590afc1452be253e23f3681dfbb362f38ed, Venia still splits the filters out into their own chunk but loads that chunk immediately instead of waiting for user interaction on the "Filter" button.

This still improves the time to interactive on the category page and doesn't negatively impact the user experience for the common interaction of filtering.

jimbo:
@brendanfalkowski 

> I'd prefer wasting some bandwidth to have the page's UI areas ready if it meant there was no lag when the user interacts with them.

I agree with you; this is definitely the goal, and we can continue to iterate toward it. We actually have quite a bit of control over the deferred loading flow. A PWA should defer the loading of as many things as possible, but we should do a bit more prefetching on render and a bit less loading on demand.

PR #1975, opened 2019-11-14
Update backend value in create-pwa package

tjwiebell:
## Description

Simple update of backing instance value suggested during `create-pwa` flow. 

## Related Issue

* [[PWA-95](https://jira.corp.magento.com/browse/PWA-95)] Clean up pwa-buildpack/sampleBackends.json backend reference url

### Verification Steps

I had some issues getting the install to complete with local init package that I didn't encounter using `npm init @magento/pwa`, but I'm confident I'm just missing a step. I am able to confirm the value is updated, but could use a second look.
1. `npm install -g /Users/twiebell/Projects/pwa-studio/packages/create-pwa`
2. From a new test directory: `npm init pwa`
3. Step through prompts and select `2.3.3-venia-cloud` backend option
4. Observe updated value in `buildpack create-project` command
5. I would suggest uninstalling when done: `npm uninstall -g @magento/create-pwa`

## Checklist

### DIFF
--- a/packages/pwa-buildpack/sampleBackends.json
+++ b/packages/pwa-buildpack/sampleBackends.json
@@ -1,9 +1,9 @@
 {
     "environments": [
         {
-            "name": "2.3.1-venia-cloud",
-            "description": "Magento 2.3.1 with Venia sample data installed",
-            "url": "https://release-dev-231-npzdaky-zddsyhrdimyra.us-4.magentosite.cloud/"
+            "name": "2.3.3-venia-cloud",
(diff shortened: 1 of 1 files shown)


tjwiebell:
@sirugh That is the long term vision, to maintain current and previous backend versions, but we only have this single instance we're ready to maintain and share right now. I'll add a P2 to be addressed before next release to add `2.3.4` to the list. Zetlen also wanted this to live somewhere public outside of the repo, because right now we can't update the list unless be publish (which I suppose we could do with a quick patch release if we felt this was high priority).

PR #1976, opened 2019-11-14
Improved Image Fit with Fastly

supernova-at:
### tl;dr

This PR adds a `fit=cover` image optimization parameter to the `srcset` urls of resource images. This will be ignored by `express-sharp`, which already produces the intended behavior, but will instruct Fastly to resize the image to entirely cover the given `width` / `height` dimensions.

See https://docs.fastly.com/api/imageopto/fit for more details.

### Overview

The `CategoryTile` images on the homepage of our staging site (upward-php + fastly) did not look good:

Screen Shot 2019-11-14 at 11 27 29 AM

Compare that to the latest `develop` (upward-js):

Screen Shot 2019-11-14 at 11 28 41 AM

### Investigation

We were sending the same image resize parameters (`width=80`, `height=100`) in both places but getting much different responses back:

**upward-php + fastly**:
80x100
Screen Shot 2019-11-14 at 8 29 42 AM

**upward-js**:
178x100
Screen Shot 2019-11-14 at 8 29 37 AM

### Resolution

This is because our onboard image optimization middleware (powered by `express-sharp`) was handling things differently than Fastly when both `width` and `height` resize parameters were present. Since we wanted Fastly to emulate the `express-sharp` way, this PR updates the urls in resource image `srcset`s to include `fit=cover`, which will resize the image to entirely cover the given `width` & `height` dimensions.

Support for this property was already present in [our makeUrl utility](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/util/makeUrl.js#L92-L98), with the following comment:

> Fit does not do anything within express-sharp, this is used within Fastly to achieve the same output of desired cropping: https://docs.fastly.com/api/imageopto/fit. Passing this to express-sharp doesn't have any side effects.

See also https://docs.fastly.com/api/imageopto/fit.

## Related Issue

Closes PWA-159.

### Verification Stakeholders

@tjwiebell @zetlen 

### Verification Steps

1. Go to the homepage of a fastly-enabled site
1. See that the images are correct (not squished)

## Checklist


packages/venia-ui/lib/components/CategoryList/categoryTile.css
```diff
@@ -15,6 +15,7 @@
     display: block;
     height: 5rem;
     object-fit: cover;
+    object-position: center;
```
supernova-at: This isn't strictly necessary I suppose, the initial value of `50% 50%` will have the same result.

packages/venia-ui/lib/util/images.js
```diff
@@ -21,7 +21,8 @@ const generateURL = (imageURL, mediaBase) => (width, height) => {
     const url = resourceUrl(imageURL, {
         type: mediaBase,
         width,
-        height
+        height,
+        fit: 'cover'
```
supernova-at: This is the change that actually fixes the issue.

packages/venia-ui/lib/components/CategoryList/categoryTile.css
tjwiebell: Unless you think it's better for readability (ie. the common developer doesn't know centering is the default), let's remove it. Not going to block my approval though.

PR #1984, opened 2019-11-18
Create a Venia styleguide

jimbo:
## Description
As we enter a new phase of Venia feature development and revisit some of our original designs, it's time for us to have a living styleguide. We need a style guide for two reasons:

- to keep track of our design intent and manage changes to it
- to maintain canonical implementations of our design in CSS

We'd like to use [Adobe Spectrum](https://spectrum.adobe.com/) as a basis for this former, and our own Venia UI as a basis for the latter.

Accordingly, this PR adds a new package to the repository: `@magento/venia-styleguide`. This new package is an application (like `venia-concept`), not a library (like `venia-ui`), though in the near future it may export some CSS for `venia-ui` to consume.

## Related Issue
PWA-152

### Verification Stakeholders
- @soumya-ashok 
- @jcalcaben 
- @zetlen 

### Specification
N/A

### Verification Steps
1. `yarn`
2. `yarn workspace @magento/venia-styleguide run start`
3. Open a browser window to the dev server URL

## Checklist
* I have *not* updated the documentation accordingly, if necessary.
* I have *not* added tests to cover my changes, if necessary.


packages/venia-styleguide/src/pages/typography/content.yml
```diff
@@ -0,0 +1,2 @@
+---
+page_description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Fusce quis euismod nisi. Morbi metus mauris, volutpat ac aliquet eget, laoreet vel tortor. Pellentesque commodo tellus."
```
jcalcaben: What do you think about using [MDX](https://github.com/mdx-js/mdx) for defining the page content/docs?

jimbo: Looks interesting! Probably a lot better than plain YAML or JSON.

supernova-at:
Yep, seems like a good start to me! 👍

packages/venia-styleguide/src/pages/typography/content.yml
jimbo: @jcalcaben Okay, I've added `@mdx-js/loader`, and I can confirm that it can successfully load and render `.md` files as MDX files.

I think we should probably restrict our usage of it to the sectional content for now. MDX files support some React basics and some JS primitives, but they're best for pure documents. We'll have to decide whether to inject components on a per-section basis or at the root with `MDXProvider`; I'm leaning toward the latter.

jimbo:
@tjwiebell @supernova-at Thanks for taking a look. Please re-review when you get a chance, since I added a loader for MDX files at the suggestion of @jcalcaben.

packages/venia-styleguide/src/pages/typography/content.yml
jcalcaben: I agree. Let's go with the latter and see what that experience is like.

PR #1985, opened 2019-11-18
Refactor User Detail Fetching to GraphQL

tjwiebell:
## Description
See PWA-138

## Related Issue

* [[PWA-138](https://jira.corp.magento.com/browse/PWA-138)] User details should be fetched from GraphQL

### Verification Steps

1. Create Account
- Go through Create Account workflow
- Verify user details are displayed correctly after successful creation (at bottom on nav bar)
2. Sign-In
- Logout and sign-in with the same account as above
- Verify you're redirected to MyAccount/Sign-out Page
- Verify user details still displayed correctly
3. Sign-out
- Sign-out
- Verify user details is replaced by signin button
4. MyAccount
- While signed in, clear browser cache while retaining local storage
- Verify user details are loading correctly on initial app render

## Checklist


packages/peregrine/lib/talons/Navigation/useNavigation.js
```diff
@@ -11,16 +12,20 @@ const ancestors = {
...
-    const [, { getUserDetails }] = useUserContext();
+    const [, { actions: userActions }] = useUserContext();
+    const { data } = useQuery(customerQuery);
 
-    // request data from server
     useEffect(() => {
-        getUserDetails();
-    }, [getUserDetails]);
+        if (data) {
+            userActions.getDetails.receive(data.customer);
```
tjwiebell: @jimbo 
This feels wrong to me, and @sirugh and I weren't sure which direction this should go.

1. Continue storing `currentUser` state in redux. We'd need to create a `setCurrentUser` action that simulates `actions.getDetails.(request|receive)`. This would be more backward compatible, but simulating a network request makes `isGettingDetails` useless.
2. Pull `currentUser` state in `UserContext` and remove a ton of redux code. If I were writing this from scratch today, I think this is where I'd put it, so just looking for arguments against that.

This problem is challenging me because query state lives in the talons, so there's really no use for it in redux anymore. Is there a better pattern we can use here? Is there a tertiary solution here where talons pass queries off the redux that hasn't been established yet?

sirugh: >  Is there a tertiary solution here where talons pass queries off to redux that hasn't been established yet?

If we could figure out a way to do this the work to convert to graphql gets a lot simpler. We would be able to replace the async REST calls with graphql requests. Right now we're fighting two battles -- replacing rest with graphql and figuring out how to refactor from redux actions to components.

sirugh: Alternatively, in this specific case we may not need to store user data in the store anymore. I was suggesting in our chat that we just have the components fetch the data they need to render through graphql. This would mean that other components who care about `currentUser` would also have to make a `customerQuery`.

That only really works if we can inform those components of some upstream change though -- `isSignedIn` is a good example of a global state we can use to trigger updates within the components.

So that would look like this:

```js
const useNavigation = () => {
  const [{ isSignedIn }] = useUserContext();
  const [getUserDetails, { error, loading, data }] = useLazyQuery(customerQuery)
  
  useEffect(() => {
    if(isSignedIn) {
      getUserDetails()
    }
  }, [...]

  if (data) {
    // render Navigation with user data
  }
}
```

jimbo: Good questions.

- There's a reason we make sync actions, such as `userActions.getDetails.receive()`, available from `UserContext`: sync actions are the simplest way to update the store. So this doesn't really feel wrong to me, and I don't think a `setCurrentUser` async action would add any value.

- The state we keep in the store is all global state, and there is no state more global than `currentUser`. Here, you've suggested "pulling" `currentUser` state into `UserContext`; how would you keep it? Would the `user` slice remain, and in what capacity?

- One of the reasons we chose to keep the store for global state is that our most orchestrated async actions—those in cart and checkout—rely on having access to the latest state during each step, via `getState()`. For example, in checkout, `getShippingMethods()` pulls the latest value of `user` directly from the store before determining which endpoint to use. If this state isn't in the store, then a component would have to read it from context and pass it as an argument (payload). The `user` state is unlikely to be stale, but in other cases, such as `cartId`, state received from the component could be stale by the time the async action used it.

---

Long story short, I think it's best to keep our global state in the store for now.

tjwiebell: - Point 1: Awesome! I had not seen any direct usage of sync actions, so good to know. This does pose a followup question below.
- Point 2: No real reason for this suggestion. `currentUser` state seemed half-baked in redux, and most components I found were triggering their own fetch before using it (also the object shape was `whatever is returned by REST`). Was really just trying to kick off the conversation about when to roll our own state into context, or use redux. Happy to leave it where it is, and try to clean up its usage.
- Point 3: Noted. I was a little narrowed on `user.currentUser`, but can still see where sharing that state might be needed to pre-fill address information, or something like that.

New Question: Do we continue adhering to the existing action map pattern? Piggy backing on @sirugh example:

```javascript
const useNavigation = () => {
  const [{ isSignedIn }, { actions: userActions }] = useUserContext();
  const [getUserDetails, { error, loading, data }] = useLazyQuery(customerQuery)
  
  useEffect(() => {
    if(isSignedIn) {
      userActions.getDetails.request(); // trigger request action
      getUserDetails()
    }
  }, [...]

  if (data || error) {
    userActions.getDetails.receive(data, error);
  }
}
```

sirugh: Do you need to use `getDetails.request`? The request sync actions are mostly used to set some global "loading data" state.

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -29,6 +32,12 @@ export const useSignIn = props => {
         errors.push(getDetailsError);
     }
 
+    useEffect(() => {
+        if (getCustomerData) {
```
sirugh: This will rerun if the data object changes at all. Is that what we want?

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -1,25 +1,28 @@
...
-        { getUserDetails, setToken }
-    ] = useUserContext();
+    const [, { actions: userActions, setToken }] = useUserContext();
 
-    const [signIn, { error: signInError }] = useMutation(query);
+    const [signIn, { error: signInError }] = useMutation(signInMutation);
+    const [
+        getUserDetails,
+        { called, data: getCustomerData, error: getDetailsError, loading }
+    ] = useLazyQuery(getCustomerQuery);
```
sirugh: I wonder if we even need to query for this. It is possible the mutation returns what we need to render.

tjwiebell: This mutation unfortunately only returns the token, customer data wasn't available.

packages/peregrine/lib/talons/SignIn/useSignIn.js
tjwiebell: Yes, though that use case should be impossible, given the component is unmounted right after this.

packages/peregrine/lib/talons/Navigation/useNavigation.js
tjwiebell: That's pretty much the question I was asking; do we continue with this pattern of storing fetching state globally? It feels like the answer is no, but that likely means we need to cleanup our action maps to remove request, and remove fetching states from `initialState` where needed.

packages/peregrine/lib/talons/SignIn/useSignIn.js
sirugh: Ok, what about doing something similar to what I've done in the `createCart` PR?
Basically we have some global/app effect that runs whenever some id necessary to get data is modified.

https://github.com/magento/pwa-studio/pull/1988/files#diff-4080b54df336f2a5838a43c80bc67391R46-R62

packages/venia-ui/lib/queries/getCustomer.graphql
```diff
@@ -0,0 +1,10 @@
+query getCustomer {
+    # This would ideally be a fragment that could be shared where needed.
+    # Please keep shape in sync with createAccount.graphql until this is supported.
+    customer {
+        id
+        email
+        firstname
+        lastname
```
sirugh: Is this all the necessary data we were using before? I know we tend to spread the entire REST response into the store so I definitely want to make sure we are reducing that "noise".

tjwiebell: Only usage of this state I found was in `useMyAccount`, `const { email, firstname, lastname } = currentUser;`, and `useAuthModal` and that damn `id`; so this should cover it. Agree we should only fetch what we need.

tjwiebell: I do anticipate this will expand, to pre-fill billing and shipping details, but that functionality does not currently exist.

packages/peregrine/lib/talons/Navigation/useNavigation.js
```diff
@@ -11,16 +12,27 @@ const ancestors = {
...
+    const { customerQuery } = props;
+
     // retrieve app state from context
     const [appState, { closeDrawer }] = useAppContext();
     const [catalogState, { actions: catalogActions }] = useCatalogContext();
-    const [, { getUserDetails }] = useUserContext();
+    const [{ isSignedIn }, { actions: userActions }] = useUserContext();
+    const [getUserDetails, { data: getCustomerData }] = useLazyQuery(
+        customerQuery,
+        { fetchPolicy: 'network-only' }
```
tjwiebell: This is necessary until we investigate why `apolloClient.resetStore` wasn't clearing browser persistence.

sirugh: Add a comment in the code please :)

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
```diff
@@ -62,20 +59,18 @@ export const useCreateAccount = props => {
...
                     variables: {
                         email: formValues.customer.email,
                         password: formValues.password
                     }
                 });
 
                 const token =
-                    response && response.data.generateCustomerToken.token;
+                    signInResponse &&
+                    signInResponse.data.generateCustomerToken.token;
```
revanth0212: Potential cause of `TypeError`? Also `generateCustomerToken` sounds like a `function` not an `object`. If possible can we change the name?

packages/peregrine/lib/talons/Navigation/useNavigation.js
```diff
@@ -11,16 +12,27 @@ const ancestors = {
...
+    );
 
-    // request data from server
     useEffect(() => {
-        getUserDetails();
-    }, [getUserDetails]);
+        if (isSignedIn) {
+            getUserDetails();
```
revanth0212: Should you not `await` for this or use the output? Does that function cause any side effects like setting the `store` or `context`?

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
tjwiebell: Seems outside of the scope of this PR, but I can investigate. I was expecting `signIn` to `reject` on `error` and enter the `catch`, but its possible it may still `resolve` and expect us to destructure `error` and handle ourselves.

Nothing we can do about that name; data comes back with the name of the mutation, which is `generateCustomerToken`. I'll sanity check that assumption though.

packages/peregrine/lib/talons/Navigation/useNavigation.js
tjwiebell: I really wish `useLazyQuery` returned a Promise like `useMutation`, but it does not ☹️. This call does cause the effect to run again though, by updating `getCustomerData`. Is that okay?

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
sirugh: > data comes back with the name of the mutation

You might be able to modify that through aliasing:

```
query storeConfigData {
  foo: storeConfig {
    id
    copyright
  }
}
```
 returns
```
{
  "data": {
    "foo": {
      "id": 1,
      "copyright": "Copyright © 2013-present Magento, Inc. All rights reserved."
    }
  }
}
```

sirugh: You could do this and alias as `customerToken: generateCustomerToken (...)` but I wonder if maybe it's a red flag that we're even asking this question. The GQL docs look like leaving the verb form query name is fine: https://graphql.org/learn/queries/#mutations

It makes it easier to grok when you consider that we may be composing several queries or mutations at once. We may need some layer that is able to understand what queries have returned with data without understanding what data they return.

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -8,26 +8,20 @@ export const useSignIn = props => {
...
 
-    const [signIn, { error: signInError }] = useMutation(query);
+    const [signIn, { error: signInError }] = useMutation(signInMutation);
 
     const errors = [];
     if (signInError) {
         errors.push(signInError.graphQLErrors[0]);
     }
-    if (getDetailsError) {
-        errors.push(getDetailsError);
```
sirugh: You may want to keep this code. You can still propagate the error to the store by calling `getDetails.receive` with an error.

packages/peregrine/lib/talons/Navigation/useNavigation.js
sirugh: Maybe this would be a good candidate for the idea @jimbo had. Create a `useCustomerData` hook that returns an async callback (and uses refs internally to ensure we don't double fetch?). It can handle storing that data too Then you would just invoke that here.

packages/peregrine/lib/talons/Navigation/useNavigation.js
```diff
@@ -11,16 +12,27 @@ const ancestors = {
...
+
+            if (getCustomerData) {
+                userActions.getDetails.receive(getCustomerData.customer);
+            }
```
sirugh:
```js
if (getCustomerError) {
  userActions.getDetails.receive(new Error('Unable to get customer details'))
}
```

tjwiebell:
Closing in favor of solution in #2004.

PR #1987, opened 2019-11-19
Use graphql to add item to cart

sirugh:
## Description

* Updates `addItemToCart` to pass mutations for adding different product types to the cart.

* Updates `updateItemInCart` to pass the necessary mutations for add item since addItem action was used as a fallback case.

* Updates retry logic for graphQL response in the addItem error catch.

## Related Issue

Closes [PWA-139](https://jira.corp.magento.com/browse/PWA-139).

### Verification Steps

1. Add a simple product to cart (no options ie jewelry). The cart should display properly with the item.
2. Add a configurable product to cart (a dress or something). The cart should display properly with the item.
3. Log in and add an item to the cart. It should display properly.
3. Log out and add an item to the cart. It should display properly.
4. Edit an item in the cart. It should update properly.
5. Remove an item from the cart. It should be removed properly.

## Checklist


packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -499,23 +497,25 @@ function toRESTCartItem(cartId, payload) {
     return cartItem;
 }
 
-export async function writeImageToCache(item = {}) {
-    const { media_gallery_entries: media, sku } = item;
+export const writeImageToCache = (item = {}) =>
+    async function thunk() {
```
sirugh: To make this useable in the talon from the cart context I had to update the signature of this function otherwise an error was thrown about not using plain objects or something. This is likely a bug with how we connect the actions but I'm not willing to solve it here because we're likely to throw away this code when we move `getCartDetails` to graphql. At that time we will just get the image url from the GQL response.

packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -1,60 +1,22 @@
...
-import { getCartDetails, removeCart } from '../cart';
+import { removeCart } from '../cart';
 import { clearCheckoutDataFromStorage } from '../checkout';
 
 import actions from './actions';
 
 const { request } = Magento2;
 const storage = new BrowserPersistence();
 
-export const signIn = credentials =>
```
sirugh: Should have deleted this in the #1904 but I didn't. Doing it now since I'm trying to clean up refs to `getCartDetails`.

packages/venia-ui/lib/components/App/app.js
```diff
@@ -33,6 +36,22 @@ const App = props => {
...
+    // TODO: Should this belong elsewhere? Basically this is global logic that says "any time the cartId is deleted, recreate the cart".
+    useEffect(() => {
+        async function initializeCart() {
+            const {
+                data: { createEmptyCart }
+            } = await createCart();
+            await setCartId(createEmptyCart);
+        }
+        if (!cartId) {
+            initializeCart();
```
sirugh: Now, as a side effect of not having a `cartId` we re-create one.

Ideally we _never_ call the `createCart` action as it uses REST. We want to use the `createCart` mutation instead.

In this PR you'll see places I [removed](https://github.com/magento/pwa-studio/pull/1987/files#diff-ca0eebf490b4ae87e97fbec37383872bL57) calls that would create a new cart through REST. Instead, we now fetch a new cart, set the cartId, and then fetch the details for that cart.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -54,9 +56,13 @@ export const useAuthModal = props => {
     }, [showMyAccount]);
 
     const handleSignOut = useCallback(() => {
+        // After logout, reset the store to set the bearer token.
+        // https://www.apollographql.com/docs/react/networking/authentication/#reset-store-on-logout
+        apolloClient.resetStore();
```
sirugh: This is important -- you'll see other calls to this function in this PR. We should have been doing it whenever the login state changed to make sure that requests through apollo used the bearer token. You can see this "bug" in [develop](https://develop.pwa-venia.com/). Make sure you're logged out, then log in and without a reload perform some action that would call a graphql endpoint. You'll see that the `authorization: Bearer <token>` header is not present on the request.

After this change, it will be.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -425,24 +404,40 @@ function toRESTCartItem(cartId, payload) {
...
         }
+    };
+
+/**
+ * This function returns the correct cartId for use by the REST endpoint.
+ * For authed users we have to use the "actual" id. For guest users we use the
+ * "masked" id. When we fully convert cart requests to graphql we can do away
+ * with this function.
+ */
+function getCartIdForREST(cart, user) {
```
sirugh: Oops need to delete. Duped from rebase.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -190,8 +181,7 @@ export const updateItemInCart = (payload = {}) => {
                     // Add the updated item to that cart.
                     await dispatch(
                         addItemToCart({
-                            ...payload,
-                            fetchCartId
+                            ...payload
```
sirugh: This concerns me. We have a case in `updateItemInCart` where we may call `addItemToCart` in the case of a newly created cart. The payload for `addItemToCart` is actually changed in this PR so this code no longer actually works.

We would have to somehow do all the logic in the [handleAddToCart](https://github.com/magento/pwa-studio/pull/1987/files#diff-27ab3c434634ee4e2b37a99c7a9772f8R191) callback right here _or_ do it wherever we invoke the `updateItemInCart` action.

Maybe we should do the variant matching and typename conditionals in the `addItemToCart` action. Then, this should function still...

Looking for thoughts.

sirugh: Moving the content of the callback into the action should work but it'll require that any user of `updateItemInCart` pass the mutations for adding items as well. 😭

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -100,12 +93,10 @@ export const addItemToCart = (payload = {}) => {
...
             dispatch(actions.addItem.receive());
         } catch (error) {
-            const { response, noCartId } = error;
-
             dispatch(actions.addItem.receive(error));
 
-            // check if the cart has expired
-            if (noCartId || (response && response.status === 404)) {
+            // If we got an error and it's not a network problem, retry.
+            if (!error.networkError) {
```
sirugh: We may need to get a list of the possible graphql errors. I'm sure there may be some where we don't want to fully recreate the cart. It would suck to get some "out of stock" type error and lose your entire cart.

packages/peregrine/lib/store/actions/cart/asyncActions.js
sirugh: Ended up simplifying this by reverting most of the logic. I just needed to deconstruct the original payloads better in `addItem`. Now `updateItemInCart` also has the `addItem` mutations in the payload so if we get here we are fine.

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
```diff
@@ -49,22 +43,6 @@ export const useMiniCart = props => {
         setIsEditingItem(false);
     }, []);
 
-    const handleUpdateItemInCart = useCallback(
```
sirugh: I moved this to `cartOptions` where it was more appropriate.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -134,6 +145,16 @@ export const addItemToCart = (payload = {}) => {
     };
 };
 
+// Returns true if the cart is invalid.
+function hasRetryableGqlError(error) {
```
sirugh: This function feels nasty. I don't like matching some string value from the API. Would rather we be able to use some error id map. GQL always returns 200 so we no longer get 404 for invalid carts.

sirugh: Turns out that `category: "graphql-no-such-entity"` is used if anything within the response is not found. To ensure we only retry on an invalid cart, and not on an invalid/oos item, we should use the `message` string.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -52,39 +52,49 @@ export const createCart = payload =>
...
+                parentSku,
+                product: item,
+                quantity,
+                sku: item.sku
+            };
+
+            let addItemMutation;
+
+            // Prepare the params to add and use the proper mutation.
+            if (productType === 'SimpleProduct') {
```
tjwiebell: This might prove to be a terrible idea once we add other product types, but hear me out. If we move this logic to `useProductFullDetail`, you can pass a single addItem mutation to the action. I hope we don't sit in this intermediate phase too long, but limiting action dependencies seems preferable here.

Whatchu think?

packages/venia-ui/lib/queries/addConfigurableProductsToCart.graphql
```diff
@@ -0,0 +1,29 @@
...
+            cart_id: $cartId
+            cart_items: [
+                {
+                    data: { quantity: $quantity, sku: $sku }
+                    parent_sku: $parentSku
+                }
+            ]
+        }
+    ) {
+        cart {
```
tjwiebell: Same comment as other PR; fetching this data has a cost on the backend (though I hope this primes the cache, and the next getCartDetails call serves from the cache 🤞)

packages/venia-ui/lib/queries/addSimpleProductsToCart.graphql
```diff
@@ -0,0 +1,23 @@
...
+    $quantity: Float!
+    $sku: String!
+) {
+    addSimpleProductsToCart(
+        input: {
+            cart_id: $cartId
+            cart_items: [{ data: { quantity: $quantity, sku: $sku } }]
+        }
+    ) {
+        cart {
```
tjwiebell: Samsies

packages/venia-ui/lib/queries/addConfigurableProductsToCart.graphql
sirugh: I replied in the other thread but basically there is no such thing as a `null` response in GQL. You can choose to return a success indicator boolean or some other "output" type, as we do in Magento GQL. You can check the schema, or you can try in graphiql yourself, but it will complain if you remove the output.

sirugh: And this *may* prevent the extra "get details" request from being necessary, so not completely a bad thing.

packages/peregrine/lib/store/actions/cart/asyncActions.js
sirugh: > If we move this logic to useProductFullDetail, you can pass a single addItem mutation to the action.

I had it like that originally but I was also doing logic to create the `variables` object within the talons rather than the action. I'll see what this looks like in the talons. Remember, `updateItemInCart` will need to behave the same. I'll make the change there as well if it looks OK here.

> limiting action dependencies seems preferable here.

I agree.  I think it may make refactoring easier later.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -203,10 +203,20 @@ export const useProductFullDetail = props => {
             appendOptionsToPayload(payload, optionSelections, optionCodes);
         }
 
+        let addItemMutation;
+
+        // Prepare the params to add and use the proper mutation.
+        if (productType === 'SimpleProduct') {
```
tjwiebell: `productType` is `undefined` in this scope, bad copy/paste from the action.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -203,10 +203,20 @@ export const useProductFullDetail = props => {
...
+            addItemMutation = addSimpleProductToCart;
+        } else if (productType === 'ConfigurableProduct') {
+            addItemMutation = addConfigurableProductToCart;
+        } else {
+            throw new Error('Unsupported product type. Cannot add to cart.');
```
tjwiebell: This is also now a unhandled exception. Since we know the add will fail, why don't we just disable the button, and `console.error`. If you have a more user friendly way to handle this, open to suggestions.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
sirugh: Ooops!

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
sirugh: I'm not expecting this case to occur -- we only use `Simple` and `Configurable` in our storefront.

That said, I'll disable the button for now so it's impossible to attempt. I'll log something with it too.

dpatil-magento:
@sirugh @tjwiebell Add item to cart and place order >> now close checkout panel by clicking close icon or by clicking in greyed out area and try adding same product again. Product is not getting added to cart.
Note - Once we open cart manually after that it starts working.

tjwiebell:
@sirugh - Looks like the cartId check was removed, so its now possible to enter with a null `cartId` after checkout. We assumed this would fall into the retry logic with error message pattern matching, but its throwing a different message because of type validation, `null !== String!`. Would like to push this through, some let me know if you'd like me to just a push a solution and get this approved w/o ya.

Possible solutions:
1) Pattern match this new message in `isInvalidCart`
2) Re-add init cart if `cartId` not present before add

Message
`Error: GraphQL error: Variable "$cartId" got invalid value null; Expected non-nullable type String! not to be null.`

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -52,39 +52,28 @@ export const createCart = payload =>
...
-                cartEndpoint = '/rest/V1/carts/mine/items';
-            }
-
-            const quoteId = getQuoteIdForRest(cart, user);
-            const cartItem = toRESTCartItem(quoteId, payload);
-            await request(cartEndpoint, {
-                method: 'POST',
-                body: JSON.stringify({ cartItem })
+            const { cart } = getState();
+            const { cartId } = cart;
```
sirugh: So this can be `null` after an order is placed, and a subsequent `add` calls this with a `null` cartId.

@tjwiebell suggested two options, but I think it may be better to figure out why the cart is deleted but never re-created after order. It should definitely be created _every_ time the cartId is deleted.

sirugh:
@dpatil-magento I believe I "fixed" the issue :) Tried it myself and I'm able to add immediately after checkout.

dpatil-magento:
@sirugh Recent commit has broken checkout flow. After placing an order as guest/registered user, error is being displayed.

image

dpatil-magento:
@sirugh Dont see error anymore but after placing order for fraction of second checkout form displayed again (Slow 3G screen captured below). Also "Create an Account" button is not functional anymore and original bug has reappeared (after placing order unable to add product to cart).

image

sirugh:
The checkout form being visible for a fraction of a second is because we `await` the completion of checkout before we show the receipt. Something must be modifying the editing state in between submission though -- I can look into it.

As for the `createAccount` button not working on receipt, I can also fix.

dpatil-magento:
@sirugh Looks good now. Can be merged once review is approved.

PR #1988, opened 2019-11-20
Uses graphQl mutation to create/fetch cartId.

sirugh:
## Description

NOTE: REST endpoint returns data for "authed" cart created through graphql. To repro, log in, add item to cart, then delete sign in token from local storage and refresh. Cart details should be fetched but you should not be able to add/update/remove. If you take an action on the cart it will reset to a new cart. This could be a bug in REST. It should not matter since those actions are moving to graphql which won't return data for unauthed user.

This PR precedes #1987.

Changes include:

* **Proper use of cart id for quote id**

GraphQL endpoints use the masked (hashed) cartId regardless of auth state. REST uses the "actual" (table key value) cartId as quoteId for authed requests and masked cartId as quoteId for guest requests. This means that while we are in this intermediate state of _some_ requests being through graphql (such as `createCart` and `addItemToCart`) we have to make sure the REST calls use the appropriate cartId.

* **Passing of fetch/invoker to `createCart`**

The majority of this PR is to pass the mutation callback into the actions that may need to call `createCart`.

* **Cleanup of reducer state**

We were passing and spreading unnecessary info to the store in the `receive` actions of certain asyncActions. I removed this as it was extraneous.

## Related Issue

Closes [PWA-140](https://jira.corp.magento.com/browse/PWA-140).

### Verification Stakeholders

@jimbo

### Verification Steps

1. Clear storage and load the app. 
2. Check network for the graphql for `createCart`. Note the response cartId.
3. Check local storage and you should see the cartId in local storage.
4. Add an item to cart.
5. Update an item in cart.
6. Remove an item from cart.
7. Remove all items from cart.
8. Sign in, add an item, then expire the cart from the admin console and try to remove the item. It should still function same as it did before.

## Checklist


packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -54,9 +56,13 @@ export const useAuthModal = props => {
     }, [showMyAccount]);
 
     const handleSignOut = useCallback(() => {
+        // After logout, reset the store to set the bearer token.
+        // https://www.apollographql.com/docs/react/networking/authentication/#reset-store-on-logout
+        apolloClient.resetStore();
```
sirugh: This is important -- you'll see other calls to this function in this PR. We should have been doing it whenever the login state changed to make sure that requests through apollo used the bearer token. You can see this "bug" in develop. Make sure you're logged out, then log in and without a reload perform some action that would call a graphql endpoint. You'll see that the authorization: Bearer <token> header is not present on the request. After this change it will be.

This is significant because if a user token is stored in local storage we should make sure the bearer token is set on the graphql call to `createCart`.

packages/venia-ui/lib/components/App/app.js
```diff
@@ -33,6 +37,30 @@ const App = props => {
...
+            const {
+                data: { createEmptyCart }
+            } = await createCart();
+            await setCartId(createEmptyCart);
+
+            getCartDetails({ forceRefresh: true });
+        }
+
+        if (!cartId) {
+            initializeCart();
```
sirugh: Now, as a side effect of not having a cartId we re-create one.

Ideally we never call the createCart action as it uses REST. We want to use the createCart mutation instead.

In this PR you'll see places I removed calls that would create a new cart through REST. Instead, we now fetch a new cart, set the cartId, and then fetch the details for that cart.

packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -1,60 +1,22 @@
...
-import { getCartDetails, removeCart } from '../cart';
+import { removeCart } from '../cart';
 import { clearCheckoutDataFromStorage } from '../checkout';
 
 import actions from './actions';
 
 const { request } = Magento2;
 const storage = new BrowserPersistence();
 
-export const signIn = credentials =>
```
sirugh: Should have deleted this in the #1904 but I didn't. Doing it now since I'm trying to clean up refs to getCartDetails.

packages/peregrine/lib/store/reducers/cart.js
```diff
@@ -2,12 +2,16 @@ import { handleActions } from 'redux-actions';
...
+
+const { BrowserPersistence } = Util;
+const storage = new BrowserPersistence();
 
 export const name = 'cart';
 
 export const initialState = {
     addItemError: null,
-    cartId: null,
+    cartId: storage.getItem('cartId'),
```
sirugh: This is a pattern also used by the `user` reducer to ensure we synchronize the state on load with local storage.

packages/peregrine/lib/store/reducers/cart.js
```diff
@@ -74,10 +76,7 @@ const reducerMap = {
             isAddingItem: false
         };
     },
-    [actions.updateItem.request]: (state, { payload, error }) => {
-        if (error) {
-            return initialState;
```
sirugh: We should not be passing error to the `request` sync actions, only the `receive`.

packages/peregrine/lib/store/reducers/cart.js
```diff
@@ -43,7 +46,6 @@ const reducerMap = {
             return {
                 ...state,
                 detailsError: payload,
-                cartId: null,
```
sirugh: Removing `cartId` should be done through `removeCart`.

packages/peregrine/lib/store/reducers/cart.js
```diff
@@ -31,10 +35,9 @@ const reducerMap = {
             cartId: String(payload)
         };
     },
-    [actions.getDetails.request]: (state, { payload }) => {
+    [actions.getDetails.request]: state => {
         return {
             ...state,
-            cartId: String(payload),
```
sirugh: The only thing setting `cartId` should be `getCart` action.

jimbo: For context, why did this action accept `cartId` as a payload in the first place, and why did the reducer take this opportunity to set it in state?

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
jimbo: Wow. I'll agree, this is important and good to know.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -54,9 +56,13 @@ export const useAuthModal = props => {
...
+
         // TODO: Get history from router context when implemented.
         signOut({ history: window.history });
-    }, [signOut]);
+    }, [apolloClient, signOut]);
```
jimbo: I would expect `apolloClient` to be a stable object, but just in case, I would consider destructuring `resetStore` and using it as the dependency.

```es6
const { resetStore } = apolloClient

const handleSignOut = useCallback(
  () => {
    resetStore()
  },
  [resetStore]
)
```

packages/venia-ui/lib/components/App/app.js
jimbo: Consider making a custom hook to contain this effect, so as not to bloat `App` with cart code.

packages/peregrine/lib/store/reducers/cart.js
sirugh: No idea. Requests are only meant to set `loading `states in redux right now. Even that will eventually move to UI I believe. At that point all we will are about are `.receive` sync actions.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -136,29 +88,29 @@ export const updateItemInCart = (payload = {}, targetItemId) => {
...
+            if (!isSignedIn) {
+                const { cartId } = cart;
 
-            const cartItem = toRESTCartItem(cartId, payload);
+                if (!cartId) {
+                    const missingCartIdError = new Error(
+                        'Missing required information: cartId'
+                    );
+                    missingCartIdError.noCartId = true;
+                    throw missingCartIdError;
```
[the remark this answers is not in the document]
sirugh: We should never not have a cartId. These are just leftover from before. I can remove them. They won't matter when we move these functions to the UI components.

packages/venia-ui/lib/components/App/app.js
sirugh: Done!

packages/peregrine/lib/store/actions/cart/asyncActions.js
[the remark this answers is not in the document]
jimbo: No, please don't remove them. It doesn't hurt to have a helpful error thrown if we reach a state we shouldn't.

I just wanted to make sure that we weren't expecting this case to come up.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -31,42 +31,19 @@ export const createCart = () =>
...
-            const guestCartEndpoint = '/rest/V1/guest-carts';
-            const signedInCartEndpoint = '/rest/V1/carts/mine';
-            const cartEndpoint = user.isSignedIn
-                ? signedInCartEndpoint
-                : guestCartEndpoint;
-
-            const cartId = await request(cartEndpoint, {
-                method: 'POST'
-            });
+            const { data } = await fetchCartId();
```
sirugh: How do we feel about directly using the `useMutation` API within a thunk? Should we implement a `useAwaitMutation`? It'd protect us from changes to the API from apollo hooks later.

@jimbo @tjwiebell @supernova-at

tjwiebell: The only reason I would propose doing this would be to match the error implementation of `useAwaitQuery`. It looks to me like `useMutation` still resolves on error state, and expects you to destructure `error` to handle that.

```
const { data, error } = await fetchCartId();
const receivePayload = error ? new Error(error) : data.cartId;
dispatch(actions.getCart.receive(receivePayload));
```

packages/peregrine/lib/store/actions/checkout/asyncActions.js
```diff
@@ -38,7 +43,7 @@ export const cancelCheckout = () =>
 export const resetCheckout = () =>
     async function thunk(dispatch) {
         await dispatch(closeDrawer());
-        await dispatch(createCart());
+        await dispatch(removeCart());
```
sirugh: This moves the creation of the cart down the line to the next action that depends on the cart id. Technically it's a tiny bit slower than how it behaved before. Do we care? I'd have to thread `fetchCartId` into this action if we want to leave it.

packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -1,60 +1,22 @@
...
     // Sign the user out in local storage and Redux.
     await dispatch(clearToken());
     await dispatch(actions.reset());
     await clearCheckoutDataFromStorage();
 
-    // Now that we're signed out, forget the old (customer) cart
-    // and fetch a new guest cart.
+    // Now that we're signed out, forget the old (customer) cart.
     await dispatch(removeCart());
-    dispatch(getCartDetails({ forceRefresh: true }));
```
sirugh: No reason to refetch cart details if we're going to refresh. This wasn't even awaited so it was a pointless call.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -31,42 +32,23 @@ export const createCart = () =>
...
-            if (user.isSignedIn) {
-                await request('/rest/V1/carts/mine/billing-address', {
-                    method: 'POST',
-                    body: JSON.stringify({
-                        address: {},
-                        cartId
-                    })
-                });
+            if (error) {
+                throw new Error('Unable to fetch cartId:', error);
```
tjwiebell: There's nothing else in the try block that throws, so it seems odd to keep it just to throw inside; get rid of the try/catch and just handle error state (similar to code sample I provided).

packages/peregrine/lib/store/reducers/cart.js
tjwiebell: This makes me nervous why this existed, but I couldn't find any components observing this state, so seems safe 🤞

packages/venia-ui/lib/components/Header/cartTrigger.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { shape, string } from 'prop-types';
 import { ShoppingCart as ShoppingCartIcon } from 'react-feather';
-
+import CREATE_CART_MUTATION from '../../queries/createCart.graphql';
```
tjwiebell: I bet vscode auto imported this and put it here, but I think we should try to stick to a consistent style of grouping imports going forward (though not documented, and loosly followed). I first saw @supernova-at do this, and really like it. What do you think?

<lib imports>

<@magento imports>

<local imports>

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -281,16 +273,25 @@ export const removeItemFromCart = payload => {
         // and create a new cart to prevent a bug where the next item added to the
         // cart has a price of 0. Otherwise refresh cart details to get updated totals.
         if (isLastItem) {
-            await clearCartId();
-            dispatch(createCart());
+            await dispatch(removeCart());
```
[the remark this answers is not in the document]
sirugh: It was but I figure we can do this logic when we work on the relevant actions.

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -43,14 +45,15 @@ export const useSignIn = props => {
                 const token =
                     response && response.data.generateCustomerToken.token;
 
-                setToken(token);
+                await setToken(token);
```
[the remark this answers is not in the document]
sirugh: Thanks to @tjwiebell :D

packages/peregrine/lib/store/actions/user/asyncActions.js
tjwiebell: Did you create a ticket to bring this action back? Would be nice if it referenced this comment on the off chance one of us doesn't pick it up.

You also left `SIGN_IN` in the action map, but I'm fine not touching it since it _should_ be coming back.

sirugh: I didn't -- I don't think it's necessary right now. You?

tjwiebell: I would probably want it moved before the next release, just so we're providing a consistent pattern for other developers to follow. Given `useSignIn` doesn't have any data fetching or effects that make our custom hooks harder to understand, doesn't have to be right now.

packages/peregrine/lib/store/actions/cart/__tests__/asyncActions.spec.js
```diff
@@ -131,8 +131,8 @@ describe('createCart', () => {
             user: { isSignedIn: false }
         }));
 
-        const error = 'AN ERROR FROM GQL';
-        fetchCartId.mockResolvedValue({ error });
+        const errors = ['AN ERROR FROM GQL'];
+        fetchCartId.mockResolvedValue({ errors });
```
tjwiebell: We should add another test to cover the throw use case; almost exactly this but with `mockRejectedValueOnce(new Error('Woof'))`.

PR #1992, opened 2019-11-21
Webpack optimization for SW.

revanth0212:
## Description

Added Terser Optimizer for Service Worker webpack config.

Using the same Terser Config from the Client's webpack config.

## Related Issue
Closes https://jira.corp.magento.com/browse/PWA-188

### Verification Steps
1. Run `yarn build && yarn stage:venia`.
2. Launch Venia in the browser and refresh to load the SW.
3. You should not see any error or console logs.

## Checklist
* Make sure `yarn stage:venia` is not broken.


supernova-at:
Code looks fine but I'm seeing

Screen Shot 2019-11-25 at 8 29 20 AM

revanth0212:
> Code looks fine but I'm seeing
> 
> Screen Shot 2019-11-25 at 8 29 20 AM

This probably does not have anything to do with the changes. Also, I am not able to replicate the error in my local and also in the deployed version of this PR. Can you check if the SSL key you have is valid? It may have expired.

PR #1993, opened 2019-11-21
Countries from GraphQL

supernova-at:
## Description

This PR updates Venia to fetch countries data from GraphQL instead of the REST endpoint.

## Related Issue

[PWA-144](https://jira.corp.magento.com/browse/PWA-144).

### Verification Stakeholders

@sirugh @jimbo @tjwiebell @revanth0212 

### Verification Steps

1. Add an item to your cart
1. Begin checkout by clicking the "Checkout" button in MiniCart

#### State Validation Still Works

1. Fill out the Shipping Address form but don't fill in a state abbreviation. Verify that you get an inline error message.
1. Fill out the Shipping Address form with an invalid state abbreviation (ex: "GO"). Verify that you get an inline error message.
1. Fill out the Shipping Address form with an invalid state abbreviation (too many characters) (ex: "GOOSE"). Verify that you get an inline error message.
1. Verify that the form submit succeeds with a valid state abbreviation
1. Repeat the steps above for the Billing Address form.

#### Submit Order Still Works

1. Submit an order both using "billing address same as shipping address" and not. Verify that both succeed.

## Checklist


packages/peregrine/lib/store/actions/checkout/__tests__/asyncActions.spec.js
```diff
@@ -116,82 +115,6 @@ describe('resetCheckout', () => {
     });
 });
 
-describe('getCountries', () => {
```
supernova-at: Add tests for `setCountries`

packages/peregrine/lib/store/actions/checkout/asyncActions.js
```diff
@@ -27,7 +27,6 @@ export const beginCheckout = () =>
             })
         );
         dispatch(getShippingMethods());
-        dispatch(getCountries());
```
supernova-at: We no longer `getCountries` as part of the `beginCheckout` action. It happens in the `Flow` component now (where `beginCheckout` is called.

packages/peregrine/lib/store/actions/checkout/asyncActions.js
```diff
@@ -317,10 +304,21 @@ export const createAccount = history => async (dispatch, getState) => {
...
+ *
+ * @param {object} address - The input address.
+ * @param {object[]} countries - The list of countries data.
+ */
+export const formatAddress = (address = {}, countries = []) => {
     const { region_code } = address;
-    const { available_regions: regions } = country;
+
+    const usa = countries.find(({ id }) => id === 'US');
+    const { available_regions: regions } = usa;
```
supernova-at: It's possible we could genericize this function not to be specific to "US" but Venia's UI remains specific to the USA so I didn't change this for now.

packages/peregrine/lib/talons/Checkout/useFlow.js
```diff
@@ -62,6 +71,12 @@ export const useFlow = props => {
         setStep('cart');
     }, [setStep]);
 
+    useEffect(() => {
+        if (countriesData) {
+            setCountries(countriesData.countries);
```
supernova-at: Unfortunately we still do have to persist countries data in redux because the `formatAddress` function needs `countries` available to format with.

supernova-at: Actually ... I might be able to pass `countries` down at the time `submitPaymentMethodAndBillingAddress` is called 👍

packages/peregrine/lib/store/actions/checkout/actions.js
```diff
@@ -1,7 +1,7 @@
 import { createActions } from 'redux-actions';
 
 const prefix = 'CHECKOUT';
-const actionTypes = ['BEGIN', 'EDIT', 'RESET'];
+const actionTypes = ['BEGIN', 'EDIT', 'RESET', 'SET_COUNTRIES'];
```
sirugh: Can you remove `GET_COUNTRIES` from the map in this file?

packages/peregrine/lib/talons/Checkout/useFlow.js
sirugh: Yea I don't think you really need to do that anymore -- it's basically local state now. You can pass it to the shipping form and the billing/payment form and they can use some utility function there.

packages/peregrine/lib/talons/Checkout/useFlow.js
```diff
@@ -31,18 +33,25 @@ export const useFlow = props => {
...
     ] = useCheckoutContext();
-    const { onSubmitError, setStep } = props;
+    const { countriesQuery, onSubmitError, setStep } = props;
+
+    const [runQuery, queryResponse] = useLazyQuery(countriesQuery);
+    const getCountries = runQuery;
+    const { data: countriesData } = queryResponse;
 
     const handleBeginCheckout = useCallback(async () => {
+        getCountries();
```
sirugh: I think we can just do this as a query on mount rather than wait for checkout to begin.

packages/peregrine/lib/talons/Checkout/useFlow.js
sirugh: Do you think `flow` is the best place for this? Or would the `form` component be the more appropriate? I wasn't sure how much prop threading we wanted to encourage.

packages/peregrine/lib/store/actions/checkout/__tests__/asyncActions.spec.js
supernova-at: Not needed anymore https://github.com/magento/pwa-studio/pull/1993/commits/68cd2b8ac59650c6a8b8a677a46d355adf1d1c0b.

packages/peregrine/lib/talons/Checkout/useFlow.js
supernova-at: Redux now no longer needs to track `countries` https://github.com/magento/pwa-studio/pull/1993/commits/68cd2b8ac59650c6a8b8a677a46d355adf1d1c0b and https://github.com/magento/pwa-studio/pull/1993/commits/0d3572a06546452552dd9e4ed6fff36b5857244a.

> Do you think flow is the best place for this? Or would the form component be the more appropriate? I wasn't sure how much prop threading we wanted to encourage.

I had just kept it in the same place it was, but yeah it can definitely be moved down a level. Working on that now.

packages/peregrine/lib/store/actions/checkout/actions.js
supernova-at: Nice catch, thanks. Resolved in https://github.com/magento/pwa-studio/pull/1993/commits/68cd2b8ac59650c6a8b8a677a46d355adf1d1c0b.

packages/peregrine/lib/talons/Checkout/useFlow.js
supernova-at: Done: https://github.com/magento/pwa-studio/pull/1993/commits/68cd2b8ac59650c6a8b8a677a46d355adf1d1c0b.

packages/peregrine/lib/talons/Checkout/useFlow.js
supernova-at: I was able to move it down to `EditableForm` 👍 

https://github.com/magento/pwa-studio/pull/1993/commits/daa0cb90a220420fa8d4ea1c5d5e9e3fe14a84ca.

The countries data isn't populated until the user starts to edit the checkout form:

Screen Shot 2019-11-22 at 8 30 16 AM

Less prop threading but also may open us up to race conditions if the GraphQL query is slow? We need the countries data by the time that "Use Address" button is clicked (for validation).

supernova-at: Okay I just simulated it with a slow 3G connection: our validators are smart and early out. The first validation is that the "State" field is required, which doesn't need the `countries` data to do its job.

I'm just thinking if there's an auto-filled form or something, we may be vulnerable to a case where the user clicks "Use Address" before we have `countries` data.

supernova-at: Ok, the countries data is now fetched when the Checkout Form is shown (when this "Checkout" button is clicked): 

Screen Shot 2019-11-22 at 8 47 46 AM

It only threads down to `EditableForm` > `AddressForm` and `EditableForm` > `PaymentsForm`, and the GraphQL call should return by the time those forms need it for validation.

supernova-at:
This is a `Major` change because we've removed some Redux actions and removed the `countries` property from Redux.

packages/peregrine/lib/store/actions/checkout/asyncActions.js
supernova-at: More details: All we have from the form is a `region_code` (ex: "AL"). Since this could match a valid region in any number of countries, I don't think we should touch this code right now.

packages/peregrine/lib/talons/Checkout/useForm.js
```diff
@@ -0,0 +1,20 @@
...
+export const useForm = props => {
+    const { countriesQuery } = props;
+
+    const [countries, setCountries] = useState(null);
+
+    const { data: countriesData } = useQuery(countriesQuery);
+
+    useEffect(() => {
+        if (countriesData) {
+            setCountries(countriesData.countries);
```
sirugh: Do you need state/effect for this? I think you can just return `countriesData` right?

packages/peregrine/lib/store/actions/checkout/asyncActions.js
```diff
@@ -113,35 +95,41 @@ export const getShippingMethods = () => {
...
-            dispatch,
-            getState
-        );
+        const { countries, formValues } = payload;
+        const { billingAddress, paymentMethod } = formValues;
+
+        submitBillingAddress({
+            billingAddress,
+            countries
+        })(dispatch, getState);
```
sirugh: This syntax looks odd. This is an async action. We should probably have this:

```js
await Promise.all([
  dispatch(submitBillingAddress({ billingAddress, countries })),
  dispatch(submitPaymentMethod(paymentMethod));
]);
```

packages/peregrine/lib/store/actions/checkout/asyncActions.js
```diff
@@ -113,35 +95,41 @@ export const getShippingMethods = () => {
...
+        submitPaymentMethod(paymentMethod)(dispatch, getState);
     };
 
 export const submitBillingAddress = payload =>
     async function thunk(dispatch, getState) {
         dispatch(actions.billingAddress.submit());
 
-        const { cart, checkout } = getState();
-        const { countries } = checkout;
+        const { cart } = getState();
```
sirugh: Nice. Each time we get rid of some slice of state that is used within an action God creates a kitten.

packages/peregrine/lib/talons/Checkout/useForm.js
supernova-at: You da man: https://github.com/magento/pwa-studio/pull/1993/commits/5cbc333c2dce7c892d75927754186d8c6dd12dd5.

packages/peregrine/lib/store/actions/checkout/asyncActions.js
supernova-at: 💪 .

Omitted the `Promise.all` because the previous functionality wasn't waiting.

https://github.com/magento/pwa-studio/pull/1993/commits/5771d0717f5ed71387c30784260699dd86a3de2a

packages/peregrine/lib/talons/Checkout/useForm.js
```diff
@@ -0,0 +1,13 @@
+import { useQuery } from '@apollo/react-hooks';
+
+export const useForm = props => {
+    const { countriesQuery } = props;
+
+    const { data } = useQuery(countriesQuery);
+
+    const { countries } = data || {};
```
sirugh: Now that I read this I realize in the past we would make requests for country data in the `beginCheckout` async action which was `await`ed in the initial opening of the submission form. This means that we don't render the form until we already have the data.

Now, we could possibly render with an empty object for `countries` which may be incorrect. I think this means you should add a loading spinner/state and use that when countries is empty (or null/undefined). In this case it's probably safe to just return `data` since it'll be undefined until then and we can block render until we get an error or it's finished loading.

packages/peregrine/lib/store/actions/checkout/asyncActions.js
```diff
@@ -112,36 +94,38 @@ export const getShippingMethods = () => {
...
-        submitPaymentMethod(payload.formValues.paymentMethod)(
-            dispatch,
-            getState
-        );
+    async function thunk(dispatch) {
+        const { countries, formValues } = payload;
+        const { billingAddress, paymentMethod } = formValues;
+
+        dispatch(submitBillingAddress({ billingAddress, countries }));
+        dispatch(submitPaymentMethod(paymentMethod));
```
sirugh: Actually I think you may need to await this as downstream we do `await` the result of this function:

```js
// in useEditableForm.js
 const handleSubmitPaymentsForm = useCallback(
        async formValues => {
            await submitPaymentMethodAndBillingAddress({
                formValues
            });
            setEditing(null);
        },
        [setEditing, submitPaymentMethodAndBillingAddress]
    );
```

packages/peregrine/lib/talons/Checkout/useForm.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1993/commits/4a253fee5ae2a149102f83edf2c7c5767d3fc1db.

**Loading:**

loading_checkout


**Error:**

error_checkout

Let me know what you think of the implementation for sure.

packages/peregrine/lib/store/actions/checkout/asyncActions.js
supernova-at: The async actions return thunks. `await`ing is not doing what we think it is.

```

export const submitPaymentMethodAndBillingAddress = payload =>
    async function thunk(dispatch) { ... }
```

sirugh: The actual function, though named the same, is actually the bound action creator through use of `bindActionCreators` in each context module. It's async :) 

https://redux.js.org/api/bindactioncreators

packages/peregrine/lib/talons/Checkout/useForm.js
```diff
@@ -0,0 +1,18 @@
...
+        loading: isLoadingCountries,
+        error: countriesError,
+        data: countriesData
+    } = useQuery(countriesQuery);
+    const { countries } = countriesData || {};
+
+    return {
+        countries,
+        countriesError,
+        isLoadingCountries
```
sirugh: I imagine we will want to make these APIs resilient to change as much as possible. To "future proof" this API you could just do something like:

```js
{
  isLoading: !!isLoadingCountries
  hasError: !!countriesError
}
```

And if we add cases for loading or error in the future we just have to extend them within the talon and the API won't change.

packages/venia-ui/lib/components/Checkout/form.js
```diff
@@ -1,21 +1,64 @@
...
+    const handleCountriesError = useCallback(() => {
+        if (!didShowToast) {
+            addToast({
+                type: 'error',
+                icon: ErrorIcon,
+                message: 'Unable to load checkout. Please try again later.',
+                timeout: 3000
+            });
+
+            setDidShowToast(true);
```
sirugh: Rather than a callback you should be able to do this with an effect based on `hasErrors`.

packages/peregrine/lib/store/actions/checkout/asyncActions.js
supernova-at: Oh awesome, thanks.

Pausing pending architecture discussion.

packages/peregrine/lib/talons/Checkout/useForm.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/1993/commits/8ff820c2f73c711c8210e3b53b7ddaa287cbdbde.

packages/venia-ui/lib/components/Checkout/form.js
supernova-at: Thank you! https://github.com/magento/pwa-studio/pull/1993/commits/8ff820c2f73c711c8210e3b53b7ddaa287cbdbde.

packages/peregrine/lib/store/actions/checkout/asyncActions.js
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/1993/commits/d36b0d4f0385d26b43237da0381e8fcd5fd10cf1.

PR #1997, opened 2019-11-21
Added email validation to ForgotPasswordForm

gauravagarwal1001:
## Description

Added Email validation in forgot password form.

## Related Issue

Closes [PWA-191](https://jira.corp.magento.com/browse/PWA-191).

### Verification Steps

    Go to pwa-studio home page
    Click on navigation
    Go for Sign in and then click on forgot password link
    Forgot password form email field has only required validation

## Expected behavior
Forgot password form email field must have email as well as required field validation


revanth0212:
Some tests have failed. Please check.

packages/venia-ui/lib/components/ForgotPassword/ForgotPasswordForm/forgotPasswordForm.js
```diff
@@ -25,7 +25,7 @@ const ForgotPasswordForm = props => {
                 <TextInput
                     autoComplete="email"
                     field="email"
-                    validate={isRequired}
+                    validate={combine([isRequired, validateEmail])}
```
revanth0212: Please import `combine`.

gauravagarwal1001:
> Please fix all the danger warnings.

ok I will fix warnings..

dpatil-magento:
@gauravagarwal1001 Observation - Enter some string and hit submit >> Error will display >> Now make it valid email and hit submit again.
Actual - Error disappeared but email has not been submitted yet. User has to hit Submit again.
Expected - Error should disappear and form should also be submitted at same time.

image
image
image

revanth0212:
> @gauravagarwal1001 Observation - Enter some string and hit submit >> Error will display >> Now make it valid email and hit submit again.
> Actual - Error disappeared but email has not been submitted yet. User has to hit Submit again.
> Expected - Error should disappear and form should also be submitted at same time.
> 
> image
> 
> image
> 
> image

Fixed.

dpatil-magento:
QA Pass. @sirugh Need your thumbs up again.

packages/venia-ui/lib/components/ForgotPassword/ForgotPasswordForm/forgotPasswordForm.js
```diff
@@ -25,8 +26,7 @@ const ForgotPasswordForm = props => {
                 <TextInput
                     autoComplete="email"
                     field="email"
-                    validate={isRequired}
-                    validateOnBlur
```
sirugh: Before @dpatil-magento merges maybe @revanth0212 can explain why we had to remove this. I feel like the validation should happen on blur _and_ before submit, but it seems like it wasn't happening correctly.

revanth0212: The issue with `validateOnBlur` is that it only validates on blur. So once you enter a new valid email address and click on the submit button, the validation is not done yet and the click won't be registered because clicking button === losing focus for the email field but the validation result will only propagate in the next render. That is the reason why @dpatil-magento  had to click twice.

PR #2000, opened 2019-11-22
[chore] Update eslint dep

sirugh:
## Description

This is a follow up to the `magento-eslint` security issue. I updated our usage to `v1.5.0` and in doing so had to handle a bunch of new warnings/errors. I also had to update some peer dependencies and install the config as a dependency to buildpack and peregrine where it was not before.

Question: Should this be a patch? People using it with custom components may start getting warnings for things.

## Related Issue

Closes [PWA-193](https://jira.corp.magento.com/browse/PWA-193).

### Verification Stakeholders
@jimbo @zetlen 

### Verification Steps

1. `yarn && yarn lint`, no errors.
2. Make sure all the disables make sense. If they don't let's fix it here.

## Checklist


packages/venia-concept/_buildpack/create.js
```diff
@@ -91,6 +91,7 @@ function createProjectFromVenia({ fs, tasks, options }) {
                     pkg.scripts[name] = toPackageScript(scriptsToInsert[name]);
                 });
 
+                // eslint-disable-next-line
```
supernova-at: Would be helpful for readers if we explicitly named the rule we were disabling when we do this.

```
// eslint-disable-next-line no-alert
```

sirugh: Good suggestion! I got disable happy :P I'll do this.

.eslintrc.js
```diff
@@ -1,6 +1,10 @@
 const config = {
     parser: 'babel-eslint',
-    extends: ['@magento']
+    extends: ['@magento'],
+    rules: {
+        'no-undef': 'off',
+        'no-useless-escape': 'off'
+    }
```
jimbo: Why do we have to define these two rules everywhere?

sirugh: Because we disregard them throughout the application. The alternative would be to add these rules to `magento-eslint`.

packages/venia-concept/_buildpack/create.js
```diff
@@ -91,6 +91,7 @@ function createProjectFromVenia({ fs, tasks, options }) {
                     pkg.scripts[name] = toPackageScript(scriptsToInsert[name]);
                 });
 
+                // eslint-disable-next-line no-extra-boolean-cast
```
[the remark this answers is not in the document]
sirugh: Rather than require a regression for code changes I was just adding disables where new eslint warnings were occurring in the app at present.

If you want me to do this you'll need to verify it doesn't break anything.

packages/upward-js/lib/resolvers/__tests__/TemplateResolver.test.js
```diff
@@ -42,6 +42,7 @@ test('throws if template engine is unsupported', async () => {
         ),
         context: {
             get: jest.fn(() => {
+                // eslint-disable-next-line no-unused-labels
```
jimbo: Look more closely! This is a bug that's been caught by the linter.

To implicitly return an object from an arrow function, we have to wrap the `{}` in `()`. Since that didn't happen here, what should have been an object property was interpreted as a label, which has been flagged by the linter as unused.

```es6
get: jest.fn(() => ({
    ENV_VAR: 'ENV_VALUE'
}))
```

sirugh: I wonder if this test is just busted then... I'll fix.

sirugh:
On hold again pending a release of magento-eslint to fix some redundancy.

See https://github.com/magento-research/magento-eslint/pull/12

.eslintrc.js
sirugh: I added these to magento-eslint and published but I'm reconsidering that... I don't think we should be disabling `no-undef`. We should be updating some config with those values somewhere.

sirugh: Created [PWA-207](https://jira.corp.magento.com/browse/PWA-207).

PR #2002, opened 2019-11-22
Handling SW actions in Dev mode.

revanth0212:
## Description

Handling SW actions in Dev mode. This PR will stop errors propagating into venia due to SW not being registered in Dev mode.

## Related Issue
Closes https://jira.corp.magento.com/browse/PWA-194

### Verification Steps
1. Run `yarn watch:venia`
2. Go to the deployed app in the browser.
3. Should not see any SW errors but SW should not be registered.
4. Run `yarn build && yarn stage:venia`
5. Go to the deployed app in the browser.
6. Should not see any SW errors but SW should be registered.

image

Should not be seen anymore.

## Checklist
None

### DIFF
--- a/packages/venia-concept/src/registerSW.js
+++ b/packages/venia-concept/src/registerSW.js
@@ -3,7 +3,6 @@ import {
     handleMessageFromSW
 } from '@magento/venia-ui/lib/util/swUtils';
 
-// process.env.DEV_SERVER_SERVICE_WORKER_ENABLED is a string representation of a boolean value
 export const registerSW = () => {
     if (SHOULD_ENABLE_SERVICE_WORKER) {
         navigator.serviceWorker
--- a/packages/venia-concept/src/registerSW.js
+++ b/packages/venia-concept/src/registerSW.js
@@ -1,11 +1,11 @@
-import { handleMessageFromSW } from '@magento/venia-ui/lib/util/swUtils';
+import {
+    SHOULD_ENABLE_SERVICE_WORKER,
+    handleMessageFromSW
+} from '@magento/venia-ui/lib/util/swUtils';
 
 // process.env.DEV_SERVER_SERVICE_WORKER_ENABLED is a string representation of a boolean value
 export const registerSW = () => {
-    if (
-        process.env.NODE_ENV === 'production' ||
-        process.env.DEV_SERVER_SERVICE_WORKER_ENABLED === 'true'
-    ) {
+    if (SHOULD_ENABLE_SERVICE_WORKER) {
         navigator.serviceWorker
             .register('/sw.js')
             .then(() => {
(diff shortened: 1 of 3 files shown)


packages/venia-ui/lib/util/swUtils.js
```diff
@@ -1,3 +1,7 @@
+export const SHOULD_ENABLE_SERVICE_WORKER =
```
sirugh: This sounds like a const that would be used to enable the service worker.

Can you name it, `SERVICE_WORKER_ENABLED`? Would be more fitting, IMO.

packages/venia-concept/src/registerSW.js
```diff
@@ -1,10 +1,11 @@
-import { handleMessageFromSW } from '@magento/venia-ui/lib/util/swUtils';
+import {
+    SHOULD_ENABLE_SERVICE_WORKER,
+    handleMessageFromSW
+} from '@magento/venia-ui/lib/util/swUtils';
 
+// process.env.DEV_SERVER_SERVICE_WORKER_ENABLED is a string representation of a boolean value
```
sirugh: Move this comment to the exported const.

packages/venia-ui/lib/util/swUtils.js
revanth0212: If `true` it does not mean the SW is enabled yet. It just is a condition to enable or not. `SERVICE_WORKER_ENABLED` sounds like a boolean to say if the SW is enabled or not.

sirugh: I'm saying if the const is being used in places as a way of saying "Is the service worker enabled? If so, do X" then the proper name should be `SERVICE_WORKER_ENABLED`.

When you say "enabled yet" I think you mean that it hasn't been created or something. It's a difference of semantics and I'm not super worried about it. Just letting you know how it reads to me.

supernova-at: I was actually going to post a similar comment. In `registerSW`, `SHOULD_ENABLE_SERVICE_WORKER` reads fine because it's job is to then enable the service worker (if `true`).

In other places (such as `useProductImageCarousel` in this PR), it reads funny because it's doing something completely different than enabling the service worker when `SHOULD_ENABLE_SERVICE_WORKER` is `true`.

Unfortunately I can't really think of anything that reads well in both scenarios. In that case I say stick with the convention of `DEV_SERVER_SERVICE_WORKER_ENABLED` and just call it `SERVICE_WORKER_ENABLED` with the understanding that `registerSW` actually has to enable it first.

revanth0212: How about `VALID_SERVICE_WORKER_ENVIRONMENT`?

supernova-at:
Code looks good but I'm getting

Screen Shot 2019-11-26 at 8 24 41 AM

on Step 6. Maybe I don't have a `/sw.js` ?

revanth0212:
> Code looks good but I'm getting
> 
> Screen Shot 2019-11-26 at 8 24 41 AM
> 
> on Step 6. Maybe I don't have a `/sw.js` ?

This probably does not have anything to do with the changes. Also, I am not able to replicate the error in my local and also in the deployed version of this PR. Can you check if the SSL key you have is valid? It may have expired.

PR #2004, opened 2019-11-25
PWA-138: Re-factor getUserDetails to use GraphQL (useAwaitQuery Edition)

tjwiebell:
## Description
See PWA-138

## Related Issue

* [[PWA-138](https://jira.corp.magento.com/browse/PWA-138)] User details should be fetched from GraphQL

### Verification Steps

1. Create Account
- Go through Create Account workflow
- Verify user details are displayed correctly after successful creation (at bottom on nav bar)
2. Sign-In
- Logout and sign-in with the same account as above
- Verify you're redirected to MyAccount/Sign-out Page
- Verify user details still displayed correctly
3. Sign-out
- Sign-out
- Verify user details is replaced by signin button
4. MyAccount
- While signed in, clear browser cache while retaining local storage
- Verify user details are loading correctly on initial app render

## Checklist


packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -69,11 +69,9 @@ export const getUserDetails = () =>
             dispatch(actions.getDetails.request());
 
             try {
-                const userDetails = await request('/rest/V1/customers/me', {
-                    method: 'GET'
-                });
+                const { data } = await fetch();
```
sirugh: This prevents other actions from calling each other as only components will have fetchers.

packages/peregrine/lib/hooks/useAwaitQuery.js
```diff
@@ -0,0 +1,16 @@
+import { useCallback } from 'react';
+import { useApolloClient } from '@apollo/react-hooks';
+
+export const useAwaitQuery = query => {
+    const apolloClient = useApolloClient();
+
+    return useCallback(
+        variables => {
+            return apolloClient.query({
```
sirugh: If this hook is the only thing we take from the spike I still like it as we can await queries now :P Any idea on the timeline for the awaitable apollo hook?

Also do you know if this works for mutations?

packages/peregrine/lib/store/actions/user/asyncActions.js
sirugh: Also we would want to ensure we handle the error appropriately. Does `fetch` throw an error or would we have to destructure `error` from the result?

packages/peregrine/lib/hooks/useAwaitQuery.js
tjwiebell: There is another function `mutation` on the ApolloClient, but `useMutation` already returns a Promise, so shouldn't need to support in our custom hook.

packages/peregrine/lib/store/actions/user/asyncActions.js
tjwiebell: ...you're absolutely right, and it would be nasty to have to pass query invokers for nested action calls, but it would more clearly define those action APIs.

`fetch` does `throw` on error, so this code should function as-is (though may need to destructure `{ error }` in the `catch`.

packages/peregrine/lib/hooks/useAwaitQuery.js
sirugh: Another option we discussed was just using `useLazyQuery` and passing the callback function to the action ie:

```js
const [runQuery, result] = useLazyQuery(GET_USER_QUERY);

useEffect(() => { 
  getUserDetails(runQuery);
}, [getUserDetails, runQuery]);
```

tjwiebell: Discussed IRL: I would feel uncomfortable using this hook for anything but passing query invokers to Redux. I think hooks should be training developers to think in the render loop cycle, and adding async data fetching seems counter to that pattern.

packages/peregrine/lib/hooks/useAwaitQuery.js
```diff
@@ -0,0 +1,16 @@
+import { useCallback } from 'react';
+import { useApolloClient } from '@apollo/react-hooks';
+
+export const useAwaitQuery = query => {
```
sirugh: Let's document this fella :)

packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -69,11 +69,11 @@ export const getUserDetails = () =>
             dispatch(actions.getDetails.request());
 
             try {
-                const userDetails = await request('/rest/V1/customers/me', {
-                    method: 'GET'
+                const { data } = await fetchUserDetails({
+                    fetchPolicy: 'network-only'
```
sirugh: I'd leave a comment explaining why we use `network-only` here.

packages/venia-ui/lib/queries/getCustomer.graphql
```diff
@@ -0,0 +1,8 @@
+query getCustomer {
```
sirugh: I'd add a comment here indicating that this query depends on the `bearer` token being present otherwise it won't return anything (or it may just fail?)

packages/peregrine/lib/hooks/useAwaitQuery.js
tjwiebell: Fixed by d9eff35

packages/peregrine/lib/store/actions/user/asyncActions.js
tjwiebell: Fixed by d9eff35

packages/venia-ui/lib/queries/getCustomer.graphql
tjwiebell: Fixed by d9eff35

supernova-at:
# QA

- [x] Verification Steps Pass
- [x] Tested in Chrome
- [x] Tested in Safari
- [x] Tested on iOS simulator

packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -69,11 +69,13 @@ export const getUserDetails = () =>
             dispatch(actions.getDetails.request());
 
             try {
-                const userDetails = await request('/rest/V1/customers/me', {
-                    method: 'GET'
+                const { data } = await fetchUserDetails({
+                    // until we can investigate some odd behavior with apollo-cache-persist
+                    // not busting the cache on sign out, avoid caching user details.
```
supernova-at: @tjwiebell did we create a ticket for this?

PR #2010, opened 2019-12-02
[SPIKE] Module Replacement Plugin - Extensibility POC

revanth0212:
## Description

Implement a Module Replacement system in `pwa-buildpack` that can be used to inject code into the bundle at build time. To prove the point, this PR replaces the venia logo with a goose logo and the loading component with the `BounceLoader` component from the [react-spinners](https://www.npmjs.com/package/react-spinners) package.

Both these changes are provided as extensions using a sample package called [sample-venia-loader-icon](https://github.com/revanth0212/sample-venia-loader-icon).

Extensions can also be added while creating a PWA App using the `create-pwa` utility provided as part of the PWA Studio package.

To call a package an extension, it has to expose `extensions.json` in its root structure which has the details that `buildpack` will need to replace modules from the storefront.

## Related Issue
Closes [PWA-209](https://jira.corp.magento.com/browse/PWA-209)

### Verification Steps

For an existing project

1. Go to PWA Studio root folder.
2. Install the extension using `yarn add sample-venia-loader-icon`
3. Build and run the app using `yarn build && yarn stage:venia` or `yarn watch:venia`.
4. When you open the app in the browser, you should see that the logo and the loading icon have changed.

For a new project

1. Create a new project using the `create-pwa` tool by running this command `DEBUG_PROJECT_CREATION=1 ./pwa-studio/packages/create-pwa/bin/create-pwa`. Setting `DEBUG_PROJECT_CREATION` to `1` will take bundles for `@magento` packages from local rather than from npm.
2. This will start the `inquirer` which will ask you questions about the sample storefront you are trying to create.
3. Once you reach the question about extensions, type `sample-venia-loader-icon`. That is the sample storefront extension that we trying to install in the app.
4. Once the `create-pwa` tool is done with its magic, you will have the app structure ready and you can navigate into that project and run `yarn build && yarn start` to build and start the app.
5. Go to the URL provided in the terminal and you should see the goose logo instead of the default venia logo. Also you should see the loading icon change to the `BounceLoader` from `sample-venia-loader-icon`.

image
Logo File downloaded is different in both versions.

image
The logo shown on the loading screen is different.

image
The logo shown in the header section is different.

## Checklist
None. This is a SPIKE.


packages/venia-concept/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+    "logo.svg": "node_modules/new_logo/logo.svg"
```
sirugh: So the plugin literally just swaps the resource with another? Interesting.

Could you extend this to a component/module rather than an asset? Something that takes props would be interesting. The API would need to be rock solid. Imagine replacing the entirety of checkout!

revanth0212: That is exactly what I am working on right now. Trying to replace the `Toast` component using an `npm` module that takes the same props and just changes the way UI looks.

But the ultimate plan is to create an environment for HTML extensions similar to the ones that exist in the current marketplace.

revanth0212:
Don't review this PR yet. Still working on it 😄

revanth0212:
Closing SPIKE.

PR #2012, opened 2019-12-04
[PWA-147] Sign Out should revoke customer token using graphql mutation

tjwiebell:
## Description

App should revoke Customer token on sign out using GraphQL mutation.

## Related Issue

* [[PWA-147](https://jira.corp.magento.com/browse/PWA-147)] Sign Out should revoke customer token using graphql mutation

### Verification Steps

1. Signin with a valid customer/user
2. Grab the oauth token from local storage
3. If you see customer details in the app, the token is confirmed working, but worth sanity checking and running an auth'd `getCustomer` query in a GraphQL client, being sure to append `Authorization: Bearer <token>` header to request
4. Signout - verify you no longer see customer details, and page should refresh and take you back to home
5. Using a GraphQL client, attempt the auth'd `getCustomer` query again. You should now get `The current customer isn't authorized.`

## Checklist


packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -54,9 +60,8 @@ export const useAuthModal = props => {
     }, [showMyAccount]);
 
     const handleSignOut = useCallback(() => {
-        // TODO: Get history from router context when implemented.
-        signOut({ history: window.history });
-    }, [signOut]);
+        signOut({ history, revokeToken });
```
sirugh: Should we be calling resetStore here?

sirugh: It may be worth rewriting this to something like 

```js
await signOut({ revokeToken });
await apolloClient.resetStore();
refresh({ history }); // from router-helpers or window?
```

packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -7,8 +7,11 @@ import actions from './actions';
 
 const storage = new BrowserPersistence();
 
-export const signOut = ({ history }) => async dispatch => {
-    // Sign the user out in local storage and Redux.
+export const signOut = ({ history, revokeToken }) => async dispatch => {
+    // Send mutation to revoke token.
+    await revokeToken();
```
sirugh: Technically this will return an error if called without a bearer token. Do we care? Probably not, but I would possibly catch and do something with the error just incase.

```json
{
  "errors": [
    {
      "message": "The current customer isn't authorized.",
      "category": "graphql-authorization",
      "locations": [
        {
          "line": 2,
          "column": 5
        }
      ],
      "path": [
        "revokeCustomerToken"
      ]
    }
  ],
  "data": {
    "revokeCustomerToken": null
  }
}
```

tjwiebell: Fixed in 1f680c4

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
tjwiebell: Went ahead and did this in 1f680c4; I like the idea of not having any effects in actions.

PR #2014, opened 2019-12-04
[bug] Footer position value defaulted to static which broke z-index styling on Edge

sirugh:
## Description

We applied a z-index to the checkout form and a lesser z-index to the cart mask but we forgot to modify the `position` of the form from the default of `static`. A statically positioned element disregards all positioning styles including `z-index`. Chrome must have some ChromeMagic™ that fixed it in browser so we never caught the issue.

## Related Issue

Closes [PWA-175](https://jira.corp.magento.com/browse/PWA-175).

### Verification Steps

1. Open this PR on Edge, add a product to cart and click the checkout button. The cart mask should apply but not over the form.
2. You should be able to interact with the checkout steps. Go through each.

## Checklist


dpatil-magento:
QA Pass, verified checkout flow on Win 10 Edge, Mac Chrome, FF, Safari, Android, iOS.

PR #2015, opened 2019-12-05
[feature] Use removeItemFromCart mutation in minicart product

sirugh:
## Description

Replaces the REST removal endpoint with use of the graphql mutation.
I also removed some prop threading of the action. Now the `useProduct` talon in minicart directly imports the cart context to use the action.

## Related Issue

Closes [PWA-142](https://jira.corp.magento.com/browse/PWA-142).

### Verification Steps

1. Add an item to your cart.
2. Remove the item.
3. Sign in and repeat steps 1 and 2.

## Checklist

### DIFF
--- a/packages/peregrine/lib/store/actions/cart/asyncActions.js
+++ b/packages/peregrine/lib/store/actions/cart/asyncActions.js
@@ -233,9 +233,8 @@ export const removeItemFromCart = payload => {
     return async function thunk(dispatch, getState) {
         dispatch(actions.removeItem.request(payload));
 
-        const {
-            cart: { cartId }
-        } = getState();
+        const { cart } = getState();
+        const { cartId } = cart;
 
         try {
             await removeItem({
--- a/packages/peregrine/lib/store/actions/cart/asyncActions.js
+++ b/packages/peregrine/lib/store/actions/cart/asyncActions.js
@@ -228,50 +228,29 @@ export const updateItemInCart = (payload = {}) => {
 };
 
 export const removeItemFromCart = payload => {
-    const { item, fetchCartId } = payload;
+    const { item, fetchCartId, removeItem } = payload;
 
     return async function thunk(dispatch, getState) {
         dispatch(actions.removeItem.request(payload));
 
-        const { cart, user } = getState();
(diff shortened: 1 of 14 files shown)


packages/venia-ui/lib/queries/removeItem.graphql
```diff
@@ -0,0 +1,13 @@
+mutation remoteItem($cartId: String!, $itemId: Int!) {
+    removeItemFromCart(input: { cart_id: $cartId, cart_item_id: $itemId }) {
+        cart {
```
tjwiebell: Seems unnecessary to make the server fetch this data if we just ignore it. Did you intend to pass to cartDetails.receive and remove the getCartDetails action call?

sirugh: You cannot ignore it, AFAICT graphQl requires you have something.

However that's not necessarily a bad thing, as you suggest. We can probably do away with the `getCartDetails` call so long as this data provides whatever we need.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -228,50 +228,29 @@ export const updateItemInCart = (payload = {}) => {
...
 
-        const { cart, user } = getState();
-        const { isSignedIn } = user;
-        let isLastItem = false;
-
-        if (cart.details && cart.details.items_count === 1) {
-            isLastItem = true;
-        }
+        const {
+            cart: { cartId }
```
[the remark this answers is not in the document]
sirugh: @jimbo how likely are we to automate the checks for these rules :)

packages/venia-ui/lib/queries/removeItem.graphql
tjwiebell: You got half of my point, but not worth another cycle. IF response data is required and we're not doing anything with it right now, might be best just to request `cart { id }` or something more minimal so the backend isn't doing all this extra work for nothing.

sirugh: Fair. Whomever does `getCartDetails` can just go through all this :)

PR #2017, opened 2019-12-05
Use updateItem mutations

sirugh:
## Description

Updates the `cartOptions` to use the `updateCartItems` mutation for simple products.
As there is no update mutation for configurable products we just remove then add the new configuration in place of a singular update call.
I also added some logic to disable the `Update Cart` button until either the quantity or the options have been touched.

## Related Issue

Closes [PWA-143](https://jira.corp.magento.com/browse/PWA-143).

#### Simple product
1. Add a simple product to the cart.
2. Open the edit form. The update button should be disabled.
3. Change the quantity. The update button should enable.
4. Click the update button. The product should update in the cart.

#### Configurable product
1. Add a configurable product to cart.
2. Open the edit form. The update button should be disabled.
3. Change the quantity. The update button should enable.
4. Click the update button. The product should update in the cart.
5. Open the edit form. The update button should be disabled.
6. Change an option like color or size. The update button should enable.
7. Click the update button. The product should update in the cart.

## Checklist


packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -134,51 +134,77 @@ export const addItemToCart = (payload = {}) => {
...
-                if (!cartId) {
-                    const missingCartIdError = new Error(
-                        'Missing required information: cartId'
-                    );
-                    missingCartIdError.noCartId = true;
-                    throw missingCartIdError;
-                }
-
-                cartEndpoint = `/rest/V1/guest-carts/${cartId}/items/${cartItemId}`;
+            if (productType === 'simple') {
```
sirugh: This check is fragile. `productType` uses the REST type. When we switch `getCartDetails` to GQL this may break so we should be aware.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -134,51 +134,77 @@ export const addItemToCart = (payload = {}) => {
...
+                );
+                await dispatch(
+                    addItemToCart({
+                        ...payload,
+                        // For now, rewrite productType since the REST cart_item
+                        // uses "configurable" and "simple" string values.
+                        productType: 'ConfigurableProduct',
+                        fetchCartId
+                    })
+                );
```
sirugh: The above actions both call `getCartDetails`. The result is a weird double fetch. There's no way around this unless we break the calls to `getCartDetails` out of those functions. Probably worth moving it to the callback invoker as a "post action" action but not worth changing in this PR.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -134,51 +134,78 @@ export const addItemToCart = (payload = {}) => {
...
-                    const missingCartIdError = new Error(
-                        'Missing required information: cartId'
-                    );
-                    missingCartIdError.noCartId = true;
-                    throw missingCartIdError;
-                }
-
-                cartEndpoint = `/rest/V1/guest-carts/${cartId}/items/${cartItemId}`;
+            if (productType === 'ConfigurableProduct') {
+                await dispatch(
```
sirugh: We _must_ remove and then add. If changes are made to quantity only and we add first then remove we end up deleting the entire cart item.

packages/peregrine/lib/store/actions/cart/asyncActions.js
tjwiebell: I would be fine with this, but it does cause a weird flash of an empty cart, and then the add action finishes and displays the updated cart. I see that the `isUpdatingItem` state is being drilled down to these components, but `isLoading` seems to be causing some extra renders.

I tried changing some logic in `MiniCart/body`, but didn't work as I wanted; only displaying loading indicator `if (isLoading && !isUpdating)`. The only option I see is to defer these cart renders, but agree we should avoid a big re-factor in this scope. Any ideas to prevent that double render on update?

PR #2018, opened 2019-12-06
Refactor checkout workflow to set shipping address/get shipping methods with graphql

sirugh:
## Description

Uses graphQL to set shipping address and, in the same mutation, receive the available shipping methods for the selected address.

Also removes email from forms for authed users as the email is already attached to cart.
For guest users it will set the guest email on the cart when submitting the address form.

## Related Issue

Closes [PWA-145](https://jira.corp.magento.com/browse/PWA-145).

### Verification Steps

1. Go through checkout -- it should complete successfully. Verify that the REST submissions still work.

## Checklist


packages/peregrine/lib/store/reducers/checkout.js
```diff
@@ -15,7 +15,7 @@ const initialState = {
     receipt: {
         order: {}
     },
-    shippingAddress: null,
+    shippingAddress: {},
```
sirugh: I may be able to revert this change, though we probably can default this to `{}` safely as it is set to that in `beginCheckout`.

packages/venia-ui/lib/components/Checkout/overview.js
```diff
@@ -57,11 +56,7 @@ const Overview = props => {
                     onClick={handleAddressFormClick}
                     showEditIcon={hasShippingAddress}
                 >
-                    <ShippingAddressSummary
-                        classes={classes}
-                        hasShippingAddress={hasShippingAddress}
-                        shippingAddress={shippingAddress}
```
sirugh: It may be overkill to have the summaries import this through context. Can probably at least do away with the threading down to the overview and to the editable forms.

packages/venia-ui/lib/components/Checkout/__tests__/__snapshots__/addressForm.spec.js.snap
```diff
@@ -422,9 +422,7 @@ exports[`renders validation block with message if address is incorrect 1`] = `
     </h2>
     <div
       className="validationMessage"
-    >
-      Oops
-    </div>
+    />
```
supernova-at: I think this is a legit regression. An error message should be rendered.

packages/venia-ui/lib/components/Checkout/addressForm.js
```diff
@@ -28,16 +31,24 @@ const fields = [
 ];
 
 const AddressForm = props => {
-    const { countries, error, isSubmitting, onCancel, onSubmit } = props;
```
supernova-at: Okay, this explains the snapshot discrepancy from earlier. Seems like this component no longer takes an `error` prop; the talon provides one instead.

Therefore I think we should:
  1. Remove `error` from this component's `propTypes`
  1. Fix up or remove the "renders validation block with message if address is incorrect" test in `addressForm.spec.js`. It's passing in an `error` prop right now. Perhaps we mock the talon's response instead.
  1. Update the snapshot again to ensure we're rendering an error message

packages/venia-ui/lib/components/Checkout/form.css
```diff
@@ -16,6 +16,12 @@
     text-transform: capitalize;
 }
 
+.disabledPrompt {
+    composes: informationPrompt;
+    /* TODO: Is this the best way to signify disabled? Contrast seems light on text. */
+    color: rgb(var(--venia-grey-dark));
```
supernova-at: I'd agree maybe we can improve the contrast but since this is consistent with the handful of other "disabled" styles we use in Venia today, I'm 👍 right now.

packages/venia-ui/lib/components/Checkout/shippingAddressSummary.js
```diff
@@ -1,8 +1,16 @@
 import React, { Fragment } from 'react';
 import { array, bool, shape, string } from 'prop-types';
 
+import { useCheckoutContext } from '@magento/peregrine/lib/context/checkout';
+
+import isObjectEmpty from '../../util/isObjectEmpty';
```
supernova-at: In another place we imported this from Peregrine - do we want to do that here too?

packages/venia-ui/lib/components/Checkout/shippingMethodSummary.js
```diff
@@ -1,15 +1,28 @@
 import React, { Fragment } from 'react';
-import { bool, shape, string } from 'prop-types';
+import { shape, string } from 'prop-types';
+
+import { useCheckoutContext } from '@magento/peregrine/lib/context/checkout';
+
+import isObjectEmpty from '../../util/isObjectEmpty';
```
supernova-at: Can import from Peregrine

packages/venia-ui/lib/components/Checkout/shippingMethodSummary.js
```diff
@@ -1,15 +1,28 @@
...
 
 const ShippingMethodSummary = props => {
-    const { classes, hasShippingMethod, shippingTitle } = props;
+    const { classes } = props;
+
+    const [
+        { shippingAddress, shippingMethod, shippingTitle }
+    ] = useCheckoutContext();
```
supernova-at: 💯

packages/venia-ui/lib/components/Checkout/form.css
sirugh: Ok I'll remove the TODO then :)

dpatil-magento:
QA pass but waiting for https://github.com/magento/pwa-studio/pull/2029 to get merged first.

PR #2027, opened 2019-12-12
Switches products filter by name to use match input type instead of eq

sirugh:
## Description

In 2.3.4, graphQL updated the product filter for name to use a `match` input rather than `eq`.

## Related Issue

Closes [PWA-233](https://jira.corp.magento.com/browse/PWA-233).

### Verification Steps

1. Set your backend in .env to a 2.3.4 instance
2. Add a **Configurable** item to your cart
3. Open the cart and click to edit the item. It should open the options form in the cart.
4. Check the network tab for the /graphql query. Make sure the `getProductDetailsByName` query has filter with `match` and not `eq`.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/queries/getProductDetailByName.graphql
+++ b/packages/venia-ui/lib/queries/getProductDetailByName.graphql
@@ -1,5 +1,5 @@
 query productDetailByName($name: String, $onServer: Boolean!) {
-    products(filter: { name: { eq: $name } }) {
+    products(filter: { name: { match: $name } }) {
         items {
             __typename
             id


sirugh:
@dpatil-magento We should only merge this against the release branch.

dpatil-magento:
@sirugh I think we should be able to merge to `develop` , dont see a reason why we need to merge only to `releasebranch`
cc - @tjwiebell

dpatil-magento:
Works only with 2.3.4 not with 2.3.3, if this is the only option then we need to merge against release branch only.

packages/venia-ui/lib/queries/getProductDetailByName.graphql
```diff
@@ -1,5 +1,5 @@
 query productDetailByName($name: String, $onServer: Boolean!) {
-    products(filter: { name: { eq: $name } }) {
+    products(filter: { name: { match: $name } }) {
```
sirugh: We need to talk about what this means. The `filter.name` type changed from "equals" type in 2.3.3 to "match" type in 2.3.4. In order to support both versions we would have to somehow switch the variable prop dynamically based on the backend _or_ GraphQL team will have to provide some method which works on both versions.

cc @nishantjariwala @awilcoxa @cherdman

cherdman: @cpartica

sirugh:
There may be a better solution. Don't use `name` at all! See this [commit](https://github.com/magento/pwa-studio/pull/2029/commits/bbcd42259e19124ec0aa969c12bf4b6980da9a66) in the PR where we switch to get cart details.

It uses `sku` for filtering, which is an `eq` type and works in both 2.3.3 and 2.3.4!

sirugh:
This can be resolved by #2029, specifically using `SKU` to search.

PR #2028, opened 2019-12-12
Removes id from getCustomer response object

sirugh:
## Description

`id` is no longer returned by the customer query. The check can just use a different field from the customer response object.

## Related Issue

Closes [PWA-228](https://jira.corp.magento.com/browse/PWA-228).

### Verification Steps

1. Reset local storage
2. Sign in. It should show you the normal authed view after and it should not infinitely spin.
3. Sign out and sign in again -- make sure nothing broke because of the removed `id`.

## Checklist


sirugh:
This can be merged to `develop` as it is compat with 2.3.3 and 2.3.4

PR #2035, opened 2019-12-17
[bugfix]: Adds meta descriptions to root pages

sirugh:
## Description

This uses the `Meta` component from `react-head` to provide the description meta tag.

I updated all the queries to return it regardless of whether we're on server or not. I then configured each component to utilize the returned value in the rendered HTML.

If lighthouse still complains with this change that means the sample data/backend data needs to be updated. 

[Image from Gyazo](https://gyazo.com/4fb9c7afc48280f71a2b28a2b0130dc9)

I'm not sure if JS crawlers wait for/allow this sort of async update of meta description. We could possibly provide a default value?

## Related Issue

Closes [PWA-217](https://jira.corp.magento.com/browse/PWA-217).

### Verification Steps

1. Go to the app, check the HTML for a meta description tag.
2. Navigate to the category and check that it is still there.
3. Do the same for a product.

You may notice that `content` is empty for some pages. If that happens it is because the backend is returning `null`. Check the response for the relevant query to see if that is true.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/RootComponents/CMS/cms.js
+++ b/packages/venia-ui/lib/RootComponents/CMS/cms.js
@@ -5,6 +5,7 @@ import { fullPageLoadingIndicator } from '../../components/LoadingIndicator';
 import RichContent from '../../components/RichContent';
 import { number } from 'prop-types';
 import CategoryList from '../../components/CategoryList';
+import { Meta } from '../../components/Head';
 
 const CMSPage = props => {
     const { id } = props;
@@ -36,7 +37,15 @@ const CMSPage = props => {
             return <RichContent html={data.cmsPage.content} />;
         }
 
-        return <CategoryList title="Shop by category" id={2} />;
+        return (
+            <>
+                <Meta
+                    name="description"
+                    content={data.cmsPage.meta_description}
+                />
+                <CategoryList title="Shop by category" id={2} />
+            </>
+        );
     }
     return null;
 };
--- a/packages/venia-ui/lib/RootComponents/CMS/cms.js
+++ b/packages/venia-ui/lib/RootComponents/CMS/cms.js
@@ -28,13 +28,16 @@ const CMSPage = props => {
     }
 
     if (data) {
+        let content;
         // Only render <RichContent /> if the page isn't empty and doesn't contain the default CMS Page text.
         if (
             data.cmsPage.content &&
             data.cmsPage.content.length > 0 &&
             !data.cmsPage.content.includes('CMS homepage content goes here.')
         ) {
-            return <RichContent html={data.cmsPage.content} />;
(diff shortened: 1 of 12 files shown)


packages/venia-ui/lib/RootComponents/CMS/cms.js
```diff
@@ -36,7 +37,15 @@ const CMSPage = props => {
             return <RichContent html={data.cmsPage.content} />;
         }
 
-        return <CategoryList title="Shop by category" id={2} />;
+        return (
+            <>
+                <Meta
+                    name="description"
+                    content={data.cmsPage.meta_description}
```
sirugh: @davemacaulay do you guys want this in the above conditional?

davemacaulay: It makes sense to me to have it for all resolved pages, the above conditional was to ensure the category list shows up on instances without Page Builder.

sirugh: Right -- I'm considering just adding this to all pages but this specific changeset was supposed to address the lighthouse score dinging us for not having the description on the root page (which happens to be CMS).

davemacaulay: It should then be included on all CMS pages, correcting meta data for all pages is more beneficial then a lighthouse score fix imo.

sirugh: Agreed :) I went ahead and added it to everything.

packages/venia-ui/lib/queries/getProductDetailBySku.graphql
```diff
@@ -38,9 +38,6 @@ query productDetailBySku($sku: String, $onServer: Boolean!) {
                     }
                 }
             }
-            meta_title @include(if: $onServer)
-            meta_keyword @include(if: $onServer)
-            meta_description @include(if: $onServer)
```
sirugh: These were unused. This query is only used for the cart and shouldn't need these details anyways.

packages/venia-ui/lib/fragments/productDetails.graphql
```diff
@@ -21,7 +21,7 @@ fragment ProductDetails on ProductInterface {
     }
     meta_title @include(if: $onServer)
     meta_keyword @include(if: $onServer)
-    meta_description @include(if: $onServer)
```
sirugh: We now want `meta_description` regardless of whether we are doing SSR. I assume there is still a path to use this in SSR as there are templates in our repo that use these meta fields but I couldn't figure out how exactly.

sirugh:
> I'm also curious what the {category, cmspage, product}-shell.mst templates are there for, which appear to fulfill this requirement; but not a blocker.

I think they are related to SSR. @zetlen may know?

PR #2036, opened 2019-12-17
[SPIKE] Use response objects instead of making numerous details requests

sirugh:
##  Description

What does it look like if we use GQL to return the data we need to render the cart rather than calling a new fetch after each mutation? This!

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist

### DIFF
--- a/packages/peregrine/lib/store/actions/cart/asyncActions.js
+++ b/packages/peregrine/lib/store/actions/cart/asyncActions.js
@@ -1,5 +1,4 @@
 import BrowserPersistence from '../../../util/simplePersistence';
-import { toggleDrawer } from '../app';
 import actions from './actions';
 
 const storage = new BrowserPersistence();
@@ -45,14 +44,7 @@ export const createCart = payload =>
     };
 
 export const addItemToCart = (payload = {}) => {
-    const {
-        addItemMutation,
-        fetchCartDetails,
-        fetchCartId,
-        item,
-        quantity,
-        parentSku
-    } = payload;
+    const { addItemMutation, fetchCartId, item, quantity, parentSku } = payload;
 
     const writingImageToCache = writeImageToCache(item);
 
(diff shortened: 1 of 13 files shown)


packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -108,14 +92,6 @@ export const addItemToCart = (payload = {}) => {
                     })
                 );
 
-                // and fetch details
-                await dispatch(
-                    getCartDetails({
```
sirugh: Do we need to fetch details on errors anymore? We remove/create cart so the state should be reset to empty anyways. I removed this from the relevant actions.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -224,27 +222,27 @@ export const useProductFullDetail = props => {
...
             }
 
-            addItemToCart({
+            await addItemToCart({
                 ...payload,
                 addItemMutation,
-                fetchCartDetails,
                 fetchCartId
             });
+            toggleDrawer('cart');
```
sirugh: Importantly, we missed this in #2017 and we were calling `toggleDrawer` when we updated a configurable item, after the add.

packages/venia-ui/lib/queries/addConfigurableProductsToCart.graphql
```diff
@@ -4,7 +4,7 @@ mutation addConfigurableProductToCart(
     $sku: String!
     $parentSku: String!
 ) {
-    addConfigurableProductsToCart(
+    response: addConfigurableProductsToCart(
```
sirugh: Product add uses different mutations depending on product type. Rename the responses here to simplify the action code for destructuring the response.

packages/venia-ui/lib/queries/addConfigurableProductsToCart.graphql
```diff
@@ -21,8 +21,33 @@ mutation addConfigurableProductToCart(
                 product {
                     name
                     sku
+                    small_image {
```
sirugh: Since we can't use fragments, each of the cart mutations/queries uses this identical response object. That's lame!

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -72,22 +64,14 @@ export const addItemToCart = (payload = {}) => {
...
-                    fetchCartId,
-                    fetchCartDetails
-                })
-            );
-            await dispatch(toggleDrawer('cart'));
-            dispatch(actions.addItem.receive());
+            // The response contains the new updated cart state.
+            const { response } = data;
+            const { cart: details } = response;
+            dispatch(actions.addItem.receive({ details }));
```
sirugh: Instead of re-spreading the `details` after each mutation, we could try to only add/remove the relevant item in the cart. With that logic in place we could take advantage of [optimistic UI](https://www.apollographql.com/docs/angular/features/optimistic-ui/) which would allow us to update the UI without having to wait for GQL to respond. We could still display the error state if the server eventually responded with one, but we wouldn't be stuck behind the network request.

PR #2037, opened 2019-12-18
Search Result Counts Match

supernova-at:
## Description

There are two places we show the count of a search result: in the autocomplete dropdown and on the search page itself. Unfortunately, these two places were calculating their values differently and so sometimes did not match.

This PR:
  * Updates the way the autocomplete dropdown was calculating its value
  * Updates the search results page to add pagination so the number of items matches the number of results indicated

## Related Issue

Closes JIRA-97.

### Verification Steps

1. Perform a search for "Tops Dresses"
1. Note how many items the autocomplete dropdown says have been found
1. Press `<Enter>` to execute the search and continue to the search page
1. Note how many items the search page says have been found
1. See that the two numbers match

### Autocomplete

Screen Shot 2019-12-18 at 11 03 07 AM

### Search Page

Screen Shot 2019-12-18 at 11 03 16 AM

## Checklist


packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -35,7 +35,7 @@ const SearchPage = props => {
     }
 
     const { products } = data;
-    const { filters, total_count, items } = products;
+    const { filters, items } = products;
```
tjwiebell: It turns out that `total_count` was actually accurate here, we're just not passing `pageSize` to `productSearch.graphql`, so its falling back on the default value of 20. Given that pagination is missing from this component, easy to see why this wasn't making sense.

It seems preferable to fix the count in the suggestion component, and add pagination to `searchPage`. Given that's a much larger undertaking, I would be fine with fixing the count on suggestion, and creating a Story to add pagination to search.

supernova-at: Ah, gotcha - I was going to dig into that as a follow up, thanks.

supernova-at: I didn't go with `total_count` because we're not showing that number of items. To a user, there'd still be a discrepancy until that pagination to search ticket is completed.

tjwiebell: Sure. I'll settle for a comment explaining we're only displaying the first twenty results, and the counts will be updated to be accurate once pagination is added.

supernova-at: I may or may not have just gone ahead and implemented Pagination.... https://github.com/magento/pwa-studio/pull/2037/commits/95ab17bdf0473277b1f2c70d39081d0e23720614

😄

dpatil-magento:
@supernova-at  Observations - 
1. Get results for say "Tops" and be on page 2 or 3.  >> Now enter new search which has only 1 page results say "VT10"  >> Issue - No results are displayed.
2. Get results for say "Tops" and be on page 2 >> Now enter new search which has multiple search products  say "Dresses"  >> Issue - Dresses results displayed but with page 2 in focus instead of page 1.

dpatil-magento:
@supernova-at Same issue exists when selecting from suggestion box. 
Enter "Tops" be on page 3 of search results >> Click back in search text field > Select "tops in Tops" suggested search. Issue - Search results does not load as it has products only for 2 pages and we searched from page 3.

supernova-at:
@dpatil-magento great job!

Fixed in https://github.com/magento/pwa-studio/pull/2037/commits/19d17c6f8bc5954004eefaa049a3f9b851fec580.

dpatil-magento:
QA pass. @tjwiebell You can merge once it passes the review.

PR #2039, opened 2019-12-18
[PWA-229] [DX] VSCode Jest Plugin Broken

tjwiebell:
## Description

After we added automated testing for service workers, the popular Jest plugin for vscode stopped working, due to the --experimental-worker flag that must be appended to test execution. This plugin essentially does what test:dev does, runs the whole suite and then watches the file system for changes, but gives you feedback right in the code base.

While it would be okay to give developers instructions on how to fix, it would probably be better to detect if this flag was passed, and not run these tests, instead of just failing.

Error Message
```
Error: Cannot find module 'worker_threads'
Starting Jest in Watch mode failed too many times and has been stopped.
```
## Related Issue

* [[PWA-229](https://jira.corp.magento.com/browse/PWA-229)] [DX] VSCode Jest Plugin Broken
### Verification Steps

1. Install Jest plugin for VSCode and enable
2. In the Output panel, verify that test runner starts up w/o error and completes

## Checklist


tjwiebell:
@revanth0212 - would like if you looked at this. While this does solve the problem I was having in vscode, I noticed that the executed test count doesn't change if `worker_threads` is missing.

jest.config.js
```diff
@@ -6,7 +6,14 @@
  */
 const path = require('path');
 
-const worker = require('worker_threads');
+let worker;
+try {
+    worker = require('worker_threads');
+} catch (e) {
+    console.log(
+        'Experimental worker flag missing, skipping execution of ServiceWorker tests.'
```
sirugh: Does this actually cause the tests to skip? How does that work :o

tjwiebell: Yeah, referenced in my comment above, I have no idea what effect this dependency had on test execution; I see the same counts regardless if `--experimental-worker` is passed or not. Will need @revanth0212 to provide some insight.

revanth0212: I remember `--experimental-worker` was need to allow node to provide `worker_threads` as a module. `worker_threads` was needed to test  Service Worker stuff. We were in the phase of mocking SW stuff instead of using the node workers, but if you are telling me that the tests are not failing even if we remove `--experimental-worker`, then maybe were have successfully mocked SW. In that case, we can just simply remove the flag.

revanth0212: > Does this actually cause the tests to skip? How does that work :o

It doesn't, the tests will still run with `MessageChannel` set to `{}`.

tjwiebell: So there is no difference between having this code in place, and just hard-coding `MessageChannel: {}` in the globals object (just looking at counts, haven't looked at actual coverage diffs). I'm not familiar with these tests, would it better to just merge this as-is, or put in this hard-coding of `MessageChannel` and remove all the `--experimental-worker` flags?

revanth0212: I would say, lets merge with this change. `MessageChannel` is used in multiple tests. Mocking it would be hard. If tests are not failing, maybe I have to update the tests to be more precise. They have to fail if `MessageChannel` is not provided or I would have mocked it already 😜

PR #2041, opened 2019-12-18
[spike] Rugh/spike cart cache

sirugh:
An exploration of what it looks like to use apollo/graphql cache instead of redux. We still use the redux store for things like `cartId` and shared UI state, but in this PR the cart details come from graphql!

Some key things:

* Instead of spreading `fetchDetails` queries around the app I put it in the cart context provider. This is a similar pattern to what I had spiked before. Alternatively, all components that care about cart state can make fetch calls with specific queries using the cart id. It should still use the gql cache.

* `cartTrigger` is now _much_ smaller. It no longer has an effect that requests details yet it still gets updated whenever cart state changes!

* We still have to remove/create cart within actions. If we did this at, say, the cart context as an effect for not having a `cartId` we would have the problem where our retries didn't have the cartId we propagate to the redux store.

TODO:

* `getCartDetails` never cached and we always cleared the stale cart after signIn. We will need to figure out a way to do this. Probably something related to `update` callback to the signIn query.

### DIFF
--- a/packages/peregrine/lib/context/cart.js
+++ b/packages/peregrine/lib/context/cart.js
@@ -1,14 +1,17 @@
-import React, { createContext, useContext, useMemo } from 'react';
+import React, { createContext, useContext, useEffect, useMemo } from 'react';
+import { useLazyQuery } from '@apollo/react-hooks';
 import { connect } from 'react-redux';
 
 import actions from '../store/actions/cart/actions';
 import * as asyncActions from '../store/actions/cart/asyncActions';
 import bindActionCreators from '../util/bindActionCreators';
 
+// TODO Where would we store this?
+import GET_CART_DETAILS_QUERY from '../../../venia-ui/lib/queries/getCartDetails.graphql';
+
 const CartContext = createContext();
 
-const isCartEmpty = cart =>
-    !cart || !cart.details.items || cart.details.items.length === 0;
+const isCartEmpty = cart => !cart || !cart.items || cart.items.length === 0;
 
 const getTotalQuantity = items =>
     items.reduce((total, item) => total + item.quantity, 0);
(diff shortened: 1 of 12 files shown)


packages/peregrine/lib/talons/App/useApp.js
```diff
@@ -40,6 +46,14 @@ export const useApp = props => {
...
 
     const [isHTMLUpdateAvailable, setHTMLUpdateAvailable] = useState(false);
+    const [fetchCartId] = useMutation(CREATE_CART_MUTATION);
+    const [, { createCart }] = useCartContext();
+
+    useEffect(() => {
+        createCart({
+            fetchCartId
+        });
+    }, [createCart, fetchCartId]);
```
sirugh: If we like the idea of fetching details within the context provider we may want to consider doing this there as well since it just makes that whole slice cleaner.

packages/peregrine/lib/context/cart.js
```diff
@@ -1,41 +1,58 @@
...
     const { actions, asyncActions, cartState, children } = props;
 
+    const { cartId } = cartState;
+    const [fetchCartDetails, { data = {} }] = useLazyQuery(
+        GET_CART_DETAILS_QUERY
+    );
+
+    useEffect(() => {
+        if (cartId) {
+            fetchCartDetails({
```
sirugh: Instead of fetching here we could, like I did in cart trigger, fetch just data we need in the components.

packages/peregrine/lib/talons/Header/useCartTrigger.js
```diff
@@ -1,32 +1,46 @@
...
+    const [getItemQuantities, { data }] = useLazyQuery(GET_ITEM_QUANTITY_GQL);
 
     useEffect(() => {
-        getCartDetails({
-            fetchCartId,
-            fetchCartDetails
-        });
-    }, [fetchCartDetails, fetchCartId, getCartDetails]);
+        if (cartId) {
+            getItemQuantities({
```
sirugh: This is only called once! After that, each time the cart is updated (as in adding/removing a product) the data from the lazy query is updated as well!

sirugh:
Posted in JIRA but also here for posterity:

Outcome:

There are multiple ways we can update data without requiring actions to dispatch changes to the store.

1) Using `refetchQueries` we can call a query to update cart data store like "getCartData", after making some mutation like "addItemToCart". Any query that is also listening for even a piece of the data requested by getCartData will be automatically updated. 

2) Instead of using refetchQueries we can request _all_ necessary data as part of the mutation response. "addItemToCart" returns a cart type object which means we can add an item and then request things like price, item counts, etc, that may have changed.

PR #2042, opened 2019-12-19
[SPIKE] [JIRA-250] Sign out alert extension using Peregrine Tap Hooks.

revanth0212:
## Description
This PR tries to approach the peregrine extensibility topic using a page from webpack's plugin system.

This framework is implemented using webpack's `tapable` package. The `tapable` package provides a sophisticated API to implement call handlers and interceptors.

**The scenario that is used as the basis for this SPIKE goes like this:**
`Assume there is an extension provider who would like to write an extension that when added to Venia, provides an alert to the user once they have clicked on the signout button.`

**Solution:**
We add a Sync hook to the `useMyAccount` which will also be provided to all extensions (in our case the `signoutExtension.js`) so they can tap and register a callback for it. When the user clicks on the sign out button, the sync hook will be called and all the register extensions will get a chance to handle it.

`Tapable` is a very complex looking package at first but it is very very useful. (Literally webpack runs on it  😐).

We can come up with a formalized way to handle and provide these tapable hooks if the team believes this is the right approach to handle peregrine extensibility.

Feel free to drop your thoughts on this PR. Cheers guys.

## Related Issue
Closes [JIRA-250](https://jira.corp.magento.com/browse/PWA-250)

### Verification Steps
1. Go to the deployed page and sign in.
2. Once signed in, click sign out.
3. You should see an alert that says `Signing Out`.

image

## Checklist
None


packages/peregrine/package.json
```diff
@@ -47,5 +47,8 @@
   "module": "lib/index.js",
   "jsnext:main": "lib/index.js",
   "es2015": "lib/index.js",
-  "sideEffects": false
+  "sideEffects": false,
+  "dependencies": {
+    "tapable": "~1.1.3"
```
sirugh: Is it necessary for us to import a dependency for what is effectively just an event/callback model? If so, do we want to use a dependency that seems to be written for server-side operations? We would need to ensure that the code as written would work in our supported browsers.

I'm also not sure how much this adds to the bundle but it would be worth looking into.

packages/venia-concept/src/signoutExtension.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * This is a mock extension. Assume this file will be provided by
+ * an extension NPM package. For now for the simplicity of the POC
+ * I have added this file as part of venia-concept.
+ */
+
+export default tapableHooks => {
+    tapableHooks.syncOnSignOut.tap('syncOnSignOut', () => {
```
sirugh: So this here is external/vendor code that a client would import/install. We will need to publish the supported callback names as well as their [supported methods](https://github.com/webpack/tapable#hookhookmap-interface).

zetlen:
I agree with @sirugh that we should audit the behavior of Tapable at runtime. I originally liked and approved the idea of using Tapable directly in the UI, since it would give us exactly the same API for both build-time and runtime extensions, maybe allowing the same code to run "isomorphically" in both situations. But I'm not sure that's where we should start.

For now, I think we should prefer build-time extensibility in all cases, so I'd rather that the hook tapping and calling take place in Webpack than in the React app.

packages/peregrine/package.json
revanth0212: Your points are valid. It might have been written keeping the server in mind. But this is far from being a simple event/callback model. This supports many more things that I have not demonstrated in this PR. Things like injecting, hijacking, piggybacking, bailing function internals in both sync and async modes. To do something like this on our own would take lot of time and resources.

Regarding browser compatibility, I don't have a lot of information to share on this topic.

revanth0212: The intended use case of this package is to allow 3rd party extensions to have ability to control function implementations and return values in Venia UI and Peregrine.

packages/venia-concept/src/signoutExtension.js
revanth0212: Yes. It is assumed that the extension provider will have a doc or a page to let the user know about these hooks.

packages/peregrine/package.json
sirugh: >This supports many more things that I have not demonstrated in this PR. Things like injecting, hijacking, piggybacking, bailing function internals in both sync and async modes.

I would love to see examples of this.

sirugh:
This spike has been done for a while, results tracked in JIRA.

PR #2055, opened 2020-01-06
[PWA-261] [SAFARI][DESKTOP][CLOUD] Default swatch image ratio is not proper on product details page 

tjwiebell:
## Description

URL - /valeria-two-layer-tank.html

/valeria-two-layer-tank.html

-Steps –

Hit any of the above cloud deployed URL in SAFARI browser on desktop.
Observe the first default swatch image.
Corrected Steps (if you don't see behavior, be sure to clear all caches)

Navigate to Category page to preload thumbnail images (/venia-tops.html)
Then navigate to a PDP (/vitalia-top.html)
Observe aspect ratio of thumbnail and main image (appears to be serving from service worker cache, then applying cropping again)
Expected - All swatch images should have right ratio.

Actual - First swatch image ration is not proper.

Note - 
1. Works fine on 4.0.0 Magento Cloud https://venia.magento.com/

2. Works fine on 5.0.0 AWS deployed instances (https://pr-2043.pwa-venia.com/valeria-two-layer-tank.html). Issue only with Magento Cloud deployments.

## Related Issue

* [[PWA-261](https://jira.corp.magento.com/browse/PWA-261)] [SAFARI][DESKTOP][CLOUD] Default swatch image ratio is not proper on product details page

### Verification Steps

1. Start Venia with Service Workers enabled, a Fastly backend, and IMAGE_OPTIMIZING_ORIGIN=backend
2. Follow steps in description for repro and verify issue no longer exists
3. Navigate around other pages with images and verify no regressions introduced with image aspect ratio

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/Image/useResourceImage.js
+++ b/packages/peregrine/lib/talons/Image/useResourceImage.js
@@ -24,13 +24,9 @@ export const useResourceImage = props => {
         widths
     } = props;
 
-    const src = useMemo(() => generateUrl(resource, type)(width, height), [
-        generateUrl,
-        height,
-        resource,
-        type,
-        width
-    ]);
+    const src = useMemo(() => {
+        return generateUrl(resource, type)(width, height);
+    }, [generateUrl, height, resource, type, width]);
 
     const srcSet = useMemo(() => {
         return generateSrcset(resource, type);
--- a/packages/peregrine/lib/talons/Image/useResourceImage.js
+++ b/packages/peregrine/lib/talons/Image/useResourceImage.js
@@ -28,7 +28,8 @@ export const useResourceImage = props => {
         return resourceUrl(resource, {
             type,
             height: height,
-            width: width
+            width: width,
+            fit: 'cover'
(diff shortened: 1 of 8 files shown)


packages/peregrine/lib/talons/Image/useResourceImage.js
```diff
@@ -28,7 +28,8 @@ export const useResourceImage = props => {
         return resourceUrl(resource, {
             type,
             height: height,
-            width: width
+            width: width,
+            fit: 'cover'
```
sirugh: This function looks very similar to [this one](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/util/images.js#L22-L28) which is used by `generateSrcset` below. We should use the utility if possible otherwise risk future changes getting out of sync.

tjwiebell: Exported this utility and replaced usage of `resourceUrl` in 39da51e. Good catch 👍

PR #2058, opened 2020-01-06
Exclude admin path from Service Worker

supernova-at:
## Description

It may be the case that a Magento store is hosted on `https://www.example.com` and its Admin UI is at `https://www.example.com/admin`.

The service worker registered by the PWA storefront living at `https://www.example.com` has a default scope of all things underneath it, which in this case includes `/admin/`.

Basically, the PWA service worker does not know that `/admin/` should not be treated as part of the PWA.

### Possible Solutions:

1️⃣ Use [registerNavigationRoute](https://developers.google.com/web/tools/workbox/reference-docs/v4/workbox.routing#.registerNavigationRoute) to redirect all requests to `index.html` except for those in `/admin/`.

```js
    workbox.routing.registerNavigationRoute('/index.html', {
        blacklist: [/^\/admin/]
    });
```

2️⃣ Use [registerRoute](https://developers.google.com/web/tools/workbox/reference-docs/v4/workbox.routing#.registerRoute) and target the admin route with a `networkOnly` matcher.

```js
    workbox.routing.registerRoute(
        new RegExp('\/admin'),
        new workbox.strategies.NetworkOnly()
    );
```

## Related Issue

Closes JIRA-263.

### Verification Stakeholders

@revanth0212 

### Verification Steps

1. Open the Admin UI and PWA storefront in separate tabs in the same browser
1. Go to PWA storefront tab
1. Open dev tools
1. In the Admin UI tab, navigate to Catalog > Categories
1. Hit the "Save" button
1. There should be no error displayed in the dev tools console

## Checklist

### DIFF
--- a/packages/venia-concept/src/ServiceWorker/registerRoutes.js
+++ b/packages/venia-concept/src/ServiceWorker/registerRoutes.js
@@ -16,6 +16,14 @@ import { cacheHTMLPlugin } from './Utilities/htmlHandler';
 export default function() {
     const catalogCacheHandler = createCatalogCacheHandler();
 
+    /**
+     * Exclude anything under the admin path from being served by the service worker
+     * by caching and mapping all routes onto index.html except for URLs starting with /admin.
+     */
+    workbox.routing.registerNavigationRoute('/index.html', {
+        blacklist: [/^\/admin/]
+    });
+
     workbox.routing.registerRoute(
         new RegExp('(robots.txt|favicon.ico|manifest.json)'),
         new workbox.strategies.StaleWhileRevalidate()


fooman:
The default behaviour of Magento is to obfuscate the admin url (having it as /admin/ is nearly only the case for development/demo instances). Additionally efforts have historically being made to keep this secret so as to make it harder for adversaries to start bruteforcing admin logins. Which likely means that even if we were able to exclude it, we probably do not want it showing publicly in the service worker.

supernova-at:
Thanks for your note, @fooman.

I was hoping this draft PR could serve as a proof of concept and perhaps we'd get the obfuscated admin path from GraphQL and inject it, but thanks for pointing out that we don't want it showing publicly in the service worker all.

PR #2059, opened 2020-01-06
[JIRA-262] Robust HTML update checking.

revanth0212:
## Description

Robust HTML update checking using Node HTML Parser.

The issue with current HTML checking is, it is a simple byte to byte comparison. Even if there is a change to an unrelated part of the HTML file that does not affect the App, SW will send a message to UI to show the update toast.

With this change, we can define what the comparison checker needs to compare to deduce if an update is needed.

As of now, this checker tracks, `style`, `link`, `script` and `title` tags along with `data-media-backend` and `data-image-optimizing-origin` attributes of the`html` tag.

We can add new validation functions as we go along.

This approach uses `node-html-parser` which adds 28KB to the bundle size before zipping.

## Related Issue
Closes [JIRA-262](https://jira.corp.magento.com/browse/PWA-262)

### Verification Steps
Can only be tested in local.

1. Run `yarn build && yarn stage:venia`
2. Go to the deployed webpage and refresh a couple times to make sure the SW is installed.
3. Go to the build directory, `venia-concept/dist/index.html` and change `title` or `data-media-backend` or any `style`.
4. Refresh the deployed web page. You should see the update toast.
5. Go back to the `index.html` file and change any entity that is not tracked, like `meta` tag.
6. Repeat step 4 and you should not see any update toast.

## Checklist
* Should not cause any issues in development mode. Ie. `yarn watch:venia`.

### DIFF
--- a/packages/venia-concept/package.json
+++ b/packages/venia-concept/package.json
@@ -36,7 +36,9 @@
     "url": "https://github.com/magento/pwa-studio/issues"
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/venia-concept#readme",
-  "dependencies": {},
+  "dependencies": {
+    "node-html-parser": "~1.1.16"
+  },
   "devDependencies": {
     "@babel/core": "~7.3.4",
     "@babel/plugin-proposal-class-properties": "~7.3.4",
--- a/packages/venia-concept/package.json
+++ b/packages/venia-concept/package.json
@@ -36,9 +36,7 @@
     "url": "https://github.com/magento/pwa-studio/issues"
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/venia-concept#readme",
-  "dependencies": {
-    "node-html-parser": "~1.1.16"
-  },
+  "dependencies": {},
   "devDependencies": {
     "@babel/core": "~7.3.4",
     "@babel/plugin-proposal-class-properties": "~7.3.4",
@@ -93,6 +91,7 @@
     "lodash.over": "~4.7.0",
     "memoize-one": "~5.0.0",
     "memory-fs": "~0.4.1",
+    "node-html-parser": "~1.1.16",
     "prettier": "~1.16.4",
     "prop-types": "~15.7.2",
     "react": "~16.9.0",
--- a/packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
+++ b/packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
@@ -437,3 +437,78 @@ test('fetchDidSucceed should not call sendMessageToWindow if data-image-optimizi
 
     expect(sendMessageToWindow).not.toHaveBeenCalled();
 });
+
+test('fetchDidSucceed should call sendMessageToWindow if a resouce has been added', async () => {
+    /**
+     * Mocking Request to return different values first 2 times.
+     * Adds a resource second time to mock a resource addition.
+     */
+    global.Response.prototype.text = jest
+        .fn()
+        .mockReturnValueOnce(
+            Promise.resolve(
+                '<html><link href="www.adobe.com/spectrumCss.css"></link></html>'
+            )
+        )
+        .mockReturnValueOnce(
+            Promise.resolve(
+                '<html><link href="www.adobe.com/spectrumCss_v0.css"></link><link href="www.adobe.com/spectrumCss_v1.css"></link></html>'
+            )
+        )
+        .mockReturnValue(
+            Promise.resolve(
+                '<html><link href="www.adobe.com/spectrumCss.css"></link></html>'
+            )
+        );
+    global.Response.prototype.clone = function() {
+        return new Response();
+    };
+
+    /**
+     * Mocking cache to have a response for 'https://develop.pwa-venia.com/'
+     */
+    matchFn.mockReturnValue(Promise.resolve(new Response()));
+
+    const url = 'https://develop.pwa-venia.com/';
+    const request = new Request(url);
+    const response = new Response();
+
+    await cacheHTMLPlugin.fetchDidSucceed({ request, response });
+
+    expect(sendMessageToWindow).toHaveBeenCalled();
+});
+
+test('fetchDidSucceed should call sendMessageToWindow if a resouce has been removed', async () => {
+    /**
+     * Mocking Request to return different values first 2 times.
+     * Removes a resource second time to mock a resource removal.
+     */
+    global.Response.prototype.text = jest
+        .fn()
+        .mockReturnValueOnce(
+            Promise.resolve(
+                '<html><link href="www.adobe.com/spectrumCss_v0.css"></link><link href="www.adobe.com/spectrumCss_v1.css"></link></html>'
+            )
+        )
+        .mockReturnValue(
+            Promise.resolve(
+                '<html><link href="www.adobe.com/spectrumCss.css"></link></html>'
+            )
+        );
+    global.Response.prototype.clone = function() {
+        return new Response();
+    };
+
+    /**
+     * Mocking cache to have a response for 'https://develop.pwa-venia.com/'
+     */
+    matchFn.mockReturnValue(Promise.resolve(new Response()));
+
+    const url = 'https://develop.pwa-venia.com/';
+    const request = new Request(url);
+    const response = new Response();
+
+    await cacheHTMLPlugin.fetchDidSucceed({ request, response });
+
+    expect(sendMessageToWindow).toHaveBeenCalled();
+});
--- a/packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
+++ b/packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
@@ -162,7 +162,7 @@ test("fetchDidSucceed should call sendMessageToWindow if 1 or more script tag's 
         return Promise.resolve(
             `<html><script src="www.adobe.com/analytics_script_${Math.random()
                 .toString()
-                .slice(2)}></script></html>`
+                .slice(2)}"></script></html>`
         );
     };
     global.Response.prototype.clone = function() {
@@ -189,9 +189,9 @@ test("fetchDidSucceed should not call sendMessageToWindow if script tag's src ha
      */
     global.Response.prototype.text = function() {
         return Promise.resolve(
-            `<html><script src="www.adobe.com/analytics_script_v1></script><script>var x = ${Math.random()
+            `<html><script src="www.adobe.com/analytics_script_v1"></script><script>var x = ${Math.random()
                 .toString()
-                .slice(2)}</script></html>`
+                .slice(2)}"</script></html>`
         );
     };
     global.Response.prototype.clone = function() {
--- a/packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
+++ b/packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
@@ -6,6 +6,8 @@ import { cacheHTMLPlugin } from '../htmlHandler';
 import { sendMessageToWindow } from '../messageHandler';
 
 const matchFn = jest.fn();
+let originalText = null;
+let originalClone = null;
 
 beforeAll(() => {
     global.caches = {
@@ -17,6 +19,22 @@ beforeAll(() => {
         this.headers = headers;
     };
     global.Response = function() {};
+});
+
+beforeEach(() => {
+    /**
+     * Collecting functions on prototype to restore after tests are done.
+     */
+    originalText = global.Response.prototype.text;
+    originalClone = global.Response.prototype.clone;
+});
+
+afterEach(() => {
+    /**
+     * Restoring prototype function definitions
+     */
+    global.Response.prototype.text = originalText;
+    global.Response.prototype.clone = originalClone;
 });
 
 test('cacheHTMLPlugin should have cacheKeyWillBeUsed, requestWillFetch, fetchDidSucceed async functions implemented', () => {
@@ -81,13 +99,7 @@ test('fetchDidSucceed should not call sendMessageToWindow if the cache does not 
     expect(sendMessageToWindow).not.toHaveBeenCalled();
 });
 
-test('fetchDidSucceed should call sendMessageToWindow if the response is different from what is in the cache', async () => {
-    /**
-     * Collecting functions on prototype to restore after tests are done.
-     */
-    const originalText = global.Response.prototype.text;
-    const originalClone = global.Response.prototype.clone;
-
+test('fetchDidSucceed should call sendMessageToWindow if the response has a different title from what is in the cache', async () => {
     /**
      * Mocking Request to return different value when .text is called.
      */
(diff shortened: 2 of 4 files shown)


revanth0212:
I have burnt my hands trying to get the `DOMParser` working in SW that all browsers provide out of the box. It is a restriction for all web workers. There is a push from the community to get the `DOMParser` API available for Workers. Till then we have to settle for a 3rd party parser. I am using `node-html-parser` which is one of the many parsers with full HTML parsing capabilities for only `28KB`. It is fast too.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -1,7 +1,144 @@
...
+ * @returns Boolean
+ */
+const haveResourcesChanged = (newDOM, oldDOM) => {
+    const newResources = generateResources(newDOM);
+    const oldResources = generateResources(oldDOM);
+    const keysToCompare = Object.keys(newResources);
+
+    return !keysToCompare.reduce((acc, key) => {
+        const hasResourceChanged =
+            newResources[key].toString() === oldResources[key].toString();
```
sirugh: What happens if one of the old resources is dropped entirely? This should be rewritten to check for removal of old keys, addition of new keys, and equivalence of unchanged values.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -115,26 +127,128 @@ test('fetchDidSucceed should call sendMessageToWindow if the response is differe
...
+    const request = new Request(url);
+    const response = new Response();
+
+    await cacheHTMLPlugin.fetchDidSucceed({ request, response });
+
+    expect(sendMessageToWindow).not.toHaveBeenCalled();
 });
 
-test('fetchDidSucceed should not call sendMessageToWindow if the response if same as that in the cache', async () => {
+test("fetchDidSucceed should call sendMessageToWindow if 1 or more script tag's src has changed", async () => {
```
supernova-at:
```suggestion
test("fetchDidSucceed should call sendMessageToWindow if one or more script tags' src has changed", async () => {
```

Super minor but the convention is to spell out numbers less than 10.

I'm really not trying to be a pain I promise 😆

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -115,26 +127,128 @@ test('fetchDidSucceed should call sendMessageToWindow if the response is differe
...
+    const url = 'https://develop.pwa-venia.com/';
+    const request = new Request(url);
+    const response = new Response();
+
+    await cacheHTMLPlugin.fetchDidSucceed({ request, response });
+
+    expect(sendMessageToWindow).toHaveBeenCalled();
+});
+
+test("fetchDidSucceed should not call sendMessageToWindow if script tag's src has not changed even though the wrapped JS has changed", async () => {
```
supernova-at: Good test 👍

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -115,26 +127,128 @@ test('fetchDidSucceed should call sendMessageToWindow if the response is differe
...
+    /**
+     * Mocking Request to return same value when .text is called.
```
supernova-at:
```suggestion
     * Mocking Request to return same value for the script tag when .text is called.
```

supernova-at: Just adding a little more clarity here because of the `Math.random`.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -115,26 +127,128 @@ test('fetchDidSucceed should call sendMessageToWindow if the response is differe
...
+    const url = 'https://develop.pwa-venia.com/';
+    const request = new Request(url);
+    const response = new Response();
+
+    await cacheHTMLPlugin.fetchDidSucceed({ request, response });
+
+    expect(sendMessageToWindow).not.toHaveBeenCalled();
+});
+
+test("fetchDidSucceed should call sendMessageToWindow if 1 or more link's href has changed", async () => {
```
supernova-at:
```suggestion
test("fetchDidSucceed should call sendMessageToWindow if one or more links' href has changed", async () => {
```

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -152,10 +266,174 @@ test('fetchDidSucceed should not call sendMessageToWindow if the response if sam
     await cacheHTMLPlugin.fetchDidSucceed({ request, response });
 
     expect(sendMessageToWindow).not.toHaveBeenCalled();
+});
 
+test('fetchDidSucceed should call sendMessageToWindow if 1 or more styles have changed', async () => {
```
supernova-at:
```suggestion
test('fetchDidSucceed should call sendMessageToWindow if one or more styles have changed', async () => {
```

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -152,10 +266,174 @@ test('fetchDidSucceed should not call sendMessageToWindow if the response if sam
...
+    const url = 'https://develop.pwa-venia.com/';
+    const request = new Request(url);
+    const response = new Response();
+
+    await cacheHTMLPlugin.fetchDidSucceed({ request, response });
+
+    expect(sendMessageToWindow).toHaveBeenCalled();
+});
+
+test('fetchDidSucceed should not call sendMessageToWindow if data-media-backend attribute in the html tag not changed', async () => {
```
supernova-at:
```suggestion
test('fetchDidSucceed should not call sendMessageToWindow if data-media-backend attribute in the html tag has not changed', async () => {
```

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -152,10 +266,174 @@ test('fetchDidSucceed should not call sendMessageToWindow if the response if sam
...
+    const url = 'https://develop.pwa-venia.com/';
+    const request = new Request(url);
+    const response = new Response();
+
+    await cacheHTMLPlugin.fetchDidSucceed({ request, response });
+
+    expect(sendMessageToWindow).toHaveBeenCalled();
+});
+
+test('fetchDidSucceed should not call sendMessageToWindow if data-image-optimizing-origin attribute in the html tag not changed', async () => {
```
supernova-at:
```suggestion
test('fetchDidSucceed should not call sendMessageToWindow if data-image-optimizing-origin attribute in the html tag has not changed', async () => {
```

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -1,7 +1,144 @@
...
+ *
+ * @returns { scripts: [string], links: [string], styles: [string] }
+ */
+const generateResources = parsedHTML => {
+    const resources = {};
+
+    resources.scripts = parsedHTML
+        .querySelectorAll('script')
+        .map(generateScriptResource)
+        .filter(identity)
```
supernova-at: What is this filtering out? What does the `identity` function do?

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -1,7 +1,144 @@
...
+ * @param {HTMLElement} oldDOM
+ *
+ * @returns Boolean
+ */
+const haveResourcesChanged = (newDOM, oldDOM) => {
+    const newResources = generateResources(newDOM);
+    const oldResources = generateResources(oldDOM);
+    const keysToCompare = Object.keys(newResources);
+
+    return !keysToCompare.reduce((acc, key) => {
```
supernova-at: I think [every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) fits better than `reduce` here.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: If an old resource is removed the condition will fail and `hasResourceChanged` will be set to true and the page will be asked to refresh. Isn't that how it is supposed to be?

sirugh: You're going to run into errors at runtime right?

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -1,7 +1,144 @@
...
+ *
+ * @returns Boolean
+ */
+const hasBackendUrlChanged = (newDOM, oldDOM) => {
+    const oldUrl = oldDOM.querySelector('html').attributes[
+        'data-media-backend'
+    ];
+    const newUrl = newDOM.querySelector('html').attributes[
+        'data-media-backend'
+    ];
```
supernova-at: There are selectors for data attributes. See here: https://css-tricks.com/almanac/selectors/a/attribute/.

This could be `querySelector('[data-media-backend]')` if you want.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -1,7 +1,144 @@
...
+
+/**
+ * Runs all verification functions with new and old
+ * HTML to determine if HTML has changed.
+ *
+ * @param {HTMLElement} newDOM
+ * @param {HTMLElement} oldDOM
+ */
+const hasHTMLChanged = (newDOM, oldDOM) =>
+    verificationFunctions.reduce((acc, fn) => {
```
supernova-at: Here again I think you can use the `every` or `some` array functions that fit better than `reduce`.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -35,15 +172,21 @@ export const cacheHTMLPlugin = {
...
         if (cachedResponseObj) {
             const cachedResponse = await cachedResponseObj.text();
             const clonedResponse = await response.clone().text();
-            if (cachedResponse !== clonedResponse) {
+
+            if (
+                hasHTMLChanged(
+                    parse(clonedResponse, { style: true }),
+                    parse(cachedResponse, { style: true })
+                )
```
supernova-at: We could save ourselves some time by keeping both conditionals here.

Only if `cachedResponse !== clonedResponse` do we check `hasHTMLChanged`. This way if they are the same, we don't do any of the more time-intensive checks.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: > ~~If an old resource is removed, it won't cause a refresh.~~ If a new resource is added, then it will cause the page to refresh by setting `hasResourceChanged`. ~~Thought i realized there will be a `TypeError` in such cases. Will be fixing it in the next commit.~~

I would like to take a couple things back.

1. If a resource/resources are removed or added, page will refresh. This is already being done. `newResources` and `oldResources` are not a list, they are of type `{scripts: [], links: [], styles: []}`. If a resource/resources are added or removed, their respective array changes and a refresh will be triggered. 

2. There won't be any `TypeErrors`. The code works as-is. Though I will be adding a couple more tests to replicate this behavior.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
revanth0212: I believe you Andy 😆

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: It removes falsy values in this case `null` or `undefined`.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: Nice catch Andy. This saves time. We can only calculate the differences if the HTMLs differ.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: Added tests to replicate the behavior.

packages/venia-concept/src/ServiceWorker/Utilities/__tests__/htmlHandler.spec.js
```diff
@@ -19,6 +21,22 @@ beforeAll(() => {
...
+    originalClone = global.Response.prototype.clone;
+});
+
+afterEach(() => {
+    /**
+     * Restoring prototype function definitions
+     */
+    global.Response.prototype.text = originalText;
+    global.Response.prototype.clone = originalClone;
+});
```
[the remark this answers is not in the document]
revanth0212: We need these changes to be done before and after each test. That is the reason why I am using beforeEach and afterEach.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: We can not use such selectors because the type of the dom object, in this case, is actually not what `DOMParser` returns, this object is different and is returned by the `node-html-parser` which does not support complex selectors.

revanth0212: We can not use such selectors because the type of the dom object, in this case, is actually not what DOMParser returns, this object is different and is returned by the node-html-parser which does not support complex selectors.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: Nice catch. Done 👍

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
revanth0212: Nice catch. Done 👍

packages/venia-concept/package.json
```diff
@@ -36,7 +36,9 @@
     "url": "https://github.com/magento/pwa-studio/issues"
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/venia-concept#readme",
-  "dependencies": {},
+  "dependencies": {
+    "node-html-parser": "~1.1.16"
+  },
```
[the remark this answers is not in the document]
revanth0212: Done.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
supernova-at: Can we give it a better name or add a comment or something? It's not really clear right now.

packages/venia-concept/src/ServiceWorker/Utilities/htmlHandler.js
```diff
@@ -181,7 +176,12 @@ export const cacheHTMLPlugin = {
             const cachedResponse = await cachedResponseObj.text();
             const clonedResponse = await response.clone().text();
 
+            /**
+             * Dont bother caculating changes if the response string
+             * has not changed. Saves CPU cycles most of the time.
+             */
             if (
+                clonedResponse !== cachedResponse &&
```
supernova-at: 💯

dpatil-magento:
Merged as part of #2086

PR #2084, opened 2020-01-07
Venia v2 Cart Page

supernova-at:
## Description

Venia v2 separates out the Cart into its own page. This PR lays out the basic structure of the page. Future PRs will populate the page with its actual content / components.

## Related Issue

Closes [PWA-130](https://jira.corp.magento.com/browse/PWA-130).

### Verification Stakeholders

@jimbo 

### Verification Steps

1. `yarn watch:venia`
1. Navigate to Venia in your browser
1. Navigate to `<your dev site>/cart`
1. Verify that the new Cart page skeleton appears
1. Verify that if you are not signed in, a "Sign In" button appears
1. Verify that clicking the "Sign In" button opens the navigation drawer.
1. Verify that if you are already signed in, the "Sign In" button does not appear
1. Switch between Desktop and Mobile (< `640px`) views

On Mobile:
1. Verify that the "Recently Viewed" container is not present
1. Verify that the page is a single column layout
1. Verify that the "Order Summary" container appears at the bottom

On Desktop:
1. Verify that the "Recently Viewed" container is present
1. Verify that the page is split into two columns
1. Scroll and verify that the "Order Summary" container "sticks" to the side / stays in view. Note that it is expected that it will disappear once you scroll past all of the content.

**Mobile**

Screen Shot 2020-01-07 at 2 44 03 PM

**Desktop**

Screen Shot 2020-01-07 at 2 44 22 PM

## Checklist


packages/venia-ui/lib/components/CartPage/cartPage.js
```diff
@@ -0,0 +1,60 @@
...
+
+import { mergeClasses } from '../../classify';
+import defaultClasses from './cartPage.css';
+
+const CartPage = props => {
+    const { handleSignIn, isSignedIn } = useCartPage();
+
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    const signInDisplay = !isSignedIn ? (
```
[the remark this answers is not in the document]
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/2084/commits/4442862c88e02f4b8c51eb197e10b36e40c57947.

packages/peregrine/lib/talons/CartPage/useCartPage.js
```diff
@@ -0,0 +1,20 @@
...
+import { useAppContext } from '../../context/app';
+import { useUserContext } from '../../context/user';
+
+export const useCartPage = () => {
+    const [, appApi] = useAppContext();
+    const { toggleDrawer } = appApi;
+    const [{ isSignedIn }] = useUserContext();
+
+    const handleSignIn = useCallback(() => {
+        // TODO: set navigation state to "SIGN_IN". useNavigation:showSignIn doesn't work.
```
sirugh: Right - the `showSignIn` callback and menu state are local to the navigation component. We would have to hoist the navigation state to a context similar to app. So we could, like we do with `app.toggleDrawer('nav')`, do something like `navigation.toggleView('signIn')`. I'm OK with this just opening the left drawer for now.

packages/venia-ui/lib/components/CartPage/cartPage.css
```diff
@@ -0,0 +1,76 @@
...
+    height: 100%;
+}
+
+.summary_contents {
+    position: sticky;
+    /*
+     * TODO: Use CSS Properties (variables) or something instead of hardcoding these. 
+     * min-height of nav header + page spacing + height of page header 
+     */
+    top: calc(3.5rem + 3.5rem + 25px);
```
sirugh: This is not sticking as I would expect. Add some extra long text like lorem ipsum and scroll. You'll see that it scrolls off the page.

packages/venia-ui/lib/components/CartPage/cartPage.css
```diff
@@ -0,0 +1,76 @@
...
+    grid-area: adjustments;
+}
+
+.recently_viewed_container {
+    grid-area: recents;
+}
+
+.sign_in {
+    font-weight: bold;
+    text-decoration: underline;
```
sirugh: This text looks a little larger than in the mocks, at least in relation to its counterpart in the header, "Cart".

packages/venia-ui/lib/components/CartPage/cartPage.css
```diff
@@ -0,0 +1,76 @@
...
+
+.items_container {
+    grid-area: items;
+}
+
+.price_adjustments_container {
+    grid-area: adjustments;
+}
+
+.recently_viewed_container {
```
sirugh: Since this is an extension/dynamic location can we just call this `cart_extension_container` and rename the grid area?

supernova-at: Do we know for sure that's the case? I wasn't clear.

sirugh: Ah yea, it was discussed when we broke the stories down that its more like a configurable area for things like "recently viewed" or "people like you purchased..." etc.

sirugh: TBH It may be better to simply remove/comment out the html since those components don't exist and I'm not sure we are expected to even implement it within this epic.

packages/venia-ui/lib/components/CartPage/cartPage.css
supernova-at: Oh yeah, that's what

> Scroll and verify that the "Order Summary" container "sticks" to the side / stays in view. Note that it is expected that it will disappear once you scroll past all of the content.

was supposed to identify 😛.

If we're talking about the same thing, the fact that you can scroll past the content entirely is a product of a [style that gets applied to all pages](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/Main/main.css#L13): the initial height of the page content area is always set to `100vh`.

I think once the components on this page are present we won't be able to repro this.

packages/venia-ui/lib/components/CartPage/cartPage.css
supernova-at: That's actually a good callout - I don't know how to open the mock in XD to get the correct size. I'll dig deeper.

sirugh: Should be able to find a link somewhere in the epic or linked issues. But @tjwiebell is also having trouble finding pixel/size direction from XD so not sure how to help from there.

packages/venia-ui/lib/components/CartPage/cartPage.css
```diff
@@ -50,10 +50,10 @@
 .summary_contents {
     position: sticky;
     /*
-     * TODO: Use CSS Properties (variables) or something instead of hardcoding these. 
-     * min-height of nav header + page spacing + height of page header 
+     * TODO: Use CSS Properties (variables) or something instead of hardcoding this. 
+     * 3.5 rem = min-height of nav header. See the "toolbar" class in header.css.
      */
-    top: calc(3.5rem + 3.5rem + 25px);
+    top: 3.5rem;
```
sirugh: Would have liked this to be a variable but the TODO will suffice

jimbo:
Looking through late on this one, but it's fine. We can patch up the CSS as part of the pricing summary work. 👍

PR #2086, opened 2020-01-08
[5.0] [JIRA-262] Robust HTML update checking.

revanth0212:
## Description

Robust HTML update checking using Node HTML Parser.

The issue with current HTML checking is, it is a simple byte to byte comparison. Even if there is a change to an unrelated part of the HTML file that does not affect the App, SW will send a message to UI to show the update toast.

With this change, we can define what the comparison checker needs to compare to deduce if an update is needed.

As of now, this checker tracks, `style`, `link`, `script` and `title` tags along with `data-media-backend` and `data-image-optimizing-origin` attributes of the`html` tag.

We can add new validation functions as we go along.

This approach uses `node-html-parser` which adds 28KB to the bundle size before zipping.

## Related Issue
Closes [JIRA-262](https://jira.corp.magento.com/browse/PWA-262)

### Verification Steps
Can only be tested in local.

1. Run `yarn build && yarn stage:venia`
2. Go to the deployed webpage and refresh a couple times to make sure the SW is installed.
3. Go to the build directory, `venia-concept/dist/index.html` and change `title` or `data-media-backend` or any `style`.
4. Refresh the deployed web page. You should see the update toast.
5. Go back to the `index.html` file and change any entity that is not tracked, like `meta` tag.
6. Repeat step 4 and you should not see any update toast.

## Checklist
* Should not cause any issues in development mode. Ie. `yarn watch:venia`.

### DIFF
--- a/packages/venia-concept/package.json
+++ b/packages/venia-concept/package.json
@@ -36,7 +36,9 @@
     "url": "https://github.com/magento/pwa-studio/issues"
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/venia-concept#readme",
-  "dependencies": {},
+  "dependencies": {
+    "node-html-parser": "~1.1.16"
+  },
   "devDependencies": {
     "@babel/core": "~7.3.4",
     "@babel/plugin-proposal-class-properties": "~7.3.4",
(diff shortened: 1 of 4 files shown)


dpatil-magento:
@supernova-at Need your approval on this. Its copy of #2059 but base branch in `release/5.0`.

PR #2090, opened 2020-01-09
Adds PriceAdjustments and Accordion components

supernova-at:
## Description

This PR adds a `PriceAdjustments` and `Accordion` components, needed for Venia v2 Cart page.
Since `Accordion` was written for re-use elsewhere throughout the app, it does not live in `CartPage/`.

## Related Issue

Closes PWA-241.

### Verification Stakeholders

@jimbo @sirugh @tjwiebell 

#### Price Adjustments

1. Go the `/cart` page
1. Verify that the price adjustments container appears
1. Verify that the price adjustments container is functional by opening and closing each accordion section and verifying the respective contents are shown only when the section is open
1. Verify that the border / divider / separators are always correct, especially when opening and closing sections (there should be no "double borders")
1. Verify that it works in mobile view
1. Verify that it is keyboard navigable (you can navigate to and open / close sections using keyboard only)

#### Accordion

1. Check out this PR locally and run `yarn workspace @magento/venia-ui run storybook`
1. Navigate to the `Accordion` stories and verify that each use case functions as you'd expect it to

priceAdjustments_accordion

## Checklist

## In Progress
- [x] Stories
- [x] Tests


supernova-at:
I'm working to simplify the logic in the `Accordion` talons at the moment. I think it can be much more straightforward.

packages/peregrine/lib/talons/Accordion/useAccordion.js
```diff
@@ -0,0 +1,82 @@
...
+    cloneElement,
+    useCallback,
+    useEffect,
+    useState
+} from 'react';
+
+export const useAccordion = props => {
+    const { canOpenMultiple, children } = props;
+
+    const [openSectionIndicies, setOpenSectionIndicies] = useState(new Set([]));
```
jimbo: Nice, good idea.

On a minor note, you'll want to find and replace all instances of `indicies` with `indices`. 😉

packages/peregrine/lib/talons/Accordion/useAccordion.js
```diff
@@ -0,0 +1,82 @@
...
+                updatedIndicies.add(sectionIndex);
+            } else {
+                // The user wants to close this section.
+                updatedIndicies.delete(sectionIndex);
+            }
+
+            setOpenSectionIndicies(updatedIndicies);
+        },
+        [canOpenMultiple, openSectionIndicies, setOpenSectionIndicies]
+    );
```
jimbo: You're on the right track here. Unfortunately, this callback will be recreated on every state update, because it depends on `openSectionIndices`, which is the very same state it's updating.

To work around this, use the functional update signature of `setOpenSectionIndices`:

```es6
const handleSectionClick = useCallback(
  index => {
    setOpenSectionIndices(prevIndices => {
      const nextIndices = new Set(prevIndices)

      if (!prevIndices.has(index)) {
        if (!canOpenMultiple) {
          nextIndices.clear()
        }

        nextIndices.add(index)
      } else {
        nextIndices.delete(index)
      }

      return nextIndices
    })
  },
  [canOpenMultiple, setOpenSectionIndices]
)
```

packages/peregrine/lib/talons/Accordion/useAccordion.js
```diff
@@ -0,0 +1,82 @@
...
+    // 2. What index it is.
+    // 3. A callback function to call when it gets clicked.
+    const controlledChildren = Children.map(children, (child, childIndex) => {
+        const isOpen = openSectionIndicies.has(childIndex);
+
+        return cloneElement(child, {
+            handleClick: handleSectionClick,
+            index: childIndex,
+            isOpen
+        });
```
jimbo: This is a clever solution, but `cloneElement` is generally an antipattern these days (barring any unusual circumstances), or at least a smell.

Why not use context to make `openSectionIndices` and a callback or two available to the child sections? That's the approach we've been using, for the most part.

packages/venia-ui/lib/components/Accordion/accordion.css
```diff
@@ -0,0 +1,11 @@
+.root {
+    border: 1px solid lightgray;
```
jimbo: There should be an existing CSS variable for our border color that you can use (`--venia-border`).

packages/venia-ui/lib/components/Accordion/accordion.css
```diff
@@ -0,0 +1,11 @@
...
+    border-radius: 0.5em;
+}
+
+.root > * {
+    border-bottom: solid 1px lightgray;
+}
```
jimbo: This is valid CSS, but we don't use this approach in our components. Prefer putting these styles directly on some classnames in `section.css`.

If you really want to emphasize that `accordion.css` is the source of the styles, create some classnames in this file and have the classnames in `section.css` extend them with `composes`.

packages/peregrine/lib/talons/Accordion/useAccordion.js
supernova-at: you_got_me

packages/peregrine/lib/talons/Accordion/useAccordion.js
supernova-at: Sweet, thanks - done!

packages/venia-ui/lib/components/Accordion/accordion.css
supernova-at: Thanks for catching this, I meant to comment on it.  I was forgetting the `rgb()` in front of `var(---venia-border)` 😨

packages/venia-ui/lib/components/Accordion/accordion.css
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2090/commits/507c44e2d3ff7678f3c5afd7a9a2a1bb4109e650.

packages/venia-ui/lib/components/Accordion/accordion.css
supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/2090/commits/ae7b1ae4cf2bdb04f9da1c9632295ffddcd37ad0.

brendanfalkowski:
@supernova-at — This is scary-similar to our accordion, which used `cloneElement` as well but we noted an issue that maybe you'll get to before we loop back.

Sometimes, for business reasons, a section of our tabs/accordions is not displayed (ex: some products have documents attached, and some do not so we hide that tab/section). When using `cloneElement` if a child is null or empty string or the calling function returns nothing, that fails because the node has no props when the section executes.

There's an old React ticket debating handling this: https://github.com/facebook/react/issues/4867 which lands on filtering out null children first then mapping (i.e. not React's problem).

We spotted it recently but haven't fixed ours yet. Just worth a mention you may want to test for "it depends" sections.

packages/venia-ui/lib/components/Accordion/accordion.js
```diff
@@ -0,0 +1,41 @@
...
+const AccordionContext = createContext(contextValue);
+const { Provider } = AccordionContext;
+
+const Accordion = props => {
+    const { canOpenMultiple = true, children } = props;
+
+    // The talon is the source of truth for the context value.
+    const talonProps = useAccordion({ canOpenMultiple, children });
+    const { handleSectionToggle, lastSectionId, openSections } = talonProps;
+    contextValue = {
```
jimbo:
```suggestion
    const contextValue = { handleSelectionToggle, openSections }
```

---
I see what you've done here, but I'd suggest just creating the `const` here. If you need to set some default values like you're doing above, I think you should do that in the talon.

packages/venia-ui/lib/components/Accordion/accordion.js
```diff
@@ -0,0 +1,41 @@
...
+        handleSectionToggle,
+        lastSectionId,
+        openSections
+    };
+
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    return (
+        <Provider value={contextValue}>
```
jimbo: Perfect. 👍

packages/venia-ui/lib/components/Accordion/section.js
```diff
@@ -0,0 +1,49 @@
...
+
+const Section = props => {
+    const { children, id, title } = props;
+
+    const {
+        handleSectionToggle,
+        lastSectionId,
+        openSections
+    } = useAccordionContext();
+    const isLast = lastSectionId === id;
```
jimbo: I apologize for the confusion caused by my last comment. It's fine to use pseudo-selectors such as `:last-child` or `:last-of-type`. I simply wanted to avoid using descendant selectors such as `.root > div` or `:root > *`. So I don't think we'll need `id` or `lastSectionId` logic. 😞

packages/venia-ui/lib/components/Accordion/accordion.js
```diff
@@ -0,0 +1,41 @@
...
+            <div className={classes.root}>{children}</div>
+        </Provider>
+    );
+};
+
+export default Accordion;
+
+export const useAccordionContext = () => useContext(AccordionContext);
```
jimbo: Also perfect. 💯

packages/venia-ui/lib/components/Accordion/section.js
```diff
@@ -0,0 +1,49 @@
...
+import defaultClasses from './section.css';
+
+const Section = props => {
+    const { children, id, title } = props;
+
+    const {
+        handleSectionToggle,
+        lastSectionId,
+        openSections
+    } = useAccordionContext();
```
jimbo: This is exactly what I had in mind. 👍

packages/venia-ui/lib/components/Accordion/section.js
supernova-at: No worries! 

`id` is still used to determine which `Section` is open, but I was able to nuke `lastSectionId` 👍

packages/venia-ui/lib/components/Accordion/accordion.js
supernova-at: I may or may not have been inspired by `Routes` 😉

packages/venia-ui/lib/components/Accordion/accordion.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/2090/commits/83fecaaaaaaf3326967ed88ee1c6fc641bf1aef1, good call.

packages/venia-ui/lib/components/Accordion/section.js
supernova-at: I made that more clear in https://github.com/magento/pwa-studio/pull/2090/commits/82e01f193d85081e3dcde5dd870bb4284b6d43fa.

jimbo:
> This is scary-similar to our accordion, which used `cloneElement` as well but we noted an issue that maybe you'll get to before we loop back.
> 
> *snip*

@brendanfalkowski Thanks, that's interesting to know about. Fortunately, I think we're able to use context here and avoid using `cloneElement`, so it shouldn't be an issue.

dpatil-magento:
@supernova-at  Whats expected behavior wrt to expand/collapse - 
1. When I hit `/cart` ?   By default Shipping method is in expanded state and rest all are collapsed.
2. If I Expand one, two or all and re-load page, state should saved or go back to default?

dpatil-magento:
@soumya-ashok @schensley  Need your call #2.  Andy clarified that #1 is expected.

PR #2092, opened 2020-01-09
[PWA-240] [feature]: Cart Price Summary

sirugh:
## Description

Adds a price summary to the cart page. Each of the more "complex" line items have been broken into separate components. To avoid having "magical" named queries being used in the parent I added a dependency called [fraql](https://github.com/smooth-code/fraql) that allows you to create unnamed queries. Look at `priceSummary.js` for usage example.

[Image from Gyazo](https://gyazo.com/eba73decceb705eeff00e32ac7a75a67)

## Related Issue

Closes [PWA-240](https://jira.corp.magento.com/browse/PWA-240).

### Verification Steps

For reference, use [the tutorial](https://devdocs.magento.com/guides/v2.3/graphql/tutorials/checkout/checkout-customer.html) for checkout w/ graphQL.

### Empty cart

Go to the `/cart` page. The summary should not appear when there are no items.

### With items

1) Add an item to your cart and obtain the cart id.
2) Using graphql playground/postman/etc, apply a coupon (you  may need to create a coupon for your storefront)
3) Using graphql playground/postman/etc, set shipping address.
4) Using graphql playground/postman/etc, set shipping method (using "flatrate"/"flatrate" for carrier/method)
5) Refresh the `/cart` page. You should see an updated summary.

[Image from Gyazo](https://gyazo.com/f835a0d4b646e624ad1fb71b97206a30)

### Scrolling view

The summary should scroll with the contents of the cart page.

### Note

* Coupons values are represented in an array with other possible values and no way to identify it as such. For now, I have represented this as a "Discount" line item.

* Gift Cards can be enabled/disabled at build time if we're on EE or not, but that work needs to be done. In this PR they are disabled entirely with a local flag but we should address that soon. Deferred the work to PWA-78 for now.

## Checklist


packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.js
```diff
@@ -0,0 +1,76 @@
...
+            <span className={classes.price}>
+                {'(-'}
+                <Price value={cards.value} currencyCode={cards.currency} />
+                {')'}
+            </span>
+        </>
+    ) : null;
+};
+
+// TODO: Gift cards are only enabled in EE, write a build time tool that turns the component into a no-op and the static fragments into __typename requests.
```
sirugh: This is a _big_ TODO. Before we merge this PR we have to have a way to determine which fragment to use, otherwise we risk breaking CE.

packages/venia-ui/lib/components/CartPage/PriceSummary/discountSummary.js
```diff
@@ -0,0 +1,69 @@
...
+                {')'}
+            </span>
+        </>
+    ) : null;
+};
+
+DiscountSummary.fragments = {
+    discounts: gql`
+        fragment _ on CartPrices {
+            discounts {
```
sirugh: In 2.3.3 we have `discount`, which is a deprecated field in 2.3.4
In 2.3.4 we have `discounts`, which 2.3.3 does not have.

If we want to support both versions we will need to provide some way (build or runtime?) to determine which field is valid for the current backend.

sirugh: I am going to proceed with the assumption that we _don't_ need to support 2.3.3, so `discounts` is valid.

packages/peregrine/lib/talons/CartPage/PriceSummary/usePriceSummary.js
```diff
@@ -0,0 +1,32 @@
...
+import { useQuery } from '@apollo/react-hooks';
+import { useCartContext } from '@magento/peregrine/lib/context/cart';
+
+export const usePriceSummary = props => {
+    const [{ cartId }] = useCartContext();
+    const { error, loading, data } = useQuery(props.query, {
+        variables: {
+            cartId
+        },
+        fetchPolicy: 'no-cache'
```
sirugh: We should verify that actions that modify the price summary values such as shipping method or adding a shipping address elsewhere also update the price summary even if `no-cache` is used here.

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummary.js
```diff
@@ -0,0 +1,136 @@
...
+
+import DiscountSummary from './discountSummary';
+import GiftCardSummary from './giftCardSummary';
+import ShippingSummary from './shippingSummary';
+import TaxSummary from './taxSummary';
+
+/**
+ * A component that fetches and renders cart data including:
+ *  - subtotal
+ *  - coupons applied
```
sirugh: s/coupons/discounts

packages/venia-ui/lib/components/CartPage/PriceSummary/shippingSummary.js
```diff
@@ -0,0 +1,55 @@
...
+const ShippingSummary = props => {
+    const { classes, data } = props;
+
+    // Don't render estimated shipping until an address has been provided and
+    // a method has been selected.
+    if (!data.length || !data[0].selected_shipping_method) {
+        return null;
+    }
+
+    const shipping = shipping_addresses[0].selected_shipping_method.amount;
```
revanth0212: Where is `shipping_addresses` defined?

packages/venia-ui/lib/components/CartPage/PriceSummary/taxSummary.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * A component that renders the discount summary line item.
+ *
+ * @param {Object} props.classes
+ * @param {Object} props.data query response data
+ */
+const TaxSummary = props => {
+    const { classes } = props;
+
+    const tax = getEstimatedTax(props.data);
```
revanth0212: Should this not be `const tax = getEstimatedTax(props.data.applied_taxes);`?

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummary.js
```diff
@@ -0,0 +1,136 @@
...
+ *  - gift cards applied
+ *  - estimated tax
+ *  - estimated shipping
+ *  - estimated total
+ */
+const PriceSummary = props => {
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const talonProps = usePriceSummary({
+        query: PriceSummary.GET_PRICE_SUMMARY
```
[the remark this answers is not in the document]
sirugh: Yes! Good catch.

packages/venia-ui/lib/components/CartPage/PriceSummary/shippingSummary.js
sirugh: Oops! Fixed.

packages/venia-ui/lib/components/CartPage/PriceSummary/taxSummary.js
sirugh: `data` is already the destructured value from the query response. I didn't do the destructuring within the line item components because they don't really know how to obtain their data from a larger object -- they only expect to receive the data in the fragment that they have asked for.

sirugh:
> Looks good to me. I'd like to know what effect fraql has on the bundle before approving this; I'm fine with it from a fragment authoring perspective.

I have removed it and used named exports instead. This just leaves a little coupling magic in terms of fragment name and the name of the export but that's not a big deal, and is preferable over using a new dependency IMO.

> Also note that we do have to support 2.3.3.

Rather than go back and forth on this, let's put that in documentation somewhere and make sure the entire team knows exactly what backend version we have to support when we are coding. I spoke with @tjwiebell and we both thought that this code wouldn't be released until our next release which would coincide with 2.3.5. And we both thought that we were supporting "current - 1", so 6.0.0 would have to support 2.3.5 and 2.3.4 whereas 5.0.0 which is releasing now will have to support 2.3.4 and 2.3.3

In regards to this specific PR, I can work around that by using the `discount` field which is deprecated in 2.3.4 but still exists. I just worry that we're going to run into complete removal of fields that we have to support since GQL doesn't really do semver "the right way".

jimbo:
Yes, my mistake. This can be for 2.3.4, so we're good here. 👍

dpatil-magento:
QA Pass, can be merged once re-approved.

PR #2088, opened 2020-01-08
Add a button section to the styleguide

jimbo:
## Description

Add a button section to the styleguide. Some of the prose content may be lorem ipsum, but the values should reflect the official Venia styles.

## Related Issue
PWA-154

## Acceptance 
- @jcalcaben should affirm document structure
- @soumya-ashok should affirm Venia typographic styles
- Others may review the code

### Verification Stakeholders
- @jcalcaben
- @soumya-ashok 

### Verification Steps
1. `yarn workspace @magento/venia-styleguide run start`
2. Navigate to `/page/button`

## Checklist
* I have updated the documentation accordingly, if necessary.
* I have *not* added tests to cover my changes, if necessary.

PR #2030, opened 2019-12-12
Add color section to styleguide

jimbo:
## Description
Add a color section to the styleguide. The prose content may be lorem ipsum, but the values should reflect the official Venia styles.

## Related Issue
PWA-224

## Acceptance 
- @jcalcaben should affirm document structure
- @soumya-ashok should affirm Venia typographic styles
- Others may review the code

### Verification Stakeholders
- @jcalcaben 
- @soumya-ashok 

### Verification Steps

1. `yarn workspace @magento/venia-styleguide run start`
2. Navigate to `/page/color`

## Checklist
* I have updated the documentation accordingly, if necessary.
* I have *not* added tests to cover my changes, if necessary.

PR #2093, opened 2020-01-10
[PWA-156] Added Drawer Footer Styleguide.

revanth0212:
## Description

Drawer Footer Styleguide.

## Related Issue
Closes [PWA-156](https://jira.corp.magento.com/browse/PWA-156)

### Verification Stakeholders
@jimbo @soumya-ashok @jcalcaben 

### Verification Steps
2. Navigate to `/page/drawer-footer`

### Screenshots
image

#### Checklist
1. To be merged only after #2088 is merged.


packages/peregrine/lib/PeregrineContextProvider/__tests__/__snapshots__/peregrineContextProvider.spec.js.snap
```diff
@@ -1,17 +1,17 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
supernova-at: This file doesn't seem related to the changes.

packages/venia-styleguide/src/pages/drawer-footer/sections/Behaviors/index.mdx
```diff
@@ -0,0 +1,38 @@
...
+import FormButtonsLayout from './FormButtonsLayout.mdx';
+
+import DrawerFooter from '../../../../components/DrawerFooter/DrawerFooter';
+import Button from '../../../../components/Button';
+
+<Columns reverse>
+    <SingleButtonLayout />
+    <ExampleGroup>
+        <DrawerFooter>
+            <Button priority="high">Signin</Button>
```
supernova-at:
```suggestion
            <Button priority="high">Sign In</Button>
```

jimbo:
@revanth0212 I've updated the base branch from `develop` to `jimbo/styleguide-buttons`.

packages/venia-styleguide/src/components/DrawerFooter/DrawerFooter.css
```diff
@@ -0,0 +1,9 @@
+.root {
+    align-items: center;
+    background-color: rgb(var(--venia-global-color-gray-75));
```
jimbo:
```suggestion
    background-color: rgb(var(--venia-global-color-gray-50));
```
---
I believe that in all cases, the drawer footer has a white background.

packages/venia-styleguide/webpack.config.js
```diff
@@ -37,8 +37,9 @@ module.exports = () => ({
                     }
                 ]
             },
+
             {
-                test: /\.md$/,
+                test: /\.md$|\.mdx$/,
```
jimbo: This was intentional; just as we avoid the `.jsx` extension, I'd like to avoid the `.mdx` extension here in favor of just `.md`.

packages/venia-styleguide/src/components/DrawerFooter/DrawerFooter.css
revanth0212: I remember picking that from another css file from the style guide. Will make the change.

packages/venia-styleguide/webpack.config.js
revanth0212: Does this mean, no more `mdx` files? I have a bunch. If so, I can update them all to `md`.

revanth0212: Done

jimbo: Correct. `.md` is a valid extension for MDX, and it's the one I chose to favor here.

jimbo:
Approved, but blocked by preceding PRs:

- #2030 
- #2088

soumya-ashok:
@revanth0212 I've just had a look at the implementation.

- I see a white background behind the buttons. Is that intentional?
- Also, I assume that the way they are placed is meant to indicate that they should be vertically and horizontally aligned to or around the center, depending on how many buttons there are but this is not obvious.
- In the first two button example, we have 2 secondary buttons. This shouldn't typically happen. The more regular case would be a secondary and primary like in the third example.
- Are we addressing break-points and relative spacing or is this for the smallest mobile view?

revanth0212:
> I see a white background behind the buttons. Is that intentional?

As far as I remember, this is intentional. I can revert it if that's not expected.

> Also, I assume that the way they are placed is meant to indicate that they should be vertically and horizontally aligned to or around the center, depending on how many buttons there are but this is not obvious.

If we need a vertical example, I can add it to the styleguide. @jimbo what do you think?

> In the first two button example, we have 2 secondary buttons. This shouldn't typically happen. The more regular case would be a secondary and primary like in the third example.

Can it not be a case where 2 buttons will definitely not be primary? As an example, next and back both can be secondary.

> Are we addressing break-points and relative spacing or is this for the smallest mobile view?

Not sure of this.

jimbo:
> I see a white background behind the buttons. Is that intentional?

Yes. Drawers (and footers) have a white background.

> Also, I assume that the way they are placed is meant to indicate that they should be vertically and horizontally aligned to or around the center, depending on how many buttons there are but this is not obvious.

Yes, there are a maximum of two buttons, vertically and horizontally centered in a single row. There is not an example of vertically stacked buttons in a footer, because a footer has a single row.

>  In the first two button example, we have 2 secondary buttons. This shouldn't typically happen. The more regular case would be a secondary and primary like in the third example.

@soumya-ashok Agreed, examples should show one high and one low priority button, since that's the typical case.

@revanth0212 You're correct, it could potentially be two low priority buttons, but that would be unusual.
 
> Are we addressing break-points and relative spacing or is this for the smallest mobile view?

In this case, there are no breakpoints, since a drawer's width (and thus the footer's width) matches our smallest breakpoint.

soumya-ashok:
@revanth0212 and @jimbo Thanks for clarifying.

We don't need vertically stacked buttons, I meant center-aligned along the height.
There could be a case when someone chooses to use 2 secondary buttons alongside one another but I don't want to advocate it. Even in the case of "next" and "back", "next" is presumably the preferred action, and would be indicated with a primary button.

revanth0212:
> There could be a case when someone chooses to use 2 secondary buttons alongside one another but I don't want to advocate it. Even in the case of "next" and "back", "next" is presumably the preferred action, and would be indicated with a primary button.

You are right here. The Drawer Footer is just 1 component in a page and sometimes it may not have a primary action but the page housing the drawer footer might have. In such situations, having 2 secondary buttons it is a valid use case.

revanth0212:
We can merge this PR for now and circle back later to discuss the missing use cases once we have a matured style guide.

PR #2094, opened 2020-01-10
[Cart v2] ProductListing Component

tjwiebell:
## Description

Within the Cart Page component, if a product has been added to cart it will display with necessary information and actions for a shopper to view/edit/delete. 

 

Requirements: 

Adhere to design specs/flows provided
Empty cart should display appropriate messaging
"Sign in" button opens sign in modal that returns back to cart once user is auth'ed
User can update quantity using increment/decrement functionality
Kebab menu should reuse options from existing Venia mini-cart
Open Questions: 

What product attributes should be listed in the view? For now, it's just size/color in the mockup. As discussed in grooming, all product options should be displayed for now. 

## Related Issue

* [[PWA-238](https://jira.corp.magento.com/browse/PWA-238)] Product Listing (Cart)

### Verification Steps

**Data fetch policy I used for ease of development has been removed so its not accidentally left in once all cart features are completed. You'll need to clear your Apollo cache if you make any cart mutations outside of this view (ie. Add to Cart).**
1. Navigate to `/cart` route, verify empty cart message is shown
2. Add Products to your cart, both simple and configurable
3. Go back to `/cart` route, verify details are displayed and aligned with mocks
4. Switch to mobile view and verify layout collapses and adheres to mocks

## Checklist


packages/venia-ui/lib/components/ProductListing/product.js
```diff
@@ -0,0 +1,77 @@
...
+            <Image
+                alt={name}
+                classes={{ image: classes.image, root: classes.imageContainer }}
+                width={IMAGE_SIZE}
+                resource={image}
+            />
+            <span className={classes.name}>{name}</span>
+            <ProductOptions
+                options={options}
+                classes={{ options: classes.options, optionLabel: {} }}
```
sirugh: An empty object is stringified to `[object Object]` in the html. You should just create an empty `.optionLabel` class in the css and pass it here.

packages/venia-ui/lib/components/ProductListing/product.js
```diff
@@ -0,0 +1,77 @@
...
+                {' ea.'}
+            </span>
+            {/** Quantity Selection to be completed by PWA-119. */}
+            <div className={classes.quantity}>- {quantity} +</div>
+            <Kebab classes={{ root: classes.kebab }}>
+                <Section
+                    text="Move to favorites"
+                    onClick={handleToggleFavorites}
+                    icon="Heart"
+                    isFilled={false}
```
sirugh: Favorites doesn't do anything right now but I think this should use a local state variable at least.

packages/venia-ui/lib/components/ProductListing/productListing.js
```diff
@@ -0,0 +1,83 @@
...
+export const GET_PRODUCT_LISTING = gql`
+    query getProductListing($cartId: String!) {
+        cart(cart_id: $cartId) {
+            ...CartBody
+        }
+        ${ProductListing.fragments.cart}
+    }
+`;
+
+export const REMOVE_ITEM_MUTATION = gql`
```
sirugh: I think this should be defined on in the `Product` component.

packages/venia-ui/lib/components/ProductListing/productListing.js
```diff
@@ -0,0 +1,83 @@
...
+        }
+    `
+};
+
+export default ProductListing;
+
+// Export query to be used by other components that may need to trigger
+// a full re-fetch.
+export const GET_PRODUCT_LISTING = gql`
+    query getProductListing($cartId: String!) {
```
sirugh: I have `queries` exported as a static member like `fragments`. Thoughts on doing that instead of exporting a const here?

```js
ProductListing.queries = {
  getProductListing: gql`...`
}
``

File: packages/venia-ui/lib/components/ProductListing/product.js (hunk 2 above)
Comment (brendanfalkowski) [reply packages/venia-ui/lib/components/ProductListing/product.js]: Maybe say "wishlist" in code (if not labels too) since that lines up to Magento's terminology.

**Comment by brendanfalkowski** [comment]
Just a note on the `ProductListing` naming. This seems a little confusing to me coming from a Magento background where "product grid" and "product list" are general terms applicable to a view mode of a category/search listing of products.

Since this component is mostly useful within cart (err, maybe withina wishlist too?) I would probably rename it `CartList`. We have all four of those to support, and each has different UI structure so just thinking ahead that this naming doesn't stretch too cleanly.

File: packages/venia-ui/lib/components/CartPage/cartPage.js [hunk 5]
```diff
@@ -3,6 +3,7 @@ import React, { useMemo } from 'react';
 import { useCartPage } from '@magento/peregrine/lib/talons/CartPage/useCartPage';
 
 import Button from '../Button';
+import ProductListing from '../ProductListing';
```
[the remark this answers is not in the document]
tjwiebell: This has a public/private API implication, but I'm okay moving if we're not considering these components public API (suppose we could always export in index to denote they are public).

packages/venia-ui/lib/components/ProductListing/product.js
sirugh: I think for now Tommy is using the existing options kebab which hardcodes the label.

packages/venia-ui/lib/components/CartPage/cartPage.js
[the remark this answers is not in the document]
sirugh: Maybe it's `CartPage` that should actually just be a RootComponent and these should all just be top-level `components`. Hmm.

packages/peregrine/lib/talons/ProductListing/useProductListing.js
```diff
@@ -0,0 +1,30 @@
...
+        fetchPolicy: 'cache-and-network'
+    });
+
+    let items = [];
+    if (!error && !loading) {
+        items = data.cart.items;
+    }
+
+    if (error) {
+        console.error(error);
```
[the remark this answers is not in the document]
jimbo: I think it's good habit, not overkill. 👍

packages/venia-ui/lib/components/CartPage/cartPage.js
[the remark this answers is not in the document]
jimbo: > Maybe it's `CartPage` that should actually just be a RootComponent and these should all just be top-level `components`. Hmm.

We're moving away from RootComponents being anything but shells bound to a route. I would prefer that `CartPage` and all pages that actually have DOM, style, talons, etc., be in `components`. See #1953.

packages/venia-ui/lib/components/ProductListing/product.css
```diff
@@ -0,0 +1,71 @@
+.root {
+    display: grid;
+    grid-gap: 0.5rem 1rem;
+    grid-template-areas:
+        'image name quantity kebab'
+        'image options quantity kebab'
+        'image price quantity kebab';
+    grid-template-columns: 100px 2fr max-content min-content;
```
jimbo:
```suggestion
    grid-template-columns: 100px 1fr max-content min-content;
```

---
If you only have one flexible column, `1fr` is enough.

packages/venia-ui/lib/components/ProductListing/product.css
```diff
@@ -0,0 +1,71 @@
...
+
+.imageContainer {
+    grid-area: image;
+}
+
+.image {
+    background-color: rgb(var(--venia-grey));
+    border: solid 1px rgb(var(--venia-border));
+    border-radius: 2px;
+    height: 100px;
```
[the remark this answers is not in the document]
jimbo: If it's for an image, I think it's fine.

packages/venia-ui/lib/components/ProductListing/product.js
jimbo: I know this API isn't pretty, but props for using it. At some point we'll make a better solution for this use case.

jimbo:
> Just a note on the `ProductListing` naming. This seems a little confusing to me coming from a Magento background where "product grid" and "product list" are general terms applicable to a view mode of a category/search listing of products.
> 
> Since this component is mostly useful within cart (err, maybe within a wishlist too?) I would probably rename it `CartList`. We have all four of those to support, and each has different UI structure so just thinking ahead that this naming doesn't stretch too cleanly (for us). But maybe it could for simpler sites.

Agree on all points. 👍

packages/venia-ui/lib/components/ProductListing/product.js
tjwiebell: I'm going to stick with the copy in the mocks, as I think this functionality is intended to be different from the auth'd wishlist feature. I'll make it usable client-side for @sirugh though 🤗

tjwiebell:
> Just a note on the `ProductListing` naming. This seems a little confusing to me coming from a Magento background where "product grid" and "product list" are general terms applicable to a view mode of a category/search listing of products.
> 
> Since this component is mostly useful within cart (err, maybe within a wishlist too?) I would probably rename it `CartList`. We have all four of those to support, and each has different UI structure so just thinking ahead that this naming doesn't stretch too cleanly (for us). But maybe it could for simpler sites.

@brendanfalkowski - I've got two suggestions related to naming and location that conflict. If I take @sirugh's advice and nest as a sub-component under `CartPage`, would that remove the overlap with existing Magento jargon? eg. `CartPage > ProductListing`.

packages/peregrine/lib/talons/ProductListing/useProductListing.js
```diff
@@ -0,0 +1,30 @@
...
+export const useProductListing = props => {
+    const { query } = props;
+
+    const [{ cartId }] = useCartContext();
+
+    const { data, error, loading } = useQuery(query, {
+        variables: { cartId },
+        // temporarily use fetch policy that validates against the network
+        // until all cart mutations are correctly updating the cache
+        fetchPolicy: 'cache-and-network'
```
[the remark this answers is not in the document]
tjwiebell: I think I'd rather just remove the fetchPolicy that made development easier, because I don't know the big picture of re-factoring all these other cart components to start monitoring Apollo cache instead of global state.

brendanfalkowski:
@tjwiebell — That's nicer from my POV. I still prefer even more explicit naming all the way down like `CartPage > CartList`. I've been on BEM naming for 9 years so I'm guilty as charged.

It's nice in an IDE being able to fuzzy search when the file name, component name, and outer context all have the same meta cues in naming. Said another way, I'd never search "productlist" to pick from product lists in multiple contexts when I'm really thinking "the product list in the cart" — hence "cartlist" is what I'd type and name it.

That "seems" to work better when we don't have generically named sub-components (ex: Thing/button, Thing/form, etc) vs (Thing/thingForm, Thing/thingButton, etc). Maybe the fuzzy sort puts more weight on file name matches the "in the path" matches. It seems to always find the exact file I meant without confusing paths that happen to match.

packages/peregrine/lib/talons/ProductListing/useProduct.js
```diff
@@ -11,7 +11,14 @@ export const useProduct = props => {
...
     const [{ cartId }] = useCartContext();
 
+    const [isRemoving, setIsRemoving] = useState(false);
+
+    const handleEditItem = useCallback(() => {
+        // Edit Item action to be completed by PWA-272.
+    }, []);
+
     const handleRemoveFromCart = useCallback(() => {
+        setIsRemoving(true);
```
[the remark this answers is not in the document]
tjwiebell: Fixed in 0ac27f8. I need to do some validation, but I'm banking on Apollo running the refetch on error, which would make this unnecessary; but if not, would put this in an odd state. Will validate that assumption, and manually trigger a refetch onError if needed.

packages/peregrine/lib/talons/ProductListing/useProduct.js
```diff
@@ -0,0 +1,54 @@
...
+        // Edit Item action to be completed by PWA-272.
+    }, []);
+
+    const handleRemoveFromCart = useCallback(() => {
+        setIsRemoving(true);
+        removeItem({
+            variables: {
+                cartId,
+                itemId: item.id
+            }
```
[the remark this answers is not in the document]
tjwiebell: Added in 0ac27f8 (and I'll fix again once your PR is merged 😉 )

packages/peregrine/lib/talons/ProductListing/useProductListing.js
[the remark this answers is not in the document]
tjwiebell: Removed in 0ac27f8

packages/peregrine/lib/talons/ProductListing/useProductListing.js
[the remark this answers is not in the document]
tjwiebell: Fixed in 0ac27f8

packages/venia-ui/lib/components/CartPage/cartPage.js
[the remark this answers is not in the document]
tjwiebell: Nested under `CartPage` in 0ac27f8

packages/venia-ui/lib/components/ProductListing/__tests__/product.spec.js
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import { act } from 'react-test-renderer';
+import { createTestInstance } from '@magento/peregrine';
+
+import Product from '../product';
+
```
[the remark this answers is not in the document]
tjwiebell: Mocked in 0ac27f8

packages/venia-ui/lib/components/ProductListing/product.js
tjwiebell: I made this change in 0ac27f8, but vscode linting is complaining about the empty rule now. Is that something we enforce?

packages/venia-ui/lib/components/ProductListing/product.js
tjwiebell: Made this functional client-side in 0ac27f8

packages/venia-ui/lib/components/ProductListing/productListing.js
tjwiebell: Moved mutation in 0ac27f8

packages/venia-ui/lib/components/ProductListing/productListing.js
tjwiebell: Discussed IRL, this is now obsolete.

packages/peregrine/lib/talons/ProductListing/useProduct.js
```diff
@@ -0,0 +1,54 @@
...
+        });
+    }, [cartId, item.id, removeItem]);
+
+    return {
+        handleEditItem,
+        handleRemoveFromCart,
+        handleToggleFavorites: () => {},
```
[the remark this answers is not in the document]
tjwiebell: By request of @sirugh, I made this functional client-side in 0ac27f8, which resolves this render problem. Good to know though 😄

packages/venia-ui/lib/components/ProductListing/product.css
tjwiebell: Fixed in 0ac27f8

packages/peregrine/lib/talons/ProductListing/useProduct.js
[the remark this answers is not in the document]
sirugh: OK, I believe price summary was merged :D

packages/peregrine/lib/talons/CartPage/PriceSummary/usePriceSummary.js
```diff
@@ -25,10 +25,7 @@ export const usePriceSummary = props => {
     const [{ cartId }] = useCartContext();
 
     const [fetchPriceSummary, { error, loading, data }] = useLazyQuery(
-        props.query,
-        {
-            fetchPolicy: 'no-cache'
```
sirugh: You removed this cause refetch refreshes this data right?

tjwiebell: Correct. Since I didn't know where to track the remaining work for making sure aggressive fetch policies are removed on completion, I just removed them entirely. You'll need to delete the apollo-cache entry in local storage and refresh, and you should see live data.

sirugh: This is on a fresh start, cleared cache and everything.

If you start from the `/cart` page I believe it never is updated after future operations.

packages/peregrine/lib/talons/CartPage/ProductListing/useProductListing.js
```diff
@@ -0,0 +1,30 @@
...
+
+import { useCartContext } from '../../../context/cart';
+import { useEffect } from 'react';
+
+export const useProductListing = props => {
+    const { query } = props;
+
+    const [{ cartId }] = useCartContext();
+
+    const { data, error, loading } = useQuery(query, {
```
sirugh: You're going to want to change this to a lazy query since `cartId` can change. Will this work if you sign out/sign in?

packages/peregrine/lib/talons/CartPage/PriceSummary/usePriceSummary.js
sirugh: After discussion we have opened [a ticket](https://jira.corp.magento.com/browse/PWA-321) to track the ideal state work.

packages/peregrine/lib/talons/CartPage/ProductListing/useProductListing.js
sirugh: However, I'm not sure if that's necessary as cartId should update and the `useQuery` hook should update appropriately...

packages/peregrine/lib/talons/CartPage/ProductListing/useProductListing.js
```diff
@@ -0,0 +1,30 @@
...
+        variables: { cartId }
+    });
+
+    let items = [];
+    if (!error && !loading) {
+        items = data.cart.items;
+    }
+
+    useEffect(() => {
```
sirugh: This is reachable code, but I like to put effects/hooks at the top since they cannot be conditional. Not super important.

packages/peregrine/lib/talons/CartPage/ProductListing/useProductListing.js
sirugh: Ok, so this will work correctly IF `cartId` is present, but you do need to make sure you aren't calling the query when `cartId` is undefined. After adding an item, go to the page and sign in/out. You should see a console error for graphql related to missing `cartId`.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -0,0 +1,80 @@
...
+        setIsRemoving(true);
+        const { error } = await removeItem({
+            variables: {
+                cartId,
+                itemId: item.id
+            },
+            refetchQueries: [
+                {
+                    query: refetchCartQuery,
+                    variables: { cartId }
```
sirugh: We talked offline about this -- this kinda sucks because this is sort of a magical coupling between query/variables. This component doesn't actually know what variables these queries expect. 🤷‍♂

packages/peregrine/lib/talons/CartPage/ProductListing/useProductListing.js
tjwiebell: I was comfortable having error state === empty cart state, but agree its best to save the network request we know will fail. Added in f622664.

packages/peregrine/lib/talons/CartPage/ProductListing/useProductListing.js
tjwiebell: Moved in f622664

packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
```diff
@@ -0,0 +1,68 @@
...
+import Product from './product';
+
+const ProductListing = props => {
+    const talonProps = useProductListing({ query: GET_PRODUCT_LISTING });
+    const { isLoading, items } = talonProps;
+
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    if (isLoading) {
+        return <LoadingIndicator>{`Fetching Cart...`}</LoadingIndicator>;
```
sirugh: Do we want a loading indicator here? I thought we were trying to get rid of them -- I just display nothing in summary when loading.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -0,0 +1,80 @@
...
+    const handleToggleFavorites = useCallback(() => {
+        setIsFavorite(!isFavorite);
+    }, [isFavorite]);
+
+    const handleEditItem = useCallback(() => {
+        // Edit Item action to be completed by PWA-272.
+    }, []);
+
+    const handleRemoveFromCart = useCallback(async () => {
+        setIsRemoving(true);
```
sirugh: You . may not have to use a local state. The `removeItem` mutation returns an object with a `loading` prop. I'm not sure if it waits for refetchQueries though.

packages/venia-ui/lib/components/ProductListing/product.js
jimbo: No, VSC's linter is not something we care about, so an empty rule is fine. That said, you may safely delete empty rules. The component should not throw an error if you try to apply `classes.foo` but there is no `.foo {}` defined in the CSS module.

dpatil-magento:
QA Pass, waiting for reviewer approval on latest commits.

packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
jimbo: It's really a case-by-case thing. You're right, we don't want loading indicators everywhere, but we also don't want to have a whole view suddenly just pop in. Ultimately, the way for us to avoid having spinners everywhere will have to be something like art direction but for loading states; until we have that, having an indicator here is fine.

schensley:
@tjwiebell the UI looks good.  The 'adjust quantity control' is yet to be implemented at this time, but the Product Listing itself is approved by UX.

PR #2095, opened 2020-01-10
[chore] Initial attempt at fixing validator for venia-ui

sirugh:
# Description

`yarn run validate-queries` was incomplete. It only validated `venia-concept` which no longer has any any graphql. That's not to say it won't always, but we should at least have had the validator check `venia-ui`.

**Note**: I added the validator to `pre-push` but I also changed the errors to warnings so as to not kill the push. We can discuss in this PR.

## Related Issue

Closes [PWA-251](https://jira.corp.magento.com/browse/PWA-251)

### Verification Stakeholders
Ya'll.
### Verification Steps

1. Run `yarn run validate-queries`. Result should display.

[Image from Gyazo](https://gyazo.com/ead83ad3c418adc8d125fa12eff71d0c)

## Checklist


package.json
```diff
@@ -90,7 +90,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "yarn run prettier:check && yarn run lint"
+      "pre-push": "yarn run prettier:check && yarn run lint && yarn run validate-queries"
```
sirugh: This is debatable. Do we want to run this on push? I think so.

packages/graphql-cli-validate-magento-pwa-queries/lib/index.js
```diff
@@ -67,7 +67,7 @@ async function validateQueries(context, argv) {
 
         // Validate our queries against that schema.
         context.spinner.start('Finding queries in files...');
-        const validator = getValidator({ clients, project });
+        const validator = getValidator({ clients, project, schemaPath });
```
sirugh: Now the validator will run the schema stored in the path rather than doing a "magical" search for the file within the directory the script is running. Basically we can run this from venia-concept but still check files in the sibling package, venia-ui.

packages/graphql-cli-validate-magento-pwa-queries/lib/index.js
```diff
@@ -81,19 +81,23 @@ async function validateQueries(context, argv) {
...
+            console.warn(chalk.red('Found some potential issues:'));
 
             const formatter = validator.getFormatter();
             console.log(formatter(report.results));
 
             console.log(getErrorResolutionDetails());
 
-            process.exit(exitCodes.FAILURE);
+            if (report.errorCount) {
+                process.exit(exitCodes.FAILURE);
```
sirugh: I only wanted to fail the exit on actual errors.

packages/graphql-cli-validate-magento-pwa-queries/lib/index.js
```diff
@@ -116,20 +120,23 @@ function getSupportedArguments(args) {
...
+function getValidator({ clients, project, schemaPath }) {
     const clientRules = clients.map(clientName => ({
         env: clientName,
-        projectName: project
+        projectName: project,
+        schemaJsonFilepath: schemaPath
     }));
 
-    const ruleDefinition = ['error', ...clientRules];
+    const ruleDefinition = ['warn', ...clientRules];
```
sirugh: This is debatable. I know some folks think warnings are code smells. If we want to leave this as `error` we should then fix all the issues in the current graphql files (see the screenshot in the PR).

packages/venia-concept/.graphqlconfig
```diff
@@ -8,7 +8,7 @@
                 },
                 "validate-magento-pwa-queries": {
                     "clients": ["apollo", "literal"],
-                    "filesGlob": "src/**/*.{js,graphql,gql}"
+                    "filesGlob": "../{venia-ui,venia-concept}/{lib,src}/**/**/*.{js,graphql,gql}"
```
sirugh: I'm bad with patterns...do I need the extra `/**`?

sirugh:
I will fix tests once we land on a consensus for the debatable bits.

package.json
jimbo: I mean, we don't currently. I'm okay with adding it since I always push with `--no-verify` anyway.

packages/venia-concept/.graphqlconfig
jimbo: I don't think you do.

package.json
sirugh: Let's ask the team

sirugh: @dpatil-magento can we add this script to the CI checks?

sirugh: https://jira.corp.magento.com/browse/PWA-341 was created -- team says NO to prepush :D

packages/graphql-cli-validate-magento-pwa-queries/lib/index.js
jimbo: Yes, in this repo, we do prefer errors over warnings. That said, if you'd like to create an issue to fix all the warnings (and then flip it back from `warn` to `error`), that would be fine.

packages/graphql-cli-validate-magento-pwa-queries/lib/index.js
```diff
@@ -128,8 +133,18 @@ function getValidator({ clients, project }) {
...
-            'graphql/no-deprecated-fields': ruleDefinition
+            'graphql/capitalized-type-name': ruleDefinition,
+            'graphql/named-operations': ruleDefinition,
+            'graphql/no-deprecated-fields': ['warn', ...clientRules],
+            'graphql/required-fields': [
+                'error',
+                ...clients.map(clientName => ({
+                    env: clientName,
+                    schemaJsonFilepath: schemaPath,
+                    requiredFields: ['id']
```
sirugh: > The Required Fields rule validates that any specified required field is part of the query, but only if that field is available in schema. This is useful to ensure that query results are cached properly in the client.

https://github.com/apollographql/eslint-plugin-graphql#required-fields-validation-rule

dpatil-magento:
@jimbo Don't see any performance degradation on this PR, so merging this one.

PR #2101, opened 2020-01-14
Object Accessor Util

revanth0212:
## Description

Adding Object Accessor utilities to ease up code blocks like `obj && obj.bar && obj.bar.baz` with a simple and powerful API.

`(obj && obj.bar && obj.bar.baz) || 'defaultValue'` -> `pathOr('defaultValue', ['bar', 'baz'], obj)`

## Related Issue
Closes nothing. Created this out of self-interest.

### Verification Steps
None

## Checklist
None

### DIFF
--- a/packages/peregrine/lib/util/objectAccessor.js
+++ b/packages/peregrine/lib/util/objectAccessor.js
@@ -1,0 +1,35 @@
+function isNull(val) {
+  return val === null;
+}
+
+function isUndefined(val) {
+  return val === undefined;
+}
+
+function or(fn1, fn2) {
+  return function(val) {
+    return fn1(val) || fn2(val);
+  };
+}
+
+function isNullOrUndefined(val) {
+  return or(isNull, isUndefined)(val);
+}
+
+export function pathOr(backup, keyArray, obj) {
+  const [key, ...rest] = keyArray;
+  const result = obj[key];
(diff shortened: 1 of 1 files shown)


supernova-at:
Not sure the timetable on when this will be fully cross-browser compatible but: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

tjwiebell:
> Not sure the timetable on when this will be fully cross-browser compatible but: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

Also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

jimbo:
Thank you for the suggestion, but we made a conscious decision at the outset of the project to avoid this kind of utility belt code, and I think that choice has been effective. Plenty of [existing solutions](https://lodash.com/docs/4.17.11#get) exist, but we've chosen not to use them.

In this particular case, chaining `a && a.b && a.b.c` is verbose for a reason: it makes a lot of assumptions about the shape of an object, which is not ideal—and which is why it rarely occurs in our codebase.

revanth0212:
This was a fun utility I was working on. Forgot about this forever 🤣

PR #2112, opened 2020-01-22
bugfix: do not render things when no items are in cart

sirugh:
## Description

We were rendering the `PriceAdjustments` and `PriceSummary` components regardless of whether we _should_.

This PR adds a `getCartDetails` query that does an initial fetch for cart information, including `total_quantity` which it then uses to conditionally render the subcomponents.

This PR also removes the `refetchQueries` from the `removeItem` mutation. It now uses named fragments! This means that after the remove item mutation is called you should see no subsequent refetches for new data, and yet the UI should update!

## Related Issue

Closes [PWA-326](https://jira.corp.magento.com/browse/PWA-326).

### Verification Stakeholders
@schensley @jimbo 
### Verification Steps

1. Clear your storage and go to `/cart`. You should see no items, no adjustments, and no summary.
2. Add a product and go back to `/cart. You should see the item, adjustments, and summary.
3. Remove the item with the kebab toggle. The item, adjustments, and summary views should disappear.

[Image from Gyazo](https://gyazo.com/e0aa5e87be87899aa119cda9c5c1f191)

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/product.js
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/product.js
@@ -9,8 +9,9 @@ import ProductOptions from '../../MiniCart/productOptions';
 import Section from '../../MiniCart/section';
 import Image from '../../Image';
 import defaultClasses from './product.css';
-import { GET_PRODUCT_LISTING } from './productListing';
-import { PriceSummaryQuery } from '../PriceSummary/priceSummary';
+import { CartPageFragment } from '../cartPageFragments';
+import { ProductListingFragment } from './productListingFragments';
+import { PriceSummaryFragment } from '../PriceSummary/priceSummaryFragments';
 
 const IMAGE_SIZE = 100;
 
@@ -18,8 +19,6 @@ const Product = props => {
     const { item } = props;
     const talonProps = useProduct({
         item,
-        refetchPriceQuery: PriceSummaryQuery,
-        refetchCartQuery: GET_PRODUCT_LISTING,
         removeItemMutation: REMOVE_ITEM_MUTATION
     });
     const {
@@ -95,7 +94,13 @@ export const REMOVE_ITEM_MUTATION = gql`
         removeItemFromCart(input: { cart_id: $cartId, cart_item_id: $itemId }) {
             cart {
                 id
+                ...CartPageFragment
+                ...ProductListingFragment
+                ...PriceSummaryFragment
             }
         }
     }
+    ${CartPageFragment}
+    ${ProductListingFragment}
+    ${PriceSummaryFragment}
 `;
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/product.js
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/product.js
@@ -10,8 +10,6 @@ import Section from '../../MiniCart/section';
 import Image from '../../Image';
 import defaultClasses from './product.css';
 import { CartPageFragment } from '../cartPageFragments';
-import { ProductListingFragment } from './productListingFragments';
-import { PriceSummaryFragment } from '../PriceSummary/priceSummaryFragments';
 
 const IMAGE_SIZE = 100;
 
@@ -95,12 +93,8 @@ export const REMOVE_ITEM_MUTATION = gql`
             cart {
                 id
                 ...CartPageFragment
-                ...ProductListingFragment
-                ...PriceSummaryFragment
             }
         }
     }
     ${CartPageFragment}
-    ${ProductListingFragment}
-    ${PriceSummaryFragment}
 `;
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/product.js
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/product.js
@@ -10,6 +10,8 @@ import Section from '../../MiniCart/section';
 import Image from '../../Image';
 import defaultClasses from './product.css';
 import { CartPageFragment } from '../cartPageFragments';
+import { ProductListingFragment } from './productListingFragments';
+import { PriceSummaryFragment } from '../PriceSummary/priceSummaryFragments';
 
 const IMAGE_SIZE = 100;
 
@@ -93,8 +95,12 @@ export const REMOVE_ITEM_MUTATION = gql`
             cart {
                 id
                 ...CartPageFragment
+                ...ProductListingFragment
+                ...PriceSummaryFragment
             }
         }
     }
     ${CartPageFragment}
+    ${ProductListingFragment}
+    ${PriceSummaryFragment}
 `;
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
@@ -6,6 +6,7 @@ import { mergeClasses } from '../../../classify';
 import LoadingIndicator from '../../LoadingIndicator';
 import defaultClasses from './productListing.css';
 import Product from './product';
+import { ProductListingFragment } from './productListingFragments';
 
 const ProductListing = props => {
     const talonProps = useProductListing({ query: GET_PRODUCT_LISTING });
@@ -28,41 +29,13 @@ const ProductListing = props => {
     }
 };
 
-export default ProductListing;
-
-export const CartBody = gql`
-    fragment CartBody on Cart {
-        id
-        items {
-            id
-            product {
-                name
-                small_image {
-                    url
-                }
-            }
-            prices {
-                price {
-                    currency
-                    value
-                }
-            }
-            quantity
-            ... on ConfigurableCartItem {
-                configurable_options {
-                    option_label
-                    value_label
-                }
-            }
-        }
-    }
-`;
-
 export const GET_PRODUCT_LISTING = gql`
     query getProductListing($cartId: String!) {
         cart(cart_id: $cartId) {
-            ...CartBody
+            ...ProductListingFragment
         }
-        ${CartBody}
+        ${ProductListingFragment}
     }
 `;
+
+export default ProductListing;
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
@@ -24,9 +24,9 @@ const ProductListing = props => {
         ));
 
         return <ul className={classes.root}>{productComponents}</ul>;
-    } else {
-        return <h3>There are no items in your cart.</h3>;
     }
+
+    return null;
 };
 
 export const GET_PRODUCT_LISTING = gql`
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
@@ -34,8 +34,8 @@ export const GET_PRODUCT_LISTING = gql`
         cart(cart_id: $cartId) {
             ...ProductListingFragment
         }
-        ${ProductListingFragment}
     }
+    ${ProductListingFragment}
 `;
 
 export default ProductListing;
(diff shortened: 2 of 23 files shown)


packages/venia-ui/lib/components/CartPage/__tests__/__snapshots__/cartPage.spec.js.snap
```diff
@@ -0,0 +1,192 @@
...
+    >
+      <span>
+        Sign In
+      </span>
+    </button>
+  </div>
+  <div>
+    <div>
+      <h3>
+        There are no items in your cart.
```
sirugh: Oops this seems wrong

schensley:
@sirugh while this functions as it is supposed to, I think some addition finesse could be given to the 'fading out' of elements. 
Can this behave in such a way that first the removed product line fades out, and then the Pricing Summary and the accordion of modules (Shipping, Coupon, etc.) fades out after?  It feels abrupt to have all fade out (at least so quickly).  Thoughts?

packages/venia-ui/lib/components/CartPage/cartPageFragments.js
```diff
@@ -0,0 +1,7 @@
+import gql from 'graphql-tag';
+
+export const CartPageFragment = gql`
+    fragment CartPageFragment on Cart {
+        total_quantity
```
sirugh: Should this fragment just compose itself of the other block components ie:

```js
export const CartPageFragment = gql`
    fragment CartPageFragment on Cart {
        total_quantity
        ...PriceSummaryFragment
        ...PriceAdjustmentsFragment
        ...ProductListingFragment
    }
`
```

sirugh: And then in mutations, instead of requesting individual fragments we could just request the `CartPageFragment`?

```js
export const REMOVE_ITEM_MUTATION = gql`
    mutation removeItem($cartId: String!, $itemId: Int!) {
        removeItemFromCart(input: { cart_id: $cartId, cart_item_id: $itemId }) {
            cart {
                id
                ...CartPageFragment
                ...ProductListingFragment
                ...PriceSummaryFragment
            }
        }
    }
    ${CartPageFragment}
    ${ProductListingFragment}
    ${PriceSummaryFragment}
`;
```

vs
```js
export const REMOVE_ITEM_MUTATION = gql`
    mutation removeItem($cartId: String!, $itemId: Int!) {
        removeItemFromCart(input: { cart_id: $cartId, cart_item_id: $itemId }) {
            cart {
                id
                ...CartPageFragment
            }
        }
    }
    ${CartPageFragment}
`;
```



The former approach is more "slim" in that it requests specific data whereas the latter just does a bulk fetch on mutate.

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
```diff
@@ -95,7 +92,9 @@ export const REMOVE_ITEM_MUTATION = gql`
         removeItemFromCart(input: { cart_id: $cartId, cart_item_id: $itemId }) {
             cart {
                 id
+                ...CartPageFragment
```
sirugh: When performing a mutation within the `CartPage` we can just refetch the entire composed fragment rather than requiring each component to fetch specifics. This may result in "overfetch" such as in the case of adding a coupon where you only really need to update the total and discounts.

packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
```diff
@@ -0,0 +1,29 @@
+import gql from 'graphql-tag';
+
+export const ProductListingFragment = gql`
+    fragment ProductListingFragment on Cart {
+        id
+        items {
+            id
+            product {
```
sirugh: The props requested in items should _probably_ be moved to a `productListingProductFragment` object that is then used here. The idea is that each visual component declares what it needs to render, and the `productListing` component is really just a wrapper that passes the data down to the `product` component anyways.

packages/venia-ui/lib/components/CartPage/cartPage.js
```diff
@@ -37,11 +42,11 @@ const CartPage = props => {
                     <ProductListing />
                 </div>
                 <div className={classes.price_adjustments_container}>
-                    <PriceAdjustments />
+                    {hasItems ? <PriceAdjustments /> : null}
```
sirugh: Should we do the same here for `ProductListing`? That component (and `PriceSummary`, actually) have a `null` render state based on the data they receive back.

If we add a conditional before rendering though, we prevent the component from mounting/making a graphQL query, which could be a win. Thoughts?

packages/venia-ui/lib/queries/getCartDetails.graphql
```diff
@@ -1,5 +1,6 @@
 query getCartDetails($cartId: String!) {
     cart(cart_id: $cartId) {
+        id
```
sirugh: This was missing, and potentially might cause cache misses.

However, in order for the trigger to be updated after mutations like `removeItem` we would have to start using named fragments inside the remove item mutation.

sirugh:
>  I think some addition finesse could be given to the 'fading out' of elements. Can this behave in such a way that first the removed product line fades out, and then the Pricing Summary and the accordion of modules (Shipping, Coupon, etc.) fades out after? It feels abrupt to have all fade out (at least so quickly). Thoughts?

@schensley I definitely think it can be touched up, just not in the scope of this ticket. We should make a follow-up. Can you start that conversation with Andrew as it'll be good to get some mocks that show transition requirements as in what order things disappear and the exact timing of each?

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
```diff
@@ -95,7 +92,9 @@ export const REMOVE_ITEM_MUTATION = gql`
         removeItemFromCart(input: { cart_id: $cartId, cart_item_id: $itemId }) {
             cart {
                 id
```
tjwiebell: Don't need this id stub anymore since you're using the fragment.

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
tjwiebell: I'm fine with this, under the assumption that over fetching doesn't causes unnecessary re-renders if that data didn't change.

packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
```diff
@@ -28,41 +29,13 @@ const ProductListing = props => {
...
-`;
-
 export const GET_PRODUCT_LISTING = gql`
     query getProductListing($cartId: String!) {
         cart(cart_id: $cartId) {
-            ...CartBody
+            ...ProductListingFragment
         }
-        ${CartBody}
+        ${ProductListingFragment}
```
tjwiebell: Odd that this works, but the fragment should be outside of the query body.

```
query ...
${ProductListingFragment}
```

packages/venia-ui/lib/components/CartPage/cartPage.js
```diff
@@ -8,9 +9,13 @@ import PriceSummary from './PriceSummary';
 import ProductListing from './ProductListing';
 import { mergeClasses } from '../../classify';
 import defaultClasses from './cartPage.css';
+import { CartPageFragment } from './cartPageFragments';
```
tjwiebell: I would support this overfetch if it prevented network calls in sub-components OR if you drilled the data down. Via chat you confirmed the first is not true; lets use a slimmer fragment that only grabs the data this component needs (total_quantity).

packages/venia-ui/lib/components/CartPage/cartPageFragments.js
tjwiebell: I like the idea of an encapsulating page fragment, which future proofs new features. When new data is needed on the page, a new fragment is added to the page fragment, and any existing mutations automatically pick it up. Easier to maintain, but also relies on the assumption that over fetching doesn't cause unnecessary renders.

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
sirugh: So fragments should always include it? This was one of those things I wasn't sure about. Do we just always make sure to have some identifier for caching in our queries/mutations or can we assume that fragment authors will include it?

packages/venia-ui/lib/components/CartPage/ProductListing/productListing.js
sirugh: Oops

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
sirugh: I'll have to look -- I'm not sure, it seems super fast.

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
tjwiebell: Yes, I think we should make this assumption that all fragments include id when available. Ad-hoc inclusion of id would imply we're relying on heuristic fragment matching, which requires exact field matches, which is difficult to maintain. It's not causing any harm, but could lead to confusion that you always need to add an id field, when you don't.

packages/venia-ui/lib/queries/getCartDetails.graphql
tjwiebell: These cart queries use different versions of schema, so with the inclusion of id heuristic fragment matching wouldn't hit (must be exact). Even if we switched to consuming fragments here, we'd have to update the components to the new schema, which is making the scope creepy.

**If** this was a priority, I would probably just do a manual cache update in our mutations.

packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
tjwiebell: I like this idea, a great extension of what you did with PriceSummary. This did bring to my attention we're inconsistently colocating fragments though, and should probably decide on a single location.

1. Dedicated file (`<componentName>Fragments.js`)
2. Named export in component (eg. `discountSummary`)

Of course the Apollo docs use static properties, which we know has the issue of cyclical dependencies when children need parent fragments (also an issue with named export). Maybe this is another reason to switch to a single root component query, and use apollo-anywhere filter. This still would require `CartPageFragment` be exported outside of the root component though...

Tough questions, but I'm okay iterating on it and not solving in the scope of this PR.

sirugh: Yea I just wanted to call it out :)

packages/venia-ui/lib/queries/getCartDetails.graphql
sirugh: As we discovered just now this works because Apollo sees "partial data" and thus makes a network call.

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
sirugh: Do you still feel this way? I kind of prefer that we let fragments only define the data they need.

tjwiebell: I don't think fragments should need any additional config besides inclusion to work; without id in the fragment, developers would need to be aware they need it for cache updates to work.

supernova-at: It makes sense to me that the base "wrapping" query would include the `id` field, not the fragments.

sirugh: This seems like it could go either way -- either way we need to include `id` somewhere and it does not hurt to declare it multiple times so I'll just leave it as is.

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
sirugh: For now I left the fetch policy so this doesn't matter.

packages/venia-ui/lib/components/CartPage/cartPageFragments.js
sirugh: Seems like over fetching would only cause re-renders if the data changes.

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
tjwiebell: Great compromise 🤝

PR #2114, opened 2020-01-24
[PWA-178] Gift options support in cart page

revanth0212:
## Description

Gift Options Support in Venia Cart Page

## Related Issue
Closes [PWA-178](https://jira.corp.magento.com/browse/PWA-178)

### Verification Stakeholders
@soumya-ashok 
@jimbo 

### Verification Steps
1. Add an item to the cart and go to the cart page.
2. Expand the `See Gift Options` section.
3. You should see the Gift Options.

image

image

## Checklist
* GQL does not support Gift Options yet. Make sure a rest call is made to save the options.


packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -38,6 +39,65 @@ persistCache({
     storage: window.localStorage
 });
 
+const initialData = {
+    __typename: 'CartData',
```
revanth0212: I might be wrong on this kind of a use case. Let me know if that isn't how we should be doing it. Will make the changes.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions/GiftOptions.js
```diff
@@ -0,0 +1,69 @@
...
+    const [
+        { includeGiftReceipt, includePrintedCard, giftMessage },
+        {
+            toggleIncludeGiftReceiptFlag,
+            toggleIncludePrintedCardFlag,
+            updateGiftMessage
+        }
+    ] = useGiftOptions();
+
+    const setGiftMessage = useCallback(
```
revanth0212: I wasn't sure if I should be having hooks in the view layer. Thought it made more sense here than in the talon. Let me know if this needs to change.

sirugh: Yep - only UI code lives in components. Logic lives in talons. move callbacks into the talon.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -38,6 +39,65 @@ persistCache({
...
+                gift_message
+            };
+        }
+    },
+    Mutation: {
+        set_gift_options: (_, variables, { cache }) => {
+            const prevData = cache.readQuery({ query: GET_GIFT_OPTIONS_QUERY });
+            cache.writeData({
+                data: {
+                    gift_options: {
```
sirugh: You'll want to read and write this data to the current cart rather than the root.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
```diff
@@ -0,0 +1,206 @@
...
+        setIncludePrintedCard(!includePrintedCard);
+        updateGiftOptions({
+            include_printed_card: !includePrintedCard
+        });
+    }, [updateGiftOptions, includePrintedCard, setIncludePrintedCard]);
+
+    /**
+     * Fetch gift options for a given cart id.
+     */
+    const { data } = useQuery(GET_GIFT_OPTIONS_QUERY, {
```
sirugh: Use a lazy query like in the other components within `CartPage`. We can only query for data if we have a cartId, and we don't always have a cartId (at least not right now).

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
```diff
@@ -0,0 +1,206 @@
...
+            10000,
+            {
+                leading: false
+            }
+        );
+    }, []);
+
+    const updateGiftMessage = useCallback(
+        newGiftMessage => {
+            setGiftMessage(newGiftMessage);
```
sirugh: I'm not sure you actually need to use a state value here. Think about this like the apollo cache _is_ the state. Updates to the cache (after you do some mutation) are automatically picked up by the query hook, which causes a render update.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
```diff
@@ -0,0 +1,206 @@
...
+        $include_printed_card: Boolean
+        $gift_message: String
+    ) {
+        set_gift_options(
+            cart_id: $cart_id
+            include_gift_receipt: $include_gift_receipt
+            include_printed_card: $include_printed_card
+            gift_message: $gift_message
+        ) @client
+    }
```
sirugh: Your custom mutation should apply the new options to the cart cache and then return that entire object. See places where we use `...CartPageFragment`.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -38,6 +39,65 @@ persistCache({
...
+            include_gift_receipt
+            include_printed_card
+            gift_message
+        }
+    }
+`;
+
+const resolvers = {
+    Query: {
+        gift_options: (_, __, { cache }) => {
```
sirugh: I'm not sure if these can be async but you may consider moving the REST calls here to the resolvers. Basically you would only interact with graphql in the component/talon but the resolvers would actually handle doing the rest call and update the cache accordingly.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
```diff
@@ -0,0 +1,206 @@
...
+    );
+
+    /**
+     * Throttling message update. Only make 1 mutation
+     * every 10 seconds. This is to save on bandwidth.
+     *
+     * More info: https://lodash.com/docs/4.17.15#throttle
+     */
+    const throttledMessageUpdate = useMemo(() => {
+        return throttle(
```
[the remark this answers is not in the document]
revanth0212: Debounce just delays the calls. Throttle combines all the calls into a single call for a given `wait`. Throttle makes more sense in this case. Check out the difference between debounce and throttle for more info.

packages/venia-ui/lib/drivers/adapter.js
revanth0212: Yes, I am aware of that. But I was having a hard time writing the local resolvers and local state with the current cart. Hence I moved it to the top level. I was in a view that this code is only useful until the Gql team comes up with the support for gift options. Once that happens, this whole block of code is scrapped off.

packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.js
```diff
@@ -84,6 +89,7 @@ export const GiftCardSummaryFragment = IS_EE
       `
     : gql`
           fragment GiftCardSummaryFragment on Cart {
+              id
```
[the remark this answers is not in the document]
revanth0212: It wasn't working. I keep seeing Apollo errors about unknown fragment because the fragment definition was missing from the query. Apparently, as I mentioned `__typename` is not enough. When I added `id`, it started working. Without a key in the fragment other than __typename, this is not working.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
```diff
@@ -0,0 +1,206 @@
...
+            (updateGiftOptions, newGiftMessage) => {
+                updateGiftOptions({
+                    gift_message: newGiftMessage
+                });
+            },
+            10000,
```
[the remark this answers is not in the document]
revanth0212: I agree, 10 seconds is a lot. But the REST endpoint is throwing errors if we make the call to save the data, too quickly. I have tried 3 seconds, 5 seconds and 10 seconds. We can use a `Apply Gift Options` button like in Luma or save only when the user clicks on Checkout.

@soumya-ashok @schensley

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions/giftOptions.css
```diff
@@ -0,0 +1,7 @@
+.root {
+}
+
+ul {
+    padding: 0rem;
+    padding-top: 1rem;
```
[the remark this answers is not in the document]
revanth0212: Well, the section component already has `1.5rem` padding all around it, so I thought this padding will add into the section padding which sits above it. Will check and fix if I am wrong.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
revanth0212: Cart cache is temp, only till the GQL team comes up with gift options coverage. I don't think we should be using it.

Additionally, waiting for a whole mutation/REST call to check or uncheck a box, is very snappy and does not provide a good UI experience.

packages/venia-ui/lib/drivers/adapter.js
revanth0212: Resolvers are temporary. I went with the easiest temporary option here. Making a network call from the hook gives me more control over when to make the call, for instance in case of gift message only send the request every 10 seconds. Yes, I can do this from the resolver as well, but the resolver can not differentiate between the checkbox and gift message.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
revanth0212: As I mentioned in the other comment, waiting for a mutation to change the state of a UI component creates a bad user experience. Mutations on an average are taking around 600-800ms in Venia as of today, that is not an acceptable wait to check a box. My point of using a local state is to create a better UI experience over a couple of state variables.

We can use optimistic updates from apollo as well which I have explored briefly. Our apollo usage is not matured enough to handle optimistic updates. Doing it as part of this PR is not valid. I deserves to be its own task.

Let me know if I am missing something here.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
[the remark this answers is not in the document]
sirugh: I don't think that's true. https://css-tricks.com/debouncing-throttling-explained-examples/#article-header-id-0. Similar functions but slightly different. If the delay was 300ms, denounce would only income 300ms after the last keypress but throttle would call multiple times if you have keypresses over say 1000ms

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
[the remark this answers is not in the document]
sirugh: Ok yea I think it'd be good to use a button here then. Interested in ux opinions

packages/venia-ui/lib/drivers/adapter.js
sirugh: The flow I imagined was:

User input > call mutation > call rest > augment cache with rest results

Right now we're doing both sort of simultaneously, which isn't necessarily wrong I guess, but it puts all the logic of doing the fetch within the component in a manner that we're trying to move away from in the long run.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
sirugh: I can see the confusion here. If gift options are a single operation that are guaranteed not to modify the cart prices or other values then this is fine. I'm not sure what the docs say about it with REST. Do you just make a single call (apply gift options) and continue with checkout? Or is there a second step like "apply gift options **and then** fetch the new total"?

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
sirugh: You're right -- I was thinking about this from the perspective of how I implemented "coupons". Those have a button that, when clicked, displays a wait state while the mutation resolves.

This is a bit different though since UX didn't provide direction about any submit button, which means this would have to submit on blur, or after some timeout :/

packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.js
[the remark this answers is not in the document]
sirugh: Haha so for some reason I was seeing the error still after checking out your PR. Turns out I didn't fetch the right PR :P Nice find.

packages/venia-ui/lib/drivers/adapter.js
sirugh: > this code is only useful until the Gql team comes up with the support for gift options.

True -- but that may be a very long time, and there are other advantages to containing the options or other cart-specific state within the cart. We wouldn't want other carts to start showing state from a previous one :o

packages/venia-ui/lib/drivers/adapter.js
sirugh: @tjwiebell @jimbo lemme know if I'm trying to scope creep here.

schensley:
The latest screen caps of the UI look pretty good, and functions as designed.
_Being fussy in terms of design, there is a tab too much white space between section header and first checkbox option._
UX Approved.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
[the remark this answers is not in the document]
revanth0212: Yeah, isn't that what we want? Autosave for every certain time period? As you said, in 1000 ms throttle would make the call 3 times, but with debounce, the user might leave the page once he is done with the typing before the timeout and the whole data is gone. Let me know what is the right use case, will make the change 👍

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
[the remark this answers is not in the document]
revanth0212: I had a talk with @soumya-ashok. She feels, we can wait for the GQL team to be done with their work and then save the data to cart using GQL mutations. Till then, local cache should be good. Meaning no button on the side or no REST calls. Yeehh

packages/venia-ui/lib/drivers/adapter.js
revanth0212: Done. Now the gift options are scoped to a cart id.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
revanth0212: We should be calling the gift options rest call when the user clicks on checkout. Till then, it will be stored in the apollo cache.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -38,6 +39,74 @@ persistCache({
...
+            include_gift_receipt
+            include_printed_card
+            gift_message
+        }
+    }
+`;
+
+const resolvers = {
+    Query: {
+        gift_options: (_, { cart_id }, { cache }) => {
```
sirugh: There is an [alternative approach](https://www.apollographql.com/docs/react/development-testing/client-schema-mocking/#gatsby-focus-wrapper) which is to define a `typeDef` that extends some existing schema, like `Cart` and then uses resolvers off that type to handle the new fields. I wonder which is "better"?

```js
const typeDefs = gql`
    extend type Cart {
        include_gift_receipt: Boolean
        include_printed_card: Boolean
        gift_message: String
    }
`;

const resolvers = {
    Cart: {
        include_gift_receipt: (...) => {...}
      ...
    }
}
```

When you use a new typedef you do several things including:

1) Inform other developers of your intent behind the new fields by defining their types ie String, Boolean, etc.
2) Lets Apollo Dev Tools (and maybe GraphIQL?) "see" the new fields so you can actually interact with them through the browser: 

[Image from Gyazo](https://gyazo.com/3a152fc50e1ea5b61fff5841dd5a0b1d)

That said, I attempted to use `typeDefs` in the client definition and I ran into several issues with the cache initializing/persisting properly.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+/**
+ * Local query. GQL support is not available as of today.
+ *
+ * Once available, we can change the query to match the schema.
+ */
+const GET_GIFT_OPTIONS_QUERY = gql`
+    query getGiftOptions($cart_id: String) {
+        gift_options(cart_id: $cart_id) @client {
```
sirugh: Is it possible to use the `cart` query from the server schema and just use `@client` fields on the custom props? I think that's the approach you'd want to take if you defined your own typedefs.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
```diff
@@ -0,0 +1,126 @@
...
+     * More info: https://lodash.com/docs/4.17.15#throttle
+     */
+    const throttledMessageUpdate = useMemo(() => {
+        return throttle(
+            (updateGiftOptions, newGiftMessage) => {
+                updateGiftOptions({
+                    gift_message: newGiftMessage
+                });
+            },
+            5000,
```
sirugh: 5 seconds is pretty long. Since this no longer depends on a network request and is now just saving to cache maybe we can reduce this to half a second?

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -38,6 +39,74 @@ persistCache({
...
+                        include_printed_card,
+                        gift_message,
+                        id: cart_id,
+                        __typename: 'Cart'
+                    }
+                }
+            });
+
+            /**
+             * We do not return anything on purpose.
```
sirugh: This is a good call-out. Since the component renders based on `data` changing if we start returning data we will cause a weird UX. 👍

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -38,6 +39,74 @@ persistCache({
...
+             * This is how the `cacheKeyFromType` saves the
+             * cart data in the `InMemoryCache`.
+             */
+            const cartIdInCache = `Cart:${cart_id}`;
+
+            const {
+                include_gift_receipt,
+                include_printed_card,
+                gift_message
+            } = cache.data.data[cartIdInCache];
```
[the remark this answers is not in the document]
revanth0212: It wasn't working. I tried them all, maybe my GQL is bad. I had to drop down to this as my last resort.

packages/venia-ui/lib/drivers/adapter.js
revanth0212: I did try the typedefs and ran into all sorts of troubles. According to some GitHub issues on Apolo's page, typedefs and other UI features are still not matured enough to use. Hence I stayed away from them.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
revanth0212: The intent behind 5 seconds and other decisions I took while writing this hook were to save on dev time once the gift options support from the GQL team is available. The plan was to just delete the local resolvers and remove `@client` directive.

`useGiftOptions` should not be touched once gift options are available through GQL.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
```diff
@@ -0,0 +1,126 @@
...
+            cartId,
+            includeGiftReceipt,
+            includePrintedCard,
+            giftMessage
+        ]
+    );
+
+    /**
+     * Throttling message update. Only make 1 mutation
+     * every 5 seconds. This is to save on bandwidth.
```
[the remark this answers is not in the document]
revanth0212: The end goal is to save on bandwidth, if not we can make the call every second or every millisecond.

Regarding using local state, ill leave some comments for other developers to know.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -38,6 +39,74 @@ persistCache({
...
+`;
+
+const resolvers = {
+    Query: {
+        gift_options: (_, { cart_id }, { cache }) => {
+            /**
+             * This is how the `cacheKeyFromType` saves the
+             * cart data in the `InMemoryCache`.
+             */
+            const cartIdInCache = `Cart:${cart_id}`;
```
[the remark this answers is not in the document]
revanth0212: Nice 👍

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -38,6 +39,63 @@ persistCache({
     storage: window.localStorage
 });
 
+const GIFT_OPTIONS_QUERY = gql`
+    query GiftOptions {
```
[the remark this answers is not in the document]
revanth0212: Done 👍 

Now resolvers will be served from `GiftOptions.js` where the queries reside.

Nice modification @sirugh

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions.js
revanth0212: Do you mean to use it as a fragment on the Cart from remote schema? When we switch to GQL support, this will be a fragment instead of a query so I tried to use local fragments but didn't work dude. I had to settle with the query. I am not sure what you mentioned is possible or I am not aware of that. Let me know if I am missing something here.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/index.js
```diff
@@ -1 +1,3 @@
 export { default } from './priceAdjustments';
+
+export { giftOptionsResolvers } from './GiftOptions/giftOptions';
```
sirugh: Move this to an `index.js` within the `GiftOptions` folder.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/priceAdjustments.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 import { Accordion, Section } from '../../Accordion';
 
-import GiftOptions from './giftOptions';
+import GiftOptions from './GiftOptions/giftOptions';
```
sirugh: Add an `index.js` within the `GiftOptions` folder and export the default from `giftOptions.js`. Then you can just import this like `import GiftOptions from './GiftOptions'`

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions.js
sirugh: We discussed in person and concluded that getting typeDefs working locally to augment server schema was not a good use of our time :P Apollo was breaking and it doesn't tell you why.

packages/venia-ui/lib/drivers/adapter.js
sirugh: As of now we are only storing gift option choices and text to local storage through Apollo. When we work on Checkout we will need to make sure to add some step that pulls this data from the cache and submits it via REST when you click the checkout button.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
```diff
@@ -0,0 +1,130 @@
+import { useState, useCallback, useEffect, useMemo } from 'react';
+import throttle from 'lodash.throttle';
+import { useMutation, useLazyQuery } from '@apollo/react-hooks';
+
+import { useCartContext } from '@magento/peregrine/lib/context/cart';
+
+const useGiftOptions = ({ getGiftOptionsQuery, saveGiftOptionsQuery }) => {
```
jimbo: Avoid destructuring here. The pattern in our codebase is to just call this object `props`, and destructure it in the function body.

packages/peregrine/lib/talons/CartPage/GiftOptions/useGiftOptions.js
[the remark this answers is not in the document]
jimbo: Yes, I think `throttle` is the right choice.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions/giftOptions.js
```diff
@@ -0,0 +1,166 @@
...
+        toggleIncludePrintedCardFlag,
+        updateGiftMessage
+    } = useGiftOptions({
+        getGiftOptionsQuery: GIFT_OPTIONS_QUERY,
+        saveGiftOptionsQuery: SET_GIFT_OPTIONS_QUERY
+    });
+
+    const classes = useMemo(() => mergeClasses(defaultClasses, props.classes), [
+        props.classes
+    ]);
```
jimbo: Don't bother memoizing this.

packages/venia-ui/lib/drivers/adapter.js
jimbo: I would normally have done the API call on submit of the "step", which is `Proceed to Checkout` or some variation of `Continue`. With the way the forms are broken out, though, it may be easier to wait until `Place Order`.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions/giftOptions.js
revanth0212: Sure. Any particular reason?

dpatil-magento:
@revanth0212 I can see gift options even after disabling this setting in Admin UI. Is this expected?

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions/giftOptions.js
jimbo: Just consistency. We don't memoize `mergeClasses()` in any of the existing components (afaik), and I have minor concerns about things slipping through the cracks when we get to refactoring `classify` one day.

sirugh:
> @revanth0212 I can see gift options even after disabling this setting in Admin UI. Is this expected?

Oh man I didn't even realize this was a thing. I think that we should create a fast-follow that goes back and enables/disables cart features based on their admin toggles. This will probably require either some build or runtime GQL check.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions/giftOptions.js
```diff
@@ -0,0 +1,164 @@
...
+        giftMessage,
+        toggleIncludeGiftReceiptFlag,
+        toggleIncludePrintedCardFlag,
+        updateGiftMessage
+    } = useGiftOptions({
+        getGiftOptionsQuery: GIFT_OPTIONS_QUERY,
+        saveGiftOptionsQuery: SET_GIFT_OPTIONS_QUERY
+    });
+
+    const classes = () => mergeClasses(defaultClasses, props.classes);
```
sirugh:
```suggestion
    const classes = mergeClasses(defaultClasses, props.classes);
```

Ah I didn't catch this originally. Omit the function creation. `classes` should be an object, which `mergeClasses` returns.

PR #2115, opened 2020-01-24
PWA-119: add product quantity stepper to cart

### LINKED ISSUE #1281: [feature]: Implement quantity dropdown as increment decrements with text-box on PDP

brendanfalkowski:
## Description

Adds a product quantity selector to each line item in /cart.

## Related Issue

Closes #1281 

### Verification Stakeholders

@soumya-ashok @sirugh 

### Verification Steps

1. Add item to cart.
2. Visit /cart
3. Play with the stepper and try to break it (it should be pretty resilient)

[Image from Gyazo](https://gyazo.com/db79480aa06de3fef882e374411af392)

## Checklist


brendanfalkowski:
Screenshots!

Screen Shot 2020-01-24 at 2 43 58 PM
Screen Shot 2020-01-24 at 2 44 10 PM

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.js
```diff
@@ -0,0 +1,159 @@
...
+                    aria-label={'Decrease ' + label}
+                    className={classes.button_decrement}
+                    disabled={isDecrementDisabled}
+                    onClick={decrement}
+                    type="button"
+                >
+                    <Icon className={classes.icon} src={MinusIcon} size={22} />
+                </button>
+
+                <input
```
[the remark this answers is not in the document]
jimbo: @sirugh Yes, fields should use `Form` from `informed` and `Field` from `venia-ui`, rather than storing the value in local state.

Ordinarily, I'd also expect our existing input components to be used, but in this case we're building a fairly unique one.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -32,17 +33,45 @@ export const useProduct = props => {
...
+                        itemId: item.id,
+                        quantity
+                    }
+                });
+
+                if (error) {
+                    throw error;
+                }
+            } catch (err) {
+                // TODO: Toast?
```
supernova-at: Toast?

😄

supernova-at:
Manually entering zero (0) in the text field and watching the entire page disappear is ... fun?

😆

supernova-at:
Probably out of scope here but we may want to add a `max` prop (set to current stock count, perhaps) to avoid

Screen Shot 2020-02-06 at 2 06 44 PM

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
sirugh: Error states are coming later. No UX direction so far on cart page error states :(

packages/peregrine/lib/talons/CartPage/ProductListing/useQuantity.js
```diff
@@ -0,0 +1,70 @@
...
+    ]);
+
+    const debouncedOnChange = useMemo(() => debounce(onChange, 100), [
+        onChange
+    ]);
+
+    const handleDecrement = useCallback(() => {
+        const newQuantity = quantity - 1;
+        quantityFieldApi.setValue(newQuantity);
+        debouncedOnChange(newQuantity);
```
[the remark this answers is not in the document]
sirugh: @jimbo was this actually breaking for you? This is all sync so spamming the button should still work and the debounced change should get the value from the last click anyways.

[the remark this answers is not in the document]
jimbo: Yes. Ideally, the buttons would not be disabled while the request is in flight—I'd be able to click the button a dozen times, and while it would be firing requests in the background every few seconds, none of them would prevent me from continuing to click the button.

I think Revanth's approach of throttling the update may work best, since it saves on both the leading edge and the trailing edge.

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.css
```diff
@@ -0,0 +1,69 @@
...
+    justify-items: center;
+}
+
+.wrap {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.label:not(:focus):not(:active) {
```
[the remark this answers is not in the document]
brendanfalkowski: We use this a lot to hide an element visually, but not hide it for a screen reader. It's not accessible to omit the element or hide it with `display: none`. In our repo, we compose an `srOnly` class to abstract this where possible (doesn't work across breakpoints), but I didn't want to introduce a stand-in concept for our old Sass mixins.

packages/peregrine/lib/talons/CartPage/ProductListing/useQuantity.js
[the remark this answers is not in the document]
sirugh: The issue was that my debounce delay was only 100ms. I made it 1000ms but we can discuss this with UX to fine tune.

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.css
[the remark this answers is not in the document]
jimbo: Ah, my mistake. As you said, the label contains `"Quantity"`. 👍

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.css
```diff
@@ -0,0 +1,69 @@
...
+    overflow: hidden;
+    width: 1px;
+    height: 1px;
+    clip: rect(0 0 0 0);
+    clip-path: inset(100%);
+    white-space: nowrap;
+}
+
+.button {
+    --size: 2.5rem;
```
jimbo: Leaving aside the mockups for a second, shouldn't this value match the input? (`2.25rem`) It looks off to me, as-is.

sirugh: Fair. I'll make that change in good faith. Will still need UX to sign off on this iteration.

dpatil-magento:
@sirugh What should be expected behaviour in following cases :
1. When I set qty to 0 and click outside > Product getting removed.
2. [Safari Browser] when I enter 1,000 as qty > both steppers are disabled.
3. [Safari Browser] also allows to enter alphabets and other characters. Chrome blocks all expect character "e"
4. When I enter 2.5, steppers increase(3.5)/decrease(1.5) in fractions and price summary does not get updated.
In 2nd, 3rd, 4th cases, or in generic just allow only numbers as inputs?   Also if user is able to enter something other than number then just revert to previous state?

sirugh:
> When I set qty to 0 and click outside > Product getting removed.

This is intentional. Updating a quantity to zero is the same as removing it.

> [Safari Browser] when I enter 1,000 as qty > both steppers are disabled.

Not intentional. Will investigate.

> [Safari Browser] also allows to enter alphabets and other characters. Chrome blocks all expect character "e"

Not intentional. Will investigate.

> When I enter 2.5, steppers increase(3.5)/decrease(1.5) in fractions and price summary does not get updated.

Not intentional. Will investigate.

> In 2nd, 3rd, 4th cases, or in generic just allow only numbers as inputs? Also if user is able to enter something other than number then just revert to previous state?

We have the ability to mask the input after a user enters something. I'll have to add a case for if a user somehow enters a non-integer.

dpatil-magento:
@sirugh Fixes on qty field looks good and also the updating shipping amount in shipping accordion wrt qty change.
But available Shipping methods does not stay in sync. After getting first available methods, sometimes shipping methods automatically get disappeared, sometime after selecting a shipping method remaining gets disappeared.

dpatil-magento:
@sirugh  When we have more shipping methods enabled like Fedex, USPS, Fixed, Free and user has say 2 products in cart >  Updating qty via stepper for both products > Price Summary and Shipping accordion does not get updated for 2nd request.

https://i.gyazo.com/0aabb0d32fbaa75f04b29630cde6a398.gif

sirugh:
Looks like luma requires the user click "update cart" after changing quantities...

dpatil-magento:
@sirugh Qty updates looks good but https://pr-2115.pwa-venia.com/cart  Kebab `Remove item` from cart is not working anymore.

dpatil-magento:
Firefox UI broken :( 

image

dpatil-magento:
Firefox issue will addressed as part of https://jira.corp.magento.com/browse/PWA-386.  This one is good to merge.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -28,17 +28,23 @@ export const useProduct = props => {
...
+            const { error } = await removeItem({
+                variables: {
+                    cartId,
+                    itemId: item.id
+                }
+            });
 
-        if (error) {
+            if (error) {
+                throw error;
```
supernova-at: This change seems unrelated - does decreasing the quantity to zero hit this callback?

Also, what does the UI do with this error, anything?

I was expecting to see L15:

```js
const [removeItem, { data, loading, error }] = useMutation(removeItemMutation);
```

and then the talon returns `data` and `error` back to the UI.

I can understand `await`ing the mutation so we can `setIsUpdating` but our handling here isn't consistent with the rest of the app.

This can also be out of scope of this ticket.

sirugh: Here I opted to mimic the `handleUpdateItemQuantity` callback. I assumed when we worked on handling errors in the UI we could revisit how this is implemented.

supernova-at: It's at least better than what we have today, where the UI can get "stuck" in the `isUpdating` state. I'm cool for revisiting.

PR #2119, opened 2020-01-27
Styleguide: App bar

jimbo:
## Description
Add an app bar section to the styleguide. Some of the prose content may be lorem ipsum, but the values should reflect the official Venia styles.

## Related Issue
PWA-155

## Acceptance 
- @jcalcaben should affirm document structure
- @soumya-ashok should affirm Venia typographic styles
- Others may review the code

### Verification Stakeholders
- @jcalcaben
- @soumya-ashok 

### Verification Steps
2. Navigate to `/page/app-bar`

## Checklist
* I have *not* added tests to cover my changes, if necessary.


soumya-ashok:
@jimbo Looks good overall, just a couple of things to consider:

On this - "The main app bar is affixed to the top of the viewport, so it stays fixed in position as the user scrolls." - I believe we had discussed a few months ago that it was better not to fix the app bar, rather let it scroll and if the user indicated that they were moving back up the page by scrolling in reverse it would re-appear. I might be mis-remembering but I think this would be the preferred behavior?

Not sure I understand this - "Within a modal, an app bar is affixed to the top of the modal." - The modal would be an overlay so wouldn't the app bar technically be behind the mask?

"The internal grid of an app bar is always as wide as the container to which it's affixed, even at higher resolutions." - We had discussed setting a max. width on the site for larger monitors. This would mean that there would be a max. width on the app bar as well, right? Do you think this needs to be called out here?

"At wider resolutions, its height increases to 5rem" - This is based on our discussion from today I presume?

jimbo:
> On this - "The main app bar is affixed to the top of the viewport, so it stays fixed in position as the user scrolls." - I believe we had discussed a few months ago that it was better not to fix the app bar, rather let it scroll and if the user indicated that they were moving back up the page by scrolling in reverse it would re-appear. I might be mis-remembering but I think this would be the preferred behavior?

We had discussed that, but declined that option. We want to avoid attaching a scroll listener to the page; affixing (or not affixing) can be done with CSS alone, no scroll listener required.

> Not sure I understand this - "Within a modal, an app bar is affixed to the top of the modal." - The modal would be an overlay so wouldn't the app bar technically be behind the mask?

Just a terminology thing. We happen to use the app bar for our modal (drawer) headers as well, which is what "within a modal" references.

> "The internal grid of an app bar is always as wide as the container to which it's affixed, even at higher resolutions." - We had discussed setting a max. width on the site for larger monitors. This would mean that there would be a max. width on the app bar as well, right? Do you think this needs to be called out here?

Yes, we've discussed setting a max width on the page container. That does not necessarily need to affect the main app bar, though; it could be exempt, as it is on some sites (and Venia today). Other sites do constrain the app bar's contents, though (Twitter being one example). If you'd like to constrain the app bar to a max width, we can certainly do that, and I think it would look better if we did. Should I update the copy?

> "At wider resolutions, its height increases to 5rem" - This is based on our discussion from today I presume?

Yes, I derived this from the new designs and tentatively included it here.

soumya-ashok:
> > On this - "The main app bar is affixed to the top of the viewport, so it stays fixed in position as the user scrolls." - I believe we had discussed a few months ago that it was better not to fix the app bar, rather let it scroll and if the user indicated that they were moving back up the page by scrolling in reverse it would re-appear. I might be mis-remembering but I think this would be the preferred behavior?
> 
> We had discussed that, but declined that option. We want to avoid attaching a scroll listener to the page; affixing (or not affixing) can be done with CSS alone, no scroll listener required.
> 
OK
> > Not sure I understand this - "Within a modal, an app bar is affixed to the top of the modal." - The modal would be an overlay so wouldn't the app bar technically be behind the mask?
> 
> Just a terminology thing. We happen to use the app bar for our modal (drawer) headers as well, which is what "within a modal" references.
> 
OK
> > "The internal grid of an app bar is always as wide as the container to which it's affixed, even at higher resolutions." - We had discussed setting a max. width on the site for larger monitors. This would mean that there would be a max. width on the app bar as well, right? Do you think this needs to be called out here?
> 
> Yes, we've discussed setting a max width on the page container. That does not necessarily need to affect the main app bar, though; it could be exempt, as it is on some sites (and Venia today). Other sites do constrain the app bar's contents, though (Twitter being one example). If you'd like to constrain the app bar to a max width, we can certainly do that, and I think it would look better if we did. Should I update the copy?
> 
Yes please

> > "At wider resolutions, its height increases to 5rem" - This is based on our discussion from today I presume?
> 
> Yes, I derived this from the new designs and tentatively included it here.
Sounds good.

dpatil-magento:
Deployment failed because of AWS env issue. Site is up at https://pr-2119.pwa-venia.com/. Ignore error.

soumya-ashok:
@jimbo thanks for making the change on the max-width text. This is good to go.

PR #2121, opened 2020-01-28
[SPIKE] [PWA-339] Conditionally bundle CE or EE features in Venia.

revanth0212:
## Description

Conditionally bundle CE or EE features in Venia.

This is a webpack plugin that at build time, swaps code depending on Magento backend edition. It uses `FILENAME.ce.js` in case of CE and `FILENAME.ee.js` in case of EE. This solution is scalable and maintainable even if we come up with different versions of the backend in the future.

As an example, I have added a couple examples of different components and different queries.

## Related Issue
Closes [PWA-339](https://jira.corp.magento.com/browse/PWA-339)

### Verification Stakeholders
@zetlen @jimbo 

### Specification
Docs need to be updated to reflect the new variable in .env `MAGENTO_BACKEND_EDITION`. @jcalcaben 

### Verification Steps
1. Run `yarn build`.
2. Inspect the build, EE code should not be present in the bundle, or in case of EE, the CE code should not be present.

image

image

## Checklist
* Docs need to be updated to reflect the new `MAGENTO_BACKEND_EDITION` environment variable.


packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummaryFragments.js
```diff
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag';
 
 import { DiscountSummaryFragment } from './discountSummary';
-import { GiftCardSummaryFragment } from './giftCardSummary';
+import { GiftCardSummaryFragment } from './queries/giftCardSummary.js';
```
revanth0212: `./queries/giftCardSummary.js` instead of `./queries/giftCardSummary` because we are only tapping on the resolution of js files in webpack.

packages/venia-ui/lib/components/CartPage/PriceSummary/queries/GiftCardSummary.ce.js
```diff
@@ -0,0 +1,7 @@
+import gql from 'graphql-tag';
+
+export const GiftCardSummaryFragment = gql`
+    fragment GiftCardSummaryFragment on Cart {
+        __typename
```
sirugh: Interesting -- I was hoping to eventually avoid having to maintain a "no-op" fragment. Is it possible for the plugin to replace it with an empty fragment instead?

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummaryFragments.js
sirugh: What does this mean?

packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.js
```diff
@@ -71,21 +70,4 @@ const GiftCardSummary = IS_EE
       }
     : () => null;
 
-export const GiftCardSummaryFragment = IS_EE
```
sirugh: The component itself also should no-op if non EE. Is that something we can do with the plugin too?

packages/pwa-buildpack/envVarDefinitions.json
```diff
@@ -11,6 +11,18 @@
                 }
             ]
         },
+        {
+            "name": "Magento Store Edition",
+            "variables": [
+                {
+                    "name": "MAGENTO_BACKEND_EDITION",
```
sirugh: What if we could do this automatically through schema validation instead of having to use a flag?

I have a PR that gets the validation to function -- and it does alert if a fragment/component has invalid gql.

packages/venia-ui/lib/components/CartPage/PriceSummary/queries/GiftCardSummary.ce.js
revanth0212: The plugin has nothing to do with Queries or GQL in general. It resolves files based on env variables. If Apollo allows empty strings you make this export an empty string.

packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.js
revanth0212: Yes, we can just have an empty export from its CE file and a valid component from its EE file.

packages/pwa-buildpack/envVarDefinitions.json
revanth0212: If the schema validation happens before bundling, we can definitely use that.

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummaryFragments.js
revanth0212: `./queries/giftCardSummary.js` does not exist. There is only `./queries/giftCardSummary.ce.js` or `./queries/giftCardSummary.ee.js`. But we still write `./queries/giftCardSummary.js` because the plugin is only looking for `js` file imports, once it comes across this import, it realizes that `ce` and `ee` versions of it are available and resolves to them based on the env variable.

revanth0212:
Tests are failing. This is expected. Working on fixing them.

fooman:
Really would love to see another spike that implements EE functionality as an extension to venia. Currently for the rest of Magento EE is essentially CE + extra modules. It would be great if PWA Studio could follow along the same path as this would immediately validate that PWA Studio extensibility works for real life uses.

sirugh:
>  implements EE functionality as an extension to venia.

@fooman that's _exactly_ what we've been discussing in person the last few days! This spike is just showing an intermediate state on the road to extensibility :)

revanth0212:
@fooman well said. As @sirugh said, this an intermediate step and what you mentioned about EE being CE + Extra Features should be possible.

fooman:
Great to hear this is the general direction. One thing you mentioned is scalability. I do believe having calls like isEE in the code base will not scale well. Even just adding Magento's other version B2B the combination options start to add up

`isB2B?'b2b.js':isEE?'ee.js':'ce.js'`

packages/pwa-buildpack/lib/WebpackTools/MagentoResolver.js
```diff
@@ -5,13 +5,22 @@ const validateConfig = optionsValidator('MagentoResolver', {
 module.exports = {
     validateConfig,
     async configure(options) {
-        validateConfig('.configure()', options);
+        const { isEE, ...restOptions } = options;
+        const extensions = [
+            '.wasm',
+            '.mjs',
+            isEE ? '.ee.js' : '.ce.js',
```
larsroettig: We should this also support for `.graphql`

packages/venia-ui/lib/components/CartPage/PriceSummary/GiftCardSummary.ee.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+};
+
+/**
+ * A component that renders the gift card summary line item.
+ *
+ * @param {Object} props.classes
+ * @param {Object} props.data fragment response data
+ */
+
```
larsroettig: Remove empty line

packages/venia-ui/lib/components/CartPage/PriceSummary/queries/GiftCardSummary.ce.js
```diff
@@ -0,0 +1,8 @@
+import gql from 'graphql-tag';
+
```
larsroettig: Use `.graphql` file extensions it is a query

packages/venia-ui/lib/components/CartPage/PriceSummary/queries/GiftCardSummary.ee.js
```diff
@@ -0,0 +1,13 @@
+import gql from 'graphql-tag';
```
larsroettig: Use `.graphql` file extensions it is a query

larsroettig:
> Great to hear this is the general direction. One thing you mentioned is scalability. I do believe having calls like isEE in the code base will not scale well. Even just adding Magento's other version B2B the combination options start to add up
> 
> `isB2B?'b2b.js':isEE?'ee.js':'ce.js'`

Very nice Idea we should add this also

revanth0212:
> I do believe having calls like isEE in the code base will not scale well. Even just adding Magento's other version B2B the combination options start to add up
> 
> `isB2B?'b2b.js':isEE?'ee.js':'ce.js'`

Well put @fooman 

We can do something like this to prevent clumsy if-else cases

```
const extensions = ['.wasm', `FILENAME.${MAGENTO_BACKEND_EDITION}.js`, '.js', '.jsx']
```

If that is not the case because, in the case of B2B, `MAGENTO_BACKEND_EDITION` might not be the same as your file extension, you could use a mapper.

We din't have other examples other than CE and EE, hence we went ahead with a boolean check for now. I hope this helps.

luke-denton-aligent:
I just watched the community update, and I have a question about the implementation of this. It seems that, for components that are EE only, we have to manage a CE file _aswell_, which is simply going to return `null`. Could we instead have _the component_ control when it is rendered, and use webpack's dead code elimination to strip out irrelevant code?
I ran a quick test with a `create-react-app` application, which I ejected but I don't think I needed to, as I had all 
So my idea would be a EnterpriseEdition higher order component (or render prop, or whatever is popular in React land at the moment) that wraps the parts of the component that should only be available in EE. This would allow us to have a whole component omitted from a bundle, or only parts of the bundle.
In practice, I found that the dead code elimination doesn't work as well for _parts_ of code, but works well with a HOC. Using `create-react-app`, I tested the following

```
// MyComponent.js
import React from 'react';
import EnterpriseEdition, { withEnterpriseEdition } from './EnterpriseEdition';

const MyComponent = () => {
    return (
    <div>This message is only for EE versions</div>
    );
};

export default withEnterpriseEdition(MyComponent);
```
and
```
// EnterpriseEdition.js
import React from 'react';
const isEnterprise = false; // This will need to be populated using an ENV variable

const EnterpriseEdition = (props) => {
    if (isEnterprise === false) {
        return null;
    }

    if (isEnterprise === true) {
        return props.children;
    }
}

export const withEnterpriseEdition = (WrappedComponent) => {
    return class extends React.Component { // Seems to be the only way to do a HOC?
        render() {
            if (isEnterprise === false) {
                return null;
            }

            if (isEnterprise === true) {
                return <WrappedComponent />
            }
        }
      }
};

export default EnterpriseEdition;
```
When using `withEnterpriseEdition`, the code inside `MyComponent.js` would only ever appear in the build folder if `isEnterprise` in `EnterpriseEdition.js` is true (to be completely transparent, it would still appear in the source maps, but not the JS bundles).
I was hoping that wrapping code in `MyComponent.js` with `<EnterpriseEdition></EnterpriseEdition>`, for example `<EnterpriseEdition><div>Some message for EE stores</div></EnterpriseEdition>`, would also eliminate that code from the resulting bundle, but it didn't seem to work as well as the HOC.. Perhaps I need some tweaking to get it working too.

I think managing the code this way would be far better than having to manage multiple files, as those 2 files could quickly become 3, especially if we're dealing with a component that has similar functionality that we want to abstract, and then include into both `.ce.js` and `.ee.js`

revanth0212:
Hey @luke-denton-aligent. Thanks for the write-up. Your approach makes complete sense if a feature if not available in CE but available in EE. Our goal with this PR was to not have the if-else cluster as part of the codebase and support different implementations for CE and EE features not just exclusive EE features.

We felt it is maintainable to have the code separated into different files instead of having if-else cases in 1 file. This way, we can generalize for all code types (JS, JSX, CSS, GQL, maybe even docs one day).

Here are a couple of reasons why we don't think HoCs work for this use-case:
1. HoCs are good design patterns for react components. They are not generally used for CSS modifications. Nor can they be used for GQL. 
2. They add a lot of wrapper code commonly known as the wrapper hell in the react community. Consider the bellow use case, it would add a new HoC every time we need a CE and EE separation because they can not be bundled into single HoC.
```
<ul>
  <CommonComponent />
  <CE_OR_EE_Component />
  <CommonComponent />
  <CE_OR_EE_Component />
</ul>
```
3. It will be hard to extend an HoC in the future if a new edition comes up (Maybe a Gold Edition). This would create a cluster of if-else inside the HoC. With the file separation structure, we just have to create a new file something like FILENAME.ge.js (Some wiring of filetypes in webpack and jest is needed though).

Like @sirugh mentioned we are working on improving this structure with the new extensibility framework that you should be hearing soon in future community talks.

Hope this helps.

luke-denton-aligent:
Hi @revanth0212,

Thanks for the response!

I'll have to disagree with the feeling that multiple files is more maintainable - as it means there has to be special magic in the webpack config in order for it to work, and like I said, a single feature could be across 3 files. HoC's certainly aren't the greatest thing in the world, and in most cases have been superseded by other approaches, so I was only using them as a quick example.

As for the if/else cluster, hasn't it just been moved to the webpack config now anyway?

I'll await the final implementation of the CE/EE approach, hopefully the chosen path doesn't come back to bite us in the butt :)

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummaryFragments.js
```diff
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag';
 
 import { DiscountSummaryFragment } from './discountSummary';
-import { GiftCardSummaryFragment } from './giftCardSummary';
+import { GiftCardSummaryFragment } from './queries/giftCardSummary';
```
dpatil-magento: @revanth0212 If I change import to `./queries/GiftCardSummary.ee` AWS builds pass. Tested in https://github.com/magento/pwa-studio/pull/2138.  Do we need to define any env variable?

revanth0212:
> We can do something like this to prevent clumsy if-else cases
> 
> ```
> const extensions = ['.wasm', `.${MAGENTO_BACKEND_EDITION}.js`, '.js', '.jsx']
> ```
> 
> If that is not the case because, in the case of B2B, `MAGENTO_BACKEND_EDITION` might not be the same as your file extension, you could use a mapper.
> 
> We din't have other examples other than CE and EE, hence we went ahead with a boolean check for now.
> 
> Once we have more matured use cases of EE and B2B, we can change this to support all the permutations and combinations.

@luke-denton-aligent we don't have B2B or other examples as of now but if we did, this is how we want to handle it in the future without an `if-else` cluster.

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummaryFragments.js
revanth0212: Found the issue @dpatil-magento 

We are referencing `queries/giftCardSummary` but the file name is `queries/GiftCardSummary`. Mac does not throw an error if the file name case is wrong, but Linux machines do. Git did not recognize the file name case change and did not push the changes. Hence it is working in our locals but fails in AWS.

I will be pushing the changes soon.

luke-denton-aligent:
> > We can do something like this to prevent clumsy if-else cases
> > ```
> > const extensions = ['.wasm', `.${MAGENTO_BACKEND_EDITION}.js`, '.js', '.jsx']
> > ```
> > 
> > 
> > If that is not the case because, in the case of B2B, `MAGENTO_BACKEND_EDITION` might not be the same as your file extension, you could use a mapper.
> > We din't have other examples other than CE and EE, hence we went ahead with a boolean check for now.
> > Once we have more matured use cases of EE and B2B, we can change this to support all the permutations and combinations.
> 
> @luke-denton-aligent we don't have B2B or other examples as of now but if we did, this is how we want to handle it in the future without an `if-else` cluster.

This would mean for a feature that is _only_ B2B, for example, we have to have 3 files; `component.b2b.js`, `component.ee.js` and `component.ce.js`, with `component.ee.js` and `component.ce.js` both just returning `null` right? I don't understand how this is a nice solution..

Also, this approach breaks tab completion in IDE's right? I have to know _how_ Magento works in order to know that I omit the `ee`/`ce` from the filename when importing a component, which an IDE isn't going to know

Just trying to make sure we're making this as easy on developers as possible

revanth0212:
> @revanth0212 This would mean for a feature that is _only_ B2B, for example, we have to have 3 files; `component.b2b.js`, `component.ee.js` and `component.ce.js`, with `component.ee.js` and `component.ce.js` both just returning `null` right? I don't understand how this is a nice solution..

@luke-denton-aligent check out this doc to get a clear understanding of how `resolve.extensions` works in webpack. If you have a B2B only feature, you can have `FILENAME.b2b.js` that has the feature code and `FILENAME.js` which can return `null`. You don't need to have all files. That is how webpack resolve works.

> Also, this approach breaks tab completion in IDE's right? I have to know _how_ Magento works in order to know that I omit the `ee`/`ce` from the filename when importing a component, which an IDE isn't going to know

The whole point of separating them into different files is for developers to know by looking at the filename what edition this feature belongs to. 

>Just trying to make sure we're making this as easy on developers as possible.

Every new feature comes with a learning curve and no solution is a swiss knife to solve everything. There are compromises that we had to make as a team for the greater good of the product and the developers (core team and community). The advantages of separating features into their own files made life easier by not having to look inside each file every time. Also, this separation happens at build time and reduces the bundle size eventually (not today, we don't have enough code to remove to make a significant point).

Which this pattern we are not mandating the usage of it, you can still use the `if-else` structure.

PR #2123, opened 2020-01-28
[PWA-239] Shipping Method (Cart)

tjwiebell:
## Description

When a user (shopper) is in the cart flow for Venia, a shipping method selection/estimate component should display in order to allow a shopper to determine the impact of shipping to their order summary. 

 
Requirements: 

Adhere to design specs/flows
Initial design accounts for unauthenticated cart/checkout, authenticated to be a fast follow
out of scope: 

detecting location based on customer device

## Related Issue

* [[PWA-239](https://jira.corp.magento.com/browse/PWA-239)] Shipping Method (Cart)

### Verification Steps

If you want to verify that third-party shipping providers are working, you'll need inventory items that have weight, and your shipping methods configured in the backend. I've done this with a staging environment, you'll have to use Carmina Necklace: https://mcstaging-jnz3dtiuj77ca.dummycachetest.com/
1. Add product to your cart
2. Navigate to `/cart` page and verify that estimate shipping fields are displayed (country, state, zip)
3. Enter values for these fields and shipping methods should be displayed based on available methods
4. Select a method and verify that PriceSummary is updated to reflect this selection

## Checklist

* I have updated the documentation accordingly, if necessary.
* I have added tests to cover my changes, if necessary.

PR #2124, opened 2020-01-29
Gift Cards

supernova-at:
## Description

This PR adds Gift Card support to the new Cart page.

Known issues out of scope:
* The `PriceAdjustments` accordion resets back to the original set of sections that are open (`isOpen={true}`) after mutations occur.
* Applying an invalid card spits an error to the console

## Related Issue

Closes JIRA-78.

### Verification Stakeholders

@schensley

### Verification Steps

Since Gift Cards is an enterprise edition feature, first set your `MAGENTO_BACKEND_EDITION` to `EE`. 

1. `yarn watch:all`
1. Go to the `/cart` page
1. Check the balance of a valid card: `02Y0LCHUNDRZ `.
1. Attempt to check the balance of an invalid card: `no_work`.
1. Apply a valid card: `02Y0LCHUNDRZ `.
1. Attempt to apply an invalid card: `no_work`.
1. Remove a card.
1. Simulate a remove card failure by applying a card, turning off your network, and clicking "remove". You should see an error message Toast appear.

Set your `MAGENTO_BACKEND_EDITION` back to `CE` (`CE` is also the default), and observe that the Cart page doesn't even show a Gift Card section in the Accordion.

Screen Shot 2020-01-29 at 11 19 54 AM

## Checklist


packages/peregrine/lib/talons/CartPage/GiftCards/useApplyButton.js
```diff
@@ -0,0 +1,18 @@
+import { useCallback } from 'react';
+import { useFormState } from 'informed';
+
+export const useApplyButton = props => {
+    const { handleApplyCard } = props;
+
+    const giftCardEntryFormState = useFormState();
+
+    const handleApplyCardWithCode = useCallback(() => {
+        const giftCardCode = giftCardEntryFormState.values['card'];
```
supernova-at: The `ApplyButton` and `CheckBalanceButton` are child components / separated out so that they can `useFormState` to get this card value from the `Form`.

This is different from Coupons, which could use the `Form`'s `onSubmit` handler directly. Since gift cards has multiple action buttons (apply and check balance), it needed to do it this way.

I suppose we could make one of them (Apply, I'd say) a submit button and only do this for `CheckBalanceButton`.

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCard.js
```diff
@@ -0,0 +1,13 @@
+import { useCallback } from 'react';
+
+export const useGiftCard = props => {
+    const { code, handleRemoveCard } = props;
+
+    const handleRemoveCardWithCode = useCallback(() => {
+        handleRemoveCard(code);
+    }, [code, handleRemoveCard]);
```
supernova-at: Am I overcomplicating things here? The `handleRemoveCard` callback function defined in the `useGiftCards` talon needs to be passed a gift card code.

Is this the easiest way to do that?

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
```diff
@@ -0,0 +1,167 @@
...
+            cartResult.data && cartResult.data.cart.applied_gift_cards,
+        handleApplyCard,
+        handleCheckCardBalance,
+        handleRemoveCard,
+        handleTogglePromptState,
+        isLoadingGiftCards: cartResult.loading,
+        isApplyingCard: applyCardResult.loading,
+        isCheckingBalance: balanceResult.loading,
+        isRemovingCard: removeCardResult.loading,
+        shouldDisplayCardBalance: Boolean(balanceResult.data),
```
supernova-at: The logic for determining this value will likely change.

packages/venia-ui/lib/components/CartPage/GiftCards/giftCard.js
```diff
@@ -0,0 +1,39 @@
...
+
+import { mergeClasses } from '../../../classify';
+import defaultClasses from './giftCard.css';
+
+const GiftCard = props => {
+    const {
+        // appliedBalance,
+        code,
+        // currentBalance,
+        // expirationDate,
```
supernova-at: I can remove these here and from the query, they are available but aren't currently used.

packages/venia-ui/lib/components/CartPage/PriceSummary/index.js
```diff
@@ -1 +1 @@
-export { default, PriceSummaryQuery } from './priceSummary';
+export { default, GET_PRICE_SUMMARY } from './priceSummary';
```
supernova-at: This change has nothing to do with Gift Cards but the browser was complaining.

packages/venia-ui/lib/components/CartPage/GiftCards/giftCards.css
```diff
@@ -0,0 +1,89 @@
...
+.card_input_container {
+    align-items: center;
+    display: grid;
+    grid-template-columns: 1fr min-content min-content;
+    gap: 1em;
+}
+
+.card_input {
+    composes: root from '../../Field/fieldIcons.css';
+    width: 100%;
```
sirugh: https://github.com/magento/pwa-studio/pull/2123/files#diff-a77804cf0b1634bdd0e60b2ae8576eaeR7

sirugh: I think the right way to approach this is to just apply the width to the `FieldIcon` component. We should be in the practice of "fixing" them when they don't quite meet our needs anymore. Reach for customization only after concluding that it's not something we can apply generally. In this particular case, it is :)

packages/peregrine/lib/talons/CartPage/GiftCards/useApplyButton.js
jimbo: I discussed this with @sirugh last week, but I think multiple submit is the way to go here. From the docs:

> **`submitForm`**
> 
> This function will submit the form and trigger some lifecycle events.
> 1. It will set all the fields to touched.
> 2. It will call all validators.
> 3. It will call `onSubmit` if the form is valid. This function can be called manually however it is also called if you have a `<button type='submit'>` within the `<Form>.`

As you can see, form submission involves more than just calling an endpoint with some values. And we certainly don't want to be in the business of replicating these effects ourselves. So we have two options here:

- Let the buttons trigger a `submit` event on the form as normal, or
- Get a handle on the form and call the `submitForm()` method

The calls should be done inside of a `handleSubmit` callback passed to the `Form` element via `onSubmit`. The first argument to that handler is `values`, so it should be very easy to get the value and make the call. You should also have access to event (and from it, the button element that triggered the event), which you can use to determine which call to make. Let me know if you have difficulty with this part.

The official attribute for distinguishing between multiple submit buttons is [`formaction`](https://css-tricks.com/separate-form-submit-buttons-go-different-urls/), I believe.

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
```diff
@@ -0,0 +1,180 @@
...
+
+    // We need the cartId for all of our queries and mutations.
+    const [{ cartId }] = useCartContext();
+
+    /*
+     * Apollo hooks.
+     *
+     * Immediately execute the cart query and set up the other graphql actions.
+     */
+    const cartResult = useQuery(cartQuery, {
```
sirugh: I'd make this a lazy query and nest this in an effect since there's a chance the cart won't have been created yet.

packages/venia-ui/lib/components/CartPage/GiftCards/giftCards.css
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2124/commits/e70ba23133ae0d7f6339b39488105bebe3df9e6a.

packages/venia-ui/lib/components/CartPage/GiftCards/giftCard.js
```diff
@@ -0,0 +1,39 @@
...
+        handleRemoveCard,
```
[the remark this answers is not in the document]
supernova-at: Awesome, thanks!

Resolved in https://github.com/magento/pwa-studio/pull/2124/commits/cf6febc4a3f26c4163680bc947c2ba987e4b6ff8.

packages/peregrine/lib/talons/CartPage/GiftCards/useApplyButton.js
supernova-at: > You should also have access to event (and from it, the button element that triggered the event), which you can use to determine which call to make. Let me know if you have difficulty with this part.

When I tried it previously I definitely was not getting this value but I'll take another crack at it.

supernova-at: Ok, the callback passed to `Form` `onSubmit` only gets the values of the `Form`, not the `event`: https://github.com/joepuzzo/informed/blob/d797b3d5a51f2125039cd4c315429097b41aa14a/src/hooks/useForm.js#L51

But I was able to get this working: https://github.com/magento/pwa-studio/pull/2124/commits/41da0823a2f82141c2ff27f8dd485c62ea596d89.

Basically since I was already keeping track of the "last action taken" (apply, check balance, remove, toggle), I was able to piggyback on that and combine it with the `values` passed to `onSubmit` to make the appropriate call.

It muddles the talon a bit, but it lets me remove child components and keep everything centralized.

packages/venia-ui/lib/components/CartPage/GiftCards/giftCards.js
```diff
@@ -0,0 +1,178 @@
...
+        <div className={classes.card}>
+            <Field
+                classes={{ root: classes.entry }}
+                id={classes.card}
+                label="Gift Card Number"
+            >
+                <div className={classes.card_input_container}>
+                    <TextInput
+                        id={classes.card}
+                        classes={{ icons: classes.card_input }}
```
sirugh: I think this is leftover, right?

packages/venia-ui/lib/components/TextInput/textInput.css
```diff
@@ -1,3 +1,6 @@
+.icons {
```
sirugh: I think this is leftover, right?

packages/venia-ui/lib/components/CartPage/GiftCards/giftCards.js
supernova-at: Yes, thanks! 

Resolved in https://github.com/magento/pwa-studio/pull/2124/commits/67a1061ae9e205d3f676c3b035c5cd3048dcc8e0.

packages/venia-ui/lib/components/TextInput/textInput.css
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/2124/commits/67a1061ae9e205d3f676c3b035c5cd3048dcc8e0.

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
supernova-at: This is the last piece of feedback left but I'm not sure I understand it fully.

Currently, if the cart hasn't been created yet `cartResult.error` will be populated and the UI shows a "There was an error loading gift cards. Please try again." message.

If we make this a lazy query and put it in a `useEffect`, what does that buy us? Some extra time   before we check for the existence of a cart?

supernova-at: From reviewing the shipping methods PR, I see

```js
useEffect(() => {
        if (cartId) {
            fetchShippingMethods({
                variables: {
                    cartId
                }
            });
        }
    }, [cartId, fetchShippingMethods]);
```

And I think that answers my question: we can save ourselves a call to the backend if we already know that we don't have a `cartId`.

packages/venia-ui/lib/components/CartPage/GiftCards/giftCards.css
```diff
@@ -0,0 +1,109 @@
...
+        'card_balance card_balance card_balance'
+        'check_balance_button check_balance_button check_balance_button';
+}
+
+.card_input_container {
+}
+
+.cards_container {
+    align-items: baseline;
+    display: grid;
```
jimbo: Consider adding a `gap` here so the items don't have to accomplish it with `margin`.

supernova-at: Nice, thanks!

Resolved in https://github.com/magento/pwa-studio/pull/2124/commits/9173c3b25090bdec84cfd7c0d0e4b6205beacf95.

dpatil-magento:
@supernova-at https://pr-2124.pwa-venia.com/cart running on EE but dont see Gift Cards accordion. Also when run on local with CE, dont see any of the accordion.

supernova-at:
> @supernova-at https://pr-2124.pwa-venia.com/cart running on EE but dont see Gift Cards accordion. Also when run on local with CE, dont see any of the accordion.

Resolved in-person. New `MAGENTO_BACKEND_EDITION` environment variable is needed.

dpatil-magento:
@supernova-at EE looks good but CE does not show any of the accordion (Shipping, Coupon, Gift options).  Does this work for you?

supernova-at:
> @supernova-at EE looks good but CE does not show any of the accordion (Shipping, Coupon, Gift options). Does this work for you?

It does:

Screen Shot 2020-02-07 at 8 09 38 AM

Do you have items in your cart? That's the only thing I can think of that hides the entire accordion.

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
supernova-at: This didn't get done because if we don't have a `cartId` the `GiftCard` component won't render because we hide the entire price adjustments `Accordion`.

sirugh: We talked offline, just to comment here: I think its safer to just do the lazyQuery check since eventually we may hide this with css and not by rendering null.

supernova-at: Ok, nevermind - resolved in https://github.com/magento/pwa-studio/pull/2124/commits/bbc73b8a23f2894dc9eefa743724e089f95aaca3.

supernova-at:
Fixed the issue!

Check out commit https://github.com/magento/pwa-studio/pull/2124/commits/24b7fb5d56e1bea1190fac5d356502e0c4c559aa.

Verify:
1. View the `graphql` network call to `getCartDetails`
1. In CE, verify the `GiftCardFragment` and `GiftCardSummaryFragment` in the `query`:

```
fragment GiftCardFragment on Cart {
  id
  __typename
}
```

```
fragment GiftCardSummaryFragment on Cart {
  id
  __typename
}
```

3. In EE, verify the `GiftCardFragment` and `GiftCardSummaryFragment` in the `query`:

```
fragment GiftCardFragment on Cart {
  applied_gift_cards {
    code
    __typename
  }
  id
  __typename
}
```

```
fragment GiftCardSummaryFragment on Cart {
  id
  applied_gift_cards {
    applied_balance {
      value
      currency
      __typename
    }
    __typename
  }
  __typename
}
```

Thanks @dpatil-magento @sirugh @tjwiebell !

PR #2129, opened 2020-01-30
[bug] Fix w10 issues

sirugh:
## Description
Informed had a [bug](https://github.com/joepuzzo/informed/issues/215), introduced in version 2, that broke edge on w10.

This PR updates our version of informed to the latest minor. Informed has released a major but I didn't want to introduce that much risk in this PR. That said, there are some semi risky changes that involved changing how we interact with the `formApi`. At some point between our informed versions the author changed the root component to be a function component which does not have an `instance`. So our references in tests to `form.instance` breaks. This also removes the ability to pass a `ref` directly. To remedy this I replaced these usages of instance methods with usage of the `getApi` callback.

We also had a problem with spreading a nodelist which is invalid syntax in edge. Converting it to `Array.from` fixes the issue of an infinite loop of rendering which was occurring due to the problematic code living in the wrapping component of `App`.

## Related Issue

Closes [PWA-266](https://jira.corp.magento.com/browse/PWA-266).

### Verification Steps

1. Open a PDP on W10/Edge.
2. Change the quantity - the UI should reflect the selection.

## Checklist


packages/venia-ui/lib/components/Head/headProvider.js
```diff
@@ -10,7 +10,7 @@ const removeExistingTitleTags = titleTags => {
 const VeniaHeadProvider = props => {
     useEffect(() => {
         const titleTagsToRemove = document.getElementsByTagName('title');
-        removeExistingTitleTags([...titleTagsToRemove]);
+        removeExistingTitleTags(Array.from(titleTagsToRemove));
```
sirugh: This fixes an issue where we would infinitely error in Edge.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/CouponCode/__tests__/__snapshots__/couponCode.spec.js.snap
```diff
@@ -18,6 +18,7 @@ exports[`disables remove button on click 1`] = `
 exports[`disables submit button on coupon entry 1`] = `
 <form
   className="entryForm"
+  onKeyDown={[Function]}
```
sirugh: These are additions by `informed`.

packages/venia-ui/lib/components/CreateAccount/__tests__/createAccount.spec.js
```diff
@@ -67,34 +65,3 @@ test('attaches the submit handler', () => {
 
     expect(typeof onSubmit).toBe('function');
 });
-
-test('calls onSubmit if validation passes', async () => {
```
sirugh: This function was testing `informed` functionality. Instead of adding a new formApiRef to this component I just deleted this test.

packages/venia-ui/lib/components/MiniCart/__tests__/__snapshots__/cartOptions.spec.js.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`it renders the correct tree 1`] = `
-<t>
+<Q>
```
sirugh: I genuinely don't know why this changed but I imagine it has something to do with informed.

packages/venia-ui/lib/components/SignIn/__tests__/signIn.spec.js
```diff
@@ -165,30 +165,3 @@ test('changes view to ForgotPassword', () => {
     expect(setDefaultUsername).toHaveBeenCalledTimes(1);
     expect(showForgotPassword).toHaveBeenCalledTimes(1);
 });
-
-test("avoids reading the form if it doesn't exist", () => {
```
sirugh: This also seems like something we don't need to actually test. I removed it, but I could be convinced to add it back if someone feels strongly about it (and has a way to actually make it work).

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -36,7 +36,8 @@ export const useSignIn = props => {
         errors.push(getDetailsError);
     }
 
-    const formRef = useRef(null);
+    const formApiRef = useRef(null);
+    const setFormApi = useCallback(api => (formApiRef.current = api), []);
```
tjwiebell: We should not memoize callbacks that do not change; I believe this is actually more costly than `api => (formApiRef.current = api)`.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/CouponCode/__tests__/__snapshots__/couponCode.spec.js.snap
tjwiebell: I bet this is related to the new `preventEnter` prop added in `2.4.0`. The previous default behavior allowed enter which is consistent with the default now, so this seems safe.

packages/venia-ui/lib/components/Head/headProvider.js
tjwiebell: We'll have to chat about this; we surely use the spread operator more than just this.

sirugh: Right, spread is allowed, just not spread on a `NodeList` :)

packages/peregrine/lib/talons/SignIn/useSignIn.js
sirugh: Oops! Just a habit now, though I really doubt it's _that_ costly :P

PR #2132, opened 2020-01-31
Set fixed `yarn` version `yarn@berry`

larsroettig:
## Description
-  if you already update to `npm install -g yarn@berry` you need 
`yarn policies set-version` for get the correct `yarn`version 

## Acceptance 
-  `pwa studio` works with a installed yarn@berry


zetlen:
Hey @larsroettig ! I appreciate that we need to make sure this repository uses Yarn v1, but I don't think it's a good idea to include an entire copy of Yarn in our source code.

Our practice is to use NVM to switch between Node (and npm and yarn) versions. It looks like Yarn supports a command which sets a version per repository; is there a reason to also include a Yarn version?

https://yarnpkg.com/getting-started/install

larsroettig:
@zetlen I close we don't need should be working if you use yarn berry correct

PR #2133, opened 2020-02-01
Add new categorySort Element #360

### LINKED ISSUE #360: [feature]: Category Sorting with Venia (PWA)

[X] `venia-concept`
[ ] `pwa-buildpack`
[ ] `peregrine`
[ ] `pwa-module`
[ ] `pwa-devdocs`

## Description

https://magento.invisionapp.com/share/WQN5F7BYBPG#/324119989_Product_Sort

### JIRA issue for design
https://jira.corp.magento.com/browse/UXD-181

Screen: Category with Filter and Sort
image

Screen: Sort By with Category
image

As a shopper, I want to filter and sort products within a category so that I can make a purchase on Venia.

- Price: Low to High
- Price: High to Low
- Featured
- Newest
- Best Selling
- Ratings: High to Low (non-MVP)

#### Assumptions
- Dependent on Category page and ability to retrieve products by category
- Ratings not part of MVP (concept only)
- Sorting will be done SSR
- Dependency on collection filtering variable support; follow-up with graphql-ce
 
## Tasks
- [ ] Discovery for GraphQL coverage w/r/t searching and filtering queries (input types)
- [ ] Discovery for how sorting affects and shown with URL (linking+routing)
- [ ] Implement Sort By panel with basic options (not Ratings)
- [ ] Testing and documentation (https://github.com/magento-research/pwa-studio/wiki/Documentation-Guidelines)

larsroettig:
## Description
Implements sorting for Categorie Page  

## Related Issue
Closes #360 

## Acceptance 
- Price: Low to High
- Price: High to Low
- Featured
- Best Selling
- Ratings: High to Low (non-MVP)

### Verification Steps

1. Go to the Categorie sorting works as Expected 

sort

## Checklist

- [x]  I have updated the documentation accordingly, if necessary.


larsroettig:
@soumya-ashok @cherdman we should have an extra sort button `venia` has also the availability to
sort all attributes in both directions. If we don't do it than we lost functionality.

schensley:
Hi @larsroettig I'm with the UX team for PWA.  I agree with you that sorting should have provide for incremental or decremental sorting when sorting by price, as for sorting alphabetically (A-Z)  I am skeptical of how often B2C shoppers sort products in this manner, but that is Luma functionality for sorting.  I'm not sure what the screen cap above is or where it's from, but it is non-standard to say the least. A new design exists. Based on your feedback, I'd like to run it through another internal review to ensure we are accounting for parity in terms of functionality.

jimbo:
@larsroettig Thanks for your contribution. 🙂 

As Scott mentioned, we take Venia's UX pretty seriously, so we scrutinize any PR that proposes new interactions or patterns more heavily than one that simply constructs a new screen out of existing interactions and patterns.

I understand you're hoping to get this PR merged quickly, but we'd like to take our time evaluating it, and we may request revisions. Please be patient with us—it could take some time. 😅

brendanfalkowski:
For what it's worth, in B2C and B2B projects I've generally reduced the available sorting to:

Best Match (aka the merchant or algorithm picked)
Price: High to Low
Price: Low to High

This is usually combined with additional filters like "only in stock" or "only in stock at store X", but before the user can set those our search provider does boost/bury products so for example low or zero stock at a location is pushed to the bottom of the list.

We haven't found it helpful to allow sorting control over inventory or name (alpha).

soumya-ashok:
@larsroettig We looked at a number of sites before designing the new sorting functionality for Venia. It appears that the alphabetical sort is used rarely in the B2C context. Have you come across differently in your experience?

The other sort parameter in question for me is "position" - It seems that this is merely a notation in the Magento product catalog on the admin side based on conversations with quite a few merchants and a few partners and does not indicate any clear sorting preference on the shopper side. I see that you do have that in your implementation as Luma currently has this and I'd like to understand if you see value in including this.

larsroettig:
Hi will change to design luma design from 2.3.4 at this weekend is equals to design from adobe UX team sorry I was not aware of it

packages/venia-ui/lib/RootComponents/Category/category.js
```diff
@@ -18,6 +18,16 @@ const Category = props => {
     const { currentPage, totalPages } = paginationValues;
     const { setCurrentPage, setTotalPages } = paginationApi;
 
+    const [sort, setSort] = useState({
+        sortAttribute: 'relevance',
+        sortDirection: 'ASC'
+    });
```
jimbo: Generally, we avoid storing objects in state unless they're stored outside of the component as constants.

```jsx
const Foo = props => {
  const [bar, setBar] = useState({ a: 1, b: 2 })

  // causes an infinite loop
  useEffect(() => {
    // what if the next state is the "same" as the current one?
    const nextBar = { a: 1, b: 2 }

    // always true, since they're different objects
    if (nextBar !== bar) {
      setBar(nextBar)
    }
  }, [bar, setBar])

  return <div>{`${bar.a} ${bar.b}`}<div>
}
```

We want to be able to know when the state has changed and when it hasn't, so it's better to use a boolean, number, or string if possible. Coincidentally, if we have a collection of objects, we generally need to derive a key for each object, and that key should also be a boolean, number, or string. So deriving a proper key for each object can also give you a useful value to store in state.

packages/venia-ui/lib/components/CategorySort/categorySort.css
```diff
@@ -0,0 +1,68 @@
...
+    margin: 0.125rem 0 0;
+    font-size: 1rem;
+    color: black;
+    text-align: left;
+    list-style: none;
+    background-color: #fff;
+    background-clip: padding-box;
+    border: 1px solid rgb(var(--venia-grey-dark));
+    border-radius: 0.25rem;
+    box-shadow: 0 1px 3px #3f3b4a26;
```
jimbo: Clever use of extended hex, but let's stick with `rgba()`. Also, we do try to avoid arbitrary colors in our CSS. 😅 

Have a look at [`autocomplete.css`](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/SearchBar/autocomplete.css) for an example of the shadow for a dropdown menu.

packages/venia-ui/lib/components/CategorySort/categorySort.css
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.child {
```
jimbo: Consider choosing a more semantic or descriptive name for this classname. If this is a menu item, `menuItem` will work.

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,125 @@
...
+            setSortDirection(sortAttribute.sortDirection);
+        };
+
+        if (visible === false) {
+            return;
+        }
+
+        const sortElements = [];
+        for (const element of availableSortMethods) {
+            const key = uuidV4();
```
jimbo: This may circumvent the error, but it defeats the purpose of keys, especially if not memoized. As I suggested in an earlier comment, you should derive a string key from each of these objects, which can be used for the `key` prop as well as for storing the current sort value in state.

I suggest concatenating `attribute` and `sortDirection`, which will give you a unique string that's representative of the object. Since you'll want this key for multiple purposes, it may serve you better to transform the array into a map right away, as we often do in this codebase.

```es6
const availableSortMethods = new Map()

for (method of props.availableSortMethods) {
  // derive a unique key for each method
  const key = `${attribute}--${sortDirection}`

  // add each method to the map
  availableSortMethods.set(key, method)
}
```

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,125 @@
...
+            sortElements.push(
+                <li key={key} className={classes.child}>
+                    <button
+                        className={classes.childButton}
+                        onClick={onChildButtonClick.bind(null, element)}
```
jimbo: If you search the codebase for `.bind(`, you won't find any results. 😅 

In years past, binding handlers was fairly common practice in React, but it's no longer recommended. I recommend creating a component specifically for these buttons, and letting that component define its own `handleClick` handler (and memoize it with `useCallback`!). There should be a number of examples in the codebase you can use for reference.

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,125 @@
...
+                    >
+                        <span className={classes.childText}>
+                            {element.text}
+                        </span>
+                        {element.attribute === sortAttribute &&
```
jimbo: Generally, we avoid multiline inline JSX expressions, since they defeat the readability advantages of JSX. Consider moving them out of inline expressions and into constants.

```jsx
const conditionalFoo = a && b !== "bar" && c.items
  ? <span>hello</span>
  : null

return (
  <div>
    {conditionalFoo}
  </div>
)
```

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,125 @@
...
+import defaultClasses from './categorySort.css';
+import { mergeClasses } from '../../classify';
+import Icon from '../Icon/icon';
+
+const CategorySort = props => {
+    const classes = mergeClasses(defaultClasses);
+    const { availableSortMethods, sortControl } = props;
+    const { currentSort, setSort } = sortControl;
+
+    const [visible, setVisible] = useState(false);
```
jimbo: We have a utility just for this! It's a hook called `useDropdown`(https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/hooks/useDropdown.js), and it will handle closing the dropdown whenever a click happens outside of the menu. I recommend using it to avoid duplicating that code.

packages/venia-ui/lib/RootComponents/Category/category.js
larsroettig: Main Problem we should have default `sortState`  on every category page and i dont know solve this better.

larsroettig: We need `attribute` and  `direction` to build a valid graph query 

```
 sortAttribute: 'relevance',
 sortDirection: 'ASC'
```

jimbo: > We need attribute and direction to build a valid graph query

Yes, right. I was merely suggesting that we store the state differently. Honestly though, it's not a big deal. This is probably fine.

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,112 @@
...
+
+import { mergeClasses } from '../../classify';
+import CategorySortItem from './categorySortItem';
+import defaultClasses from './categorySort.css';
+
+const CategorySort = props => {
+    const classes = mergeClasses(defaultClasses);
+    const { availableSortMethods, sortControl } = props;
+    const { currentSort, setSort } = sortControl;
+    const { elementRef, expanded, setExpanded } = useDropdown();
```
jimbo: Nice work incorporating all of the feedback I had in this file. 😄

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,112 @@
...
+            setSort({
+                sortAttribute: sortAttribute.attribute,
+                sortDirection: sortAttribute.sortDirection
+            });
+            setExpanded(false);
+        };
+
+        // should be not render item in collapsed mode.
+        if (expanded === false) {
+            return;
```
jimbo: One minor thing here: we tend to return `null` in these cases (rather than `undefined`).
```suggestion
            return null;
```

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,112 @@
...
+        const onMenuItemClick = sortAttribute => {
+            setSort({
+                sortAttribute: sortAttribute.attribute,
+                sortDirection: sortAttribute.sortDirection
+            });
+            setExpanded(false);
+        };
+
+        // should be not render item in collapsed mode.
+        if (expanded === false) {
```
jimbo:
```suggestion
        if (!expanded) {
```

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,112 @@
...
+            return;
+        }
+
+        const sortElements = [];
+        for (const element of availableSortMethods) {
+            const attribute = element.attribute;
+            const sortDirection = element.sortDirection;
```
jimbo: These last couple of lines can be destructured pretty easily.

```es6
const { attribute, sortDirection } = element
```

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,112 @@
...
+
+            const key = `${attribute}--${sortDirection}`;
+            sortElements.push(
+                <li key={key} className={classes.menuItem}>
+                    <CategorySortItem
+                        sortItem={element}
+                        active={isActive}
+                        onClick={function(element) {
+                            onMenuItemClick(element);
+                        }}
```
jimbo: When you pass an inline function to `onClick`, it will be recreated on every render. That's especially not optimal since it has to be rebound in the DOM.

The general rule here is to never create a "version" of the handler in the parent and pass it down to the child. Instead, pass the necessary data (and any static handlers) to the child and let it compose its own handler.

```jsx
// parent
const List = props => {
  // memoize to avoid recreating this function on render
  // make sure its deps aren't going to change
  const handleItemClick = useCallback(() => {
    setExpanded(false)
  }, [setExpanded])

  // iterate over data to create elements
  // note that we don't define any handlers in the loop
  const itemElements = Array.from(data, item => {
    return (
      <Item item={item} onClick={handleItemClick} />
    )
  })
}

// child
const Item = props => {
  const { item } = props
  const { foo } = item

  // define a handler at the item level, not the list level
  const handleClick = useCallback(() => {
    // do something with the data we received as a prop
    console.log(foo)

    // if we accept a callback, call it here
    // note that received callbacks are `onEvent`
    // while defined callbacks are `handleEvent`
    onClick()
  }, [foo, onClick])

  return <button onClick={handleClick} />
}
```

packages/venia-ui/lib/components/CategorySort/categorySortItem.js
```diff
@@ -0,0 +1,42 @@
...
+    const handleClick = useCallback(() => {
+        onClick(sortItem);
+    }, [sortItem, onClick]);
+
+    const activeIcon =
+        active === true ? (
+            <span className={classes.activeIcon}>
+                <Icon src={Check} size={14} />
+            </span>
+        ) : null;
```
jimbo: Perfect. 💯

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -0,0 +1,120 @@
...
+    const handleItemClick = useCallback(
+        sortAttribute => {
+            setSort({
+                sortAttribute: sortAttribute.attribute,
+                sortDirection: sortAttribute.sortDirection
+            });
+            setExpanded(false);
+        },
+        [setSort, setExpanded]
+    );
```
jimbo: Yes, this is the right way to do it. 👍

packages/venia-ui/lib/components/CategorySort/categorySortItem.js
```diff
@@ -0,0 +1,42 @@
...
+
+import defaultClasses from './categorySortItem.css';
+
+const CategorySortItem = props => {
+    const { onClick, sortItem, active } = props;
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    const handleClick = useCallback(() => {
+        onClick(sortItem);
+    }, [sortItem, onClick]);
```
jimbo: Yes, much better. As long as `sortItem` is stable (which it appears to be here), this will be properly memoized: created once and stable across renders.

PR #2137, opened 2020-02-03
[PWA-334] Implement PageBuilder as an extension

zetlen:
## Description

PageBuilder has moved into its own package `@magento/pagebuilder`. It now integrates into Venia using a **brand new plugin pipeline!**

### Dependencies

The `@magento/venia-ui` package does **not** depend on `@magento/pagebuilder`. All direct references to `@magento/pagebuilder` in that package have been removed. Instead, `@magento/venia-ui` exposes a new [extension point](#Current_Extension_Points) and PageBuilder integrates into it.

The new `@magento/pagebuilder` package depends on `@magento/venia-ui` as a **peerDependency**. It contains a few direct imports of `@magento/venia-ui` components in its code. **We want PageBuilder to be independent of Venia in the future, but we need to refactor its code to be more independent first.**

The `@magento/venia-concept` package depends on `@magento/pagebuilder` and `@magento/venia-ui`. **Venia "installs" `@magento/pagebuilder` just by putting it in its dependencies list!**

This satisfies the use case where a developer should be able to say `npm install @magento/pagebuilder` and it will automatically integrate into their UI on the next build. Magento!

### Extension Model

The extension model works entirely at build time, making it superior to other extension systems in performance and bundle size. It's powered by a new concept called the **BuildBus**. The **BuildBus** is an object much like an event bus, but it uses the interceptor patterns from [Tapable](https://github.com/webpack/tapable) to enable modules to declare, and share, extension points. This is good because it's the same as the popular, successful system in the build tool we already use: [Webpack compiler and compilation hooks](https://webpack.js.org/api/compiler-hooks/). _(In our project we will call these objects "targets", to avoid confusion with the substantially different concept of React hooks.)_

Modules which are extensions can integrate with BuildBus by putting a custom property `pwa-studio` in their `package.json` files:

```
{
  "name": "@magento/venia-ui",
  [...other properties]
  "pwa-studio": {
    "targets": {
      "declare": "./targets/declare",
      "intercept": "./targets/intercept"
    }
  }
```

The above configuration tells BuildBus that this package both **declares** its own targets and **intercepts** other targets. Specifically:

- BuildBus will first run the file at `@magento/venia-ui/targets/declare`. There, Venia UI exposes its own targets for other extensions to modify it.
- BuildBus will then run the file at `@magento/venia-ui/targets/intercept`. There, Venia UI injects itself into other modules' targets, and invokes its own targets to enable other modules to interact with its build.

BuildBus does the following:

- runs all modules in dependency order (e.g. `@magento/pagebuilder` runs after `@magento/venia-ui`
- runs all "declare" files in order, _then_ all "intercept" files in order, so that extensions are guaranteed access to all other targets, even dependents' targets
- expects each "declare" and "intercept" file to export a default function, which it will call with a special API.

#### BuildBusStop API

- Declare files and intercept files have access to a "BuildBusStop" facade for the BuildBus, so that BuildBus can track their requests and declarations. The facade has two methods:

- `api.declareTarget(targetName, target)`: Declare a named target that other modules can use. `targetName` must be a string, which will be the public name of this extension point. `target` must be an instance of a `Tapable` class.
- `api.getTarget(moduleName, targetName)`: Retrieve a target from your own declared targets or another module. Calling `api.getTarget()` with one argument will try to retrieve the current module's named target. Calling `api.getTarget()` with two arguments will try to retrieve the targets from a different module.

Once the targets are retrieved, they can be tapped in many ways; see the [Tapable docs](https://github.com/webpack/tapable#readme) for the complete API.

#### Declaring and Intercepting Targets

Modules which declare targets can choose how to invoke those targets. This should usually happen in the `intercept` phase: for a module's target to do anything, it must in turn intercept an upstream target in one of its dependencies, and in its own handler it can invoke its named target with any arguments it chooses.

### Current Extension Points

_(The list will get a lot longer; this PR only implements enough extension points to do a PageBuilder integration.)_

#### BuildBus
BuildBus has some builtin targets; these targets hook directly in to the build process, and all extension targets are ultimately powered by interceptors on these base targets.

- `specialFeatures`: Runs before the compilation begins. Interceptors will receive the `special` configuration object that the root project argued to `configureWebpack()`. They can add to that object, indicating that the current package contains `esModules`, `cssModules`, `rootComponents`, et cetera.
- `webpackCompiler`: Runs when a Webpack compiler is created. Interceptors will receive the [compiler object](https://webpack.js.org/api/compiler-hooks/), which has a `hook` property containing Tapable objects with the exact same API as BuildBus targets. Interceptors of `webpackCompiler` will usually tap into the compiler's hooks, which allows them to interact with the build.

#### Venia UI
Venia UI declares a `richContentRenderers` target. Interceptors to this target receive an API object which allows them to add additional content renderers to the `RichContent` component. Each content renderer must be a module that exposes a `canRender` function, which receives CMS content data and should return true if the renderer can display it, and a `Component` React component, which will be used to render. These renderers are added to a list in dependency order, where the first one which can render will be used. It builds that list manually by editing source code at build time, injecting direct references to all rich content renderers.

## Related Issue

Closes [PWA-334]..

### Verification Stakeholders
@davemacaulay 
@jimbo 
@sirugh 
@supernova-at 
@revanth0212 
@jcalcaben 
@tjwiebell 
@awilcoxa 

### Specification
PageBuilder functionality must be reviewed.
Fallback HTML render must be reviewed.
Build times and bundle sizes must be analyzed and compared.

### Verification Steps

1. Set your Venia instance `MAGENTO_BACKEND_URL` to https://dave-mhvaqay-vzsrtettsztvg.us-4.magentosite.cloud/ to access an instance with PageBuilder data on the home page.
1. Manually remove `@magento/pagebuilder` from `packages/venia-concept/package.json`.
1. Run `yarn watch:venia`.
1. Visit the homepage.
1. Verify that content exists, but it's not attractively styled as hero images by PageBuilder. Scroll around and look for a parallax effect; it should not be there.
1. Add `@magento/pagebuilder` back to the dependencies list in `packages/venia-concept/package.json`.
1. Run `yarn watch:venia`.
1. Visit the homepage.
1. Verify PageBuilder content renders beautifully..
1. Run a full production build and verify staging as well.

## Checklist


brendanfalkowski:
Our team started testing PageBuilder in PWA last week, and I'm collating some questions we had after seeing these changes.

PageBuilder components seemed to be re-redefined in three places, and now four potentially.

1. code for the Magento admin
2. code for the PHTML frontend (not a concern for us)
3. code for the PWA frontend (in /components/RichContent)

Now (3) is split into: 
- code for the PWA frontend (from @magento/pagebuilder)
- code for the PWA frontend (that customizes @magento/pagebuilder or is new)

We **DO NOT** have a significant amount of content to migrate, and could ignore PageBuilder (or CMS entirely by deploying content as code) as a shortcut. What we don't want to do is sink time into modifying PageBuilder components in PWA or PHP then see a consolidation of these redefinitions.

As of this PR, we're not sure if we should freeze our work customizing in RichContent component because it's not clear how we'll extend this code once it's a dependency. Any advice?

packages/venia-ui/targets/intercept.js
```diff
@@ -0,0 +1,46 @@
...
+
+const loaderIdent = __filename;
+
+const loaderIsInstalled = mod => mod.loaders.some(l => l.ident === loaderIdent);
+
+const name = '@magento/venia-ui';
+
+module.exports = api => {
+    const richContentRenderers = api.getTarget('richContentRenderers');
+    api.getTarget('@magento/pwa-buildpack', 'webpackCompiler').tap(
```
fooman: I imagine for other parts of the code that declare their own targets in the future this implementation of handling the target will look very similar and won't change that much. Any chance this can be generalised so it can be more easily reused?

jimbo:
> So it looks like two paradigms are implied "application components" + "CMS components" and they don't share. Is this an intentional split for separate concerns? Again we don't really want to customize heavily if there's a long-term consolidation planned

Astute observation.

Venia and PageBuilder components are separate because they're two existing, independent systems that work differently. They're not separate concerns—both are responsible for rendering interactive, responsive UI—but they differ on a few core implementation principles.

- PageBuilder uses a blend of global styles and inline styles to achieve the final presentation, and many of those inline styles are stored as data and passed down the tree as component props.
- Venia uses scoped styles and custom properties to achieve the final presentation, so the component code often knows little about the appearance of the end result.

It would be hard to consolidate these components without reconciling these differences and others. For example, PageBuilder's browser support matrix includes Internet Explorer versions that Venia's does not. This doesn't mean consolidating our components is impossible, just that it would amount to a refactor of PageBuilder's rendering code (major version, regression, etc.), which is not something we're undertaking at the moment.

That said, there's opportunity for it to happen. We'll be expanding Venia's usage of custom properties soon, so that style values (e.g., colors) will properly derive from design tokens. We're doing that to offer an easier way to "theme" Venia, but it'll also do a better job of exposing our design architecture to PageBuilder and other extensions—and they may opt to make use of those properties.

jonaohana:
Hi All, 

Wondering if anyone can point me in the right direction ,  is there a means of validating a URL with the cms to return the page ID of a page created within Pagebuilder ?

packages/pwa-buildpack/lib/WebpackTools/plugins/BuildBusPlugin.js
```diff
@@ -0,0 +1,13 @@
+class BuildBusPlugin {
+    constructor(bus) {
+        this.bus = bus;
```
jimbo: No default value necessary?

packages/venia-ui/lib/components/RichContent/richContent.js
```diff
@@ -21,20 +18,26 @@ const toHTML = str => ({ __html: str });
...
 const RichContent = props => {
-    const { html } = props;
     const classes = mergeClasses(defaultClasses, props.classes);
-
-    if (detectPageBuilder(html)) {
-        return (
-            <div className={classes.root}>
-                <PageBuilder masterFormat={html} />
-            </div>
+    const newProps = {
```
jimbo: Consider avoiding `new` in vars. 👍 

```suggestion
    const nextProps = {
```

packages/pagebuilder/package.json
```diff
@@ -0,0 +1,67 @@
...
+  },
+  "repository": "github:magento-research/pwa-studio",
+  "author": "Magento Commerce",
+  "license": "(OSL-3.0 OR AFL-3.0)",
+  "bugs": {
+    "url": "https://github.com/magento-research/pwa-studio/issues"
+  },
+  "homepage": "https://github.com/magento-research/pwa-studio/tree/master/packages/pagebuilder#readme",
+  "dependencies": {},
+  "devDependencies": {
```
jimbo: This seems inaccurate to me. I would expect most of these to be actual dependencies, not dev dependencies. No?

Related, are you using NPM instead of Yarn?

packages/pagebuilder/intercept.js
```diff
@@ -0,0 +1,17 @@
...
+            };
+        }
+    );
+    api.getTarget('@magento/venia-ui', 'richContentRenderers').tap(
+        '@magento/pagebuilder',
+        richContentRenderers => {
+            richContentRenderers.add('PageBuilder', '@magento/pagebuilder');
+        }
+    );
+};
```
jimbo: This is really concise.

packages/pagebuilder/lib/ContentTypes/Products/products.css
```diff
@@ -26,7 +26,7 @@
 }
 
 .galleryItems {
-    composes: items from '../../../../Gallery/gallery.css';
+    composes: items from '~@magento/venia-ui/lib/components/Gallery/gallery.css';
```
tjwiebell: Accidental tilde?

jimbo: > Accidental tilde?

Not accidental. Required for absolute `node_modules` refs since [`css-loader@2.0.0`](https://github.com/webpack-contrib/css-loader/tree/v2.0.0#url).

packages/venia-ui/targets/intercept.js
zetlen: There's more than a chance; there's a plan! We plan to leave these code snippets in place for a while, as we implement more and more functionality as extensions. As patterns appear, we will abstract them into a library of extension utility functions. This way, we can grow towards a more terse, declarative model, while still supporting lower-level customizations.

packages/venia-ui/lib/components/RichContent/__tests__/richContent.spec.js
```diff
@@ -2,33 +2,12 @@ import React from 'react';
...
-    };
-});
-
 test('renders a RichContent component', () => {
     const component = createTestInstance(<RichContent />);
 
     expect(component.toJSON()).toMatchSnapshot();
 });
 
-test('renders a RichContent component with Page Builder content', () => {
```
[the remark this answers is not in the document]
zetlen: Yeah

packages/venia-concept/package.json
```diff
@@ -12,7 +12,7 @@
   "scripts": {
     "build": "yarn run build:prod",
     "build:analyze": "yarn run clean && mkdir dist && webpack -p --profile --no-progress --env.mode production --json > dist/build-stats.json && webpack-bundle-analyzer dist/build-stats.json",
-    "build:dev": "yarn run clean && yarn run validate-queries && webpack --progress --env.mode development",
+    "build:dev": "yarn run clean && webpack --no-progress --env.mode development",
```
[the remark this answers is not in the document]
zetlen: I have a reason, but it's not relevant to this PR, so I'm reverting.

packages/venia-concept/package.json
```diff
@@ -88,6 +89,9 @@
     "graphql-tag": "~2.10.1",
     "html-webpack-plugin": "~3.2.0",
     "informed": "~2.1.13",
+    "jarallax": "~1.11.1",
+    "load-google-maps-api": "~2.0.1",
+    "lodash.escape": "~4.0.1",
```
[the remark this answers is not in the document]
zetlen: I'm inclined to declare as many dependencies as possible as `peerDependencies`, which is what they are in `@magento/pagebuilder`. That's a safer model, to me, for building an efficient app. It enforces that only one copy of `jarallax` and other such libraries can be in the same bundle at one time.

When a dev installs `@magento/pagebuilder`, they'll see warnings that they also need to add these packages as dependencies. So then they can, and they'll have control over those package versions. If another package wants an incompatible version of `jarallax`, it'll fail the build, rather than silently putting two copies of the same library into the bundle.

packages/pwa-buildpack/lib/WebpackTools/plugins/BuildBusPlugin.js
zetlen: Nah, nobody's creating these but us.

packages/pagebuilder/package.json
zetlen: In the core repo I'm using Yarn, but this system works with either.

packages/venia-ui/lib/components/RichContent/richContent.js
zetlen: Thanks. Renamed to be more specific.

packages/pagebuilder/package.json
zetlen: Also, all of the dependencies that PageBuilder needs at runtime are listed as `peerDependencies`, not literal dependencies. See [my other comment](https://github.com/magento/pwa-studio/pull/2137#discussion_r375536983) for an explanation of that.

>It enforces that only one copy of jarallax and other such libraries can be in the same bundle at one time.

The NPM/Yarn model is built totally around the idea of tolerating mutually incompatible versions of a library. That's great for Node, and browser apps without strict performance budgets, but we need to be more careful--and peer deps are the NPM-supported way of doing that.

packages/venia-concept/package.json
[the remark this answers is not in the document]
jimbo: Ah, I see why. If these are both peer dependenies *and* dev dependencies in `@magento/pagebuilder`, that does explain why they would be listed as dependencies of `@magento/venia-concept`. Wouldn't they be real dependencies here, though, rather than dev dependencies?

supernova-at:
> BuildBus has some builtin targets
> specialFeatures: Runs before the compilation begins...
> webpackCompiler: Runs when a Webpack compiler is created....

What would you think of naming these things for when they run, like event handlers? We could export targets for use like `onCompilationStart`, `onCompilationEnd`, `onCompilerCreated`, etc.

packages/pwa-buildpack/lib/BuildBus/BuildBus.js
```diff
@@ -0,0 +1,123 @@
+const makeDebug = require('debug');
+const Tapable = require('tapable');
+const pertain = require('pertain');
+
+const VALID_HOOKS = new Set(
+    Object.values(Tapable).filter(v => typeof v === 'function')
+);
```
supernova-at: Do we want to only support a subset of hooks?

But I suppose `BuildBus` supporting all of the hooks is different from saying Venia will use / declare all of the hooks.
This would allow plugins to support any Tapable hook they wanted, and other plugins to tap into them.

packages/pwa-buildpack/lib/BuildBus/declare-base.js
```diff
@@ -0,0 +1,6 @@
+const { SyncHook } = require('tapable');
+
+module.exports = api => {
+    api.declareTarget('webpackCompiler', new SyncHook(['compiler']));
+    api.declareTarget('specialFeatures', new SyncHook(['featuresByModule']));
```
supernova-at: > BuildBus has some builtin targets
> specialFeatures: Runs before the compilation begins...
> webpackCompiler: Runs when a Webpack compiler is created....

What would you think of naming these things for when they run, like event handlers? We could export targets for use like `onCompilationStart`, `onCompilationEnd`, `onCompilerCreated`, etc.

packages/pagebuilder/lib/ContentTypes/Block/__tests__/__snapshots__/block.spec.js.snap
```diff
@@ -23,11 +23,6 @@ exports[`renders a Block component 1`] = `
 >
   <div
     className="root"
-    dangerouslySetInnerHTML={
```
davemacaulay: Seems weird for this snapshot to be modified for moving the logic to another directory.

packages/pagebuilder/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@magento/pagebuilder",
```
davemacaulay: We might want to think of a naming convention for packages like this, while I love the simplicity of @magento/pagebuilder, I think it could cause some confusion on it's intent, even with the description.

packages/pwa-buildpack/lib/BuildBus/BuildBus.js
```diff
@@ -0,0 +1,123 @@
...
+
+class BuildBusStop {
```
davemacaulay: I'd love some inline documentation to understand the logic of BuildBus

packages/pwa-buildpack/lib/BuildBus/declare-base.js
davemacaulay: @supernova-at I love this suggestion! Let's make our API super obvious of when things run :)

packages/venia-concept/package.json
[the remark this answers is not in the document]
davemacaulay: @zetlen if the application did not have `@magento/pagebuilder` installed, would the bundle still include Jarallax etc? Or would these only be included if the bundler discovered their usage in the application?

packages/venia-ui/lib/components/RichContent/plainHtmlRenderer.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+const toHTML = str => ({ __html: str });
+
+function PlainHtmlRenderer({ html, classes }) {
```
davemacaulay: 🔥 love it!

packages/venia-ui/lib/components/RichContent/richContent.js
```diff
@@ -21,20 +18,26 @@ const toHTML = str => ({ __html: str });
...
-    if (detectPageBuilder(html)) {
-        return (
-            <div className={classes.root}>
-                <PageBuilder masterFormat={html} />
-            </div>
+    const rendererProps = {
+        ...props,
+        classes
+    };
+    for (const Renderer of richContentRenderers) {
```
davemacaulay: this is beautiful 😍

packages/venia-concept/package.json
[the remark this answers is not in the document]
zetlen: @davemacaulay That is an interesting question. In this peerDeps driven workflow, here's what the workflow would be:

1. Without PageBuilder installed, none of those other deps are installed either
1. `yarn add @magento/pagebuilder`
1. Install process shows warnings about installing `jarallax` and other dependencies
1. Build fails if those dependencies are not installed
1. Developer must install those deps as well

This happens with Babel and other common tools, but I would understand if it feels too complex for a new NPM/yarn user. There may be other good ways to prevent multiple copies of a library in the same bundle; do you have an alternative suggestion?

packages/pwa-buildpack/lib/BuildBus/BuildBus.js
zetlen: @supernova-at This is an early draft API, and in my new suggestion (I'll post it in another comment) it makes a little more sense. I do think we should enforce that you have to use Hooks so that the API is consistent; Tapable's `HookMap` seems to have little utility to me, so I wanted to disallow it.

packages/pwa-buildpack/lib/BuildBus/declare-base.js
zetlen: I'd be more than happy to do that for our own targets. Webpack's internal hooks do a more `beforeThing` / `afterThing` model; what do you think of that?

packages/pagebuilder/package.json
zetlen: I know, but I was really into the simplicity of this.

Other plugin systems, like `eslint` and `babel`, rely on verbose package naming conventions to auto-detect the plugins. That annoys me and it forces a given package to plug in to one and only one thing. I chose the `pwa-studio` property in `package.json` so that package names could be more freely chosen.

Nevertheless, we have our own namespacing to think about. Suggestions?

packages/pwa-buildpack/lib/BuildBus/BuildBus.js
zetlen: You'll get it! And this feature needs a lot more documentation before it's releasable. This PR is small and incremental, and I'm planning on adding better docs in a future release.

zetlen:
Hey reviewers and stakeholders!

@brendanfalkowski 
@fooman 
@jimbo 
@tjwiebell 
@supernova-at 

Thanks so much for taking the time to review this PR. I've made requested changes and answered a few questions, and I have one more thing to ask of you.

I've reviewed this to myself a few times, and I think that new developers would benefit from a revision to the API. I've worked on it over the last few days, and I'm going to ask how you all feel about it before I decide to push the change.

## Alternative BuildBus API suggestion

The `api` object that extension authors use to declare and intercept extensions doesn't feel very intuitive to me. Here are some of the changes I've made in my experimental branch.


### Declaring

#### Old
```js
const { SyncHook } = require('tapable');

module.exports = api => {
    api.declareTarget('webpackCompiler', new SyncHook(['compiler']));
    api.declareTarget('specialFeatures', new SyncHook(['featuresByModule']));
};
```

#### New
```js
module.exports = targets => {
    targets.declare({
        webpackCompiler: new targets.types.Sync(['compiler']),
        specialFeatures: new targets.types.Sync(['special'])
    });
};
```

#### Notes

- The convention should be to call the API `targets`, not the generic `api`.
- The only things you can declare are targets, so `declareTarget` becomes`declare`.
- You can only declare one target per call to `declareTarget`, but many extensions will declare multiple targets. `declare` takes a dictionary, supporting single and multiple target declarations in a readable format.
- You have to use a specific version of Tapable, so you shouldn't have to import it yourself. The new version provides the set of available target types as part of the API.

---

### Intercepting

#### Old
```js
module.exports = api => {
    const richContentRenderers = api.getTarget('richContentRenderers');
    richContentRenderers.tap('@magento/venia-ui', api =>
        api.add({
            componentName: 'PlainHtmlRenderer',
            importPath: './plainHtmlRenderer'
        })
    );
    api.getTarget('@magento/pwa-buildpack', 'webpackCompiler').tap(
        '@magento/venia-ui',
        compiler => // etc.
    );
};
```

#### New
```js
module.exports = targets => {
    const { richContentRenderers } = targets.own;
    richContentRenderers.tap(renderers =>
        renderers.add({
            componentName: 'PlainHtmlRenderer',
            importPath: './plainHtmlRenderer'
        })
    );

    targets.of('@magento/pwa-buildpack').webpackCompiler.tap(
        compiler => // etc.
    );
};
```

#### Notes

- The `getTarget` method for getting other extensions' targets is not ideal.
  - The purpose of the `targets` API is to work with targets, so `getTarget` is verbose.
  - The one-argument and two-argument overloads for getting own and other targets are confusing.
  - Extensions will use both their own targets and other extensions' targets in their intercept code, _but_ they will use each category of target differently.
  - They will tap/intercept the targets of other extensions, whereas they will `call()` their own targets. Different purposes should mean different accessors.
  - In the new API, an extension's own targets are available at `targets.own.targetName`, and other extensions' targets are available at `targets.of('extension-package-name')`.
- Retrieving targets themselves is awkward and verbose.
  - In the current API, you pass a string name of a target to `getTarget` to retrieve the target object itself. This differs from Webpack's pattern, where hooks are simple object properties.
  - In the new API, you retrieve your own targets or others' targets as an object dictionary, not one at a time.
  - The plain object is much more "discoverable" during debugging; when paused on a breakpoint, it's easy to see what targets are available through inspecting local objects.
- Calling the tap/intercept methods is wordy and repetitive.
  - Tapable hooks require two arguments to tap/intercept methods: a name for the plugin, e.g. `"HtmlWebpackPlugin"`, and then the interceptor handler itself. The name is required for debugging and error tracking.
  - For our extension system, these names should be consistent and informative; it makes sense to use the package names here. However, it is very verbose to pass the same string `"@magento/venia-ui"` multiple times.
  - In the new API, the name argument is optional. If the intercept code provides no name argument, the handler will be named after the package name itself automatically.
  - If intercept code _does_ provide a named argument, it will be appended to the package name, to do simple namespacing. In `venia-ui/targets/intercept.js`, for instance:
    ```js
     const buildpackTargets = targets.of('@magento/pwa-buildpack');
     buildpackTargets
       .specialFeatures.tap(handler);
     buildpackTargets
       .webpackCompiler.tap("VeniaWebpackStuff", handler2);
    ```
    The tap on `specialFeatures` will be named `@magento/venia-ui` in logging. The tap on `webpackCompiler` will be named `@magento/venia-ui:VeniaWebpackStuff`.

---

This API feels better to me, and it's certainly more succinct and readable. In this experimental branch, I've also improved the logging and tracing of extensions and how they tap into each other. The BuildBus feeds those details into the Webpack stats object, which will enable future debugging with sophisticated tools, and even helpful visualizations of the current extension setup.

I'm interested in your opinions on this revision before we merge, but if I hear from many of you that you'd rather it be merged in its current state and this change come in a followup PR, I would understand that. Still, I think we have a good chance now to make it better, before we start using it in many other places.

jimbo:
> This API feels better to me, and it's certainly more succinct and readable. In this experimental branch, I've also improved the logging and tracing of extensions and how they tap into each other. The BuildBus feeds those details into the Webpack stats object, which will enable future debugging with sophisticated tools, and even helpful visualizations of the current extension setup.
> 
> I'm interested in your opinions on this revision before we merge, but if I hear from many of you that you'd rather it be merged in its current state and this change come in a followup PR, I would understand that. Still, I think we have a good chance now to make it better, before we start using it in many other places.

I agree. I think the new API is more predictable, gets to the point more quickly, and should be easier to use in practice. The extra tracing should also help during extension development.

I'm in favor of making the change in this PR; I'd rather not merge a short-lived API just to rush this into `develop`. But I'll wait for others to chime in.

davemacaulay:
@zetlen I love the new API, the method calls and reasoning behind the changes make a lot of sense, I'm happy to move ahead with making those alterations if everyone else is.

packages/pagebuilder/package.json
davemacaulay: Yeah I love the simplicity too, if people get confused we'll deal with that later.

I dislike naming conventions but sometimes they're important, so I have no good suggestions.

packages/pwa-buildpack/lib/BuildBus/declare-base.js
davemacaulay: @zetlen I'm happy with either, as long as it's reasonably clear from the event name when the target will be executed I think we're cool. if webpack is doing before / after maybe that makes the most sense as this is a webpack-y type thing-y

packages/venia-concept/package.json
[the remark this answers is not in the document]
davemacaulay: @zetlen I guess I'm used to the world of composer where it'll automatically resolve and install all dependencies on install, but you're right in saying if they required different versions things would get a little messy.

I don't have an alternative and if this is the way to go it makes enough sense to me. I'd be inclined to say we shouldn't include the deps for Page Builder in `venia-concept`, as any other package not authored by us won't have this luxury, but as we're going to include `@magento/pagebuilder` by default for now we probably have too.

If you have any suggestions of how we can avoid this that'd be great, but not necessary for this iteration.

packages/pwa-buildpack/lib/BuildBus/declare-base.js
zetlen: `upon` it is, then!

fooman:
`targets.of('@magento/venia-ui').richContentRenderers` does look a lot easier to grasp. If I am not mistaken the use of `targets.own` will ensure that different packages can't interfere with each others extensions points as easily. This would be a win as well.

zetlen:
@fooman Yes, that's the idea. There exist "backdoors" for packages to inspect each others' extension points, but this API has a few safeguards for that:

1) The declare and intercept phases run sequentially, and the packages run in dependency order
2) The `targets` API passed to each package's declare and intercept files is actually a facade object. It's bound to that particular package and it tracks any interactions between that package and others.

With this architecture we can add rules in the future pretty easily, such as preventing any package from running `call` on another package's interceptor.

dpatil-magento:
@zetlen Ran through verification steps looks good.  Stage server looking good.
PWA - https://pr-2137.pwa-venia.com/
PB-PWA - https://pr-2168.pwa-venia.com/

So pending issue is about env variables, ex - GOOGLE_MAPS_API_KEY. Also one more minor thing `package.json` has reference to `magento-research/pwa-studio` instead of `magento/pwa-studio`.

@dhaecker Would you like to take a quick look?

zetlen:
To support extensible environment variables, I need to add overloads to some public methods. My changeset has gotten large; we can either merge a lot of complex code or make this another major release, so I can introduce breaking changes to Buildpack APIs. @dpatil-magento we can talk about this internally.

dpatil-magento:
Created bug https://jira.corp.magento.com/browse/PWA-379 for env variable issue. 

Its good to merge once @dhaecker approves.

dhaecker:
@dpatil-magento have you tested anything page builder related in this PR (what have you tested)? Thank you

dpatil-magento:
@dhaecker I had deployed these changes to https://pr-2168.pwa-venia.com/ and compared home page with https://pb.pwa-venia.com/ on Chrome.

dhaecker:
@dpatil-magento After speaking with Dave M. about the changes, no functionality is expected to change in Page Builder content so your testing is good enough. I approve ✅

PR #2150, opened 2020-02-07
[bugfix] Hide stale suggestions

jimbo:
## Description
Fix a bug with `Suggestions` wherein stale results would continue to be displayed when the field had been cleared. The proper fix would entail resetting the query, such that `called` would be `false` and `data` would be `undefined`, but Apollo does not yet offer that functionality. Instead, this fix recognizes when the field has been cleared and incorporates that into the conditional rendering logic of `Autocomplete` and `Suggestions`.

This fix involves changes to the talon APIs, so it's probably a major. As a result, I've also taken the opportunity to have `SearchBar` obtain `history` and `location` from hook calls rather than render props.

## Related Issue
[PWA-361](https://jira.corp.magento.com/browse/PWA-361)

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Enter `athena` in the search input.
2. Verify that suggested categories and products appear.
3. Click the button to clear the search input.
4. Verify that suggestions are no longer displayed, but a prompt is.

## Checklist
* I have not yet added tests to cover my changes, if necessary.


packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -33,21 +34,26 @@ export const useAutocomplete = props => {
     // run the query once on mount, and again whenever state changes
     useEffect(() => {
         if (visible && valid) {
+            setCleared(false);
```
tjwiebell: This will actually still display the stale results if the clear and next keystroke is done within 500ms. I'm not exactly sure why, but I suspect that while the query is in flight, it might be clearing out the contents of data, but haven't dug into it.

Very minor problem, but if you wanted to fix, you could use the `onCompleted` callback of `useLazyQuery`, or even easier just move the state update to `debouncedRunQuery`.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -33,21 +34,26 @@ export const useAutocomplete = props => {
...
             debouncedRunQuery({ variables: { inputText: value } });
+        } else if (called && !value && !visible) {
+            console.log('value was cleared');
```
tjwiebell: Remove debug code

packages/peregrine/lib/talons/SearchBar/useSearchBar.js
```diff
@@ -1,13 +1,15 @@
...
 
 const initialValues = { search_query: '' };
 
-export const useSearchBar = props => {
-    const { history } = props;
-    const { push } = history;
+export const useSearchBar = () => {
     const { elementRef, expanded, setExpanded } = useDropdown();
+    const history = useHistory();
+    const location = useLocation();
```
tjwiebell: Move this to `useSearchField`, then you don't even need to drill it down.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
jimbo: > This will actually still display the stale results if the clear and next keystroke is done within 500ms.

I can't reproduce this. Are you encountering a situation in which the input is empty *and* the dropdown is open *and* stale results are displayed?

Stale results are always going to be present in the dropdown—that's the intent, given that I have no way to actually reset the query state—but this PR attempts to ensure that the dropdown is hidden (and thus, stale results are hidden) while the input is empty.

tjwiebell: I believe so. The effect immediately sets cleared to false, but doesn't account for the debounced delay. Very minor, but I think I captured it here.

jimbo: Right. As you can see in the gif, when the input is empty, the stale results are not shown. Instead, "Search for a product" is shown as a prompt. This is the new expected behavior.

The key thing I'm trying to communicate here is that we can't actually blow away the stale results of the query. Apollo gives us no method to do that. So we can only hide them, and I only want to hide them while the input is empty. 👍

tjwiebell: Fair enough. I think this debounced function provides a better experience; give it a try, and if you're still comfortable with the existing behavior, I'm fine with this as-is. Clean up the unit test failures and remove `location` propType in `SearchField` and I'll approve.

```js
    () =>
        debounce(options => {
            setCleared(false);
            runQuery(options);
        }, 500),
    [runQuery]
);
```

jimbo: Good suggestion. I'll incorporate it.

dpatil-magento:
@jimbo I can still reproduce the issue in one case - Enter "Tops Dress" > Wait for suggestions and delete the entire input via delete key. Now click outside the textfield and click back in.  Search suggestions are shown for "Tops"

image

tjwiebell:
> @jimbo I can still reproduce the issue in one case - Enter "Tops Dress" > Wait for suggestions and delete the entire input via delete key. Now click outside the textfield and click back in. Search suggestions are shown for "Tops"

Looked at this with Dev; looks like its easiest to reproduce by holding down backspace to remove the search value.

jimbo:
> > @jimbo I can still reproduce the issue in one case - Enter "Tops Dress" > Wait for suggestions and delete the entire input via delete key. Now click outside the textfield and click back in. Search suggestions are shown for "Tops"
> 
> Looked at this with Dev; looks like its easiest to reproduce by holding down backspace to remove the search value.

This is so bizarre. I can replicate this with `tops dresses` but not with `athena`. I'll have to look into it.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -15,39 +15,48 @@ import debounce from 'lodash.debounce';
...
+            }, 500),
+        [runQuery, setCleared]
+    );
 
     // run the query once on mount, and again whenever state changes
     useEffect(() => {
         if (visible && valid) {
             debouncedRunQuery({ variables: { inputText: value } });
+        } else if (called && !value && !visible) {
+            setCleared(true);
```
sirugh: What's happening is that as you delete the text of `tops dresses` the last "valid" search text is `top` which is used as the debounced query. That explains why you see product suggestions for `top` which are based on the last search data, but you don't see any text in the `<input> in <Category>` suggestions.

sirugh: I fixed this by moving the `setCleared` logic into the debounced run query. The last call that executes should control the UI. In this case, we don't need to validate input before the debounce, we can do that inside the debounce, and if it is invalid we just clear the view.

jimbo:
@dpatil-magento I think this is fixed.

dpatil-magento:
After clearing if I click back > stale results display for fraction of second and goes away.
https://gyazo.com/4d31b77f18f0c7f07b6b6c126b3dff4c

dpatil-magento:
Now the valid search suggestions also not displaying. 

<a href="https://gyazo.com/2d480f029ff47bff282c11835d0d2b48">Image from Gyazo</a>

jimbo:
@dpatil-magento I've fixed that issue. Please give it another test.

**Important:** we had logic that was reading the value of `query` from the URL and writing it to the search field. This logic was flawed and had a negative effect on the UX anyway, so after discussing it with @soumya-ashok, I removed this logic. As a result, if you navigate directly to the search page, it will no longer prepopulate the search field.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -61,6 +60,8 @@ export const useAutocomplete = props => {
         messageType = 'RESULT_SUMMARY';
     }
 
+    console.log(messageType === 'PROMPT' && { valid, hasResult });
```
sirugh: O.O

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -11,43 +11,48 @@ import debounce from 'lodash.debounce';
...
 
     // Create a debounced function so we only search some delay after the last
     // keypress.
-    const debouncedRunQuery = useMemo(() => debounce(runQuery, 500), [
-        runQuery
-    ]);
+    const debouncedRunQuery = useMemo(
+        () =>
+            debounce(inputText => {
+                runQuery({ variables: { inputText } });
```
sirugh: This now runs with the last 3 characters in the input when you start deleting some text value. For example "top shirt" will search "top shirt" but when you delete that character by character it'll query for "top" since it is the last valid value on the way to an empty string.

I would prefer we don't make an extra call for text that may never be searched but it's also a minor optimization so :shrug:

dpatil-magento:
@jimbo Looks good but still there is minor issue.  Input `dress` wait for search suggestion to show, now clear the search field and enter "Top" > For fraction of seconds old search suggestions display and then new search suggestions. 

<a href="https://gyazo.com/1ebd419d337e80ae1edd9741f622aeff">Image from Gyazo</a>

jimbo:
> @jimbo Looks good but still there is minor issue. Input `dress` wait for search suggestion to show, now clear the search field and enter `Top` > For fraction of seconds old search suggestions display and then new search suggestions.

That's fine. We don't have any defined loading behavior yet, so it's going to show stale results while the next query is resolving.

PR #2151, opened 2020-02-08
Added real swatch color resolves #1633

### LINKED ISSUE #1633: [feature]: Utilize graphql returned color for swatches
Currently we randomize the colors in the product component because GQL does not return the hex code or any way for the UI to determine what color to display. There is an open issue for this in GQL: https://github.com/magento/graphql-ce/issues/196.

When the GQL issue is resolved/merged/released I would like to update our product query and swatch code to use the hex value returned.

If there was a mapping of colors returnable by the gql endpoint and hex colors we could maintain it on the UI and render a map. I don't think this is scaleable or a good idea, but it is an alternative :P

Another more realistic alternative would be to get the data from the rest endpoint instead of graphql. I'm not sure how to obtain that information or whether it even exists, but it should be a fallback incase we need to be "production ready" and don't have this data available in gql yet.

**Additional context**
https://github.com/magento/graphql-ce/issues/196
https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/ProductOptions/swatch.js#L58-L59

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

deiserh:
## Description
Changed the random color of the swatch to the real one.

## Related Issue

Closes #1633 .

### Verification Steps

1. Go to a configurable product.
2. The colors of the swatches are the real ones.

## Checklist

* I have updated tests to cover my changes.

## Compatibility
With Magento 2.3.5 or higher


deiserh:
What should I do regarding node failed?

jimbo:
Awesome. Looks like a good, straightforward implementation.

Does GQL always return a single hex color? Can it return multiple colors or an image?

sirugh:
> What should I do regarding node failed?

Run the `validate-queries` script locally. See if the output indicates the failure reason. If you need help after that, let us know.

larsroettig:
> Awesome. Looks like a good, straightforward implementation.
> 
> Does GQL always return a single hex color? Can it return multiple colors or an image?

Hi @jimbo, 
your are correct ** image**  is allowed  see: https://docs.magento.com/m2/ce/user_guide/catalog/swatch-create.html

For Hex color you can only have one by default. 

I will test this Ticket tomorrow with an image swatch  and update this comment if it is
working.

best regards, 

Lars

deiserh:
I Updated the Pull Request the error of validate-queries is that some functions are depricated and the Magento Version should be between 2.3.3-2.3.4 but the swatch graphql api was introduced in 2.3.5.

sirugh:
>  the swatch graphql api was introduced in 2.3.5.

Ah, there is the problem. We develop against the latest released version ie 2.3.4, so this will cause our CI to start failing if we merge it as is.

This may be a good use case for the new extensibility framework though I don't think we have published our docs for it. If you urgently need this feature in your own line of work I would suggest making the change on your own fork for now.

If you want to see how we implemented page builder as an extension, check out the `packages/pagebuilder` directory, specifically the `intercept` file and how it taps into `richContentRenderers`. I'd love to hear your thoughts on how we might approach implementing a 2.3.5 feature as an extension.

zetlen:
I think this PR is absolutely a candidate for core functionality; it should be our default swatch implementation, instead of the one that uses random colors!

This kind of thing _can_ be implemented as an extension, but I think this changeset in particular should be merged (once 2.3.5 is out).

In the meantime, @deiserh, please maintain this code in your own fork and we'll merge it as soon as our CI is rebased on to 2.3.5.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2151.pwa-venia.com : LH SEO Expected 1 Actual 0.9

packages/venia-ui/lib/components/ProductOptions/swatch.css
```diff
@@ -1,6 +1,6 @@
 .root {
     composes: root from './tile.css';
-    background-color: rgb(var(--venia-swatch-bg));
+    background: var(--venia-swatch-bg);
```
sirugh, line 3: The `background` property is shorthand and allows us to use a single variable to provide either the URL or the color.

packages/venia-ui/lib/util/makeUrl.js
```diff
@@ -21,7 +21,8 @@ const joinUrls = (base, url) =>
 
 const mediaBases = new Map()
     .set('image-product', 'catalog/product/')
-    .set('image-category', 'catalog/category/');
+    .set('image-category', 'catalog/category/')
+    .set('image-swatch', 'attribute/swatch');
```
sirugh, line 25: Is this where swatch images are served? I'd have to test this but I don't have a magento instance configured to try it out. @dpatil-magento do you know how to test this?

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2151.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.71, LH Accessibility Expected 0.9 Actual 0.87

packages/venia-ui/lib/fragments/productDetails.graphql
```diff
@@ -49,6 +49,9 @@ fragment ProductDetails on ProductInterface {
                 store_label
                 use_default_value
                 value_index
+                swatch_data {
+                    value
```
tjwiebell, line 53: Magento already has an optimized thumbnail which I think would be better to use; this also gives you better indicator that its an image instead of color swatch. Apply same to below.

```suggestion
                    ... on ImageSwatchData {
                        thumbnail
                    }
                    value
```

packages/venia-ui/lib/util/makeUrl.js
tjwiebell, line 25:
```suggestion
    .set('image-swatch', 'attribute/swatch/');
```

tjwiebell: I've confirmed on my local that both image swatch `value` and `thumbnail` are in this path. `value` returns a relative path, while `thumbnail` returns an absolute path (which I think is less fragile to bad config).

packages/venia-ui/lib/components/ProductOptions/swatch.js
```diff
@@ -43,18 +43,28 @@ const Swatch = props => {
 
     const classes = mergeClasses(defaultClasses, props.classes);
 
-    // TODO: use the colors from graphQL when they become available.
-    //   https://github.com/magento/graphql-ce/issues/196
-    //   https://github.com/magento/pwa-studio/issues/1633
-    const randomColor = memoizedGetRandomColor(value_index);
+    let swatchValue;
+    const isImage = swatch_data && swatch_data.value.charAt(0) !== '#';
```
tjwiebell, line 47:
```suggestion
    const isImage = swatch_data && swatch_data.thumbnail;
```

packages/venia-ui/lib/components/ProductOptions/swatch.js
```diff
@@ -43,18 +43,28 @@ const Swatch = props => {
...
+
+    if (isImage) {
+        const imagePath = generateUrlFromContainerWidth(
```
tjwiebell, line 50: So, having provided feedback on image swatch, I'm a bit apprehensive to include it in this scope. If I use one of our 4x5 images this is the result as-is:
Screen Shot 2020-05-05 at 4 26 07 PM

This is the result if I use the optimized thumbnail from GraphQL:
Screen Shot 2020-05-05 at 4 26 53 PM

And finally if I change the implementation to hard-code width/height to the exact dimensions of these tiles (3rem/3rem or 48px/48px):
Screen Shot 2020-05-05 at 4 27 12 PM

We're making progress on not assuming every images aspect ratio, but a square box now forces us to actually support this natively. Would like the team's thoughts on this; but I'm going to propose we remove support for image swatch until we can support different aspect ratios.

I would be willing to accept this if we put in a comment that says you **must** upload swatch images with a 1x1 aspect ratio, and change implementation to this (only way I could get images to render correctly):
```
const imagePath = generateUrl(swatch_data.thumbnail, 'image-swatch')(48);
```

sirugh: >I'm going to propose we remove support for image swatch until we can support different aspect ratios.

I think that is fair. The initial scope of this ticket was to replace random color with the actual color.

However it does seem like `swatch_data.value` can be an image url. We still need to have a fall-back thing to display.

sirugh:
Upon further review it looks like this needs some more work to get the color filter swatches and the product edit swatches functional.

sirugh:
That said, we may have to remove the filter swatches because the `aggregation` response from the server does not include color values:

[Image from Gyazo](https://gyazo.com/fe1d08b475f4a7c44c2d437ed70370bb)

[Image from Gyazo](https://gyazo.com/dafe421a524ee4c0b3501da273fbec62)

packages/venia-ui/lib/components/ProductOptions/swatch.js
```diff
@@ -43,30 +46,31 @@ const Swatch = props => {
...
-            swatch_data.value,
-            imageWidths.get('ICON'),
-            'image-swatch'
-        );
-
-        swatchValue = `url("${imagePath}")`;
-    } else {
-        swatchValue = swatch_data.value;
+    let finalStyle = style;
+    let swatchValue = '';
```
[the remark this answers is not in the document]
sirugh: Oh oops :D

packages/venia-ui/lib/components/ProductOptions/swatch.js
```diff
@@ -43,30 +46,31 @@ const Swatch = props => {
...
+    if (swatch_data) {
+        const isImage = swatch_data && swatch_data.thumbnail;
```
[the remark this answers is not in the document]
sirugh: Yea I can clean this up :)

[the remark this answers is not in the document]
sirugh: Err `swatch_data` could be `undefined` so I'll just try to handle this gracefully.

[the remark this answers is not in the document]
sirugh: Resolved in 69386ab

packages/venia-ui/lib/components/ProductOptions/swatch.js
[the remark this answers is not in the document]
sirugh: 69386ab

deiserh:
Thank you for finalizing the pr :grin:

PR #2156, opened 2020-02-10
Applied Gift Cards show current balance

supernova-at:
## Description

This PR adds current applied gift card balances to the UI to improve the UX.

The balance font size is slightly smaller than the code, so I consulted the Venia Styleguide for its value.

## Related Issue

Closes PWA-364.

### Verification Steps

1. Go to the `/cart` page
1. Apply a gift card or two
1. See that the card's balance appears along with its code

Screen Shot 2020-02-10 at 12 22 20 PM

## Checklist


packages/venia-ui/lib/components/CartPage/GiftCards/giftCard.css
```diff
@@ -1,4 +1,13 @@
 .card_info {
+    display: flex;
+    flex-direction: column;
+}
+
+.balance {
+    font-size: 0.875rem;
```
supernova-at: This is computed based on the Venia Styleguide:

Screen Shot 2020-02-10 at 11 52 40 AM

Since we don't have CSS Properties (variables) for font sizes yet, `0.875rem` = `14px`.

packages/venia-ui/lib/components/CartPage/GiftCards/__tests__/__snapshots__/giftCard.spec.js.snap
```diff
@@ -16,9 +22,15 @@ Array [
...
 Array [
-  <span>
-    unit test card code
-  </span>,
+  <div>
+    <span>
+      unit test card code
+    </span>
+    <span>
+      Balance:
```
sirugh: There is no balance in these snaps, not even a zero. Expected?

schensley:
@supernova-at the implementation looks very good.  One fussy thing, the balance could use a little bit more separation for the card number once it has been applied.

packages/venia-ui/lib/components/CartPage/GiftCards/__tests__/__snapshots__/giftCard.spec.js.snap
supernova-at: Expected because the value is rendered via the `Price` component, which gets mocked here to an empty `<div />` (see next line).

packages/venia-ui/lib/components/CartPage/GiftCards/__tests__/__snapshots__/giftCards.spec.js.snap
```diff
@@ -3,18 +3,52 @@
...
+        <span>
+          99
+        </span>
+        <span>
+          .
+        </span>
+        <span>
+          00
+        </span>
+      </span>
```
supernova-at: @sirugh I think this `Price` component doesn't get mocked because we `requireActual` Peregrine in this `.spec.js`. Apparently that will ignore mocks in `__mocks__`?

supernova-at:
> @supernova-at the implementation looks very good. One fussy thing, the balance could use a little bit more separation for the card number once it has been applied.

@schensley , resolved by https://github.com/magento/pwa-studio/pull/2156/commits/af7baf67a686b6d57b2be05239bc936a41cf3e3d.

PR #2157, opened 2020-02-10
PWA-213: [UNIT TEST] Email Validation on Signin, Create Acount, Forgot Password

tjwiebell:
## Description

Reviewed this task with team lead, and it turns out this feature should have never been implemented in the first place. There is little value in doing client-side validation, and the most we should do in our app is validate the email address isn't in use, which isn't currently possible with the form library we're using. He requested we strip this out, which we're doing in the scope of this task now. Article for reference: https://medium.com/hackernoon/the-100-correct-way-to-validate-email-addresses-7c4818f24643

## Related Issue

* [[PWA-213](https://jira.corp.magento.com/browse/PWA-213)] [UNIT TEST] Email Validation on Signin, Create Acount, Forgot Password

### Verification Steps

For each of the forms that had this validation (SignIn, Forgot Password, Create Account, Checkout Shipping/Billing/Payment)
1. Submit clean form and verify required validation still fires
2. Enter an "invalid" email address in email address field and verify it no longer complains (eg. `test@domain`)

## Checklist


supernova-at:
✅ Code

Verification steps pass, except when I attempt enter "test@domain" for a shipping address during checkout and submit the form, I'm met with an error toast and the following console message:

Screen Shot 2020-02-11 at 9 47 58 AM

Is that in scope here or should we split it out into a separate issue?

sirugh:
IMO we should cover that here since we are "breaking" the current checkout flow.

tjwiebell:
> IMO we should cover that here since we are "breaking" the current checkout flow.

Already on it, but may ping you tomorrow morning for some weirdness. Even though I'm catching these errors they're still hitting the error boundary and popping a toast 😢

tjwiebell:
@supernova-at I've resolved the uncaught exception issue in fac582c.

@jimbo This is ready for review if you wanted to give it a once over.

dpatil-magento:
QA Pass. After removing client side validation encountered bug in Billing information form. Which will be addressed as part of separate ticket.

PR #2164, opened 2020-02-11
Images Appear

supernova-at:
## Description

This PR updates some CSS handling in our `Image` component to no longer apply `display: none` to images that we think haven't loaded yet, since Chromium (as of version 80) is no longer requesting those images.

Instead, we set `visibility: hidden`. Chromium still sends off the request and nothing else downstream needs to change.

## Related Issue

Closes PWA-367.

### Verification Stakeholders

@jimbo 

#### Venia

1. `yarn watch:venia`
1. Navigate to `/pomona-skirt.html`
1. See that all images load correctly
1. Verify that all images across the entire Venia site load correctly

#### Storybook

1. `yarn workspace @magento/venia-ui run storybook`
1. Verify that all the stories in the `Image` component load correctly

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/Image/image.css
+++ b/packages/venia-ui/lib/components/Image/image.css
@@ -16,16 +16,11 @@
      */
 }
 
-.loaded {
+.image_positioned {
     display: block;
     position: absolute;
     top: 0;
     left: 0;
-}
-
-.notLoaded {
-    composes: loaded;
-    display: none;
 }
 
 .placeholder {
--- a/packages/venia-ui/lib/components/Image/image.css
+++ b/packages/venia-ui/lib/components/Image/image.css
@@ -16,11 +16,16 @@
      */
 }
 
-.image_positioned {
-    display: block;
+.loaded {
     position: absolute;
     top: 0;
     left: 0;
+    visibility: visible;
+}
+
+.notLoaded {
+    composes: loaded;
+    visibility: hidden;
 }
 
 .placeholder {
--- a/packages/venia-ui/lib/components/Image/image.js
+++ b/packages/venia-ui/lib/components/Image/image.js
@@ -65,7 +65,8 @@ const Image = props => {
 
     const classes = mergeClasses(defaultClasses, propsClasses);
     const containerClass = `${classes.root} ${classes.container}`;
-    const imageClass = `${classes.image} ${classes.image_positioned}`;
+    const isLoadedClass = isLoaded ? classes.loaded : classes.notLoaded;
+    const imageClass = `${classes.image} ${isLoadedClass}`;
 
     // If we have a src, use it directly. If not, assume this is a resource image.
     const actualImage = src ? (
--- a/packages/venia-ui/lib/components/Image/image.js
+++ b/packages/venia-ui/lib/components/Image/image.js
@@ -65,8 +65,7 @@ const Image = props => {
 
     const classes = mergeClasses(defaultClasses, propsClasses);
     const containerClass = `${classes.root} ${classes.container}`;
-    const isLoadedClass = isLoaded ? classes.loaded : classes.notLoaded;
-    const imageClass = `${classes.image} ${isLoadedClass}`;
+    const imageClass = `${classes.image} ${classes.image_positioned}`;
 
     // If we have a src, use it directly. If not, assume this is a resource image.
     const actualImage = src ? (
--- a/packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
+++ b/packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
@@ -1,0 +1,168 @@
+import { useCallback } from 'react';
+import { useApolloClient, useMutation, useQuery } from '@apollo/react-hooks';
+
+import { useCartContext } from '../../../../context/cart';
+
+/**
+ * GraphQL currently requires a complete address before it will return
+ * estimated shipping prices, even though it only needs Country, State,
+ * and Zip. Assuming this is a bug or oversight, we're going to mock the
+ * data points we don't want to bother collecting from the Customer at this
+ * step in the process. We need to be very mindful that these values are never
+ * displayed to the user.
+ */
+const MOCKED_ADDRESS = {
+    city: 'city',
+    firstname: 'firstname',
+    lastname: 'lastname',
+    street: ['street'],
+    telephone: 'telephone'
+};
+
+export const useShippingForm = props => {
+    const {
+        getCountriesQuery,
+        getStatesQuery,
+        selectedValues,
+        setShippingMutation,
+        shippingMethodsQuery
+    } = props;
+
+    const [{ cartId }] = useCartContext();
+    const apolloClient = useApolloClient();
+
+    const {
+        data: countriesData,
+        error: countriesError,
+        loading: isCountriesLoading
+    } = useQuery(getCountriesQuery);
+
+    const {
+        data: statesData,
+        error: statesError,
+        loading: isStatesLoading,
+        refetch: fetchStates
+    } = useQuery(getStatesQuery, {
+        variables: { countryCode: selectedValues.country }
+    });
+
+    const [setShippingAddress, { loading: isSetShippingLoading }] = useMutation(
+        setShippingMutation
+    );
+
+    const handleCountryChange = useCallback(
+        country => {
+            if (country) {
+                fetchStates({
+                    countryCode: country
+                });
+            }
+        },
+        [fetchStates]
+    );
+
+    /**
+     * When the zip value is changed, go ahead and manually wipe out that
+     * portion of the cache, which will cause the components subscribed to
+     * that state to re-render. Ideally we would send a mutation to clear the
+     * shipping address set on the cart, to keep those states in sync, but the
+     * GraphQL API does not currently supported clearing addresses.
+     */
+    const handleZipChange = useCallback(
+        zip => {
+            if (zip !== selectedValues.zip) {
+                const data = apolloClient.readQuery({
+                    query: shippingMethodsQuery,
+                    variables: {
+                        cartId
+                    }
+                });
+
+                const { cart } = data;
+                const { shipping_addresses: shippingAddresses } = cart;
+                if (shippingAddresses.length) {
+                    const primaryAddress = shippingAddresses[0];
+                    const {
+                        available_shipping_methods: availableMethods
+                    } = primaryAddress;
+                    if (availableMethods.length) {
+                        apolloClient.writeQuery({
+                            query: shippingMethodsQuery,
+                            data: {
+                                cart: {
+                                    ...cart,
+                                    shipping_addresses: [
+                                        {
+                                            ...primaryAddress,
+                                            available_shipping_methods: []
+                                        }
+                                    ]
+                                }
+                            }
+                        });
+                    }
+                }
+            }
+        },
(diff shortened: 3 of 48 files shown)


packages/venia-ui/lib/components/Image/image.css
```diff
@@ -16,18 +16,13 @@
      */
 }
 
-.loaded {
+.image_positioned {
     display: block;
```
sirugh: So now we are just overlaying the image on top of the placeholder?

packages/venia-ui/lib/components/Image/image.css
```diff
@@ -16,18 +16,13 @@
...
     position: absolute;
     top: 0;
     left: 0;
 }
 
-.notLoaded {
-    composes: loaded;
-    display: none;
```
sirugh: I feel like the code deserves a TODO or something. Instead of changing the classes applied, perhaps we could just comment out this line and add a comment here and in the `onLoad` callback about it not functioning as expected.

sirugh: Or perhaps this is the right approach and we should always just overlay the loaded image over the placeholder...Hmm

brendanfalkowski: Just started seeing this happen in Chrome this week with our logo (which uses `Image`). Note: we are using ~v4 code currently.

FYI: overlaying probably is not safe. We have a transparent background on some assets (i.e. brand logos) passed into `Image`.

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -65,8 +65,7 @@ const Image = props => {
 
     const classes = mergeClasses(defaultClasses, propsClasses);
     const containerClass = `${classes.root} ${classes.container}`;
-    const isLoadedClass = isLoaded ? classes.loaded : classes.notLoaded;
-    const imageClass = `${classes.image} ${isLoadedClass}`;
+    const imageClass = `${classes.image} ${classes.image_positioned}`;
```
jimbo: So I know this snuck through the last couple PRs, but this is really not how we're supposed to be using classnames. Applying multiple classnames in this way creates ambiguity, since the source order of CSS *rulesets* matters but the order of *classnames* does not (and the source order of rulesets is beyond our control). That ambiguity manifests as bugs.

Instead, each element we create should have **only one** classname applied to it (e.g., `classes.image`), and we should use `composes` to ensure that all the right styles have been included. Of course, CSS Modules may map its keys to multiple classnames (as output), but [proper use of `composes`](https://github.com/css-modules/css-modules#composing-from-other-files) ensures that they're deterministic.

packages/venia-ui/lib/components/Image/image.css
supernova-at: The overlay is always onto a transparent background placeholder (unless a different placeholder is explicitly passed in), so no worries there.

packages/venia-ui/lib/components/Image/image.js
supernova-at: I actually did change this but ran into:

Callers can (and do) pass in overrides to `classes.image` and `classes.root`.

I tried setting `classes.container` to compose `root` (and `classes.image_positioned` to compose `image`), but they weren't picking up the passed-in overrides.

Let me know if that doesn't make sense - I'd love to fix this but I'm not sure how.

packages/venia-ui/lib/components/Image/image.css
supernova-at: Yes

supernova-at:
Well, I'm not sure that this is the way to go anymore. For one thing, the `PlaceholderImage` still uses the `isLoaded` flag, which isn't getting set.

And then watch as the load event doesn't fire until I alter a CSS property:

logo2

packages/venia-ui/lib/components/Image/image.js
jimbo: You're correct. When overriding classnames, a caller must pass in overrides for the final classnames being applied, not the "base" classnames. Take the following sample component:

```jsx
const Foo = props => {
  const [isOpen, setIsOpen] = useState(false)
  const classes = mergeClasses(defaultClasses, props.classes)
  const rootClass = isOpen ? classes.root_visible : classes.root_hidden

  return <div className={rootClass}>{props.children}</div>
}
```

```css
.root {
  color: red;
  display: grid;
}

.root_visible {
  composes: root;
}

.root_hidden {
  composes: root;
  display: none;
}
```

In this example, the classnames that make up `Foo`'s public API are `classes.root_hidden` and `classes.root_visible`, the only keys that `Foo` applies. The component does not apply `classes.root` to anything, so that classname is not part of `Foo`'s API at all. It's essentially a mixin that the CSS module may opt to use to avoid some repetition—nothing more.

sirugh:
I wonder if `display:none` is being used as an optimization to not load the image 😕

supernova-at:
> I wonder if display:none is being used as an optimization to not load the image 😕

Good thought. [This article](https://csswizardry.com/2018/06/image-inconsistencies-how-and-when-browsers-download-images/) is a little dated but I found

Screen Shot 2020-02-12 at 8 41 37 PM

May be worth trying again in Chrome 80 since we didn't see this bug until then.

supernova-at:
Yep, I think you're right: https://codesandbox.io/s/nervous-minsky-rwqz1

There's an image with `loading="lazy"` and `display: none` that is out of the viewport initially. When you scroll down to it, the request never fires in the network tab.

When you remove the `display: none` styling, you can see it fire the request for the image.

supernova-at:
So I'm pretty confident we can fix that in our code but the question becomes why does Chrome think our `Logo` and primary product image are out of the viewport initially - they are most definitely not.

brendanfalkowski:
I'm wondering if Chromium sees the `lazy` image occupying the same physical space as the `eager` image so it chooses not to request the lazy image.

Is there a reason you need two images with different `loading` attribute values? I would probably have started with something like:

```
<div class="wrapperWhichSetsThePlaceholderAsBackgroundImage">
    <img loading="lazy" src="xyz" />
</div>

becomes

<div class="wrapperWithNoBackgroundImage">
    <img loading="lazy" src="xyz" />
</div>
```

The placeholder `background-image` would be assigned by CSS, and would be guaranteed to load plus be cached to render instantly later. You'd still need to get the "loaded" event properly to change the class, and that seems to be complicating.

I would simplify and not have a loading image. There's a lot happening to support that, but I'd think it's really only important to anchor the physical space. Whether that's a filled by a solid, transparent, or image is secondary / depends on the design.

Aside: isn't base64 decoding a PNG bad for performance on most devices? I recall research that it was always better to take the HTTP hit and have the physical file in cache. Even for an SVG there was a limit to the gain of inlining the markup vs referencing a file source. If the SVG itself was complex the DOM size bloat from repetition was again worse than the one-time HTTP hit.

supernova-at:
> the question becomes why does Chrome think our Logo and primary product image are out of the viewport initially - they are most definitely not.

Update: editing that sandbox, it doesn't matter whether the images are in the viewport initially or not. If they have `display: none`, Chromium v80 is not requesting them.

So THEN the question becomes why do some of the images work just fine and others don't? 🤪

supernova-at:
PR Updated: 
  * Changes to using `visibility: hidden` instead of `display: none` to "hide" the image until it is loaded

Chromium@80 does request images that are `visibility: hidden`, so I think this is the smallest code change we can make that will fully solve the problem.

Speaking of small code changes, since this is a Priority 1 I'd like to keep the scope extremely small and separate out the other suggested improvements to `Image` into their own tickets.

packages/venia-ui/lib/components/Image/image.css
supernova-at: Outdated.

packages/venia-ui/lib/components/Image/image.css
supernova-at: Outdated.

packages/venia-ui/lib/components/Image/image.js
supernova-at: Hoping we can split this out into a separate ticket so we can resolve this P1 with minimal impact.

dpatil-magento:
@supernova-at  Not sure if its related to your changes but somehow its happening on this PR. On Win10-Edge - Home page image does not load.  Same works if I test on `develop`

Also same I could reproduce on Safari browser.  Whenever Response Header `Content-Type: image/webp` image loads blank.

image

jimbo:
> Is there a reason you need two images with different `loading` attribute values? I would probably have started with something like:
> 
> ```
> <div class="wrapperWhichSetsThePlaceholderAsBackgroundImage">
>     
> </div>
> 
> becomes
> 
> <div class="wrapperWithNoBackgroundImage">
>     
> </div>
> ```
> 
> The placeholder `background-image` would be assigned by CSS, and would be guaranteed to load plus be cached to render instantly later. You'd still need to get the "loaded" event properly to change the class, and that seems to be complicating.
> 
> I would simplify and not have a loading image. There's a lot happening to support that, but I'd think it's really only important to anchor the physical space. Whether that's a filled by a solid, transparent, or image is secondary / depends on the design.

The reasoning behind the technique used here is that the layout behavior of `` (namely, preserving aspect ratio while scaling) can't be accomplished with CSS unless the stylesheet is told the aspect ratio. Since we don't tell the CSS what aspect ratio our images are, we use a transparent image to allow the container to perform layout while the image loads. And since `` has that cool scaling, we can use a very tiny placeholder (`4x5`).

**However...**

Since we only use a single, fixed-size placeholder, we're not really gaining anything with this technique. If it's fixed, then the CSS might as well know the size, and we have precedent for doing so in other places by letting the component pass sizes to the stylesheet via CSS variables. Worth a PR in the future probably.

> Aside: isn't base64 decoding a PNG bad for performance on most devices? I recall research that it was always better to take the HTTP hit and have the physical file in cache. Even for an SVG there was a limit to the gain of inlining the markup vs referencing a file source. If the SVG itself was complex the DOM size bloat from repetition was again worse than the one-time HTTP hit.

I doubt it's that bad for a tiny PNG (`4x5`). At the time, though, we did this for expediency.

packages/venia-ui/lib/components/Image/image.css
```diff
@@ -17,15 +17,15 @@
...
     position: absolute;
     top: 0;
     left: 0;
+    visibility: visible;
 }
 
 .notLoaded {
     composes: loaded;
-    display: none;
+    visibility: hidden;
```
jimbo: Approving this PR, but for posterity, I have to note here that the way this was done (originally) was not the recommended approach.

- the styles on `.loaded` should be on `.image`
- `.loaded` should compose `.image`
- `.notLoaded` should not compose `.loaded`
- `. notLoaded` should compose `.image` and override `visibility`
- (most importantly) `.image` should not be applied directly to the component

brendanfalkowski:
@jimbo That makes sense.

FWIW, we might go simpler and have `<Image>` be dimensionless letting the browser figure it out (or pass params), and split a separate `<ProductImage>` component that has a fixed aspect ratio + associated styling.

I've never worked on a Magento site that didn't have a standard aspect ratio for product images as part of product data acceptance (overwhelmingly 1:1 but the fashion-torso prevails).

There are cases we carve a recess for `<Image>` with padding on a parent and absolutely position it within so repaint doesn't affect reflow, but the vast majority of image declarations would be `<ProductImage>` with a hardcoded scale.

dpatil-magento:
Chrome, Firefox, Win10-latest Edge, Android Chrome loads images fine. 
We are tracking intermittent Safari issue with Fastly team.

PR #2165, opened 2020-02-12
[bugfix]: search trigger focus

sirugh:
## Description
Clicking the search trigger should focus the search input. This was done before, but we broke it at some point in our talon refactor. I also replaced some components' use of `location` and `history` with the equivalent hooks.

Old issue: https://github.com/magento/pwa-studio/pull/1019

## Current Behavior

Clicked search trigger, press tab, then press shift tab.

[Image from Gyazo](https://gyazo.com/6d914759537d0b19ae1d04059000fe8c)

## New Behavior

Clicked search trigger, press tab, then shift tab, shift tab.
[Image from Gyazo](https://gyazo.com/795d7e87dc6e6b101dc7ff73d72dc802)

## Tab Order

I opted not to change tab order, only to focus the input element on click.

This does result in a strange behavior where tabbing out of the search input moves to the first item in the page, thus skipping over the cart trigger.

That said, in my opinion this is a better UX than before.

* Sighted users can now click the trigger and immediately start typing a search term.
* Non sighted users/keyboard-only users are automatically moved to the search input instead of having to tab further to find the field.

## Related Issue

Closes [PWA-371](https://jira.corp.magento.com/browse/PWA-371).

### Verification Stakeholders

@soumya-ashok or @schensley and @jimbo 
### Verification Steps

1. Open any venia page and click the search trigger.
2. Observe the focus moves directly to the input.
3. Click the search trigger again. Focus should now be back on the search trigger.

## Checklist


packages/venia-ui/lib/components/SearchBar/__tests__/searchField.spec.js
```diff
@@ -1,39 +1,53 @@
...
+
 jest.mock('../../../classify');
 jest.mock('../../Trigger', () => () => null);
 
 const onChange = jest.fn();
 const onFocus = jest.fn();
 
 test('renders correctly', () => {
+    const location = { pathname: '/' };
+    useLocation.mockReturnValueOnce(location);
```
revanth0212: Instead of writing this in every test, can we move this to `beforeEach` and change it in a certain test if needed? Or like you did for `useLocation` in `packages/peregrine/lib/hooks/__tests__/useSearchParam.spec.js `.

sirugh: Sure, I could just have it return this by default and only do the override in the one test.

revanth0212:
Tests failing. Can you check?

packages/peregrine/lib/talons/SearchBar/useSearchBar.js
```diff
@@ -6,10 +6,10 @@ import { useDropdown } from '../../hooks/useDropdown';
 const initialValues = { search_query: '' };
 
 export const useSearchBar = () => {
-    const [valid, setValid] = useState(false);
-    const { elementRef, expanded, setExpanded } = useDropdown();
     const history = useHistory();
     const { push } = history;
+    const [valid, setValid] = useState(false);
+    const { elementRef, expanded, setExpanded } = useDropdown();
```
jimbo: Is there some meaning in this change that I'm missing?

sirugh: Just how it rebased I guess.

jimbo: Cool, reverted it. This is approved now.

PR #2166, opened 2020-02-12
[feature]: Connect Filtering to GraphQL!

sirugh:
## Description
This PR hooks up the filter state on both the category and search pages to the expected format by graphql. There are some pretty major changes in this PR as it breaks any pages that used filtering (search and category). 

Some noteable changes:

* ~I opted to remove the ability to filter by `price` on both pages because it was a weird experience. I also removed the ability to filter by category when not on the search page (ie category). It didn't make sense to filter for categories when already having navigated to a category page. I'll open follow-up tickets incase to capture any concerns.~ Filters are static based on a single input. For category pages we get the possible filters for a given category id. For search filters we get possible filters based on the search text. This does not use "aggregations" for layered navigation, but it does provide a fairly decent filter experience.

* There was a change from `filter` to `aggregations` for filtering. This required an introspection query since aggregations and filter inputs did not align. You will see code that does introspection in order to get a union of the possible aggregations and possible input types. 

* There is "helper" code that now converts an input into the expected input types. For example, we know that `category_id` is a `FilterEqualTypeInput` but what does it mean to format that for graphql? A helper function is born :)

* I simplified the filter format in the url. Since it is possible to have multiple query params for a filter, like category, I wanted the query string to be less verbose. It now looks like `category_id[filter]=<title>,<value>` which is shorter from `category_id[title]=<title>&category_id[value]=<value>`. 

## Related Issue

Closes [PWA-369](https://jira.corp.magento.com/browse/PWA-369).

### Verification Steps

1. Regress filters on a category page.
2. Regress the search functionality.

## Checklist


packages/venia-ui/lib/queries/productSearch.graphql
```diff
@@ -2,21 +2,21 @@ query productSearch(
     $currentPage: Int = 1
     $inputText: String!
     $pageSize: Int = 6
-    $categoryId: String
+    $filters: ProductAttributeFilterInput!
```
sirugh: This is interesting - there is no "Object" type, and locally we have no knowledge of `ProductAttributeFilterInput`. We know this query requires this type though, because of the schema. I just thought it was weird.

packages/venia-ui/lib/components/FilterModal/CurrentFilters/currentFilters.js
```diff
@@ -25,7 +34,10 @@ const CurrentFilters = props => {
                         <CurrentFilter
                             group={group}
                             groupName={groupName}
-                            item={item}
+                            item={{
+                                title,
```
sirugh: Blah I don't like this

packages/venia-ui/lib/RootComponents/Category/category.js
```diff
@@ -26,14 +43,58 @@ const Category = props => {
...
+        filters.forEach((values, key) => {
+            newFilters[key] = getFilterInput(values, filterTypeMap.get(key));
+        });
+
+        // TODO: Category filtering on a category page is weird. How should we handle it? Currently if we don't have a filter, as in navigated from Home, we will just use the default category. If a user then filters by category, should we overwrite the category page category? It would be possible to be on the "Bottoms" category page but filter for "Tops", and see tops, but have nothing else on the page change ie breadcrumbs, etc.
+        // If a user is not filtering by category, just use the one for the page
+        // we are currently on.
+        if (!filters.get('category_id')) {
+            newFilters['category_id'] = { eq: String(id) };
+        }
```
sirugh: Spoke with @jimbo and the ideal UX does not seem possible using server side only (gql) response data.

For example, suppose you are on the "Dresses" category page and have selected to filter further by selecting "Shop the Look". One would expect that the new results are `Dresses` that are also within the `Shop the Look` category. What actually happens is that the server returns a union of all products in both categories. You end up with results like belts/accessories that happen to be in the `Shop the Look` category even though you are on the `Dresses` category page.

I thought one solution would be to take advantage of both the `eq` and `in` category filter through the following code, but it does not work:

```js
        const nextCategory = {
            eq: String(id),
            in: [String(id)]
        };
        if (newFilters['category_id']) {
            if (newFilters['category_id'].eq) {
                // If a single filter has been specified, move it to the `in` prop
                nextCategory.in.push(newFilters['category_id'].eq);
            } else if (newFilters['category_id'].in) {
                // Otherwise spread the selected categories.
                nextCategory.in.push(...newFilters['category_id'].in);
            }
        }

        // Replace the category filter with our modified filter.
        newFilters['category_id'] = nextCategory;
```

That results in a query with `category_id = { eq: "28", in: ["28", "11", "14"] }` for example, but it does not work.

I believe our only solution, without turning this ticket into a major refactor of the UX of filters, is to hide category filters entirely.

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -52,28 +75,66 @@ export const useSearchPage = props => {
...
-        apolloQueryVariable = {
-            ...apolloQueryVariable,
-            categoryId
-        };
-    }
-
-    const { loading, error, data } = useQuery(query, {
-        variables: apolloQueryVariable
-    });
+    useEffect(() => {
```
sirugh: This entire effect is basically a copy pasta from the category page effect. A lot of the code changes in this file are. I could extract a good amount of this into a custom hook that would only require a few inputs but I think it's more important to get feedback on the approach before optimizing.

packages/peregrine/lib/talons/SearchBar/useSuggestedCategory.js
```diff
@@ -1,13 +1,16 @@
...
     // start with the current uri
     const uri = new URL('/search.html', window.location);
 
     // update the query params
     uri.searchParams.set('query', searchValue);
-    uri.searchParams.set('category', categoryId);
+    uri.searchParams.set(
+        'category_id[filter]',
+        `${label}${DELIMETER}${categoryId}`
+    );
```
sirugh: Now that the search page is also relying on the filter modal for the category state, this is tightly coupled to the format expected by the filter components/state. The more I worked on this the more I wished we had a single context for interacting with it. I know that the true state is supposed to be held in the URL but I feel like maybe they should be synchronized.

packages/peregrine/package.json
```diff
@@ -32,6 +32,7 @@
     "@apollo/react-hooks": "~3.1.2",
     "@babel/runtime": "~7.4.2",
     "apollo-client": "~2.6.4",
+    "graphql-tag": "~2.10.1",
```
sirugh: Now we have some query code written directly in talons so this is a peer dep.

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -65,9 +61,8 @@ const SearchPage = props => {
         <article className={classes.root}>
             <div className={classes.categoryTop}>
                 <div className={classes.totalPages}>
-                    {`${total_count} items`}
+                    {`"${inputText}" - ${total_count} items`}
```
sirugh: Since we delete the search term in the input box after searching (#2150) I humbly propose this little helper for the users.

packages/venia-ui/lib/RootComponents/Category/category.js
```diff
@@ -36,14 +55,60 @@ const Category = props => {
...
+
+        const filters = getFiltersFromSearch(search);
+
+        // Construct the filter arg object.
+        const newFilters = {};
+        filters.forEach((values, key) => {
+            newFilters[key] = getFilterInput(values, filterTypeMap.get(key));
+        });
+
+        // TODO: Category filtering through aggregations seems incorrect. The result from a query using the `in` filter type results in a union of all results in all selected categories. There is no way to get a union of all results AND an intersection with the current category. For now, just search using the current page's category id.
```
sirugh: If we care about category filtering when on a category page we should turn this into a ticket, otherwise I can update this comment to reflect what is going on and why.

cc @awilcoxa @soumya-ashok @schensley

packages/venia-ui/lib/queries/productSearch.graphql
sirugh: @jimbo it's the first time I've personally had to write a query on the client that uses a custom type instead of one of the natives such as `Int` or `String`. I just felt weird writing it - like if the server changed the type name the client side would break.

I had to do some custom type matching as well for the individual filter input types and their expected props, which you'll see in the helper.js file.

I'd prefer to use an `Object` type is all I'm saying, but I don't think that exists.

packages/peregrine/package.json
sirugh: The queries in question are runtime (so a client can change filters without having to redeploy) introspection queries. I can try moving them to the components though.

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -37,20 +29,17 @@ const SearchPage = props => {
...
+
     const { products } = data;
-    const { filters, total_count, items } = products;
+    const { aggregations: filters, total_count, items } = products;
 
     if (items.length === 0) {
         return <div className={classes.noResult}>No results found!</div>;
     }
 
-    const maybeCategoryFilters = categoryId ? (
```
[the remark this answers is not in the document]
sirugh: Yes - now that the filter modal is functional on the search page it didn't make sense to _just_ present the category pills and none of the other applied filters. The component would also require a refactor to work with JUST the category filters and if we're going to refactor it I would want to also get the other filters in there as individual pills, kind of like we do with the pills in the filter modal, though those are directly tied to `filterState` and not the url.

I can see a future where we have a component that reads/interacts with the search state (in the url/etc) and renders pills for all applied filters, but I didn't want to scope creep this PR any more.

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -8,25 +8,19 @@ import Gallery from '../Gallery';
...
-import CategoryFilters from './categoryFilters';
 import defaultClasses from './searchPage.css';
 import PRODUCT_SEARCH from '../../queries/productSearch.graphql';
+import FilterIntrospectionQuery from '../../queries/introspection/filterIntrospectionQuery.graphql';
 
 const SearchPage = props => {
     const classes = mergeClasses(defaultClasses, props.classes);
 
     const talonProps = useSearchPage({
+        FilterIntrospectionQuery,
```
jimbo: Sorry to nitpick, but the naming of these things is inconsistent.

- By convention, props are camelCase
- In this same file, imported queries are ALL_CAPS

```es6
// option A:
// fully name query props when there's more than one
useSearchPage({
  filterIntrospectionQuery: FILTER_INTROSPECTION,
  productSearchQuery: PRODUCT_SEARCH
})

// option B:
// define `queries` outside the component, or memoize with `useMemo`
const queries = {
  FILTER_INTROSPECTION,
  PRODUCT_SEARCH
}

useSearchPage({ queries })
```

My preference here would be option B.

sirugh: Oops you are right - I got myself into the practice of using a different style in the cart page. I'll fix these :)

packages/venia-ui/lib/RootComponents/Category/category.js
```diff
@@ -36,14 +44,60 @@ const Category = props => {
 
     const [runQuery, queryResponse] = useLazyQuery(GET_CATEGORY);
     const { loading, error, data } = queryResponse;
+    const { search } = useLocation();
+
+    // Keep track of the search terms so we can tell when they change.
+    const [previousSearch, setPreviousSearch] = useState(search);
```
jimbo: We know when search terms change because `useLocation` will receive a new value from context, triggering a render cycle for this component. In other words, `useLocation` is the single source of truth for search params.

Ordinarily, if we want to store a previous value for some comparison purpose, [we use a ref for that with `useRef`](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/FilterModal/useFilterModal.js#L23). That way, when the source of truth changes (`useLocation`), a render is triggered and we update the ref as an effect, but since a ref update does not trigger a second render, we avoid unnecessary renders (or worse, an infinite loop).

Do we actually require a second render cycle here, necessitating `useState` over `useRef`?

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
jimbo: I don't think it's too early to create a generic hook in `peregrine/lib/hooks` called `useFilterIntrospection`, as long as all it does is the work from L54-L76:

- run the query, which it receives as a prop: `query`
- log an error, if present
- return a memoized `filterTypeMap`

It probably wouldn't even need to return an object; it could return the map directly.

packages/venia-ui/lib/RootComponents/Category/category.js
sirugh: Good point! I copied this from the search page which was pretty much [just added](https://github.com/magento/pwa-studio/pull/2037/files#diff-74df0db583d1513b2f2ae327a4776babR38). I'll fix it in both places.

packages/venia-ui/lib/RootComponents/Category/category.js
sirugh: Oops! The comment was not meant to remain. Any `TODO` comments I add while working are an indicator for myself to get back to them. In this case I actually turned price and category filtering into two new tickets:

https://jira.corp.magento.com/browse/PWA-403
https://jira.corp.magento.com/browse/PWA-404

I'll replace the comment with a reference.

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
sirugh: If I was to make a special hook who would own the query? Would all components that use the hook need to know to pass that specific query? Seems like a good case for peregrine owning `gql` as a dependency 😆

jimbo: Yes, components in `venia-ui` would import/define the query and pass it to the `peregrine` hook. If we ever decide to move our queries into `peregrine`, it'll be easy to streamline hooks & talons by putting them as default arguments, while still leaving them open to custom queries. IMO.

sirugh: In this specific case the `useFilterIntrospection` hook would have logic directly tied to the graphql fields, which just makes every component required to pass the _same_ query. In that case why not just have the query live in the hook?


This is the code that is used to generate the type map. It checks `__type.inputFields.name`:

```js
    const filterTypeMap = useMemo(() => {
        const typeMap = new Map();
        if (introspectionData) {
            introspectionData.__type.inputFields.forEach(({ name, type }) => {
                typeMap.set(name, type.name);
            });
        }
        return typeMap;
    }, [introspectionData]);
```

packages/peregrine/lib/talons/FilterModal/helpers.js
```diff
@@ -45,7 +48,17 @@ export const getStateFromSearch = (initialValue, filterKeys, filterItems) => {
 
             // map item values to items
             for (const value of params.getAll(key)) {
-                items.add(groupItemsByValue.get(value));
+                const existingFilter = groupItemsByValue.get(
+                    value.split(DELIMITER)[1]
```
jimbo: Out of curiosity, didn't you define a helper for this later in the file?

packages/peregrine/lib/talons/FilterModal/useFilterModal.js
```diff
@@ -26,32 +31,76 @@ export const useFilterModal = props => {
...
+    // that the api will understand.
+    const possibleFilters = useMemo(() => {
+        let nextFilters;
+        if (introspectionData) {
+            nextFilters = introspectionData.__type.inputFields
+                .map(field => field.name)
+                .filter(filterName =>
+                    filters
+                        .map(filter => filter.attribute_code)
+                        .includes(filterName)
```
jimbo: This is a simple functional style, but recall what the original comment here read:

```es6
// iterate over filters once to set up all the collections we need
```

One of the reasons I encourage us to use `for...of` instead of `Array` prototype methods in most places is that it encourages us to combine operations into a single loop. (Another reason is it encourages us to go ahead and use `Set` and `Map` more.) As written, there are a bunch of nested loops happening, which is pretty inefficient. Let's try to improve it.

```es6
const DISABLED = new Set(["category_id", "price"])

const Foo = () => {
  /* ... */
  
  const inputFields = introspectionData
    ? introspectionData.__type.inputFields
    : []

  // precalculate attribute codes, since that's all we need
  // avoid repeating this calculation for each input field
  const attributeCodes = useMemo(
    () => filters.map(({ attribute_code }) => attribute_code)
    [filters]
  )

  // determine which input fields are relevant
  const possibleFilters = useMemo(() => {
    const nextFilters = new Set()

    // perform mapping and filtering in the same cycle
    for (const { name } of inputFields) {
      const isValid = attributeCodes.includes(name)
      const isEnabled = !DISABLED.has(name)

      if (isValid && isEnabled) {
        nextFilters.add(name)
      }
    }

    return nextFilters
  }, [attributeCodes, inputFields])

  /* ... */
}
```

Generally, you can get it down to a couple of loops. Precalculate maps and other structures you can use in the main loop for lookups, then do the bulk of the work in that one main loop.

Also, please restore the original comment above L76. 🙂

packages/peregrine/lib/talons/FilterModal/helpers.js
sirugh: Nice observation! I'll rename and re-use.

packages/peregrine/lib/talons/FilterModal/useFilterModal.js
sirugh: DONE!

sirugh:
After speaking with @jimbo we realized we can just use the search term for the "static" filter query on the search page.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -62,9 +71,9 @@ export const useAutocomplete = props => {
 
     return {
         displayResult,
+        filters: filterResult ? filterResult.products.aggregations : null,
         messageType,
         products,
-        queryResult,
```
sirugh: Unused.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
```diff
@@ -15,23 +15,30 @@ import debounce from 'lodash.debounce';
...
     const { value } = useFieldState('search_query');
 
     // Create a debounced function so we only search some delay after the last
     // keypress.
     const debouncedRunQuery = useMemo(
         () =>
             debounce(inputText => {
-                runQuery({ variables: { inputText, filters: {} } });
+                runProductSearch({ variables: { inputText, filters: {} } });
+                runFilterSearch({ variables: { search: inputText } });
```
sirugh: If we care we can merge these two back into a single query though I would write a new one and not double up on the `productSearch` query.

jimbo: Yes, I think we should merge these into one query specific to this component.

dpatil-magento:
QA Pass, but need review approval to get merged.

packages/peregrine/lib/talons/SearchBar/useAutocomplete.js
sirugh: Done!

dpatil-magento:
@sirugh  Only for first time when search, I see "No results found" for a sec and then results load. 

<a href="https://gyazo.com/4d0d495e6980b71dfc4d6cfddbb8b787">Image from Gyazo</a>

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -172,6 +179,13 @@ export const useSearchPage = props => {
     // default (and potentially changing!) aggregations from the product query.
     const filters = filterData ? filterData.products.aggregations : null;
 
+    // Avoid showing a "empty data" state between introspection and search.
+    const loading =
+        (introspectionCalled && !searchCalled) ||
+        (searchCalled && !data) ||
```
sirugh: This line may not be necessary. Testing

dpatil-magento:
QA pass, goot to merge after review approval.

PR #2167, opened 2020-02-13
[PWA-376] Cart page items show correct configurable options

tjwiebell:
## Description

Configurable Products listed in the MiniCart and new Cart page are showing different values for their configured properties (ex. Fashion Size, Fashion Color).

**RCA:** `configurable_options.id` is not unique based on selected values, so Apollo was not caching any other options after the first.

## Related Issue

* [[PWA-376](https://jira.corp.magento.com/browse/PWA-376)] Cart page items show correct configurable options

### Verification Steps

1. Add two configurable products to your cart with different selected options
2. Navigate to `/cart` page and verify that options are correctly displayed

Screen Shot 2020-02-13 at 10 12 10 AM

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
@@ -21,6 +21,7 @@ export const ProductListingFragment = gql`
             quantity
             ... on ConfigurableCartItem {
                 configurable_options {
+                    id
                     option_label
                     value_label
                 }
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
@@ -23,6 +23,7 @@ export const ProductListingFragment = gql`
                 configurable_options {
                     id
                     option_label
+                    value_id
                     value_label
                 }
             }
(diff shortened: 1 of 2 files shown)


packages/venia-ui/lib/util/apolloCache.js
```diff
@@ -42,7 +43,8 @@ export const cacheKeyFromType = object => {
             return object.url_key
                 ? `${MagentoGraphQLTypes.ProductInterface}:${object.url_key}`
                 : defaultDataIdFromObject(object);
-
+        case MagentoGraphQLTypes.SelectedConfigurableOption:
```
sirugh: I think a comment here explaining why we have to do this is warranted.

sirugh: Also looks like we are getting `value_id` from the `getCartDetails` query. I'm not sure that this is just some magical thing we have. We may want to make sure we add it to the product details query within the `CartPage/` components

tjwiebell: Man, good catch! I had assumed Apollo automatically injected this key, but I was mistaken. The bug appeared to be fixed because we weren't caching any options, because `object.value_id` was `undefined` 🤔

I've now fixed this up in a187f13  so it uses that field if available, and explicitly won't cache if its left out.

packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
```diff
@@ -23,6 +23,7 @@ export const ProductListingFragment = gql`
                 configurable_options {
                     id
                     option_label
+                    value_id
```
sirugh: Silly enough - we now have two "required" fields. `id` per validation although it's not actually guaranteed to be unique and `value_id` because we use it for caching though you'd find no code within `CartPage/` that uses it.

PR #2169, opened 2020-02-13
Fix RootComponentsPlugin to be IE compatible

### LINKED ISSUE #675: IE11 Support

- [X] `venia-concept`
- [ ] `pwa-buildpack`
- [X] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

## This issue is a:

- [X] Bug
- [ ] Feature suggestion
- [ ] Documentation issue
- [ ] Other (Please Specify)

### Environment

| Operating System + version  | windows 10 IE11 |

## Description

As a merchant, I expect my site to work in all modern browsers + IE11 -- especially US based merchants.

I understand that "technically" IE11 falls outside of the list of supported browsers according to https://github.com/magento-research/pwa-studio/wiki/Definition-Of-Quality#browser-support but that's based on global usage which heavily skews mobile and B2C. Inside the US, IE11 is still >5% http://gs.statcounter.com/browser-market-share/all/united-states-of-america

Additionally, many B2B customers/merchants unfortunately deal with a significant user base still on IE11 (we have a client >15% IE11). Even in 2019, lack of IE11 is a complete turnoff for many merchants. Is there at least some basic level of support we can add so that it's not just a blank screen... I feel like the current state of Venia in IE11 is not ideal since it just shows a blank screen?

## Expected result:

PWA does not even load in IE11:

screen shot 2018-12-17 at 6 06 55 pm

## Possible solutions:

We might need to add a missing polyfill perhaps?  Or changing babel browsers to include >5% + IE11

luke-denton-aligent:
## Description
This PR fixes a part of the buildpack that prevents PWA studio from rendering the most basic of applications. I changed my `index.js` to look like this:

```
import React from 'react';
import ReactDOM from 'react-dom';

ReactDOM.render(
    <div>This is react!</div>,
);

```

Even with this super simplified render, there is still an error displayed in IE11.

**The reason**
Within `RootComponentsPlugin`, as the final string (`importerFactory`) is put together by calling `toString()` on variables that are functions, this means that the function itself is converted to a string, and then concatenated to the `importerFactory` variable. The problem with this is that it's _not_ transpiled, which means arrow functions and default parameters were making it to IE, causing the white screen of death.

With the fixes proposed in this PR, the very simplified application above will render in IE11 as expected.

Re-introducing all of the Venia components, however, will cause further errors in IE11, and the white screen of death again, so with that in mind, this doesn't fix **all** problems with IE, but it does at least fix the build pack, so users can build their own React application from scratch, using PWA Studio, and have it work in IE.

## Related Issue
Closes #675 . (doesn't actually close, this is just related - CI doesn't like me changing "Closes" to "Related". Please let me know if I should create a new issue instead of linking to this one)

### Verification Steps
Here are some screenshots showing before and after the changes are made to the `RootComponentsPlugin` file. As we can see in the After screenshot, with the updates, we're able to actually render content to the browser in IE

Before
image

After
image

## Checklist


packages/pwa-buildpack/lib/WebpackTools/plugins/RootComponentsPlugin.js
```diff
@@ -7,12 +7,16 @@ const micromatch = require('micromatch');
 
 const prettyLogger = require('../../util/pretty-logger');
 
-const toRootComponentMapKey = (type, variant = 'default') =>
-    `RootCmp_${type}__${variant}`;
+const toRootComponentMapKey = function(type, variant) {
+    variant = variant || 'default';
+    return 'RootCmp_' + type + '__' + variant;
```
jimbo: I'm surprised the linter didn't complain, but we should avoid mutating arguments. I suggest deleting L11 and revising L12.

```suggestion
    return 'RootCmp_' + type + '__' + (variant || 'default');
```

packages/pwa-buildpack/lib/WebpackTools/plugins/RootComponentsPlugin.js
```diff
@@ -159,10 +163,10 @@ class RootComponentsPlugin {
                                 );
                                 importerSources[
                                     key
-                                ] = `() => import(/* webpackChunkName: "${key}" */'./${relative(
+                                ] = `function () { import(/* webpackChunkName: "${key}" */'./${relative(
```
jimbo: This function should *return* the result of `import()`, as the arrow function did.

luke-denton-aligent:
@jimbo Thank you for your review! The return import is certainly something I missed 😬, and thanks for the suggestion regarding mutating arguments, it's been so long since I've had to do that I wasn't sure the best approach 😅.
The changes have now been mae, if you could please review those

luke-denton-aligent:
P.s. @jimbo, while the final code changes were few, it still took a bit of debugging, time and knowledge to figure out that this is the change required, so lets not discount the effort that has gone in to coming up with this simple fix :)

sirugh:
https://jira.corp.magento.com/browse/PWA-397

dpatil-magento:
No impact on other browsers, looks good.

PR #2178, opened 2020-02-20
Update page layout so footer will be pushed to bottom of viewport

### LINKED ISSUE #2177: [bug]: Footer gets pushed out of viewport even if page content is short
**Describe the bug**
Regardless of the height of a pages' content, the footer will never be within the viewport.

**To reproduce**
1. Go to 'https://venia.magento.com/'

**Expected behavior**
As the page's content is small, just a simple list of categories, I would expect to see the footer also within the viewport, but it's completely out of view, and I have to scroll to be able to see it

**Screenshots**
image

**Additional context**
The reason this issue is occuring is due to the `.page` element being given a `min-height` of `100vh`. Viewport units are great, but they should be used very carefully, because they have unintended consequences, one of which I believe this is.

**Possible solutions**
The fix is to go back to the "old" way of pushing a footer to the bottom of the page: `height: 100%` on `html`, `body` and any parent containers (in this case, `#root` as well)

 - All Devices
 - All OS's
 - All browsers
 - PWA Studio Version: 5

- [x] `venia-ui`

luke-denton-aligent:
## Description

This PR fixes an issue with the page content being forced to a certain height, meaning the footer will never be within the viewport on page load, even if the page being loaded is very short.
With this update, if the page's content isn't long enough to push the footer, then it will place itself at the bottom of the viewport, otherwise it will be placed at the end of the pages content. This ensures there is never "leading/trailing whitespace" around the footer

## Related Issue
Closes #2177.

### Verification Steps
1. Go to the hoempage.
2. Verify the footer is visible.

**Before:**
image

**After:**
image

**With long content:**
image

### DIFF
--- a/packages/venia-concept/src/index.css
+++ b/packages/venia-concept/src/index.css
@@ -35,6 +35,7 @@ html {
     line-height: 1;
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
+    height: 100%;
 }
 
 body {
@@ -42,6 +43,7 @@ body {
     color: rgb(var(--venia-text));
     margin: 0;
     padding: 0;
+    height: 100%;
 }
 
 html[data-scroll-lock='true'],
@@ -115,4 +117,8 @@ button {
 button:disabled {
     cursor: default;
     touch-action: none;
+}
+
+:global(#root) {
+    height: 100%;
 }
--- a/packages/venia-ui/lib/components/Main/main.css
+++ b/packages/venia-ui/lib/components/Main/main.css
@@ -3,6 +3,9 @@
     color: rgb(var(--venia-text));
     position: relative;
     z-index: 1;
+    height: 100%;
+    display: grid;
+    grid-template-rows: auto 1fr auto;
 }
 
 .root_masked {
@@ -10,7 +13,6 @@
 }
 
 .page {
-    min-height: 100vh;
 }
 
 .page_masked {


jimbo:
Hi Luke. As I mentioned in the associated issue, this is one of those "working as intended" situations. I understand how that could be counterintuitive in this case. 😅 

The thing is, we don't actually *want* the site footer to take up half the page when the page is low on content. More importantly, we don't want there to be empty space below the footer. And of course, we don't want to set `height: 100%` on any of our container elements here (such as `html`, `body`, or `#root`). Perhaps the current `min-height: 100vh` on `.page` is a bit heavy-handed, but it does meet our criteria for a solution at the moment.

If we want to say that the current solution is too extreme, though, let's look at what an acceptable alternative might look like.

```css
/* components/Main/main.css */
.root {
  display: grid;
  grid-template-rows: auto 1fr;
  min-height: 100vh;
}

.page {
  min-height: 50vh;
}
```

In this example, `Main` becomes a grid with three rows, and the middle row (`.page`) flexes to consume all of the grid's remaining space after the header and footer have been allocated. Since the grid is now the one to have `min-height: 100vh`, we ensure there can't be space below the footer—and we do so without fixing any heights to `100%`.

Meanwhile, reducing the `min-height` on `.page` to `50vh` could help achieve your goal while still ensuring the footer doesn't dominate the viewport. What do you think?

luke-denton-aligent:
Hi @jimbo,

I think that's a much better response to the one I got on the Issue, thank you actually giving an explanation. As for Magento's design decisions, that's totally valid, but I'm coming from the perspective of a developer who is going to be creating a PWA using `create-pwa` for clients. As you'll be aware, the script essentially duplicates the Venia concept implementation, so while Magento may not want to see the footer, is that going to be the consensus among all agencies/clients that are building using Venia as a base? That's the real question, and that's what I think Venia should be built to (i.e. built to what the majority of the industry wants, which may be to hide the footer, or may be to show the footer).

There are bugs when using `100vh`, but I'm not sure how applicable they'll be to PWA Studio based apps (specifically that applying `100vh` doesn't factor in the height of the browsers URL bar on mobile, so there are funky things that happen when the user scrolls, and the URL bar auto hides). Other than that, I do like the Grid approach that you've suggested.

As a side note, I am new to CSS grid, so I may be missing something, but doesn't `grid-template-rows: auto 1fr;` only set up 2 rows? Should it be `grid-template-rows: 1fr auto 1fr;`, which will correspond with `header` `content` `footer`?

awilcoxa:
Closing based on comments in issue/pr

luke-denton-aligent:
Hi @awilcoxa,

There is clearly an unanswered question that is awaiting a response, how will that question be answered if the PR is closed?

The community communication really needs some work.

PR #2181, opened 2020-02-20
[PWA-181] Venia checkout skeleton

revanth0212:
## Description

Venia checkout page skeleton. This PR creates the file system and the basic state needed to render the checkout page.

In this PR I have created a new utility called `renderIf`. It is a new methodology in our code base. Let me know what you guys think of it.

## Related Issue
Closes [PWA-181](https://jira.corp.magento.com/browse/PWA-181)

### Verification Steps
1. Go to `/checkout` on the deployed app to see the checkout page.

Desktop Checkout Page
image

Desktop Confirmation Page
image

Mobile Checkout Page
image

Mobile Confirmation Page
image

## Checklist
None


packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -0,0 +1,38 @@
...
+export const useCheckoutPage = () => {
+    const [, { toggleDrawer }] = useAppContext();
+    const [{ isSignedIn }] = useUserContext();
+    const [{ isEmpty }] = useCartContext();
+
+    /**
+     * Using local state to maintain these booleans. Can be
+     * moved to checkout context in the future if needed.
+     */
+    const [shippingInformationDone, setShippingInformationDone] = useState(
```
sirugh: One way to tell if shipping information/method/etc are done is to just make a query for the data and see if anything returns. In the case of shipping information we will probably want to do something like query for shipping address as the estimator on cart does but then wipe the cache of any of that fake data.

I'd definitely prefer that we allow the apollo cache to control the "renderable" state of things. For example:

* checkout page would query for shipping information (along with other info such as shipping method, pre-added gift card/coupons/etc) which dictate what step to display.
* shipping address component would query for any provided cart shipping address which, if found, would tell it to render in the "card" UI rather than the form UI.

As it is, on a page refresh you lose local state which means you would have to start over each time when in reality part of your data would have already been saved to gql.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -0,0 +1,38 @@
...
+        false
+    );
+    const [shippingMethodDone, setShippingMethodDone] = useState(false);
```
sirugh: In the same vein as my above comment, the checkout page itself would just query for selected shipping method, and if found, would know that the current step should open to the payment input form.

packages/venia-ui/lib/components/CheckoutPage/utilities.js
```diff
@@ -0,0 +1,22 @@
...
+const renderComponent = Comp => {
+    return isElement(Comp) ? Comp : <Comp />;
+};
+
+/**
+ * If an Else components can be React Components or Elements.
+ *
+ * React components are useful for lazy rendering.
+ */
+export const renderIf = booleanCondition => (IfComp, ElseComp) => {
```
sirugh: This is a really interesting approach to conditional rendering. Currently, throughout the app, we have code that uses memoization that returns the component. This allows us to also maintain readability in the JSX markup. Example:

```js
const maybeRenderComponent = useMemo(() -> {
  if (condition) {
    return <SomeComponent />
  }
  else {
    return null;
  }
}, [condition])

return (
  <div>
    {maybeRenderComponent}
    <SiblingComponent />
    <span>Some text</span>
  </div>
);
```

This is all to say that while I think this is a neat idea, I'm not sure we want to introduce this approach. @jimbo will likely have input here.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,190 @@
...
+    );
+
+    return (
+        <div className={classes.root}>
+            <Title>{`Checkout - ${STORE_NAME}`}</Title>
+            {/**
+             * Lazy rendering checkout page components only
+             * if the cart is not empty.
+             */
+            renderIfCartNotEmpty(
```
sirugh: Instead of using this conditional render, this should be a lazy query, triggered off `cart_id`, for `total_quantity`. Check out how `CartPage` does it.

packages/venia-ui/lib/components/App/renderRoutes.js
```diff
@@ -1,21 +0,0 @@
...
-import { Switch, Route } from '@magento/venia-drivers';
-import { Page } from '@magento/peregrine';
-import ErrorView from '../ErrorView/index';
-
-const CreateAccountPage = lazy(() => import('../CreateAccountPage/index'));
-const Search = lazy(() => import('../../RootComponents/Search'));
-
-const renderRoutingError = props => <ErrorView {...props} />;
-
-const renderRoutes = () => (
```
sirugh: We weren't using this at all? I wonder when we moved away from it and why it didn't get removed at that time.

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.css
```diff
@@ -0,0 +1,9 @@
+.textContent {
+    padding: 25px;
+    padding-top: 50px;
+    font-weight: bold;
+}
+
+.proceedButton {
+    padding: 25px;
```
sirugh: We should be using `rem` units. Not super important as this'll be modified when we implement these components, but wanted to point it out.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,190 @@
...
+        <div className={classes.signin_container}>
+            <div className={classes.heading_container}>
+                <h1 className={classes.heading}>Quick Checkout</h1>
+            </div>
+            <Button
+                className={classes.sign_in}
+                onClick={handleSignIn}
+                priority="high"
+            >
+                {'Login to Checkout Faster'}
```
sirugh: In the mocks this is a button that says "Sign In". Did you get UX feedback that differs from the mocks for this?

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.css
```diff
@@ -0,0 +1,99 @@
...
+.signin_container {
+    border-bottom: 1px solid rgb(var(--venia-border));
+    margin-bottom: 1.5rem;
+}
+
+.section_container {
+    min-height: 150px;
+    border: 1px solid rgb(var(--venia-border));
+    border-radius: 5px;
+    padding: 25px;
```
sirugh: Same idea here with using `rem`.

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.js
```diff
@@ -0,0 +1,30 @@
+import React, { useCallback } from 'react';
+
+import Button from '../../Button';
+
+import { renderIf } from '../utilities';
+
+import defaultClasses from './shippingMethod.css';
+
+export default ({ onSave, doneEditing }) => {
```
sirugh: This'll get changed when someone implements this but we don't destructure props in the signature for function components anywhere else in this repo.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,190 @@
...
+                             * Lazy rendering Shipping method component
+                             * only if shipping information component is
+                             * done editing.
+                             */
+                            renderIfShippingInformationDone(() => (
+                                <Fragment>
+                                    <div
+                                        className={`${
+                                            classes.shipping_method_container
+                                        } ${classes.section_container}`}
```
sirugh: Avoid concatenating classes like this. In `checkoutPage.css` you should compose the shipping method container class with the `section_container`

```css
.shipping_method_container {
  composes: section_container;
  //...other css specific to shipping method container
}
```

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,190 @@
...
+                                     * Lazy rendering Payment information and price
+                                     * adjustments components if the shipping method
+                                     * component is done editing.
+                                     */
+                                    renderIfShippingMethodDone(() => (
+                                        <Fragment>
+                                            <div
+                                                className={`${
+                                                    classes.payment_information_container
+                                                } ${classes.section_container}`}
```
sirugh: Same idea here re: composition.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,190 @@
...
+                                                <PriceAdjustments />
+                                            </div>
+                                        </Fragment>
+                                    ))}
+                                </Fragment>
+                            ))}
+                            <div className={classes.summary_container}>
+                                <div
+                                    className={`${classes.summary_contents} ${
+                                        classes.section_container
```
sirugh: Same idea here re: composition

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: Din't know about this approach. Nice. Will do that.

packages/venia-ui/lib/components/CheckoutPage/utilities.js
jimbo: @sirugh is correct. As we've discussed a few times before, in this codebase, we avoid utilities. The only utility our components use is `mergeClasses` (because we have to), but even that one is something we're aiming to eliminate.

I've worked in codebases where utility code was common. I've also used Ramda to create some beautiful functional code. Ultimately, that approach is not the right one for this project, and more importantly, it's not the one we selected in the first place. As a general rule of thumb, please write code that looks like the other code in the codebase. 😅

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.js
revanth0212: Yeah, I was aware of that. Missed to change this. Should have changed it. Will do 👍

revanth0212:
>@sirugh is correct. As we've discussed a few times before, in this codebase, we avoid utilities. The only utility our components use is mergeClasses (because we have to), but even that one is something we're aiming to eliminate.

>I've worked in codebases where utility code was common. I've also used Ramda to create some beautiful functional code. Ultimately, that approach is not the right one for this project, and more importantly, it's not the one we selected in the first place. As a general rule of thumb, please write code that looks like the other code in the codebase. 😅

I couldn't find examples of the use cases I have encountered in this PR in our code base. So instead of writing if-else cases in the render method, I thought of writing a `renderIf` structure which I felt was more readable. Obviously, it isn't straight forward and I am still trying to clean it up. I wasn't aware of the decision not to use any utilities, I just thought we don't want to use third-party utilities from npm. Good to know.

What do you think is the right approach in this case where each block of UI is dependent on some boolean condition?

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: I thought of leaving that to the individual taking up the respective checkout tickets.

packages/venia-ui/lib/components/App/renderRoutes.js
revanth0212: Not sure dude, that's why I was asking about it in the chat the other day.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: Yeah, the mocks are old. @soumya-ashok showed me the latest mocks which are not updated in the ticket yet.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -0,0 +1,33 @@
+import { useCallback, useState } from 'react';
+
+import { useAppContext } from '../../context/app';
+import { useUserContext } from '../../context/user';
+import { useCartContext } from '../../context/cart';
+
+export const useCheckoutPage = () => {
+    const [, { toggleDrawer }] = useAppContext();
+    const [{ isSignedIn }] = useUserContext();
+    const [{ isEmpty }] = useCartContext();
```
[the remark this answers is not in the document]
revanth0212: Good to know. For now, I will leave it to the individual working on the ticket to make the changes. The intent of this PR is to scaffold checkout.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: You are absolutely right about losing data on refresh which is not what we want. We can use apollo client local queries for that. Also, these boolean flags here are used to track progress, not data presence. Even if you have shipping method picked in cart page, and that data is available at the time of checkout, you still need them to click on proceed to payment information but use the data from prev step or from the account (shipping address for instance). This is what I understood after I spoke with @soumya-ashok a couple of days ago. That is the reason I have those progress indicator flags. They have to be moved to a persistent store, but for now, since this is just the scaffolding PR, I have left them in the hook.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Check my reply to the above comment.

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.css
revanth0212: Done 👍

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: >  Also, these boolean flags here are used to track progress, not data presence.

Right, I'm just implying that data presence itself is an indicator of the progress a user has made so there shouldn't be a need to re-submit it, which is what those "proceed to X" buttons are for. 

Assume a user has entered (and saved to the server) a bunch of information on their cart and they're on some final step like payment method. They've clicked "proceed to X step" a bunch of time at that point. If they refresh the page, for whatever reason, they will have to click through all those steps again, likely resubmitting redundant data that the server is already aware of.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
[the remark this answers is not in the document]
sirugh: I'm not sure there's a follow-up ticket for checkout refactor but I definitely don't want to scope creep you in this ticket. I'll make a ticket and reference this comment.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
sirugh: Yea, I just thought this was relevant to the initial bare bone state :P Regardless, made a ticket to follow up.

sirugh:
> What do you think is the right approach in this case where each block of UI is dependent on some boolean condition?

This is what I was alluding to in my comment: https://codesandbox.io/s/agitated-thompson-ehk3h?fontsize=14&hidenavigation=1&theme=dark

revanth0212:
> > What do you think is the right approach in this case where each block of UI is dependent on some boolean condition?
> 
> This is what I was alluding to in my comment: https://codesandbox.io/embed/agitated-thompson-ehk3h?fontsize=14&hidenavigation=1&theme=dark
> 
> @jimbo may have a better suggestion but this is the general strategy I use.

Thanks for the demo @sirugh 

One issue I see with this approach is the ambiguity of steps. Looking at step number it is hard to grasp what step that number belongs to. It would take more effort If someone wants to change or remove a certain step.

I had a talk with @jimbo and make changes to the code base. Let me know what you think?

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: You are right. Let's see what Soumya and Scott have in mind.

sirugh:
>One issue I see with this approach is the ambiguity of steps. Looking at step number it is hard to grasp what step that number belongs to. It would take more effort If someone wants to change or remove a certain step.

Right, the difference in that regards from the example code and your code with the `renderIf` function was that I defined the condition within the memo rather than outside. The step/memo is self-aware of its render condition.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,190 @@
...
+    const { classes, isGuestCheckout } = props;
+
+    return (
+        <div className={classes.signin_container}>
+            <div className={classes.heading_container}>
+                <h1 className={classes.heading}>
+                    {isGuestCheckout ? 'Guest Checkout' : 'Checkout'}
+                </h1>
+            </div>
+            <h3>There are no items in your cart.</h3>
```
sirugh: Empty cart feels sparse. Surprised there isn't a call to action like "Continue Shopping".

[Image from Gyazo](https://gyazo.com/a5259e16a785d23299ab460654869f14)

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,190 @@
...
+            </h1>
+        </div>
+    );
+};
+
+const EmptyCartMessage = props => {
+    const { classes, isGuestCheckout } = props;
+
+    return (
+        <div className={classes.signin_container}>
```
sirugh: Because this re-uses the `signin_container` style we end up with an out of place bottom border. Just make a new class for the empty cart state message without the border.

[Image from Gyazo](https://gyazo.com/a5259e16a785d23299ab460654869f14)

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -0,0 +1,95 @@
...
+
+    const [, { toggleDrawer }] = useAppContext();
+    const [{ isSignedIn }] = useUserContext();
+    const [
+        { isEmpty },
+        { createCart, getCartDetails, removeCart }
+    ] = useCartContext();
+
+    const [fetchCartId] = useMutation(createCartMutation);
+    const fetchCartDetails = useAwaitQuery(getCartDetailsQuery);
```
sirugh: Even though we are using the redux actions for create cart (as there isn't a better alternative now) I think that fetching data for the checkout page should be done identically to how the cart page does it.

Create a lazy query and call it in an effect whenever the `cartId` changes (so long as there is a cartId).

See https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/CartPage/useCartPage.js#L13-L34

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * TODO
+     */
+    const placeOrder = useCallback(async () => {
+        await cleanUpCart();
+        updateOrderPlaced(true);
+    }, [cleanUpCart, updateOrderPlaced]);
+
+    return [
+        {
```
sirugh: While we use the `[state, api]` return for context hooks, no other talons use this API that I'm aware of.

I really do like the idea, but for now maybe we should stick to the norm? What are your thoughts @jimbo? I feel like this makes the talon APIs much more resistant to breakage.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -0,0 +1,95 @@
...
+            isGuestCheckout: !isSignedIn,
+            isCartEmpty: isEmpty,
```
sirugh: If we change to use a checkout page graphql query here we can check for `total_quantity` like we do on the cart page. If you could make that change here it'll be one less thing we have to refactor later.

See https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/CartPage/useCartPage.js#L37

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.css
```diff
@@ -0,0 +1,112 @@
...
+    margin: 0 auto;
+}
+
+.body {
+    display: grid;
+    gap: 2rem;
+    grid-template-columns: 2fr 1fr;
+    /* The summary grid item spans the entire right column. */
+    grid-template-areas:
+        'shipping_information summary'
```
sirugh: Because we have named areas here for potentially hidden nodes there is some extra blank space created by the grid:

[Image from Gyazo](https://gyazo.com/5bd4ae4c52c7e0da18830e3c7cd5d086)

I think you may be able to replace the names with `.`. The structure should be the same but it will compress the blank spaces.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,190 @@
...
+                        classes={classes}
+                    />
+                    <div className={classes.body}>
+                        <div className={classes.shipping_information_container}>
+                            <ShippingInformation
+                                onSave={setShippingInformationDone}
+                                doneEditing={shippingInformationDone}
+                            />
+                        </div>
+                        {shippingInformationDone && (
```
sirugh: Just noting that I don't personally like inline conditional renders but I'm guessing you and @jimbo already talked about it so I leave it to you guys to hash out.

revanth0212: I am not sure if we have a standard for this, but I found both `if-else` and `&&` for conditional renders in our code base.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -0,0 +1,95 @@
...
+    const setShippingMethodDone = useCallback(
+        () => updateShippingMethodDone(true),
+        [updateShippingMethodDone]
+    );
+    const setPaymentInformationDone = useCallback(
+        () => updatePaymentInformationDone(true),
+        [updatePaymentInformationDone]
+    );
+    /**
+     * TODO
```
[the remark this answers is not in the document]
revanth0212: Just a piece of code I forgot to remove.

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default () => {
+    /**
+     * Use talon to get information about items in cart
+     * to show a review of them.
+     */
+    return (
+        <div className={defaultClasses.container}>
+            <div>
+                New ticket needs to be created for Review of Items in Cart
```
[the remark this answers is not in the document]
revanth0212: There was no ticket to create this component

image

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: This was never part of the XD or the epic. So I added filler text. UX needs to come up with a design for this.

@soumya-ashok @schensley drop your thoughts here.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Like I mentioned previously, I'll leave this to the individual working on the ticket. The scope of this PR is to set up a basic state and UI.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Yeah, I was confused of this as well. I personally feel `[state, api]` is a cleaner structure. I might be wrong but react (un-officially) pushes that return structure for react hooks (both built in and custom).

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: > I'm not sure there's a follow-up ticket for checkout refactor but I definitely don't want to scope creep you in this ticket. I'll make a ticket and reference this comment.
> 
> Made this: https://jira.corp.magento.com/browse/PWA-395

Let's handle such tasks in the ticket you created last week.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
brendanfalkowski: FWIW, we settled on the more explicit ternary syntax like:

`{isAlive ? (
    <Thing />
) : null}`

Over the shorter but less familiar (I don't even know what this syntax is called):

`{isAlive && (
    <Thing />
)}`

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.js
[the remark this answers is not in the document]
sirugh: I know there was no ticket, so I made one :D

My ask here is that you update the text to reflect the new ticket as you have done for the other "to be completed by PWA-X" sections.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.css
revanth0212: UI has changed. This is not relevant anymore.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: There is no name as such in react but `&&` or `||` are short circuit evaluation elements in CS. So I believe we can call it short circuit `short circuit rendering`.

jimbo: I've covered this subject a number of times in various PRs, but I suppose the real problem is that the pattern isn't clear because we haven't been perfectly consistent in the codebase. 😅 

1. Avoid multiline statements in JSX expressions. The main reason to use JSX is because trees are very readable (like XML).
2. Avoid short-circuiting JSX elements with `&&`. The alternative to a JSX element should be `null`—not `undefined`, not `false`, and not anything else. As Brendan said, use a ternary to be explicit about it.

I've opened #2200 to correct all of the cases in the codebase where we were inconsistent about this.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,238 @@
...
+
+    const orderConfirmation = useMemo(() => {
+        if (showOrderConfirmation) {
+            return <OrderConfirmationPage />;
+        } else {
+            return null;
+        }
+    }, [showOrderConfirmation]);
+
+    const emptyCart = () => {
```
sirugh: Instead of a function, just use the conditional inline as Jimmy did in his PR.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,238 @@
...
+                </div>
+            );
+        } else {
+            return null;
+        }
+    };
+
+    const showOrderConfirmation = isCartEmpty && orderPlaced;
+
+    const orderConfirmation = useMemo(() => {
```
sirugh: Missed a memo :/

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,238 @@
...
+            }
+        } else {
+            return null;
+        }
+    };
+
+    return (
+        <div className={classes.root}>
+            <Title>{`Checkout - ${STORE_NAME}`}</Title>
+            {!isCartEmpty ? (
```
sirugh: Here's another inline conditional to fixup.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: Left it on purpose, wanted to talk to the team and jimmy about it. Removed everything else.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: Left it on purpose. It is the whole component render output. Moving it into a variable and just returning that variable seamed weird. The return state will just have that variable.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: Done. Changed.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: Done.

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+import defaultClasses from './itemsReview.css';
+
+export default () => {
```
jimbo: Here as well.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
jimbo: > Yeah, I was confused of this as well. I personally feel `[state, api]` is a cleaner structure. I might be wrong but react (un-officially) pushes that return structure for react hooks (both built in and custom).

You're correct, React does favor a `[state, api]` return signature in its own hooks. That also happens to be the reason why we selected that signature for our context hooks, such as `useAppContext`. It's a very intuitive structure.

Talons are a bit different though, so we took them in a different direction. They're more like components, in that they can do pretty much anything; they can hold state, define callbacks, and perform side effects. The only thing they can't do is create elements. In order to not confine them, and to ensure they don't appear to be generic or reusable, we gave them the most basic, consistent API possible:

```es6
propsObjectIn => propsObjectOut
```

It's not that talons are incompatible with a `[state, api]` signature, though; it's that it's too late to change the pattern (for now). In the future, we can revisit the talon APIs **as a whole** and consider whether to apply more constraints to them or reshape the signature. Just now, though, adoption of this new feature will depend on consistency, so let's stick to the existing pattern. Object in, object out, please.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -0,0 +1,207 @@
...
+            <Button
+                className={classes.sign_in}
+                onClick={handleSignIn}
+                priority="high"
+            >
+                {'Login and Checkout Faster'}
+            </Button>
+        </div>
+    );
+};
```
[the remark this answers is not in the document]
revanth0212: These are private components used just in the `CheckoutPage` component. I am not exporting them in the `index.js` file.

packages/venia-ui/lib/components/CheckoutPage/PriceSummary/index.js
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export default () => <div>Price Summary Will be handled in PWA-184</div>;
```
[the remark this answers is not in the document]
revanth0212: Yeah, these are just place holders for other devs to fill in.

packages/venia-ui/lib/components/Routes/routes.js
```diff
@@ -5,6 +5,7 @@ import { fullPageLoadingIndicator } from '../LoadingIndicator';
 import MagentoRoute from '../MagentoRoute';
 
 const CartPage = lazy(() => import('../CartPage'));
+const CheckoutPage = lazy(() => import('../CheckoutPage'));
```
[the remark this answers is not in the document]
revanth0212: That was purely coincidental 😆

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderSummary.js
```diff
@@ -0,0 +1,24 @@
...
+    return (
+        <div className={classes.order_summary_container}>
+            <div className={classes.order_summary_heading}>
+                Order Number: **************
+            </div>
+            <div className={classes.order_summary_heading}>
+                Shipping Information
+            </div>
+            <div className={classes.text}>Goosey Goose</div>
+            <div className={classes.text}>12345 Lake Ln, Austin, TX, 78759</div>
```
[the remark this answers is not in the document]
revanth0212: Good to know. I will do that.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.css
```diff
@@ -0,0 +1,137 @@
...
+        /* Only one column in mobile view. */
+        grid-template-columns: 1fr;
+        gap: 1rem;
+        grid-template-areas:
+            'shipping_information'
+            'shipping_method'
+            'payment_information'
+            'price_adjustments'
+            'summary'
+            'price_adjustments_button';
```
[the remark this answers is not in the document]
revanth0212: I followed the usage of grid css in the cart page. It had a similar layout. Thought that is how we will be doing css. Ok, will look into this.

PR #2189, opened 2020-02-24
Fixes Cart page sign out error when cart does not exist in cache

supernova-at:
## Description

This PR provides a fallback empty object (`{}`) to avoid a destructuring exception thrown when the cart does not exist in our Apollo cache.

This surfaced on the `/cart` page after the user signed out: the cache was cleared out using Apollo client's `resetStore`. But `resetStore` also causes all active queries to be refetched. `GiftOptions` has one such query that was querying for the cart in the cache. Because it was just cleared out via `resetStore`, the query got back `undefined`, which the code then attempted to destructure.

Providing the fallback empty object in the case of `undefined` returns the expected result (`undefined` values for `include_gift_receipt`, `include_printed_card`, and `gift_message`), and results in the ability to sign out from the cart page without error.

## Related Issue

Closes PWA-368.

### Verification Stakeholders

@revanth0212 

### Verification Steps

1. Sign in
1. Ensure an item is added to your cart
1. Navigate to the `/cart` page
1. Verify that you can sign out successfully without error
1. Sign in
1. Verify that your cart is restored to its previous state

## Checklist


packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions/giftOptions.js
```diff
@@ -109,11 +109,8 @@ export const giftOptionsResolvers = {
                 id: cart_id
             });
 
-            const {
-                include_gift_receipt,
-                include_printed_card,
-                gift_message
-            } = cache.data.data[cartIdInCache];
+            const { include_gift_receipt, include_printed_card, gift_message } =
+                cache.data.data[cartIdInCache] || {};
```
sirugh: I wonder what GraphQL suggests for client side caching/resolvers for this situation. We will likely run into it elsewhere - I wonder if there is some sort of "reset callback" that can reset the store to an initial state. After all, we call `resetStore` on sign out which was effectively deleting the cache, causing this to throw. Maybe `resetStore` takes a callback?

This is fine, by the way - just postulating.

sirugh: Oh hey: https://www.apollographql.com/docs/react/caching/cache-interaction/#resetting-the-store

revanth0212: Yes, I tried that a while ago when I was developing gift options. The problem is our gift options are name spaced to a certain cart id. Reset store, at that time, would not know which cart id to use to reset the store with empty gift options.

sirugh: Oh yeah @revanth0212 I forgot about that. So instead of doing it on store reset we would have to know to set that initial state whenever a cart is created. Hmm interesting! I don't think we can do that with the way that carts are created right now.

PR #2196, opened 2020-02-26
Shows Error on invalid Product Quantity

supernova-at:
## Description

This PR pops up an error toast when a user enters a number into the product quantity stepper on the cart page that results in a graphQL error, usually occurring when the number entered exceeds the available stock of the product.

## Related Issue

Closes PWA-377.

### Verification Stakeholders

@schensley @sirugh 

### Verification Steps

1. Ensure an item is in your cart
1. Go to the `/cart` page
1. Manually enter a very large number into the quantity stepper input text field
1. Remove focus from the quantity stepper input text field
1. See that an error toast appears with a sensible error message

Screen Shot 2020-02-26 at 12 02 50 PM

Screen Shot 2020-02-26 at 12 02 55 PM

## Checklist

## Notes / Risks

The error message comes straight from the GraphQL backend. "Qty" isn't a great thing to show an end user, but we don't want to be in the business of massaging these error strings.


packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -18,6 +18,7 @@ export const useProduct = props => {
     const [{ cartId }] = useCartContext();
 
     const [isFavorite, setIsFavorite] = useState(false);
+    const [updateItemErrorMessage, setUpdateItemErrorMessage] = useState(null);
```
sirugh: My ignorance with apollo errror handling is showing but could we not just monitor the original mutation state? I don't know if that is cleared between requests but if it is it would prevent the need for this local state.

Something like

```js
const [updateItemQuanitty, { error: updateItemErrorMessage }] = useMutation(updateItemQuantityMutation);
useEffect(() => {
 if (updateItemErrorMessage) { ... }
}, [updateItemErrorMessage]);
```

supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/2196/commits/47dd63e778e8576be33ef27c626cdf492a022deb.

PR #2200, opened 2020-02-27
Fix conditionals in JSX

jimbo:
## Description
In this codebase, we avoid multiline statements in JSX because they defeat the main purpose of using JSX: readable trees. In practice, this only comes up when conditionally rendering JSX elements. We've established a pattern of first defining conditional elements as variables (constants), then inserting those variables into the final tree, and we've been pretty good about using this pattern. But we haven't been perfect, and that makes the pattern unclear.

This PR corrects all known cases in which we were not adhering to this pattern.

```jsx
// ❌
return (
  <div>
    {condition ? (
        <Foo />
      ) : (
        <Bar />
      )
    }
  </div>
)

// ✅
const element = condition ? (
  <Foo />
) : (
  <Bar />
)

return (
  <div>
    {element}
  </div>
)
```

## Related Issue
PWA-415

### Verification Steps
1. `yarn test`

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingMethods.js
+++ b/packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingMethods.js
@@ -24,6 +24,38 @@ const ShippingMethods = props => {
 
     const classes = mergeClasses(defaultClasses, props.classes);
 
+    const radios =
+        isShowingForm && hasMethods ? (
+            <Fragment>
+                <h3 className={classes.prompt}>Shipping Methods</h3>
+                <Form>
+                    <ShippingRadios
+                        selectedShippingMethod={selectedShippingMethod}
+                        shippingMethods={shippingMethods}
+                    />
+                </Form>
+            </Fragment>
+        ) : null;
+
+    const formOrPlaceholder = isShowingForm ? (
+        <Fragment>
+            <ShippingForm
+                hasMethods={hasMethods}
+                selectedShippingFields={selectedShippingFields}
+            />
+            {radios}
+        </Fragment>
+    ) : (
+        <Button
+            classes={{ root_lowPriority: classes.estimateLink }}
+            priority="low"
+            type="button"
+            onClick={showForm}
+        >
+            I want to estimate my shipping
+        </Button>
+    );
+
     return (
         <div className={classes.root}>
             <p className={classes.message}>
@@ -31,36 +63,7 @@ const ShippingMethods = props => {
                 provide the Country, State, and ZIP for the destination of your
                 order.
             </p>
-            {isShowingForm ? (
-                <Fragment>
-                    <ShippingForm
-                        hasMethods={hasMethods}
-                        selectedShippingFields={selectedShippingFields}
-                    />
-                    {hasMethods ? (
-                        <Fragment>
-                            <h3 className={classes.prompt}>Shipping Methods</h3>
-                            <Form>
-                                <ShippingRadios
-                                    selectedShippingMethod={
-                                        selectedShippingMethod
-                                    }
-                                    shippingMethods={shippingMethods}
-                                />
-                            </Form>
-                        </Fragment>
-                    ) : null}
-                </Fragment>
-            ) : (
-                <Button
-                    classes={{ root_lowPriority: classes.estimateLink }}
-                    priority="low"
-                    type="button"
-                    onClick={showForm}
-                >
-                    I want to estimate my shipping
-                </Button>
-            )}
+            {formOrPlaceholder}
         </div>
     );
 };
--- a/packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.ee.js
+++ b/packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.ee.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import { Price } from '@magento/peregrine';
 
 import { mergeClasses } from '../../../classify';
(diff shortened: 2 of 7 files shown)


packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.ee.js
```diff
@@ -39,14 +39,14 @@ export default props => {
     const cards = getGiftCards(props.data);
 
     return cards.value ? (
-        <>
+        <Fragment>
```
sirugh: Is this preferred over `<>`?

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingMethods.js
```diff
@@ -24,43 +24,46 @@ const ShippingMethods = props => {
 
     const classes = mergeClasses(defaultClasses, props.classes);
 
+    const radios =
+        isShowingForm && hasMethods ? (
```
revanth0212: How about using `useMemo` here? @sirugh and I had a conversation yesterday about this. He can expand more on the `useMemo` use case.

packages/venia-ui/lib/components/CartPage/cartPage.js
```diff
@@ -33,6 +33,15 @@ const CartPage = props => {
         ) : null;
     }, [classes.sign_in, handleSignIn, isSignedIn]);
 
+    const productListing = hasItems ? (
+        <ProductListing />
+    ) : (
+        <h3>There are no items in your cart.</h3>
+    );
+
+    const priceAdjustments = hasItems ? <PriceAdjustments /> : null;
```
revanth0212: Same here. Can we use `useMemo`? At most we will only be caching 2 versions of `priceAdjustments` either the component or null.

packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.ee.js
jimbo: Yes, it's a bit more explicit and portable. Generally, you can paste our code into a sandbox and it'll work. This particular transform isn't as popular as you might think.

Also, we've used `Fragment` almost everywhere else.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingMethods.js
jimbo: We have to draw the line somewhere on memoization, and in this codebase, I've generally drawn that line at element creation. The only place we should be memoizing element creation (afaik) is when we're doing it in a loop.

sirugh: > when we're doing it in a loop

Like...a _render_ loop? 😆

sirugh: But seriously, imagine a toggleable element that we expect a user to interact with quite often. Shouldn't we memoize the state to avoid re-creating the component unnecessarily?

This is the difference we're talking about, I think. Which of these is "more performant"/"better"?
```js
const MyComponent = ({ renderChildren }) => {
    const FooComponent = useMemo(() => {
        return renderChildren ? (
            <Foo />
        ) : null;
    });

    const BarComponent = renderChildren ? (
        <Bar />
    ) : null

    return (
        <Fragment>
            {FooComponent}
            {BarComponent}
        </Fragment>
    )
}
```

revanth0212: Yeah, I feel the same. `useMemo` can save on garbage collection if we are smart about the dependency array element. `useMemo` can quickly go out of hand if we abuse it by putting too many elements in the dependency array or elements that have a wide range of possible values. I don't think there is an eslint rule that can help us in this case but I feel `useMemo` has more potential that we are not using at this point in our code base.

jimbo: > This is the difference we're talking about, I think. Which of these is "more performant"/"better"?

> I feel `useMemo` has more potential that we are not using at this point in our code base.

Yes, I understand. The difference here is negligible—it's one `createElement()` call. The benefit is not zero, but it's insignificant.

```jsx
  // this is also a `createElement` call
  // should we memoize it too?
  return <div>{props.text}</div>
}
```

If we wanted to realize the full potential of memoization, we could reproduce `PureComponent`.

```jsx
// let's create a custom hook
// it'll only re-render when props change
const usePureComponent = (Component, props) => useMemo(
  () => <Component {...props} />
  [props]
)

// looks pure to me
  return <div>{props.text}</div>
}

// neat
export default const pure = props => usePureComponent(MyComponent, props)
```

This is cool, but it's overkill. We're not gaining much from the memoization, and now we're doing something that's pretty unusual. So this is where we draw the line.

We use it to avoid identities changing across renders, and as an early optimization when mapping props to elements in a loop, but not for literally every `createElement` call that happens to be pure (suitable for memoization).

jimbo: > But seriously, imagine a toggleable element that we expect a user to interact with quite often. Shouldn't we memoize the state to avoid re-creating the component unnecessarily?

Situationally, any time you have a particularly complex tree, you can consider memoizing it to improve the render performance. Toggling doesn't factor into it much, though.

sirugh: A question @revanth0212 had in the office was why not just memoize all the time since you may not know if a tree is complex or not.

PR #2206, opened 2020-03-03
[bugfix]: Coupon code with leading/trailing whitespace works

sirugh:
## Description

We now trim the whitespace around the entered coupon code.

## Related Issue

Closes [PWA-399](https://jira.corp.magento.com/browse/PWA-399).

### Verification Steps

1. Add an item to your cart and visit the `/cart` page.
2. Add a valid coupon code but with a leading space. It should work.
3. Remove and add the code but with a trailing space. It should work.
4. Remove and add the code but with a leading _and_ trailing space. It should work.

## Checklist


packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
```diff
@@ -30,10 +30,11 @@ export const useCouponCode = props => {
         async ({ couponCode }) => {
             if (!couponCode) return;
             try {
+                // Trim whitespace before submitting the coupon code
                 await applyCoupon({
                     variables: {
                         cartId,
-                        couponCode
+                        couponCode: couponCode.trim()
```
tjwiebell: I would prefer to see this behavior prevented in the UI, not applied before the network request. Add `mask={value => value && value.trim()}` to `couponCode.js`, and this shouldn't be needed. If you can think of an edge case where this would still be needed, I'm fine leaving it around.

sirugh: Oh...that's much cleaner :D

packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
```diff
@@ -30,6 +30,7 @@ export const useCouponCode = props => {
         async ({ couponCode }) => {
             if (!couponCode) return;
             try {
+                // Trim whitespace before submitting the coupon code
```
tjwiebell:
```suggestion
```

packages/venia-ui/lib/components/CartPage/GiftCards/giftCards.js
```diff
@@ -120,6 +120,8 @@ const GiftCards = props => {
                         id={classes.card}
                         disabled={isApplyingCard || isCheckingBalance}
                         field="card"
+                        mask={value => value && value.trim()}
```
tjwiebell: Someone on the team has advised me against this; I believe a concern about this function being created on every render. @supernova-at or @revanth0212, is this style of inline function okay? I have a feeling someone would say this should be a returned `talonProp`, but shouldn't be wrapped in `useCallback` since it won't change.

sirugh: Normally I would have returned these from the talon but a mask function is this weird middle ground between logic and UI. Curious what @jimbo thinks.

Personally I'd like to leave it or make it a util function but the savings either way make me feel like it's not super important.

PR #2213, opened 2020-03-04
SW PDP error fix for IMAGE_OPTIMIZING_ORIGIN=backend

revanth0212:
## Description

When **IMAGE_OPTIMIZING_ORIGIN** is set to **backend**, **resourceUrl** function of **venia-drivers** package returns an absolute URL instead of a relative URL in **useProductImageCarousel**. SW would not accept any origin other than the origin of the app that created it due to CORS restrictions.

Fixed this by stripping away the backend origin set by **resourceUrl** before sending it to the SW in case of **IMAGE_OPTIMIZING_ORIGIN** set to **backend**.

## Related Issue
Closes [PWA-392](https://jira.corp.magento.com/browse/PWA-392)

### Verification Steps
1. Set `IMAGE_OPTIMIZING_ORIGIN` to `backend` in `.env` file.
2. Build and stage venia using `yarn build && yarn stage:venia`.
3. Go to any PDP page.
4. Should not see any errors in the console.

## Checklist
* Should not change the app's behavior in watch mode.
* No app functionality has been changed.


tjwiebell:
There was some optimization need to send these requests directly to Magento instead of proxying them through UPWARD (which is what this change has done). I would need @zetlen to provide some insight into that requirement, but I wouldn't feel comfortable accepting this change without his input.

**If** that is no longer a requirement, I think we should just go ahead and make this change directly in `makeUrl.js:69`, which should be a simple one-line change of removing `&& !useBackendForImgs`.

```js
if (baseURL.href.startsWith(mediaBackend) && !useBackendForImgs) {
    baseURL = new URL(baseURL.href.slice(baseURL.origin.length), origin);
}
```

zetlen:
The only reason CORS matters here is because our smart image cache code uses `fetch()` to manually request and cache an image response:

```js
const fetchAndCacheImage = imageURL =>
    fetch(imageURL).then(response =>
        caches
            .open(CATALOG_CACHE_NAME)
            .then(cache => cache.put(imageURL, response.clone()))
            .then(() => response)
    );
```

Regular image tags don't need CORS, because there's no way for the local JS environment to spy on the HTTP response anyway. But for `fetch()`, you need to specify that you don't want to access the response data, you only want to cache it!

```js
const fetchAndCacheImage = imageURL =>
    fetch(imageURL, { mode: 'no-cors' }).then(response =>
        caches
            .open(CATALOG_CACHE_NAME)
            .then(cache => cache.put(imageURL, response.clone()))
            .then(() => response)
    );
```

zetlen:
If we set the `no-cors` mode on the ServiceWorker fetch, and we use the `URL` constructor to tolerate both relative and absolute URLs, we can make this work in `backend` and `auto` mode.

Opening a PR to this PR!

revanth0212:
Thanks, guys for looking into this.

PR #2215, opened 2020-03-04
[feature]: Disable cart mutations while others are in flight

sirugh:
## Description

~When any mutation occurs on the cart page we need to disable other inputs/mutations otherwise it is very possible (and likely) that the cache data gets out of sync with the server.~

When any mutation occurs, queue it to avoid a race condition where cart data gets out of sync with the server. While any mutation is in flight (ie `called && loading`), hide the price summary values until complete.

Instead of applying a mask to basically enforce that a user cannot make a change while another is in flight we can use an Apollo [link](https://www.apollographql.com/docs/link/stateful/) to prevent mutations firing at the same time, regardless of how many are "queued". I created one specifically for this purpose, [apollo-link-mutation-queue](https://github.com/adobe/apollo-link-mutation-queue).

The idea is simple - if the link receives a mutation and there are no other mutations in flight, execute it immediately. If the link receives a mutation and there is a mutation already in flight, enqueue it to be executed after the in-flight mutation resolves.

If we ever need to allow a mutation to skip the queue, just provide `{ context: { skipQueue: true }` and it will be requested in parallel with any other requests.

## Related Issue

Closes [PWA-385](https://jira.corp.magento.com/browse/PWA-385).

### Verification Steps

1. Add a product and go to `/cart`.
2. Make a mutation such as updating quantity, applying coupon/gift card etc. The prices should disappear while the mutation is in flight.

[Image from Gyazo](https://gyazo.com/e8f647efb00b0471d5b86d50e8a88c93)

## Checklist


packages/venia-concept/package.json
```diff
@@ -60,6 +60,7 @@
     "apollo-client": "2.6.4",
     "apollo-link-context": "~1.0.17",
     "apollo-link-http": "~1.5.11",
+    "apollo-link-mutation-queue": "~1.0.0",
```
sirugh: This PR won't work until I officially publish this file.

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummary.css
```diff
@@ -26,6 +26,12 @@
     font-weight: bold;
 }
 
+.priceUpdating {
+    @composes: price;
+    color: white;
+    background: white;
```
sirugh: This just covers the "stale" price data. Should I use a different method to cause this effect?

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -6,17 +6,39 @@ import { useCartContext } from '@magento/peregrine/lib/context/cart';
 export const useProduct = props => {
     const {
         item,
-        removeItemMutation,
+        mutations: { removeItemMutation, updateItemQuantityMutation },
```
tjwiebell: I recall a guideline opposing nested destructuring, but can't find it. I'm definitely guilty of doing this with Apollo hooks, so if that is the case a lot of instances have slipped through. Just calling it out, might be a tech debt thing we should fix, or even better, add a lint rule for it if this is something we do prefer.

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummary.css
```diff
@@ -26,6 +26,12 @@
     font-weight: bold;
 }
 
+.priceUpdating {
+    @composes: price;
```
tjwiebell:
```suggestion
    composes: price;
```

packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummary.css
tjwiebell: Preferable to use `opacity` here. I also saw some older XD mocks that had a transition which I think we should add. This seems to work, but causes an odd cascade; review with UX and if its too much trouble we can leave as-is.
```css
.price {
    justify-self: end;
    transition: opacity 384ms var(--venia-anim-standard);
    opacity: 1;
}

.priceUpdating {
    composes: price;
    opacity: 0;
}
```

packages/venia-ui/lib/components/CartPage/cartPage.js
```diff
@@ -34,13 +42,18 @@ const CartPage = props => {
     }, [classes.sign_in, handleSignIn, isSignedIn]);
 
     const productListing = hasItems ? (
-        <ProductListing />
+        // TODO: Pass setter to other mutation-invoking components
```
tjwiebell:
```suggestion
```

packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
```diff
@@ -21,11 +21,10 @@ const MOCKED_ADDRESS = {
...
 export const useShippingForm = props => {
     const {
-        getCountriesQuery,
-        getStatesQuery,
         selectedValues,
-        setShippingAddressMutation,
-        shippingMethodsQuery
+        setIsCartUpdating,
+        mutations: { setShippingAddressMutation },
+        queries: { getCountriesQuery, getStatesQuery, shippingMethodsQuery }
```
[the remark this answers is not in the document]
sirugh: I think so, though the naming of the properties within the operations props is debatable. Should they be called `setShippingAddressMutation` or `setShippingAddress`? What should the mutate function returned from the hook be named in that case?

yarn.lock
```diff
@@ -6384,7 +6391,7 @@ cross-spawn@6.0.5, cross-spawn@^6.0.0, cross-spawn@^6.0.5:
     shebang-command "^1.2.0"
     which "^1.2.9"
 
-cross-spawn@^5.0.1:
+cross-spawn@^5.0.1, cross-spawn@^5.1.0:
```
tjwiebell: I just noticed that running a `yarn install` on my local reverts all of these lock file changes. Did you upgrade some other dependencies in this scope by mistake?

packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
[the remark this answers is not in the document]
tjwiebell: The old ATM Machine problem; I don't like the redundancy, but I dislike more having to come up with a new naming convention for the execute functions (unless you like `executeSetShippingAddress`).

yarn.lock
sirugh: Do you mean when you check this branch out locally?

sirugh: Fixed.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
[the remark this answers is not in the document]
sirugh: https://github.com/magento/pwa-studio/wiki/Project-coding-standards-and-conventions#talons

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
sirugh: Let's leave this for now until we can enforce a linting rule for it.

PR #2216, opened 2020-03-04
[PWA-388] Billing Information form does not have email validation as in shipping information form.

tjwiebell:
## Description

Steps - 
1. Add product to cart and go to Checkout panel.
2. Click on "Add billing Information" 
3. Uncheck the “Billing address same as shipping address” check box.
4. Enter valid card details, in address section enter invalid email address.
5. Hit Save button 

Expected - GraphQl validation should happen and email field should show error.
Actual - There is no GraphQl validation. When user tries to checkout > Checkout fails with message in network response.
{“message”:“The shipping information was unable to be saved. Verify the input data and try again.“}

This what not seen before as we had client side validation. 

## Implementation Note
It seems a little odd that to solve the bug we're removing the field entirely that is causing it; but this has been approved by UX (@soumya-ashok) as a pattern that will continue in the new checkout page, and it didn't make sense to have this field, since Magento doesn't support billing email address. I suspect that may have been why we got such a general error message; the REST endpoint docs had email incorrectly as a payload property, but it confused the backend when it failed validation, because it shouldn't have been there.

## Related Issue

* [[PWA-388](https://jira.corp.magento.com/browse/PWA-388)] Billing Information form does not have email validation as in shipping information form.

### Verification Steps

1. Add Simple Product to cart
2. Open MiniCart and checkout, using a different billing address from shipping
3. Verify that email address field is no longer present
4. Verify that you can still submit an Order with billing address details filled in
## Checklist


dpatil-magento:
@tjwiebell  If I place order with different billing address and hit `Create  Account` button, email field has `undefined` text instead of showing email address entered in shipping form.

image

sirugh:
Oh no, I think there may be more involved with this - I recall now that if you select "same as shipping", redux tries to copy some state from the address form into the billing form. We should double check where we are reading from the billing address slice and make sure it's not using email anymore.

dpatil-magento:
QA Pass, good to merge after review approval.

PR #2222, opened 2020-03-05
[SPIKE] Move Fragments and Queries to GraphQL Files

tjwiebell:
## Description

Discovery work to see if moving all fragments and queries/mutations to graphql files.

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist


fooman:
Seeing that this is some exploratory work I wondered how feasible it was to implement a Fragment Collector as part of an intercept target

```
module.exports = targets => {
    targets.declare({
        cartPageFragmentCollector: new targets.types.Sync(['fragments'])
    });
};
```

this could be either specific per query or the fragment collector works for all queries. In the latter case the intercepting code would have to specify the query it wants to add a fragment onto.

PR #2235, opened 2020-03-10
[PWA-428] Update upward-connector PWA install docs

tjwiebell:
## Description

Current https://magento.github.io/pwa-studio/tutorials/cloud-deploy/ docs is based on pre built dist. But we have removed dist from publishing. So, doc needs update on how to build part of magento cloud installation and upward.yml file path to use.

## Related Issue

* [[PWA-428](https://jira.corp.magento.com/browse/PWA-428)] Update upward-connector PWA install docs

### Verification Steps

1. Follow instructions in article
2. Verify no errors during build and deployment
3. Verify Venia storefront actually works following these steps

## Checklist


pwa-devdocs/src/tutorials/cloud-deploy/index.md
```diff
@@ -175,44 +135,26 @@ These commands modify the `composer.json` file and adds the sample data modules
...
-```sh
-yarn install
-```
-
-A successful installation creates a `yarn.lock` file in your project directory.
-
 ### Composer dependencies
 
-Install the new Composer dependencies added to the `composer.json` file.
+_Optional_: If you manually copied the `composer.json` file changes from previous steps, you'll need to update your `composer.lock` with these new dependencies.
```
jcalcaben:
```suggestion
_Optional_: If you manually modified the `composer.json` file to make the changes described in the previous steps, update your `composer.lock` with these new dependencies:
```

pwa-devdocs/src/tutorials/cloud-deploy/index.md
```diff
@@ -221,7 +163,7 @@ dependencies:
 The `venia-concept` storefront uses a newer version of Node than the default environment provides.
 It also uses Yarn to install its dependencies.
 
-The following update to the `hooks.build` section of `.magento.app.yaml` installs NVM, installs the latest LTS release of Node, and uses Yarn to install the project dependencies.
+The following update to the `hooks.build` section of `.magento.app.yaml` installs NVM and installs the latest LTS release of Node. NPX runs a `buildpack` command  `create-project` that takes arguments to scaffold the application using the `venia-concept` template. Install and build the application, then move that artifact to an area on the file system that is persisted after the build phase.
```
jcalcaben:
```suggestion
The following update to the `hooks.build` section of `.magento.app.yaml` installs NVM and the latest LTS release of Node.

It also uses NPX to run the `buildpack` CLI tool with the `create-project` sub-command to scaffold the application using the `venia-concept` template.
After scaffolding completes, it installs and builds the application and moves the build artifact to a persistent area on the file system.
```

pwa-devdocs/src/tutorials/cloud-deploy/index.md
```diff
@@ -235,7 +177,13 @@ hooks:
         [ -s "$NVM_DIR/nvm.sh"  ] && \. "$NVM_DIR/nvm.sh"
         nvm install --lts=dubnium
 
-        yarn install
+        npx @magento/pwa-buildpack create-project /tmp/pwa --template venia-concept --backend-url https://[your-cloud-url-here] --name Venia --author Magento --install false --npm-client yarn
```
jcalcaben: Does this mean developers will have to set up their project so that it can be deployed using the `create-project` tool? Which would mean that they will need to create their own [`create.js`](https://magento.github.io/pwa-studio/pwa-buildpack/scaffolding/custom-project-templates/) instruction file for the command?

tjwiebell: I would consider these instructions for people wanting to use `venia-concept` as a turn key solution; but this was a reason I wanted @zetlen to take a peek at these instructions.

It looked like `create-project` is capable of accepting any npm package as a template, but now requires you select one from a list (`venia-concept` being the only option). If we removed that list restriction, anyone who started with a `venia-concept` scaffolded app could publish and use `--template @awesome-si/super-cool-pwa`.

This falls in line with that production deployment story where we've positioned ourselves as unopinionated. We give you instructions for how to produce your bundle, mostly in a local context, but don't get into too much detail on how to make it public.

pwa-devdocs/src/tutorials/cloud-deploy/index.md
```diff
@@ -11,51 +11,14 @@ title: Magento Cloud deployment
...
-
-```json
-{
-    "dependencies": {
-        "@magento/venia-concept": "~4.0.0"
-    }
-}
-```
+These instructions provide a method for building your application bundle in the Magento Cloud, but
+you can get the same results by building locally and checking your local build artifacts into source control.
```
zetlen:
```suggestion
you can get the same results by building locally with the same environment variables as your targeted production environment, and then checking your local build artifacts into source control.
```

zetlen: My edit could maybe be phrased better, but I think this is important to note, since environment variables could cause the remote build and the local build to produce different artifacts.

You'd hope they would not, because the local simulated "production" environment should include all the non-secret production variables, you don't want to hardcode any secrets into the bundle.

pwa-devdocs/src/tutorials/cloud-deploy/index.md
```diff
@@ -175,44 +138,26 @@ These commands modify the `composer.json` file and adds the sample data modules
...
-### Install Yarn
+### Install NPX and Yarn
 
-If your project uses Yarn, which is the case for `venia-concept`, add the following entry to the `.magento.app.yaml` file.
+If your project supports Yarn, which is the case for `venia-concept`, add the following entry to the `.magento.app.yaml` file.
 
 ```yaml
 dependencies:
     nodejs:
+      npx: "*"
```
zetlen: Did you find that `npx` actually needs to be installed? On recent versions of Node it comes built-in.

pwa-devdocs/src/tutorials/cloud-deploy/index.md
zetlen: The `create-project` template functionality is not officially supported yet; there are some kinks to work out, and I haven't notated them yet because I jumped hurriedly off the project to work on BuildBus. So I wouldn't document that quite yet.

tjwiebell: Yeah, it looked like the logic was built out far enough to support `create-pwa`, but I could see the long term intention. That should be fine for now, with the title note that local builds are suitable if you're not down for turn key `venia-concept`.

pwa-devdocs/src/tutorials/cloud-deploy/index.md
tjwiebell: To my knowledge, the only runtime variables are for UPWARD, so just `NODE_ENV` and `MAGENTO_BACKEND_URL`, all others are baked in. This includes `CHECKOUT_BRAINTREE_TOKEN`, but I always considered that a public key that wasn't useful by itself. Definitely worth auditing if that's not the expectation.

pwa-devdocs/src/tutorials/cloud-deploy/index.md
tjwiebell: Doesn't seem so, must have got left behind when experimenting with removal of NVM steps. Removed explicit dependency in f934608.

PR #2255, opened 2020-03-17
Removed cart and checkout routes from 6.0 branch.

revanth0212:
## Description

Deleted Cart and Checkout routes. We let webpack deal with not including cart and checkout code into the build.

Here are the build sizes before and after the changes for perspective:

Before

image

After

image

## Related Issue
Closes [PWA-429](https://jira.corp.magento.com/browse/PWA-429)

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Go to cart page. Should not render.
2. Go to checkout page. Should not render.

## Checklist
None

### DIFF
--- a/packages/venia-ui/lib/components/Routes/routes.js
+++ b/packages/venia-ui/lib/components/Routes/routes.js
@@ -4,8 +4,6 @@ import { Route, Switch } from 'react-router-dom';
 import { fullPageLoadingIndicator } from '../LoadingIndicator';
 import MagentoRoute from '../MagentoRoute';
 
-const CartPage = lazy(() => import('../CartPage'));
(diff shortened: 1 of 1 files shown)


brendanfalkowski:
Just curious — because we have dozens of custom routes — am I understanding it correctly that lazy-loading a route does not exclude those assets from the bundle? That's the opposite of what I assumed `React.lazy()` does.

I thought this was the preferred approach, but based on this PR something in webpack magic (maybe in 6.0?) is able to understand those routes exist without a definition of them in `routes.js`.

I'm missing part of the context for these changes, but I like what it suggests (minus the dead-simple clarity of one "routes" file).

sirugh:
I would have expected lazy loading to chunk. Maybe webpack bundle analyzer would show different results. Something to look into for sure :o

PR #2256, opened 2020-03-17
[Chore] Remove experimental-worker flag for performance.

revanth0212:
## Description

Removing the `experimental-worker` flag from jest scripts for performance reasons. Also, it is not needed anymore because we are mocking `MessageChannel` for service worker tests.

`experimental-worker` eats up a lot of memory as shown in the below screen grabs and occasionally jest runs out of memory.

**Jest Heap before (1169 MB):**

`node --experimental-worker --expose-gc ./node_modules/.bin/jest --runInBand --no-cache --logHeapUsage --coverage=false`

image

**Jest Heap after changes (361 MB):**

`node --expose-gc ./node_modules/.bin/jest --runInBand --no-cache --logHeapUsage --coverage=false`

image

## Related Issue
Closes [PWA-452](https://jira.corp.magento.com/browse/PWA-452)

### Verification Stakeholders
@zetlen 

### Verification Steps
1. Run `node --expose-gc ./node_modules/.bin/jest --runInBand --no-cache --logHeapUsage --coverage=false` to see the heap usage per test. Tests should not fail.

## Checklist
None


package.json
```diff
@@ -36,10 +36,10 @@
     "stage:venia": "yarn venia run start",
     "stats:venia": "yarn venia run build:analyze && yarn venia run stats",
     "storybook:venia": "yarn workspace @magento/venia-ui run storybook",
-    "test": "node --experimental-worker node_modules/.bin/jest",
-    "test:ci": "node --experimental-worker node_modules/.bin/jest --max-workers=3 --json --outputFile=test-results.json",
+    "test": "jest",
+    "test:ci": "jest --no-cache --max-workers=3 --json --outputFile=test-results.json",
```
tjwiebell: Did you intentionally add the `--no-cache` flag here?

zetlen: No, that was me, sorry. It seemed wise not to rely on cache for the CI tests, so I updated it while I was testing things out.

PR #2257, opened 2020-03-17
Checkout Items Review Component

revanth0212:
## Description

Created the review of items in order component.

## Related Issue
Closes [PWA-406](https://jira.corp.magento.com/browse/PWA-406)

### Verification Steps
1. Add items to the cart.
2. Go to the checkout page.
3. Complete address and payment steps. You should see review of the items section before you place the order.

**Desktop (Showing only 2):**

image

**Desktop (Showing All):**

image

**Mobile (Showing only 2):**

image

**Mobile (Showing All):**

image

## Checklist
None


packages/peregrine/lib/talons/CheckoutPage/ItemsReview/useItemsReview.js
```diff
@@ -0,0 +1,69 @@
...
+             * + will convert it to 0
+             */
+            setTotalQuantity(+data.cart.total_quantity);
+        }
+    }, [data, called, loading, error, setItems]);
+
+    return {
+        isLoading: !!loading,
+        items: showAllItems ? items : items.slice(0, 2),
+        setItems,
```
sirugh: Do you need to return this function from the talon?

packages/peregrine/lib/talons/CheckoutPage/ItemsReview/useItemsReview.js
```diff
@@ -0,0 +1,69 @@
...
+            /**
+             * if total_quantity turns out to be null
+             * + will convert it to 0
+             */
+            setTotalQuantity(+data.cart.total_quantity);
+        }
+    }, [data, called, loading, error, setItems]);
+
+    return {
+        isLoading: !!loading,
```
sirugh: It may be worth it to also return `hasError: !!error`. You could use that in the UI component to render some error display like "Oops! Please reload the page to see your items".

packages/peregrine/lib/talons/CheckoutPage/ItemsReview/useItemsReview.js
```diff
@@ -0,0 +1,69 @@
...
+            /**
+             * if for some reason items turns out to be
+             * null set items to []
+             */
+            setItems(data.cart.items || []);
+            /**
+             * if total_quantity turns out to be null
+             * + will convert it to 0
+             */
+            setTotalQuantity(+data.cart.total_quantity);
```
sirugh: I don't think we need local state for `items` or `totalQuantity`. Should just be able to return the values from `data` or some default:

```js
const items = data ? data.cart.items : [];
return {
  items: showAllItems ? items : items.slice(0, 2),
  totalQuantity: data ? data.cart.total_quantity : 0
  ...
}

File: packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReviewQuery.js [hunk 4]
```diff
@@ -0,0 +1,40 @@
+import gql from 'graphql-tag';
```
sirugh: This file should be named `itemsReview.gql.js` per the convention we discussed in slack. I just added it today in the wiki :)

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReviewQuery.js
```diff
@@ -0,0 +1,40 @@
...
+
+const ProductListingFragment = gql`
```
sirugh: This fragment should live in a file named `itemsReviewFragments.gql.js` and the fragment itself should be named `ItemsReviewFragment`.

packages/peregrine/lib/talons/CheckoutPage/ItemsReview/useItemsReview.js
revanth0212: No, added it by mistake. Nice catch.

packages/peregrine/lib/talons/CheckoutPage/ItemsReview/useItemsReview.js
revanth0212: Nice idea. Will do that.

packages/peregrine/lib/talons/CheckoutPage/ItemsReview/useItemsReview.js
revanth0212: Makes sense, data is probably a state variable coming from the apollo hooks. Done.

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.css
```diff
@@ -1,12 +1,60 @@
...
+
+.item_root {
+    display: grid;
+    grid-template-columns: 1fr 4fr;
+    gap: 0.5rem;
+}
+
+.item_thumbnail {
+    grid-column: 1 / span 1;
+    grid-row: 1 / span 3;
```
jimbo: Good syntax; we do prefer these properties and this value structure.

That said, explicitly placing items into the grid is usually a sign of overcomplicating things. I'll check out the branch and make a call in a bit.

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.css
```diff
@@ -1,12 +1,60 @@
...
 
-.container {
-    min-height: 20rem;
+.show_all_items_button {
+    composes: root_normalPriority from '../../Button/button.css';
+    width: 100%;
+    border: none;
+}
+
+.show_all_items_button > span {
```
jimbo: Avoid targeting tagnames—ever. In this case, it's not necessary anyway, because the element you're targeting will inherit `font-size` from its parent, the button.

```suggestion
.show_all_items_button {
```

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.css
```diff
@@ -1,12 +1,60 @@
...
-    padding: 25px;
-    padding-top: 50px;
+.total_quantity {
     font-weight: bold;
 }
 
-.container {
-    min-height: 20rem;
+.show_all_items_button {
+    composes: root_normalPriority from '../../Button/button.css';
```
jimbo: Our normal button styles have very little to do with the button you're styling here. I would suggest not extending our normal button and instead styling this button from scratch.

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.css
```diff
@@ -1,12 +1,60 @@
...
+
+.items_container {
+    display: grid;
+    gap: 2rem;
     padding: 1.5rem 1rem;
 }
+
+.item_root {
+    display: grid;
+    grid-template-columns: 1fr 4fr;
```
jimbo: Layouts made up of entirely flexible tracks do exist, but they're very unusual. This isn't one of them. The image has a fixed width and doesn't have much use for extra space, so all the extra space should go to the second column.

```suggestion
    grid-template-columns: auto 1fr;
```

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.css
jimbo: After review, I think this is a good case for explicit placement. 👍

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/showAllItemsFooter.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import { mergeClasses } from '../../../classify';
+import Button from '../../Button';
+
+import defaultClasses from './itemsReview.css';
```
jimbo: This component should have its own CSS file. Also, the classnames should be as simple as possible here: `root` and `button`.

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/showAllItemsFooter.js
```diff
@@ -0,0 +1,23 @@
...
+
+const ShowAllItemsFooter = props => {
```
jimbo: I'd suggest reworking this component a bit.

- Make it just a button, rather than a button wrapped in a container
- Rename it to `ShowAllButton` or something
- Have it accept one prop, `onClick`
- Render a `button` element rather than our `Button` component
- Render the container element as part of the parent component instead

It should have its own CSS file. If it's just a button, it can have one classname: `root`.

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.css
revanth0212: Nice one. I was not aware of this 👍

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/showAllItemsFooter.js
revanth0212: Got it. Done.

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.css
```diff
@@ -1,12 +1,60 @@
...
-    min-height: 20rem;
+.show_all_items_button {
+    composes: root_normalPriority from '../../Button/button.css';
+    width: 100%;
+    border: none;
+}
+
+.show_all_items_button > span {
+    font-weight: bold;
+    font-size: 0.9rem;
```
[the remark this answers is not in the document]
tjwiebell: @revanth0212 - I replied to your comment in another PR, but wanted to clarify here too just in case; still use `rem` units, do not use units that result in fractional pixel values.

soumya-ashok:
@revanth0212 - Listing the options how @jimbo suggested is fine. Let's keep that.
I am seeing a ton of space separating the second product from the horizontal rule below. In the mock I have 24px between them.

Additionally, if I go to the cart from the review step and return to checkout, the review order button shows up between the payment card and the items card. Maybe this is a caching issue as well? Not sure.

Screen Shot 2020-03-25 at 3 42 09 PM

tjwiebell:
@revanth0212 - This is very close to QA pass. Please see [this comment](https://github.com/magento/pwa-studio/pull/2257#issuecomment-604084138) about too much white space.

I am also still seeing `SHOW ALL ITEMS` regardless of my viewport size, which I believe is supposed to be mobile only. If that requirement didn't change, please investigate.

jimbo:
> I am also still seeing SHOW ALL ITEMS regardless of my viewport size, which I believe is supposed to be mobile only. If that requirement didn't change, please investigate.

Personally I'd be fine with it on desktop as well.

soumya-ashok:
@tjwiebell and @jimbo - "Show all items" is intended for all viewports. The default state would be to display up to 2 items with the option to see more if they exist.

tjwiebell:
> @tjwiebell and @jimbo - "Show all items" is intended for all viewports. The default state would be to display up to 2 items with the option to see more if they exist.

@revanth0212 - I'm not sure where I pulled that requirement from then, sorry 😬Fix the white space problem and this is ready to merge.

revanth0212:
> > @tjwiebell and @jimbo - "Show all items" is intended for all viewports. The default state would be to display up to 2 items with the option to see more if they exist.
> 
> @revanth0212 - I'm not sure where I pulled that requirement from then, sorry 😬Fix the white space problem and this is ready to merge.

Hey @tjwiebell it will be handled in a different PR. You can merge this for now.

revanth0212:
The white space issue should be gone now. Also the animation can not be done for rectangle expansion. It can only be done for the visibility of the hidden items.

image

soumya-ashok:
@revanth0212 and @jimbo UX approved

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/item.css
```diff
@@ -0,0 +1,48 @@
+.root {
+    display: grid;
+    grid-template-columns: auto 1fr;
+    grid-template-rows: auto auto 1fr;
+    column-gap: 1rem;
+    opacity: 1;
+    visibility: visible;
+    transition-duration: 350ms;
+    transition-property: opacity visbility;
```
jimbo:
```suggestion
    transition-property: opacity, visbility;
```

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/item.css
```diff
@@ -0,0 +1,48 @@
+.root {
+    display: grid;
+    grid-template-columns: auto 1fr;
+    grid-template-rows: auto auto 1fr;
+    column-gap: 1rem;
+    opacity: 1;
+    visibility: visible;
+    transition-duration: 350ms;
```
jimbo: Venia's animation durations are generally in increments of 128ms. Half-increments are also acceptable for fine tuning. Here, I recommend slightly increasing the duration and adding a small delay to allow the container's reflow to complete before fading in.

```suggestion
    transition-delay: 64ms;
    transition-duration: 384ms;
```

revanth0212: Oh, good to know. Wasn't aware of that. Thank you.

PR #2267, opened 2020-03-18
[security] minimist >= 1.2.2

sirugh:
## Description

1) Migrated `pwa-devdocs` to `yarn` from `npm`.
2) Used `resolutions` to resolve security vulnerability in `pwa-devdocs` and `docker` packages.

## Related Issue

Closes [PWA-449](https://jira.corp.magento.com/browse/PWA-449).

### Verification Steps

1. Verify docker did not break.
2. Verify pwa-devdocs did not break.
3. Verify that vulnerable versions of minimist are no longer included.

## Checklist


docker/package.json
```diff
@@ -6,5 +6,8 @@
   "license": "ISC",
   "dependencies": {
     "devcert": "~1.1.0"
+  },
+  "resolutions": {
+    "minimist": "^1.2.2"
```
sirugh: Using `resolutions` effectively "steps over" any transitive dependency requirements on older versions. This may cause issues, especially in this case where the resolution has a major bump.

The only other alternative to fixing this would be to update down the chain of transitive deps, but each of those packages also needs to get fixed/published. So this is the fastest way to resolve the security issues but it may cause a bug/break in the app.

dpatil-magento:
Below tests pass.
`bash docker/run-docker`
`npm run develop` (pwa-devdocs)
`yarn develop` (pwa-devdocs)

PR #2272, opened 2020-03-19
Style icons with CSS

jimbo:
## Description
Style the `Icon` component with CSS rather than SVG attributes.

## Related Issue
PWA-445

## Acceptance 
- @tjwiebell 

### Verification Stakeholders
- @larsroettig 
- @sirugh 

### Verification Steps
1. View icons throughout the site, verifying none have changed.
2. Fill and empty your cart, verifying that the icon changes when you do.

## Checklist


packages/venia-ui/lib/components/Header/cartTrigger.css
```diff
@@ -8,3 +8,12 @@
     font-weight: 600;
     margin-left: 0.3rem;
 }
+
+.icon--empty {
+    composes: root from '../Icon/icon.css';
+}
+
+.icon--filled {
```
sirugh: So instead of passing a color prop, if you want to override the filled color you would provide your own `icon--filled` class, right?

packages/venia-ui/lib/components/Header/cartTrigger.js
```diff
@@ -42,21 +37,16 @@ const CartTrigger = props => {
...
         </button>
     );
 };
 
+export default CartTrigger;
+
 CartTrigger.propTypes = {
-    iconColor: string,
     classes: shape({
         root: string
```
sirugh: Can you add any missing classes? This is our only documentation of the change.

packages/venia-ui/lib/components/Icon/icon.js
```diff
@@ -1,39 +1,41 @@
...
-    }
-}
+export default Icon;
 
-export default classify(defaultClasses)(Icon);
+Icon.propTypes = {
+    attrs: shape({}),
+    classes: shape({
+        'icon--empty': string,
+        'icon--filled': string,
```
sirugh: I don't think this is accurate - we aren't passing `icon--empty` or `icon--filled` with `classes`. We apply one or the other to the `Icon` itself via `className`. It just happens to get spread onto the Icon through `restProps`.

jimbo: Oh, I see what I did. I put the new classnames in the prop types for `Icon` rather than `CartTrigger`. I'll fix that.

As for `className` as a prop, you're right. The `classify` HOC allowed us to pass `className` as a shortcut override for `classes.root`, but its `mergeClasses` utility is simpler and doesn't have that feature. I'll update `CartTrigger` to override `classes.root` more explicitly.

dpatil-magento:
@jimbo  Pls take on following observations , PR (left) vs develop (Right)
1 No green border in Kebab icons, qty stoppers ui.
2. No X icon, right mark in price check box and color .
3. Pagination icons differ.
image

jimbo:
> 1. No green border in Kebab icons, qty steppers ui

Fixed both. Kebab icons are now slightly larger; stepper icons are now slightly smaller.

> 2. No X icon, right mark in price check box and color

Fixed both. The synthetic checkboxes now use raw feather icons with a new appearance.

> 3. Pagination icons differ.

Prefer the new unfilled style, so this is fine.

> 4. Sort select icon, right mark missing.

Fixed. Checkmark now uses raw feather icons with a new appearance.

dpatil-magento:
@jimbo Fixes looks good but problem with paginations icons issue is,  they don't get disabled when user is first or last page.

image

jimbo:
> Fixes looks good but problem with paginations icons issue is, they don't get disabled when user is first or last page.

Good catch. Fixed.

dpatil-magento:
UI looks good on Chrome, Safari, firefox, android chrome. Good to merge once latest commits are reviewed.

packages/venia-ui/lib/components/Pagination/navButton.css
```diff
@@ -1,10 +1,18 @@
...
     composes: button from './tile.css';
     display: grid;
     padding-bottom: 1px;
 }
 
-.buttonInactive {
-    composes: buttonArrow;
-    color: #999;
+.root_disabled {
+    composes: root;
```
tjwiebell: This class seems a bit redundant. Should this instead just be, or did you intend to add additional styles?
```css
.root:disabled {
  pointer-events: none;
}
```

jimbo: Fair point. I had included it as an optional classname to mirror the one for the icon, but the pseudo is probably sufficient for the root itself.

jimbo: While I was fixing this, I noticed a few more bits of weird CSS in the pagination styles. These have been fixed. Please check those out too. @tjwiebell

PR #2283, opened 2020-03-25
fix(pagebuilder): Html content type unescapes when GraphQL does not

zetlen:
## Description

Due to an apparent regression in M2.3.5, the GraphQL API now passes HTML content (in CMS blocks, anyway) through at least one PageBuilder-related render. The effect is that HTML content nodes, whose content itself was HTML encoded (`<div>` became `&lt;div&gt;`), was no longer HTML encoded. This broke the PageBuilder Html content type by stripping all HTML tags from it.

The Html "config aggregator" was using `node.textContent` to get and unescape its HTML, but when GraphQL has already unescaped it, we must use innerHTML instead. This PR updates the config aggregator with that logic.

Because the bug in MC-32787 may indicate a misconfiguration, the config aggregator logs a message in development mode.

## Related Issue

Closes https://jira.corp.magento.com/browse/MC-32787.

### Verification Stakeholders
@sirugh 
@davemacaulay 
@dpatil-magento 
### Verification Steps

1. Set your `MAGENTO_BACKEND_URL=https://pwa-test1-scjyl6a-vqnoqmbf7mivy.us-4.magentosite.cloud/` a 2.3.4 instance.
2. `yarn run watch:venia`
3. Visit homepage.
4. Verify it looks normal.
1. Set your `MAGENTO_BACKEND_URL=https://magento232-qwt7c3a-vqnoqmbf7mivy.us-4.magentosite.cloud/` a **2.3.5** instance
2. `yarn run watch:venia`
3. Visit homepage.
4. Verify it still looks normal, whereas in `develop` this formatting is bonkers.

### DIFF
--- a/packages/pagebuilder/lib/ContentTypes/Html/configAggregator.js
+++ b/packages/pagebuilder/lib/ContentTypes/Html/configAggregator.js
@@ -1,12 +1,23 @@
 import { getAdvanced } from '../../utils';
 
 export default node => {
-    const dom = new DOMParser().parseFromString(
-        '<!doctype html><body>' + node.textContent,
-        'text/html'
-    );
+    let html;
+    if (node.dataset.decoded) {
+        html = node.innerHTML;
+        if (process.env.NODE_ENV !== 'production') {
+            console.warn(
+                'PageBuilder HTML content was unescaped! This may be a Magento configuration error.'
+            );
+        }
+    } else {
+        const dom = new DOMParser().parseFromString(
+            '<!doctype html><body>' + node.textContent,
+            'text/html'
+        );
+        html = dom.body.innerHTML;
+    }
     return {
-        html: dom.body.innerHTML,
+        html,
         ...getAdvanced(node)
     };
 };


packages/pagebuilder/lib/ContentTypes/Html/configAggregator.js
```diff
@@ -1,12 +1,23 @@
...
 
 export default node => {
-    const dom = new DOMParser().parseFromString(
-        '<!doctype html><body>' + node.textContent,
-        'text/html'
-    );
+    let html;
+    if (node.dataset.decoded) {
+        html = node.innerHTML;
+        if (process.env.NODE_ENV !== 'production') {
```
sirugh: I think we remove logs in production bundles anyways - shouldn't need to gate with a condition. If I'm wrong we probably have to go clean up our app :D

zetlen: We might, but I just wanted to be explicit. This is a pattern we ought to follow. We may later remove the drop_console in some cases!

dpatil-magento:
Works fine. Tested this PR against - 
1. 2.3.5 with core fix.
2. 2.3.5 without core fix.
3. 2.3.5 with Core fix and without PB data changes (PWA home page)
4. 2.3.4 
5. 2.3.4 without PB data changes (PWA home page)

PR #2285, opened 2020-03-25
[PWA-244] Shipping Information (Guest)

tjwiebell:
## Description

As an unauthenticated user I want to enter my data in order to complete my purchase. Form fields should include (see mockups):
 * Shipping address
 * Edit shipping address
 * Double check that headers adhere to mock

*Important:*
* Do not display temp data from shipping estimate fields. Compare gql response to `MOCKED_ADDRESS` to verify.

*Nice to have: *

* Auto city/state detection by zip code entry  (is this available through existing API) 
* UX would like to see the different sections numbered if possible. Include in this scope of doesn't present any problems.

eg.
1. Shipping Information
2. Shipping Method
3. Payment Information

### Developer Note
Included in this scope is the decomposition of Country and Region fields into their own components, extracted from the Estimate Shipping Methods components on the CartPage. 

## Related Issue

* [[PWA-244](https://jira.corp.magento.com/browse/PWA-244)] Shipping Information (Guest)

### Verification Steps

This PR decomposed Country and Region into its own components, so we'll need to test both `/cart` Estimate Shipping functionality and Checkout
**Cart Verification**
1. Add a few items to cart of different types
2. Navigate to `/cart` and open Estimate Shipping panel
3. Verify that Country still defaults to US and all options loading w/o Error
4. Verify that Region loaded state options
5. Switch to a country with no region options and verify text input displayed
6. Select France and verify new set of regions is loaded
7. Switch back to US, enter State/ZIP, and estimate your shipping. Verify no errors.

**Checkout Verification**
1. Go to `/checkout`. Shipping form should automatically render with values from cart pre-filled
2. Enter Shipping Information, verify that it saves, and the card view is then rendered with entered data and it automatically opened the Shipping Methods step (this should go directly to Payment once Shipping Methods PR is merged, since you already selected from Cart)
3. Edit shipping information, verify everything saves and is re-rendered correctly
4. Repeat these steps for mobile viewports (only difference is a drawer instead of modal for edit, and first/last should collapse to single column)

Screen Shot 2020-03-26 at 3 38 23 PM
Screen Shot 2020-03-26 at 3 38 46 PM
Screen Shot 2020-03-26 at 3 38 54 PM

## Checklist


dpatil-magento:
**Performance Test Results**

https://pr-2285.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.74

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/useShippingInformation.js
```diff
@@ -0,0 +1,79 @@
...
+                    email,
+                    ...primaryAddress
+                };
+            }
+        }
+
+        return filteredData;
+    }, [data]);
+
+    const doneEditing = shippingData && shippingData.email;
```
sirugh: This is potentially flaky - while we don't require an email in cart right now it's possible we may add it later which would cause this form to render the card view. Also considering auth checkout that may have an email pre-attached to the cart but not other info. I know there is probably a follow up ticket for auth but I wanted to call this out here.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/shippingInformation.js
```diff
@@ -1,50 +1,59 @@
...
+    if (loading) {
+        return <span>Loading...</span>;
+    }
 
-    /**
-     * TODO
-     *
-     * Change this to reflect diff UI in diff mode.
-     */
     const shippingInformation = doneEditing ? (
```
sirugh: It occurs to me that `doneEditing` should really be something like `submittedInitialData` or something else indicative of a difference between the initial edit and subsequent "updates".

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/shippingInformation.css
```diff
@@ -1,24 +1,31 @@
...
 
-.container_edit_mode {
-    composes: container;
+.root--editMode {
+    composes: root;
     border: none;
     border-radius: 0px;
     border-bottom: 1px solid rgb(var(--venia-border));
-    padding: 1.5rem 0rem;
+    padding: unset;
```
sirugh: `padding: 0`?

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/card.css
```diff
@@ -0,0 +1,19 @@
...
+    row-gap: 0.5rem;
+    padding-top: 1rem;
+}
+
+.address {
+    display: grid;
+    gap: 0.5rem;
+    grid-template-areas:
+        '.'
+        'area';
```
sirugh: I think we can accomplish this vertical alignment without the use of `areas`. 

`grid-template-columns: 1fr;` should do the trick and allow you get rid of the added class within the address div.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/editModal.css
```diff
@@ -0,0 +1,73 @@
+.root {
+    background-color: white;
+    height: auto;
+    left: calc(50% - 35vw);
+    max-height: 85vh;
+    max-width: 70vw;
```
sirugh: The desktop modal should probably have a pixel max-width. I know an ultrawide isn't the best example of the average user but even a high resolution/large monitor will force the input fields to be extremely wide.

[Image from Gyazo](https://gyazo.com/3f86f32fc5067e727ca73a5ab1bcf2c0)

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
```diff
@@ -0,0 +1,133 @@
...
+                classes={{ root: classes.field }}
+                id="street1"
+                label="Street Address 2"
+            >
+                <TextInput field="street[1]" />
+            </Field>
+            <Field classes={{ root: classes.field }} id="city" label="City">
+                <TextInput field="city" validate={isRequired} />
+            </Field>
+            <Region classes={{ root: classes.field }} validate={isRequired} />
```
sirugh: Using `Ukraine` which has no regions and then trying to estimate yields form validation errors. A similar error occurs in the edit form.

[Image from Gyazo](https://gyazo.com/dfaa5690df6fa7c20fa3f64fb6cfa8a7)

Can we conditionally require region/postcode based on the length of the region response data?

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/EditForm/useEditForm.js
```diff
@@ -0,0 +1,57 @@
...
+    const { code: countryCode } = country;
+    const { code: regionCode } = region;
+
+    const initialValues = {
+        ...shippingData,
+        country: countryCode,
+        region: regionCode
+    };
+
+    const isUpdate = !!shippingData.email;
```
sirugh: A similar note here as earlier -- is presence of email enough to know if this is an update or an initial input?

packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -0,0 +1,38 @@
...
+    let formattedRegionsData = [];
+    if (!loading && !error) {
+        const { country } = data;
+        const { available_regions: availableRegions } = country;
+        if (availableRegions) {
+            formattedRegionsData = availableRegions.map(region => ({
+                key: region.id,
+                label: region.name,
+                value: region.code
+            }));
```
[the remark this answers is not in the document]
tjwiebell: Inconsistent responses from GraphQL; country seems to be in a random order, regions came back alphabetized.

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/useShippingInformation.js
```diff
@@ -0,0 +1,79 @@
...
+
+    useEffect(() => {
+        if (doneEditing) {
+            onSave(doneEditing);
+        }
+    }, [doneEditing, onSave]);
+
+    const handleEditShipping = useCallback(() => {
+        toggleDrawer('shipping.edit');
```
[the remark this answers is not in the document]
tjwiebell: We'll discuss tomorrow with the group. I would avoid snake case in this name, as that's only supposed to be used to indicate backend data, but the future need to use `shippingInformation.edit` makes sense.

packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -0,0 +1,38 @@
...
+            formattedRegionsData.unshift({
```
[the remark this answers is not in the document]
tjwiebell: Yes, UX wanted it to default to unselected, additionally enforcing the field requirement.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/card.css
tjwiebell: This accounts for multiple address rows without having to hard code any `address.length` logic.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/editModal.css
tjwiebell: Changed to fixed width in 456f4e8.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
tjwiebell: This is actually configuration we should pull out of the backend, but we haven't addressed that Story yet about how to properly do that (runtime vs. buildtime etc...). Sometimes region is still required, even though there are no options; sometimes there are options and its not required, each Country can be different. Deferring for now, but create a feature request, this could be a good task for community.

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/EditForm/useEditForm.js
tjwiebell: Yes, currently in the cart/checkout experience, you can't set an email any other way.

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/useShippingInformation.js
tjwiebell: This duck typing seems fine in this scope. I agree it may need to change if there are edge cases introduced by the auth flow or other interactions in the app, and should be addressed in that scope.

tjwiebell: I just saw your comment on Andy's PR, and see the need to standardize this now. This might be a good candidate for another `@client` field on the cart, so each component doesn't have to guess. Let's discuss more during our peer review tomorrow.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/card.css
sirugh: Oh you meant multiple `street` entries :) I gotcha.

tjwiebell: I'm having some grid critters flashbacks; maybe I could have done `grid-row-start: -1` instead.

packages/venia-ui/lib/components/Country/country.js
```diff
@@ -0,0 +1,41 @@
...
+            getCountriesQuery: GET_COUNTRIES_QUERY
+        }
+    });
+    const { countries, loading } = talonProps;
+    const { classes: propClasses, validate } = props;
+
+    const classes = mergeClasses(defaultClasses, propClasses);
+
+    return (
+        <Field id="country" label="Country" classes={{ root: classes.root }}>
```
revanth0212: A suggestion here, to make this component useful in all use cases, take `field` as a prop. If not provided then use `country`.  The same for `Region` component.

revanth0212: Thanks for writing this component, it is very useful.

tjwiebell: Made these adjustments in 5be51aa, good suggestion 👍

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/useShippingInformation.js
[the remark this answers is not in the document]
tjwiebell: More descriptive name added in 32ea1ca

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/shippingInformation.css
tjwiebell: Per the peer review feedback, I went through a fixed any usages of `unset`.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/card.css
tjwiebell: I don't even know why this was needed, the default flow works fine. Removed in 32ea1ca.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
sirugh: https://jira.corp.magento.com/browse/PWA-484

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
```diff
@@ -0,0 +1,163 @@
...
+            <Field
+                classes={{ root: classes.lastname }}
+                id="lastname"
+                label="Last Name"
+            >
+                <TextInput field="lastname" validate={isRequired} />
+            </Field>
+            <Country classes={{ root: classes.field }} validate={isRequired} />
+            <Field
+                classes={{ root: classes.field }}
```
jimbo, line 74: Each of these `field` classnames should be unique classnames, not shared between elements.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
```diff
@@ -0,0 +1,163 @@
...
+        </Button>
+    ) : null;
+
+    return (
+        <Form
+            className={classes.root}
+            initialValues={initialValues}
+            onSubmit={handleSubmit}
+        >
+            <Field classes={{ root: classes.field }} id="email" label="Email">
```
jimbo, line 54: Consider applying these classnames to some wrapping div elements around the fields rather than as classname overrides to the fields themselves.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -86,7 +86,7 @@ const CheckoutPage = props => {
                 <ShippingMethod onSave={setShippingMethodDone} />
             ) : (
                 <h2 className={defaultClasses.shipping_method_heading}>
-                    Shipping Method
+                    2. Shipping Method
```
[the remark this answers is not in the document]
tjwiebell: It sounded like any output should start using this pattern so we can easily identify strings to translate. Should this rule also apply to `label` props?

packages/venia-ui/lib/components/Field/message.js
```diff
@@ -5,6 +5,10 @@ import classify from '../../classify';
 import defaultClasses from './message.css';
 
 export class Message extends Component {
+    static defaultProps = {
```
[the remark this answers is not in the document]
tjwiebell: Converted in 0d829a5

packages/venia-ui/lib/components/Country/country.js
```diff
@@ -0,0 +1,48 @@
...
+Country.propTypes = {
+    classes: shape({
+        root: string
+    }),
+    field: string,
+    label: string,
+    validate: func
+};
+
+export default Country;
```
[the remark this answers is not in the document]
tjwiebell: Updated all new code to follow this pattern in 0d829a5

packages/venia-ui/lib/components/Country/country.js
```diff
@@ -0,0 +1,48 @@
...
+            getCountriesQuery: GET_COUNTRIES_QUERY
+        }
+    });
+    const { countries, loading } = talonProps;
+    const { classes: propClasses, field, label, validate } = props;
+
+    const classes = mergeClasses(defaultClasses, propClasses);
+
+    return (
+        <Field id={field} label={label} classes={{ root: classes.root }}>
```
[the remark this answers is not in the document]
tjwiebell: Extracted to props object in 0d829a5.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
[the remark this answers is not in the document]
tjwiebell: Updated in 0d829a5. Will wait to confirmation of labels.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
```diff
@@ -0,0 +1,163 @@
...
+    const talonProps = useEditForm({
+        afterSubmit,
+        ...EditFormOperations,
+        shippingData
+    });
+    const { handleSubmit, initialValues, isSaving, isUpdate } = talonProps;
+
+    const classes = mergeClasses(defaultClasses, propClasses);
+
+    const messageComponent = !isUpdate ? (
```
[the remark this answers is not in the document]
tjwiebell: Renamed in 0d829a5.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
```diff
@@ -0,0 +1,163 @@
...
+            >
+                <TextInput field="telephone" validate={isRequired} />
+            </Field>
+            <div className={classes.buttons}>
+                {cancelButton}
+                <Button
+                    classes={{
+                        root_normalPriority: classes.submit,
+                        root_highPriority: classes['submit--update']
+                    }}
```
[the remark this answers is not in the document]
tjwiebell: Extracted the whole button in 0d829a5.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
```diff
@@ -0,0 +1,163 @@
...
+        </Message>
+    ) : null;
+
+    const cancelButton = isUpdate ? (
+        <Button
+            classes={{
+                root_normalPriority: classes.submit
+            }}
+            disabled={isSaving}
+            onClick={onCancel}
```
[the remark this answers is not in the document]
tjwiebell: Done in 0d829a5.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
tjwiebell: Created unique classnames for all in 0d829a5.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/EditForm/editForm.js
tjwiebell: Wrapped in divs to avoid class overrides in 0d829a5.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2285.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.64

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/useShippingInformation.js
```diff
@@ -0,0 +1,79 @@
...
+    const [fetchShippingInformation, { called, data, loading }] = useLazyQuery(
+        getShippingInformationQuery
+    );
+
+    useEffect(() => {
+        if (cartId) {
+            fetchShippingInformation({
+                variables: {
+                    cartId
+                }
```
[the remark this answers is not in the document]
sirugh: Tommy fixed this - a user re-estimating on the cart page now causes this to display as the original edit form rather than the card/update view. This is because we resubmit the mock data when estimating so when we determine the initial state of shipping data and strip the mock data we don't have a `city`, which is the compared value that we are using to indicate the lack of data.

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/useShippingInformation.js
```diff
@@ -0,0 +1,79 @@
...
+                    email,
+                    ...primaryAddress
+                };
+            }
+        }
+
+        return filteredData;
+    }, [data]);
+
+    const doneEditing = !!shippingData && !!shippingData.city;
```
sirugh, line 61: This is fragile but works - a comment here and wherever else we use it explaining why we use `city` to indicate completeness would be nice. If someone customizes their cart form with a city field this would no longer work.

packages/peregrine/lib/talons/Country/useCountry.js
```diff
@@ -0,0 +1,24 @@
...
+
+export const useCountry = props => {
+    const {
+        queries: { getCountriesQuery }
+    } = props;
+
+    const { data, error, loading } = useQuery(getCountriesQuery);
+
+    let formattedCountriesData = [{ label: 'Loading Countries...', value: '' }];
+    if (!loading && !error) {
```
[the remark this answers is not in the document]
tjwiebell: I suspect the hook will either immediately return data (from the cache), or return `loading = true` and not enter this block; I don't think there's a render cycle possibility where `error/loading = false` and `data = null`.

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/EditForm/useEditForm.js
```diff
@@ -0,0 +1,63 @@
...
+    const { code: countryCode } = country;
+    const { code: regionCode } = region;
+
+    const initialValues = {
+        ...shippingData,
+        country: countryCode,
+        region: regionCode
+    };
+
+    const isUpdate = !!shippingData.city;
```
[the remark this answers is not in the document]
tjwiebell: Added in 14a23f7

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/useShippingInformation.js
tjwiebell: Added in 14a23f7

dpatil-magento:
QA Pass, good to merge after review approval.

PR #2288, opened 2020-03-26
[feature]: Order confirmation Page

sirugh:
## Description

The order confirmation page displays data from the order that was just placed. It's different from most of our other recent components in that it must render "past" data that is no longer stored/cached.

I approached this by prefixing a query on the submission handler that fetches _current_ order details right before placing the order. The details are then passed to the order confirmation component for rendering. The only queries/mutations that exist within the order confirmation page and its children are those around creating an account.

**Note:**

We combined the subscribe and create account forms into a single form due to restrictions from graphql.

Refer a friend is also deferred.

## Related Issue

Closes [PWA-186](https://jira.corp.magento.com/browse/PWA-186).

### Verification Steps

1. As a guest, place an order. See that the order confirmation page appears.
2. Create an account using the form. It should log you in but not redirect you for now. Eventually this will redirect to an account view or orders page.
3. As an authed user, place an order. See that the order confirmation page appears but there is no  create account form.

If shipping info/method/payment info sections are not merged you may need to manually submit this information to the cart through graphql queries. I'll provide some that should work, just replace cart id with your own from the store. 

**To add shipping address/method**

Go to the `/cart` page and estimate your shipping. That will attach a fake address as well as select a shipping method for your cart.

**To add payment method:**

```gql
mutation {
    setPaymentMethodOnCart(input: {
        cart_id: "e3ih8dF98xpiykcHkHqb9XSsppOSymMK"
        payment_method: {
            code: "checkmo"
        }
    }) {
        cart {
        selected_payment_method {
            code
        }
        }
    }
}
```

**To add billing address:**
```gql
mutation {
  setBillingAddressOnCart(
    input: {
      cart_id: "e3ih8dF98xpiykcHkHqb9XSsppOSymMK"
      billing_address: {
        address: {
          firstname: "John"
          lastname: "Doe"
          company: "Company Name"
          street: ["320 N Crescent Dr", "Beverly Hills"]
          city: "Los Angeles"
          region: "CA"
          postcode: "90210"
          country_code: "US"
          telephone: "123-456-0000"
          save_in_address_book: false
        }
      }
    }
  ) {
    cart {
      billing_address {
        firstname
        lastname
        company
        street
        city
        region{
          code
          label
        }
        postcode
        telephone
        country {
          code
          label
        }
      }
    }
  }
}
```

[Image from Gyazo](https://gyazo.com/54f108a2bb064c3ebd9ea0427d0e6d45)

[Image from Gyazo](https://gyazo.com/aa1d347742af044a5fe1f526cc7a5394)

## Checklist


packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.js
```diff
@@ -19,7 +23,8 @@ const ItemsReview = props => {
     const talonProps = useItemsReview({
         queries: {
             getItemsInCart: LIST_OF_PRODUCTS_IN_CART_QUERY
-        }
+        },
+        data: props.data
```
sirugh: I want to re-use this component's UI and other logic but not the queries. I'm open to other approaches but this one seemed the most direct. If data is passed to the talon it will use it instead of making a query.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/createAccount.js
```diff
@@ -1,21 +1,148 @@
...
+        });
+
+        addToast({
+            type: 'info',
+            message: 'Account successfully created.',
+            timeout: 5000
+        });
+    }, [addToast]);
+
+    const talonProps = useCreateAccount({
```
sirugh: Similarly, I re-used this talon but not the generic `CreateAccount` component to which it belongs. I think that's a no-no from our previous guidance. Different from `ItemsReview`, I actually want a slightly different UI than the old `CreateAccount` view. However the entirety of the logic can be re-used with exception of `isSignedIn` which this new component doesn't care about.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.gql.js
```diff
@@ -1,14 +1,57 @@
...
+                country {
+                    label
+                }
+
+                selected_shipping_method {
+                    carrier_title
+                    method_title
+                }
+            }
+            ...ItemsReviewFragment
```
sirugh: I reused a fragment here and could possibly do the same for the other pieces of information that are needed, once those PRs are merged.

packages/peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/useCreateAccount.js
```diff
@@ -0,0 +1,147 @@
...
+import { useCartContext } from '@magento/peregrine/lib/context/cart';
+import { useAwaitQuery } from '@magento/peregrine/lib/hooks/useAwaitQuery';
+
+/**
+ * Returns props necessary to render CreateAccount component. In particular this
+ * talon handles the submission flow by first doing a pre-submisson validation
+ * and then, on success, invokes the `onSubmit` prop, which is usually the action.
+ *
+ * This talon is almost identical to the other useCreateAccount but does not
+ * return `isSignedIn`.
```
sirugh, line 13: Read this.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -1,34 +1,121 @@
...
+    });
+
+    useEffect(() => {
+        window.scrollTo({
+            left: 0,
+            top: 0,
+            behavior: 'smooth'
+        });
+    }, []);
+
```
brendanfalkowski, line 51: @sirugh — Just submitted a PR that avoids needing code to scroll to top in every module on route change. See #2309

sirugh: Awesome! I'll leave this here for now and we can go clean up extra stuff later depending on which PR gets merged first :) I will say that this isn't a route, so it may not actually trigger the effect.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/createAccount.js
```diff
@@ -1,23 +1,157 @@
...
+import GET_CUSTOMER_QUERY from '../../../queries/getCustomer.graphql';
+import SIGN_IN_MUTATION from '../../../queries/signIn.graphql';
+
+const CreateAccount = props => {
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    const [, { addToast }] = useToasts();
+
+    const onSubmit = useCallback(() => {
+        // TODO: Redirect to account/order page when implemented.
```
supernova-at, line 33: Is this TODO still valid?

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/createAccount.js
```diff
@@ -1,23 +1,157 @@
...
+        onSubmit
+    });
+
+    const { errors, handleSubmit, isDisabled, initialValues } = talonProps;
+
+    // Map over any errors we get and display an appropriate error.
+    const errorMessage = errors.length
+        ? errors
+              .map(({ message }) => message)
+              .reduce((acc, msg) => msg + '\n' + acc, '')
```
supernova-at, line 71: Consider using [join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) for this.

supernova-at:
Two other comments that didn't really have a line number to go with them:

1. This PR adds a new dependency `graphql-anywhere`
2. Do we want the order confirmation page to live inside the `CheckoutPage` ?

sirugh:
> This PR adds a new dependency graphql-anywhere

Correct - this allows us to define propTypes using fragment data much easier than listing them all out.

> Do we want the order confirmation page to live inside the CheckoutPage ?

This was something we talked about in review. It is possible to define a route like `/receipt` that could display order information but that requires we pass state using react router like `history.push('/receipt', orderDetails)`. First, I'm not sure if this is secure. Right now the order data is purely in memory and when you refresh the page after checkout the details disappear. If we passed with router I believe it will store state in memory which means another vector for accessing PII. We also have to provide the order data from the active cart **before** we submit the cart and go to the receipt page. So this data MUST be fetched in the `CheckoutPage` component and passed through props or state to the receipt page.

It gets pretty convoluted so rather than try to implement a route for it I simply return `OrderConfirmation` component without any of the wrapping DOM/CSS so it's at least separated in that way.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/createAccount.js
sirugh: Yes.

packages/peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/useOrderConfirmationPage.js
```diff
@@ -0,0 +1,34 @@
+import { useUserContext } from '../../../context/user';
+
+export const flatten = data => {
```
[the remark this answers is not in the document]
sirugh: Yep. @jimbo requested that in our team review

packages/peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/useOrderConfirmationPage.js
```diff
@@ -0,0 +1,34 @@
...
+    const { cart } = data;
+    const { shipping_addresses } = cart;
+    const address = shipping_addresses[0];
+
+    const shippingMethod = `${
+        address.selected_shipping_method.carrier_title
+    } - ${address.selected_shipping_method.method_title}`;
```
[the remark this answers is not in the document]
sirugh: Correct - `data` must be truthy, but I can define this in the propTypes.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -1,34 +1,121 @@
...
+        email: string,
+        name: string,
+        addressAdditional: string,
+        shippingMethodHeading: string,
+        shippingMethod: string,
+        itemsReview: string,
+        additionalText: string,
+        sidebarContainer: string
+    }),
+    data: propType(GET_ORDER_DETAILS),
```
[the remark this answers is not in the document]
sirugh: Yep! It's a little confusing that you have to pass the query and not the fragment but basically this is saying that `data` should match the result of this query.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -1,34 +1,121 @@
...
-            <Subscribe />
-            <div className={classes.email_review_container}>
-                <div className={classes.email_review_text}>
+                <div
+                    className={classes.orderNumber}
+                >{`Order Number: ${orderNumber}`}</div>
+                <div className={classes.shippingInfoHeading}>
+                    Shipping Information
+                </div>
+                <div className={classes.shippingInfo}>
```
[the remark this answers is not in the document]
sirugh: No, unfortunately the "completed" views of these components do not match the UX for the order confirmation page. They are just text blocks anyways so it's not like we gain that much by sharing the code. The ItemsReview component was pretty involved and had some logic for the UX which is why I reused it.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -1,34 +1,121 @@
...
+                    >{`${firstname} ${lastname}`}</span>
+                    {streetRows}
+                    <span
+                        className={classes.addressAdditional}
+                    >{`${city}, ${region} ${postcode} ${country}`}</span>
+                </div>
+                <div className={classes.shippingMethodHeading}>
+                    Shipping Method
+                </div>
+                <div className={classes.shippingMethod}>{shippingMethod}</div>
```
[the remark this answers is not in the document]
sirugh: Yea - I also would not suggest rewriting the components until we HAVE to do it. As we develop iteratively we don't tend to make sweeping refactors that often so I think allowing "static" data to all talons might be one of those things we end up doing in a big PR at some later time.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.css
```diff
@@ -1,97 +1,67 @@
...
+
     display: grid;
-    gap: 1.5rem 2rem;
+    gap: 2rem;
     grid-template-columns: 2fr 1fr;
-    grid-template-areas:
-        'heading create_account'
-        'order_summary create_account'
-        'items_review subscribe'
-        'email_review subscribe';
```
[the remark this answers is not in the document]
sirugh: The page is now simplified into a grid with two columns: main and side. The children position automatically within those columns so the best I could do for you is to add areas to those columns which I don't think is helpful.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/createAccount.js
sirugh: Done in ff792f8

packages/peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/useOrderConfirmationPage.js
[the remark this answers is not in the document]
sirugh: Which I didn't do....

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.css
```diff
@@ -1,97 +1,67 @@
...
-.heading_container {
-    grid-area: heading;
-    display: flex;
-    align-items: baseline;
-    justify-content: space-between;
-    margin-bottom: 1.5rem;
-}
+.mainContainer {
+    /* Positioned in CheckoutPage's grid */
+    grid-column: 1 / span 1;
```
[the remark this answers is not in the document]
sirugh: Done in ff792f8

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.css
```diff
@@ -1,97 +1,67 @@
...
 }
 
-.subscribe_container {
-    grid-area: subscribe;
-    min-height: 10rem;
-}
+.sidebarContainer {
+    /* Positioned in CheckoutPage's grid */
+    grid-column: 2 / span 1;
+    grid-row: 1 / span 1;
```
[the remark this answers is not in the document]
sirugh: Done in ff792f8

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -1,34 +1,121 @@
...
+                <div className={classes.itemsReview}>
+                    <ItemsReview data={data} />
+                </div>
+                <div className={classes.additionalText}>
                     You will also receive an email with the details and we will
                     let you know when your order has shipped.
```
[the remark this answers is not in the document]
sirugh: Done in ff792f8

supernova-at:
Looks good to me. Probably want @jimbo 's sign off on a new dependency though.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -1,34 +1,122 @@
-import React from 'react';
+import React, { useEffect } from 'react';
+import { propType } from 'graphql-anywhere';
```
jimbo, line 2: Maybe rename this on import? Open to suggestions here, but I find this pretty vague.

```suggestion
import { propType as queryData } from 'graphql-anywhere';
```

jimbo:
As for the new dependency, it's fine as long as it's only being used for its `propType` export. Would like confirmation that it's not being included in the production bundle though.

sirugh:
You can see that most of the code is not included. Not sure why the small amount is leftover in prod.

Develop:

[Image from Gyazo](https://gyazo.com/d1947a597d6b66dbb524321199dd526d)

Prod:

sirugh:
Couldn't find the source code for `graphql-anywhere` and since there is 1.5kb in the prod bundle for some reason we removed the dependency. Goodnight sweet prince.

dpatil-magento:
@sirugh Pls check the conflict and following observation > on clicking `Create Account` button.

image

PR #2303, opened 2020-04-01
[bugfix]: check types in isRequired validate rule (#2302)

### LINKED ISSUE #2302: [bug]: checkbox doesn't work with isRequired rule
**Describe the bug**
rule isRequired from `venia-ui/lib/util/formValidators.js` doesn't work with something else than string. 

**To reproduce**
1. Use checkbox in your templates with isRequired validate rule
```js
<Checkbox field="agreement" label="check this" validate={isRequired} validateOnChange/>
```

**Expected behavior**
Validate rule isRequired works with checkbox also.

**Additional context**
TypeError: (value || "").trim is not a function

**Possible solutions**
We should check types at least - not every field is string. 

Device Mac
OS Catalina 10.15.2
Browser Chrome
Browser Version 79.0.3945.130
Magento Version:
PWA Studio Version: develop
NPM version 6.13.4
Node Version v10.18.0

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

mercs600:
## Description
I added checking types for isRequired rule to detect true value also 
## Related Issue

Closes #2302 

### Verification Steps
Just add checkbox with isRequired rule
```html
<Checkbox field="agreement" label="check this" validate={isRequired} validateOnChange/>
```


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2303.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.66

jimbo:
@mercs600 Hello, thanks for your contribution!

If I understand the intent correctly here, you're proposing to use the existing `isRequired` validator to ensure that a checkbox has a *specific* value for `checked`: `true`. When it comes to other fields, the `isRequired` is only used to ensure that a field has *any* value. I can see how you might construe this as being the case for binary fields such as checkboxes, but I think the distinction does have some meaning here.

Instead of modifying `isRequired` to have an alternate implication for checkboxes, why not create a new validator, `isChecked`, that ensures the value is *specifically* boolean `true`? I might prefer that solution.

brendanfalkowski:
@jimbo It also fails if the value is a number (not string as expected). We have some cases where a select input needs that. See: https://github.com/magento/pwa-studio/issues/2302#issuecomment-607400386

We ran into both of these issues and modified `isRequired` to handle them. I'm not really opposed to `isChecked` and `isRequired` being separate validators, but it's nice having only one thing to remember. I can't think of a case where "mustBeChecked" would not mean "isRequired".

mercs600:
I totally agree with @brendanfalkowski we should have one method to check field as required - there is no difference between isCheckboxRequired or isRequired. What stop us to merge it ?

tjwiebell:
Doesn't the need for this PR just reveal invalid markup in your code? The checkbox value is of type `DOMString`, so there should be no need to handle a boolean type. I'll give @jimbo the final call here, but I see a couple reasons here why this isn't something we would want in core.

mercs600:
@tjwiebell yes - if it comes to regular checkbox, I used checkbox from venia-ui where we can see

packages/venia-ui/lib/components/Checkbox/checkbox.js: 

```js
export class Checkbox extends Component {
    static propTypes = {
       .....
        fieldState: shape({
            value: bool
        }).isRequired,
        ....
    };
```

Additionally it based on `BasicCheckbox` from informed:
https://github.com/joepuzzo/informed/blob/master/src/components/form-fields/Checkbox.js
Where you can see their set values is based on `checked` attribute what is `bool` also.

tjwiebell:
@mercs600 - I was not aware of Informed's casting, boolean type is not a concern then, thanks for pointing that out; probably still valid for the other edge case for `number`. 

So, we will accept a new validator `isChecked` without question, as it shows clear intent. Ergonomics is a good reason to merge, but I don't see that convincing @jimbo to change his mind. Happy to continue discussing if there are other good reasons for this rule to serve two unique use cases.

jimbo:
> It also fails if the value is a number (not string as expected). We have some cases where a select input needs that. See: [#2302 (comment)](https://github.com/magento/pwa-studio/issues/2302#issuecomment-607400386)
> 
> We ran into both of these issues and modified `isRequired` to handle them. I'm not really opposed to `isChecked` and `isRequired` being separate validators, but it's nice having only one thing to remember. I can't think of a case where "mustBeChecked" would not mean "isRequired".

@brendanfalkowski Yes, I agree that the `isRequired` validator should be updated to support cases where `value` is a number. We should cast `value` to a string, trim it, and measure it—and probably unit test the new edges.

> So, we will accept a new validator `isChecked` without question, as it shows clear intent. Ergonomics is a good reason to merge, but I don't see that convincing @jimbo to change his mind. Happy to continue discussing if there are other good reasons for this rule to serve two unique use cases.

Honestly, I can be convinced on this one. The issue here is that `isRequired` is inherently ambiguous; you can't say that an unchecked checkbox is devoid of user input, but you also can't say an empty text field is devoid of input, either.

I would recommend avoiding `isRequired` entirely, in favor of more specific validators such as `hasLengthAtLeast` or `mustBeChecked`, which are unambiguous. But if people are going to use it anyway, and they prefer expected output over accurate output, I can accept that.

Unit tests for the new edges and I think we'll be good.

brendanfalkowski:
> I would recommend avoiding isRequired entirely, in favor of more specific validators such as hasLengthAtLeast or mustBeChecked, which are unambiguous. But if people are going to use it anyway, and they prefer expected output over accurate output, I can accept that.

@jimbo Point taken on specific validators. I think I'd still prefer the catch-all `isRequired` validator because that feels _de rigueur_ for every form validation library I remember using.

I'd say 50% of our form inputs are these same props, so it's nice having one phrasing to remember even if `validate={hasLengthAtLeast(1)}` is more accurate.

``` jsx
<Field fieldId="phone" label="Phone" required={true}>
    <TextInput
    field="phone"
    id="phone"
    validate={isRequired}
    validateOnBlur />
</Field>
```

supernova-at:
Hi @mercs600 , I have been assigned to review this but it seems some time has elapsed since any discussion here.

It looks like there are a few remaining items to be addressed here:
  1. Add an `isChecked` or `mustBeChecked` validator
  1. Update the `isRequired` validator that composes the `isChecked` / `mustBeChecked` validator
  1. Update the relevant unit tests

Kindly let me know if you would like to perform this work. If not I can help move this across the finish line 👍

packages/venia-ui/lib/util/formValidators.js
```diff
@@ -32,8 +32,32 @@ export const hasLengthExactly = (value, values, length) => {
...
+ */
 export const isRequired = value => {
-    return (value || '').trim() ? SUCCESS : 'The field is required.';
+    const FAILURE = 'The field is required.';
+
+    // If it's a boolean, it must be `true`.
+    if (typeof value === 'boolean') {
+        if (!value) return FAILURE;
+        return SUCCESS;
+    }
```
jimbo, line 46: This is fine, but this would be clearer if we just let it fall through to success.

```es6
// if it's a boolean, it must be true
if (value === false) return FAILURE

// if it's a number or string
// ... return FAILURE

return SUCCESS
```

packages/venia-ui/lib/util/formValidators.js
```diff
@@ -32,8 +32,32 @@ export const hasLengthExactly = (value, values, length) => {
...
+
+    // If it is a number or string, it must have at least one character of input (after trim).
+    let stringValue = String(value);
+    stringValue = stringValue.trim();
```
jimbo, line 50: I'm not sure we need the reassignment.

```es6
const stringValue = String(value).trim()
```

packages/venia-ui/lib/util/formValidators.js
```diff
@@ -32,8 +32,32 @@ export const hasLengthExactly = (value, values, length) => {
...
+    const measureResult = hasLengthAtLeast(stringValue, null, 1);
+
+    if (measureResult) return FAILURE;
+    return SUCCESS;
+};
+
+export const mustBeChecked = value => {
```
jimbo, line 57: Thanks for adding this.

packages/venia-ui/lib/util/formValidators.js
```diff
@@ -32,8 +32,32 @@ export const hasLengthExactly = (value, values, length) => {
...
 
+/**
+ * isRequired is provided here for convenience but it is inherently ambiguous and therefore we don't recommend using it.
+ * Consider using more specific validators such as `hasLengthAtLeast` or `mustBeChecked`.
+ */
 export const isRequired = value => {
-    return (value || '').trim() ? SUCCESS : 'The field is required.';
+    const FAILURE = 'The field is required.';
+
+    // If it's a boolean, it must be `true`.
```
jimbo, line 42: Is it possible for a checkbox field to hold a value of `undefined`? Is it always `true` or `false`?

supernova-at:
PR Updated:
  * All feedback addressed in https://github.com/magento/pwa-studio/pull/2303/commits/55e5267c8a336ea78a1b4375fe9d0ff8a74277e2.

supernova-at:
> Is it possible for a checkbox field to hold a value of undefined? Is it always true or false?

Man, great catch - it absolutely can be `undefined`. Handled it and added tests for this case in https://github.com/magento/pwa-studio/commit/55e5267c8a336ea78a1b4375fe9d0ff8a74277e2.

> The field is required has always stood out to me

I'm so glad you pointed this out, I almost changed it on my own and your comment helped me decide to do it. Changed it to "Is required" in https://github.com/magento/pwa-studio/commit/55e5267c8a336ea78a1b4375fe9d0ff8a74277e2.

PR #2307, opened 2020-04-03
Fixing css var in the checkbox.css

### LINKED ISSUE #2306: [bug]: Checkbox.css has an error in a variable usage
**Describe the bug**
It has a small bug on the usage of a variable in the checkbox.css

**To reproduce**
1. Go to `packages/venia-ui/lib/components/Checkbox/checkbox.css`, line 23

This code has a syntax error `z-index: var(--base-z-index + 1, 1);`

**Expected behavior**
Just use the --base-z-index variable with a default value like `z-index: var(--base-z-index, 1);`

**Possible solutions**
`z-index: var(--base-z-index, 1);`

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

LucasCalazans:
## Description

Small fix on the usage of CSS variable in the checkbox.css file

## Related Issue
Closes #2306.

### Verification Steps

1. Use a Checkbox in somewhere in the page.
2. Click in the checkbox to be checked.
3. Inspect the span that wraps the check Icon.
4. Check this `z-index` property

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/Checkbox/checkbox.css
+++ b/packages/venia-ui/lib/components/Checkbox/checkbox.css
@@ -20,7 +20,7 @@
     height: 1.25rem;
     justify-content: center;
     width: 1.25rem;
-    z-index: var(--base-z-index + 1, 1);
+    z-index: var(--base-z-index, 1);
 }
 
 .input {
--- a/packages/venia-ui/lib/components/Checkbox/checkbox.css
+++ b/packages/venia-ui/lib/components/Checkbox/checkbox.css
@@ -20,7 +20,6 @@
     height: 1.25rem;
     justify-content: center;
     width: 1.25rem;
-    z-index: var(--base-z-index, 1);
 }
 
 .input {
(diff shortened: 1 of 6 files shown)


packages/venia-ui/lib/components/Checkbox/checkbox.css
```diff
@@ -20,7 +20,7 @@
     height: 1.25rem;
     justify-content: center;
     width: 1.25rem;
-    z-index: var(--base-z-index + 1, 1);
+    z-index: var(--base-z-index, 1);
```
jimbo, line 23: Good catch, this was certainly an error, and this does fix it.

However, I think while we're here, I'd prefer to set this to the lowest possible value, if at all. One of the following solutions would be preferable, I think:

- Set `z-index` directly to `1` here, **or**
- Render the icon element *after* the input element and delete this `z-index`

In the latter option, the DOM order alone will cause them to stack properly. Generally we want our DOM order to default to the most reasonable stacking, so that's probably the best choice here. But I would also accept the former solution.

LucasCalazans: Yeah, I also agree removing the z-index is better.

jimbo:
Thanks for making the requested changes. Moving this to QA for validation.

PR #2309, opened 2020-04-03
[feat] Always scroll to top on router change

### LINKED ISSUE #2289: [feature]: Always scroll to top on a router change
When a long page links to a short page, the former scroll-y position doesn't exist, so the browser keeps the lowest scroll position possible. This means links from anywhere to anywhere can often land on the footer of a new page render.

When routing (link or redirect) the page should always scroll to top so the user's context is reset.

**Additional context**
A fix was added for this in #832 but it only affects PDP. We have dozens of modules in our build it affects all of them, so the solution should be implemented at the router-level.

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

brendanfalkowski:
## Description

On router changes, the page should always scroll to top.

This also removes prior attempts to manage scrolling-to-top per route. This should be a universal behavior.

## Related Issue
Closes #2289.

### Verification Steps

1. Go to any page.
2. Scroll down.
3. Click a link.
4. The page should scroll up to top.

## Checklist


brendanfalkowski:
I removed `window.scrollTo` from the PDP since that is handled by the global routing behavior.

I did not remove `window.scrollTo` from the category/search areas because some changes like sorting don't invoke the router. These inline calls are still needed.

packages/venia-ui/lib/components/ScrollToTop/scrollToTop.js
```diff
@@ -0,0 +1,24 @@
+import React, { useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
+
+// See: https://reacttraining.com/react-router/web/guides/scroll-restoration/scroll-to-top
+
+// Note: using { behavior: 'smooth' } in scrollTo() options often won't have an
+// effect because the re-render happens before smooth scrolling starts. Some
+// browsers try to do this internally but it's not universal.
+
+const ScrollToTop = () => {
```
sirugh, line 10: Now that we're migrated to function components this should be a peregrine hook, `useScrollTop`. It could even be `useScrollTopOnChange` and take a parameter like `pathname` which it uses in the effect. Then any component could configure when it should change such as `useScrollTopOnChange(user.isSignedIn)`, etc.

Could also have `useScrollTopOnMount` that scrolls only on mount of a component.

sirugh: I just saw Jimmy said basically the same thing a few days ago 😆

packages/venia-ui/lib/components/Routes/routes.js
```diff
@@ -1,15 +1,19 @@
...
+import { useScrollTopOnChange } from '@magento/peregrine/lib/hooks/useScrollTopOnChange';
 
 const CartPage = lazy(() => import('../CartPage'));
 const CheckoutPage = lazy(() => import('../CheckoutPage'));
 const CreateAccountPage = lazy(() => import('../CreateAccountPage'));
 const Search = lazy(() => import('../../RootComponents/Search'));
 
 const Routes = () => {
+    const { pathname } = useLocation();
+    useScrollTopOnChange(pathname);
```
sirugh, line 15: Some other examples I saw out on the web included also observing `search`. I'm not sure if we want to implement that here or in the components that actually care about search such as pagination or the search/filter components.

supernova-at: I think it makes sense to do it here. Those components are changing `search` but scrolling is a page-level concern; I don't think they should know / care / control the scrolling.

supernova-at: I think it makes sense to do it here. Those components are changing `search` but scrolling is a page-level concern; I don't think they should know / care / control the scrolling.

jimbo: Agree with @supernova-at here. I wouldn't scroll to top when search params change.

sirugh: > We'd have to accept that it'll happen in every case where we ever change search...and I don't have that level of confidence. In contrast, I do have that level of confidence when it comes to pathname.

To be specific, the `Router` component can confidently control scrolling on path changes but not search changes. That said, other components such as the Category or Search pages should be able to utilize the new hook to handle scrolling to top as they do already.

```js
// CURRENT
const Category = props => {
    const { search } = useLocation();
    const filtersFromSearch = getFilters(search);
    useEffect(() => {
        runFilterQuery()
        window.scrollTop();
    }, [filtersFromSearch])    
}
// PROPOSED
const Category = props => {
    const { search } = useLocation();
    const filtersFromSearch = getFilters(search);
    useScrollTopOnChange(filtersFromSearch);
    useEffect(() => {
        runFilterQuery()
    }, [filtersFromSearch])
}
```

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2309.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.65

PR #2313, opened 2020-04-05
[feature] Pass rest attributes to radio group items

### LINKED ISSUE #2270: [bug]: RadioGroup cannot pass a disabled attribute to the Radio component
**Describe the bug**
When using the RadioGroup component, we cannot specify if its items are disabled or not.

https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/RadioGroup/radioGroup.js#L34
In the RadioGroup component, the items only accepts a label and a value.

**Expected behavior**
We should have a way to render the Radio items as disabled or not.

**Possible solutions**
We could pass additional attributes to the item via an _attributes_ key of the items prop.
```
const options = items.map(({ label, value, attributes }) => (
    <Radio
        classes={{
            label: classes.radioLabel,
            root: classes.radio
        }}
        key={value}
        label={label}
        value={value}
        {...attributes}
    />
));

/* --- */

const items = [
    { label: 'Enabled item', value: 'value_1' },
    {
        label: 'Disabled item',
        value: 'value_2',
        attributes: { disabled: true }
    }
];

/* --- */

<RadioGroup
    field="disabled_example"
    items={items}
/>
```

 - Magento Version: 2.3.4
 - PWA Studio Version: 5.0.1
 - NPM version: `6.11.3`
 - Node Version: `10.17.0`

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

LucasCalazans:
## Description

Forwarding additional attributes to Radio component using RadioGroup.

## Related Issue
Closes #2270.

### Verification Steps
1. Use the RadioGroup on a page.
2. Pass other attributes in the item object like this:
```
<RadioGroup items={[
    {
        value: 'value 1',
        label: 'label 1',
        disabled: true
    },
    {
        value: 'value 2',
        label: 'label 2',
        role: 'radio'
    }
]}/>
```
3. Check the new attributes rendered in the page. The first should be disabled the second should be enabled.

image

## Checklist


packages/venia-ui/lib/components/RadioGroup/radioGroup.js
```diff
@@ -31,15 +31,15 @@ export class RadioGroup extends Component {
...
+        const options = items.map(({ value, ...item }) => (
             <Radio
                 classes={{
                     label: classes.radioLabel,
                     root: classes.radio
                 }}
                 key={value}
-                label={label}
                 value={value}
+                {...item}
```
sirugh, line 42: Do we need to allow for a spread here? I feel the API is pretty limited. If you just need to allow the `disabled` prop then maybe that's all we need to add. If you want to future proof this a bit more, then spreading does accomplish the same, though it is less documentable as you can include `disabled` in the propTypes.

If you do keep this, spread `item` first and then provide the other values such as classes, key, and value.

```js
<Radio
   {...item}
   classes={...}
   key={value}
   value={value}
</Radio>
```

PR #2320, opened 2020-04-07
Checkout Page Payment Information

revanth0212:
## Description

Guest Payment Workflow for the new Venia Checkout workflow.

As a guest, I should be able to add a credit card in the checkout page and use it for the checkout process.

Without a payment method selected, the review order button will be disabled. Once the user clicks the review order button, the braintree drop-in will make a request for the payment nonce and save it in the apollo cache.

If a payment nonce is present in the cache, it is an indication of the payment information step being complete. Hence the payment information summary should be shown instead.

The payment information summary component should have an edit option to edit the original payment method's data. Note: For security purposes, credit card details will not be stored. They should be entered again to complete the payment edit workflow.

This component saves some stuff in the cache and some on the remote GQL server. It stores payment details like type of card and last 4 digits in the cache but the nonce in the remote server due to incompatibilities between what GQL supports and what UX wants.

Sample Credit Cards for testing: https://developers.braintreepayments.com/guides/credit-cards/testing-go-live/ruby

Deployed version: http://pr-2320.pwa-venia.com

## Related Issue
Closes [PWA-183](https://jira.corp.magento.com/browse/PWA-183)

### Verification Stakeholders
@soumya-ashok 
@schensley 
@jimbo 

### Verification Steps
1. Go to the Checkout page.
2. Complete the first 2 steps.
3. The payment information section should show up.

image

image

image

## Checklist
1. Once order confirmation PR is merged, we will have to make sure that it is deleting the payment nonce and billing address data from the cache that this component is adding.


packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
```diff
@@ -0,0 +1,125 @@
...
+             */
+            <div
+                id={'addressSummary'}
+                className={classes.address_summary_container}
+            >
+                <div>
+                    <span className={classes.first_name}>
+                        {billingAddress.firstName}
+                    </span>
+                    <span className={classes.last_name}>
```
sirugh, line 35: I think this could be simplified. Check out [how Tommy did it](https://github.com/magento/pwa-studio/pull/2285/files#diff-746541a4b0f7d43762fed892f9cec1fa). You just need one parent container and then some child spans that include template strings:

```js
<div className={classes.addressSummary>
  <span className={classes.name}{`${firstname lastname}`}</span>
  ...
  ...
</div>
```

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
```diff
@@ -0,0 +1,125 @@
...
+                </div>
+            </div>
+        ) : null;
+
+    return (
+        <div className={classes.root}>
+            <div className={classes.heading_container}>
+                <h5 className={classes.heading}>Payment Information</h5>
+                <button className={classes.edit_button} onClick={onEdit}>
+                    <Icon size={16} src={EditIcon} attrs={{ fill: 'black' }} />
```
sirugh, line 67: Do you need fill here? I think we talked about this in the developer review but I don't remember what we said. If you do need it, this doesn't seem to be applying the fill anyways.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
```diff
@@ -0,0 +1,125 @@
...
+                </button>
+            </div>
+            <div className={classes.card_details_container}>
+                <span className={classes.payment_type}>Credit Card</span>
+                <span className={classes.payment_details}>{`${
+                    paymentNonce.details.cardType
```
sirugh, line 73: Pull out multi-line template strings into a const.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
```diff
@@ -0,0 +1,125 @@
...
+    const classes = mergeClasses(defaultClasses, propClasses);
+
+    const talonProps = useSummary({ operations: summaryGQLOperations });
+
+    const { billingAddress, isBillingAddressSame } = talonProps;
+
+    const billingAddressSummary =
+        !isBillingAddressSame && billingAddress ? (
+            /**
+             * Added ID for tests
```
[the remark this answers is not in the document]
revanth0212: I was not able to do that because of mergeClasses but it looks like once I mock it I can do like you mentioned above.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -0,0 +1,120 @@
...
+        ? paymentNonceData.cart.paymentNonce
+        : null;
+
+    /**
+     * Helper Functions
+     */
+
+    const handleReviewOrder = useCallback(() => {
+        setShouldRequestPaymentNonce(true);
+        if (onSave) {
```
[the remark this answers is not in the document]
revanth0212: Got it. Then we will have to move the `onSave` to `handlePaymentSuccess` in `creditCardPaymentMethod` component and similarly other payment methods in the future. That way we only call onSave once the payment method has been accepted. If there has been an error, the user will stay in the edit mode.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
```diff
@@ -0,0 +1,78 @@
...
+import Radio from '../../RadioGroup/radio';
+import CreditCardPaymentMethod from './creditCardPaymentMethod';
+import PaypalPaymentMethod from './paypalPaymentMethod';
+import { mergeClasses } from '../../../classify';
+
+import paymentMethodsOperations from './paymentMethods.gql';
+
+import defaultClasses from './paymentMethods.css';
+
+const PaymentMethods = props => {
```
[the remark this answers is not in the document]
revanth0212: I don't think we should be using this query. It might be a false positive. Depending on the store it might return payment methods that are not handled in the current checkout workflow. We only support Credit Card as of now. We can and should add this query in the future and render the components based on the return value. Also `braintree` in the case above can only be a braintree credit card component, not every payment method that it supports.

I will need more info from @soumya-ashok and @jimbo on this before proceeding.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paypalPaymentMethod.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { string, bool, shape } from 'prop-types';
+
+import defaultClasses from './paypalPaymentMethod.css';
+import { mergeClasses } from '../../../classify';
+
+const PaypalPaymentMethod = props => {
```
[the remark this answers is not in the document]
revanth0212: Yeah after our discussion yesterday, I have decided to delete the Paypal payment method altogether. Will remove this and other files related to paypal.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.css
```diff
@@ -0,0 +1,67 @@
+.root {
+    display: grid;
+    gap: 2rem;
+    padding: 2rem 1rem;
+}
+
+.heading_container {
```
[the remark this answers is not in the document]
revanth0212: For now ill leave them to be `_` casing. @jimbo mentioned he will make a pass and change all of them. To be consistent I have used `_` casing everywhere.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/editModal.js
```diff
@@ -0,0 +1,104 @@
...
+            <Button
+                className={classes.cancel_button}
+                onClick={handleClose}
+                priority="normal"
+            >
+                {'Cancel'}
+            </Button>
+            <Button
+                className={classes.update_button}
+                onClick={handleUpdate}
```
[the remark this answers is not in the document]
revanth0212: Yeah, similar to the prev comment I made above, Ill move this to `handlePaymentSuccess` so the update will go through only if the payment is a success.

Nice catch dude.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
revanth0212: I had that but purposefully changed it because @jimbo mentioned we have get as granular as we can with `classNames` so end users can have full control over the styling. That is the reason why I broke them up into individual spans.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
revanth0212: This is the old syntax. I made a note about changing this to the new syntax. I don't think we need the fill anymore.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.gql.js
```diff
@@ -0,0 +1,23 @@
...
+
+// We disable linting for local fields because there is no way to add them to
+// the fetched schema.
+// https://github.com/apollographql/eslint-plugin-graphql/issues/99
+/* eslint-disable graphql/template-strings */
+export const GET_SELECTED_PAYMENT_METHOD = gql`
+    query getSelectedPaymentMethod($cartId: String!) {
+        cart(cart_id: $cartId) {
+            id
+            selectedPaymentMethod @client
```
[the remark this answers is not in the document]
revanth0212: The thing with getting `selected_payment_method` from `cart` is, we have to use the `setPaymentMethodOnCart` mutation but that mutation does not accept the payment nonce's details section. It only takes the nonce, but we need it to save the details as well like `type of the card and its last 4 digits`. If I make the change Ill have to maintain 2 diff versions, half will be in remote saving `selected_payment_method, nonce` and `payment_details` in the local.

This will create more confusion than needed. I would say we update the cart with the payment information it needs but also update it in the cache.

image

The same goes for billing address. Remote does not store if the billing address is the same as the shipping address. It just returns you the address. Instead, we checking the addresses manually every time which is a time-consuming process, I would like to leave the boolean in the cache and update the appropriate address on the cart when the user submits.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
sirugh: Ok fair point - I will say that nowhere else in the app do we split presentation of firstname and lastname into separate spans. That said, I can see the confusion in the guidelines you were given :)

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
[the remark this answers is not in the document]
sirugh: I don't think pinging folks in Github is working. Can you set up some time with them to discuss this?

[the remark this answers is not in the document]
soumya-ashok: @revanth0212 It is hard to keep track of github mentions and I had missed this one. Per our discussion with @awilcoxa I believe the only payment method we are supporting first is credit card, and possibly including others as a follow up. I am not sure if there would be a change to that approach based on what payment methods braintree supports. In any case, since design effort would be required to construct a good experience around multiple payment methods, it would likely need to be a follow up anyway.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.gql.js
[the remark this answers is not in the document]
sirugh: > The thing with getting selected_payment_method from cart is, we have to use the setPaymentMethodOnCart mutation but that mutation does not accept the payment nonce's details section. It only takes the nonce, but we need it to save the details as well like type of the card and its last 4 digits

I'd like to understand this a bit more - is the problem that we need to display type of card and last 4 digits to the user but we don't get that information from graphql?

I think it is reasonable to maintain and read from two different slices of state if you don't get everything back from the server query. I don't think you should be overloading the slice names though. `selectedPaymentMethod` is confusing since there is already a returned `selected_payment_method` in the response. 

> Remote does not store if the billing address is the same as the shipping address. It just returns you the address. Instead, we checking the addresses manually every time which is a time-consuming process

I see what you are saying here. That is confusing :/ I wish they set `same_as_shipping` in the response object of `cart.billing_address`.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2320.pwa-venia.com/valeria-two-layer-tank.html : LH Best Practices Expected 1 Actual 0.93

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
jimbo: I'm fine with how @revanth0212 has done it here. 👍

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.css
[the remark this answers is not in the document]
jimbo: Yes, one day I'll make a pass and change them. 😅  

In the meantime, though, @sirugh is right: we prefer camelcase. Not much reason to use something else for new code you're submitting.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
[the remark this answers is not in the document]
jimbo: I'm fine with hardcoding one method (Braintree credit card) for now, since that's what we've built.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2320.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.67

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.gql.js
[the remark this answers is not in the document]
revanth0212: Yeah, the billing address data is being maintained at 2 diff sources. Billing address at remote and the boolean if billing is the same as shipping address in the local.

Also for the payment details. Payment nonce and selected payment method are stored at the remote and credit card details in the local.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -0,0 +1,439 @@
...
+     * `6` Payment information mutation completed
+     * `7` All mutations done
+     */
+    const [stepNumber, setStepNumber] = useState(0);
+
+    const client = useApolloClient();
+    const formState = useFormState();
+    const [{ cartId }] = useCartContext();
+
+    const { data: countriesData } = useQuery(getAllCountriesQuery);
```
tjwiebell, line 112: Why was this needed? I also can't find a `creditCard.js` component for this talon, so I'm thinking this might need re-named to match the component that uses it.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -0,0 +1,439 @@
...
+            lastName = '',
+            country = '',
+            street1 = '',
+            street2 = '',
+            city = '',
+            state = '',
+            postalCode = '',
+            phoneNumber = ''
+        } = formState.values;
+
```
tjwiebell, line 244: Wouldn't this end up suppressing any errors from GraphQL? Is that the behavior we want?

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -0,0 +1,439 @@
...
+                cart: {
+                    __typename: 'Cart',
+                    id: cartId,
+                    isBillingAddressSame
+                }
+            }
+        });
+    }, [client, cartId, getIsBillingAddressSameQuery, isBillingAddressSame]);
+
+    const setShippingAddressAsBillingAddress = useCallback(() => {
```
tjwiebell, line 199: Mentioned in the GraphQL file, seems we can greatly clean this up by syncing the two before you get to this step. Sorry for missing that in my scope 😐

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -0,0 +1,439 @@
...
+        } = rawAddressData;
+
+        return {
+            firstName,
+            lastName,
+            city,
+            postalCode,
+            phoneNumber,
+            street1: street[0],
+            street2: street[1],
```
tjwiebell, line 27: I would consider maintaining these indices; labels are whatever, but the code should be consistent.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCardPaymentMethod.gql.js
```diff
@@ -0,0 +1,177 @@
...
+            id
+        }
+    }
+`;
+
+export const GET_IS_BILLING_ADDRESS_SAME = gql`
+    query getIsBillingAddressSame($cartId: String!) {
+        cart(cart_id: $cartId) {
+            id
+            isBillingAddressSame @client
```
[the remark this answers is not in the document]
revanth0212: We can not because this needs to be persisted through refreshes.

revanth0212:
> This decision sounds fine in this initial scope, but I would like to see rendering driven from this GraphQL query at some point. There is nothing to implementing Check/Money Order, so easy to break off a Story to make this extensible, and driven off of configuration from the backend. This will enable Merchants to easily swap out Braintree, as most will probably do.

This was discussed earlier, but to recap, once we have UX and GQL support for other payment methods, once that is in scope, we can write a new talon for the `paymentMethods.js` file which will let the `PaymentMethods` component know which components to render. At this point, the decision was made to render only braintree credit card. For now, it will be hardcoded but in a future story we can extend this.

revanth0212:
> Overall I think this looks fine, but from first glance the step logic seems very complex. There are lot of areas where I feel like you're maintaining your own state, when you already have all the information you need from GraphQL data to derive that from. You're just being more explicit, which isn't a bad thing; in the future I would try to rely on GraphQL data first, because that data **is** just state.
> 
> Glossed over some areas that I know Jimmy will dig into. Saw you started on breaking up those forms, which will also make your step state easier to maintain.

You are right, I could have used the data to determine the state of the payment submission, which I did previously, but since the data submission needs to happen in sequential order, the only way I could do this without ending up in a corrupted state is to use `useEffects`. When I use `useEffects` there will be a small time gap between the effects being called, so the UI was flashing between loading screens and payment method UI for brief moments. To avoid it, @jimbo asked me to use a stepper logic. After I implemented the stepper logic, the loading state determination became very easy to understand and maintain. At first, it is confusing because this is a complex component but eventually I felt the stepper logic made it simpler.

Let me know how can we make it even simpler. I am up for it.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCardPaymentMethod.gql.js
```diff
@@ -0,0 +1,177 @@
...
+                }
+                postalCode: postcode
+                phoneNumber: telephone
+            }
+        }
+    }
+`;
+
+export const GET_SHIPPING_ADDRESS = gql`
+    query getSelectedShippingAddress($cartId: String!) {
```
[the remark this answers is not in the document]
revanth0212: That is an interesting edge case but irrespective of that, if the user checks the box to use shipping as billing address, GQL mutation mandates to still submit the whole address. So I still need to fetch the shipping address and use it to submit as billing address with a boolean letting GQL know about it (not sure why we need to submit the address when we are submitting the boolean clearly to tell it the same). Anyways, I feel once we all have all our components completed, we need to handle edge cases like if someone changes shipping address, fetching and changing shipping methods and replaying billing address mutation.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2320.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.73

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
revanth0212: My bad, `creditCardPaymentMethod.js` is the file you are looking for. I should have named it appropriately.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
revanth0212: You are right, Ill change this.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
revanth0212: I am sorry. I dint get this.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCardPaymentMethod.gql.js
```diff
@@ -0,0 +1,177 @@
...
+                name
+            }
+            id
+        }
+    }
+`;
+
+export const GET_IS_BILLING_ADDRESS_SAME = gql`
+    query getIsBillingAddressSame($cartId: String!) {
+        cart(cart_id: $cartId) {
```
sirugh, line 23: In #2322 I'm adding `@connection` directives to all queries and mutations that fetch cart or use a cart ids. You could get ahead of that by doing it in your PR now. I'm not sure if it'll operate entirely correctly if you do it before #2322 gets merged though but once it is merged you definitely should make that change.

soumya-ashok:
I have reviewed the work with @revanth0212 . The feedback was to
- Always have credit card pre-selected by default, with the form exposed
- To pre-populate the country field based on shipping address when the billing address = shipping address is unchecked
Both of those have been implemented, and are UX approved.

**Some things that came up:**

- The "edit" button on the payment and other checkout info cards needs to be a unit, without the underlined text - a button rather than icon+text link combo
- In auth checkout, a payment method when added is automatically applied to the checkout without the shopper clicking "apply" on the list page
- After adding a new address, if the shopper clicks the "edit" on the card, there is currently an issue displaying the address that was just added
- GraphQL does not allow the assignment of a "default" payment method, the first one added will always be the default. This is not optimal for the experience to be competitive for what is offered as a standard auth experience on other shopping sites.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -0,0 +1,85 @@
...
+        toggleDrawer('edit.payment');
+    }, [toggleDrawer]);
+
+    const hideEditModal = useCallback(() => {
+        closeDrawer('edit.payment');
+    }, [closeDrawer]);
+
+    const handlePaymentSuccess = useCallback(() => {
+        if (onSave) {
+            onSave();
```
sirugh, line 67: `onSave` is really just setting the initial checkout step. Use an effect to call it when data has been submitted or on your initial mount after you fetch data and see that data is all present. In this way we can set the step correctly.

See my changes here: https://github.com/magento/pwa-studio/pull/2354/files#diff-bc147cc50ccb76b2ee26bf61bbb89c56R68

revanth0212: `onSave` in case of `paymentInformation.js` sets the step to 4. This should only happen after payment success. I understand what you meant by setting the step using `useEffect` and it might work but that seems like a hack. Ideally, the component above should set the step to 3 when it is done and that should be stored in the cache. So even if we refresh the page or perform operations (edit), that should not change.

revanth0212: Having `useEffect` might confuse the component in case of authenticated checkout. If the user has a `default` shipping address assigned, and the component fetches it, it might understand that as the presence of data and call the `onSave` function which will set the step to 2 and not perform the mutation of setting shipping address on the cart.

tjwiebell: You would also do that mutation in an effect. Just solved for this in Shipping Information:

https://github.com/magento/pwa-studio/commit/7396499bb8796b5e5aea0d4e1a22bbecc2880d6a#diff-bc147cc50ccb76b2ee26bf61bbb89c56R103-R116

sirugh: We discussed over chat so logging here:

* Calling the `onSave` or "hasData" callback as it really should be named may be OK so long as the "automatic" flow makes sense for auth.
* Calling `onSave` automatically for payment info in an effect does not make sense. All prior steps are editable except for price adjustments in the "review" step so if we automatically proceed from payment to the review step on a refresh the user will never have a chance to go back if they mess up.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -0,0 +1,440 @@
...
+    const [shouldRequestPaymentNonce, setShouldRequestPaymentNonce] = useState(
+        false
+    );
+    /**
+     * `stepNumber` depicts the state of the process flow in credit card
+     * payment flow.
+     *
+     * `0` No call made yet
+     * `1` Billing address mutation intiated
+     * `2` Billing address mutation completed
```
sirugh, line 112: Do you ever set step 2 or 4 or 6?

For that matter could these be simplified into a single callback with some async/await usage? Using state and triggering render cycles seems excessive.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -0,0 +1,440 @@
...
+
+    /**
+     * Step 2 effect
+     *
+     * Billing address mutation has completed
+     */
+    useEffect(() => {
+        const billingAddressMutationCompleted =
+            billingAddressMutationCalled && !billingAddressMutationLoading;
+        if (billingAddressMutationCompleted && !billingAddressMutationErrors) {
```
sirugh, line 365: Can you simplify step 1 and step 2 into a single effect using async await? I get that there is some messiness occurring due to the fact that the dropin is a child yet we want it to invoke a callback (ie `shouldRequestPaymentNonce` but I think we can simplify some of these effects a little.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
```diff
@@ -0,0 +1,276 @@
...
+         * `5` Payment information mutation intiated
+         * `6` Payment information mutation completed
+         * `7` All mutations done
+         */
+        stepNumber,
+        initialValues,
+        shippingAddressCountry
+    } = talonProps;
+
+    const isLoading = isDropinLoading || (stepNumber >= 1 && stepNumber <= 6);
```
sirugh, line 87: Can we simplify steps into `isSubmitting`? Seems like we only use the state here and that's all it means, right?

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
sirugh: OK I did see that we have to use multiple effects due to the nature of the braintree dropin callback. Still, steps seem to be used just for loading or not loading, so I might suggest changing them.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
revanth0212: It is kind of opinionated. I feel we can leave it as is. We have given enough information on the top for a future developer to understand this step.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
revanth0212: I feel is verbose enough to understand what is happening. Having them separated into diff code blocks makes it easy to maintain/modify if needed.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
revanth0212: It was a suggestion from @jimbo to have granular steps instead of only submit steps. Hence I broke it into granular ones.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -0,0 +1,440 @@
...
+                            type
+                        }
+                    }
+                }
+            });
+        },
+        [cartId, client, getPaymentNonceQuery]
+    );
+
+    const updateCCDetailsOnCart = useCallback(
```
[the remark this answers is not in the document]
revanth0212: Again this is to separate out code blocks to better readability. If a block of code does something very specific I tend to move it into a function to have more clarity. If that does not confine with our coding standards let me, ill change it.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
```diff
@@ -0,0 +1,109 @@
...
+import summaryGQLOperations from './summary.gql';
+
+import defaultClasses from './summary.css';
+
+const Summary = props => {
+    const { classes: propClasses, onEdit } = props;
+
+    const classes = mergeClasses(defaultClasses, propClasses);
+
+    const talonProps = useSummary(summaryGQLOperations);
```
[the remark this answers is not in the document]
revanth0212: I am sure the future developers working on this will know that the prop takes an object and they can send more if needed. Not to sound dramatic but this is kind of stuff they need to know to work on our/any codebase. But if that is a standard we want to stick to, ill change.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
sirugh: Ok that's fine - at least delete the unused steps.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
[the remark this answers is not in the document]
sirugh: We don't really have function size limits defined anywhere. If it was my code I would condense it unless I used some specific functionality elsewhere, but I'm fine with your reasoning :) Readability is important.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
sirugh: Sure no problem.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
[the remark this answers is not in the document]
joturako: @revanth0212 hi, thanks for the PR, I waited for 2 weeks for this. Do you guy mind if I ask about the plan to refactor this step to support multiple payment methods? I'm adding more payment methods based on this PR and I just want to know if is there a task/issue for that that an outsider can take?

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
sirugh: All I'm saying is that the steps aren't used as triggers to the effect, they're just used to indicate whether to render a loading state or not. The effects are triggered based on a bunch of state within `useCreditCard`, which is also fine, but just makes the existence of `step` kind of moot. You may as well just call `setLoading(true)` or `setLoading(false)` and  leave everything else since the comments above the effects already explain that they are happening in some order.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }, [onSave]);
+
+    const handlePaymentError = useCallback(() => {
+        resetReviewOrderButtonClicked();
+    }, [resetReviewOrderButtonClicked]);
+
+    return {
+        doneEditing: checkoutStep > 3,
```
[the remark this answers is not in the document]
revanth0212: I remember mentioning this in the past, but the presence of data is not the same as the completion of a step in case of payment information. Even if we have all the data we need, payment information can actually result in an incomplete step due to multiple submit steps internally. That is the reason why I am using `checkoutStep` to determine `doneEditing` and that is also the reason why I do not call it `hasData` because they are diff. Like you mentioned I can use a local variable to determine this but I was asked not to create local state when you can derive it. Not sure which one to follow. Honestly, I don't see a reason why it is bad to use `checkoutStep` in this case. I feel it is present for that exact reason. If not, what is the point of it? If every other step before payment information sets the step properly we should not have any corrupted step state.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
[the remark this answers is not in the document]
sirugh: This is such a minor change, and we do it in the rest of the app. Please make the change.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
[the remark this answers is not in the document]
revanth0212: Also payment information is plugable step. Credit card has diff data and Paypal in the future will have diff data. What should the `usePaymentInformation.js` talon look for to determine if data is actually present. We can not write if-else cases to look for diff data for diff payment methods. Either have local variable called `doneEditing` and set it to true on`handlePaymentSuccess` or use `checkoutStep`.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -0,0 +1,85 @@
...
+     * Query Fetches
+     */
+
+    const { data: checkoutStepData } = useQuery(getCheckoutStepQuery, {
+        variables: cartId
+    });
+
+    const checkoutStep = checkoutStepData
+        ? checkoutStepData.cart.checkoutStep
+        : 1;
```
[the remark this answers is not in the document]
revanth0212: Not sure what you meant by getting it from checkout step.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
revanth0212: Let's stick to `street1` and `street2`. Because `street0` and `street1` don't really make sense in the context of CSS classes, components, and labels. I wanted to maintain the same name codes actors to maintain consistency.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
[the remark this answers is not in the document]
revanth0212: At this point, it is just credit card, but if you want to add a new payment method, consider checking out `paymentMethods.js` file.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
revanth0212: I had this conversation with @jimbo a while ago. He preferred step logic. Also, it is not just to render loading. If you check, we change the text inside the loading indicator based on the step giving more info to the user about what is happening. Not totally needed but as a by-product of using steps. @jimbo had plans of using the step logic in the future. That is also the reason why the steps are more granular than needed.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentInformation.gql.js
```diff
@@ -0,0 +1,24 @@
+import gql from 'graphql-tag';
+
+// We disable linting for local fields because there is no way to add them to
+// the fetched schema.
+// https://github.com/apollographql/eslint-plugin-graphql/issues/99
+/* eslint-disable graphql/template-strings */
+
+export const GET_CHECKOUT_STEP = gql`
```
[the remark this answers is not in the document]
revanth0212: We need this. `doneEditing` is fine as is. No need of refactoring there.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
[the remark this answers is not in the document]
joturako: @revanth0212 thanks, I'm half way done with it, but i'm in kind of a rush, so the code quality is not that good. I saw in the mock up you added Paypal, and i'm adding similar methods. If adding multiple payment methods is not taken by someone else, I'd like to contribute.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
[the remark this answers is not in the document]
sirugh: We talked about this - meant passing down as a prop from `CheckoutPage`. Since you no longer use it it's not important.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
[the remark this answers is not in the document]
sirugh: This callback should be unecessary now. You have an efffect that now sets the step to review once data is submitted. If the user edits and saves we're already on the review step, so no need to call this function. Infact it won't do anything because the callbac has a gate that says not to invoke the step setter if you're not on that step but in the function.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
revanth0212: Done, there should only be 4 steps in the credit card component.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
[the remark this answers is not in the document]
revanth0212: Hey, @joturko thanks for chipping in. We do not have any UX mocks right now for other payment methods. Paypal was just a place holder and we have removed it from the UI a week ago. Stay tuned and we will update the community GitHub board once the mocks are up so the community can chip in.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
sirugh: Thanks!

fooman:
The following doesn't seem quite right yet.

1. Add product to cart
2. Head to cart page
3. Click Proceed to Checkout
4. Enter a shipping address
5. Select a shipping method
6. Decide to browse the store a bit more
7. Return to checkout
8. Note Payment Information is not able to be interacted with

image

Note: After editing Shipping Method it becomes right
image

joturako:
@fooman i repored it here https://github.com/magento/pwa-studio/issues/2357#issue-606045526

fooman:
@joturako indeed looks to be the same issue. Looks like it will be hard to provide feedback from the outside as it's tricky to understand what is and isn't in scope for a given PR.

revanth0212:
> @fooman i repored it here #2357 (comment)

> @joturako indeed looks to be the same issue. Looks like it will be hard to provide feedback from the outside as it's tricky to understand what is and isn't in scope for a given PR.

Thanks, guys for reporting this. It is a known issue not due to payment information but due to a logic bug in the checkout page. It is being handled in #2354. Once that is merged, you won't see the issue.

dpatil-magento:
@revanth0212  Don't enter any info on credit card and hit `Review Order` button, You see error > now enter details and hit review again  >> Infinite loading spinner is displayed.

<a href="https://gyazo.com/455bd9672d665aa315b77388c6d817d0">Image from Gyazo</a>

soumya-ashok:
@revanth0212 Had a look - Just one small change. Could we change the message that appears in the payment card when the "review order" button is clicked from "submitting credit card information" to "checking credit card information"?

I know we're not validating the card at this point, but "submitting" might make people think they could've been charged but they still would need to review the order before placing it and it causes some confusion there.

revanth0212:
@soumya-ashok the requested UX change is done.

image

@dpatil-magento the issues have been handled.

soumya-ashok:
@revanth0212 That looks good, thank you.

Question - if someone were to "leave" checkout by navigating away from the page after they entered their payment info and then return would it "save" the payment info they had entered in the card or would they need to enter it again?

revanth0212:
> @revanth0212 That looks good, thank you.
> 
> Question - if someone were to "leave" checkout by navigating away from the page after they entered their payment info and then return would it "save" the payment info they had entered in the card or would they need to enter it again?

If they clicked the `Review Order` button, then it is saved. So if they refreshed the page or came back to the page later, they should see the payment information summary. This is because the payment nonce has been updated on the Cart in the remote so it is connected to the cart.

If they haven't clicked the `Review Order` button, then the data is lost because we can not save the information on cart till they click the review button. Neither can we save the info in the cache due to PII issues.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * This is because step numbers start from 1
+     * but array indexes start from 0. Hence setting
+     * index 0 to `null`.
+     */
+    null,
+    'Checking Credit Card Information',
+    'Checking Credit Card Information',
+    'Checking Credit Card Information',
+    'Checking Credit Card Information'
```
sirugh, line 34: Is this array still necessary anymore? Now all the steps result in the same text being rendered.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/brainTreeDropIn.js
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }, [dropinInstance, onError, onSuccess, shouldRequestPaymentNonce]);
+
+    useEffect(() => {
+        if (shouldTeardownDropin) {
+            dropinInstance
+                .teardown()
+                .then(() => {
+                    resetShouldTeardownDropin();
+                    createDropinInstance();
```
sirugh, line 132: What is the difference between this effect and [this teardown on line 71](https://github.com/magento/pwa-studio/pull/2320/files#diff-85ae277ce64dc8b57308b14d64ea72f1R71)?

revanth0212: I am not sure of the internals of the dropping, but Andy added it as part of the initial drop in work that he did for the mini cart. Here I had to move it into a diff function because of a case that can arise due to payment service failing.

Here is some background information.

The credit card checkout process is a 3 step process:
1. Save billing address information on the cart
2. Get payment nonce from braintree by submitting credit card information
3. Save the payment nonce that we got from braintree in step 2.

After step 2, the braintree dropin changes it UI from a form to a summary kind of component which does not confine with out UX.

If the save service in step 3 fails, the dropin needs to be reset to the form. This is where tear down comes into picture. When that happens, we send a signal to the braintree dropin to teardown, which in braintree terms is reset so the user can enter the new payment information and submit again.

dpatil-magento:
@revanth0212  
2 (Continuing) : When total amount is $0.00, i see Graphql call returns error instead of accepting it. Order total can be $0.00 when gift card is applied.

3. Enter Payment info and hit review order button. Now either reload page ...or come back to `Checkout` page.... user is asked to enter Payment info again, its not saved as Shipping info/Shipping method does.

revanth0212:
> @revanth0212
> 2 (Continuing) : When total amount is $0.00, i see Graphql call returns error instead of accepting it. Order total can be $0.00 when gift card is applied.
> 
> 1. Enter Payment info and hit review order button. Now either reload page ...or come back to `Checkout` page.... user is asked to enter Payment info again, its not saved as Shipping info/Shipping method does.

The first one is a graphql bug. Second is being handled in PWA-523

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
revanth0212: Done 👍 

Changed to something more meaningful.

soumya-ashok:
Thanks for clarifying @revanth0212 ! The change is UX approved.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/brainTreeDropIn.js
sirugh: The reason I asked was to try to get an understanding of both cases and see if there was some simplification we could make. In this instance it looks like the original `.teardown()` call was used as a sort of cleanup where the modal/dropin was unmounted. That doesn't seem to need the same logic as you've done in your async calls so there's nothing to do here.

There is an edge case I can forsee causing some errors be thrown. If `shouldTeardownDropin` is `true` the effect will run triggering teardown/recreation of the dropin instance. If a user closes the form during this effect's execution an error will throw because the dropin will mount without a target. That's why we added the `didClose` logic in the past.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/brainTreeDropIn.js
```diff
@@ -0,0 +1,173 @@
...
+        // this component. We do this via onmount because we have to be sure
+        // the container id has mounted per braintree's API.
+        createDropinInstance(didClose);
+
+        // If we close the form quickly after opening we may end up in a
+        // semi-mounted state so we need a local variable to make sure we
+        // behave correctly.
+        return () => {
+            didClose = true;
+        };
```
sirugh, line 105: Now that I look a little closer at this effect I realize it is not functioning as intended with `didClose`. The original effect from the dropin file defined `didClose` within the same closure as the definition for the `createDropinInstance` function. This meant that when the component unmounted and the effect's callback was invoked the `didClose` referenced within the `createDropinInstance` function was updated and handled correctly.

Here's a simplified example. The whole point is to create a `thing` if we're mounted or ignore it if we're unmounted.

```js
const [thing, setThing] = useState();

useEffect(() => {
    let unmounted = false;
    async function attach() {
        const ref = await createThingThatNeedsDom();

        // If `unmounted` is still "false" by the time the awaited function resolves we set the ref in state
        // but if it is "true" because we unmounted then we should just cleanup and never set state.
        if (unmounted) {
            ref.cleanup();
        }
        else {
            setThing(ref);
        }
    }

    // Run the async attach.
    attach();

    return () => { 
        unmounted = true;
    }
}, [
    // No dependencies because this component runs on mount.
]);
```

I've provided this example as a way to show you how it was used in the past because as implemented, `createDropinInstance` can break if we unmount before we get to `setDropinInstance(...)`. I realize you broke out the creation function because you need the logic twice but it doesn't make sense as written now. `createDropinInstance(didClose);` will always be called with `false`.

Please refactor this to correctly utilize the previous logic.

dpatil-magento:
@revanth0212 As discussed, check if you can handle this scenario same as Shipping info.  On hitting review order button on blank billing fields... we are submitting to GraphQL which return error on UI like . Its good if we can check all mandorty fields are netered before we submit to graphql. 

image

dpatil-magento:
Above reported fixes looks good. This UI validation needs to be handled - If you try to save any address other than US. Like India, Switzerland which have state drop down >> I see error - 

image

dpatil-magento:
QA Pass, need review on recent commits.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -392,13 +388,25 @@ export const useCreditCard = props => {
...
         try {
             if (shouldSubmit) {
-                setStepNumber(1);
-                if (isBillingAddressSame) {
-                    setShippingAddressAsBillingAddress();
+                /**
+                 * Validate billing address fields and only process with
+                 * submit if there are no errors.
+                 */
+                validateBillingAddressForm();
```
sirugh, line 395: Is there no way to validate these fields on blur? Check out how [other forms](https://github.com/magento/pwa-studio/blob/2a56b839666426b8edb50964778e70a7488c0e4c/packages/venia-ui/lib/components/CreateAccount/createAccount.js#L77) do it. I'm not sure why we didn't use validation on the old mini cart.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
```diff
@@ -122,6 +116,29 @@ const CreditCard = props => {
...
+                 * Informed validator functions return `undefined` is
+                 * validation is `true`
+                 */
+                return undefined;
+            } else {
+                return isRequired(value);
+            }
+        },
+        [isBillingAddressSame]
+    );
```
sirugh, line 140: Why is this form being submitted if `isBillingAddressSame` is `true`? Do we need to submit anything here? I would think if the billing and shipping address are the same (as in the user clicked the checkbox) then we would have sent that information via the graphQL cart mutation which means that this form shouldn't even render.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.gql.js
```diff
@@ -5,19 +5,6 @@ import gql from 'graphql-tag';
 // https://github.com/apollographql/eslint-plugin-graphql/issues/99
 /* eslint-disable graphql/template-strings */
 
-export const GET_ALL_COUNTRIES = gql`
```
sirugh: Why did you remove this query? We were previously validating the countries but that has now changed to just checking that the field is there, not that it's valid. If the change to validation was necessary (I'm not sure it is as we can just hide the form right?) then lets try to find a way to include country validation so we don't lose this functionality.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/brainTreeDropIn.js
```diff
@@ -39,70 +39,40 @@ const BraintreeDropin = props => {
...
-        // the container id has mounted per braintree's API.
-        createDropinInstance(didClose);
-
-        // If we close the form quickly after opening we may end up in a
-        // semi-mounted state so we need a local variable to make sure we
-        // behave correctly.
-        return () => {
-            didClose = true;
-        };
+        createDropinInstance();
```
sirugh, line 75: There is an edge case scenario that can still occur where a user navigates out of the checkout page through clicking some link like the home page. At least I'm fairly confident that's how to trigger the error. Let's check on this tomorrow and see if my theory holds.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -392,13 +388,25 @@ export const useCreditCard = props => {
     useEffect(() => {
         try {
             if (shouldSubmit) {
-                setStepNumber(1);
-                if (isBillingAddressSame) {
-                    setShippingAddressAsBillingAddress();
+                /**
+                 * Validate billing address fields and only process with
+                 * submit if there are no errors.
```
sirugh, line 393: Update this comment explaining why we are manually invoking validation (no form submit) and also why we can't just validate on blur (a user can skip the form entirely and just click review)

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
revanth0212: Discussed this on the call, commenting for notes.

We can not use the same techniques as in the previous forms, because of the way the form is submitted. In other forms, we submit using the `onSubmit` prop on the form so the validation is done by the form. In this case, we submit manually, form just serves as a context provider.  Hence we have to perform the form validation manually before submitting it.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/brainTreeDropIn.js
```diff
@@ -40,40 +40,54 @@ const BraintreeDropin = props => {
...
+        });
+
+        return dropinInstance;
+    }, [containerId]);
+
+    useEffect(() => {
+        let unmounted = false;
+
+        const renderDropin = async () => {
+            const instance = await createDropinInstance();
```
sirugh, line 71: This isn't quite right. We still get an error because it's possible for the component to attempt to attach to a container which is no longer mounted.

[Image from Gyazo](https://gyazo.com/f8bb7d5085707f5ac718b3675ab58dcd)

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/brainTreeDropIn.js
```diff
@@ -40,40 +40,54 @@ const BraintreeDropin = props => {
...
+            setDropinInstance(instance);
```
sirugh, line 72: The gate is in the wrong place. You need to put the `if(!unmounted)` check here _within_ the async function. The bug is the same as your earlier race condition - attempting to set state on an unmounted component.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/brainTreeDropIn.js
```diff
@@ -94,22 +108,40 @@ const BraintreeDropin = props => {
...
-                });
+        let unmounted = false;
+
+        const teardownAndRenderDropin = async () => {
+            dropinInstance.teardown();
+            resetShouldTeardownDropin();
+
+            const instance = await createDropinInstance();
+
+            setDropinInstance(instance);
```
sirugh, line 119: Same here.

sirugh: Something like this:
```js
    useEffect(() => {
        let unmounted = false;

        const teardownAndRenderDropin = async () => {
            try {
                dropinInstance.teardown();
                resetShouldTeardownDropin();

                const instance = await createDropinInstance();

                if (!unmounted) {
                    setDropinInstance(instance);
                    onReady(true);
                }
            } catch (err) {
                console.error(
                    `Unable to tear down and re-initialize Credit Card form (Braintree). \n${err}`
                );
                if (!unmounted) {
                    setIsError(true);
                }
            }
        };

        if (shouldTeardownDropin) {
            teardownAndRenderDropin();
        }

        return () => {
            unmounted = true;
        };
    }, [
        shouldTeardownDropin,
        dropinInstance,
        resetShouldTeardownDropin,
        createDropinInstance,
        onReady
    ]);
```

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.js
revanth0212: The billing address fields are always rendered, they are just hidden if the checkbox is checked. This was due to a review comment way up on the top 😅 . Hence when I have written a wrapper around the `isRequired` to only validate if the boolean is false.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.gql.js
revanth0212: We were using this query to get the list of countries. The list of countries are being used to validate the region code. But the region code validator that we have only works for US. It does not work for other countries. Hence I am doing what Shipping Address form does, only validate if it is present or not.

This is not the right place for that. A robust validate function needs to be written into the new Region component to handle such cases. It will be handled in a new ticket.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/brainTreeDropIn.js
revanth0212: I tried the same before committing my code and I was not able to reproduce. I tried now, and it is intermittent. Maybe I am clicking it before the initializing is done. Changing now.

joturako:
Finally 🥳 congrats!

PR #2321, opened 2020-04-07
[bug]: Fix (remove) OOTB tests from scaffold

### LINKED ISSUE #2244: [bug]: Tests don't work or pass OOTB when creating new project with `@magento/create-pwa`
**Describe the bug**
Directly after `npm init @magento/pwa` has finished, running `yarn test` will not work. (mixing `npm` and `yarn` commands because the `yarn create @magento/pwa` [currently doesn't work](https://github.com/magento/pwa-studio/issues/2238))
There are multiple reasons for this:
- `jest` isn't an included packge that is installed!
  - Therefore the config also isn't present
- `eslint-config-prettier` isn't installed
- The tests just fail

**To reproduce**
1. Run `npm init @magento/pwa`
2. Run `yarn test`
3. See error

Please see screenshots below as a step by step guide to get past the initial test blocker, to actually get the instance to a state where the tests are running

**Expected behavior**
The tests should pass, without any further packages having to be installed or config to be set up

**Screenshots**
Configuration for new project: 
image

Installation complete, attempt to run `yarn test`
image
Failed because prettiers eslint plugin isn't part of the install package

Remove prettier from test command, because it's not important right now, then re run `yarn test`
image

🤦 Install `jest` and `jest-cli` for good measure, re run `yarn test`
image
As we can see above, the tests in `_buildpack` are failing because a file is trying to be required, but it's not part of this install

Delete the `_buildpack` folder, re run `yarn test`
image
Now we've got issues with jest config not being set up to use babel

Add the following jest config to `package.json`
```
"jest": {
    "moduleNameMapper": {
      "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
      "\\.(css|less)$": "identity-obj-proxy"
    },
    "transformIgnorePatterns": [
      "/node_modules/(?!@magento).+\\.js$"
    ]
  }
```
Then re-run `yarn test`

image
It seems the tests are finally actually running, but there's failures

**Possible solutions**
There are a few things that need to be done right away:
1. `jest` needs to be added to the packages that are installed
2. Appropriate config for Jest needs to be provided either in `package.json`, or in an external `jest.config.js` file
3. `esling-config-prettier` also needs to be added to the packages that are installed (because it's part of the OOTB test command)
4. We need to figure out why `_buildpack` is in the initial install, either removing it if it isn't important, or adding a file that the tests rely on
5. Fix the failing tests, so that new projects initialised using `create-pwa` can actually rely on the core tests, and confidently add their own

 - Device [e.g. iPhone6, PC, Mac, Pixel3]: PC
 - OS [e.g. iOS8.1, Windows 10]: Centos 7
 - Browser [e.g. Chrome, Safari]: N/A
 - Browser Version [e.g. 22]:N/A
 - Magento Version: N/A
 - PWA Studio Version: 5.0.1
 - NPM version `npm -v`: 6.10.1
 - Node Version `node -v`: 10.13.0

- [x] `create-pwa`

sirugh:
## Description

The out of the box tests were failing to run. The tests that were copied were really internal tests for venia-concept components and were not intended to be scaffolded. There will be a follow-up ticket to add a proper test suite to the scaffolded app.

## Related Issue

Closes [PWA-476](https://jira.corp.magento.com/browse/PWA-476) and closes #2244.

### Verification Stakeholders
@luke-denton-aligent @zetlen 

### Verification Steps

1. Check out this PR locally.
2. In a sibling directory from where you checked out pwa-studio, run the following command to mimic a scaffold: 

```
../pwa-studio/packages/pwa-buildpack/bin/buildpack create-project . --template "venia-concept" --name "example" --author "example <example@example.com>" --backend-url "https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/" --braintree-token "sandbox_8yrzsvtm_s2bg8fs563crhqzk" --npm-client "yarn"
```
3. Once the command completes (it may throw an error, but thats ok) your directory should have the scaffolded app but there should be no tests copied from `venia-concept` and there should be no `test` script in the scaffolded `package.json`. As a bonus I also removed the `_buildpack` directory.

[Image from Gyazo](https://gyazo.com/b7989e5c6dbb9eee22094cb5acc9cfda)

## Checklist

### DIFF
--- a/packages/venia-concept/_buildpack/create.js
+++ b/packages/venia-concept/_buildpack/create.js
@@ -25,7 +25,6 @@ function createProjectFromVenia({ fs, tasks, options }) {
         'prettier:fix',
         'start',
         'start:debug',
-        'test',
         'validate-queries',
         'watch'
     ];
@@ -105,7 +104,8 @@ function createProjectFromVenia({ fs, tasks, options }) {
                 delete config.projects.venia;
                 fs.outputJsonSync(targetPath, config, { spaces: 2 });
             },
-            '{CHANGELOG*,LICENSE*,_buildpack/*}': tasks.IGNORE,
+            '{CHANGELOG*,LICENSE*,_buildpack,_buildpack/*,**/__tests__,**/__tests__/*}':
+                tasks.IGNORE,
             '**/*': tasks.COPY
         }
     };
--- a/packages/venia-concept/_buildpack/create.js
+++ b/packages/venia-concept/_buildpack/create.js
@@ -29,6 +29,19 @@ function createProjectFromVenia({ fs, tasks, options }) {
         'watch'
     ];
     const scriptsToInsert = {};
+
+    const filesToIgnore = [
+        'CHANGELOG*',
+        'LICENSE*',
+        '_buildpack',
+        '_buildpack/**',
+        // These tests are teporarily removed until we can implement a test
+        // harness for the scaffolded app. See PWA-508.
+        '**/__tests__',
+        '**/__tests__/**'
+    ];
+    const ignoresGlob = `{${filesToIgnore.join(',')}}`;
+
     return {
         after({ options }) {
             // The venia-concept directory doesn't have its own babel.config.js
@@ -104,8 +117,8 @@ function createProjectFromVenia({ fs, tasks, options }) {
                 delete config.projects.venia;
                 fs.outputJsonSync(targetPath, config, { spaces: 2 });
             },
-            '{CHANGELOG*,LICENSE*,_buildpack,_buildpack/**,**/__tests__,**/__tests__/**}':
-                tasks.IGNORE,
+            // These tasks are sequential so we must ignore before we copy.
+            [ignoresGlob]: tasks.IGNORE,
             '**/*': tasks.COPY
         }
     };
--- a/packages/venia-concept/_buildpack/create.js
+++ b/packages/venia-concept/_buildpack/create.js
@@ -104,7 +104,7 @@ function createProjectFromVenia({ fs, tasks, options }) {
                 delete config.projects.venia;
                 fs.outputJsonSync(targetPath, config, { spaces: 2 });
             },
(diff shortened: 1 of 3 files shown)


packages/venia-concept/_buildpack/create.js
```diff
@@ -105,7 +104,8 @@ function createProjectFromVenia({ fs, tasks, options }) {
                 delete config.projects.venia;
                 fs.outputJsonSync(targetPath, config, { spaces: 2 });
             },
-            '{CHANGELOG*,LICENSE*,_buildpack/*}': tasks.IGNORE,
```
sirugh, line 108: The changes remove the top-level `_buildpack` and any `__tests__` directories as well as their entire contents. Previously we were just contents and not directories which is why sub directories were being included.

packages/venia-concept/_buildpack/create.js
```diff
@@ -105,7 +104,8 @@ function createProjectFromVenia({ fs, tasks, options }) {
...
+            '{CHANGELOG*,LICENSE*,_buildpack,_buildpack/**,**/__tests__,**/__tests__/**}':
```
sirugh, line 107: This seems verbose because it is. We need to tell the matcher to ignore contents AND the top-level directory otherwise we end up with a bunch of empty `__tests__` folders :)

tjwiebell, line 107: Lets extract this out into a list that's easier to manage, and then build this pattern string.

```js
const ignorePatterns = ['CHANGELOG*', 'LICENSE*', ...];
const ignoreGlob = `{${ignorePatterns.join(',')}}`; // is glob the correct term here?
```

I would also like to see a comment that these are temporarily removed until we can implement test harness into the scaffolded app.

PR #2322, opened 2020-04-08
[bug]: Clears sensitive data on logout, login, and checkout

sirugh:
## TODO

Waiting for other checkout page PRs to get merged to verify that all necessary connection keys have been set.

## Description

We store data in several places - local storage (using persist cache AND browser persistence) and internal memory. When a user performs some action that indicates a change of user or a completion of some task (such as checkout) we should clear all data that is sensitive or no longer necessary. The purpose of this PR is to ensure that we clear/reset **all** of these possible data stores and not just one or the other. Since we use cachePersist we have to take some [manual steps](https://github.com/apollographql/apollo-cache-persist/issues/34#issuecomment-371177206).

Here's the events and the actions we should take:

Login:
* Clear stale cart (or merge cart?) - done, we delete cached cart on login.

Logout:
* Clear cart - done, we delete cached cart on logout.
* Clear user data - we use no-cache for user data outside of cart so this won't be necessary.

Checkout:
* Clear cart - TODO - waiting for #2288 

Gotchas:

Our use of `useAwaitQuery` does not handle replaying queries so we can't use `resetStore`. Our options thus are to either blow the entire thing away with `clearStore` or wipe specific cache items.

Since we're in here we may want to consider adding logic to [handle schema changes](https://github.com/apollographql/apollo-cache-persist#ive-had-a-breaking-schema-change-how-do-i-migrate-or-purge-my-cache). Would love thoughts on this.

## Related Issue

Closes [PWA-457](https://jira.corp.magento.com/browse/PWA-457).

### Verification Steps

1. See that there is now a single `Cart` entry within the apollo cache.
2. See that there are no longer cart ids used in keys within ROOT_QUERY or ROOT_MUTATION within apollo cache.
3. Regress all things cart and user.  An authed user should not see guest user info such as cart or other info from a previous checkout. A guest user should also not see data from a previously logged in user. To check this, log in and use mini cart to attach some data such as addresses/etc to the cart. Log out. After the page refreshes the old cart data should be gone from apollo memory (use apollo dev tools) and local storage. There should be a new cart id for the guest user. Add some more data to the new cart and log in. The cart you just added items/data to should be gone and the new cart should not show any old data (it'll probably show cart data from the initial cart interaction.

## Checklist


packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -63,12 +63,19 @@ export const useAuthModal = props => {
...
         await signOut({ revokeToken });
 
-        // Refresh this page.
+        // Pause automatic persistence.
+        client.persistor.pause();
+        // Delete everything in the storage provider.
+        await client.persistor.purge();
+
+        // Wipe the apollo store. Do not replay queries.
+        await client.clearStore();
```
sirugh, line 74: Instead of clearStore/refresh we should be able to just call `resetStore` which replays any queries. Unfortunately this replays await queries such as `getCartDetails` that use the old cart id and results in a thrown auth error. I'd love to figure this out, but it really only matters if we don't want to refresh the page. You can see the error for yourself if you replace lines 74-77 with the below code.

```js
await client.resetStore();
client.persistor.resume();
closeDrawer();
 ```

File: packages/peregrine/lib/talons/SignIn/useSignIn.js [hunk 2]
```diff
@@ -54,12 +57,30 @@ export const useSignIn = props => {
...
+                // Delete everything in the storage provider.
+                await client.persistor.purge();
+
+                // Delete any cached cart data.
+                Object.keys(client.cache.data.data).forEach(key => {
+                    if (key.match(/^Cart/)) {
+                        // Delete the cache entry
+                        client.cache.data.delete(key);
+
+                        // TODO: Delete the root query.
```
sirugh, line 77: If we care about deleting old cart ids/data on sign in then we will need to figure out some way to do this. Iterating over the keys for cart entries works to delete the actual data but the query itself remains in the `ROOT_QUERY` object. I'm not sure how relevant this. See below images:

Before sign in:

[Image from Gyazo](https://gyazo.com/a876736a31856c96c51d4c740ebc0378)

After sign in:

[Image from Gyazo](https://gyazo.com/3778805c9df5d31383d13f825183fea1)

Notice how after sign in the cache entry is gone but the `ROOT_QUERY` entry remains for the old cart.

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -56,10 +56,20 @@ export const useSignIn = props => {
...
-                // Then remove the old, guest cart and get the cart id from gql.
+                // Then remove the old guest cart and get the cart id from gql.
                 // TODO: This logic may be replacable with mergeCart in 2.3.4
                 await removeCart();
 
+                // Delete any cached cart data.
+                Object.keys(client.cache.data.data).forEach(key => {
+                    if (key.match(/^Cart/)) {
+                        // Delete the cache entry
+                        client.cache.data.delete(key);
```
sirugh, line 67: If this approach works we will need to apply it to the `placeOrder` handler in #2288

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2322.pwa-venia.com : LH Accessibility Expected 1 Actual 0.88

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
sirugh: We may be able to replace this with specific cache deletion.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -53,32 +38,59 @@ persistCache({
 const VeniaAdapter = props => {
     const { apiBase, apollo = {}, children, store } = props;
 
-    const apolloClient = useMemo(() => {
-        // If we already have a client instance, use that.
-        if (apollo.client) {
-            return apollo.client;
-        }
+    const [apolloClient, setApolloClient] = useState(undefined);
```
sirugh, line 41: Use `null` as initial.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -53,32 +38,59 @@ persistCache({
...
+
+    useEffect(() => {
```
sirugh, line 43: Pull out into hook for initializing client

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -63,12 +63,19 @@ export const useAuthModal = props => {
...
+
```
sirugh, line 75: Extract to function and add `persistor.resume()`

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -56,10 +56,20 @@ export const useSignIn = props => {
                 await setToken(token);
                 await getUserDetails({ fetchUserDetails });
 
-                // Then remove the old, guest cart and get the cart id from gql.
+                // Then remove the old guest cart and get the cart id from gql.
                 // TODO: This logic may be replacable with mergeCart in 2.3.4
                 await removeCart();
 
+                // Delete any cached cart data.
+                Object.keys(client.cache.data.data).forEach(key => {
```
sirugh, line 64: Extract to function.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -53,32 +38,59 @@ persistCache({
...
+        const initData = {
+            // Any initial cache state can go here!
+        };
+
+        cache.writeData({ data: initData });
+
+        const persistor = new CachePersistor({
+            cache,
+            storage: window.localStorage,
+            debug: true
```
sirugh, line 72: Remove.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -53,32 +38,59 @@ persistCache({
...
+            client.apiBase = apiBase;
+        }
+
+        const initData = {
+            // Any initial cache state can go here!
+        };
+
+        cache.writeData({ data: initData });
+
+        const persistor = new CachePersistor({
```
sirugh, line 69: Look into awaiting this init

packages/peregrine/lib/talons/SignIn/useSignIn.js
sirugh: I was able to resolve this by using `@connection` directive on EVERY query and mutation that uses cart id :o

packages/venia-ui/lib/drivers/adapter.js
sirugh: Oh, it's synchronous, so not necessary.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2322.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.63

packages/venia-ui/lib/components/MiniCart/cartOptions.gql.js
```diff
@@ -4,7 +4,8 @@ import { CartPageFragment } from '../CartPage/cartPageFragments.gql';
 
 export const REMOVE_ITEM_MUTATION = gql`
     mutation remoteItem($cartId: String!, $itemId: Int!) {
```
supernova-at, line 6:
```suggestion
    mutation removeItem($cartId: String!, $itemId: Int!) {
```

I know you didn't modify this line but as long as we're in here 😝

packages/venia-ui/lib/queries/removeItem.graphql
```diff
@@ -1,5 +1,6 @@
 mutation remoteItem($cartId: String!, $itemId: Int!) {
```
supernova-at, line 1:
```suggestion
mutation removeItem($cartId: String!, $itemId: Int!) {
```

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.gql.js
```diff
@@ -4,7 +4,7 @@ import { ItemsReviewFragment } from './itemsReviewFragments.gql';
 
 const LIST_OF_PRODUCTS_IN_CART_QUERY = gql`
     query getItemsInCart($cartId: String!) {
-        cart(cart_id: $cartId) {
+        cart(cart_id: $cartId) @connection(key: "Cart") {
```
supernova-at, line 7: Man, after seeing all these it would be great if we could enforce this somehow going forward. Guess we're just gonna have to keep a sharp eye during reviews.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -61,14 +60,9 @@ export const useAuthModal = props => {
     }, [showMyAccount]);
 
     const handleSignOut = useCallback(async () => {
-        // After logout, reset the store to set the bearer token.
-        // https://www.apollographql.com/docs/react/networking/authentication/#reset-store-on-logout
-        await resetStore();
```
supernova-at, line 66: So where do we actually clear everything out now?

packages/peregrine/lib/Apollo/clearStore.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Wipes the apollo cache in memory and persistence.
+ *
+ * @param {ApolloClient} client client with persistor objec
```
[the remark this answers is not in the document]
sirugh: he attac! he protec! most importantly he objec!






Oops :D

packages/peregrine/lib/Apollo/clearStore.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+export const clearStore = async (client, persistor) => {
```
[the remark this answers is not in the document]
sirugh: I had thought about this. In our implementation with a `apollo-cache-persist`, `client.clearStore` doesn't actually do what the apollo documentation say it should. So I think the overlap in name is fine here but I definitely will add a comment explaining this. Thanks!

packages/peregrine/lib/Apollo/deleteCacheEntry.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Deletes specific entry/entries from the apollo cache.
+ *
+ * @param {ApolloClient} client from useApolloClient, etc.
+ * @param {Function} predicate a matching function
+ */
+export const deleteCacheEntry = (client, predicate) => {
+    Object.keys(client.cache.data.data).forEach(key => {
```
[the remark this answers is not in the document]
sirugh: Hmm good point... I'll try to make this resilient with some fail safe checks :)

packages/peregrine/lib/Apollo/useCreateClient.js
```diff
@@ -0,0 +1,52 @@
...
+        if (propsClient) {
+            client = propsClient;
+        } else {
+            client = new ApolloClient({ cache, link, resolvers });
+            client.apiBase = apiBase;
+        }
+
+        const initData = {
+            // Any initial cache state can go here!
+        };
```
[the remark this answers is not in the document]
sirugh: Sure!

packages/venia-ui/lib/components/CheckoutPage/ItemsReview/itemsReview.gql.js
sirugh: There may be some way to include this with the graphql schema check tool. I'll open an investigation ticket.

https://jira.corp.magento.com/browse/PWA-527

packages/peregrine/lib/Apollo/clearStore.js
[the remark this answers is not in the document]
sirugh: doc added: 5d2dc81f183b09ecb2787a47aeeb5beb1fdfb279

packages/peregrine/lib/Apollo/deleteCacheEntry.js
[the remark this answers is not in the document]
sirugh: Added check: 5d2dc81f183b09ecb2787a47aeeb5beb1fdfb279

packages/peregrine/lib/Apollo/useCreateClient.js
```diff
@@ -0,0 +1,52 @@
...
+        cache.writeData({ data: initData });
+
+        const persistor = new CachePersistor({
+            cache,
+            storage: window.localStorage,
+            debug: process.env.NODE_ENV === 'development'
+        });
+
+        client.persistor = persistor;
+        client.onResetStore(async () => cache.writeData({ data: initData }));
```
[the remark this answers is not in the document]
sirugh: Thanks :D

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
sirugh: Thanks for pointing this out. This is magical so let me explain (and potentially apply a change).

When we call the `signOut` action we delete cart data from our **redux** browser persistence layer. If you throw a debugger on `history.go(0)` and inspect local storage you'll see that **apollo's** persistence still contains the `Cart` entry with the old id, however there will be no other reference. So when the page refreshes we still have the old cart entry. However, on page start the code that is responsible for initializing the cart checks **redux's** persistence for an existing id, finds nothing, and calls the `createCart` mutation. Our authLink automatically sends a bearer token (which is null because we're now a guest) with the mutation which sees us as a guest and responds with a brand spankin' new cart id. We then make a query for `getCartDetails` using this new cart id, and because I rewrote all the connection keys to store cart responses in a single location, we blow away the old cart data that was there from the authed user.

SO

TL;DR - A "proper" thing to do would be to call `deleteCacheEntry(...)` here before we refresh. :)

supernova-at: > brand spankin' new cart id.

😄

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2322.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.63

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -25,7 +25,9 @@ export const createCart = payload =>
 
         try {
             // errors can come from graphql and are not thrown
-            const { data, errors } = await fetchCartId();
+            const { data, errors } = await fetchCartId({
+                fetchPolicy: 'no-cache'
```
sirugh, line 29: We were storing the cart id result under ROOT_MUTATION. We don't need to/shouldn't store this.

sirugh:
@supernova-at I ended up moving the apollo client initialization back into the adapter. It is no longer async so doesn't need to occur in an effect or a custom hook.

dpatil-magento:
@sirugh As discussed `Create Account` flow does not clear Cart entry from cache also mini cart icon shows the item entry.

dpatil-magento:
ROOT_MUTATION.placeorder has cart id after placing order from /checkout. 

image

dpatil-magento:
QA Pass, good to merge after review on recent commits.

sirugh:
> ROOT_MUTATION.placeorder has cart id after placing order from /checkout.

resolved before you commented ;)

PR #2329, opened 2020-04-13
[PWA-480] Price summary does not get updated after removing an item.

tjwiebell:
## Description

Steps -
1. Add 2 products to cart and go to /cart page
2. Remove one product from cart page

Expected - Price summary should get updated for remaining products.
Actual - Price summary is blank, only labels no values.

## Related Issue

* [[PWA-480](https://jira.corp.magento.com/browse/PWA-480)] Price summary does not get updated after removing an item.

### Verification Steps

1. Go through repro steps in description, verify bug no longer present
2. Step through other price adjustment panels that affects cart updating state (totals go away while mutations are in flight), verify they still work. 

## Checklist


packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -32,6 +32,9 @@ export const useProduct = props => {
             // If a product mutation is in flight, tell the cart.
             setIsCartUpdating(updateItemLoading || removeItemLoading);
         }
+
+        // Reset updating state on unmount
+        return () => setIsCartUpdating(false);
```
sirugh, line 37: Should we use the `onCompleted` callback instead? 

https://www.apollographql.com/docs/react/data/mutations/#options

PR #2330, opened 2020-04-13
[bugfix]: Adds white stroke to checkout button css in mini cart

sirugh:
## Description

We updated icon styles but missed the checkout button. This uses the new style, specifically `--stroke` to make the icon stroke white.

## Related Issue

Closes [PWA-506](https://jira.corp.magento.com/browse/PWA-506).

### Verification Steps

1. Add an item to your cart.
2. Verify the mini cart checkout button shows the icon to the left as white outline on a black background.

[Image from Gyazo](https://gyazo.com/c45242aa5c629d9ab652b5f0406ad5b0)

## Checklist


packages/venia-ui/lib/components/Checkout/checkoutButton.css
```diff
@@ -0,0 +1,4 @@
+.icon {
+    composes: root from '../Icon/icon.css';
+    --stroke: white;
```
sirugh, line 3: An alternative would be to update the `button.css` for "high priority" (ie black background) buttons to automatically have a stroke of white for svgs:

```css

.filled svg {
    --stroke: white;
}
```

That's a bit opinionated though so I prefer our standard method of providing overwritable styles.

tjwiebell, line 3: Do we prefer using variable overrides on root instead of just targeting the icon directly? This seems easier to me, was just curious.

```css
.icon {
    stroke: white;
}
```

sirugh: I was following the pattern set forth by @jimbo in his recent icon styling PR: https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/MiniCart/section.css#L29

PR #2341, opened 2020-04-16
[Doc] css modules tutorial

jcalcaben:
## Description

This PR revises and publishes the css modules tutorial contributed by @rossmc 

## Related Issue

Closes PWA-304

## Acceptance 

any developer

### Verification Stakeholders

any developer

### Verification Steps

1. Navigate to the pwa-devdocs directory: `cd pwa-devdocs`
2. Run linting on the new topic: `yarn lint src/tutorials/pwa-studio-fundamentals/css-modules/index.md`
3. Verify no linting issues
4. Start HTML preview server: `yarn develop`
5. Navigate to the tutorials section
6. Verify left nav entry for the new topic and clicking on it takes you to the correct page

## Checklist


pwa-devdocs/src/tutorials/pwa-studio-fundamentals/css-modules/index.md
```diff
@@ -0,0 +1,145 @@
...
+import buttonOverrides from './myButtonWrapper.css';
+
+const MyButtonWrapper = props => {
+    const {
+        content,
+        root,
+        root_highPriority,
+        root_lowPriority,
+        root_normalPriority
+    } = buttonOverrides;
```
tjwiebell, line 93: Seems odd to destructure here only to create the exact same object shape below, and with long hand. A better example might be to show the contents of `myButtonWrapper.css`, and pass it directly without destructuring, which might be more effort than it's worth (maybe don't mock up fake properties, just show those class names exist in that file).

Another reason to destructure would be when names are not exactly the same, but that's not the case here.

PR #2343, opened 2020-04-19
#2331 Add new useSort hook and test

### LINKED ISSUE #2331: [feature]: Add Sort Status lable also to Category Page
PR #2294 was also a status text for sort added. We need to move this logic in a `useSort` Hook. Also, include them on the packages/venia-ui/lib/RootComponents/Category/categoryContent.js

A Good example of a hook you can find here: packages/peregrine/lib/hooks/usePagination.js

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

larsroettig:
## Description
Implement a hook to simplify usage of sort 

## Related Issue
Closes #2331.

## Acceptance 
* Implement new useSort refactoring of current logic to use the new hook

### Verification Stakeholders
- @jimbo 

### Verification Steps
- Visted a Category sorting should work
- Visited a Search Page sorting should work 

## Checklist


packages/peregrine/lib/hooks/__tests__/useSort.spec.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+test('should render without an error', () => {
+    createTestInstance(<TestComponent />);
+    const result = log.mock.calls[0][0];
+    const [{ sortText }] = result;
+    expect(sortText).toBe('Best Match');
+});
+
+test('should render with a differnt sort order', () => {
```
supernova-at, line 25:
```suggestion
test('should render with a different sort order', () => {
```

packages/peregrine/lib/hooks/__tests__/useSort.spec.js
```diff
@@ -0,0 +1,55 @@
...
+            sortText: 'Price: Low to High',
+            sortAttribute: 'price',
+            sortDirection: 'ASC'
+        });
+    });
+
+    const nextResult = log.mock.calls[1][0];
+    const [{ sortText: nextSortText }] = nextResult;
+    expect(nextSortText).toBe('Price: Low to High');
+});
```
supernova-at, line 55: 💯 great test

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
```diff
@@ -39,18 +40,24 @@ const CategoryContent = props => {
...
-                onClick={handleOpenFilters}
-                onFocus={handleLoadFilters}
-                onMouseOver={handleLoadFilters}
-                type="button"
-            >
-                {'Filter'}
-            </button>
-            <CategorySort sortControl={sortControl} />
-        </div>
+        <>
```
tjwiebell, line 43:
```suggestion
        <Fragment>
```

packages/venia-ui/lib/RootComponents/Category/category.js
```diff
@@ -98,7 +95,7 @@ const Category = props => {
                 filters: newFilters,
                 onServer: false,
                 pageSize: Number(pageSize),
-                sort: { [String(sortAttribute)]: String(sortDirection) }
+                sort: { [currentSort.sortAttribute]: currentSort.sortDirection }
```
supernova-at, line 98: I wonder if we couldn't modify the `pageControl` to just take `currentSort` and do this so that callers don't need to know about it.

packages/venia-ui/lib/RootComponents/Category/category.js
```diff
@@ -178,7 +174,7 @@ const Category = props => {
                 classes={classes}
                 data={loading ? null : data}
                 pageControl={pageControl}
-                sortControl={sortControl}
+                sortProps={sortProps}
```
supernova-at, line 177: Seems like surface area for an inconsistency / bug to creep in when both the `CategoryContent` and the `pageControl` passed to the `CategoryContent` need sort data.

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -8,20 +8,21 @@ import defaultClasses from './categorySort.css';
 
 const CategorySort = props => {
```
tjwiebell, line 9: This could probably use a rename to `ProductSort`.

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -24,17 +24,14 @@ export const useSearchPage = props => {
         }
     } = props;
 
-    const [sort, setSort] = useState({
+    const sortProps = useSort({
```
tjwiebell, line 27: Seems fine to be explicit here, but this is the same as the default in the custom hook now. I would consider removing.

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
```diff
@@ -90,19 +97,13 @@ const CategoryContent = props => {
...
     classes: shape({
         filterContainer: string,
         gallery: string,
         headerButtons: string,
         pagination: string,
         root: string,
         title: string
-    })
+    }),
+    sortProps: array
```
supernova-at, line 108: Since this component needs to know a couple levels deep of how this prop is structured, it would be helpful to expand this propType and define `sortProps` here using `arrayOf` and `shape`.

packages/venia-ui/lib/components/CategorySort/categorySort.js
```diff
@@ -88,13 +89,7 @@ CategorySort.propTypes = {
...
         })
     ),
-    sortControl: shape({
-        currentSort: shape({
-            sortAttribute: string,
-            sortDirection: sortDirections
-        }),
-        setSort: func.isRequired
-    })
+    sortProps: array
```
supernova-at, line 92: Same feedback here. Would be useful to go in more detail on this propType.

packages/venia-ui/lib/RootComponents/Category/category.js
larsroettig: @supernova-at i can try to do it but this will expand the scope much

packages/peregrine/lib/hooks/__tests__/useSort.spec.js
larsroettig: thank you

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
larsroettig: I try both `arrayOf` leads to error
`Failed prop type: Invalid prop `sortProps[1]` of type `function` supplied to `ProductSort`, expected `object`.`

Shap will not work because we need to use array is much better for renaming variables (destruction).

supernova-at:
@larsroettig there are some [CI failures](https://github.com/magento/pwa-studio/pull/2343#issuecomment-616892566) and merge conflicts to resolve as well.

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
supernova-at: Did you want to add a comment like we talked about instead?

packages/venia-ui/lib/components/CategorySort/categorySort.js
jimbo: I'm the one who added this line, since we changed the return signature of the hook. I know the following will work, but I don't know how to specifically indicate we expect an array of length `2` in which the first element is an object and the second is a function.

```es6
sortProps: arrayOf(
  oneOf(
    shape({}),
    func
  )
)
```

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2343.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.72

packages/venia-ui/lib/RootComponents/Category/category.js
supernova-at: Yep, definitely

packages/venia-ui/lib/components/CategorySort/categorySort.js
supernova-at: Ok, I've asked @larsroettig to add a comment instead.

packages/venia-ui/lib/RootComponents/Category/categoryContent.js
```diff
@@ -1,5 +1,6 @@
 import React, { Fragment, Suspense } from 'react';
 import { array, shape, string } from 'prop-types';
+import RichContent from '../../components/RichContent';
```
supernova-at, line 3: How was this working before 😝

supernova-at:
CI is reporting a failing unit test:

> packages/venia-ui/lib/RootComponents/Category/__tests__/categoryContent.spec.js

supernova-at:
I noticed that changing the sort doesn't reset the page you're on, but I verified that exists on `develop` too and is not a regression introduced here. As such, I've created a new bug in our backlog to fix.

PR #2354, opened 2020-04-22
[bug] Use gql persistence for checkout step between refreshes.

sirugh:
## Description

The checkout step is used to inform guest checkout flow which step we're in and to prevent early submission before all the data necessary is present. We use gql to store and persist the `checkoutStep` across refreshes so that a user returning to checkout is presented at the point in time they left off before.

## Related Issue

Closes [PWA-523](https://jira.corp.magento.com/browse/PWA-523).

### Verification Steps

1. Go to the cart page.
2. Check boxes in gift options and add a message. 
3. Refresh the page and check that section. Your selections/message should persist.
4. Go to checkout and refresh the page after you submit each step. The page should render the step you were just on instead of cascading from each previous step.
5. Delete `apollo-cache-persist` entry ONLY from local storage and refresh. The page should render what looks like a cascade while the effects trigger each `onSave` after fetching data. As there is no way to store nonce info for the payment info summary view, this step will appear as if you never submitted it now.
6. Log in/out. All `$Cart` should be deleted and the `Cart` entry should be recreated with a new id.

## Checklist


packages/venia-ui/lib/components/CartPage/PriceAdjustments/GiftOptions/giftOptions.js
```diff
@@ -93,7 +93,7 @@ const SET_GIFT_OPTIONS_QUERY = gql`
             include_gift_receipt: $include_gift_receipt
             include_printed_card: $include_printed_card
             gift_message: $gift_message
-        ) @client
+        ) @client @connection(key: "set_gift_options")
```
sirugh, line 96: These were missed during the PII pass in #2322

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -81,6 +81,25 @@ const VeniaAdapter = props => {
...
+        }
+        if (!initialized && persistor) {
+            initialize();
+        }
+    }, [initialized, persistor]);
+
+    if (!initialized) {
+        // TODO: Replace with app skeleton. See PWA-547.
+        return null;
+    }
```
sirugh, line 101: I broke this in #2322. We have to wait persistence otherwise we end up not restoring the cache from local storage which is counter to why we even include persistence layer.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.gql.js
```diff
@@ -64,7 +50,17 @@ export default {
...
     }
 };
+
+export const checkoutPageResolvers = {
+    Cart: {
+        checkoutStep: cart => {
+            // TODO: Replace with heuristic check against cart data. Requires
+            // fetching more than just total quantity for checkout details query
+            // "cart" arg will have server result.
+            return cart.checkoutStep || 1;
```
sirugh, line 63: This seems fine to me right now but if we wanted to get rid of the `onSave` callbacks that we require each component make we could have some heuristic that checks all cart data and sets the current step according to whatever data is missing. Good idea from @tjwiebell :)

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -56,31 +52,33 @@ export const useCheckoutPage = props => {
...
+        {
+            data: checkoutData,
+            called: checkoutCalled,
+            client,
+            loading: checkoutLoading
+        }
+    ] = useLazyQuery(getCheckoutDetailsQuery);
 
-    const [getCheckoutStep, { data: stepData, client }] = useLazyQuery(
-        getCheckoutStepQuery
```
sirugh, line 67: I replaced this query with a client side resolver for the `checkoutStep` property of `Cart` types.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -159,7 +156,7 @@ export const useCheckoutPage = props => {
         hasError: !!placeOrderError,
         isCartEmpty: !(checkoutData && checkoutData.cart.total_quantity),
         isGuestCheckout: !isSignedIn,
-        isLoading: checkoutLoading,
+        isLoading: !checkoutCalled || (checkoutCalled && checkoutLoading),
```
sirugh, line 159: There's a moment in render where we aren't loading but also haven't yet called the query. We want to wait to render anything until we've loaded the step at a minimum.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -67,6 +67,7 @@ export const useAuthModal = props => {
 
         // Also delete the Cart data from the apollo cache.
         await deleteCacheEntry(apolloClient, key => key.match(/^Cart/));
+        await deleteCacheEntry(apolloClient, key => key.match(/^\$Cart/));
```
tjwiebell, line 70: Maybe add a comment about the dual rule (Apollo caches weird when it can't compute a key).
```suggestion
        await deleteCacheEntry(apolloClient, key => key.match(/^\$?Cart/));
```

packages/peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/useCreateAccount.js
```diff
@@ -99,6 +99,9 @@ export const useCreateAccount = props => {
 
                 // Delete stale cart data from apollo
                 await deleteCacheEntry(apolloClient, key => key.match(/^Cart/));
+                await deleteCacheEntry(apolloClient, key =>
+                    key.match(/^\$Cart/)
+                );
```
tjwiebell, line 104:
```suggestion
                await deleteCacheEntry(apolloClient, key =>
                    key.match(/^\$?Cart/)
                );
```

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -56,31 +52,33 @@ export const useCheckoutPage = props => {
...
+            });
+
             client.writeQuery({
-                query: getCheckoutStepQuery,
+                query: getCheckoutDetailsQuery,
                 data: {
                     cart: {
-                        __typename: 'Cart',
-                        id: cartId,
+                        ...previousCart,
```
tjwiebell, line 75: Without the spread, would this blow away the entire Cart in the cache?

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -118,6 +124,7 @@ export const useCheckoutPage = props => {
 
         // Delete stale cart data from apollo
         await deleteCacheEntry(apolloClient, key => key.match(/^Cart/));
+        await deleteCacheEntry(apolloClient, key => key.match(/^\$Cart/));
```
tjwiebell, line 127: Merge the rules together, suggestions above.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -177,6 +175,20 @@ export const useShippingMethod = props => {
         }
     }, [chosenShippingMethodData]);
 
+    // Simple heuristic to check shipping data existed prior to this render.
+    // On first submission, when we have data, we should tell the checkout page
+    // so that we set the next step correctly.
+    const hasData =
```
tjwiebell, line 181: Nice rename; will do this in scope of my authenticated flow.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -177,6 +175,20 @@ export const useShippingMethod = props => {
...
+        !!chosenShippingMethodData &&
+        !!chosenShippingMethodData.cart.shipping_addresses[0]
+            .selected_shipping_method;
```
tjwiebell, line 184: Our components together guard against this use case, but someone just using this talon might still enter this block w/o setting an address. Let's guard against it (don't apply this from GitHub, I'm sure formatting is off).
```suggestion
        !!chosenShippingMethodData.cart.shipping_addresses.length && chosenShippingMethodData.cart.shipping_addresses[0].selected_shipping_method;
```

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
```diff
@@ -97,6 +97,9 @@ export const useCreateAccount = props => {
 
                 // Delete stale cart data from apollo
                 await deleteCacheEntry(apolloClient, key => key.match(/^Cart/));
+                await deleteCacheEntry(apolloClient, key =>
+                    key.match(/^\$Cart/)
```
tjwiebell, line 101: Merge rules

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -63,6 +63,9 @@ export const useSignIn = props => {
 
                 // Delete stale cart data from apollo
                 await deleteCacheEntry(apolloClient, key => key.match(/^Cart/));
+                await deleteCacheEntry(apolloClient, key =>
+                    key.match(/^\$Cart/)
```
tjwiebell, line 67: Merge rules

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: I believe so but I can double check

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
sirugh: This is an interesting point - we're explicit about which keys to delete, but by combining them in a single regex/function call that point is kind of obscured. Do you think its worth it to combine? I can add a comment above each call just to explain, at least.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
sirugh: Yea, I think it's a better indicator of what the condition means. I also want to rename `onSave` to something like.. `onHasData`.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
sirugh: Good call. I'll make the change.

dpatil-magento:
@sirugh If I click `Review Order` button and then do browser refresh >> Both `Review Order` and `Place Order` buttons are displayed.

image

sirugh:
>  If I click Review Order button and then do browser refresh >> Both Review Order and Place Order buttons are displayed.

Correct - the logic for when to display that button is currently held within the Payment Information component. When either #2320 is merged or this PR we will have to double check this case. Just keep that in mind after you merge one or the other.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
tjwiebell: Using regex obscures what you're doing for most anyway, I don't see any problem with being more efficient. If you were using a string method I would agree, eg. `key.startsWith`.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
sirugh: Fixing in 6130923b2

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: Without the spread the write still applies the new step but results in some warnings because the query provided expects certain fields. We can either pull off the fields from the `previousCart` and apply them or just spread the entire thing.

[Image from Gyazo](https://gyazo.com/d6ef8a5514790d4f16fc541d7688e32f)

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
sirugh: acf791049

packages/peregrine/lib/talons/SignIn/useSignIn.js
sirugh: acf791049

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: acf791049

packages/peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/useCreateAccount.js
sirugh: acf791049

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
sirugh: acf791049

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2354.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.66

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
tjwiebell: `proceedToNextStep`
`incrementStep`
`galavantForward`
`waltzToPayment`

Agreed on the rename, but can happen after this code is more stable.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -56,31 +52,33 @@ export const useCheckoutPage = props => {
...
-        // TODO: Purposely overfetch and hit the network until all components
-        // are correctly updating the cache. Will be fixed by PWA-321.
-        fetchPolicy: 'cache-and-network'
-    });
+        {
+            data: checkoutData,
+            called: checkoutCalled,
+            client,
+            loading: checkoutLoading
+        }
```
jimbo, line 60: Try to avoid nested destructuring.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -89,16 +87,24 @@ export const useCheckoutPage = props => {
     }, [toggleDrawer]);
 
     const setShippingInformationDone = useCallback(
-        () => setCheckoutStep(CHECKOUT_STEP.SHIPPING_METHOD),
-        [setCheckoutStep]
+        () =>
+            checkoutStep === CHECKOUT_STEP.SHIPPING_ADDRESS &&
+            setCheckoutStep(CHECKOUT_STEP.SHIPPING_METHOD),
```
jimbo, line 92: If you have a function that doesn't return anything, such as this one, just open a function body and use an `if` statement.

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
```diff
@@ -96,7 +96,9 @@ export const useCreateAccount = props => {
                 await removeCart();
 
                 // Delete stale cart data from apollo
-                await deleteCacheEntry(apolloClient, key => key.match(/^Cart/));
+                await deleteCacheEntry(apolloClient, key =>
+                    key.match(/^\$?Cart/)
+                );
```
jimbo, line 101: Consider extracting this function into a utility, since we're doing this three times in this PR alone. Perhaps `deleteStaleCartData`?

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: 9deb452e0

packages/peregrine/lib/talons/CreateAccount/useCreateAccount.js
sirugh: 9deb452e0

dpatil-magento:
QA Pass, good to merge after recent commits are reviewed.

PR #2360, opened 2020-04-27
Add input section to styleguide

jimbo:
## Description
Add an input section to the styleguide. The prose content may be lorem ipsum, but the values should reflect the the official Venia styles. There are four basic inputs included in this scope:

- checkbox
- radio
- select
- text

## Related Issue
PWA-436.

## Acceptance 
- @jcalcaben should affirm document structure
- @soumya-ashok or @schensley should affirm Venia typographic styles

### Verification Steps
2. Navigate to each of the four new pages from the left nav
3. Verify the normal and focus states match the new design intent

## Checklist


soumya-ashok:
@jimbo Looks good overall, a couple of minor things

- We had talked about adding the treatment for focused state. Is that coming later?
- I think it would be good to have an example of a form field with the button to it's right rather than under it
- In the XD style guide, we had also discussed and finalized a text area field for comments / reviews type use cases. Are we including it on this page eventually?

jimbo:
> @jimbo Looks good overall, a couple of minor things
> 
>     * I think it would be good to have an example of a form field with the button to it's right rather than under it
> 
>     * In the XD style guide, we had also discussed and finalized a text area field for comments / reviews type use cases. Are we including it on this page eventually?

Thanks Soumya; both of these are done now. 👍

soumya-ashok:
Just had a look, and both look good! @jimbo

PR #2365, opened 2020-04-29
The Dialog Component

supernova-at:
## Description

This PR introduces a new component into VeniaUI: The `Dialog` component.

Screen Shot 2020-05-01 at 10 44 57 AM

This component is meant to standardize the way Venia shows Dialogs all over the application and make them easier to use for developers. Now when you need a Dialog you can simply 

```jsx
        <Dialog
            onCancel={onCancel}
            onConfirm={onConfirm}
            title={'A Simple Dialog'}
        >
            <p>Hi, I'm the Dialog component.</p>
        </Dialog>
```

Instead of having to write or copy / paste a bunch of JSX and matching CSS.

The Storybook entries included in this PR illustrate the capabilities of `Dialog`:
  * It masks the page content behind it
  * It is configurable via the `isModal` prop whether clicking the mask closes the Dialog
  * It has a "heading" section that displays a title and a close "X" button
  * Its body section displays the `props.children` passed to it, and appends "Cancel" and "Confirm" buttons at the end, sometimes requiring the user to scroll to see them<sup>1</sup>
  * The "Cancel", "X", and mask buttons invoke the provided `onCancel` callback function
  * The "Confirm" button invokes the provided `onConfirm` callback function
  * The title, cancel, and confirm texts are all customizable
  * It renders and wraps its children in an `informed` `Form` component
  * It supports passing initial values to its internal `Form` component
  * The `onConfirm` callback is passed to the internal `Form` element via its `onSubmit` prop. In other words, clicking the "Confirm" button will submit the form
  * The "Cancel" and "X" buttons reset the form back to its initial state
  * It supports disabling all buttons (useful for when network requests are in flight)
  * Supports being in the right drawer for mobile mode

<sup>1</sup>: This is per UX. The "Cancel" and "Confirm" buttons should _not_ be sticky / always appear.

This PR also implements this new `Dialog` component on the Checkout page Shipping Methods section. Edit shipping methods on the checkout page to see it.

Note that there are other instances that should be converted that have been broken out into their own issues: Shipping Information and Payment Information on Checkout Page, and the Product Filters.

### Feedback Requested

Here are some specific areas to focus on for feedback:

#### The Internal Form

Dialog renders a `Form` element internally. 

##### For Buttons' Sake

This is primarily so that the Dialog buttons can be part of and control the `form`. For example, the close "X" and "Cancel" buttons reset the form, and the "Confirm" button submits the form.

The obvious alternative is to have the caller pass in a form of their own as `children` to the Dialog.
But then the Dialog buttons would be outside of the form and wouldn't have any way to access / control it.

There are some options to explore here, like having the `Dialog` export its buttons separately for consumers to inject into their forms:

```jsx
/* MyComponent.js */

import Dialog, { DialogButtons } from 'Dialog';

// render...
<Dialog>
  <Form>
    <MyFormFields />
    <DialogButtons />
  </Form>
</Dialog>
```

This pattern feels pretty non-idiomatic to me though, so I chose not to introduce it.

##### But the Mask

Unfortunately the button that is the mask is outside of the form, and therefore cannot / does not reset the form when clicked. 

You can reproduce this by changing a form field value and clicking the Mask to close the Dialog.
When you re-open the Dialog, your change persists. This is in direct contrast to how the close "X" and "Cancel" buttons work (they reset the form back to its initial values on click).

Here's the tradeoff I ran into:

```jsx
/*
 * This PR.
 * Bad because clicking the Mask doesn't reset the form.
*/

<Modal>
  <Mask />
  <aside>
    <Form>{children}</Form>
  </aside>
</Modal>
```

```jsx
/*
 * Alternate approach.
 * Bad because semantic HTML?  Fullscreen Mask inside Form feels bad.
 */
<Modal>
  <aside>
    <Form>
      <Mask />
      {children}
    </Form>
  </aside>
</Modal>
```

_Update_: We've decided to go with the Mask inside the Form. It isn't bad semantically.

##### Props to the Form

Currently, `Dialog` supports setting the `initialValues` and `onSubmit` props of the `Form` via its `initialFormValues` and `onConfirm` props respectively.

But `Form` [does support other props](https://joepuzzo.github.io/informed/?path=/story/form--props). Should `Dialog` take a `form` prop (or similar) that it just spreads out onto the `Form`?

```jsx
/* Dialog */

const { form: formProps } = props;

<Form {...formProps} onSubmit={onConfirm}>
</Form>
```

This would allow for more flexibility in working with `Dialog`'s internal `Form` at the cost of having a slightly more convoluted call site:

```jsx
const formProps = {
  initialValues: { name: 'some name' }
};

<Dialog formProps={formProps} />
```
_Update_: Support added for the pass through of `formProps` to `Dialog`'s internal `Form` in https://github.com/magento/pwa-studio/pull/2365/commits/5d8032b3d16d8aa6dc3e8bec6588d4bdcf322005.

#### Dialog Buttons

Should `Dialog` be configurable to show only one button? To not show any buttons (just the close "X")? These could be useful for more informational-type system messages. There is overlap with `Toasts` here.

_Update_: For now the Dialog will always show two buttons. In Modal mode, where the user is intentionally forced to interact with the content, the close "X" button is not rendered and the mask behind the dialog is not clickable. This makes the "Cancel" button an explicit, intentional rejection by the user and the "Confirm" button an explicit, intentional acceptance.

## Related Issue

Closes PWA-486.

### Verification Stakeholders

@soumya-ashok @jimbo 

#### Storybook

To see the Storybook for the `Dialog` component, run 

```
yarn workspace @magento/venia-ui run storybook
```

and click on `Dialog` in the left navigation.

Screen Shot 2020-04-29 at 10 01 18 AM

#### The Checkout Page

The `Dialog` component is also implemented on the Checkout page, for editing shipping methods.

1. Add an item to your cart
1. Go to the `/checkout` page
1. Set a shipping address and an initial shipping method
1. Click to "Edit" your Shipping Method
1. Verify that the Dialog appears
1. Verify that "Cancel" and "Update" buttons appear at the end of the body content of the Dialog (you likely have to scroll)
1. Verify that clicking the gray mask closes the Dialog
1. Note your selected shipping method
1. Open the Dialog back up and change your selection
1. Verify that clicking "Update" updates your shipping method (it is updated on the page when the Dialog closes)
1. Open the Dialog again and change your selection
1. Verify that clicking "Cancel" and the "X" do *not* update your selected shipping method
1. Open the Dialog again
1. Verify that your current shipping method is selected *not* the one you cancelled out of
1. Switch to mobile view
1. Verify that the Dialog appears in the right drawer and functions correctly

This new `Dialog` component as seen on the Checkout page:

Screen Shot 2020-04-29 at 10 10 44 AM

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2365.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.55

joturako:
I have a comment: with ```informed```, every time you want to manipulate the form from within the parent of the ```<Form>``` element, you need to wrap it with another element / use HOC. This introduce extra element and prop passing.  

I have looked at ```https://github.com/wsmd/react-use-form-state#examples```  and seems like you can use/manipulate the form without the extra layer/HOC.

With this you don't need to embed a form element to the dialog itself.

Regards,
L.

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
...
+    /* The container is the only grid item in the Dialog grid. */
+    grid-area: container;
+    background-color: rgb(var(--venia-background-color));
+
+    /* It sets maximum sizes so its body can handle overflow. */
+    height: 100%;
+    max-height: 90vh;
+    /* Minimum keeps a 16:9 aspect ratio. */
+    min-height: 416px;
+    width: 740px;
```
jimbo, line 83: These values feel pretty arbitrary to me, given that we often aim to stick to increments of `1rem`. What about `768px` by `432px`, which is `48rem` by `27rem`?

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
...
+
+    /* Nothing is allowed to overflow container itself. */
+    overflow: hidden;
+
+    /* Container is itself a grid container for its children. */
+    display: grid;
+    grid-template-rows: auto 1fr;
+    grid-template-areas:
+        'header'
+        'body';
```
jimbo, line 93: Likewise, I don't know that `grid-template-areas` provides value here, since repositioning the header and body of a dialog doesn't really qualify as a use case. Consider dropping this and just allowing `grid-template-rows` to set up the tracks for grid items to flow in naturally.

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
...
+
+.header {
+    grid-area: header;
+
+    border-bottom: 1px solid rgb(var(--venia-border));
+
+    /* The Header is itself a grid container for its children. */
+    display: grid;
+    grid-auto-flow: column;
+    padding: 0.875rem;
```
jimbo, line 111: As discussed during our review, I try to avoid top/bottom padding in grid layouts, since we can center things vertically without the padding and avoid requiring grid items to use negative margins if they want to use the full height.

Also, with rare exception, we tend to use increments of `0.25rem` or `4px` for gutters (gaps and padding). Since this header is already a bit tight, try increasing this to at least `1rem` on the sides.

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
...
+}
+
+.headerText {
+    /* Vertically align the title in the middle. */
+    align-self: center;
```
jimbo, line 116: Since this isn't really an exception—all items in the header are more or less vertically centered—you should probably just set this on `.header` instead, with `align-items` or `align-content`.

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
...
+ */
+
+.header {
+    grid-area: header;
+
+    border-bottom: 1px solid rgb(var(--venia-border));
+
+    /* The Header is itself a grid container for its children. */
+    display: grid;
+    grid-auto-flow: column;
```
jimbo, line 110: This is good. It's probably also a good idea to define the columns here.

```css
/* could even leave off the last `auto` */
grid-template-columns: 1fr auto;
```

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/*
+ *  Dialog Body styles.
+ */
+
+.body {
+    grid-area: body;
+
+    padding: 1em;
```
jimbo, line 133: Likewise, increments of `rem`, so probably at least `1rem` here.

packages/venia-ui/lib/components/Dialog/dialog.js
```diff
@@ -0,0 +1,134 @@
...
+ * @param {Object}  props.formProps - Props to apply to the internal form. @see https://joepuzzo.github.io/informed/?path=/story/form--props.
+ * @param {Boolean} props.isModal - Determines behavior of clicking on the mask. False cancels Dialog.
+ * @param {Boolean} props.isOpen - Whether the Dialog is currently showing.
+ * @param {Func}    props.onCancel - A function to call when the user cancels the Dialog.
+ * @param {Func}    props.onConfirm - A function to call when the user confirms the Dialog.
+ * @param {String}  props.title - The title of the Dialog.
+ */
+const Dialog = props => {
+    const {
+        areButtonsDisabled,
```
jimbo, line 33: I see what your intent is here, but this doesn't feel right. If i were writing this component, I might have called this one `shouldDisableButtons`, since props are essentially recommendations that the component uses to determine final state.

```es6
// disable the mask if a parent says we should
const maskIsDisabled = isModal || shouldDisableButtons
```

packages/venia-ui/lib/components/Dialog/dialog.js
```diff
@@ -0,0 +1,134 @@
...
+        cancelText,
+        children,
+        confirmText,
```
jimbo, line 36: It's good to be able to customize these buttons. Perhaps we should also have another prop for whether to render them at all?

```es6
const { includeButtons } = props

const buttons = includeButtons
  ? // create button elements
  : null
```

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
...
+    grid-template-columns: 1fr 1fr;
+    column-gap: 2rem;
+
+    justify-self: center;
+}
+
+.confirmButton {
+    composes: root_highPriority from '../Button/button.css';
+    overflow: hidden;
+}
```
jimbo, line 155: Do we really need to do this? I have a feeling these overrides aren't necessary and may be working around either a bug or a mistake.

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
...
+     *  The Dialog is a grid container with its container
+     *  horizontally and vertically centered.
+     */
+    display: grid;
+    grid-template-rows: 1fr auto 1fr;
+    grid-template-columns: 1fr auto 1fr;
+    grid-template-areas:
+        '. ......... .'
+        '. container .'
+        '. ......... .';
```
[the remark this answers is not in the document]
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/405e489deb627b646e9fe123e274a9e4c60f5532.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2365.pwa-venia.com/valeria-two-layer-tank.html : WPT TTFB Expected 300 Actual 557.33333333333

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,198 @@
+.mask {
+    /* The mask takes up and masks the entire screen. */
+    position: fixed;
+    top: 0;
+    left: 0;
+    height: 100vh;
+    width: 100vw;
```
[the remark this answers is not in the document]
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

packages/venia-ui/lib/components/Dialog/dialog.js
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * The Dialog component shows its children content in a dialog,
+ * encoding the look-and-feel and behavior in one place for consistency across the app.
+ *
+ * @typedef Dialog
+ * @kind functional component
+ *
+ * @param {Object}  props
+ * @param {BOolean} props.areButtonsDisabled - A toggle for whether the buttons are enabled.
```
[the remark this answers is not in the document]
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

packages/venia-ui/lib/components/Dialog/dialog.css
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

packages/venia-ui/lib/components/Dialog/dialog.css
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

packages/venia-ui/lib/components/Dialog/dialog.css
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

packages/venia-ui/lib/components/Dialog/dialog.css
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

packages/venia-ui/lib/components/Dialog/dialog.css
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

packages/venia-ui/lib/components/Dialog/dialog.css
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

packages/venia-ui/lib/components/Dialog/dialog.js
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2365/commits/d5841abb26d7a014b3bf9d50bdd5fe38c4dc0b77.

supernova-at: Fun side note: I changed this and a snapshot that I wasn't expecting to change changed and I was able to fix the bug before anyone else ever saw it - yay snapshots!

packages/venia-ui/lib/components/Dialog/dialog.js
larsroettig: Maybe it makes sense to have own buttons as a list with your own actions make it very changeable

jimbo: The issue with accepting any list of buttons is that this component is responsible for defining and attaching listeners to at least the standard buttons. If button customization is important, we may need a slightly more sophisticated API.

```es6
const { buttons } = props

const buttonElements = Array.from(buttons, button => {
  const { key, props } = button

  // `DialogButton` could render a `Button`,
  // attach a listener, and call an optional callback
  return <DialogButton key={key} {...props} />
})
```

supernova-at: I think we decided against this.

supernova-at:
> I do prefer the alternate masking approach, in which the mask is a child of the dialog.

Resolved in https://github.com/magento/pwa-studio/pull/2365/commits/c84ad0c02c9da932c8ceca4e85aea1f2b3168a90.

packages/venia-ui/lib/components/Dialog/dialog.css
supernova-at: Yeah I don't know why I couldn't get `<Button priority="high">` to work. I'll try again.

supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/2365/commits/79dc8d11b82ad31fbaa7e1b377ffab65cc8f1c7b.

supernova-at:
I just realized that the UpdateModal implementation on the Checkout page is still using the old `toggleDrawer`, which makes the app mask appear, meaning we have two masks on screen, impacting the opacity of the background mask.

packages/venia-ui/lib/components/Dialog/dialog.css
jimbo: Thanks for the fix. When we reviewed this with UX, didn't we propose shrinking the dialog to something like 640x360?

supernova-at: Yes - that change is in here: https://github.com/magento/pwa-studio/pull/2365/commits/805a399462f056f7bbb784cef9f212e9f0e38793.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2365.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.74

packages/venia-ui/lib/components/Dialog/dialog.css
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    .container {
+        max-height: 100%;
+        width: 100%;
+    }
+
+    .contents {
+        /* The contents should be no wider than the body. */
+        max-width: calc(360px - 2rem);
```
supernova-at, line 195: I would love to not have to do this but this `contents` div was extending out past its parent `body` div, causing the dialog to clip some content / scroll horizontally.

The `body` class does have `padding` on it but we apply `box-sizing: border-box` to all of our elements so I thought that was supposed to take the padding into account.

At any rate, this is the width of the `body` minus its horizontal `padding`.

supernova-at: This has been removed.

supernova-at:
PR Updated:
  * The HTML hierarchy has been simplified and the mask is inside the form, allowing it to act as a "reset" button
  


Unfortunately I still can't get the buttons in the button container to behave how I want them to when long button texts are set. The easiest place to see this is in the "Customizing the Button Texts" story in Storybook:

Screen Shot 2020-05-12 at 3 23 19 PM

I'm pretty confident now that the problem is in the interplay of our `Button` component and the grid.
You can see that the buttons extend past their grid cells, making it appear as though the grid gap isn't being honored.

If someone *cough* @jimbo *cough* wants to mess around with it further that would be awesome but since I've been looking at it for two days straight now I'm inclined to go with the simple approach of stacking the buttons on mobile and putting them side-by-side on desktop, no matter how long their content is.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2365.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.73

jimbo:
> If someone _cough_ @jimbo _cough_ wants to mess around with it further that would be awesome but since I've been looking at it for two days straight now I'm inclined to go with the simple approach of stacking the buttons on mobile and putting them side-by-side on desktop, no matter how long their content is.

I think that's a reasonable solution, and I'd defend it. I'll riff on the layout for a bit and see if I can identify what's causing the buttons to not participate correctly, but we should probably just take your solution and switch from columns to rows on mobile.

supernova-at:
I believe the build is failing due to the [lighthouse performance check](https://github.com/magento/pwa-studio/pull/2365#issuecomment-627581846).

@dpatil-magento I thought this was more of a warning than an error?

dpatil-magento:
@supernova-at In mobile Shipping methods button are vertically displayed instead of horizontal.

image

PR #2369, opened 2020-04-29
Revert "Merge master to develop"

dpatil-magento:
Reverts magento/pwa-studio#2362


dpatil-magento:
Dont need as https://github.com/magento/pwa-studio/pull/2368 will do it.

PR #2371, opened 2020-04-30
Fixes Checkout Shipping Methods Inadvertent Error Message

supernova-at:
## Description

This PR updates the way the Shipping Methods component on the Checkout page loads its data which solves two problems:
  1. It was showing an error state when it didn't need to
  2. It was inefficient

### 1. Incorrectly Showing Error State

The talon determines the proper `displayState` for the component to be in by looking at `cart.shipping_addresses[0].selected_shipping_method`.

But it was also inadvertently waiting for the results of fetching _all_ shipping methods before setting a `selectedShippingMethod`.

Therefore the `CompletedView` was being rendered before it had a `selectedShippingMethod`, which caused it to show an error message; it didn't have any data to show.

This PR updates how the talon sets `selectedShippingMethod` - it no longer waits for _all_ shipping methods. This keeps the talon internally consistent and fixes the inadvertent display of an error message by the `CompletedView`.

### 2. Inefficiency

Previously it was only querying for `carrier_code` and `method_code` of the `cart.shipping_addresses.selected_shipping_method`.
Therefore in order to display what it needed to in the `CompletedView`, it had to wait for the query that fetched _all_ shipping methods to return, and then `.find` the `selected_shipping_method` in the list of all available methods.

Instead, this PR modifies the query for `selected_shipping_method` to include all the fields it needs to display. This means the `CompletedView` can render immediately without having to wait for all the shipping methods to load first.

## Related Issue

Closes PWA-550.

#### Bug Resolution

1. Add an item to your cart
1. Go to the `/checkout` page
1. Select a shipping address
1. Select a shipping method
1. Refresh the page
1. See that the Shipping Method card no longer shows an error before displaying the selected shipping method

#### Shipping Method Full Regression

Because this PR changes how the `selectedShippingMethod` is determined, the Shipping Method component should get a full regression, including:

* Changing the shipping address (to an international country) after selecting a shipping method. Ensure you are presented with the shipping radios again to re-select a shipping method. Ensure the order summary stays up to date.
* Update the selected shipping method via the update modal by clicking "Edit". Verify that the method updates correctly.

## Checklist


packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -81,28 +81,37 @@ export const useShippingMethod = props => {
...
-                    // so the first one is the lowest cost one.
-                    selectedMethod = shippingMethods[0];
-                }
-            }
+            selectedMethod =
+                chosenShippingMethodData.cart.shipping_addresses[0]
+                    .selected_shipping_method;
+        } catch (err) {
+            // We don't have data from our specific query to fetch the selected shipping method.
+            // Intentionally swallow this error.
```
sirugh, line 97: Since we hide logs in production we can and should log the error _just in case_. Swallowing errors can make future us sad.

```js
if (process.env.NODE_ENV === 'development') {
  console.log(err);
}

File: packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js [hunk 2]
```diff
@@ -81,28 +81,37 @@ export const useShippingMethod = props => {
...
+        if (!selectedMethod) {
+            // From the lowest cost shipping method.
+            if (shippingMethods.length) {
+                // We sorted the shipping methods by price,
+                // so the first one is the lowest cost one.
+                selectedMethod = shippingMethods[0];
+                console.log(
+                    'selected method from list of methods',
+                    selectedMethod
+                ); // has all display data needed
```
sirugh, line 109:
```suggestion
```

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingRadios.js
```diff
@@ -46,11 +46,15 @@ const ShippingRadios = props => {
         return { label, value };
     });
 
+    // Match the serialization of method objects from the useShippingMethod talon.
+    const { carrier_code, method_code } = selectedShippingMethod;
+    const selectedShippingMethodSerializedValue = `${carrier_code}|${method_code}`;
```
sirugh, line 51: This logic should probably move to `useShippingRadios` and then be handled by some utility function that is shared between `useShippingMethod` and `useShippingRadios` talons.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
supernova-at: Fixed in https://github.com/magento/pwa-studio/pull/2371/commits/7318d2bd0602c7c5bbbacad19747c57247c7b0fa.

This was never really an error worth showing anyone: I was using that `try / catch` as a way not to have to check each object in `chosenShippingMethodData.cart.shipping_addresses[0].selected_shipping_method;` for truthiness but it turns out the only one we really have to check for is `chosenShippingMethodData` itself, so I just went ahead and did that.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
supernova-at: Oops, thanks!

Fixed in https://github.com/magento/pwa-studio/commit/7318d2bd0602c7c5bbbacad19747c57247c7b0fa.

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingRadios.js
supernova-at: 👍 I had the same thought and was on the fence about it. 

Fixed in https://github.com/magento/pwa-studio/commit/7318d2bd0602c7c5bbbacad19747c57247c7b0fa.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2371.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.66

dpatil-magento:
@supernova-at  After selecting Shipping method if we change to other country then > Previously selected shipping methods loads for a sec and then new available shipping method loads. 

<a href="https://gyazo.com/29cdc2a882e582fd38f7592ed38f3ca2">Image from Gyazo</a>

dpatil-magento:
Above mentioned will be handled separately.

PR #2372, opened 2020-04-30
 #2333 Add optional ratio for ressources

### LINKED ISSUE #2333: [feature]: Image component should have an optional 'ratio' prop
As a developer I'd love to re-use the `<Image/>` component for all custom components that render images. Unfortunately that's not ideal since images produced by the `<Image/>` component are always in a 4:5 ratio meaning that a banner image that needs to be 16:9 won't work properly.

I'd like to see a solution where the implementation decides what ratio the image needs to be instead of the internal image logic forcing it to use a set ratio. Something like this would be sufficient:
`<Image ratio={4/3} />`

**Additional context**
This would tackle 2 problem:
1: Not having to overwrite the `images.js` util if you want to change the ratio on catalog images.
2: Being able to create all sorts of ratios for different scenarios 

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

larsroettig:
## Description

As a developer I'd love to re-use the `<Image/>` component for all custom components that render images. Unfortunately that's not ideal since images produced by the `<Image/>` component are always in a 4:5 ratio meaning that a banner image that needs to be 16:9 won't work properly.

## Related Issue

Closes #2333.

## Acceptance 
I'd like to see a solution where the implementation decides what ratio the image needs to be instead of the internal image logic forcing it to use a set ratio. Something like this would be sufficient:
`<Image ratio={4/3} />`

### Verification Steps

1. Go to Storybook 
2. An Image using a Magento resource with resource constraints with a ratio
-> should display a 16 / Image

screencapture-localhost-9001-1588283259300

## Checklist


Jordaneisenburger:
Sweet! I remember a short convo with Tommy saying he didn't like the explicit "ratio={1/2}" API. Not sure why, but maybe @tjwiebell could elaborate it a bit more ?

Jordaneisenburger:
To really test this scenario you should change the story test case to 

```
stories.add(
    'An Image using a Magento resource with resource constraints with ratio',
    () => {
        const widths = new Map()
            .set(320, 240)
            .set(768, 480)
            .set(1024, 960)
            .set(UNCONSTRAINED_SIZE_KEY, 1280);
        return (
            <div className={classes.container}>
                <Image
                    alt="An Image using a Magento resource with resource constraints"
                    classes={{root: classes.root}}
                    resource={getSrc19('480x270')}
                    ratio={16 / 9}
                    widths={widths}
                />
            </div>
        )
    }
);
```

Now what's going to happen is that the `<Image/>` is going to use the `images.js` util that has a "hardcoded" `imageWidths` map aswell. Which it's going to map against to find the best width that is closest to the intrinsicWidth. But this particular map also needs to be adjustable per image component since we want to use this component for all our images. So right now in our test case for the 768px breakpoint it's not going to load the most optimal image what would be 480 but rather load `LARGE: 640,` which is suboptimal. 

@tjwiebell this is the issue I addressed in the meeting.

packages/venia-ui/lib/util/images.js
```diff
@@ -57,17 +57,22 @@ export const generateUrlFromContainerWidth = (
...
 
-export const generateSrcset = (imageURL, type) => {
+export const generateSrcset = (
+    imageURL,
+    type,
+    ratio
+) => {
     if (!imageURL || !type) return '';
 
+    ratio = ratio || DEFAULT_WIDTH_TO_HEIGHT_RATIO;
```
revanth0212, line 67: It works but is not prescribed to re-assign arguments.

```suggestion
    const imageRatio = ratio || DEFAULT_WIDTH_TO_HEIGHT_RATIO;
```

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -91,6 +93,7 @@ const Image = props => {
             type={type}
             width={talonResourceWidth}
             widths={widths}
+            ratio={ratio}
```
revanth0212, line 96: I see you are sending the `ratio` to the `ResourceImage` component and [receiving in the useResourceImage](https://github.com/magento/pwa-studio/pull/2372/files#diff-ed5db5a909eb7db786654de1573e23b6R25) component, but I don't see the code change to [receive the prop in `ResourceImage`](https://github.com/larsroettig/pwa-studio/blob/2333-image-ratio/packages/venia-ui/lib/components/Image/resourceImage.js#L34) component. Is there something that I am missing here?

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -29,6 +29,7 @@ import SimpleImage from './simpleImage';
  * @param {string}   props.src the source of the image, ready to use in an img element
  * @param {string}   props.type the Magento image type ("image-category" / "image-product"). Used to build the resource URL.
  * @param {number}   props.width the intrinsic width of the image & the width to request for the fallback image for browsers that don't support srcset / sizes.
+ * @param {number}      props.ratio the bild formate to generated the srcset for image default (4 / 5).
```
revanth0212, line 32:
```suggestion
 * @param {number}   props.ratio is the image width to height ratio. Defaults to 4/5.
```

packages/peregrine/lib/talons/Image/useResourceImage.js
```diff
@@ -21,16 +21,17 @@ export const useResourceImage = props => {
         resource,
         type,
         width,
-        widths
+        widths, 
+        ratio
```
revanth0212, line 25: Can you add `ratio` to the JS Docs for this talon?

revanth0212:
Weird, all tests passed in local. But danger says tests have failed.

image

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2372.pwa-venia.com : LH Performance Expected 0.85 Actual 0.57, LH Best Practices Expected 1 Actual 0.92
https://pr-2372.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.4, LH Best Practices Expected 1 Actual 0.92
https://pr-2372.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.51, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT TTFB Expected 300 Actual 3083

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -45,6 +46,7 @@ const Image = props => {
         type,
         width,
         widths,
+        ratio,
```
tjwiebell, line 49: I would expect that ratio should also work with the `SimpleImage` component as well. I don't want to get into a huge rabbit hole re-factor, but the `useImage` talon seems like the best place to have this logic; if height isn't provided but ratio is, calculate height and drill that down.

packages/venia-ui/lib/util/images.js
```diff
@@ -57,17 +57,18 @@ export const generateUrlFromContainerWidth = (
     );
 };
 
-export const generateSrcset = (imageURL, type) => {
+export const generateSrcset = (imageURL, type, ratio) => {
     if (!imageURL || !type) return '';
 
+    const imageRatio = ratio || DEFAULT_WIDTH_TO_HEIGHT_RATIO;
```
tjwiebell, line 63: This API would need to change, so maybe drilling ratio is still best to keep things simple.

packages/venia-ui/lib/util/images.js
```diff
@@ -57,17 +57,18 @@ export const generateUrlFromContainerWidth = (
     );
```
tjwiebell, line 57: Why isn't the line above also taking ratio into account?

packages/peregrine/lib/talons/Image/useResourceImage.js
```diff
@@ -21,16 +22,17 @@ export const useResourceImage = props => {
         resource,
         type,
         width,
-        widths
+        widths,
+        ratio
     } = props;
 
     const src = useMemo(() => {
         return generateUrl(resource, type)(width, height);
```
tjwiebell, line 30: Ratio should also be accounted for here. Some browsers load src and srcset side by side, and could cause the container to be the wrong size (without the right ratio).

Jordaneisenburger:
> To really test this scenario you should change the story test case to
> 
> ```
> stories.add(
>     'An Image using a Magento resource with resource constraints with ratio',
>     () => {
>         const widths = new Map()
>             .set(320, 240)
>             .set(768, 480)
>             .set(1024, 960)
>             .set(UNCONSTRAINED_SIZE_KEY, 1280);
>         return (
>             <div className={classes.container}>
>                 <Image
>                     alt="An Image using a Magento resource with resource constraints"
>                     classes={{root: classes.root}}
>                     resource={getSrc19('480x270')}
>                     ratio={16 / 9}
>                     widths={widths}
>                 />
>             </div>
>         )
>     }
> );
> ```
> 
> Now what's going to happen is that the `<Image/>` is going to use the `images.js` util that has a "hardcoded" `imageWidths` map aswell. Which it's going to map against to find the best width that is closest to the intrinsicWidth. But this particular map also needs to be adjustable per image component since we want to use this component for all our images. So right now in our test case for the 768px breakpoint it's not going to load the most optimal image which would be 480 but rather load `LARGE: 640,` which is suboptimal.
> 
> @tjwiebell this is the issue I addressed in the meeting.

Hey guys, just wanna make sure the issue I point out isn't forgotten. Maybe I misinterpret the image helper, if so you can ignore this comment.

packages/venia-ui/lib/util/images.js
revanth0212: I have made sure to add default props for the ratio in the components that end up using this util.

packages/venia-ui/lib/components/Image/image.js
revanth0212: I haven't thought of that use case. Is it possible? Neverthless, it is good to have a check. Done. Nice suggestion @tjwiebell. Fixed in 5536d6156d5d4546aa56a7914e76a764e10464b3

packages/peregrine/lib/talons/Image/useResourceImage.js
revanth0212: Because, after this commit 5536d6156d5d4546aa56a7914e76a764e10464b3, the height sent to `ResourceImage` and then to `useResourceImage` is checked against the ratio. I don't think we will need this any more.

packages/venia-ui/lib/util/images.js
revanth0212: Nice question. No reason not to have it. Fixed in b5126ab3509f6266e325fd7179c5d91362759b3e.

Note: I have not gone into the `useProductImageCarousel` talon, `ProductImageCarousel` component, and the components using this component to make the change of sending it as a prop. Felt out of scope for this PR. If someone wants to change this, they can send it as a prop and then as an argument to this function, if not the function uses the default ratio.

revanth0212:
> Hey guys, just wanna make sure the issue I point out isn't forgotten. Maybe I misinterpret the image helper, if so you can ignore this comment.

Added the story in 791c526e39f1e82de0cb519b62314f377004fa23

@Jordaneisenburger is this the expected result?

image

packages/venia-ui/lib/components/Image/image.js
```diff
@@ -108,7 +108,7 @@ const Image = props => {
                 alt={alt}
                 classes={classes}
                 displayPlaceholder={displayPlaceholder}
-                height={talonResourceHeight}
+                height={height}
```
revanth0212: Not setting the calculate talon height on the image because this will override the `srcset` on the `img` tag. `srcset` works best with width. By providing height we are making the container have a fixed height and because of that, the image stretches.

Before:
image

After:
image

dpatil-magento:
Regression looks good on image load w/o fastly image opt auto/onboard/backend. Good to merge once review approved and @Jordaneisenburger confirms on the fix.

PR #2379, opened 2020-05-04
When auth token is invalid reset cart and refetch details

### LINKED ISSUE #2378: [bug]: Expired token results in graphql errors in mini cart
**Describe the bug**
Cart query cached by apollo in local storage needs to be purged after token expired.

**To reproduce**
1. Login as customer add products to cart
2. Wait for token expiry time, https://venia.magento.com is 1 hour I believe
3. After token  is expired, open the site and check the cart, the minicart will show items in cart but when you open it, graphql error will be thrown, denying permission to access customer's cart 
4. Either you can clear local storage to continue/ login to continue.

**Expected behavior**
Cart cache should be cleared once token is expired.

**Screenshots**
Screenshot from 2020-05-02 19-06-05
Screenshot from 2020-05-02 19-06-32

**Additional context**
https://github.com/magento/pwa-studio/pull/2322.  This PR handles only on signIn, signOut and other customer triggered scenarios, we have to handle other scenarios like token expiry.

**Possible solutions**
Whenever app is launched, cartDetails query should be fetched and if it fails because of cart permission, cart section should be cleared like above PR

 - Device [e.g. iPhone6, PC, Mac, Pixel3]:PC
 - OS [e.g. iOS8.1, Windows 10]:ubuntu 18.04
 - Browser [e.g. Chrome, Safari]: Firefox
 - Browser Version [e.g. 22]: 
 - Magento Version: 2.3.4
 - PWA Studio Version: 6.0.1
 - NPM version `npm -v`:
 - Node Version : 10

- [X] `venia-concept`
- [X] `venia-ui`
- [ ] `pwa-buildpack`
- [X] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

sirugh:
## Description

The retry logic was not functioning when a guest user tried to perform an operation on an "authed" cart. This situation could arise if a user signs in and then the auth token expires naturally (1 hour ttl). The old, authed, cart id persisted and was used but since there was no more token in store the request failed.

## Related Issue

Closes #2378 .

### Verification Steps

1. Log in and add an item to your cart.
2. In local storage, change the `M2_VENIA_BROWSER_PERSISTENCE__signin_token` entry to have a `timeStored` of some low number like `1`.
3. Add another item to your cart. The operation should succeed but the cart should only have the item you just added. The cart trigger should be updated to show `1` item in the cart.

**Rerun the above steps but on step 3 _refresh_ the page and then add an item.**

## Checklist


packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -391,8 +398,12 @@ export async function writeImageToCache(item = {}) {
 function isInvalidCart(error) {
     return !!(
         error.graphQLErrors &&
-        error.graphQLErrors.find(err =>
-            err.message.includes('Could not find a cart')
+        error.graphQLErrors.find(
+            err =>
+                err.message.includes('Could not find a cart') ||
+                err.message.includes(
+                    'The current user cannot perform operations on cart'
```
sirugh, line 405: This addition is all that is necessary for the user flow to "work". The other changes in this PR are to make the UI (cart trigger item qty) update properly without having to refresh the page or add another item

packages/peregrine/lib/talons/Header/useCartTrigger.js
```diff
@@ -9,18 +13,24 @@ export const useCartTrigger = props => {
...
         queries: { getCartDetailsQuery, getItemCountQuery }
     } = props;
+
+    const apolloClient = useApolloClient();
     const [, { toggleDrawer }] = useAppContext();
     const [{ cartId }, { getCartDetails }] = useCartContext();
 
-    const [getItemCount, { data }] = useLazyQuery(getItemCountQuery);
+    const [getItemCount, { data }] = useLazyQuery(getItemCountQuery, {
+        fetchPolicy: 'cache-and-network'
```
tjwiebell, line 22: @sirugh - I couldn't get the CartTrigger to update without a refreshing unless I added this fetch policy. I suspect because the manual cache deletions don't trigger lifecycle events, it was still pulling the value from the cache.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2379.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.65

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -319,8 +322,21 @@ export const getCartDetails = payload => {
...
-                // Delete the cached ID from local storage.
-                await dispatch(removeCart());
+                if (isSignedIn) {
+                    // Since simple persistence just deletes auth token without
+                    // informing Redux, we need to perform the sign out action
+                    // to reset the store back to initial state here.
+                    await dispatch(signOut());
+                } else {
+                    // Delete the cached ID from local storage.
+                    await dispatch(removeCart());
```
supernova-at, line 332: I was a little surprised to see this as an `else`. I'm assuming both `removeCart` and `signOut` delete the auth token from local storage?

supernova-at: Ok nevermind, just read `user/asyncActions`. Might be useful to add that `removeCart` is called as part of `signOut` to the comment here.

tjwiebell: I'm going to investigate this more; this change should be resetting customer global state back to initial guest values, so if there are still components rendering like you're logged in, there's a bug somewhere.

Will add that comment and resolve the one area I've observed stale data (left drawer is still pinned like you're logged in). Did you see any other components still rendering with customer data?

supernova-at: That was the one I noticed too. No others I think.

tjwiebell: Fixed in ae8fe42

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -62,9 +71,10 @@ export const useAuthModal = props => {
     }, [showMyAccount]);
 
     const handleSignOut = useCallback(async () => {
+        setIsSigningOut(true);
```
supernova-at, line 74: Does this ever go back to `false`?

supernova-at: Or do we rely on unmount for that?

supernova-at:
After adding the second item in the verification steps I'm still appearing as signed in when I open the left drawer.

So I think I'm signed in with item #2 as the only thing in my cart.
But if I log out and log back in, #1 is the only thing in my cart.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
tjwiebell: Correct, this callback will trigger a page refresh, performing the reset for us.

tjwiebell:
@supernova-at - I didn't even run through those verification steps 🤦 this has been resolved in e41d29f.

supernova-at:
This is much better - thanks!

I still think there's some awkwardness here, but that may just be the nature of the issue. Off the top of my head I was thinking a toast or something to inform the user that they've been signed out, or better yet a prompt asking if they'd like to "stay signed in".

Since we don't have UX for any of that though, I think there's enough value here to get this in.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2379.pwa-venia.com : LH Performance Expected 0.85 Actual 0.57, LH Best Practices Expected 1 Actual 0.92
https://pr-2379.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.37, LH Best Practices Expected 1 Actual 0.92
https://pr-2379.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.51, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

PR #2384, opened 2020-05-06
fix(install): Use npm-is to require Yarn before install. Fixes PWA-505.

zetlen:
## Description

Add a preinstall script that prevents `npm install`, requiring `yarn install` instead.

The monorepo won't work right if installed with `npm`. Several users have reported that `npm install` at repo root causes somewhat mysterious errors. This is because the repo uses Yarn Workspaces, and Yarn automatically installs workspace dependencies along with root dependencies. NPM doesn't, so packages are missing deps when they run.

I, uh, found a utility called [`npm-is`](https://www.npmjs.com/package/npm-is) which detects and asserts package managers, and used it to add a preinstall script `npx npm-is yarn`. This command fails if the NPM script wasn't invoked with Yarn, thus preventing `npm install`.

## Related Issue
Closes PWA-505.

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Run `yarn run clean:all` to remove all node_modules folders.
2. Run `npm install` at package root.
3. Observe that it fails before installing dependencies, throwing an informative error message.

## Checklist


prod.dockerfile
```diff
@@ -23,6 +23,9 @@ COPY packages/venia-concept/package.json ./packages/venia-concept/package.json
 COPY package.json yarn.lock babel.config.js magento-compatibility.js ./
 COPY scripts/monorepo-introduction.js ./scripts/monorepo-introduction.js
 
+# To handle "could not get uid/gid"
+RUN npm config set unsafe-perm true
+
```
dpatil-magento, line 28: @zetlen On CI `yarn install` was failing with error https://console.aws.amazon.com/codesuite/codebuild/projects/pwa-pull-request-build/build/pwa-pull-request-build%3A4b8d13c6-fd9e-42d3-83a3-287a61eea4bd/log?region=us-east-1 .
I tried updating to latest docker images but ended up in one or other error.  You can check those errors in aws build logs for test pr https://github.com/magento/pwa-studio/pull/2385 commits.

I gave up on that and applied resolution from https://stackoverflow.com/questions/52196518/could-not-get-uid-gid-when-building-node-docker.

package.json
```diff
@@ -30,6 +30,7 @@
     "danger": "danger-ci",
     "lint": "eslint '@(packages|scripts)/**/{*.js,package.json}' --ignore-pattern node_modules --ignore-pattern storybook-dist",
     "postbuild": "rimraf \"./packages/*/dist/{,**/}__*__\"",
+    "preinstall": "node -e 'process.env.CI||process.exit(1)' || npx npm-is npm",
```
jimbo, line 33: Makes sense to me. Does `npm-is` not need to be added to the list of dependencies?

dpatil-magento: @zetlen I think you meant `node -e 'process.env.CI||process.exit(1)' || npx npm-is yarn`?

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2384.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.74

prod.dockerfile
```diff
@@ -9,6 +9,9 @@ RUN apk --no-cache --virtual add \
     g++ \
     yarn
 
+# set env variable for CI
+ENV CI=true
+
```
dpatil-magento, line 14: AWS has CI env variable but dockerfile needs to have also.

package.json
zetlen: @dpatil-magento I did, sorry. I was testing it and forgot to put it back.

@jimbo It does not! That’s why we use npx to run it, because it has to work before any dependencies are installed.

PR #2394, opened 2020-05-12
[featbug]: Implement UX around zero total checkout

sirugh:
## Description

Through a multitude of use cases you may find yourself with a $0 cart total. This PR handles those cases but also the inverse - going from a $0 cart to some total greater than zero. It also handles the case where a user has refreshed (or left the checkout page) after going to the review step.

## Related Issue

Closes [PWA-563](https://jira.corp.magento.com/browse/PWA-563).

### Verification Steps

To verify this PR you need to check **all** the ways a user may interact with a cart and a $0 total.

* Check how the Payment Information component behaves when the total changes within `/checkout` and also if the total changes outside of checkout. 
* Check how it behaves after having already had some info such as a CC that was submitted (click through to "review order" but then go back to `/cart` and remove the GC).

Basically any way that the cart could get to $0 or become > $0 needs to be checked. 

In development I found the best way to mimic this was to use the gift card component to add/remove a gift card that covered the balance. You can then add/remove that card on the /cart page or the /checkout page and navigate back and forth to see how the checkout view responds.

### Use Cases

I've tried to enumerate the use cases I had to work through for this code. I hope I covered them all.

* Zero Total Checkout must be enabled.
* "Adding a gift card" in this case is effectively used to make the total $0. You could accomplish the same with a coupon, or by having two items, a gift card with a value that covers one item and then remove another item that causes the total to become $0.
* We avoid making unnecessary gql calls. If a valid method is already selected on cart then we don't need to re-submit data when clicking "review order".

**User gets to "payment info" and total is $0**

1) On /cart, add a gift card which causes total to be $0. 
2) Go to /checkout and enter shipping info and proceed.
3) Select shipping method and proceed. You should see "free" summary view.
4) Click "review order". You should see review order view.
5) Refresh the page. You should still see "free" summary view.
6) Delete `apollo-cache-persist` and refresh the page. You should see the "free" summary view.
7) Click "review order" and "place order". It should work.

**When "free" is selected but cart total changes on checkout**

1) On /cart, add a gift card which causes total to be $0.
2) Go to /checkout and enter shipping info and proceed.
3) Select shipping method and proceed. You should see "free" summary view.
4) Remove the gift card. The braintree form should render.

**When "free" is selected but cart total changes outside of checkout**

1) On /cart, add a gift card which causes total to be $0.
2) Go to /checkout and enter shipping info and proceed.
3) Select shipping method and proceed. You should see "free" summary view.
4) Go back to /cart and remove the gift card.
5) Go /checkout. The braintree form should render.

**User submits a CC but then adds a gift card**

1) Go through /checkout with a non-zero total.
2) Enter CC data and click "review order". You should see the CC summary view.
3) Refresh the page. The CC summary view is still visible.
4) Add a gift card. The "free" summary should be visible.

### Verify existing functionality

**User submits a CC and places order**
1) Go through /checkout with a non-zero total.
2) Enter CC data and click "review order". You should see the CC summary view.
3) Refresh the page. The CC entry form is visible. Enter CC data again.
4) Click "review" and "place order". It should work.

[Image from Gyazo](https://gyazo.com/8a3182bee7dac5adc5b4a4e39260b924)

## Checklist


packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.gql.js
```diff
@@ -161,3 +161,10 @@ export default {
         setCreditCardDetailsOnCartMutation: SET_CC_DETAILS_ON_CART
     }
 };
+export const creditCardResolvers = {
+    Cart: {
+        paymentNonce: cart => {
+            return cart.paymentNonce || null;
```
sirugh, line 167: Whenever you write a `@client` field you have to define a resolver for it or Apollo "breaks" without an error. In my experience I found that this would result in a successful network call for the non-client fields but the `data` response would always be undefined after the query/mutation was completed.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/editModal.gql.js
```diff
@@ -4,7 +4,7 @@ export const GET_SELECTED_PAYMENT_METHOD = gql`
     query getSelectedPaymentMethod($cartId: String!) {
         cart(cart_id: $cartId) @connection(key: "Cart") {
             id
-            selectedPaymentMethod: selected_payment_method {
```
sirugh, line 7: I removed the alias to avoid potential misses of cache info. I wasn't sure if aliases are used when caching data but Apollo docs specify that aliases are intended for collisions such as having two `cart` queries for different `ids` within a single template:

```js
query getCarts {
  firstCart: cart(id: 1) { ... }
  secondCart: cart(id: 2) { ... }
}
```

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2394.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.74

sirugh:
There's a bug that exists on `develop` but I may be able to fix it in this PR. If the braintree token is expired and you submit an order, GQL responds with an error but we don't handle it. We should probably just set the checkout step to payment info to have the user resubmit.

[Image from Gyazo](https://gyazo.com/b89bb480f2d4adb678f20cf7becc3a76)

soumya-ashok:
@sirugh - I've tested a few different paths to make the cart total 0 vs. non-zero and the payment component behaves as desired except for that one bug I pointed out to you.

Once that's fixed, we can consider this UX approved. Thanks for the great work!

sirugh:
> Once that's fixed, we can consider this UX approved. 

I was only able to repro this once and then never again :/ I may have fixed it in 1a956a8.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -117,25 +93,36 @@ export const useCheckoutPage = props => {
...
+        } catch (err) {
+            console.error(
+                'An error occurred during when placing the order',
+                err
+            );
+            setReviewOrderButtonClicked(false);
+            setCheckoutStep(CHECKOUT_STEP.PAYMENT);
+            // TODO: Delete nonce? The nonce might be expired and why the order
+            // failed. If we delete it the payment info section will render as
+            // if it was not filled, thus prompting the user to enter new info.
```
sirugh, line 124: Should I resolve this here? The checkout page is not responsible for handling the payment info/credit card component's details such as a `nonce`. Similarly, the payment info/CC component have already submitted billing info/nonce to the cart when you click "review order" and it was successful at that time.

So we have some options. In either case we should probably throw a more descriptive toast error message such as "Unable to place order. Re-enter your payment information.". The question becomes whether we reset the payment form to the initial empty state or as the done state. Resetting the form requires that we delete the nonce here in the checkout page talon but I don't think that should be the checkout page's responsibility.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -60,64 +63,115 @@ export const usePaymentInformation = props => {
...
+            if (paymentMethod === 'free') {
+                // Only proceed to the next step if the total is still zero.
+                if (cart.prices.grand_total.value === 0) {
+                    onSave();
+                }
+            } else if (paymentMethod === 'braintree') {
+                // Only proceed to the next step if we have a nonce.
+                if (cart.paymentNonce) {
+                    onSave();
+                }
```
fooman, line 161: I know it's probably not in scope but could we add an `} else { onSave();}` in preparation for payment method support beyond free and braintree?

sirugh: I'm actually reconsidering this logic - I want to prevent as much coupling of this `PaymentInformation` parent component with it's children as possible. @revanth0212 had the right idea to use a success callback that is passed to the children and then can be called on success. The "Review Order" button invokes the chain of submission calls for whatever method is selected. We just had to handle `free` because it's not something that occurs from user input, but from a change in the backing data (whenever the cart total is $0).

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
```diff
@@ -17,72 +18,104 @@ const Summary = props => {
...
-        !isBillingAddressSame && billingAddress ? (
-            <div className={classes.address_summary_container}>
-                <div>
-                    <span className={classes.first_name}>
-                        {billingAddress.firstName}
-                    </span>
-                    <span className={classes.last_name}>
-                        {billingAddress.lastName}
-                    </span>
+    if (selectedPaymentMethod.code !== 'braintree') {
```
sirugh, line 37: I think we should rethink payment info selection and summary, but maybe not in this PR. Consider what happens if you have multiple options like braintree or check. We hardcoded the summary and edit forms to display only braintree, which means there is no way to change payment method once you've selected one.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
```diff
@@ -3,51 +3,99 @@ import { shape, string, bool, func } from 'prop-types';
...
+                onSubmit={onPaymentSuccess}
+                setDoneEditing={setDoneEditing}
+                shouldSubmit={reviewOrderButtonClicked}
+            />
+        )
+    };
+
+    const radios = availablePaymentMethods.map(({ code, title }) => {
+        // Always enable "free".
+        const isSelected = selectedPaymentMethod === code || code === 'free';
```
sirugh, line 74: This is weird. We don't render `component` until the radio is clicked, so the component logic that tells the UI "I'm done!" is only invoked if the component is clicked by a user (or if we automatically select it such as in the case of `free`.

How do I get out of that?

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/PaymentMethods/CheckmoMethod/checkmoMethod.gql.js
```diff
@@ -0,0 +1,26 @@
+import gql from 'graphql-tag';
+// Sets the provided payment method object on the cart.
+export const SET_PAYMENT_METHOD = gql`
+    mutation setPaymentMethodOnCart(
+        $cartId: String!
+        $method: PaymentMethodInput!
```
sirugh, line 6: This is duplicated gql from the other components that set payment method. We could define a callback in the parent `paymentMethods` for setting this value and then thread it to children. 

I'm leaning towards just doing that.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.gql.js
revanth0212: Good to know

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: Opened an issue for this: https://jira.corp.magento.com/browse/PWA-556

packages/venia-ui/lib/components/CartPage/PriceAdjustments/CouponCode/couponCode.js
```diff
@@ -44,6 +49,11 @@ const REMOVE_COUPON_MUTATION = gql`
             cart {
                 id
                 ...CartPageFragment
+                # If this mutation causes "free" to become available we need to know.
```
sirugh, line 52: We _must_ do this for mutations that could cause the cart total to become $0 or change from $0. I'm not sure how to enforce this but I applied it to gift card and coupon mutations. We probably need to add this to the add/remove item mutations as well.

I'd really appreciate thoughts here.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -57,67 +60,114 @@ export const usePaymentInformation = props => {
...
+    // payment method. However, "free" is automatically applied so we have an
+    // effect that handles Review Order clicks for it.
+    useEffect(() => {
+        if (reviewOrderButtonClicked && selectedPaymentMethod) {
+            if (selectedPaymentMethod === 'free') {
+                // Since `selectedPaymentMethod` is not actually set on non-free
+                // carts until submission is successful (ie braintree) we have
+                // to gate this `onSave` call. You could get here after adding
+                // an item to a cart that was previously free.
+                if (isTotalZero) {
```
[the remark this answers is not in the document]
sirugh: I refactored since this comment so I think it's no longer relevant.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -57,67 +60,114 @@ export const usePaymentInformation = props => {
...
+                    }
+                });
+            }
+        }
+        handleZeroTotal();
+    }, [cartId, isTotalZero, selectedPaymentMethod, setSelectedPaymentMethod]);
+
+    // Normally the Review Order button fires off the "submission" of the selected
+    // payment method. However, "free" is automatically applied so we have an
+    // effect that handles Review Order clicks for it.
```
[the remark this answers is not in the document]
sirugh: I ended up not using a child component but I did simplify this a bit.

revanth0212:
How about a free component? Somthing like this: c0372bb

It is a rough draft of my idea. Not the final solution.

sirugh:
> How about a free component? Somthing like this: c0372bb

I tried to do this in the most generic way possible relying purely on backend `available_payment_methods` to decide which components to display as options/etc. There were some odd use cases that presented problems so I ended up going for an easier, almost identical solution where the effect of handling free (just show summary!) is done in the payment information talon rather than delegating.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentMethods.js
```diff
@@ -0,0 +1,25 @@
+import { useQuery } from '@apollo/react-hooks';
+import { useCartContext } from '../../../context/cart';
+
+export const usePaymentMethods = props => {
+    const { queries } = props;
+    const { getPaymentMethodsQuery } = queries;
+    const [{ cartId }] = useCartContext();
```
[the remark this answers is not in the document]
sirugh: I think we actually do need `useFieldState` since the "current selected method" is actually the "current checked radio" and not what is selected in the gql data/cart. So for now I've at least moved the use of the field state into the `usePaymentMethods` talon to reduce unnecessary prop threading.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
```diff
@@ -2,52 +2,89 @@ import React from 'react';
...
+        initialSelectedMethod,
+        isLoading
+    } = talonProps;
+
+    if (isLoading) {
+        return null;
+    }
+
+    const COMPONENTS = {
+        braintree: (
```
sirugh, line 43: Per team review discussion I am rewriting this logic to be a bit more explicit.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -380,6 +394,32 @@ export const useCreditCard = props => {
...
+    const hasNonce = !!(paymentNonceData && paymentNonceData.cart.paymentNonce);
+    const selectedMethodIsBraintree =
+        paymentMethodData &&
+        paymentMethodData.cart.selected_payment_method.code === 'braintree';
+
+    // On first render we should wait to know if we have previously saved data.
+    // If so, we can just tell the parent component to display the summary.
+    useEffect(() => {
+        if (paymentMethodData && paymentNonceData) {
+            if (isFirstMount && selectedMethodIsBraintree && hasNonce) {
```
sirugh, line 406: Per team review discussion I am `isFirstMount` and `shouldRender` to be more explicit. Also I might have to update this effect as I don't think it's quite correct.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.css
```diff
@@ -13,5 +13,5 @@
 
 .radio_label {
     font-weight: 600;
-    width: 6rem;
+    min-width: 16rem;
```
sirugh, line 16: Stretched to fit longer text since it comes from the back end. This doesn't account for wrapping text though. Any ideas on how to handle this more gracefully would be welcome :)

*Before*
[Image from Gyazo](https://gyazo.com/f0c7092b7ffe2c0d2b8bcd9b874485fd)

*After*
[Image from Gyazo](https://gyazo.com/d3afd612df1d6c9a863707df61c5c9e2)

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -57,67 +60,94 @@ export const usePaymentInformation = props => {
...
+        null;
+
+    useEffect(() => {
+        // Whenever available methods change we should reset to the editing view
+        // so that a user can see the newly available methods. This could occur
+        // if a user causes their cart total to become $0.
+        setDoneEditing(false);
+    }, [availablePaymentMethods]);
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
sirugh: Added!

revanth0212:
> How about a free component? Somthing like this: [c0372bb](https://github.com/magento/pwa-studio/commit/c0372bb36f7afdb54f860186efd6535ae243aaaf)
> 
> It is a rough draft of my idea. Not the final solution. This just addresses the free handling logic, there is work to be done in Summary and EditModal (Should not be shown in method is free) to handle free.

I feel we can achieve this by considering `free` as payment method just like `braintree` and others in the future. That way we won't be adding any special conditions to support `free` in the `PaymentInformation` component. This will make the parent component generic and open for others to add new payment methods even if they do not know the inner workings of `Payment Information`. I tried to do the same in this PR #2409. Check it out and let me know what you think about it.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -57,71 +55,107 @@ export const usePaymentInformation = props => {
...
+    // directly. What results is a potential flash of selectable payment methods
+    // and then the "free" summary.
+    useEffect(() => {
+        const setFreeIfAvailable = async () => {
+            const freeIsAvailable = !!availablePaymentMethods.find(
+                ({ code }) => code === 'free'
+            );
+            if (freeIsAvailable) {
+                if (selectedPaymentMethod !== 'free') {
+                    await setPaymentMethod({
```
revanth0212, line 116: I feel we can move this logic into the Free component's talon and handle it as any other payment method would. For instance: https://github.com/magento/pwa-studio/pull/2409/files#diff-34f11d4c09662e347b956c693ec2ca43

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
```diff
@@ -2,52 +2,86 @@ import React from 'react';
...
-                        }}
-                        checked={selectedPaymentMethod === 'braintree'}
-                    />
-                    {creditCard}
-                </div>
+            <RadioGroup
+                field="selectedPaymentMethod"
+                initialValue={initialSelectedMethod}
+            >
+                {radios}
```
revanth0212, line 84: Similar to how you are rendering the radios and the related components, you can render the Free component as well. That way, instead of having PaymentInformation component and talon deal with Free as a special case, you can consider Free like any other payment method and pass the same props that you send to other payment methods. For instance: https://github.com/magento/pwa-studio/pull/2409/files#diff-c81c1a48df91716fa25273ee8d1fb34bR40

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
```diff
@@ -17,72 +18,104 @@ const Summary = props => {
...
+                            src={EditIcon}
+                            classes={{ icon: classes.edit_icon }}
+                        />
+                        <span className={classes.edit_text}>{'Edit'}</span>
+                    </button>
+                </div>
+                <div className={classes.card_details_container}>
+                    <span className={classes.payment_type}>Credit Card</span>
+                    <span className={classes.payment_details}>
+                        {paymentSummary}
```
revanth0212, line 112: Probably not related here/out of scope for this PR, but it would be great if we can separate out summary just like how you have done with payment methods to make it easy for future developers to add new summary components. For instance: https://github.com/magento/pwa-studio/pull/2409/files#diff-25736d2fa0849cf5b71064fe3e09a608

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -57,71 +55,107 @@ export const usePaymentInformation = props => {
...
+    // this effect downstream.
+    useEffect(() => {
+        setDoneEditing(false);
+    }, [availablePaymentMethods]);
+
+    // Along the lines of the above effect, since we don't have a "free method"
+    // component we check if free is available and set it as the selected method
+    // directly. What results is a potential flash of selectable payment methods
+    // and then the "free" summary.
+    useEffect(() => {
```
sirugh, line 109: As @revanth0212 showed in his commit it is possible to have free payment as a method and separate this effect out into a component. See here: https://github.com/magento/pwa-studio/compare/rugh/pwa-563-free-payment-method...rugh/pwa-563-free-payment-method-as-method?expand=1

I'm not really sure which method I prefer but we can discuss the pros/cons here.

revanth0212: Thanks for the diff @sirugh. I would prefer it to be a separate payment method to make the payment workflow generic for all current and future methods.

Pros:
1. Generic workflow for all payment methods. All payment methods get the same props and they deal with how to submit the info and let the parent know when they are done.
2. Leaner Payment Info component and talon. Easy to maintain and scale.

Cons:
1. Rework

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.css
jimbo: Looks better. It just needs a bit more `line-height` in case it does wrap.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
```diff
@@ -2,52 +2,86 @@ import React from 'react';
...
+            <div key={code} className={classes.payment_method}>
+                <Radio
+                    label={title}
+                    value={code}
+                    classes={{
+                        label: classes.radio_label
+                    }}
+                    checked={isSelected}
+                />
+                {isSelected && (
```
jimbo, line 65: Try to avoid multiline expressions in JSX like this.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -99,6 +103,7 @@ export const useCreditCard = props => {
      */
 
     const [isDropinLoading, setDropinLoading] = useState(true);
+    const [shouldRender, setShouldRender] = useState(false);
```
jimbo, line 106: Looks better.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentMethods.js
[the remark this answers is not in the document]
jimbo: Good change. 👍

jimbo:
@revanth0212 After a lot of review here, and even though I initially floated the idea of treating `Free` as just another payment method, I think I prefer to model `Free` as a special case:

- it's a naturally implicit method, whereas all of the others are naturally explicit, and
- it overrules the user's expressed input when it becomes available

I do think it's _possible_ to implement it as a more generic payment method, but I'm quite satisfied with this approach as written and ready to move forward with it. I will have a look your spike, though, since I'm curious about what the solution looks like.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
sirugh: All else aside Free _is_ a special case in that it is always implicitly the selected method on the cart if it is available. Because of that we must have special handling for it, regardless of where that logic lives. 

That said, it would be easy to add other payment methods and I think the API for adding methods will become clearer as we add less complicated ones like "Check/Money Order". It's just out of scope for this PR.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
sirugh: I'll leave my response in the other thread.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.js
sirugh: When we implement additional methods this will become a necessity but definitely out of scope for now I think.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
sirugh: To add to @revanth0212's pros and cons, the current implementation allows for a generic workflow for all "normal" payment methods right now. A new component will receive the same props as braintree and should "submit" when `shouldSubmit` or call `setDoneEditing(true)` when they don't need to render input forms. Free is a special case and needs to be applied as soon as it becomes available so it actually does not fit into the UX in the same way.

In the comparison I linked above the free payment method works as desired but there is additional work to be done within the credit card component to have the solution work for all the use cases I listed in the verification steps.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentMethods.js
sirugh: I'll change this. I wrote it as is because of the naming of the generic `Component` variable above. I wasn't really sure what to call it but I'll figure something out.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -380,6 +393,37 @@ export const useCreditCard = props => {
...
+    // go ahead and render the CC form.
+    useEffect(() => {
+        if (paymentMethodData && paymentNonceData) {
+            if (!shouldRender) {
+                // It's possible to get here with saved data but not be the
+                // selected method if a user enters CC and then changes payment
+                // method. In that case we should just re-render the CC form
+                // to have the user "re-select" the braintree payment method.
+                if (selectedMethodIsBraintree && hasNonce) {
+                    setDoneEditing(true);
```
sirugh, line 412: I just realized that this will get called by the credit card mounted within the edit modal. It's not really that important because by this time we're already displaying the summary view but in the future it could represent an issue.

I discussed this in a response to a thread here but this is one of two ways I can see that we can handle refresh/revisit logic.

I guess another alternative would be to never store nonce in cache and to always force users to re-enter CC data on refresh/revisit of checkout page.

dpatil-magento:
@sirugh Edit payment flow is broken. Infinite spinner.

d51503730a4c3200376ec26e3d24804bc073650c-attachment

sirugh:
> Edit payment flow is broken. Infinite spinner.
Caused by [this commit](https://github.com/magento/pwa-studio/pull/2394/commits/42e524dfd45acd2e11b6f12d1e230288962e4169). Will investigate.

> Add item to cart. Go to cart page > Edit product details say select different size and update. Total amount is wrong. (Show amount for 2 qty's instead on 1)

This is a graphql bug - when updating a cart item we do an `add` of the new configuration and then a `remove` of the old one. I added `available_payment_methods` to the response for this so that if we happened to make a change that caused `free` to become available it would become selected on the checkout page. However there seems to be a bug that requesting available_payment_methods in the `remove` mutation causes it to return the cart total from after the addition but not the remove operation. Here's some screenshots. Pay attention to the `prices` object.

Response on `develop`:



Response on this pr:

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
sirugh: Oops this is actually an issue - now it never renders in modal if you've got a saved CC :(

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -73,7 +75,8 @@ export const usePaymentInformation = props => {
         data: paymentInformationData,
         loading: paymentInformationLoading
     } = useQuery(getPaymentInformation, {
-        variables: { cartId }
+        variables: { cartId },
+        fetchPolicy: 'cache-and-network'
```
sirugh, line 79: This has introduced another "bug" in that there we get a render cycle where we aren't loading since it uses the cache, which causes the CC/braintree components to render causing a few issues like the nonce is deleted and there is an error in console because the CC component renders, then unmounts, then renders.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.gql.js
sirugh: To come back to this, I was sort of wrong here - You either need to have defined a local resolver or prime your cache with a default/initial state. Since we don't write a cart object to cache on initialization we must define local resolvers and have a try-catch since the `readQuery` will throw an error the first time:

https://www.apollographql.com/docs/react/data/local-state/#querying-local-state

dpatil-magento:
@sirugh  Some conflicts to resolve  with latested `develop`.  Also now once I apply gift card in cart page > Go to checkout and submit shipping information > I see infinite error taost.

image

sirugh:
Seems like there is a bug where billing address is required even when selected method is "free". This error comes back from the `placeOrder` mutation:

```
"Unable to place order: Please check the billing address information. "firstname" is required. Enter and try again. "lastname" is required. Enter and try again. "street" is required. Enter and try again. "city" is required. Enter and try again. "telephone" is required. Enter and try again. "postcode" is required. Enter and try again. "countryId" is required. Enter and try again."
```

sirugh:
Fun with "free" payment method

1. User enters a gift card causing their total to be zero.

2. We (the front end) _must_ now select `free` as the payment method, *if available* (it is configurable - see Zero Total Checkout).

3. To check if "free" is available, we query for `available_payment_methods` which returns "braintree", "checkmo", and "free".

4. Now that "free" is available, we select it via the `setPaymentMethodOnCart` mutation.
  
    * If we submit any of the `available_payment_methods` besides "free" at this point, GQL will allow it but when we make the `placeOrder` mutation it will throw an error.

5. We must then set billing address to some value (even though cart total is $0). 

    * If we try to place order without setting billing address, GQL will throw an error.

6. At this point, we query GQL for the cart `shipping_addresses`, pick the first one from the array of returned `shipping_addresses`, and send that as the billing address via `setBillingAddressOnCart` mutation.

  * If we try to set billing address with `same_as_shipping: true` to simply indicate to re-use the shipping address previously entered. GQL throws an error because this property is backwards. https://jira.corp.magento.com/browse/MC-34683

7. Now we should be able to place the order.

soumya-ashok:
Discussed the page "refresh" use case with @sirugh -
If the shopper makes a change to their cart, we would clear the payment method that was previously chosen.
This would also hold good if they did a hard refresh on the page after entering the payment information. (This seems like an edge case to me)
Either way, it is quite easy for them to fill out the payment info again so this is fine unless we get feedback otherwise.

jimbo:
It appears there are still a few GQL-related issues with this one. Thanks for enumerating them. For now, this is progress as far as the shopper being able to check out with a `Free` payment method under the correct conditions. We'll come back to this when GQL is ready to accept such an order without a billing address.

PR #2400, opened 2020-05-13
Add pjpg format only if image type of not png.

### LINKED ISSUE #2340: [bug]: transparent png's get a black background
**Describe the bug**
When you have .png file with a transparent background they end up with a black background when rendered on the frontend.

**To reproduce**
1. upload a .png (transparent background) file as a product image
2. go to product and you should see the background black instead of transparent

**Expected behavior**
When using .png's I'd want them to keep their transperancy 

**Possible solutions**
The reason this is happening is because we are forcing images to be a .jpeg format if they are not .webP. This happens at the following place:
https://github.com/magento/pwa-studio/blob/6d15db4a346b41b5e5b1b0967d3155a4a558f44b/packages/venia-ui/lib/util/makeUrl.js#L74-L76

To fix this we can either add an extra statement like this:
```
if (path.endsWith('.png')) {
    params.set('format', 'png');
} else {
    params.set('format', 'jpeg'); // Use progressive JPGs at least
}
```

But I'd prefer to see a solution where we get the actual file extension and use that as a format. Also we should check if that file extension is supported by the image optimizer/resizer, if not we can still fallback to .jpeg

 - Browser Version [e.g. 22]:
 - Magento Version:
 - PWA Studio Version:
 - NPM version `npm -v`:
 - Node Version `node -v`:

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

revanth0212:
## Description

This PR adds logic in `makeUrl.js` file in `venia-ui` to handle png image requests.

Currently, the `makeOptimizedUrl` function does not consider the file type while adding URLSearchParams. It adds `webp` and `pjpg` in that respective order. This means if the server can not handle `webp` it will consider `pjpg`. `webp` is compatible with `png`, so when the user requests a `png` it will be returned with a transparent background.

image

If the server does not support `webp` since the current code does not regard `png` and applies `pjpg` as the format, instead of a transparent background, we will get the image with a black background.

image

But with the current change, instead of `pjpg` the UI will request a `png` instead.

image

All this happens only if the backend does not support `webp` and is served by a backend that does not handle `pjpg` properly. For instance, we were not able to reproduce this issue till we switched off fastly, because fastly handles `pjpg` well and does not return a black background. When we switched off fastly, the backend image optimization was handled by hastily which was retuning the black background.

I am not sure if the approach taken by fastly or hastily is correct but one thing is for sure, we have to send format as `png` for a `png` image request irrespective of the backend.

## Troubleshooting Tips:
1. Check if the backend service supports `webp`. If it does, everything should be fine.
2. If it does not, check the file format requested in the `URLSearchParams`.

## Related Issue
Closes #2340 

### Verification Stakeholders
@zetlen 
@jimbo 
@Jordaneisenburger 

### Verification Steps

To test use this, use this backend URL:
`MAGENTO_BACKEND_URL=https://staging-5em2ouy-mfwmkrjfqvbjk.us-4.magentosite.cloud/`

1. Go to https://magento-venia-concept-7bnnn.local.pwadev:8914/simple13.html
2. With fastly enabled/disabled everything should work fine. You should see a transparent background when you open the image in the network tab.

## Checklist
None


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2400.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.74

packages/venia-ui/lib/util/makeUrl.js
```diff
@@ -76,7 +83,14 @@ const makeOptimizedUrl = (path, { type, ...opts } = {}) => {
     // Append image optimization parameters
     const params = new URLSearchParams(baseURL.search);
     params.set('auto', 'webp'); // Use the webp format if available
-    params.set('format', 'pjpg'); // Use progressive JPGs at least
+
+    const imageFileType = getFileType(baseURL);
+    if (imageFileType === 'png') {
+        params.set('format', 'png'); // use png if webp is not available
```
revanth0212, line 89: Piggybacking on the [comment](https://github.com/magento/pwa-studio/issues/2340#issuecomment-625337800) @Jordaneisenburger made on the issue, what should we do to make this extensible?

Should we do something like this?

```es6
if (isValidImageFileType(imageFileType)) {
    params.set('format', imageFileType);
} else {
    params.set('format', 'pjpg');
}
```

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2400.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.66

PR #2402, opened 2020-05-18
Preselect lowest cost shipping method for authed users without one

supernova-at:
## Description

This PR pre-selects the lowest cost shipping method for authenticated (signed in) users who don't yet have a saved / preferred shipping method. 

I also added a new `INITIALIZING` value to the `displayState` enumeration to remove a visual hiccup that occurred during auto-selection of a shipping method. The `ShippingMethod` component should be `INITIALIZING` before fetching its data and while the auto-selection is in progress.

Adding this `INITIALIZING` state allowed us to greatly simplify downstream logic as well 🎉 .

## Related Issue

Closes PWA-573.

### Verification Stakeholders

@soumya-ashok 

### Verification Steps

1. Sign in
1. Add an item to your cart
1. Set a shipping address (via mini-cart if #2380 isn't merged yet)
1. Go to the `/checkout` page
1. Verify that you're never actually presented with the shipping radios to select; the lowest cost shipping option (FREE) is selected for you (after some loading) and you are taken directly to the payment section
1. Fill out the payment section
1. Verify that you can successfully place your order

573_update

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2402.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.61

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.js
```diff
@@ -99,6 +69,53 @@ const ShippingMethod = props => {
...
+                    shippingMethods={shippingMethods}
+                />
+                {maybeContinueButton}
+            </Form>
+        );
+
+        const bodyContents =
+            displayState === displayStates.INITIALIZING
+                ? initializingContents
+                : editingContents;
```
supernova-at, line 111: The only logic that actually changed is the addition of what to do when `displayState` is `INITIALIZING`. The other changes here are just for readability.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -159,15 +167,73 @@ export const useShippingMethod = props => {
...
+            : isBackgroundAutoSelecting
+            ? displayStates.INITIALIZING
             : displayStates.EDITING;
         setDisplayState(nextDisplayState);
-    }, [data]);
+    }, [data, isBackgroundAutoSelecting]);
+
+    // If an authenticated user does not have a preferred shipping method,
+    // auto-select the least expensive one for them.
+    useEffect(() => {
```
supernova-at, line 178: This is the meat of the PR, everything else is minor changes to support a seamless view for the user.

supernova-at:
I should note that the Update Modal also has a new behavior:

The call to get all available shipping methods is `cache-and-network`, and previously this modal was showing a loading spinner until the network responded. Now, the modal shows the data that is in the cache (if present) immediately - it no longer waits for the network to respond. This results in a better, faster experience for the user.

In the case where the network data actually differs from what is in the cache, the modal still disables the "submit" button until the network responds.

In other words, you can't actually make a different selection until we know we have the most up-to-date data, but we no longer show a loading spinner until that happens.  This does introduce the potential for the selection list to change while the user is looking at it.

soumya-ashok:
@supernova-at I followed the steps, and the shipping method behaves as you described. That part is UX approved.

The use-case I wanted to test but can't until Tommy's work is merged is that very first time someone comes through auth checkout and doesn't have a saved shipping address. Once they enter the shipping address in the new checkout flow, I assume the shipping method behavior will stay consistent?

supernova-at:
> Once they enter the shipping address in the new checkout flow, I assume the shipping method behavior will stay consistent?

Yes, it should behave as it does here 👍

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -61,10 +64,15 @@ export const useShippingMethod = props => {
...
-    const [selectedShippingMethod, setSelectedShippingMethod] = useState(null);
+    const [displayState, setDisplayState] = useState(
+        displayStates.INITIALIZING
+    );
+    const [isBackgroundAutoSelecting, setIsBackgroundAutoSelecting] = useState(
+        false
+    );
     const [isUpdateMode, setIsUpdateMode] = useState(false);
+    const [selectedShippingMethod, setSelectedShippingMethod] = useState(null);
+    const [shippingMethods, setShippingMethods] = useState([]);
```
tjwiebell, line 75: I know this pattern existing prior to this PR, but wanted to discuss why I think it might be an anti-pattern, and something we should avoid. Apollo's hooks are effects themselves, so you should never need to wrap that returned state in an effect, or duplicate it into its own state. I'm a bit unique in that I didn't write much React before hooks, but I think the pattern you're going for here from previous paradigms is derivedState.

Two use cases for derived state:
1. Data needs manipulated - `useMemo` if complex, but option 2 also works, wrap logic in a function:
```js
const memoizedShippingMethods = useMemo(() => {
    let mutatedShippingMethods = [];
    if (data) {
        const primaryShippingAddress = data.cart.shipping_addresses[0];

        // Shape the list of available shipping methods.
        // Sort them by price and add a serialized property to each.
        const rawShippingMethods =
            primaryShippingAddress.available_shipping_methods;
        const shippingMethodsByPrice = [...rawShippingMethods].sort(
            byPrice
        );
        mutatedShippingMethods = shippingMethodsByPrice.map(
            addSerializedProperty
        );
    }

    return mutatedShippingMethods;
}, [data]);
```

2. Data just needs extracted - no special treatment
```js
const derviedSelectedShippingMethod =
    data &&
    addSerializedProperty(
        data.cart.shipping_addresses[0].selected_shipping_method
    ); // you may want a || null default value here
```

I would like to earmark this to be discussed during tomorrow's team time, as I think we should make sure we're all aligned on this. I'm not 100% I'm right, so worth discussing for my own benefit as well.

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.js
```diff
@@ -99,6 +69,47 @@ const ShippingMethod = props => {
...
+                        {'Continue to Payment Information'}
+                    </Button>
+                </div>
+            </Form>
+        );
+
+        const bodyContents =
+            displayState === displayStates.INITIALIZING
+                ? initializingContents
+                : editingContents;
```
tjwiebell, line 105: Recent feedback from Jimmy, very minor. We've already paid the cost of creating the `editingContent` element and this means we may never render it. Best to only create element we intend to render.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -159,9 +167,67 @@ export const useShippingMethod = props => {
...
+            });
+
+            // And re-fetch our data so that our other effects fire (if necessary).
+            fetchShippingMethodInfo({
+                variables: { cartId },
+                onCompleted: () => setIsBackgroundAutoSelecting(false)
+            });
+        };
+
+        const primaryAddress = data.cart.shipping_addresses[0];
```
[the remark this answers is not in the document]
fooman: Should be a possibility once virtual products are supported.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -159,9 +167,67 @@ export const useShippingMethod = props => {
...
+        // Functions passed to useEffect should be synchronous.
+        // Set this helper function up as async so we can wait on the mutation
+        // before re-querying.
+        const autoSelectShippingMethod = async shippingMethod => {
+            const { carrier_code, method_code } = shippingMethod;
+
+            setIsBackgroundAutoSelecting(true);
+
+            // Perform the operation on the backend.
+            await setShippingMethodCall({
```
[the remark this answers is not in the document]
sirugh: Tommy is right - if the data can be returned by a mutation and assuming the response data is correct (as in no bug in GQL) we should opt to query for that data in the response rather than firing a second request/query. If you're able to get all you need in the response you can get rid of the extra loading state variable too. Seems like a win-win!

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
sirugh: We use the derived pattern often throughout the codebase. [Here](https://github.com/magento/pwa-studio/blob/52fcaafab570dd34f3a42620defd96bf7feb2385/packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/useShippingInformation.js#L31-L59) it is in the shipping info talon and [here](https://github.com/magento/pwa-studio/blob/52fcaafab570dd34f3a42620defd96bf7feb2385/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js#L42-L54) it is in the breadcrumb talon.

Efficiency aside I think it's more readable which is why I'd prefer it :)

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -69,13 +72,47 @@ export const useShippingMethod = props => {
...
+        : DEFAULT_SELECTED_SHIPPING_METHOD;
+
+    // Determine the component's display state.
+    const nextDisplayState = derivedSelectedShippingMethod
+        ? displayStates.DONE
+        : isLoadingShippingMethods || (isSettingShippingMethod && isBackgroundAutoSelecting)
+        ? displayStates.INITIALIZING
+        : displayStates.EDITING;
+    
+    if (nextDisplayState !== displayState) {
```
jimbo, line 107: You don't need this check; if React detects the new state matches the previous [it _should_ skip the update](https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update).

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/__tests__/updateModal.spec.js
```diff
@@ -40,7 +40,7 @@ test('it renders correctly', () => {
     expect(instance.toJSON()).toMatchSnapshot();
 });
 
-test('it renders correctly during loading', () => {
+test('it disables the submit butting while loading', () => {
```
jimbo, line 43: Butting whom?

supernova-at:
PR updated with derived state!

This greatly simplifies the whole talon thank you so much @tjwiebell !

I am also trying to sneak in the disabling of the input radios when the page is updating 😉 . I think Shipping Methods is in a much better place now 💪

dpatil-magento:
Could not test edit address flow for Auth users but other than that it look good.

PR #2403, opened 2020-05-18
[PWA-562] Checkout (auth): Stored Address Follow Up

tjwiebell:
## Description

In PWA-245 we ran into an issue where address information could not be stored due to a graphql bug: https://jira.corp.magento.com/browse/MC-33854

This story is meant to address delivering a workaround so that we're not hard coding the region to 'AL' 

### Approach ###
In this solution, I made the Region component support any key from the response data to be the values of the dropdown. This allows:

1. Backwards Compatibility - all components previously using `code` will continue to function with no change.
2. Working Customer Address Mutations - a few GraphQL bugs were requiring that region.id is passed in mutations, even though mutations would succeed without any error indicating this. 

It appears that GraphQL is backtracking on the deprecation of region.id, so once that work has been completed, I anticipate that we will need to re-factor Region to always use `id` as the value. Since Cart addresses do not return `region.id`, it would be impossible to preemptively replace `code` with `id` across the code base.

### Alternative ###
Another solution would be to leave the existing usage of `code` in place, but before sending the mutation, performing a lookup that maps that code to the corresponding `id`. While this data should already be in the cache for zero network cost to lookup, I thought it would be bad to make that the default behavior. Happy to explore if the current solution doesn't seem like an ideal state until we can solely rely on `id` in the future.

## Related Issue

* [[PWA-562](https://jira.corp.magento.com/browse/PWA-562)] 

### Verification Steps

1. Login, add product, navigate to `/checkout`
2. Edit an existing address, change the State selection
3. Verify that state selection actually persists and is no longer hard-coded to Alabama

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.gql.js
+++ b/packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.gql.js
@@ -23,10 +23,8 @@ export const UPDATE_CUSTOMER_ADDRESS_MUTATION = gql`
         updateCustomerAddress(id: $addressId, input: $address)
             @connection(key: "updateCustomerAddress") {
             id
-            ...CustomerAddressFragment
         }
     }
-    ${CustomerAddressFragment}
 `;
 
 export const GET_CUSTOMER_QUERY = gql`
--- a/packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.gql.js
+++ b/packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.gql.js
@@ -15,6 +15,13 @@ export const CREATE_CUSTOMER_ADDRESS_MUTATION = gql`
     ${CustomerAddressFragment}
 `;
 
+/**
+ * We would normally use the CustomerAddressFragment here for the response
+ * but due to GraphQL returning null region data, we return minimal data and
+ * rely on refetching after performing this mutation to get accurate data.
+ *
+ * Fragment will be added back after MC-33948 is resolved.
+ */
 export const UPDATE_CUSTOMER_ADDRESS_MUTATION = gql`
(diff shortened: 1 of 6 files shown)


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2403.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.73

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.gql.js
```diff
@@ -23,10 +23,8 @@ export const UPDATE_CUSTOMER_ADDRESS_MUTATION = gql`
         updateCustomerAddress(id: $addressId, input: $address)
             @connection(key: "updateCustomerAddress") {
             id
-            ...CustomerAddressFragment
```
sirugh, line 26: Shouldn't we be updating the UI when we make a change to the address?

packages/venia-ui/lib/components/Region/region.js
```diff
@@ -49,7 +52,8 @@ export default Region;
 
 Region.defaultProps = {
     field: 'region',
-    label: 'State'
+    label: 'State',
+    optionValueKey: 'code'
```
sirugh, line 56: Add a comment here explaining that this is probably temporary while we wait on some backend bugs to be resolved. Add the MC ticket number too :)

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.gql.js
tjwiebell: Added comment explaining the removal in 64cc0d5.

packages/venia-ui/lib/components/Region/region.js
tjwiebell: Added in 64cc0d5.

PR #2405, opened 2020-05-19
Handling invalid braintree nonce error while placing order.

revanth0212:
## Description

This PR handles the use case of an invalid nonce being submitted to the backend during the place order mutation. This can happen if a user has created a payment nonce and has not checkout for a while, enough for the nonce to expire.

The UI would never know to refresh the payment information until the backend fails with the invalid payment nonce error when the user clicks on the place order button.

I have handled payment information errors in this PR, but also made it possible for other developers to add other error handlers in Shipping Information and  Shipping Method.

```
Checkout Page
     \__ useCheckoutPage
     \__ Shipping Information 
     \__ Shipping Method
     \__ Payment Information
              \__ usePaymentInformation
```

This PR introduces a new error class called `CheckoutError` which will be passed from the `useCheckoutPage` talon to Shipping Information, Shipping Method, and Payment Information. Each of those components should have handlers to handle this kind of an error. In this PR, I have added a handler for the `CheckoutError` in the `usePaymentInformation` talon.

## Related Issue
Closes [PWA-556](https://jira.corp.magento.com/browse/PWA-556)

### Verification Stakeholders
@jimbo 
@sirugh 

### Verification Steps
1. Go to the deployed app and add an item into the cart.
2. Go to the checkout page.
3. Complete Shipping Information, Shipping Method, and Payment Information.
4. A payment nonce should have been created and update on the cart in the backend. Now we have to invalidate that nonce to create the error. Use grapiql or any method to submit a mutation on the cart to replace the nonce with an invalid nonce.

Here is the mutation:

```
mutation setSelectedPaymentMethod($cartId: String!, $paymentNonce: String!) {
  setPaymentMethodOnCart(input: {cart_id: $cartId, payment_method: {code: "braintree", braintree: {payment_method_nonce: $paymentNonce, is_active_payment_token_enabler: false}}}) {
    cart {
      id
      selected_payment_method {
        code
        title
      }
    }
  }
}
```

and the query params:

```
{
  "cartId": "CURRENT_CART_ID",
  "paymentNonce": "ANY_STRING"
}
```

5. Once that is a success, click the place order button, it should show an error toast saying the payment method is invalid or something on those lines and the UI should ask you to re-enter the payment information.
6. Now if you click the place order button, it should work fine.

[Demo LIink](https://adobe-my.sharepoint.com/personal/annavara_adobe_com/_layouts/15/guestaccess.aspx?guestaccesstoken=1ygavSSmPUGT49OrH8n7RL8m0z6S7ISCaDhse6LAgS0%3D&docid=2_07526f07501314c2fadb5785729a85d27&rev=1&e=sQTFBi) - The error message in the demo has been changed. I have stripped off `GraphQL error:` string from the error message. Below is the updated screen shot for that.

Error screenshot:

image

## Checklist


packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -116,6 +123,18 @@ export const useCheckoutPage = props => {
...
+    const revertPaymentInformationDone = useCallback(() => {
+        /**
+         * This step will reset the checkout step to PAYMENT.
+         *
+         * This function will be called when the placeOrder mutation
+         * has failed due to payment information issues. By resetting
+         * the step to PAYMENT, we will make the user re-enter the
+         * payment information.
+         */
+        setCheckoutStep(CHECKOUT_STEP.PAYMENT);
```
sirugh, line 135: Instead of threading a callback to the child you can just do this in a try/catch inside `handlePlaceOrder`.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -57,25 +57,28 @@ const CheckoutPage = props => {
...
+        revertPaymentInformationDone
     } = talonProps;
 
     const [, { addToast }] = useToasts();
 
     useEffect(() => {
         if (hasError) {
+            const message =
+                error && error.message
+                    ? error.message
```
revanth0212, line 70: Can we display the same error we receive from the GraphQl server? At times, it can contain text like `GQL Error: *******` which might not be useful for the user.

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
```diff
@@ -0,0 +1,42 @@
...
+    };
+
+    hasPaymentExpired = () => {
+        /**
+         * TODO
+         *
+         * This is temp stuff, need to have a better check.
+         */
+        return this.error.graphQLErrors.map(({ message }) =>
+            message.includes(paymentErrorMessage)
```
sirugh, line 23: Maybe it doesn't matter that the payment is specifically expired. I'm not sure of all the ways that the specific error could be thrown but perhaps we don't even need a custom error class. With GQL mutations we can handle errors as they occur which means we won't ever be using this class for handling something like a `submitShippingMethodError` since that error will be handled by the mutation invoker, `useShippingMethod.js`.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2405.pwa-venia.com : LH Performance Expected 0.85 Actual 0.62
https://pr-2405.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.65

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -115,6 +136,27 @@ export const usePaymentInformation = props => {
         onCompleted: onPaymentDetailsQueryCompleted
     });
 
+    const handleExiredPaymentError = useCallback(() => {
+        clearPaymentDetails({ variables: { cartId } });
+        resetReviewOrderButtonClicked();
+        setHasData(false);
+        onError();
```
sirugh, line 145: Use a try-catch in the checkout page so you don't have to thread the `onError` down to the child nor do you need to reset the review order button.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -115,6 +136,27 @@ export const usePaymentInformation = props => {
...
+     * Effects
+     */
+
+    useEffect(() => {
+        if (
+            checkoutError &&
+            checkoutError instanceof CheckoutError &&
+            checkoutError.hasPaymentExpired()
+        ) {
+            handleExiredPaymentError();
```
sirugh, line 158: Since `handleEx(p)iredPaymentError` is only called here I would just call the `clearPaymentDetails` and `setHasData` functions directly here.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -105,6 +113,19 @@ export const usePaymentInformation = props => {
...
     );
 
+    const clearPaymentDetails = useCallback(() => {
+        client.writeQuery({
+            query: getPaymentNonceQuery,
+            data: {
+                cart: {
+                    __typename: 'Cart',
+                    id: cartId,
+                    paymentNonce: null
```
sirugh, line 125: I don't think that `usePaymentInformation` should know about the nonce as it is specific to braintree/credit card.

That said, after working on payment information I understand the component heirarchy a bit more. `CreditCard` is not mounted if you are in a "done editing" view which means that there's no way to pass the error to that component and have it wipe CC specific data like nonce or billing address.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -105,6 +113,19 @@ export const usePaymentInformation = props => {
         [onSave]
     );
 
+    const clearPaymentDetails = useCallback(() => {
```
revanth0212, line 116: This will make sure even if the user refreshes the page, the summary won't be shown because `hasData` will be false.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentInformation.gql.js
```diff
@@ -17,11 +17,21 @@ export const GET_PAYMENT_DETAILS = gql`
     }
 `;
 
+export const GET_PAYMENT_NONCE = gql`
```
sirugh, line 20: You should just get this from `creditCard.gql.js`

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -57,25 +57,28 @@ const CheckoutPage = props => {
...
                 type: 'error',
                 icon: errorIcon,
-                message:
-                    'Oops! An error occurred while submitting. Please try again.',
+                message,
                 dismissable: true,
                 timeout: 7000
             });
 
-            if (process.env.NODE_ENV !== 'production') {
```
sirugh, line 76: Why remove this check?

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -115,6 +136,27 @@ export const usePaymentInformation = props => {
         onCompleted: onPaymentDetailsQueryCompleted
     });
 
+    const handleExiredPaymentError = useCallback(() => {
```
revanth0212, line 139: I have tried to reset the `selected_payment_method` on the cart as well, but the GQL server does not allow us to do so. Once set, the `selected_payment_method` can not be changed to `""` which is the default value when a cart is created.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
sirugh, line 70: We don't want to be displaying backend-specific things to users like `GraphQL error:...`. Keep the original error message/error log please.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
revanth0212: If I have the try-catch and the throw statement in the `useCheckoutPage` talon, how will the children catch it? It will only propagate upwards, right?

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
revanth0212: There are different errors that can happen during the place order mutation and some of them are not related to payment information. Check this out: https://devdocs.magento.com/guides/v2.3/graphql/mutations/place-order.html#errors

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: I am not sure how will throwing an error in the parent will propagate to the child.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
revanth0212: It is just a choice of style. I would prefer to leave the 4 function calls in their own function. So when someone reads the if condition, they will know that the expired token will be handled. They don't need to know how, unless if they want to, they can check the function itself.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
revanth0212: Exactly. Also, it is not a rule but it can be an understanding that any payment method which has to save something related to payment on the cart in the cache, should place it in the key called `paymentNonce`. That way, we can handle clearing off of the cache for any payment method in the payment information component.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentInformation.gql.js
revanth0212: It is not specific to CC. In general any payment method. Also, my understanding is that we were fine with duplicating GQL queries in the interest of leaving component-specific GQL queries in `component.gql.js`.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: We have `drop-consoles: true` for production env in webpack. It will drop all console.* statements in production and not purge them in development.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: Yeah, I asked the same question above. But one advantage of showing the error would be to let them know what exactly is the error because an error can be any one of [these](https://devdocs.magento.com/guides/v2.3/graphql/mutations/place-order.html#errors).

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: `placeOrderError` is still that same error that is thrown but my point is that logic specific to checkout page should be done within the checkout page callback and logic specific to error handling for payment information such as removing the nonce should be done down the line. You've got a combination of these two things, I'm just asking that you split them.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentInformation.gql.js
sirugh: Technically the `paymentNonce` is our own construct that we created to handle persisted payment details for display _within_ the credit card component. My point is that the parent, `paymentInformation` does not and should not be aware of what information a specific payment method needed in order to be rendered "complete".

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
sirugh: I commented this in multiple places, but for posterity: `placeOrderError` can be passed down (or some combination of the error).

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
sirugh: Yea I remember this being a thing. Fortunately we just display "done" state based on the completion of the child method and not based on `selected_payment_method`.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
sirugh: Is `error.message` just the message string or the one with `GraphQL: <message>`? I just would prefer we strip off `GraphQL Error:` at least.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
sirugh: Yea - I called the same thing out to @zetlen a while back but I can't find the specific comment. I think it's fine to just be explicit with our console logging. It'll get removed in a prod build either way so there's no reason to make this change.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
sirugh: >  any payment method which has to save something related to payment on the cart in the cache, should place it in the key called `paymentNonce`.

I like the idea here but I don't think `paymentNonce` is a good name for it 😄Probably worth a small refactor here to rename this property. Perhaps `selectedPaymentDetails`?

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
zetlen: I think we shouldn't rely on drop_console because it's a little bit magic, and console logging is not the only kind of development-only code we may write. There's potentially other third-party libraries, or functions with different names, that we'll want to use when showing friendly errors, and the minifier won't catch those.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentInformation.gql.js
revanth0212: I agree with you totally but like you mentioned since the payment method component will be unmounted, how can we clear something from the parent which only the child has information about?

To make this simple, I said, maybe if the parent knows that all the child components if they need to save something, they will save in the `paymentNonce` key and the parent can simply delete it.

I am up for suggestions on this.

sirugh: Yea lets talk about what that prop will be named in the other thread :)

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: It is the while unmodified message string from GQL. That's a nice idea to strip off stuff that we might not need but are we sure that this is the only way the error message will be sent to us?

revanth0212:
How about a `free` component? Somthing like this: https://github.com/magento/pwa-studio/commit/c0372bb36f7afdb54f860186efd6535ae243aaaf

It is a rough draft of my idea. Not the final solution.

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
sirugh: Oh nice find on the doc! May as well add a link in a comment at the top of this file.

Also funny enough, the `transaction was declined` error isn't on that page 😆

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
sirugh: > are we sure that this is the only way the error message will be sent to us?

Good question for the GQL team - will the error format ever change?

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
revanth0212: You are right, I can have the try-catch and send the rest of the props down the `PaymentInformation` component/talon. But I feel, sending down a function to call when an error occurs is better than having some logic in try-catch in the parent and some in the child even though the function sent down to the child executes in the parent context.

That said, the function in question, `revertPaymentInformationDone` sent as `onError` is called as the last function in `handleExiredPaymentError`. If instead of sending it as a prop, we call it in the try-catch, it will be called first. The child component won't have control over when to call the function.

TLDR, I feel comfortable sending it as a prop and having the child components handle the error and the onError handler. In this case, it is `PaymentInformation`, but once we expand this to check for errors in `ShippingInformation` and `ShippingMethods` they can call the handlers themselves when they think is the right time.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Commented below in another thread about my thoughts.

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
revanth0212: We may have to refine/add new error checks to extend this error class to other checkout page steps.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
revanth0212: image

revanth0212: Looks much better now.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
sirugh: @revanth0212, I think the suggestion here is to add the condition back in.

revanth0212: Yeah I have done in it a commit that I haven't pushed yet.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
sirugh: OK I get. it! In the future we may get other errors during `placeOrder` that are related to other checkout components besides payment info. So this pattern would delegate the responsibility for handling the error to the component that caused it. 

I'm on board with this but I do think we can at least move `resetReviewOrderButtonClicked` to `useCheckoutPage` handler as that's a global concern we will want to do every time we get an error.

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
```diff
@@ -1,11 +1,20 @@
...
     'Unable to place order: Transaction has been declined. Please try again later.';
 
+const removeGQLTag = rawMessage => rawMessage.replace(/GraphQL error:/, '');
+
 class CheckoutError extends Error {
     constructor(gqlError, ...params) {
         super(params);
         this.name = 'CheckoutError';
-        this.message = gqlError.message;
+        this.message = removeGQLTag(gqlError.message);
```
sirugh, line 17: I don't think we have to explicitly strip out the gql tag. There is a property on the graphql errors that is just the message after the colon. Probably something like `graphqlErrors[0].message`. I'd have to debug it to see the specific props but I know it's there.

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
```diff
@@ -0,0 +1,46 @@
...
+
+class CheckoutError extends Error {
+    constructor(gqlError, ...params) {
+        super(params);
+        this.name = 'CheckoutError';
+        this.message = removeGQLTag(gqlError.message);
+        this.error = gqlError;
+    }
+
+    isPaymentInfoError = () => {
```
sirugh, line 21: Are you using this function? I couldn't find reference. Also I see that the below name `paymentHasExpired` doesn't exist. I think you mean `hasPaymentExpired`?

revanth0212: I don't see any `paymentHasExpired` anywhere. Sorry I didn't get you, buddy.

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
revanth0212: `gqlError.message` is the generic error from the server. Since `graphqlErrors` is an array and each one of them has a `meesage` field, how would we know which one to display? Either we can display all the `graphqlErrors.message`'s  or just show the generic message for the whole error object which is `gqlError.message`.

sirugh: I wonder if `error.message` always equivalent to the message from the first element in `graphQLErrors` with `GraphQL Error`.

[Image from Gyazo](https://gyazo.com/9f210ef668d7c6615e652460283e011f)

In all other cases that we reference `graphqlErrors` in this codebase we either iterate over each element in the array and handle it separately or we treat the first element in the array as the one we will "handle" either by displaying a relevant message or logging it.

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
sirugh: You found it ;)

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
sirugh: Gotcha - I'd just leave the comment unresolved then. When you push it'll resolve it for you and we avoid confusion.

tjwiebell: The argument that we should add this condition because we can't rely on drop_console is contradictory. If someone removes that flag, I feel like their intention is to see console output, which this would prevent since its behind a condition. Are we 100% on this decision?

revanth0212: Nice point @tjwiebell. Haven't thought about it that way. @zetlen any thoughts?

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: #2394 had logic in the place order catch block for setting review order button to false and setting checkout step to payment. If we're going to delegate the errors to the children for handling then we don't need the catch logic anymore (though we still should log the error for development purposes). So I would remove lines 139 and 140 in this file and perhaps replace them with a comment about how the child components handle submission errors.

Additionally, we also started passing `setCheckoutStep` and `resetShouldSubmit` to [payment information](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js#L141-L142).

Instead of writing the error handler here in checkout, could you define it in the `usePaymentInformation` talon? We don't need to define it here and pass it down as nothing else uses it.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
revanth0212: @sirugh should we still change the prop? If so, I would like to handle it as part of a diff PR. I would not like to expand the scope of this PR. What do you think?

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -1,13 +1,16 @@
...
 import { useAppContext } from '../../../context/app';
 import { useCartContext } from '../../../context/cart';
+import CheckoutError from '../CheckoutError';
 import { CHECKOUT_STEP } from '../useCheckoutPage';
 
 /**
  *
  * @param {Function} props.onSave callback to be called when user clicks review order button
+ * @param {Object} props.checkoutError an instance of the `CheckoutError` error that has been generated using the error from the place order mutation
+ * @param {DocumentNode} props.queries.getPaymentDetailsQuery query to fetch selected payment method and payment nonce details
```
sirugh, line 13: Replace this with the nonce query doc. the other query is already included in the docs on line 16.

dpatil-magento:
QA Pass, good to merge after approval.

packages/peregrine/lib/talons/CheckoutPage/CheckoutError.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Other checkout place order mutation errors can include
+ * the following list:
+ *
+ * https://devdocs.magento.com/guides/v2.3/graphql/mutations/place-order.html#errors
+ */
+
+const paymentErrorMessage =
```
jimbo, line 8: Maybe leave a comment explaining that this a heuristic, rather than something to be rendered?

jimbo: This may be the correct use case for an enum containing all of the constants from the [documentation linked elsewhere](https://devdocs.magento.com/guides/v2.3/graphql/mutations/place-order.html#errors). Perhaps import this one?

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
jimbo: I think `paymentNonce` should be fine as a name. I'd also be concerned about tearing that name out of every place we have it currently.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
jimbo:
```suggestion
    const handleExpiredPaymentError = useCallback(() => {
```

Make sure to update all references to this constant.

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentInformation.js
```diff
@@ -207,10 +231,22 @@ export const usePaymentInformation = props => {
...
+    const handleExiredPaymentError = useCallback(() => {
+        setDoneEditing(false);
+        clearPaymentDetails({ variables: { cartId } });
+        resetShouldSubmit();
+        setCheckoutStep(CHECKOUT_STEP.PAYMENT);
+    }, [resetShouldSubmit, setCheckoutStep, clearPaymentDetails, cartId]);
+
+    useEffect(() => {
+        if (
+            checkoutError &&
```
jimbo, line 243: Maybe remove this, if it's redundant with the next line.

PR #2298, opened 2020-03-31
[PWA-355]: Add richer extension points and documentation

zetlen:
## Description

Add features to the extensibility framework to support rich integration with Peregrine talons. This enables extensions that can deeply integrate analytics and segmentation code.

### Features
- Changed Buildpack `wrapEsmModules` target to a lower-level, more versatile `transformModules` target
  - Original module-wrap functionality is still available, using a transform of type `source` and with transformModule `wrap-esm-loader`.
  - New transform type `babel` allows for AST manipulation.
  - Interceptors of `transformModules` can only intercept their own files.
- Added Peregrine `talons` target for wrapping talon functions
  - Only `App.useApp` and `ProductFullDetail.useProductFullDetail` implemented so far
  - Infinitely composable
- Added VeniaUI `routes` target for adding custom routes
  - Refactored Venia's existing custom routes to dogfood this target
- Added a suite of test helpers for developers making their own extensions
  - Mock builds and build buses for simulating a full build with codemods
  - Mock targets for testing intercept behavior
  - Mock Webpack loaders for testing source code transforms

### Changes
- Refactored `MagentoResolver` to support some common development scenarios
  - Symlinked packages
  - Packages referencing each other from targets
  - Resolving a file path using Webpack rules, before running the Webpack compiler
  - Legacy behavior still supported
- Refactored `configureWebpack` into smaller modules for creating different parts of Webpack config
  - Test helpers can now simulate a closer-to-real Webpack config by composing these together
  - transformModules target implementation now encapsulated in JS module rules
- Upgraded Jest to v25
  - Performance and snapshot reliability were taking a dive with the extra tests
  - Changed some harnesses to be compatible with v25
  - Changed some snapshots to be more tolerant of versioning

### Docs
- Refactored several Buildpack modules to be more readable structures
- Put JSDoc comments on everything
  - Properly formatted JSDoc comments will enable intellisense in VSCode and IntelliJ environments
  - @jcalcaben may also be able to generate formatted documentation from the JSDoc comments
- Wrote an introductory developer and user tutorial for Targets

## Related Issue
Closes #PWA-355.

### Verification Stakeholders
@magento/pwa-studio-team 

### Verification Steps
There is a step-by-step tutorial in the getting started guide! This can serve as a verification repro.
[Quick Start](https://github.com/magento/pwa-studio/blob/zetlen/talon-targets/pwa-devdocs/_drafts/pwa-studio-fundamentals/extensibility-with-targets/index.md#quick-start)

[...]

PR #2406, opened 2020-05-19
update docker image

dpatil-magento:
## Description

Current docker image is too old and once https://github.com/magento/pwa-studio/pull/2298 is merged, build and deploy jobs will fail. Updated to near latest node 12 image.

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. AWS build/deploy jobs should pass.
2. on local `bash docker/run-docker` should work.

## Checklist

### DIFF
--- a/dev.dockerfile
+++ b/dev.dockerfile
@@ -2,7 +2,7 @@
 # This file is intended to be used with ./docker-compose.yml #
 ##############################################################
 
-FROM node:10.14.2-alpine as build
+FROM node:10.20.1-alpine as build
 # working directory
 WORKDIR /usr/src/app
 
@@ -43,7 +43,7 @@ RUN yarn install --frozen-lockfile
 RUN yarn run build
 
 # MULTI-STAGE BUILD
-FROM node:10.14.2-alpine
+FROM node:10.20.1-alpine
 # working directory
 WORKDIR /usr/src/app
 # node:alpine comes with a configured user and group
--- a/dev.dockerfile
+++ b/dev.dockerfile
@@ -2,7 +2,7 @@
 # This file is intended to be used with ./docker-compose.yml #
 ##############################################################
 
-FROM node:10.20.1-alpine as build
+FROM node:12.16.3-alpine as build
 # working directory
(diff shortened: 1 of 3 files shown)


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2406.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.68

dpatil-magento:
Tested with #2298 changes in https://github.com/magento/pwa-studio/pull/2408

fooman:
If #2298 breaks on node 10 shouldn't we also bump https://github.com/magento/pwa-studio/blob/develop/package.json#L89 to 12?

dpatil-magento:
> If #2298 breaks on node 10 shouldn't we also bump https://github.com/magento/pwa-studio/blob/develop/package.json#L89 to 12?

 Specifically, Its node 10.14.1, which is too old. anything > 10.14.1 works. 
`error jest@26.0.1: The engine "node" is incompatible with this module. Expected version ">= 10.14.2". Got "10.14.1"`

zetlen:
@dpatil-magento We could choose to declare what point-release ranges of LTS we support. It's not realistic for us to say we support all of 10.x, or all of 12.x, but we could choose a realistic set.

dev.dockerfile
```diff
@@ -2,7 +2,7 @@
 # This file is intended to be used with ./docker-compose.yml #
 ##############################################################
 
-FROM node:10.14.1-alpine as build
+FROM node:12.16.3-alpine as build
```
zetlen, line 5: Is it our policy to always specify an exact version when pulling the docker image? This could auto-update to the most recent LTS version if you used the alias `node:lts-alpine`.

dpatil-magento: We use exact working version to avoid any bugs in latest version might have.  In this way our CI env would be more stable.

dpatil-magento: But I can have test pr to try `node:lts-alpine`  and track it for 10-15 days and then discuss the outcomes.

PR #2420, opened 2020-05-22
Fix dev server unhandled error

jimbo:
## Description
Fix an unhandled error wherein the dev server attempts to read tap into the webpack compilation to read GraphQL query files. It looks like the shape of an argument changed somewhere.

## Related Issue
Maybe related to PWA-169.

### Verification Steps

1. `yarn watch:venia`
2. Verify that the node process does not warn about an unhandled error.
3. Verify that the app is successfully built and served.

## Checklist


packages/pwa-buildpack/lib/WebpackTools/PWADevServer.js
```diff
@@ -164,54 +164,41 @@ const PWADevServer = {
...
+
                 const gatheringQueryTabs = new Promise((resolve, reject) => {
                     const { compiler } = server.middleware.context;
-                    compiler.hooks.done.tap(
-                        'PWADevServer',
-                        async ({ stats }) => {
-                            /**
-                             * Stats in an array because we have 2 webpack child
-                             * compilations, 1 for client and other for service worker.
-                             */
```
jimbo, line 180: Is this still true? The argument to this function certainly wasn't an array; perhaps this is still the case in production, though?

Also, the argument to this function appears to be `Stats {}` itself. I don't see how we could have ever been destructuring `stats` from the argument in the first place.

revanth0212: Nice catch @jimbo. We changed the [webpack config to only have 1 compilation](https://github.com/magento/pwa-studio/pull/2390/files#diff-1acbde853507ec5d5dac59ef5301a593R81) during the build process.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2420.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92
https://pr-2420.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.37, LH Best Practices Expected 1 Actual 0.92
https://pr-2420.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.48, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT TTFB Expected 300 Actual 1790.6666666667

packages/pwa-buildpack/lib/WebpackTools/PWADevServer.js
revanth0212: Funny, the tests failed now. May be we were mocking stuff and using that to test the plugin. Hence it never failed when I changed the `webpack.config.js` in #2390.

jimbo: Fixed!

PR #2421, opened 2020-05-22
[PWA-595] GraphQL Build Warning

tjwiebell:
## Description

During build our graphql validation script is throwing the following warning:

useEditableForm.js
Warning: fragment with name SelectedShippingMethodFragment already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names

## Related Issue

* [[PWA-595](https://jira.corp.magento.com/browse/PWA-595)] GraphQL Build Warning

### Verification Steps

1. `yarn build`
2. Verify warning about fragments are no longer displayed

This change touched both Cart and Checkout experiences for Shipping Methods, so test all areas intersecting with Shipping Method selection.

Cart
1. Add items to cart and navigate to `/cart`
2. Estimate your shipping
3. Verify that you can select methods and selection is displayed in order summary

Checkout
1. Add items to cart and navigate to `/checkout`
2. Enter shipping information
3. Select shipping method
4. Verify selection is displayed in order summary

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2421.pwa-venia.com : LH Performance Expected 0.85 Actual 0.57, LH Best Practices Expected 1 Actual 0.92
https://pr-2421.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.38, LH Best Practices Expected 1 Actual 0.92
https://pr-2421.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.45, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

tjwiebell:
@supernova-at - That was almost exactly an alternative approach I was going to take, but didn't want to make any architectural decisions about how those would be colocated in this scope. There are also some rules we would need to align on, as I don't think I would try to share fragments between these components, since they would never be mounted at the same time.

I also had a concern that this fragment library would just become a mirror of schema types, where at some root layer we'd have a fragment for each complex field type. This feels like we'd be going right back to the structure we're getting away from, where we'd just have a `Fragments` directory at the app root that all components would use. To me, that would encourage REST-like behavior where someone might choose the Fragment that was close enough to what they needed, or we would default to fetching more than we needed out of convenience.

Definitely a good topic for discussion, but need to align on what the objectives of shared fragments should be.

PR #2433, opened 2020-05-26
Order confirmation page multi refresh fix

revanth0212:
## Description

`useCheckoutPage` talon has a `useEffect` that fetches checkout details everytime the `cartId` changes. When the user places an order, a new `cartId` will be created because the old one will be purged. Due to this, the `useEffect` was being triggered and in the process `isLoading` was changing to `true`. Because of this, the user will see a loading screen for a bit and the order confirmation page will be reloaded. This is un-intended behavior.

To fix this, I have added a check to only call the checkout details query if the order has not been placed.

The talon involved in this PR does not have a tests file defined. I need to create and add some tests.

## Related Issue
Closes [PWA-580](https://jira.corp.magento.com/browse/PWA-580)

### Verification Stakeholders
@sirugh 

### Verification Steps
1. Go to the deployed app and try to check out a product.
2. When the order confirmation page loads, there should not be page flashes.

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2433.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92, WPT TTFB Expected 300 Actual 484.66666666667
https://pr-2433.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.37, LH Best Practices Expected 1 Actual 0.92
https://pr-2433.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.5, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -56,12 +56,7 @@ export const useCheckoutPage = props => {
 
     const [
         getCheckoutDetails,
-        {
-            data: checkoutData,
-            called: checkoutCalled,
-            client,
-            loading: checkoutLoading
-        }
+        { data: checkoutData, called: checkoutCalled, loading: checkoutLoading }
```
revanth0212, line 59: Removed `client` here since we can use the `client` from `useApolloClient` which is the same object.

revanth0212:
I am gonna add more tests soon.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -147,14 +142,26 @@ export const useCheckoutPage = props => {
...
+         * Get checkout details everytime `cartId` changes
+         * but not if the place order mutation is called
+         * and `cartId` changes because of that.This is
+         * because after the place order mutation is completed
+         * old `cartId` will be purged and a new one will be
+         * created. we wont need checkout details for that
+         * `cartId`, since the order has been placed and the UI
+         * will be in the order confirmation step.
+         */
+        if (cartId && !placedOrder) {
```
sirugh, line 157: Tommy had a really good idea in his auth shipping PR. Instead of conditions in effects to invoke lazy queries we can just use the `skip` property. 

```
https://github.com/magento/pwa-studio/pull/2380/files#r430714912
11:09
I wonder - could you use this instead of the effect conditional?
11:10
If you replace the checkout page query with this we won't need the effect to call the lazy query anyways.
const {
    data: checkoutData
} = useQuery(getCheckoutDetails, {
    skip: !cartId,
    variables: {
        cartId
    }
});
```

revanth0212: I am fine doing it either way. I like this style as well. Just a small change though, the skip variable value will be `cartId && !!placeOrderData`.

> However after I mention this I realize that it is currently possible for a user to log out/log in while in cart/checkout which would require we re-fetch the data. I don't know if this would re-fetch in those cases.

Ideally speaking it should. Technically speaking, it should since the `cartId` changes once you log in or out.

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -0,0 +1,209 @@
...
+});
+
+/**
+ * Tests
+ */
+
+test('Should return correct shape', () => {
+    const { talonProps } = getTalonProps(props);
+
+    expect(talonProps).toMatchSnapshot();
```
supernova-at, line 163: I haven't been using snapshots for these, for what it's worth.

Here's an example: https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/CheckoutPage/__tests__/useShippingMethod.spec.js#L105-L120.

That said there are plenty examples of both approaches so no worries if you don't want to change this.
I just think the snapshot requires more human effort and is less explicit.

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -0,0 +1,209 @@
...
+};
+
+/**
+ * beforeAll
+ */
+
+beforeAll(() => {
+    useLazyQuery.mockImplementation(query => {
+        if (query === getCheckoutDetailsQuery) {
+            return getCheckoutDetailsQueryResult();
```
supernova-at, line 135: This is a nice way to do this 💯

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -0,0 +1,209 @@
...
+            loading: false,
+            data: {
+                placeOrder: { order: { order_number: '000000' } }
+            }
+        }
+    ]);
+
+    update();
+
+    expect(getCheckoutDetails.mock.calls.length).toBe(2);
```
supernova-at, line 208: Consider splitting this out into multiple tests so that you can follow the Arrange, Act, Assert pattern (AAA).

More details here: https://freecontent.manning.com/making-better-unit-tests-part-1-the-aaa-pattern/.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: I have tried this and it worked like a charm. When I logged in and out, I could see that the query was being fired.

What do you guys think? Should we stick with this approach?

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -147,14 +142,26 @@ export const useCheckoutPage = props => {
     ]);
 
     useEffect(() => {
-        if (cartId) {
+        const placedOrder = !!placeOrderData;
```
supernova-at, line 145: I was expecting to see the `called` property of the mutation used here, not the presence of `data`.

This opens up a window between `called` and `data` where the `cartId` could change and this bug could resurface (say, if the place order network call takes a while).

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: Did it work when logging in and logging out? The checkout page reflected the user state ie showed authed user cart after logging in and showed empty cart after logging out?

revanth0212: Yeah, it did. Because when we log we use the `cartId` associated with the user and the query would pick that up and do its magic. Once the user logs out, a new cartId is created so the query checks and realizes there are not cart items on it. That's my understanding.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Nice point @supernova-at. But my doubt is, what happens if the `place order mutation` is called but fails? In that case `called` will be true but that is a false positive for our code. We only want to skip doing this if the mutation is successful.

Also, I don't think the cart id will be cleared without the place order mutation being a success, because of the way the `handlePlaceOrder` is written. `createCart` will await till `placeOrder` is successful.

```es6
    () => {
        await getOrderDetails({
            variables: {
                cartId
            }
        });

        await placeOrder({
            variables: {
                cartId
            }
        });

        await removeCart();

        await clearCartDataFromCache(client);

        await createCart({
            fetchCartId
        });
    }
```

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
tjwiebell: I'm 100% in support of this change, and was about to implement something similar across all lazy queries in scope of Checkout. I've tested that logging in and out triggers a re-fetch as expected, and have not found any problems so far. Please proceed with this change, but try to keep your scope tight; nearly everything in here is a conflict with what I have in flight with Auth'd Shipping Info.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -27,7 +27,7 @@ export const useCheckoutPage = props => {
         false
     );
 
-    const apolloClient = useApolloClient();
```
sirugh, line 30: I'd just leave this variable named `apolloClient` as it's less ambiguous.

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -0,0 +1,209 @@
+import React from 'react';
```
tjwiebell, line 1: Thank you @revanth0212 for adding this test coverage 🙏 was just about to do the same in scope of my PR, glad I took a peek at this.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Quick point, I have tried both the use cases we are discussing here, `useEffect` and `skip`, unfortunately, they are a bit diff. [I have documented it in this video](https://adobe-my.sharepoint.com/personal/annavara_adobe_com/_layouts/15/guestaccess.aspx?guestaccesstoken=052Ws%2BmIduH8kYSiDI2V3buQ%2BMMkwRi2P8pTA%2Bb9IJE%3D&docid=2_088750a8d85aa4b56a605d54aea8537ef&rev=1&e=q74ebI).

TLDR, `useEffects` does not get invoked if the value of the variable changes to the same value like, `123` -> `123` but the `skip` variable does not recognize that. It just sees that the value has changed but not the actual value. It does not compare the new value to the old value. When ever the value changes, if that value passes the conditional, the `useQuery` will perform its action.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
supernova-at: Yeah, I was thinking about that too ... 

If you leave it as is you may want to leave a comment, mostly because the block comment below it says "wait until place order is **called**".

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Found the link of code that is causing the loading screen flash.

The loading screen is controlled by the boolean `isLoading` in the `useCheckoutPage` talon. This is how the `isLoading` is defined:

```es6
isLoading: !checkoutCalled || (checkoutCalled && checkoutLoading)
```

It reads as:
1. if the checkout details query is not called, show the loading screen
2. if it is called and is loading, show the loading screen

If you pay attention to the 2 screenshots below, in case of `useEffect`, `checkoutCalled` is not set to `true` till it is called, ie. the `useEffect` is executed.

In case of `useQuery`, it is called when the component is rendered, hence `checkoutCalled` is set to `true` even on the first render but it is not loading yet. Hence the `isLoading` boolean is false first, then when `checkoutLoading` is actually set to true, `isLoading` is set to true.

I hope this explanation of mine is understandable 🤣 

`useEffect`

image

`skip`

image

sirugh: Maybe we should revisit the apollo docs for loading states: https://www.apollographql.com/docs/react/data/queries/#inspecting-loading-states

revanth0212: Pstephen the savior. Using the `networkStatus` variable from the `useQuery` worked dude.

revanth0212: Should I be happy that the tests failed to catch the change or sad that I have to update them? 🤔

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
revanth0212: Not relevant anymore. Updated the tests to replicate the new loading check. 0efb6563523d2a24a9a8d58b3f98ec6633c436ff

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
revanth0212: I do like that approach @supernova-at. Thanks for recommending it. It is quite explicit about the shapes of the returned props. It does save us time and effort of not updating the snapshots every now and then.

But, in this component, there are a lot of props being returned and I am kind of leaning towards snapshots than explicit type checks in the interest of protecting my laziness 😉.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Not relevant anymore because the loading logic has changed.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -54,25 +55,41 @@ export const useCheckoutPage = props => {
...
+    });
+
+    /**
+     * For more info about network statues check this out
+     *
+     * https://www.apollographql.com/docs/react/data/queries/#inspecting-loading-states
+     */
+    const isLoading = useMemo(
+        () =>
+            checkoutQueryNetworkStatus ? checkoutQueryNetworkStatus < 7 : true,
```
supernova-at, line 80: The comment is nice but I'd also like us to have a utility enumeration that exports Apollo's network status codes to make this more readable.

https://github.com/apollographql/apollo-client/blob/master/src/core/networkStatus.ts

This would become something like

```js
checkoutQueryNetworkStatus ? checkoutQueryNetworkStatus < ApolloNetworkStatuses.READY : true
```

supernova-at: That source file also exposes an `isNetworkStatusInFlight` utility function that seems to be performing the same check you are here. Other places in our code could benefit from this as well.

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -0,0 +1,358 @@
...
+    getCheckoutDetailsQueryResult.mockReturnValueOnce({
+        data: null,
+        networkStatus: 4
+    });
+
+    const { talonProps, update } = getTalonProps(props);
+
+    expect(talonProps.isLoading).toBeTruthy();
+
+    getCheckoutDetailsQueryResult.mockReturnValueOnce({
```
supernova-at, line 182: Starting from this line down should be its own separate test. "isLoading should be set to false if the checkout details query networkStatus is greater than or equal to 7".

I'd also test seven (7) explicitly since that's where the errors will likely occur.

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -0,0 +1,358 @@
...
+            loading: false
+        });
+
+        const { talonProps } = getTalonProps(props);
+
+        expect(talonProps.isCartEmpty).toBeFalsy();
+    });
+});
+
+test('isGuestCheckout should be negation of isSignedIn from useUserContext', () => {
```
supernova-at, line 315: Careful of tests like this, in my opinion this is a little too implementation-specific and therefore brittle.

sirugh: @supernova-at this is a good point and leads me to believe we should drop the `isGuestCheckout` value and  just be returning `isSignedIn` for use in the UI component.

This raises a question for dev time though. In most of our talons we return some simple boolean states such as loading or render flags such as `isSignedIn`. How should we test those?

revanth0212: I wanted to test the fact that it is dependent on `isSignedIn` which is from another talon. I wanted to do this because, if someone changes the logic, the test has to fail.

All this is based on my assumptions about testing such booleans. 

@sirugh nice idea. Good topic to discuss.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Nice idea @supernova-at. Will do.

Also minor adjustments to the comment above, I don't think we should export the network status from this talon. It can be something from a util or defaults file. I can do that if that is what you are expecting here.

Also another point, `isNetworkStatusInFlight` differs from our implementation. The else case in `isNetworkStatusInFlight` returns false but in our case it has to return true.

supernova-at: Yeah, I was definitely thinking a separate file that this and other talons could use.

revanth0212:
Reverted `NETOWRK_STATUS` changes 🤣 b176795be3d8f3f6fe73aa66615e77269c4949a1

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -69,15 +69,32 @@ export const useCheckoutPage = props => {
...
 
+    /**
+     * For more info about network statues check this out
+     *
+     * https://www.apollographql.com/docs/react/data/queries/#inspecting-loading-states
+     */
+    const isLoading = useMemo(() => {
+        const checkoutQueryInFlight = checkoutQueryNetworkStatus
+            ? checkoutQueryNetworkStatus < 7
+            : true;
```
supernova-at, line 93: Since [network status](https://www.apollographql.com/docs/react/api/react-apollo/#datanetworkstatus) is 1-based, here's an alternate approach that reads a little easier in my opinion:

```js
const checkoutQueryInFlight = !checkoutQueryNetworkStatus || checkoutQueryNetworkStatus < 7;
```

revanth0212: Nice approach Andy, but I would like to stick to the ternary operator approach. In my opinion, in this case, it is easy to read than the `||` approach.

PR #2436, opened 2020-05-27
Portal Component

supernova-at:
## Description

This PR creates a new `Portal` component for rendering content into a DOM node that exists
outside of the DOM hierarchy of the parent / calling component.

The `Portal` component is a simple wrapper around [React Portals](https://reactjs.org/docs/portals.html) that provides a default target if needed.

We had this functionality already actually but the component was called `Modal` and it lead to confusion, especially after we introduced the `Dialog` component which has a "modal" mode.

In order to preserve backwards compatibility, we kept the `Modal` component around for now. It  exists as an alias for `Portal`. In Venia though, all instances of `Modal` have been replaced by `Portal`.

Another option we discussed was to have the `Modal` component compose the `Dialog` component and set its `isModal` prop to `true`. This may be a future improvement and is out of scope here.

## Related Issue

Closes PWA-590.

### Verification Stakeholders

@jimbo 

### Verification Steps

All of the places where the new `Portal` component is used should be checked for regressions.
Please check that the following places function correctly on desktop and on mobile:

1. Filter Modal on the PDP
1. Edit an Item on the Cart Page
1. Edit Shipping Information on Checkout Page
1. Edit Shipping Method on Checkout Page
1. Edit Payment Information on Checkout Page

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2436.pwa-venia.com : LH Performance Expected 0.85 Actual 0.58, LH Best Practices Expected 1 Actual 0.92
https://pr-2436.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.35, LH Best Practices Expected 1 Actual 0.92
https://pr-2436.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.51, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/venia-ui/lib/components/Modal/index.js
```diff
@@ -1 +1,2 @@
-export { default as Modal } from './modal';
+// To keep backwards compatibility, Modal is an alias for Portal.
```
jimbo, line 1: Good work. Perhaps leave a deprecation notice or at least a TODO. I think Zetlen has some examples of deprecation notices in the codebase.

packages/venia-ui/lib/components/Modal/index.js
```diff
@@ -1 +1,3 @@
-export { default as Modal } from './modal';
+// For now, Modal is an alias for Portal.
+// This is deprecated behavior and may be changed in the future.
+export { Portal as Modal } from '../Portal';
```
fooman, line 3: I don't think Modal was part of the last release so I don't think this would be needed.

supernova-at: Unfortunately `Modal` has been around for a while (at least `5.0.0`), so we're not sure who may be using it. I'll be sure to bring this up in community sync.

packages/venia-ui/lib/components/Modal/index.js
```diff
@@ -1 +1,3 @@
-export { default as Modal } from './modal';
+// For now, Modal is an alias for Portal.
+// This is deprecated behavior and may be changed in the future.
```
sirugh, line 2: https://jsdoc.app/tags-deprecated.html

PR #2443, opened 2020-05-30
root component type need to know in some cases by component

sivakumarkoduru:
in child component,custom components we need to know what type of page such as 'PRODUCT','CATEGORY','CMS_PAGE'.

## Description

in child component,custom components we need to know what type of page such as 'PRODUCT','CATEGORY','CMS_PAGE'.

## Related Issue

Closes PWA-861.

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist


tjwiebell:
Created in internal JIRA - [PWA-861](https://jira.corp.magento.com/browse/PWA-861)

supernova-at:
Hi @sivakumarkoduru , thanks for the contribution!

Some feedback:

> i think it is better to save this type in context
> pageType=useContext(type);
> so child components can get the values.

I think a page-level (`RootComponent`-level) context, accessible by a hook like `usePage` or similar is an interesting idea. In addition to the page's type, what other data would be useful to you at this level?

On the housekeeping side of things, please also:

1. Resolve the merge conflicts
1. Run `yarn lint` and `yarn prettier`

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -40,7 +40,7 @@ export const useMagentoRoute = () => {
                         const nextMap = new Map(prevMap);
                         const nextValue = routeError
                             ? { hasError: true, routeError }
-                            : { component, id };
```
sirugh, line 43: Instead of passing to this map, consider using a `reactiveVar`: https://www.apollographql.com/docs/react/local-state/reactive-variables/ that provides access through apollo client.

supernova-at:
I didn't realize this PR had been open for a while. I can make the required changes and move this along.

PR #2445, opened 2020-06-01
Requesting price summary fragment in shipping info mutation.

revanth0212:
## Description

The price summary is not updated on submitting shipping information on the checkout page. This is because the price summary is not being fetched in the mutation result.

Solved by adding price summary fragment to shipping information mutation.

## Related Issue
Closes [PWA-614](https://jira.corp.magento.com/browse/PWA-614)

### Verification Stakeholders
@tjwiebell 

### Verification Steps
1. Go to the deployed app's checkout page.
2. Update the shipping address to a state that has no tax (Texas). There should be no tax section in the price summary.
3. Update the shipping address to a state that has tax (California). There should be a tax section in the price summary.

@fooman mentioned there is a setting in the admin panel to perform tax calculation based on billing address instead. @dpatil-magento when you are testing, try to check for that option as well. The code has logic to render tax changes on billing address change as well.

Lone Star State (No Tax)
image

Golden State (😮 Tax)
image

## Checklist
No tests needed. It is a GQL update.


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2445.pwa-venia.com : LH Performance Expected 0.85 Actual 0.58, LH Best Practices Expected 1 Actual 0.92
https://pr-2445.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.38, LH Best Practices Expected 1 Actual 0.92
https://pr-2445.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.5, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

fooman:
It is possible to configure the back-end to calculate tax based on billing address. Would it make sense to add the same PriceSummaryFragment to the set billing address mutation?

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/shippingInformation.gql.js
```diff
@@ -34,11 +35,13 @@ export const SET_CUSTOMER_ADDRESS_ON_CART = gql`
                 id
                 ...ShippingInformationFragment
                 ...ShippingMethodsCheckoutFragment
+                ...PriceSummaryFragment
             }
         }
     }
     ${ShippingInformationFragment}
     ${ShippingMethodsCheckoutFragment}
+    ${PriceSummaryFragment}
```
supernova-at, line 44: Do we actually need to include the `PriceSummaryFragment` in this query?

Shouldn't the price summary / order summary component already be requesting this data for itself?

revanth0212: I don't think so. This is the same price summary component from the cart page and the `usePriceSummary` talon only re-fetches the price summary data when the cart id changes. If we add checks inside the talon to re-fetch on other variable changes, we will be breaking the existing code. I can still go ahead and make the change if you guys think that is the right way to handle this issue.

jimbo:
> It is possible to configure the back-end to calculate tax based on billing address. Would it make sense to add the same PriceSummaryFragment to the set billing address mutation?

Good callout. I could see how it would make more sense to have it configured that way, too.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/shippingInformation.gql.js
revanth0212: I do think the easy and scalable approach is to let the components fetch what they want in this case shipping information to fetch order summary because it knows that order summary might change because of its operations and let apollo do its magic to let react know which components to re-render on new data, in this case, the order summary component since new data is available. This approach makes things simple and make order summary component/talon less bloated.

supernova-at: Lol oops, GitHub compare messed me up - you didn't actually add the `PriceSummaryFragment` to the `GET_SHIPPING_INFORMATION` query 😝

dpatil-magento:
@revanth0212 Zero checkout flow breaks! 
1. Add `Emilia Cropped Lace Top` product to cart .
2. Go to checkout and enter Texas address.
3. Select Free Shipping 
4. Apply gift card `0BPAZOGI82B8 `, no payment required as order total is < gift card amount.
5. Now edit shipping address to California.

Issue - Even though order total is greater than gift card amount, payement module still stale.
(and vice-versa, if CA address entered first then changed to TX.)

image

revanth0212:
> @revanth0212 Zero checkout flow breaks!
> 
> 1. Add `Emilia Cropped Lace Top` product to cart .
> 2. Go to checkout and enter Texas address.
> 3. Select Free Shipping
> 4. Apply gift card `0BPAZOGI82B8 `, no payment required as order total is < gift card amount.
> 5. Now edit shipping address to California.
> 
> Issue - Even though order total is greater than gift card amount, payement module still stale.
> (and vice-versa, if CA address entered first then changed to TX.)

Nice catch @dpatil-magento.

There are 2 things we have to note here.
1. We have to fetch available payment methods when the shipping address change, which I will add in the next commit. This brings me to another question, should we re-fetch available shipping methods as well, because shipping methods can change when the shipping address has changed. @jimbo @supernova-at should I fetch shipping methods as well along with payment methods in the mutation response?
2. When I added the GQL fragment to fetch the payment methods on shipping address change, you can see that the `Free` payment method is still available even though the total is not 0, but when I refresh the page, the GQL query on the same cart id returns a diff response for available payment methods (does not return free). That seems like a bug in GQL.

On shipping address submit:
image

On refresh:
image

supernova-at:
> @jimbo @supernova-at should I fetch shipping methods as well along with payment methods in the mutation response?

Just to update here, that fragment should exist already and the list of available shipping methods should already update on shipping address change.

revanth0212:
> > @jimbo @supernova-at should I fetch shipping methods as well along with payment methods in the mutation response?
> 
> Just to update here, that fragment should exist already and the list of available shipping methods should already update on shipping address change.

Yeah, you are right. Found it. Thanks Andy.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentInformation.gql.js
```diff
@@ -29,8 +36,10 @@ export const GET_PAYMENT_INFORMATION = gql`
                 }
                 telephone
             }
+            ...AvailablePaymentMethodsFragment
         }
     }
+    ${AvailablePaymentMethodsFragment}
 `;
 
 /* eslint-disable graphql/required-fields */
```
tjwiebell, line 45: Not related to this PR, but couldn't help but notice this rule get disabled, and never re-enabled. This is probably preventing any file after this from being properly linted. Consider fixing in this scope, or please create a followup task to resolve this.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/creditCard.gql.js
```diff
@@ -1,5 +1,8 @@
 import gql from 'graphql-tag';
 
+import { PriceSummaryFragment } from '../../CartPage/PriceSummary/priceSummaryFragments';
```
tjwiebell, line 3: On line 28 in this file too, I think we intended to turn the lint rule back on.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/paymentInformation.gql.js
revanth0212: Good catch Tommy. Havn't noticed that. Also looks like my linter is not working 😓 . Fixed it in https://github.com/magento/pwa-studio/pull/2445/commits/06608bff1cbdf6ac405673efc877af8d56813f59

PR #2453, opened 2020-06-02
Update Home Page content after cache hit (#2230)

### LINKED ISSUE #2230: [bug]: Home Page query is not fetching from network/updating cache
**Describe the bug**
I've noticed that Apollo in PWA Studio isn't updating its local cache (`apollo-cache-persist` in localStorage) from the network (for page content in my case), it's only serving from cache. This means that making changes to a page doesn't ever get reflected in the app until this cache is manually cleared.

**To reproduce**
1. Navigate to homepage in app (allowing content to be cached)
2. Open Admin
3. Go to Content > Pages > Home page
4. Edit the home page and save
5. Refresh the application (the old content will display, which is expected for the _first_ reload)
6. Refresh the application again
7. The old content will be displayed again, as it will for any number of times that the page is refreshed
8. Manually clear the localStorage property, `apollo-cache-persist`
9. Refresh the application
10. _Now_ the updated home page data will be displayed

**Expected behavior**
On the first reload, the old data _should_ be displayed, but in the background a request should be sent to the server to get the latest version of the homepage. The returned data should then being pushed into the localStorage cache, so that on the second reload (either a user is prompted through an action similar to "There is a new version of the app available, please refresh", or they just happen to reload again), they will see the new data on the page

**Possible solutions**
The solution is to make a network request to update Apollo cache after it has server up its cache. I'm not familiar enough with Apollo to say how this would be done specifically

 - Device [e.g. iPhone6, PC, Mac, Pixel3]: PC
 - OS [e.g. iOS8.1, Windows 10]: Centos 7
 - Browser [e.g. Chrome, Safari]: Chrome
 - Browser Version [e.g. 22]: 80.0.3987.132
 - Magento Version: 2.3.3
 - PWA Studio Version: 5.0.1

- [x] `venia-concept`
- [x] `venia-ui`

davemacaulay:
## Description

Apollo will not query the GraphQL endpoint for a new version of CMS Page content due to the usage of the default fetch policy, cache only. This PR updates this cache policy to use `cache-and-network` (https://www.apollographql.com/docs/react/api/react-apollo/#optionsfetchpolicy) which will ensure the cached data is rendered, then a request is made to retrieve any new data if available.

This fetch policy ensures a quick user experience while ensuring up to date content is rendered on the homepage. If there was a change in content between the cache and GraphQL Apollo automatically handles re-rendering this.

## Related Issue
Closes #2230 .
https://jira.corp.magento.com/browse/PWA-471

### Verification Stakeholders
@jimbo 
@sirugh 
@brendanfalkowski 
@dhaecker 

### Specification

https://www.apollographql.com/docs/react/api/react-apollo/#optionsfetchpolicy

### Verification Steps
- Navigate to homepage in app (allowing content to be cached)
- Open Admin
- Go to Content > Pages > Home page
- Edit the home page and save
- Refresh the application (the old content will display, which is expected for the _first_ reload)
- Refresh the application again
- The old content will be displayed again, as it will for any number of times that the page is refreshed
- Manually clear the localStorage property, {{apollo-cache-persist}}
- Refresh the application
- _Now_ the updated home page data will be displayed

**New Experience:**

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2453.pwa-venia.com : LH Performance Expected 0.85 Actual 0.61, LH Accessibility Expected 1 Actual 0.89, LH SEO Expected 1 Actual 0.9, LH Best Practices Expected 1 Actual 0.92
https://pr-2453.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2453.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.49, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/venia-ui/lib/components/Header/header.css
```diff
@@ -39,6 +39,9 @@
 .primaryActions {
     grid-area: primary;
     justify-self: start;
+    display: grid;
+    grid-template-columns: 1fr 1fr;
```
jimbo, line 43: I think `auto` is the better size for these columns, and it's okay for the number of columns to be implicit here, so I think this declaration can be safely removed.
```suggestion
```

jimbo:
> Did you look at LoadingIndicator first, or just make a simple PageLoadingIndicator real fast?

This is essentially a leaf-node component, and I'm okay if those proliferate. It's a thin wrapper around `Icon`; I'm afraid of getting too much more abstract than that. 😅

packages/peregrine/lib/store/reducers/app.js
```diff
@@ -39,6 +40,12 @@ const reducerMap = {
             isOnline: false,
             hasBeenOffline: true
         };
+    },
+    [actions.setPageLoading]: (state, { payload }) => {
```
tjwiebell, line 44: I'm apprehensive to add additional functionality to redux if we plan on removing it. I know we've talked about removal of `cart`, `checkout`, and `user` slices, but we haven't talked about the `app` slice, so I could be mistaken and we do intend to leave it around.

Let's talk about this PR in team time tomorrow, and just make sure we're on the same page. I would want to see this as a `@client` directive in Apollo, with resolvers for querying and mutating. The uncharted territory is if we want to continue using these patterns we're all used to in redux, and add actions and state to AppContext for convenience, or set the expectation that people understand how to query and mutate data with Apollo.

packages/venia-ui/lib/RootComponents/CMS/cms.js
```diff
@@ -27,12 +30,15 @@ const CMSPage = props => {
...
         return fullPageLoadingIndicator;
     }
 
-    if (!data) {
-        return null;
+    // Ensure we mark the page as loading while we check the network for updates
+    if (loading && data && !isPageLoading) {
+        setPageLoading(true);
+    } else if (!loading && data && isPageLoading) {
+        setPageLoading(false);
```
tjwiebell, line 41: These truthy `data` checks are redundant now.

```suggestion
    if (loading && !isPageLoading) {
        setPageLoading(true);
    } else if (!loading && isPageLoading) {
        setPageLoading(false);
```

packages/peregrine/lib/store/reducers/app.js
sirugh: Ah I should have mentioned this option to Dave yesterday. I didn't even consider client state :( I wouldn't want him to have to rewrite all of this but we should raise awareness about the possiblity of using Apollo's cache instead of redux. It's a great topic for team time tomorrow, hopefully everyone can make it :)

packages/venia-ui/lib/RootComponents/CMS/cms.js
tjwiebell: Are we opposed to bitwise operators? I don't see any usage in the code base, but this logic piqued my interest, and I wanted to try and simplify.

```js
if (loading ^ isPageloading) {
    setPageLoading(loading);
}
```

packages/peregrine/lib/store/reducers/app.js
```diff
@@ -39,6 +40,12 @@ const reducerMap = {
...
+        return {
+            ...state,
+            isPageLoading: payload
```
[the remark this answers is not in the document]
davemacaulay: I think a boolean of true / false is more transparent, getting into weird cases where `setPageLoading();` toggles it or flips it to true feels weird to me

packages/venia-ui/lib/components/PageLoadingIndicator/indicator.css
```diff
@@ -0,0 +1,17 @@
+.root {
+    justify-self: start;
+}
+
+.indicator {
+    animation: spin 0.75s linear infinite;
```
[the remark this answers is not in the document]
davemacaulay: Out of interest why do you use base 2?

packages/venia-ui/lib/components/PageLoadingIndicator/indicator.js
```diff
@@ -0,0 +1,22 @@
...
+
+const PageLoadingIndicator = props => {
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    return (
+        <div className={classes.root}>
+            <Icon
+                src={LoaderIcon}
+                size={24}
+                classes={{ icon: classes.indicator }}
```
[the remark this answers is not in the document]
davemacaulay: I moved the animation onto the `.root` style, but we still require the `color` style to be applied to the `<Icon />`

packages/venia-ui/lib/components/Header/header.css
davemacaulay: Just removing causes them to stack into rows.


I also added `grid-auto-flow: column;` to resolve this :)

packages/venia-ui/lib/RootComponents/CMS/cms.js
davemacaulay: @tjwiebell that's really cool, and would certainly work, but I think it'd reduce peoples ability to easily understand the code, as it's not a common operator used in JS.

packages/venia-ui/lib/RootComponents/CMS/cms.js
```diff
@@ -17,8 +18,10 @@ const CMSPage = props => {
         variables: {
             id: Number(id),
             onServer: false
-        }
+        },
+        fetchPolicy: 'cache-and-network'
     });
+    const [{ isPageLoading }, { setPageLoading }] = useAppContext();
```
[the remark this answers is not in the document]
davemacaulay: That's awesome, thank you :)

packages/peregrine/lib/store/reducers/app.js
davemacaulay: I'll be there in the meeting! Sounds like we should make a decision and make sure it's documented and used in all similar features moving forward 😌

packages/venia-ui/lib/components/PageLoadingIndicator/indicator.css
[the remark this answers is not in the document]
jimbo: It's more evenly divisible, which helps when figuring out what a given increment should be. It's the same reason most grid systems are 12-column or 16-column instead of 10-column.

```
# easy to keep dividing by 2
16 => 8 => 4 => 2 = > 1

# becomes increasingly harder to do quick math
10 => 5 => 2.5 => 1.25 => 0.625
```

packages/venia-ui/lib/components/Header/header.css
jimbo: Good call. 👍

packages/peregrine/lib/store/reducers/app.js
jimbo: I almost left a comment about this issue specifically, so thanks for raising it.

As of writing, the Redux store's `app` slice appears to be the best place we have to put this new piece of state. But like @tjwiebell said, our goal is to pare down the usage of our Redux store until we can safely eliminate it, so ordinarily we wouldn't want to expand it like we're doing in this PR.

Obviously though, I don't want to task @davemacaulay with doing the exploratory work of creating a parallel app state context, much less a replacement. However, if we feel like the Apollo-based approach will work in this case, and that it has enough examples elsewhere in the codebase that we could reasonably ask Dave to use it here, I'd be open to it.

packages/venia-ui/lib/components/PageLoadingIndicator/indicator.css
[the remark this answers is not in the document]
davemacaulay: Ah yeah, that makes a lot of sense, thanks for the explanation

packages/peregrine/lib/store/reducers/app.js
davemacaulay: Happy to change this implementation if there is enough good other examples of how it works, let's discuss in developer team time later and work out a plan to move forward 🙂

packages/venia-ui/lib/RootComponents/CMS/cms.js
jimbo: @tjwiebell I'm not opposed to bitwise operators, as long as we're not inventing reasons to use them. It would be totally fine in this case.

That said, I certainly wouldn't ask anyone to use them. Since the code is perfectly clear already, we can leave it as written.

packages/venia-ui/lib/components/PageLoadingIndicator/indicator.js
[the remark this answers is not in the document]
jimbo: Understood. Instead of overriding `icon` here, probably override `root`, and supply `--stroke` instead of `color`. Make sure `--stroke` is in `rgb(x, y, z)` format.

packages/peregrine/lib/store/reducers/app.js
davemacaulay: As discussed Geese will do a follow up story to explore how to use Apollo local cache instead of appContext for instances like this.

dhaecker:
QA approved

PR #2454, opened 2020-06-02
Set cart page is loading while fetching data from network

supernova-at:
## Description

This PR sets the cart page's `isCartUpdating` flag while we are fetching data from the network.

This will not allow users to "Proceed to Checkout" until we know we have the latest data from the network and the user is seeing the most up-to-date information.

Notably, this PR does *not* solve the cart page flash problem - the cart page _purposefully_ shows data returned from its cache first, and updates it with data fetched from the network. It purposefully does not show a loading spinner while we wait for network data (that would defeat the purpose of having cached data).

The "flash" is the page showing the cached data first and then switching to the network data.

PWA-321 will properly add items to the cache and actually solve the flash problem, but even then there is a chance the cache and the network are out of sync, and the changes in this PR are useful to have in place regardless.

## Related Issue

Closes PWA-416.

### Verification Steps

1. Add an item to your cart
1. Change your browser's network settings to "Fast 3G"
1. Navigate to the `/cart` page
1. See that the "Proceed to Checkout" button is disabled until the "get cart details" network call returns

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2454.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92
https://pr-2454.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2454.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.48, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/peregrine/lib/talons/CartPage/__tests__/useCartPage.spec.js
```diff
@@ -4,15 +4,14 @@ import { createTestInstance } from '@magento/peregrine';
 import { useCartPage } from '../useCartPage';
```
revanth0212, line 4: Not sure how granular can we get with testing but I feel we have to add tests around conditionals and effects. For instance, check if `isCartUpdating` is being updated on `loading` change.

dpatil-magento:
@supernova-at Looks good other than being on `/cart` page as Auth user and then logout.

PR #2456, opened 2020-06-03
[PWA-586] Country/Region Occasionally Resets initialValue

tjwiebell:
## Description

There's an empty cache state for these components where the form they are attached to renders before the backend results for the list of available options in these dropdown components is returned. When that backend response arrives, the components re-render with those options, but the initialValue from the parent context doesn't apply, and the component renders an empty selection. I've more commonly seen this in the Region component, where State will be cleared out after an initial flash of the component.

Once this component has loaded once in the app, it's not reproducible, since the components first render now uses that immediate cached response.

Repro:

1. Add Simple item to cart and go to `/cart`
2. Open Estimate Shipping section and submit an estimate (State = TX)
3. Open Dev Tools > Local Storage > Delete `apollo-cache-persist` entry
4. Refresh the page and open Estimate Shipping section

Expected: State data entered is selected in dropdown
Actual: State is empty

## Related Issue

* [[PWA-586](https://jira.corp.magento.com/browse/PWA-586)] Country/Region Occasionally Resets initialValue

### Verification Steps

⬆️  Repro steps can be used for verification above ⬆️ 

## Checklist

- [x] I have updated/added tests to cover my changes, if necessary.

### DIFF
--- a/packages/peregrine/lib/talons/Region/__tests__/useRegion.spec.js
+++ b/packages/peregrine/lib/talons/Region/__tests__/useRegion.spec.js
@@ -1,13 +1,18 @@
+import React from 'react';
+import { act } from 'react-test-renderer';
 import { useQuery } from '@apollo/react-hooks';
+import { useFieldApi, useFieldState } from 'informed';
 
+import createTestInstance from '../../../util/createTestInstance';
 import { useRegion } from '../useRegion';
 
 jest.mock('informed', () => {
     const useFieldState = jest.fn().mockReturnValue({
         value: 'US'
     });
+    const useFieldApi = jest.fn();
 
-    return { useFieldState };
+    return { useFieldApi, useFieldState };
 });
 
 jest.mock('@apollo/react-hooks', () => ({
@@ -25,17 +30,30 @@ jest.mock('@apollo/react-hooks', () => ({
     })
 }));
 
+const Component = props => {
+    const talonProps = useRegion(props);
+    return <i talonProps={talonProps} />;
+};
+
 const props = {
     queries: {}
 };
 
 test('returns formatted regions', () => {
-    const talonProps = useRegion(props);
+    const tree = createTestInstance(<Component {...props} />);
+    const { root } = tree;
+    const { talonProps } = root.findByType('i').props;
+
     expect(talonProps).toMatchSnapshot();
 });
 
 test('returns formatted regions with id as the key', () => {
-    const talonProps = useRegion({ ...props, optionValueKey: 'id' });
+    const tree = createTestInstance(
+        <Component {...props} optionValueKey={'id'} />
+    );
+    const { root } = tree;
+    const { talonProps } = root.findByType('i').props;
+
     expect(talonProps).toMatchSnapshot();
 });
 
@@ -50,6 +68,26 @@ test('returns empty array if no available regions', () => {
         loading: false
     });
 
-    const talonProps = useRegion(props);
+    const tree = createTestInstance(<Component {...props} />);
+    const { root } = tree;
+    const { talonProps } = root.findByType('i').props;
+
     expect(talonProps).toMatchSnapshot();
+});
+
+test('resets value on country change', () => {
+    const mockSetValue = jest.fn();
+
+    useFieldState.mockReturnValueOnce({ value: 'FR' });
+    useFieldApi.mockReturnValue({ setValue: mockSetValue });
+
+    const tree = createTestInstance(<Component {...props} />);
+
+    expect(mockSetValue).not.toHaveBeenCalled();
+
+    act(() => {
+        tree.update(<Component {...props} />);
+    });
+
+    expect(mockSetValue).toHaveBeenCalledWith();
 });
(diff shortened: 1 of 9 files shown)


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2456.pwa-venia.com : LH Performance Expected 0.85 Actual 0.57, LH Best Practices Expected 1 Actual 0.92
https://pr-2456.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2456.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.49, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

tjwiebell:
> Technically speaking `loading` is the only change right? At first, I thought you removed the `initialValue` and `validate` props but they are fed into the same component inside `...inputProps` or `...regionProps`. Nice.

Using `loading` to disable the field while fetching options + the magical `keepState` prop that preserves the value between mounts. This might even fix the issue you had in scope of billing address, where you were passing initalValue to the field instead of the form (this bug might have been why you did that).

> Would it be possible to add snapshot tests to the country and region components?

Absolutely! Since we do tests in QA, is this code approved to get started on those?

tjwiebell:
@revanth0212 - Added tests that are a bit coupled to these form input components, but wanted to put out a challenge, maybe you've solved this problem before. I really _really_ tried to mock the `Field` and `Select` components in these tests, and could not get it to work. With the following mocks, I was getting an unhelpful `PrettyFormatPluginError: val.getMockName is not a function` error.

```js
jest.mock('../../Field', () => props => <mock-Field {...props} />);
jest.mock('../../Select', () => 'Select');
```

revanth0212:
> @revanth0212 - Added tests that are a bit coupled to these form input components, but wanted to put out a challenge, maybe you've solved this problem before. I really _really_ tried to mock the `Field` and `Select` components in these tests, and could not get it to work. With the following mocks, I was getting an unhelpful `PrettyFormatPluginError: val.getMockName is not a function` error.
> 
> ```js
> jest.mock('../../Field', () => props => <mock-Field {...props} />);
> jest.mock('../../Select', () => 'Select');
> ```
> 
> I don't know what's different about the `Select` component, but it seems impossible to mock without getting this error.

`Select` is a component but you are returning a string. I guess that is the error. You are returning a component in case of `Field` though.

tjwiebell:
> `Select` is a component but you are returning a string. I guess that is the error. You are returning a component in case of `Field` though.

dpatil-magento:
QA Pass.

packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -1,21 +1,37 @@
...
+            }
+        }
+    }, [country, regionFieldApi]);
 
     const { data, error, loading } = useQuery(getRegionsQuery, {
         variables: { countryCode: country }
     });
 
-    let formattedRegionsData = [];
+    let formattedRegionsData = [{ label: 'Loading Regions...', value: '' }];
```
revanth0212, line 34: Do we have a test for this? Looks like it is being manipulated in multiple places. A set of tests can be of help.

I see you have added snapshot tests but more verbal ones can be helpful.

packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -1,21 +1,37 @@
...
     const countryFieldState = useFieldState(countryCodeField);
     const { value: country } = countryFieldState;
+    const regionFieldApi = useFieldApi(field);
+
+    // Reset region value when country changes. Because of how Informed sets initialValues,
+    // we want to skip the first state change of the value being initialized.
+    useEffect(() => {
+        if (country) {
+            if (hasInitialized.current) {
+                regionFieldApi.setValue();
```
jimbo, line 23: Maybe indicate this is a reset with a comment.

packages/peregrine/lib/talons/Region/useRegion.js
tjwiebell: Added in ab0adc9

packages/peregrine/lib/talons/Region/__tests__/useRegion.spec.js
```diff
@@ -50,6 +68,26 @@ test('returns empty array if no available regions', () => {
...
+
+    const tree = createTestInstance(<Component {...props} />);
+
+    expect(mockSetValue).not.toHaveBeenCalled();
+
+    act(() => {
+        tree.update(<Component {...props} />);
+    });
+
+    expect(mockSetValue).toHaveBeenCalledWith();
```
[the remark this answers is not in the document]
tjwiebell: Changed in ab0adc9.

packages/peregrine/lib/talons/Region/useRegion.js
tjwiebell: `reset()` from the `fieldApi` had the behavior I wanted, so I made that change in ab0adc9, which should be more self documenting. Thanks for pointing out there was much more to that API than the docs lead you to believe.

PR #2462, opened 2020-06-05
PWA-560: Fix gift card flash on error

davemacaulay:
## Description

Maintain state for whether or not to show the gift card error so it can be correctly updated based on other state in the gift card talon.

## Related Issue
https://jira.corp.magento.com/browse/PWA-560

### Verification Stakeholders
@jimbo 
@dhaecker
@sirugh 

### Verification Steps
1. Add a card
2. Enter the same card and click "Apply Card". You'll get an error.
3. Click "Remove" to remove the original card. The error should disappear.
4. Click "Apply Card". You will not see a flash of error between states.
_See JIRA ticket for video of issue_.

See JIRA ticket for video

## Checklist


packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
```diff
@@ -197,6 +193,15 @@ export const useGiftCards = props => {
         setIsCartUpdating
     ]);
 
+    useEffect(() => {
+        // If either operation results in an error, show the error
+        setShouldDisplayCardError(
+            (mostRecentAction === actions.APPLY && applyCardResult.error) ||
+                (mostRecentAction === actions.CHECK_BALANCE &&
+                    balanceResult.error)
+        );
```
davemacaulay, line 202: Do I need to clean up in this effect? Returning `shouldDisplayCardError` to false when the component is unmounted?

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2462.pwa-venia.com : LH Performance Expected 0.85 Actual 0.57, LH Best Practices Expected 1 Actual 0.92
https://pr-2462.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.37, LH Best Practices Expected 1 Actual 0.92
https://pr-2462.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.49, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
revanth0212: I guess you are fine with that. Since the component using this will be unmounted, we should be good.

One suggestion would be to convert this to `if-else` condition and then set `setShouldDisplayCardError` in those. It makes it easy to read.

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
```diff
@@ -79,49 +78,42 @@ export const useGiftCards = props => {
...
-            setMostRecentAction(actions.CHECK_BALANCE);
-        }
-    }, [formApi, mostRecentAction]);
+        });
+
+        // Clear the input form after successful apply.
+        formApi.reset();
+    }, [formApi, applyCard, cartId]);
+
+    const checkGiftCardBalance = useCallback(async () => {
```
tjwiebell, line 100: Doesn't look like this needs to be a Promise anymore.

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
```diff
@@ -197,6 +150,21 @@ export const useGiftCards = props => {
         setIsCartUpdating
     ]);
 
+    const shouldDisplayCardBalance = useMemo(
```
tjwiebell, line 153: I would expect this memo to be calculated every time, not sure this is an improvement; what was wrong with the current implementation of derived state?

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
```diff
@@ -197,6 +150,21 @@ export const useGiftCards = props => {
...
+        () =>
+            mostRecentAction === actions.CHECK_BALANCE &&
+            Boolean(balanceResult.data),
+        [mostRecentAction, balanceResult]
+    );
+
+    // We should only display the last card error if the most recent action was apply or check and they had an error
+    const shouldDisplayCardError = useMemo(
```
tjwiebell, line 161: Same here; why the memos?

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
davemacaulay: Great catch, thank you!

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
davemacaulay: We went on an adventure fixing this bug, it seems the my refactoring to the rest of the talon resolved the original error, so you're correct. The `useMemo` is no longer needed, I've updated the code :)

packages/peregrine/lib/talons/CartPage/GiftCards/useGiftCards.js
davemacaulay: https://github.com/magento/pwa-studio/pull/2462#discussion_r437593106

sirugh:
You need to remove the `useMemo` import as well, I think. Linter is complaining.

dhaecker:
QA approved

PR #2466, opened 2020-06-07
fix: remove unintentionally spammy loadEnvironment warnings

zetlen:
## Description

Simplify the breaking change warnings in `loadEnvironment`, and add a fail-safe to make sure they don't log forever.

Changes:

- Remove `defaultChanged` and `exampleChanged` warning types, because they're not useful and I was wrong to create them. When a default changes, you either don't care or you've overridden it. When an example changes, you really never care.
- Remove the old warnings about changes from many releases ago.

Additions:

- Add a required "dateAdded" and an optional "warnForDays" property to change entries. They will now only log for an interval of "warnForDays", up to 180 days.
- Add a test for `envVarDefinitions` themselves, to ensure they're compliant. 

## Related Issue
Fixes [PWA-502](https://jira.corp.magento.com/browse/PWA-502)
Fixes underlying problem from #2308 

### Verification Stakeholders
@brendanfalkowski @revanth0212 
### Verification Steps

1. Check out the branch.
2. Run a build. Observe that the CHECKOUT_BRAINTREE_TOKEN is gone.
3. Examine the new tests. Break them experimentally if you want to see what's going on. Or, add new change entries and see what happens.

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2466.pwa-venia.com : LH Performance Expected 0.85 Actual 0.58, LH Best Practices Expected 1 Actual 0.92
https://pr-2466.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.37, LH Best Practices Expected 1 Actual 0.92
https://pr-2466.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.49, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/pwa-buildpack/lib/Utilities/CompatEnvAdapter.js
```diff
@@ -0,0 +1,91 @@
...
+    constructor(definitions) {
+        this._sortedChanges = definitions.changes
+            .slice()
+            .sort((a, b) => a.name > b.name);
+    }
+    apply(env) {
+        const warnings = [];
+        const mappedLegacyValues = {};
+        for (const change of this._sortedChanges) {
+            //
```
sirugh, line 25: Extra // :O

packages/pwa-buildpack/lib/Utilities/CompatEnvAdapter.js
```diff
@@ -0,0 +1,91 @@
+const ONE_DAY_MS = 24 * 60 * 60 * 1000;
+
+const buildpackVersion = require('../../package.json').version;
+
+class CompatEnvAdapter {
+    static isExpired({ dateChanged, warnForDays }) {
```
revanth0212, line 6: Quick question JZ. Why did you make this a `static` method instead of a function outside the `class`? Not that it would make any difference, just to pick your brain.

packages/pwa-buildpack/envVarDefinitions.json
```diff
@@ -184,132 +184,5 @@
...
-            "type": "removed",
-            "reason": "The standard filename 'sw.js' is better for caching, stability, and compatibility, and a PWA Studio project should not register more than one ServiceWorker."
-        },
-        {
-            "name": "PWA_STUDIO_PUBLIC_PATH",
-            "type": "removed",
-            "reason": "The new environment variables DEV_SERVER_HOST and STAGING_SERVER_HOST now provide the host override functionality for the Docker setup."
-        }
-    ]
+    "changes": []
```
revanth0212, line 187: Now that we have the `EnvAdapter` class looking at the expiry date and frequency, how about a sample change for that? Maybe even the brain tree token in question, maybe remind it every 30 days.

packages/pwa-buildpack/lib/Utilities/CompatEnvAdapter.js
zetlen: It's a good question though. Two reasons:
1. It's a pure function that relies on no instance variables, BUT it's only used by this class
2. I needed to [use it in tests](https://github.com/magento/pwa-studio/pull/2466/files#diff-c71856324fef05d7946db9d0cb69ed5dR48-R52)

packages/pwa-buildpack/envVarDefinitions.json
zetlen: It's actually not a frequency measure! It's a time duration during which it would warn every time you run the command.

What if we used the unit test to demonstrate that instead of adding a change here that we don't need?

PR #2477, opened 2020-06-09
[PWA-621] Carry over design tokens from styleguide

anthoula:
## Description

Many of the core values of Venia's new design have been defined in the styleguide. We need to port these to `venia-ui` so we can incorporate these values consistently as we apply the new design. Most of these tokens can be found in `venia-styleguide/src/styles/tokens.css`.

## Related Issue
https://jira.corp.magento.com/browse/PWA-621

## Acceptance 
CSS should use the same patterns as the rest of `venia-ui`:
- Use CSS modules' scoped classes
- Use CSS Grid
- Use CSS custom properties
- Use transitions responsibly
### Verification Stakeholders

@jimbo 
@sirugh 
@dhaecker 
### Verification Steps
- Copy `tokens.css` from `venia-styleguide` to `venia-ui` to carry over the CSS custom properties
- Move `index.css` from `venia-concept` into `venia-ui`
- Ensure that `index.css` imports `tokens.css`
- Update `@font-face` in our templates to load `400`, `600`, and `700` weights of `Muli`
- Update `@font-face` in our templates to load the `600` weight of `Source Serif Pro`
- Update existing CSS custom properties to use the new values
- Fix CSS as necessary in places where the new values cause a regression

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2477.pwa-venia.com : LH Performance Expected 0.85 Actual 0.55, LH Best Practices Expected 1 Actual 0.92
https://pr-2477.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.35, LH Best Practices Expected 1 Actual 0.92
https://pr-2477.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.49, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/venia-ui/lib/tokens.css
```diff
@@ -0,0 +1,136 @@
+/* global tokens */
+:global(:root) {
```
sirugh, line 2: @jimbo I saw in the JIRA ticket that the first task was to copy these from the styleguide to venia-ui. Was the wording intentional? This is an opportunity for the guide and venia-ui to get out of sync. Should we be using tokens as an export of styleguide rather than a copy?

packages/venia-concept/src/.storybook/config.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { configure, addDecorator } from '@storybook/react';
 import { Adapter } from '@magento/venia-drivers';
 import store from '../store';
-import '../index.css';
+import '../../../venia-ui/lib/index.css';
```
sirugh, line 5: We should use the alias/absolute path rather than relative.

```suggestion
import '@magento/venia-ui/lib/index.css';
```

Please update all imports of `venia-ui` in `venia-concept` to match this pattern.

packages/venia-ui/lib/components/ButtonGroup/button.css
```diff
@@ -1,37 +1,37 @@
 .root {
     composes: root from '../clickable.css';
     background: none white;
-    border: 1px solid rgb(var(--venia-grey-dark));
-    color: rgb(var(--venia-text));
+    border: 1px solid rgb(var(--venia-global-color-gray-dark));
+    color: rgb(var(--venia-global-color-text));
     font-size: 0.75rem;
```
sirugh, line 6: Should we also be replacing all uses of `font-size: X` with `--venia-global-fontSize-X`?

It seems like the tokens we changed are the old `--venia-FOO` tokens, which is a great start but there are a few tokens that we could also globally apply such as colors font size, weight and line height.

@jimbo what do you think?

packages/venia-ui/lib/tokens.css
jimbo: That's correct. They could be out of sync, but `venia-styleguide` is not intended to be a dependency of `venia-ui` in any way. It shouldn't export anything for consumption.

packages/venia-ui/lib/components/ButtonGroup/button.css
jimbo: Yes, we need to do that, but no, not as part of this story.

packages/venia-ui/lib/tokens.css
```diff
@@ -0,0 +1,133 @@
...
+    --venia-global-color-gray-800: 51, 63, 71;
+    --venia-global-color-gray-900: 23, 32, 38;
+    --venia-global-color-gray: var(--venia-global-color-gray-100);
+    --venia-global-color-gray-dark: var(--venia-global-color-gray-300);
+    --venia-global-color-gray-darker: var(--venia-global-color-gray-600);
+    --venia-global-color-green-400: 51, 171, 132;
+    --venia-global-color-green-500: 45, 157, 120;
+    --venia-global-color-green-600: 38, 142, 108;
+    --venia-global-color-green-700: 18, 128, 92;
+    --venia-global-color-orange: 241, 99, 33;
```
anthoula, line 32: The orange property is used only for the toast area, and have left this as is, since there's no option for that as of yet in the new style

packages/venia-ui/lib/tokens.css
```diff
@@ -0,0 +1,133 @@
...
+    --venia-global-color-red-400: 236, 91, 98;
+    --venia-global-color-red-500: 227, 72, 80;
+    --venia-global-color-red-600: 215, 55, 63;
+    --venia-global-color-red-700: 201, 37, 45;
+    --venia-global-color-teal: var(--venia-global-color-blue-400);
+    --venia-global-color-teal-dark: var(--venia-global-color-blue-600);
+    --venia-global-color-teal-light: var(--venia-global-color-gray-100);
```
anthoula, line 39: Teal colors now point to the new blue values. We have a stories forthcoming (PWA-622, PWA-623) that addresses the areas these are used: inputs and buttons. I've decided to leave the naming convention as is for now, as it's likely to be addressed there.

davemacaulay:
QA Approved 😄

PR #2481, opened 2020-06-10
[PWA-612]: Fixed shipping method loading error.

revanth0212:
## Description

Fixed a false negative in shipping method component.

The component has a check to show an error message is shipping methods array is empty, but the array can be empty if the query is still loading. That is a false negative.

Fixed it by adding the query loading state to the if check.

## Related Issue
Closes [PWA-612](https://jira.corp.magento.com/browse/PWA-612)

## Acceptance 
The component should not show any error message while loading.

### Verification Stakeholders
@supernova-at 
@jimbo 

### Verification Steps
1. Go to the checkout page with at least 1 item in cart.
2. Continue to step 2. Select a shipping method.
3. After the shipping method is set, refresh the page, there should not be any error message shown before the shipping method summary is shown.

Before:
image

After:
image

## Checklist


packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingRadios.js
```diff
@@ -10,11 +10,11 @@ const ERROR_MESSAGE =
     'Error loading shipping methods. Please ensure a shipping address is set and try again.';
 
 const ShippingRadios = props => {
-    const { disabled, shippingMethods } = props;
+    const { disabled, shippingMethods, isLoading } = props;
```
sirugh, line 13: This seems to be addressing a symptom of the issue which is that somehow the children are rendered even when the parent is not loading. 

It's not clear to me how this was able to render since the parent should be rendered `initializingContents` when initializing. This is determined in the talon:

```js
    const displayState = derivedSelectedShippingMethod
        ? displayStates.DONE
        : isLoadingShippingMethods ||
          (isSettingShippingMethod && isBackgroundAutoSelecting)
        ? displayStates.INITIALIZING
        : displayStates.EDITING;
```

At a glance it seems like `displayState === displayStates.EDITING` even when loading so maybe the calculation of the derived display state is incorrect somehow.

revanth0212: I have realized that `isLoadingShippingMethods` which is basically a different name for `loading` of the shipping methods query can be `undefined`, in which case the `or` condition fails and `displayState=displayStates.EDITING` is executed. Not sure why is it being set as something other than a boolean value. I'll have dig a bit more.

revanth0212: Since we have the fetch policy as `cache-and-network`, loading can actually be false at first and then true when the query is fetching from the network. In that process, when the loading is false, displayState will be editing and if the cached response for the shipping methods is empty, the shipping radios component will show the error.

All of this is a theory. But, I took your advice and removed loading because it felt like a patch rather than a fix, and when I tried with `skip` instead of `useEffect`, I was not able to reproduce the error. I tried multiple network speeds (fast 3g, slow 3g etc) and deleted the local apollo cache, still, I was not able to reproduce the error.

I guess I fixed it but I am not sure how. @sirugh help me out. How did it got fixed?

sirugh: > I guess I fixed it but I am not sure how. @sirugh help me out. How did it got fixed?

Andy's last comment on the JIRA ticket says he was unable to repro. Were you able to repro on `develop`?

revanth0212: Yeah i was able to reproduce in develop before i started making changes.

PR #2483, opened 2020-06-10
[PWA-413] [Cart Page Edit] Price Does not get updated after changing the color/size.

tjwiebell:
## Description

Steps -
1. Add a configurable product to Cart and go Cart page.
2. Go to Edit panel and switch to different color.

Expected - Price should get updated based on selection.
Actual - Price is static, user will get confused after seeing the update on Cart page.

Note - Same issue exists on Mini Cart Edit window as well but will be part of mini cart refactor?

## Related Issue

* [[PWA-413](https://jira.corp.magento.com/browse/PWA-413)] [Cart Page Edit] Price Does not get updated after changing the color/size.

### Verification Steps

⬆️  Use repro steps in description ⬆️ 

## Checklist


packages/peregrine/lib/talons/CartPage/ProductListing/EditModal/useEditModal.js
```diff
@@ -1,11 +1,17 @@
+import { useState } from 'react';
+
 import { useAppContext } from '../../../../context/app';
 
 export const useEditModal = () => {
     const [{ drawer }, { closeDrawer }] = useAppContext();
     const isOpen = drawer === 'product.edit';
 
+    const [variantPrice, setVariantPrice] = useState();
```
sirugh, line 9: This is `undefined` initially. What happens if the item in the cart is a configurable item? I know it falls back to `price` but could that be wrong? Perhaps we need to do the calculation here.

packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/editModal.js
```diff
@@ -11,16 +11,20 @@ import ProductForm from './productForm';
 
 const EditModal = props => {
     const { item, setIsCartUpdating } = props;
-    const talonProps = useEditModal();
-    const { handleClose, isOpen } = talonProps;
+    const talonProps = useEditModal({ item });
```
sirugh, line 14: I notice you pass the item here - is that to do the calculation of the variant price on initial render? You aren't using it anymore in the talon though.

packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/editModal.js
```diff
@@ -11,16 +11,20 @@ import ProductForm from './productForm';
...
+    const { handleClose, isOpen, setVariantPrice, variantPrice } = talonProps;
 
     const classes = mergeClasses(defaultClasses, props.classes);
     const rootClass = isOpen ? classes.root_open : classes.root;
 
     const bodyComponent = item ? (
         <div className={classes.body} key={item.id}>
-            <ProductDetail item={item} />
-            <ProductForm item={item} setIsCartUpdating={setIsCartUpdating} />
+            <ProductDetail item={item} variantPrice={variantPrice} />
```
sirugh, line 22: `variantPrice` is undefined if you open the edit modal and don't click on anything.

packages/peregrine/lib/talons/CartPage/ProductListing/EditModal/useEditModal.js
tjwiebell: The first render for this component will have `item = undefined`, so it wouldn't be that easy. I could add an effect, but I didn't really see the purpose when `ProductDetail` already had the data it needed to render price. Instead of having to handle a bunch of different states for opening, then resetting on close, I passed this responsibility down to `ProductForm`.

BTW - you can only edit Configurable product right now (assuming you meant _isn't a configurable item_)

packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/editModal.js
tjwiebell: I attempted your suggestion to initialize in the talon at first, but went the simple route. Will remove this prop.

packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/editModal.js
tjwiebell: Yes. We render `cartItem` price until a new variant is selected.

packages/peregrine/lib/talons/CartPage/ProductListing/EditModal/useEditModal.js
sirugh: Ah yes, I mistyped.

What I meant was that when you first click on a product to edit the value for variant was "undefined". It seemed to render correctly regardless, but maybe that was by coincidence?

packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/editModal.js
sirugh: So if the item in the cart is a variant of the default, `cartItem.price` will be the selected variant price not the default price?

packages/peregrine/lib/talons/CartPage/ProductListing/EditModal/useEditModal.js
tjwiebell: By [design](https://github.com/magento/pwa-studio/pull/2483/files#diff-75bfab0706e8e99bd7c97dbd3548145dR19). I'm willing to make this change, but did want to do some validation. If I add a product to my cart that is on sale, but don't checkout until a day later when that product is no longer on sale, do I still get that product at the sale price? I don't know the answer to that, but that would be a reason to use cart price instead of variant price. There might be other use cases as well where cart price is the accurate value, and is what we want to render.

sirugh: If variant price reflects product specific sales I think it's fine to only use that. Would be odd to reflect cart sales such as BOGO or <$X discounts in an edit view.

jimbo: > If I add a product to my cart that is on sale, but don't checkout until a day later when that product is no longer on sale, do I still get that product at the sale price? I don't know the answer to that, but that would be a reason to use cart price instead of variant price.

We can assume that if a sale price expires before a shopper completes a purchase, the sale price is forfeit. If there are special cases where that's not true, someone would have to handle those cases as exceptions.

dpatil-magento:
Went ahead and QA'ed this, looks good.

PR #2486, opened 2020-06-11
Use effect to place order after details have been fetched.

sirugh:
## Description

The invoker returned as `const [getOrderDetails] = useLazyQuery` is not actually a promise as we previously believed. We want to only place the order once the order details have been fetched. The code was causing a race condition because both operations were happening at the same time.

## Related Issue

Closes [PWA-677](https://jira.corp.magento.com/browse/PWA-677).

### Verification Steps

1. Place an order.
2. Make sure the page doesn't flicker between cart, loading, and the receipt page.
2. In the network tab, ensure that the "place order" mutation was not made until the "getOrderDetails" query completes.

[Image from Gyazo](https://gyazo.com/8e8579796629c52ff69147f883d98ae2)

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
+++ b/packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
@@ -148,7 +148,6 @@ const getTalonProps = props => {
  */
 
 beforeEach(() => {
-    jest.clearAllMocks();
     useQuery.mockImplementation(query => {
         if (query === getCheckoutDetailsQuery) {
             return getCheckoutDetailsQueryResult();
--- a/packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
+++ b/packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
@@ -112,7 +112,8 @@ const placeOrderMutationResult = jest.fn().mockReturnValue([
     {
         error: null,
         loading: false,
-        data: null
+        data: null,
+        called: false
     }
 ]);
 
@@ -146,7 +147,8 @@ const getTalonProps = props => {
  * beforeAll
  */
 
-beforeAll(() => {
+beforeEach(() => {
+    jest.clearAllMocks();
     useQuery.mockImplementation(query => {
         if (query === getCheckoutDetailsQuery) {
             return getCheckoutDetailsQueryResult();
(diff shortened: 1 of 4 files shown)


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2486.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92
https://pr-2486.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2486.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.48, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -94,17 +94,16 @@ const CheckoutPage = props => {
     const isMobile = windowSize.innerWidth <= 960;
 
     let checkoutContent;
-    if (isLoading) {
-        return fullPageLoadingIndicator;
-    }
 
-    if (!placeOrderLoading && !hasError && orderDetailsData) {
+    if (orderNumber) {
```
jimbo, line 98: Makes sense. 👍

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -142,43 +143,54 @@ export const useCheckoutPage = props => {
...
-            );
-            setReviewOrderButtonClicked(false);
-            setCheckoutStep(CHECKOUT_STEP.PAYMENT);
-            // TODO: Delete nonce? The nonce might be expired and why the order
-            // failed. If we delete it the payment info section will render as
-            // if it was not filled, thus prompting the user to enter new info.
+        // Fetch order details and then use an effect to actually place the
+        // order. If/when Apollo returns promises for invokers from useLazyQuery
+        // we can just await this function and then perform the rest of order
+        // placement.
```
jimbo, line 149: Yeah, the [Apollo docs](https://www.apollographql.com/docs/react/api/react-hooks/#result-1) say the following about the invoker:

> Function that can be triggered to execute the suspended query. After being called, `useLazyQuery` behaves just like `useQuery`.

So I think they do intend for us to use `called`, `data`, and `loading` as returned by the hook, since they're updated by the hook itself as part of its internal lifecycle. And if we want to use the hook's return values, we have to do it as an effect, just as you've done here.

revanth0212:
Other than test failures, code looks fine.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -142,43 +143,54 @@ export const useCheckoutPage = props => {
...
+                    'An error occurred during when placing the order',
+                    err
+                );
+                setReviewOrderButtonClicked(false);
+                setCheckoutStep(CHECKOUT_STEP.PAYMENT);
+            }
+        }
+
+        if (orderDetailsData && !placeOrderCalled) {
+            placeOrderAndCleanup();
```
revanth0212, line 184: Should this not be `await placeOrderAndCleanup()`?

jimbo: > Should this not be `await placeOrderAndCleanup()`?

The function provided to `useEffect` isn't an `async` function, so we couldn't `await` this call.

dpatil-magento:
QA Pass, good to merge after test changes and approval.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: Nope - this is the pattern for writing effects that contain async code. You can't have a top level `await` because the callback passed to `useEffect` cannot be asynchronous itself.

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -146,7 +147,8 @@ const getTalonProps = props => {
  * beforeAll
  */
 
-beforeAll(() => {
+beforeEach(() => {
+    jest.clearAllMocks();
```
jimbo, line 151: This shouldn't be necessary, since [we do this at the config level](https://github.com/magento/pwa-studio/blob/develop/jest.config.js#L245).

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -243,50 +245,49 @@ test('returned error prop should be error from place order mutation', () => {
...
-        await talonProps.handlePlaceOrder();
+    await talonProps.handlePlaceOrder();
 
-        expect(clearCartDataFromCache).toHaveBeenCalledWith(client);
+    expect(placeOrder).toHaveBeenCalledWith({
+        variables: { cartId: '123' }
     });
+    expect(removeCart).toHaveBeenCalled();
+    console.log(clearCartDataFromCache);
+    expect(clearCartDataFromCache).toHaveBeenCalled(); //toHaveBeenCalledWith(client);
```
jimbo, line 289: Is this comment relevant?

sirugh: I've cleaned this up - must have forgotten to push.

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
sirugh, line 151:
```suggestion
```

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
```diff
@@ -243,50 +244,49 @@ test('returned error prop should be error from place order mutation', () => {
...
 
-        await talonProps.handlePlaceOrder();
+    await talonProps.handlePlaceOrder();
 
-        expect(clearCartDataFromCache).toHaveBeenCalledWith(client);
+    expect(placeOrder).toHaveBeenCalledWith({
+        variables: { cartId: '123' }
     });
+    expect(removeCart).toHaveBeenCalled();
+    console.log(clearCartDataFromCache);
```
sirugh, line 287:
```suggestion
```

packages/peregrine/lib/talons/CheckoutPage/__tests__/useCheckoutPage.spec.js
sirugh:
```suggestion
    expect(clearCartDataFromCache).toHaveBeenCalled();
```

PR #2495, opened 2020-06-17
[PWA-336] Cart/Checkout Input Errors UI cases

tjwiebell:
## Description

After Cart/Checkout is done we should make a general pass over the error states to make sure we represent them correctly according to UX desires and the style guide.

Cases to consider:
 * User errors
 * System errors

h5. Open Questions:
 * How should we handle form fields that are out of view of the shopper when the error occurs?
 * Are we going to have both form and field level reporting?
 * Do we have the most common error scenarios / use-cases documented?

Summary of GraphQL Operations that may have error state: https://wiki.corp.magento.com/pages/viewpage.action?pageId=144314396

## Related Issue

* [[PWA-336](https://jira.corp.magento.com/browse/PWA-336)] Cart/Checkout Input Errors UI cases

### Verification Steps

This guy is going to be a little difficult to test, it's quite tricky to simulate errors. You might be able to go offline and use network errors (we've discovered this actually doesn't work in most places), but modifying code to break the queries or mutations is what I did. I'll outline the components touched, and the verification steps are all the same; simulate an error somehow and verify output. I have also included screenshots in the HLD for UX, to make visual review easier.
- CartPage > Product Edit > Loading Options (now renders error)
- CartPage > Remove Item (throws Toast)
- CartPage > Product Edit > Update (renders quantity and option errors now and doesn't close form)
- CartPage > Remove Coupon (now renders error)
- CartPage > Estimate Shipping > Set Address (now renders error)
- CheckoutPage > Select Shipping Method (now renders error and doesn't close if in modal)
- CheckoutPage > Address Book > Apply Address (throws Toast, doesn't close)
- CheckoutPage > Auth/Guest Address > Add/Update Address (renders error, auto scrolls to form top, doesn't close anymore)

See [HLD](https://wiki.corp.magento.com/pages/viewpage.action?pageId=144314396) for tons of screenshots.

## Checklist


packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -48,21 +52,23 @@ export const useProduct = props => {
...
-            return updateError.graphQLErrors
+            return errorTarget.graphQLErrors
                 .map(({ message }) => message)
                 .join(', ');
         }
 
         // A non-GraphQL error occurred.
-        return updateError.message;
-    }, [updateError]);
+        return errorTarget.message;
```
tjwiebell, line 70: I kept this pattern around where I was re-factoring and it existed, but noticed on more recent code we don't use it anymore. Can this be removed, and we only need to care about GraphQL errors?

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2495.pwa-venia.com : LH Performance Expected 0.85 Actual 0.57, LH Best Practices Expected 1 Actual 0.92
https://pr-2495.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2495.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.49, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
sirugh: How would we get here, network error?

tjwiebell: After further review, you are correct! This is needed.

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/AddressForm/useGuestForm.js
```diff
@@ -11,12 +11,36 @@ export const useGuestForm = props => {
...
+                ? graphQLErrors.map(({ message }) => message).join(', ')
+                : message;
+        }
+
+        return errorMessage;
+    }, [error]);
+
+    useEffect(() => {
+        if (error) {
+            errorRef.current.scrollIntoView({
```
sirugh, line 37: Does this sort of logic belong in the talon? It's presentational which was my concern.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
```diff
@@ -32,9 +32,24 @@ export const useShippingForm = props => {
 
     const [
         setShippingAddress,
-        { called: setShippingAddressCalled, loading: isSetShippingLoading }
+        {
+            called: setShippingAddressCalled,
+            error: setShippingAddressError,
```
jimbo, line 37: It's a bit counterintuitive to see `setShippingAddressCalled` and `setShippingAddressError` used as object names, since it's our most common format for naming functions.

Consider `calledSetShippingAddress` or `errorSettingShippingAddress`, I suppose?

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingForm.css
```diff
@@ -27,6 +27,10 @@
     padding: 0.75rem 1rem;
 }
 
+.error {
+    color: rgb(var(--venia-error));
```
jimbo, line 31: I'm not sure this is a property anymore.

- See [`tokens.css`](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/tokens.css#L44) for new tokens
- If you replace this one, make sure to replace all of the others as well

```suggestion
    color: rgb(var(--venia-global-color-error));
```

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingForm.js
```diff
@@ -30,16 +30,22 @@ const ShippingForm = props => {
...
+        errorMessage,
         handleOnSubmit,
         handleZipChange,
         isSetShippingLoading
     } = talonProps;
 
     const classes = mergeClasses(defaultClasses, props.classes);
 
+    const errorElement = errorMessage ? (
+        <span className={classes.error}>{errorMessage}</span>
```
jimbo, line 42: I might have liked to see a `FormError` component for form-level errors, since I imagine the presentation will be nearly identical for all of them. Such a component could render `null` when `props.message` was empty, too, cutting down on a lot of the boilerplate.

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/AddressForm/useCustomerForm.js
```diff
@@ -39,6 +47,30 @@ export const useCustomerForm = props => {
         }
     }, [getCustomerError]);
 
+    useEffect(() => {
+        if (createCustomerAddressError || updateCustomerAddressError) {
+            errorRef.current.scrollIntoView({
```
jimbo, line 52: I think `scrollIntoView()` is the right choice here, but I don't think it belongs in the talon.

The only other place we use this, [`PaymentsFormItems`](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/Checkout/paymentsFormItems.js#L55), has a talon of its own as well, but the component is the one to call `scrollIntoView` because this kind of effect is really just presentational. There's very little logic we put in the component rather than the talon, but I think this qualifies.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
sirugh: This pattern is used all over the place. If we want to switch it up let's make sure everyone is aware.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingForm.css
tjwiebell: @jimbo - that token doesn't seem to exist. Would you like me to add it, or pick one of the `--venia-global-color-red-xxx` values?

tjwiebell: Strike that, we cut the release branch before design tokens landed, so the new tokens are not available. Going to leave this as-is, and we'll resolve when we merge release back into `develop`.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
```diff
@@ -32,9 +32,24 @@ export const useShippingForm = props => {
...
 
+    let derivedError;
+    if (setShippingAddressError) {
+        if (setShippingAddressError.graphQLErrors) {
+            derivedError = setShippingAddressError.graphQLErrors
+                .map(({ message }) => message)
+                .join(', ');
+        } else {
+            derivedError = setShippingAddressError.message;
+        }
```
sirugh, line 50: I smell a utility function!

packages/peregrine/lib/talons/CartPage/ProductListing/EditModal/useProductForm.js
```diff
@@ -18,26 +18,44 @@ export const useProductForm = props => {
...
+        }
     ] = useMutation(updateConfigurableOptionsMutation);
 
     const isSaving =
         (updateQuantityCalled && updateQuantityLoading) ||
         (updateConfigurableCalled && updateConfigurableLoading);
 
-    useEffect(() => {
-        if (formApi) {
-            formApi.setValue('quantity', cartItem.quantity);
```
sirugh, line 38: What was this doing and why is it gone now?

packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
tjwiebell: Renamed in 57ab1a7.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingForm.js
tjwiebell: Added `FormError` component in 57ab1a7, and changed all usages in this PR to use it (where it didn't dramatically alter the UX).

I thought we had discussed deferring a dedicated component to a followup Story, but I've gone ahead and implemented one in this scope.

packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/AddressForm/useCustomerForm.js
tjwiebell: Moved this logic to `FormError` and put scroll and ref logic in component instead of talon (57ab1a7).

packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/useShippingForm.js
tjwiebell: We are probably not going to render Apollo errors going forward, since they can't be trusted to be human readable. I have moved this logic to a new `FormError` component which should reduce the boiler plate though, but you still gotta handroll this logic if you don't like the presentation of that component.

packages/peregrine/lib/talons/CartPage/ProductListing/EditModal/useProductForm.js
tjwiebell: The intention was to make sure quantity was updated between renders, but two things made it obsolete:

1. I started un-mounting between renders, so this wasn't needed
2. This was fixed in the Quantity component in the same way, so this was redundant anyway (even if we didn't unmount between renders)

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingForm.css
jimbo: Ah, my mistake. I assumed this was targeting `develop` but it's actually targeting the release branch. Carry on. 👍

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingForm.js
jimbo: Thanks. I apologize if we backtracked on an agreement. It's not the duplication of code that bothered me; I just saw that we were adding new classnames to a lot of form components as a result of the new error elements, and thought it'd be nice to avoid that churn. Glad to have a component now.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/shippingForm.js
```diff
@@ -40,6 +42,7 @@ const ShippingForm = props => {
     return (
         <Fragment>
             <h3 className={classes.formTitle}>Destination</h3>
+            <FormError errors={formErrors} />
```
jimbo, line 45: You don't have to do this, but consider just rendering the `FormError` element as the first child of the `Form` element, rather than wrapping these forms in fragments. Might help if the form wants to have the form error participate in its layout.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
```diff
@@ -87,26 +91,28 @@ export const useCouponCode = props => {
...
             // which we don't want to show to an end user.
             // Build up the error message manually without the prepended text.
-            applyErrorMessage = applyError.graphQLErrors
+            derivedErrorMessage = errorTarget.graphQLErrors
                 .map(({ message }) => message)
                 .join(', ');
         } else {
             // A non-GraphQL error occurred.
-            applyErrorMessage - applyError.message;
+            derivedErrorMessage = errorTarget.message;
```
sirugh, line 108: Do we need to handle network errors? They'd be on the target as `errorTarget.networkError`.

sirugh: Discussed offline - I am for moving this sort of logic into a utility that is used by `useFormError` and by any other components that need to turn apollo errors into a string.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/__tests__/shippingMethods.spec.js
```diff
@@ -23,7 +23,7 @@ jest.mock('../shippingForm', () => 'ShippingForm');
 jest.mock('../shippingRadios', () => 'ShippingRadios');
 
 test('renders description and confirm link w/o shipping address set', () => {
-    useLazyQuery.mockReturnValue([
+    useQuery.mockReturnValue([
```
supernova-at, line 26: I'm surprised this is working - `useQuery` just returns an object, not an array like `useLazyQuery` does.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
```diff
@@ -87,26 +91,28 @@ export const useCouponCode = props => {
...
 
-    let applyErrorMessage;
+    let derivedErrorMessage;
+
+    if (applyError || removeCouponError) {
+        const errorTarget = applyError || removeCouponError;
 
-    if (applyError) {
-        if (applyError.graphQLErrors) {
+        if (errorTarget.graphQLErrors) {
```
sirugh, line 99: OOPS - as we just discovered, `graphQLErrors` is an array. All these checks will need to be updated to `length`

packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
supernova-at: I was going to suggest that too, 💯  agree

packages/venia-ui/lib/components/CartPage/PriceAdjustments/ShippingMethods/__tests__/shippingMethods.spec.js
tjwiebell: Very odd. I haven't full covered this with tests yet, so I'll be sure to investigate this and update the mock to accurately reflect the implementation of useQuery.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
tjwiebell: Great catch! Added this check to `FormError`, and we'll address the other areas once we create a utility.

dpatil-magento:
QA Approved.

PR #2498, opened 2020-06-18
Scroll top when payment information is done

sirugh:
## Description

It was unclear what happens when a user clicks "Review Order". The page stays locked to its scroll position but the dom changes when the items component renders and the payment done card is displayed. The desired effect is to scroll the page once the payment information section is done.

## Related Issue

Closes [PWA-697](https://jira.corp.magento.com/browse/PWA-697).

### Verification Steps

1. Go to checkout and proceed through steps up to payment info.
2. Click Review Order after filling payment info.
3. See that page scrolls to top.

[Image from Gyazo](https://gyazo.com/a179b0f574bb2c6864c807241b62fdd3)

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2498.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92
https://pr-2498.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.37, LH Best Practices Expected 1 Actual 0.92
https://pr-2498.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.49, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

tjwiebell:
I'm sure there's a Slack conversation somewhere (if there is, add it to Description), but we're sure this is the behavior we want? None of the information at the top of the page changed, so I would expect a scroll to what did, which is the list of "Items in your order".

sirugh:
Per @soumya-ashok [here](https://magento.slack.com/archives/G8N21AQTV/p1592495608430900?thread_ts=1592495082.427900&cid=G8N21AQTV):

> My instinct would be to say since the action is to review, we should deposit them at the top of the page on mobile once the button is clicked so they can review the page and then hit the "place order" button at the bottom.

Maybe we want a clarification between mobile/desktop UX. Personally I think this is better than the existing UX for now.

tjwiebell:
> Maybe we want a clarification between mobile/desktop UX. Personally I think this is better than the existing UX for now.

I saw another followup Story about step transitions, so we can re-visit in that scope for desktop/mobile; this appears to be what UX wants, all good 👍

soumya-ashok:
Works as expected, UX approved.

PR #2504, opened 2020-06-23
Respect URL Redirects

### LINKED ISSUE #2502: [feature]: Make the pwa handle redirect urls
As a merchant I'd like to be able to make use of magento's redirect urls, right now those are not being handled by PWA Studio

I'd like to see a solution where urls get redirected to the right page when there is a redirect set in magento.

**Additional context**
I will submit a PR, so this issue is just for reference.

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

Jordaneisenburger:
## Description
This will handle redirects when set in Magento

## Related Issue

Closes #2502 

### Verification Steps
1. Go to a Magento backend and create a redirect for a product
2. Go to the old url and notice you'll be redirected

## Checklist


Jordaneisenburger:
**note:** Closed the previous PR because it wouldn't update.

Jordaneisenburger:
If the solution provided is the prefered way to go I'll add the tests.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2504.pwa-venia.com : LH Performance Expected 0.85 Actual 0.54, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 79
https://pr-2504.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.38, LH Best Practices Expected 1 Actual 0.92
https://pr-2504.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -14,9 +14,19 @@ const MESSAGES = new Map()
...
+        component: RootComponent,
+        id,
+        isLoading,
+        routeError,
+        redirectCode,
+        relativeUrl
+    } = talonProps;
 
-    if (isLoading) {
+    if (redirectCode === 301 || redirectCode === 302) {
```
supernova-at, line 26: According to the [urlResolver query docs](https://devdocs.magento.com/guides/v2.4/graphql/queries/url-resolver.html?itm_source=devdocs&itm_medium=search_page&itm_campaign=federated_search&itm_term=urlresolver#output-attributes), this value will only ever be `0`, `301`, or `302` 👍

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -14,9 +14,19 @@ const MESSAGES = new Map()
...
+        window.location = relativeUrl;
```
supernova-at, line 27: I'm wondering if this logic shouldn't be in `useMagentoRoute`. In other words, have it return the details for the component that matches `relativeUrl` if the redirect condition is satisfied.

I'm thinking it would save us a redirect on the front-end (page load time) in the case of redirects.

supernova-at: Use `history.replace` here instead of directly assigning to `window.location` for consistency

supernova-at: Consider putting this in a `useEffect` in the talon

supernova-at: Resolved by https://github.com/magento/pwa-studio/pull/2504/commits/c56798d3a174fd0d2ffeaada12f263fd3c827981.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -1,61 +1,98 @@
...
 import getRouteComponent from './getRouteComponent';
 
+const CODE_PERMANENT_REDIRECT = 301;
+const CODE_TEMPORARY_REDIRECT = 302;
+const REDIRECT_CODES = [CODE_PERMANENT_REDIRECT, CODE_TEMPORARY_REDIRECT];
+
+const DEMO = new Map().set('/penelope-peasant-blouse.html', {
+    isRedirect: true,
+    relativeUrl: '/susanna-draped-tank.html'
+});
```
tjwiebell, line 15:
```suggestion
```

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -1,61 +1,98 @@
...
-        const isKnown = componentMap.has(pathname);
+        // Avoid setting state if unmounted.
+        if (!isMountedRef.current) {
+            return;
+        }
 
-        // ask again following a prior failure
-        const isNotFound = isKnown && componentMap.get(pathname).id === -1;
-        const shouldReload = isNotFound && navigator.onLine;
+        const shouldFetch = data => {
```
tjwiebell, line 42: There's a lot going on in this effect, and your changes have definitely improved the readability. I would consider more simplification, and separating responsibilities. `shouldFetch` can probably be derived from state outside of the effect.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -1,61 +1,98 @@
...
+        };
 
+        // Get the data for this pathname from our Map in local state.
+        const routeData = componentMap.get(pathname);
+
+        // If we have data but it's a redirect, perform the redirect.
+        if (routeData && routeData.isRedirect) {
+            history.replace(routeData.relativeUrl);
+            return;
+        }
```
tjwiebell, line 61: This should probably go in its own effect.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -1,61 +1,98 @@
...
+            // Ask again following a prior failure.
+            if (data.isNotFound && navigator.onLine) {
+                return true;
+            }
+
+            return false;
+        };
 
+        // Get the data for this pathname from our Map in local state.
+        const routeData = componentMap.get(pathname);
```
tjwiebell, line 55: Good idea to memoize and derive this for the effects to use.

packages/peregrine/lib/talons/MagentoRoute/getRouteComponent.js
```diff
@@ -25,7 +25,7 @@ const getRouteComponent = async (apiBase, pathname) => {
             throw new Error('404');
         }
 
-        const { type, id } = resolvedRoute;
+        const { type, id, redirectCode, relative_url } = resolvedRoute;
```
tjwiebell, line 28: Very minor, but I have a preference to alias on destructure to be consistent with variable naming. I don't think I would write `const relative_url = resolvedRoute.relative_url` if we were missing the sugar.

```suggestion
        const { type, id, redirectCode, relative_url: relativeUrl } = resolvedRoute;
```

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -1,61 +1,98 @@
...
 
-                    // then add the pathname and result to local state
+                    // Update our Map in local state for this path.
                     setComponentMap(prevMap => {
                         const nextMap = new Map(prevMap);
+
                         const nextValue = routeError
                             ? { hasError: true, routeError }
+                            : id === -1
+                            ? { isNotFound: true }
```
tjwiebell, line 83: Was this fixing a bug that's not jumping out? I assume it related to the offline retry, but kinda lost track of that in the refactor.

supernova-at: Not fixing a bug, just improving readability for the logic after the lookup.  Basically all of the "inside knowledge" of what `-1` means, or "is this a redirect?" is here, so that when we retrieve this data we can just look at bools like `hasError`, `isNotFound`, `isRedirect`, etc.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/2504/commits/5be73656c08286410d34942110d9546cf67f28b0.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/2504/commits/5be73656c08286410d34942110d9546cf67f28b0.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
supernova-at: Resolved in https://github.com/magento/pwa-studio/pull/2504/commits/5be73656c08286410d34942110d9546cf67f28b0.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
supernova-at: I didn't end up memoizing this but I did pull it out of the effect 👍

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
supernova-at: Made this a bit easier to read in https://github.com/magento/pwa-studio/pull/2504/commits/5be73656c08286410d34942110d9546cf67f28b0.

packages/peregrine/lib/talons/MagentoRoute/getRouteComponent.js
supernova-at: I didn't end up changing this one not because I disagree necessarily, but because the underscores in variable names isn't how we generally name variables in our codebase. Basically, we're getting a name from GQL and turning it into something our codebase is going to use, and this feels like the best place to make that transition.

dpatil-magento:
Note - Currently Product page shows stale data (redirect not works if browser has original url-key loaded already) and would be handled separately.

PR #2505, opened 2020-06-23
PWA-602: Add to Shopping Bag from PDP

sirugh:
## Description

Adding an item to the shopping bag should add the item and update the cart trigger.
If there is a problem, it should show a message in the form. If you're on `develop` mode then there will be an error logged to console.

## Related Issue

Closes [PWA-602](https://jira.corp.magento.com/browse/PWA-602).

### Verification Steps

1. Go to the PDP and add an item. It should work.
2. Add an item that is out of stock. It should show an error about quantity.
3. Add an item but trigger an unknown error such as an invalid user token. It should render a form level error.

## Screenshots / Screen Captures (if appropriate)

**Successful Add**
[Image from Gyazo](https://gyazo.com/ed2c91346a847ffba5f1d2aa622eb0fc)

**Quantity Error**
[Image from Gyazo](https://gyazo.com/12841847b303aa93cb528bc1a2f419d8)

**Unknown Error**
[Image from Gyazo](https://gyazo.com/bbb192882b1824d05d1eb7fbfa6b4b60)

## Checklist

- [ ] I have added tests to cover my changes, if necessary.
* I have updated the documentation accordingly, if necessary.

PR #2506, opened 2020-06-23
Mini Cart V2 / Shopping bag product listing

revanth0212:
## Description

Added product listing component in the new Mini Cart component.

**Should be merged ahead of #2509 and #2511** 

## Related Issue
Closes [PWA-605](https://jira.corp.magento.com/browse/PWA-605)

### Verification Stakeholders
@jimbo 
@soumya-ashok 
@schensley 

### Verification Steps

1. run `yarn watch:venia`.
2. Add an item to the cart and click the new cart button.
3. We should be able to see the product listing.

image

## Checklist

* **Should be merged ahead of #2509 and #2511**


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2506.pwa-venia.com : LH Performance Expected 0.85 Actual 0.54, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 79
https://pr-2506.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2506.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.42, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/lib/components/ShoppingBag/shoppingBag.css
```diff
@@ -42,10 +42,8 @@
     gap: 1rem;
     /* The header and footer heights may change based on PWA-116 and PWA-601. */
     grid-template-rows: 2rem 1fr minmax(2rem, auto);
-    /* These heights may change based on PWA-605 and should be based on 1.5 and 2.5 visible items. */
-    height: 100%;
-    min-height: 31rem;
-    max-height: 45rem;
+    min-height: 10rem;
```
revanth0212, line 45: I have calculated the height for each item in the shopping bag and for 2.5 items it is 32rem. Hence the `max-height` is 32rem. I am not sure if that is the right way to handle this, I am open to suggestions.

packages/venia-ui/lib/components/ShoppingBag/shoppingBag.js
```diff
@@ -5,8 +5,16 @@ import { useScrollLock } from '@magento/peregrine';
...
 import { mergeClasses } from '@magento/venia-ui/lib/classify';
 
+import ProductListing from './ProductListing';
+
+import ShoppingBadOperations from './shoppingBag.gql';
+
 import defaultClasses from './shoppingBag.css';
 
+const Error = () => {
+    return <div>TBD</div>;
```
revanth0212, line 15: Do we wanna have an error message component/text in case of any errors while performing the remove mutation? Ill delete it if not needed.

packages/venia-ui/lib/components/ShoppingBag/shoppingBag.js
```diff
@@ -21,23 +29,41 @@ const ShoppingBag = props => {
...
     return (
         <aside className={rootClass}>
             {/* The Mask. */}
             <button className={classes.mask} onClick={onDismiss} type="reset" />
             {/* The Contents. */}
             <div className={classes.contents}>
-                <div className={classes.header}>Header TBD</div>
+                <div
+                    className={classes.header}
+                >{`Header TBD total quantity: ${totalQuantity}`}</div>
```
revanth0212, line 60: I have added `totalQuantity` to the GQL query as well. It is not needed in this PR but helpful for the tickets in the shopping bag epic.

packages/venia-ui/lib/components/ShoppingBag/ProductListing/productListing.js
```diff
@@ -0,0 +1,32 @@
...
+                    key={item.id}
+                    {...item}
+                    handleRemoveItem={handleRemoveItem}
+                />
+            ));
+        }
+    }, [listings, loading, handleRemoveItem]);
+
+    if (loading) {
+        return <LoadingIndicator>{`Fetching Items in Cart`}</LoadingIndicator>;
```
revanth0212, line 26: Should we show a loading indicator for product listing or for the whole shopping bag in case of a loading state?

revanth0212:
> When a user adds an item to the cart, the shopping bag is not updated immediately. It is getting updated after the page refresh.
> 
> I have checked and the add item to cart mutation does fetch new list of items in the cart and that should have updated the shopping bag. I am assuming I am missing something in the query or in the `useShoppingbag` talon where I am using the query.
> 
> Any idea why is that the case?

@sirugh I see you a [PR](https://github.com/magento/pwa-studio/pull/2505) for the shopping bag component. Does that handle the issue I am talking about?

sirugh:
> When a user adds an item to the cart, the shopping bag is not updated immediately. It is getting updated after the page refresh.

Consider the UX requirements when deciding how/when to fetch data. If a component is permanently rendered on the page, such as the cart trigger/icon, then other components must refetch the data necessary. In the case of the add to cart calls, the mutations fetch `total_quantity` which is what the trigger is listening for, and so it updates "automagically".

So my question is this - is the shopping bag permanently rendered and just hidden? Or is it mounted once a user clicks the trigger icon? If the latter then a simple `useQuery` call should work. If it's permanently mounted then all data necessary to render the components must be fetched by any mutations made on the `cart` object. If that's the case you'll want to expose mini cart fragments and then the add action will need to fetch them in the mutation response.

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
```diff
@@ -0,0 +1,67 @@
...
+
+export const useMiniCart = props => {
+    const { queries, mutations } = props;
+    const { shoppingBagQuery } = queries;
+    const { removeItemMutation } = mutations;
+
+    const [{ cartId }] = useCartContext();
+
+    const {
+        data: shoppingBadData,
```
sirugh, line 14: Bad data! Go to your room!

revanth0212: Classic.. thanks for the find dude 🤣

revanth0212:
> So my question is this - is the shopping bag permanently rendered and just hidden?

Yes, you are right, it is permanently rendered and hidden when not needed.

> If it's permanently mounted then all data necessary to render the components must be fetched by any mutations made on the `cart` object. If that's the case you'll want to expose mini cart fragments and then the add action will need to fetch them in the mutation response.

The add mutation already has fragments to fetch the list of cart items like I mentioned in the original comment. Since the mutation has the cart listing fragment mentioned, it is expected that the shopping bag will update on adding a new item to the cart.

Ill check if I am missing something.

packages/venia-ui/lib/components/MiniCart/miniCart.js
```diff
@@ -2,10 +2,19 @@ import React from 'react';
 import { bool, shape, string } from 'prop-types';
 
 import { useScrollLock } from '@magento/peregrine';
+import { useMiniCart } from '@magento/peregrine/lib/talons/MiniCart/useMiniCart';
 import { mergeClasses } from '@magento/venia-ui/lib/classify';
 
+import ProductListing from './ProductListing';
+
+import ShoppingBadOperations from './miniCart.gql';
```
jimbo, line 10: How bad are we talking about here? 😂

packages/venia-ui/lib/components/MiniCart/ProductListing/productListing.js
```diff
@@ -0,0 +1,32 @@
+import React, { useMemo } from 'react';
+
+import LoadingIndicator from '../../LoadingIndicator';
+import Item from './item';
+import { mergeClasses } from '../../../classify';
+
+import defaultClasses from './productListing.css';
+
+const ProductListing = props => {
+    const { listings, loading, handleRemoveItem, classes: propClasses } = props;
```
jimbo, line 10: I think we should aim to use the same terminology as before, rather than introducing `listing` as a new term for the same thing.

This component should be `ProductList`, and `props.listings` should really just be `props.items`.

packages/venia-ui/lib/components/MiniCart/miniCart.css
```diff
@@ -43,9 +43,8 @@
     /* The header and footer heights may change based on PWA-116 and PWA-601. */
     grid-template-rows: 2rem 1fr minmax(2rem, auto);
     /* These heights may change based on PWA-605 and should be based on 1.5 and 2.5 visible items. */
-    height: 100%;
-    min-height: 31rem;
-    max-height: 45rem;
+    min-height: 10rem;
+    max-height: 32rem;
```
jimbo, line 47: The maximum looks right to me. Does the minimum have to be this low? When the component enters and exits the loading state, it shrinks and grows by quite a bit.

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
```diff
@@ -0,0 +1,67 @@
...
+
+export const useMiniCart = props => {
+    const { queries, mutations } = props;
+    const { shoppingBagQuery } = queries;
+    const { removeItemMutation } = mutations;
+
+    const [{ cartId }] = useCartContext();
+
+    const {
+        data: shoppingBagData,
```
jimbo, line 14: The "shopping bag" terminology was nice while it lasted, but we've abandoned it. Probably want to update these names.

packages/venia-ui/lib/components/MiniCart/ProductListing/item.css
```diff
@@ -0,0 +1,45 @@
...
+    grid-row: 1 / span 1;
+    margin-bottom: 0.5rem;
+}
+
+.options {
+    display: grid;
+    gap: 0.5rem;
+    grid-column: 2 / span 1;
+    grid-row: 2 / span 2;
+    font-size: 85%;
```
jimbo, line 27: We should have global tokens for font size in `tokens.css` now.

packages/venia-ui/lib/components/MiniCart/miniCart.js
revanth0212: Man, I should utilize the shutdown to improve my spellings, @sirugh pointed out [something similar](https://github.com/magento/pwa-studio/pull/2506#discussion_r445246028) earlier today. 🤣

soumya-ashok:
@revanth0212 Some feedback -
The scroll bar in the mini cart when >2 products are added is very close to the "delete" icon. Is it possible to add some more padding there?
The product name and the details need to be the same font size with only the weights being different. The mockup has 14px for both.
Could we increase the padding between the header and the first product listing?

packages/venia-ui/lib/components/MiniCart/miniCart.css
revanth0212: I was going with the idea of `min-height=1 ITEM HEIGHT` and `max-height=2.5 ITEMS HEIGHT`. If the `min-height` is any bigger than that, there seems to be an empty space in case of 1 item in the cart. My CSS knowledge is kind of rusty, let me know if there is a cleaner solution than that.

revanth0212:
@jimbo and @soumya-ashok I have addressed your comments. Ready for re-review.

revanth0212:
> > When a user adds an item to the cart, the shopping bag is not updated immediately. It is getting updated after the page refresh.
> 
> Consider the UX requirements when deciding how/when to fetch data. If a component is permanently rendered on the page, such as the cart trigger/icon, then other components must refetch the data necessary. In the case of the add to cart calls, the mutations fetch `total_quantity` which is what the trigger is listening for, and so it updates "automagically".
> 
> So my question is this - is the shopping bag permanently rendered and just hidden? Or is it mounted once a user clicks the trigger icon? If the latter then a simple `useQuery` call should work. If it's permanently mounted then all data necessary to render the components must be fetched by any mutations made on the `cart` object. If that's the case you'll want to expose mini cart fragments and then the add action will need to fetch them in the mutation response.
> 
> I'm pretty sure this it's permanently rendered as you can see the `ShoppingBagQuery` being made when you first load the page.

@sirugh I have fixed this in c580b8b

soumya-ashok:
@revanth0212 Listings in the mini-cart look good, UX approved.

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
```diff
@@ -0,0 +1,37 @@
+import gql from 'graphql-tag';
+
+import { ProductListFragment } from './ProductList/productList.gql';
+
+export const MINI_CART_QUERY = gql`
+    query MiniCartQuery($cartId: String!) {
+        cart(cart_id: $cartId) @connection(key: "Cart") {
+            id
+            total_quantity
+            ...ProductListFragment
```
sirugh, line 10: Would be good to pull these out into a `MiniCartFragment` that we can then import for the `addProductToCart` mutation response.

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
```diff
@@ -0,0 +1,67 @@
...
+            try {
+                const { error } = await removeItem({
+                    variables: {
+                        cartId,
+                        itemId: id
+                    }
+                });
+
+                if (error) {
+                    throw error;
```
sirugh, line 51: Check out how we're handling removal mutations in [Tommy's PR](https://github.com/magento/pwa-studio/blob/49cc9950a1a50f7c42cd3b8197dccde7e2a6e44b/packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js#L55). We don't catch errors and we handle the messaging through deriving from the hook `error` object.

packages/venia-ui/lib/components/MiniCart/miniCart.js
```diff
@@ -2,10 +2,19 @@ import React from 'react';
...
 import { mergeClasses } from '@magento/venia-ui/lib/classify';
 
+import ProductList from './ProductList';
+
+import MiniCartOperations from './miniCart.gql';
+
 import defaultClasses from './miniCart.css';
 
+const Error = () => {
+    return <div>TBD</div>;
```
sirugh, line 15: You're gonna handle this right?

packages/venia-ui/lib/components/MiniCart/ProductList/productList.js
```diff
@@ -0,0 +1,32 @@
...
+import { mergeClasses } from '../../../classify';
+
+import defaultClasses from './productList.css';
+
+const ProductList = props => {
+    const { items, loading, handleRemoveItem, classes: propClasses } = props;
+    const classes = mergeClasses(defaultClasses, propClasses);
+
+    const cartItems = useMemo(() => {
+        if (!loading && items) {
```
sirugh, line 14: Did you get something figured out for when there are no items? Looks a bit weird right now:

[Image from Gyazo](https://gyazo.com/a4555428984f2cdc074f14d217940708)

packages/venia-ui/lib/components/MiniCart/ProductList/item.js
```diff
@@ -0,0 +1,67 @@
...
+import { Price } from '@magento/peregrine';
+
+import ProductOptions from '../../LegacyMiniCart/productOptions';
+import Image from '../../Image';
+import Icon from '../../Icon';
+import { mergeClasses } from '../../../classify';
+
+import defaultClasses from './item.css';
+
+const Item = props => {
```
sirugh, line 13: Add prop types please :)

packages/venia-ui/lib/components/MiniCart/ProductList/productList.js
revanth0212: Nice catch, It is being handled in #2511

packages/venia-ui/lib/components/MiniCart/miniCart.js
revanth0212: I asked a question on the top of this PR, may be lost in the conversation. What do you guys think should happen in case of an error? Show error toast or the GQL message in the contents section?

revanth0212:
> How was this causing the add mutation to not update the product list?

The Fragments were not same byte-wise. They differ in what image they fetch. The one in add to cart fetches `small_image` but the mini cart fragment needs thumbnail instead. Since the fragments are not the same, GQL was not updating the mini cart.

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
tjwiebell: One step further, I would say you only need to wrap mutations in a try/catch if you need block execution of the rest of the callback. I would say [this pattern](https://github.com/magento/pwa-studio/pull/2495/files#diff-f834782b75e39e20a9fc59a568a061d5R119-R143) is a better reference for doing that (just return), and no need to try/ catch otherwise, just rely on `error` in the `useMutation` response.

packages/venia-ui/lib/components/MiniCart/miniCart.js
sirugh: A toast is a good starting point. @tjwiebell did some work recently to handle errors, you may want to check out his pr: https://github.com/magento/pwa-studio/pull/2495/files

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
revanth0212: Handled in a811662

packages/venia-ui/lib/components/MiniCart/miniCart.js
revanth0212: Handled in a811662

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
revanth0212: Can you elaborate? I didn't understand what you meant here.

revanth0212:
Done handled PR comments. @sirugh it is up for re-review.

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
sirugh: I'm just suggesting we try to use composition where possible for the queries/fragments. It removes the guess work for other components that need to update global things. For example:

```gql
mutation AddToCart(args) {
    cart {
       ...MiniCartFragment
    }
}
```

`MiniCartFragment` would just be `id`, `total_quantity` and include `...ProductListFragment`.

revanth0212: I have created the `MiniCartFragment` in 018543e. At first, I did not understand the use of it but kind of made sense when I took a look at https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CartPage/cartPageFragments.gql.js#L8. I am not sure if that is what you meant when you said removing the guesswork for components.

sirugh: It's just one of those pre-optimizations so we don't have to keep the fragments in sync whenever we update. For example, if we later add some display data to the mini cart we would need to make sure it also gets fetched when we add to cart. Now, add to cart can just import the mini cart fragment and it'll automatically work when we update the fragment. Without this we would have to update multiple gql queries.

packages/venia-ui/lib/components/MiniCart/ProductList/productList.js
```diff
@@ -0,0 +1,63 @@
...
+                <Item
+                    key={item.id}
+                    {...item}
+                    handleRemoveItem={handleRemoveItem}
+                />
+            ));
+        }
+    }, [items, loading, handleRemoveItem]);
+
+    if (loading) {
```
sirugh, line 26: Was the desire to display this loading indicator when we remove items? It feels a little excessive. Maybe run it by UX?

[Image from Gyazo](https://gyazo.com/ac33311bf57e1ae48bc4102b4eb512d8)

packages/venia-ui/lib/components/MiniCart/ProductList/productList.js
```diff
@@ -0,0 +1,63 @@
...
+        return <LoadingIndicator>{`Fetching Items in Cart`}</LoadingIndicator>;
+    }
+
+    return <div className={classes.root}>{cartItems}</div>;
```
sirugh, line 30: Did we ever land on a desired UX for empty cart? The current looks a bit strange as there's a box with a scroll bar and nothing in it.

[Image from Gyazo](https://gyazo.com/bb0f2c8c846e6b71437bb995c490ffd4)

soumya-ashok: Yes, I believe we wanted a simple message in there that says "There are no items in your cart"?
Ignore the height of the cart in the mock, and just use the min-height we have determined.

Desktop - Mini Cart - No items

packages/venia-ui/lib/components/MiniCart/ProductList/productList.js
soumya-ashok: This is definitely quite extra. @schensley had done work on loading indicators a while ago I think. Could we use a pattern from that here?

jimbo: I had hoped for the background loading indicator to have some use here, replacing some or all of the content in the minicart's header (count and/or subtotal) Is it not usable here?

packages/venia-ui/lib/components/MiniCart/ProductList/productList.js
revanth0212: This has been handled in #2511 already.

packages/venia-ui/lib/components/MiniCart/ProductList/productList.js
revanth0212: What do you guys think should be the alternative here? Should we handle it here or as part of a different ticket? I am not sure if the loading indicator is part of the original ticket, so if possible, it can be handled in a new one instead of expanding the scope of the current.

revanth0212: @soumya-ashok @schensley I have removed the loading indicator while deleting an item in the cart. The new behavior is to remove the delete icon and change the opacity of the item to 0.5 to indicate that it is being deleted. Check it out and let me know if that is how we want to handle it.

jimbo: Looks pretty good to me.

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
```diff
@@ -1,14 +1,17 @@
...
 
 import { ProductListFragment } from './ProductList/productList.gql';
+import { PriceSummaryFragment } from '../CartPage/PriceSummary/priceSummaryFragments';
 
 export const MiniCartFragment = gql`
     fragment MiniCartFragment on Cart {
         id
         total_quantity
         ...ProductListFragment
+        ...PriceSummaryFragment
```
sirugh, line 11: Will need to refetch the entire cart/checkout page fragments, not just the price summary.

Shipping values change, overall cart value can change and go to $0 if there is a GC covering some of the cost, etc.

revanth0212: I purposefully didn't add the GC, Coupons, etc because those won't change anyway based on the cart. But you might be on to something here. To avoid weird edge cases, it is better to fetch the whole cart fragment.

Do we need to have the mini cart fragment same as the cart fragment?

revanth0212: Updated in 3a399cd

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
```diff
@@ -1,17 +1,13 @@
...
-import { PriceSummaryFragment } from '../CartPage/PriceSummary/priceSummaryFragments';
+import { CartPageFragment } from '../CartPage/cartPageFragments.gql';
 
 export const MiniCartFragment = gql`
     fragment MiniCartFragment on Cart {
         id
-        total_quantity
-        ...ProductListFragment
-        ...PriceSummaryFragment
+        ...CartPageFragment
```
sirugh, line 8: You should include `CheckoutPageFragment` for the same reason you need to include `CartPageFragment`. You should also include every top-level fragment explicitly that you know about, even if they overlap:

```suggestion
        ...CartPageFragment
        ...CheckoutPageFragment
        ...ProductListFragment
```

drc0:
Is it possible to link the product title/thumbnail in the minicart or cart edit page to the product detail page? so if a client want to edit a product in the cart can do it easily.

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
revanth0212: Product List Fragment is already part of the cart fragment and sure it is also part of the checkout fragment. I will do this, but I have a bad feeling we are requesting too much stuff in preparations of edge case bugs and that is gonna decrease the performance of the delete operation.

revanth0212:
> Is it possible to link the product title/thumbnail in the minicart or cart edit page to the product detail page?

Hey @drc0, nice question. I'll leave the linking part to @soumya-ashok and @schensley. 

>  so if a client want to edit a product in the cart can do it easily.

What do you mean by this? If we provide a link to the item in the cart, once you land on the page, you can not edit the item, you can only add a new item with a new/existing options.

You can edit an existing item in the cart page though.

drc0:
> What do you mean by this? If we provide a link to the item in the cart, once you land on the page, you can not edit the item, you can only add a new item with a new/existing options.
> 
> You can edit an existing item in the cart page though.

for example one could add again a product with different combinations or review the product it is buying

revanth0212:
> > What do you mean by this? If we provide a link to the item in the cart, once you land on the page, you can not edit the item, you can only add a new item with a new/existing options.
> > You can edit an existing item in the cart page though.
> 
> for example one could add again a product with different combinations or review the product it is buying

So you mean adding new items not editing the existing ones. Got it. I'll have to talk to @soumya-ashok and @schensley about this.

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
sirugh: > Product List Fragment is already part of the cart fragment and sure it is also part of the checkout fragment. 

This is by coincidence only. The fragment you defined shares properties with the fragment used in cart and checkout to retrieve product details but it is only a coincidence. If cart or checkout decide they need more or less information, the product list in the mini cart may break. 

> I will do this, but I have a bad feeling we are requesting too much stuff in preparations of edge case bugs and that is gonna decrease the performance of the delete operation.

This is the concern I raised in slack yesterday but the decision was made by @jimbo and UX. Perhaps consider adding a comment with some details about why `CartPageFragment` and `CheckoutPageFragment` are included.

drc0:
> So you mean adding new items not editing the existing ones. Got it. I'll have to talk to @soumya-ashok and @schensley about this.

I have a proof of concept of those changes https://github.com/drc0/pwa-studio/commit/24aa2c7c4d073e43364bafb0fcf287bf24e88fa5

soumya-ashok:
@drc0 
Our intent is that the edit action is available only on the full shopping cart page, and not in the mini cart. This is an explicit CTA that will allow the shopper to edit the configuration / options of the product already in their cart.

The image and product name in the listing will be linked to the product's PDP, and yes, adding to cart from there will create a new listing in the shopping cart and not edit the existing listing.

revanth0212:
> Is it possible to link the product title/thumbnail in the minicart or cart edit page to the product detail page? so if a client want to edit a product in the cart can do it easily.

@drc0 we have a ticket in our backlog to work on this. Thank you for the suggestion.

dpatil-magento:
QA Approved.

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
jimbo: To be clear, I think it's a problem that the current approach requires us to overfetch unrelated data here in order to ensure that other components reflect new data. I consider that a problem to solve.

However, it's better to have a slow delete operation and accurate data than to have fast operations and inaccurate data. Especially when it comes to price.

packages/venia-ui/lib/components/MiniCart/miniCart.gql.js
```diff
@@ -34,10 +30,14 @@ export const REMOVE_ITEM_MUTATION = gql`
             cart {
                 id
                 ...MiniCartFragment
+                ...CartPageFragment
+                ...CheckoutPageFragment
```
sirugh, line 34: I imagine that checkout page won't actually update correctly because this fragment is _just_ `total_quantity`. Does shipping methods and the value change when you remove an item? What happens if you have a gift card applied that does not cover the value and then remove an item that makes the gift card cover the entire value?

revanth0212: Mini Cart will not be available on the checkout page. That should solve the problem but if someone in the future wants mini cart on the checkout page they should make sure to request required data in the mutation.

sirugh: Is there a ticket for disabling it on the checkout page? 

Also, if it won't be available on the checkout page then remove the fragment.

revanth0212: Here it is: https://jira.corp.magento.com/browse/PWA-737

I wanted to remove it while working on that ticket. Till then let's have it.

PR #2510, opened 2020-06-24
[PWA-623]: Update Venia's inputs

anthoula:
## Description

Adjust input styles to match the new design

## Related Issue
https://jira.corp.magento.com/browse/PWA-623

## Acceptance 
* Verify inputs are updated throughout the app
* Ensure there is no broken functionality with the styling change

### Verification Stakeholders
@soumya-ashok 
@jimbo 
@dpatil-magento 

### Verification Steps

1. Navigate around the app following all known user flows
2. Verify all inputs previously present are present and work
3. Verify all inputs are positioned and relevant labels are readable

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2510.pwa-venia.com : LH Performance Expected 0.85 Actual 0.5, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2510.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.35, LH Best Practices Expected 1 Actual 0.92
https://pr-2510.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.4, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.css
```diff
@@ -66,6 +61,15 @@
     composes: root from '../../Icon/icon.css';
 }
 
+.button:hover .icon svg {
+    stroke: rgb(var(--venia-global-color-gray-900));
+}
+
+.button:disabled .icon svg,
+.button:disabled:hover .icon svg {
```
jimbo, line 69: Ah, were we not able to pass a `--stroke` here for some reason?

packages/venia-ui/lib/components/RadioGroup/radio.js
```diff
@@ -13,24 +14,30 @@ export class RadioOption extends Component {
...
-        const { classes, id, label, value, ...rest } = props;
+        const { classes, id, initialChecked, label, value, ...rest } = props;
 
         return (
             <label className={classes.root} htmlFor={id}>
                 <Radio
                     {...rest}
                     className={classes.input}
                     id={id}
+                    defaultChecked={initialChecked}
```
jimbo, line 34: Good catch.

packages/venia-ui/lib/components/SearchBar/autocomplete.css
```diff
@@ -16,15 +16,18 @@
...
+    border-top-style: none;
+    box-shadow: -6px 6px rgb(var(--venia-brand-color-1-100));
     display: grid;
     font-size: 0.8125rem;
     gap: 0.75rem;
     left: 0;
     padding: 1rem 1rem;
     position: absolute;
     right: 0;
+    top: 2.2rem;
```
jimbo, line 30: Can this be rounded to an integer?

```suggestion
    top: 2.25rem;
```

packages/venia-ui/lib/index.css
```diff
@@ -2,6 +2,8 @@
 
 :global(*) {
     box-sizing: border-box;
+    margin: 0;
+    padding: 0;
```
jimbo, line 6: I used this shortcut on the styleguide app, but this seems like something that could cause widespread or hard-to-find regressions. Does it seem otherwise to you?

anthoula: This was to fix an issue with radio buttons positioning. This looked fine in multiple browsers. However, to be safe, I will move this reset into the RadioGroup component.

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.css
anthoula: Yeah, it's hard to pass in the custom variable for `--stroke` here, possibly due to the `.icon` override from above, which also declares a different value for `--stroke`

soumya-ashok:
@jimbo @anthoula 

On first glance, looking at the fields, particularly on pages like checkout where there are a number of them - the combination of the 2px outline and the darker gray color feels visually overwhelming to me.

We could keep the gray and bump the outline down to 1px or, keep the outline at 2px but change the color to Hex #B5B8BA.

Screen Shot 2020-07-08 at 9 47 14 AM

I have the same comment about the search field (I know we addressed that as a unique case), but an additional thing there is that the text "search" in the header turns a blue that isn't in our palette when the search field is engaged.

Screen Shot 2020-07-08 at 9 25 06 AM

anthoula:
> I have the same comment about the search field (I know we addressed that as a unique case), but an additional thing there is that the text "search" in the header turns a blue that isn't in our palette when the search field is engaged.

@soumya-ashok this was actually introduced in another change and is present on develop: https://develop.pwa-venia.com

anthoula:
> looks like the Account sign in and creation forms are still reflecting the old style of required fields.

> The field "street address 2" is an optional field, but is currently not reflecting the optional field style that is in the XD file.

@soumya-ashok Good catch. I've updated the areas to remove required symbols, and added optional symbols where applicable

soumya-ashok:
@anthoula Thanks for making the spacing adjustments, the field dense forms are looking better!

Per discussion, the field borders will be adjusted to be a lighter gray as part of an accessibility story

I found a few things I wanted to bring to your attention.

The header icons, which are supposed to be the darkest gray are now all a lighter gray - not sure why this happened but thought it may be a result of us trying to adjust the border color on fields.

The ZIP code field in shipping estimation on the cart works even when the field value is wrong - I tried a 4 digit zip and it still gave me the estimate

I'm seeing a white background on checkboxes that are particularly evident in the focus state

Screen Shot 2020-07-09 at 1 47 46 PM

anthoula:
> The ZIP code field in shipping estimation on the cart works even when the field value is wrong - I tried a 4 digit zip and it still gave me the estimate

@soumya-ashok this was actually introduced in another change and is present on develop: https://develop.pwa-venia.com

soumya-ashok:
@anthoula The changes to inputs are UX approved for this round - We'll come back and adjust the grays as a follow up. Thanks for the great work!

jimbo:
This looks good to me. I'm sure there'll be a few edges to find in QA; I'll do some QA myself in the meantime.

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.css
tjwiebell: That icon override was overriding the `root` class with the `root` class from icon 🤯   Seemed redundant, so I deleted the override and was able to get variable cascade to work as expected.

```css
.button:hover {
    --stroke: rgb(var(--venia-global-color-gray-900));
}

.button:disabled,
.button:disabled:hover {
    --stroke: rgb(var(--venia-global-color-gray-400));
}
```

packages/venia-ui/lib/components/ForgotPassword/ForgotPasswordForm/__tests__/__snapshots__/forgotPasswordForm.spec.js.snap
```diff
@@ -22,6 +21,7 @@ exports[`renders correctly 1`] = `
       <span>
         <input
           autoComplete="email"
+          className="undefined "
```
tjwiebell, line 24: Consider mocking classify here too.

packages/venia-ui/lib/components/RadioGroup/radio.js
```diff
@@ -13,24 +14,30 @@ export class RadioOption extends Component {
...
+                    type="radio"
```
tjwiebell, line 35: Was this prop needed?

https://github.com/joepuzzo/informed/blob/master/src/components/form-fields/Radio.js#L49

packages/venia-ui/lib/components/RadioGroup/radio.js
tjwiebell: A radio should get its checked state from the parent RadioGroup value. Just curious why this was needed.

packages/venia-ui/lib/components/CartPage/GiftCards/giftCards.js
```diff
@@ -122,7 +122,11 @@ const GiftCards = props => {
                 id={classes.card}
                 label="Gift Card Number"
             >
-                <div className={classes.card_input_container}>
+                <div
+                    className={`${classes.card_input_container} ${
+                        shouldDisplayCardError ? classes.error : ''
```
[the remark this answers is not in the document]
jimbo: @tjwiebell No, I definitely overlooked it. You're right, we avoid classname concatenation in all cases.

As for your suggestion, I don't recommend using the `:not` pseudo selector.

packages/venia-ui/lib/components/RadioGroup/radio.js
anthoula: It's redundant, good call

packages/venia-ui/lib/components/CartPage/GiftCards/giftCards.js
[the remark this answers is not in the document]
anthoula: All cases should now use modifiers, with the exception of `filterDefault.js`. Those checkboxes should really be `<input type="checkbox">`, instead of `<button>`. Semantics of filters will be addressed in upcoming story PWA-713.

packages/venia-ui/lib/components/TextInput/textInput.js
```diff
@@ -30,13 +30,17 @@ export class TextInput extends Component {
             ...rest
         } = this.props;
 
+        const inputClass = fieldState.error
+            ? classes.input_error
+            : classes.input;
+        console.log(inputClass);
```
tjwiebell, line 36:
```suggestion
```

dpatil-magento:
@jimbo   we see minor issue in Android device where Radios in Qty drop down, County/State in shipping form dropdown does not look same as rest of site.  But its very specific to how android handles it, iOS don't have issue as there is no Radios on these fields. Does this needs to be handled?

cc @anthoula 

image

jimbo:
> @jimbo we see minor issue in Android device where Radios in Qty drop down, County/State in shipping form dropdown does not look same as rest of site. But its very specific to how android handles it, iOS don't have issue as there is no Radios on these fields. Does this needs to be handled?
> 
> cc @anthoula
> 
> image

That's working as intended. No fix necessary. 👍

dpatil-magento:
QA Approved.

PR #2513, opened 2020-06-25
Adjust page-level max-width

jimbo:
## Description

Update the max-width on a few page-level elements: header, main page, footer. Also fix some header spacing while we're in there.

## Related Issue

PWA-734

### Verification Stakeholders
@dpatil-magento 
@soumya-ashok 

### Verification Steps
1. Visit the home page in a window with a viewport wider than `1440px`.
2. Verify the header content maxes out at `1440px`, while its border stretches edge-to-edge.
3. Verify the footer content maxes out at `1440px`, along with its border.
4. Verify no change at lower viewport widths.
5. Verify `/venia-new-home` still looks good.
6. Verify a product page still looks good.

desktop

mobile

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2513.pwa-venia.com : LH Performance Expected 0.85 Actual 0.5, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 79
https://pr-2513.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.35, LH Best Practices Expected 1 Actual 0.92
https://pr-2513.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.4, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

soumya-ashok:
@jimbo The header and footer behavior seems correct, there is padding on either side at 1440 - assume that is intentional
Footer content gets center aligned in a single column layout but the design intent is that it should all be left aligned.

Screen Shot 2020-07-22 at 5 06 19 PM

Screen Shot 2020-07-22 at 5 07 39 PM

On the new home page, could we link all the buttons on it to go to various category pages instead of the same for all?

There's some jank in the product description on PDP. Looks like the heading SKU is center aligned in the column but should be left aligned?
Screen Shot 2020-07-22 at 5 11 30 PM

jimbo:
> The header and footer behavior seems correct, there is padding on either side at 1440 - assume that is intentional

Yes, it's intentional.

> Footer content gets center aligned in a single column layout but the design intent is that it should all be left aligned.

I've changed the layout to fix this.

> On the new home page, could we link all the buttons on it to go to various category pages instead of the same for all?

Not related to this PR, but I assume we could update the backend links at some point.

> There's some jank in the product description on PDP. Looks like the heading SKU is center aligned in the column but should be left aligned? Same for the product description heading, actually.

Also not related to this PR. I've fixed it anyway, though.

packages/venia-ui/lib/components/Main/main.css
```diff
@@ -10,6 +10,8 @@
 }
 
 .page {
+    margin: 0 auto;
+    max-width: 1440px;
```
sirugh, line 14:
```suggestion
    max-width: var(--venia-global-maxWidth);
```

dpatil-magento:
@jimbo Search box width is reduced, expected?

image

jimbo:
> Search box width is reduced, expected?

Good catch. I've fixed this. I also adjusted the spacing above & below the search box slightly—less above it, more below it now.

PR #2517, opened 2020-06-26
[PWA-720] Error Case Tests

tjwiebell:
## Description

Test cases for #2495, merged ahead of tests to unblock regression.

## Related Issue

* [[PWA-720](https://jira.corp.magento.com/browse/PWA-720)] 

### Verification Steps

Since all files changes are all test files (`*.spec.js` + snapshots), this really just needs green builds to verify.
1. `yarn test`
2. Verify builds pass

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/__tests__/__snapshots__/useShippingForm.spec.js.snap
+++ b/packages/peregrine/lib/talons/CartPage/PriceAdjustments/ShippingMethods/__tests__/__snapshots__/useShippingForm.spec.js.snap
@@ -1,0 +1,23 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`returns correct shape 1`] = `
+Object {
+  "formErrors": Array [
+    undefined,
+  ],
+  "handleOnSubmit": [Function],
+  "handleZipChange": [Function],
+  "isSetShippingLoading": false,
+}
+`;
+
+exports[`returns errors and loading state from Apollo 1`] = `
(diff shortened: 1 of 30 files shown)


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2517.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92
https://pr-2517.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.32, LH Best Practices Expected 1 Actual 0.92
https://pr-2517.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.49, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92

tjwiebell:
> One pattern stood out to me which is that we return the errors in an array made from the objects that are potentially undefined. This results in `[undefined, SomeOtherError]` in certain cases. This might cause issues with mapping over the error object if people customize their form error handler and don't realize it.
> 
> https://github.com/magento/pwa-studio/blob/da2d819e5c6c937b5e9e411de48024b7a9225c1e/packages/peregrine/lib/talons/FormError/useFormError.js#L7

Good point. My goal was to remove all the boilerplate from talons and let `FormError` handle parsing, which means default values for `error` of `undefined` in that response. I would propose documentation over code change, but I'm open to adding the `filter(Boolean)` to the talon responses instead.

dpatil-magento:
Tests pass on node 10 and 12. QA Approved.

PR #2523, opened 2020-07-06
PWA-722: Venia button follow up changes

davemacaulay:
## Description
Resolve remaining issues with Buttons implementation 

## Related Issue
https://jira.corp.magento.com/browse/PWA-722

### Verification Stakeholders
@sirugh 
@jimbo 
@dpatil-magento 

### Verification Steps
Issue 1 - In Mobile view "Reset Filter" button does not get disabled after first tap.
Steps - 
1. Go any category in mobile view
2. Hit filter button and apply a filter
3. Tap on Reset filter button
Issue 2 - Mini Cart Edit product Cancel button UI and order of display is not consistent with other buttons of Venia.
Steps- 
1. Add any product to cart.
2. In mini cart go to edit product window

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/Button/button.css
+++ b/packages/venia-ui/lib/components/Button/button.css
@@ -34,7 +34,8 @@
     --stroke: var(--venia-brand-color-1-800);
 }
 
-.root:disabled {
+.root:disabled,
+.root:hover:disabled {
     pointer-events: none;
     --stroke: var(--venia-global-color-gray-400);
 }
--- a/packages/venia-ui/lib/components/Button/button.css
+++ b/packages/venia-ui/lib/components/Button/button.css
@@ -34,6 +34,10 @@
     --stroke: var(--venia-brand-color-1-800);
 }
 
+/**
+ * Some browsers retain the :hover state after a click, this ensures if a button becomes disabled after
+ * being clicked it will be visually disabled.
+ */
 .root:disabled,
 .root:hover:disabled {
     pointer-events: none;
--- a/packages/venia-ui/lib/components/Button/button.js
+++ b/packages/venia-ui/lib/components/Button/button.js
@@ -24,13 +24,19 @@ const Button = props => {
         priority,
         type,
         negative,
+        disabled,
         ...restProps
     } = props;
     const classes = mergeClasses(defaultClasses, propClasses);
     const rootClassName = classes[getRootClassName(priority, negative)];
 
     return (
-        <button className={rootClassName} type={type} {...restProps}>
+        <button
+            className={rootClassName}
+            type={type}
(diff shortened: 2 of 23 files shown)


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2523.pwa-venia.com : LH Performance Expected 0.85 Actual 0.51, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2523.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2523.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/lib/components/Button/button.css
```diff
@@ -34,7 +34,8 @@
     --stroke: var(--venia-brand-color-1-800);
 }
 
-.root:disabled {
+.root:disabled,
+.root:hover:disabled {
```
jimbo, line 38: Did you have a specific browser or use case that required the second line here? If not, can we remove it? It looks redundant, and a quick test in Chrome & Firefox seems to bear that out.

packages/venia-ui/lib/components/Button/button.js
```diff
@@ -64,13 +72,15 @@ Button.propTypes = {
...
+    negative: bool,
+    disabled: bool
 };
 
 Button.defaultProps = {
     priority: 'normal',
     type: 'button',
-    negative: false
+    negative: false,
+    disabled: false
```
jimbo, line 83: Probably a good idea. 👍

packages/venia-ui/lib/components/Button/button.css
davemacaulay: @jimbo chrome keeps the `:hover` state on the button once you press it. Safari does something similarly weird where you have to move the mouse by at least 1 pixel to remove the `:hover` state. This caused a button which became disabled to remain active until something removed the hover state.

Having the `:disabled` state override the `:hover` state seemed the easier solution here. There might be an underlying problem here with the hover states that I'm unaware of, so if there's a better way to resolve this please let me know!

jimbo: No, that's enough of a reason to keep it as-is. If you add a comment, I'll approve this. 👍

davemacaulay: @jimbo comment added 😄

PR #2540, opened 2020-07-08
[spike]: Using apollo client state to store cart id

sirugh:


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2540.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2540.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.37, LH Best Practices Expected 1 Actual 0.92
https://pr-2540.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/peregrine/lib/context/cart.js
```diff
@@ -72,3 +97,31 @@ export default connect(
...
+                    const cacheData = context.cache.readQuery({
+                        query: GET_LOCAL_CART_ID
+                    });
+                    cartId = cacheData;
+                } catch (err) {
+                    console.log('No cached cart id, fetching from server');
+                    const fetchedData = await context.client.mutate({
+                        mutation: createCartMutation
+                    });
+                    cartId = fetchedData.data.cartId;
```
sirugh, line 118: This handles the creation of a cart id if one doesn't exist at all. When the app loads for the first time the cache is an empty object so the read fails and we end up here, which write the value to the cache.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -163,10 +166,34 @@ export const useProductFullDetail = props => {
...
+                    // We could specify to only re-create cart id for specific
+                    // errors.
+                    const { data } = await apolloClient.mutate({
+                        mutation: createCartMutation
+                    });
+
+                    apolloClient.writeQuery({
+                        query: GET_LOCAL_CART_ID,
+                        data
+                    });
```
sirugh, line 191: Initially I thought I could just write an empty object or `null` here to trigger the re-query in the cart context but that didn't work. It feels bad to have to handle recreation of the cart id in this way, especially when the retry pattern should be:

1) Validate/recreate auth token
2) Validate/recreate cart id

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -163,10 +166,35 @@ export const useProductFullDetail = props => {
...
+                    // We could specify to only re-create cart id for specific
+                    // errors.
+                    const { data } = await apolloClient.mutate({
+                        mutation: createCartMutation
+                    });
+                    console.log('Replacing local cart id!');
+                    apolloClient.writeQuery({
+                        query: GET_LOCAL_CART_ID,
+                        data
+                    });
```
sirugh, line 192: Does it make sense to recreate the cart id like this in every mutation/query where the cart id may be invalidated? Feels as if it could get unwieldy. Would prefer to do it somewhere more global like the cart context resolver.

sirugh:
Rolling into #2541

PR #2539, opened 2020-07-08
[spike]: Using local storage to hold cart id

sirugh:

PR #2541, opened 2020-07-09
[spike]: Uses apollo resolver/cache without need for context

sirugh:
**Spike Writeup**

1) How is the `cartId` created?

A local resolver will query for the cached value. If one is not found it will make a `createCart` mutation, store the value in cache, and return it to the initial query.

2) Where should the cart id be stored?

Storing the `cartId` in local storage [manually](https://github.com/magento/pwa-studio/pull/2539) is unnecessary. Cache persist will handle this for us anyways when we use the resolver to [writeQuery](https://github.com/magento/pwa-studio/pull/2541/files#diff-48b842d9d63f76fa70820dcae24bb31cR126-R129).

3) How do queries/mutations access cartId?

For queries, by using the `@client` directive to indicate the field is to be resolved locally and by the `@export` directive to automatically pass the value as the query variable. NOTE: You still have to write your queries with the `cartId` arg, even if the components might not pass it in.

```
query getCartDetails($cartId: String!) {
    cartId @client @export(as: "cartId")
    cart(cart_id: $cartId) @connection(key: "Cart") {
        id
        ...CartPageFragment
    }
}
```

For mutations we are unable to make a query and mutation in a single operation. The following will not work:

```
mutation addSimpleProductToCart($quantity: Float!, $sku: String!) {
    query getCartId {
        cartId @client @export(as: "cartId")
    }
    addSimpleProductsToCart(
        input: {
            cart_id: $cartId
            cart_items: [{ data: { quantity: $quantity, sku: $sku } }]
        }
    ) @connection(key: "addSimpleProductsToCart") {
        cart {
            id
            ...CartPageFragment
        }
    }
}
```

**This means we still need to expose the cart id to components for use in mutations.**

We can accomplish this by [exporting the value](https://github.com/magento/pwa-studio/pull/2541/files#diff-48b842d9d63f76fa70820dcae24bb31cR57) in cart context.

4) How to we prevent "thundering herd" on the very first load?

Using the `skip` option combined with an initial query made by the root, CartContextProvider. This allows the rest of the app to load/render without blocking. The non-performant alternative would be to return `null` from CartContextProvider until the cartId query is complete. Additionally, the automatically included `apollo-link-dedupe` does not work for local resolvers so we _must_ use either `skip` or some other method to prevent the queries.


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2541.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2541.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.39, LH Best Practices Expected 1 Actual 0.92
https://pr-2541.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.42, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

sirugh:
Apollo v3 brings some cool features such as [reactive variables](https://www.apollographql.com/docs/react/local-state/reactive-variables/) that may obviate much of this work. 

https://www.apollographql.com/docs/react/local-state/managing-state-with-field-policies/#storing-local-state-in-reactive-variables

PR #2494, opened 2020-06-17
New MiniCart

supernova-at:
## Description

This PR introduces a skeleton for the new `MiniCart`.

The new `MiniCart`'s display is controlled by the `CartTrigger` button in the site header.
On Desktop, the `CartTrigger` will show this new `MiniCart`.
On mobile, the `CartTrigger` will take the user directly to the Cart page.

Once the `MiniCart` is visible, clicking anywhere outside of it will dismiss it.

It also:
  * Renames the original `MiniCart` to `LegacyMiniCart`
  * Creates a new Storybook entry in the `Venia` category

Follow up tasks in the new MiniCart epic will implement this component's actual functionality.

## Related Issue

Closes [PWA-600](https://jira.corp.magento.com/browse/PWA-600).

### Verification Stakeholders

@jimbo 

#### On Desktop

1. Add an item to your cart
1. Verify that the MiniCart no longer automatically appears
1. Click the CartTrigger (shopping bag icon in the site header)
1. Verify that the new MiniCart skeleton appears
1. Verify that clicking outside of the MiniCart dismisses it

#### On Mobile

1. Click the CartTrigger (shopping bag icon in the site header)
1. Verify that you are directed to the new Cart page
1. Verify that the new MiniCart does *not* appear

#### In Storybook

1. `yarn workspace @magento/venia-ui run storybook`
1. Verify a new "MiniCart" entry appears in the "Venia"-specific category

## Screenshots / Screen Captures (if appropriate)

Screen Shot 2020-06-24 at 1 56 02 PM

## Checklist

- [x] I have added tests to cover my changes, if necessary.
* I have updated the documentation accordingly, if necessary.

PR #2543, opened 2020-07-09
Scroll lock effect avoid body shift when activated

drc0:
## Description

Fix scroll lock page shifting https://github.com/magento/pwa-studio/pull/2494#issuecomment-655589395
this works by always showing a scrollbar even if disabled

*@jimbo edit: Disabling scroll lock in the CSS now. Leaving `useScrollLock()` calls in place.*

## Related Issue

Closes PWA-735.


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2543.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2543.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.39, LH Best Practices Expected 1 Actual 0.92
https://pr-2543.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.42, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

sirugh:
The alternative to this would be to allow the background page to scroll while the mini cart is open. Shouldn't be a problem since the header is sticky. It might present an issue when trying to scroll within the mini cart.

cc @jimbo @schensley

drc0:
if you allow the page to scroll, when reaching the bottom of the minicart, usually most browsers lets you scroll the behind page and that is annoying.

jimbo:
Hello, thanks for your contribution! As @sirugh said, we've been discussing this very issue lately. And as the one who [originally implemented this behavior](https://github.com/magento-research/pwa-studio/pull/1449), I can provide some context. 😅 

> if you allow the page to scroll, when reaching the bottom of the minicart, usually most browsers lets you scroll the behind page and that is annoying.

Right. It can be a bit annoying, depending on the browser, so we've aimed to prevent window scrolling while any sort of scrollable overlay is open—as long as the solution meets some criteria. Namely:

1. it doesn't attach a scroll listener or read the DOM (bad for performance)
2. it doesn't affect page layout
3. it doesn't lose scroll position
4. it works in major mobile & desktop browsers (Chrome, Firefox, Edge, Safari)

As it turns out, we haven't found any solutions that meet all of these criteria. The simplest solution is something like the following:

```stylus
html[data-scroll-lock="true"] body
  overflow: hidden
```

Unfortunately, this fails in mobile Safari (criteria 4), and workarounds (such as the fixed positioning we have now) introduce issues that cause criteria 2 and 3 to fail instead. Your proposal in this PR (setting `overflow: scroll` on the root) is an interesting approach to address criteria 2, but we're still dealing with criteria 3 failing because we lose `window.scrollY` when the height of `html` and `body` change to `100%`.

As a result, we'd like to remove the CSS for scroll locking. I think we can leave the calls to `useScrollLock()` in, so it should be easy for implementors to decide how aggressively to pursue scroll locking.

drc0:
lock scrolling is really useful in case of modal dialogs, is this useScrollLock used also in those cases? There are libs out there that put overflow hidden on body, check is a scrollbar is visible and adjust body and modal padding accordingly https://github.com/twbs/bootstrap/blob/main/js/src/modal.js#L441   https://getbootstrap.com/docs/4.0/components/modal/#scrolling-long-content, is a solution like this viable?

jimbo:
> lock scrolling is really useful in case of modal dialogs, is this useScrollLock used also in those cases?

Yes, `useScrollLock` should currently be called by anything we have that masks the page—dialogs, drawers, etc. It simply sets the `data-scroll-lock` attribute on the `html` element.

> There are libs out there that put overflow hidden on body, check is a scrollbar is visible and adjust body and modal padding accordingly https://github.com/twbs/bootstrap/blob/main/js/src/modal.js#L441 https://getbootstrap.com/docs/4.0/components/modal/#scrolling-long-content, is a solution like this viable?

Good thinking. Unfortunately, checking if a scrollbar is visible would require us to measure the DOM, and we try not to do that unless it's absolutely essential. But it's also worth noting that Bootstrap's solution also fails in mobile Safari, [according to their notes here](https://getbootstrap.com/docs/4.0/getting-started/browsers-devices/#overflow-and-scrolling), for the same reason I mentioned before. That's kind of a dealbreaker for us.

As long as we continue to set the data attribute, developers using PWA Studio will be able to restore scroll locking for non-Safari browsers with one line of global CSS, or pursue a more advanced solution if cross-browser scroll locking is super important to them. For Venia, and for other design systems such as Material, I think the solutions are more problematic than the original problem.

I'd like to retain your commits here, though. I can either merge your PR as-is and make my own changes in a follow-up PR, or we can add new commits to this PR. Do you have a preference?

drc0:
ok thank you for you comment, no preference, continue in the cleanest way you prefer.

dpatil-magento:
When there are 2 scroll bar on webpage, it gets bad/odd user experiences, below are some observations with this implementation. Once these observations are addressed  I would like to see UX approval on this approach before it gets  merged. 

Chrome/FF - Add 3 products to cart, open mini cart, move mini cart scroll a bit , make an empty click and try to scroll inside mini cart, browser main scroll takes effect.

<a href="https://gyazo.com/cc6264d3f03b4b86643a77aac17faf29">Image from Gyazo</a>

Safari -  Add 3 products to cart, open mini cart and try scroll indeside cart, both scroll bars take effect.

<a href="https://gyazo.com/07b818e7f0de9145836a43ae06aeb25a">Image from Gyazo</a>

Android (samsung s10e) Chrome - Add product to cart, got to cart page and go to edit item window and try scrolling. Parent window also gets scrolled.

schensley:
@dpatil-magento it appear that in the instance of this PR demo the PDP is allowed to scroll while the mini-cart is open. When the shopper is within the scroll area of the mini-cart the scroll bar appears in the mini-cart, when the shopper moves the cursor over the page and begins the scroll the scroll bar quickly fades in the mini-cart and a scroll bar for the page appears.  It's a slight jump, but is a good way to inform the shopper of the ability to scroll and in what area. This fading from one to the other seems slightly slower in Firefox and Safari than in Chrome, but accomplishes the same result. This meets UX approval.

dpatil-magento:
QA Approved.

PR #2545, opened 2020-07-09
PWA-706: Update Venia's cards to match the style guides cards

eug123:
## Description

Adjust `Card` styles to match the new design (mainly used on Cart & checkout)

## Related Issue

https://jira.corp.magento.com/browse/PWA-706

### Verification Stakeholders

@jimbo
@dpatil-magento
@anthoula
### Verification Steps

1. Navigate to checkout, checkout address book and verify that cards meet the style guide
2. Verify all card instances are working as expected

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2545.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2545.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2545.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.42, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.css
```diff
@@ -77,11 +77,17 @@
...
+    grid-column: 1 / span 1;
+}
+
+@media (min-width: 961px) {
+    .summaryContainer {
+        grid-column: 2 / span 1;
+        grid-row: 2 / span 3;
+        position: sticky;
+        top: 5.5rem;
+        height: min-content;
```
jimbo, line 89: The technique used here to position the summary container was intentional. Does this change attempt to resolve an edge case?

- this element shouldn't need to know how many rows are in the grid
- this element shouldn't affect row sizes in the grid

```stylus
.summaryContainer
  grid-row-: 1 / span 1 // place the element into only one row
  height: 0 // avoid contributing to implicit row height
```

packages/venia-ui/lib/components/CheckoutPage/AddressBook/addressBook.css
```diff
@@ -40,16 +49,24 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    color: rgb(var(--venia-brand-color-1-700));
+}
+.addText {
+    margin-left: 0.625rem;
 }
 
+.addButton:focus,
```
jimbo, line 66: I think focus treatment should be the same as normal buttons (the shadow).

eug123: Changed to be the same as other buttons.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.css
eug123: Yes, this is fixing edge case when summary can overlap footer when scrolling.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/shippingInformation.css
```diff
@@ -27,6 +27,11 @@
     font-weight: 600;
 }
 
+.editButton {
+    padding: 1rem;
+    margin: -1rem;
```
supernova-at, line 32: I see this repeated for all the edit buttons - what problem does this solve?

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.css
```diff
@@ -77,11 +77,21 @@
...
+    }
+
+    .signin_container ~ .summaryContainer {
```
supernova-at, line 92: We try to avoid descendent selectors like this. Can we target the container another way?

eug123: We need `.summaryContainer` to be aligned with shipping information, but in case of guest checkout login section is added (+1 row), so we need to move summary down one row, and using such selector is actually solves this.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/shippingInformation.css
eug123: It was requested by UX to have padding 16px (1rem) padding for these buttons it . Negative margins compensate this added padding, avoiding change of the the heading row dimensions and button positioning.

packages/venia-ui/lib/components/CheckoutPage/OrderSummary/orderSummary.css
```diff
@@ -1,17 +1,23 @@
...
-    .title {
-        font-size: 1.25rem;
-    }
+.title {
+    position: absolute;
+    height: 1px;
+    width: 1px;
+    overflow: hidden;
+    clip: rect(1px, 1px, 1px, 1px);
+    white-space: nowrap;
```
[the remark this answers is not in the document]
eug123: This code visually hides the title, but it still available for screen readers.

packages/venia-ui/lib/components/CheckoutPage/AddressBook/addressBook.css
```diff
@@ -40,16 +49,26 @@
...
     font-size: 0.875rem;
     font-weight: 600;
     transition: border-color 384ms var(--venia-global-anim-standard);
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    color: rgb(var(--venia-brand-color-1-700));
+}
+.addText {
+    margin-left: 0.625rem;
```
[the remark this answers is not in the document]
eug123: Changed to grid, as well as edit buttons with labels.
Thank you

packages/venia-ui/lib/components/CheckoutPage/OrderSummary/orderSummary.css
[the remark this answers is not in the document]
jimbo: Presumably for accessibility reasons. The technique used here allows screen readers to read the element normally (and with full context, as an `h2`). It's an alternative to `display: none` or `visibility: hidden`.

I'm fine with leaving it as it is.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.css
jimbo: True. We should probably just use a media query instead.

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/shippingInformation.css
jimbo: It increases the clickable area without affecting the alignment.

dpatil-magento:
@schensley Let us know if any of these needs to be fixed - 

1. Gift Card, Coupon label is wrapping - 
image

2. Order total card spacing/position when shipping address form is present does not look good - 
image

schensley:
@dpatil-magento in the design mock-ups we did not recommend showing an "applied" status in association with Coupons and Gift Cards, since these would not be shown as line items in the pricing summary until they had been applied to the order. Once a shopper adds a Coupon or Gift Card (and applies it to the order) it is then reflected in the pricing summary -- "Discount  -$15.00" --  providing confirmation and confidence to the shopper that it has in fact been applied to the order.

schensley:
@dpatil-magento regarding point 2 about the spacing of the pricing summary, I agree initially it looks a little odd, but as the shopper progresses through the checkout work flow and previous section assume the review state of cards (to the left) the pricing summary aligns to that and looks better. If it is possible to have the pricing summary container be top aligned with the horizontal line just about the "Guest Checkout" title (and sign in link) initially and then align with the Shipping Information card as the shopper progresses, that would look nice. However, since this a "zombie" container that follows down the page as the shopper scrolls, any alignment is temporary.

dpatil-magento:
@eug123 In Safari edit link is wrapping. 

image

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.css
```diff
@@ -15,6 +15,16 @@
     grid-template-columns: 1fr;
 }
 
+.edit_button {
+    display: grid;
```
jimbo, line 19: Unfortunately, `button` elements don't work consistently as `grid` or even `flex` containers, so this needs a child `span` to serve as the grid container.

I recommend using our `LinkButton` component for these.

packages/venia-ui/lib/components/CheckoutPage/AddressBook/addressBook.css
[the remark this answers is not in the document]
eug123: Needed to revert this changes as buttons does not support `display: grid` in safari

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/summary.css
eug123: Rebuild to use `LinkButton` component, thank you

dpatil-magento:
QA Approved

Issue #2546, opened 2020-07-10
[bug]: Minicart container is preventing filters to be used on category pages


jemoon:
**Describe the bug**
Latest minicart changes introduce issue on each category page while browsing on mobile devices - filters button doesn't trigger Filters Modal anymore.

**To reproduce**
1. Visit category page on mobile device (480px width for example)
2. Click on 'Filters'

**Expected behavior**
Modal with filters should be shown

**Screenshots**
https://www.dropbox.com/s/932ei1s2yd7clce/Screen%20Recording%202020-07-10%20at%2012.52.38.mov?dl=0

**Possible solutions**
Minicart should use display (none/block) instead of visibility (hidden/visible) on close/open actions on larger screens.

**Please complete the following device information:**
 - Device [e.g. iPhone6, PC, Mac, Pixel3]: Dev Tools with Device toolbar set to On and with resolution set to 480px width
 - OS [e.g. iOS8.1, Windows 10]: MacOs
 - Browser [e.g. Chrome, Safari]: Chrome
 - Browser Version [e.g. 22]: 84.0.4147.85

**Please let us know what packages this bug is in regards to:**
- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

sirugh:
I'm unable to repro this on my Windows 10 box using Chrome@ 83.0.4103.116 (Official Build) (64-bit) or on my Pixel 3 with Chrome.

If you are able to repro consistently could you record and share a video of the steps from a fresh page load with a clean cache?

tjwiebell:
I ran into this today reviewing some mobile views, his suggestion seems correct; I don't understand why some elements remain clickable, but likely some z-index luck.

![](https://gyazo.com/c7969b1efe586c2d270c017d9805ee68.gif)

sirugh:
OH! I saw this happen then, but on product page. When I selected the right most color option on a product, then selected the one to the left, the right most option would stop being clickable. 

It stopped happening after a while so I just figured my mac had a problem.

dpatil-magento:
I can reproduce it on reduced window size, part of button is not clickable.

[...]

PR #2547, opened 2020-07-10
Fix MiniCart Interaction Blocking

supernova-at:
## Description

When closed, the MiniCart was blocking interactions on the page behind it.
This PR moves the MiniCart offscreen when it is closed to avoid this behavior.

Alternative solutions:

- We purposefully don't want to unmount / remount it.
- Since we're animating `visibility` we can't switch to using `display` either because CSS can't animate the `display` property

## Related Issue

Closes PWA-743 / #2546 .

### Verification Steps

1. Go to a Product Detail Page (PDP)
1. Open the MiniCart
1. Take note of page elements that it covers (color swatches most likely)
1. Close the MiniCart
1. Verify that you can interact with the elements an open MiniCart covers up

Screen Shot 2020-07-10 at 12 31 00 PM

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/MiniCart/miniCart.css
+++ b/packages/venia-ui/lib/components/MiniCart/miniCart.css
@@ -1,7 +1,7 @@
 .root {
     /* The root aside element takes up the whole screen. */
     position: fixed;
-    left: 0;
+    left: 100vw;
     top: 0;
     height: 100%;
     width: 100%;
@@ -11,7 +11,6 @@
     visibility: hidden;
 
     /* It animates to being closed. */
-    transform: scale(0);
     transition-duration: 192ms;
     transition-timing-function: var(--venia-global-anim-out);
     transition-property: opacity, visibility;
@@ -23,11 +22,11 @@
 .root_open {
     composes: root;
 
+    left: 0;
     opacity: 1;
     visibility: visible;
 
     /* It animates to being open. */
-    transform: scale(1);
     transition-duration: 224ms;
     transition-timing-function: var(--venia-global-anim-in);
 }
--- a/packages/venia-ui/lib/components/MiniCart/miniCart.css
+++ b/packages/venia-ui/lib/components/MiniCart/miniCart.css
@@ -49,7 +49,7 @@
     overflow: hidden;
     padding: 1rem;
     position: fixed;
-    right: 0.5rem;
+    right: -100vw;
     /* This value is equal to the site header height. */
     top: 5rem;
     transform: translate3d(0, -8px, 0);
@@ -62,6 +62,8 @@
 
 .contents_open {
     composes: contents;
+
+    right: 0.5rem;
 
     /* It animates "down" from the header. */
     transform: translate3d(0, 0, 0);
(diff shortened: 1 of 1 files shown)


packages/venia-ui/lib/components/MiniCart/miniCart.css
```diff
@@ -11,6 +11,7 @@
     visibility: hidden;
 
     /* It animates to being closed. */
+    transform: scale(0);
     transition-duration: 192ms;
     transition-timing-function: var(--venia-global-anim-out);
     transition-property: opacity, visibility;
```
supernova-at, line 17: Note that `transform` is purposefully omitted from the `transition-property`: we purposefully don't want to animate the growing and shrinking.

jimbo: Wasn't there supposed to be a little bit of translation going on here, though? I'd swear we had something like `transform: translate3d(0, -0.5rem, 0)`.

Anyway, generally I'd like to leave `transform` on the table for animation without known bugs waiting to happen if it's used. Could we just use positioning (`left: 100vw`) or something?

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2547.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2547.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.38, LH Best Practices Expected 1 Actual 0.92
https://pr-2547.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

dpatil-magento:
@supernova-at MFTF test failed again on one scenario. Able to reproduce locally too. 

1. Set browser width to 1280px.
2. Add 2 configurable products to cart
3. Go to Cart page and observe `Proceed to Checkout` button.

Issue - Middle portion of button is not clickable because mini cart image swatch overlapping it mostly.

<a href="https://gyazo.com/a62e87df9d2e3d72b273184138844a9b">Image from Gyazo</a>

packages/venia-ui/lib/components/MiniCart/miniCart.css
```diff
@@ -49,7 +49,7 @@
     overflow: hidden;
     padding: 1rem;
     position: fixed;
-    right: 0.5rem;
+    right: -100vw;
```
jimbo, line 52: Hmm, this doesn't feel like the solution. I think the problem here is that both `.root` and `.contents` have `position: fixed`, which is relative to the viewport. The latter should probably have `position: absolute`, so that it's relative to the former. That way, when the parent moves offscreen, it'll take the child with it.

```stylus
.root
  position: fixed

.contents
  position: absolute
```

supernova-at: Awesome, thanks. Resolved in https://github.com/magento/pwa-studio/pull/2547/commits/c9812fd6bd0fad45edf71cd89bb55b84af3bfb62.

PR #2548, opened 2020-07-10
[PWA-680] Webpagetest security score is too low

tjwiebell:
## Description

Steps -
1. Run webpagetest for https://develop.pwa-venia.com/ WPT URL - https://www.webpagetest.org/

Expected - New Security should show better results.
Actual - Security score is F.

## Related Issue

* [[PWA-680](https://jira.corp.magento.com/browse/PWA-680)] Webpagetest security score is too low

### Verification Steps

Passing score on WPT
1. https://snyk.io/test/website-scanner/?test=200713_3N_d910d2505fdbd03725cba0a30172769c&utm_medium=referral&utm_source=webpagetest&utm_campaign=website-scanner

Since we're introducing a very opinionated CSP rule in this PR, it would also be good to verify this plays nicely with various setups (Both UPWARDs, onboard/backend image opt).

Screen Shot 2020-07-13 at 10 34 38 AM

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2548.pwa-venia.com : LH Performance Expected 0.85 Actual 0.51, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2548.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.35, LH Best Practices Expected 1 Actual 0.92
https://pr-2548.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.42, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/upward.yml
```diff
@@ -109,6 +109,16 @@ veniaAppShell:
...
+                content-security-policy:
+                    inline: upgrade-insecure-requests
+                strict-transport-security:
+                    inline: max-age=31536000
+                x-content-type-options:
+                  inline: nosniff
+                x-frame-options:
+                  inline: SAMEORIGIN
+                x-xss-protection:
+                  inline: '1; mode=block'
```
jimbo, line 121: This may be [unnecessary](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection). Does WPT expect it?

tjwiebell: The ones included are the five specific headers WPT looks for, dings for each missing one. The browsers that don't support CSP and need this rule we probably don't support anyway, so that's a fair point. We have the option to not include them and defend whatever score WPT decides to give us.

packages/venia-concept/package.json
```diff
@@ -52,6 +52,7 @@
     "@magento/peregrine": "~6.0.0",
     "@magento/pwa-buildpack": "~5.1.1",
     "@magento/upward-js": "~4.0.1",
+    "@magento/upward-security-headers": "*",
```
zetlen, line 55: This probably should follow the pattern of other deps and be `~0.0.1`.

packages/extensions/upward-security-headers/intercept.js
```diff
@@ -0,0 +1,26 @@
...
+                } could not find its own definition in the emitted upward.yml`
+            );
+        }
+
+        for (const headerName in definitions[SECURITY_HEADER_DEFINITION]
+            .inline) {
+            definitions.veniaAppShell.inline.headers.inline[
+                headerName
+            ] = `${SECURITY_HEADER_DEFINITION}.${headerName}`;
+        }
```
zetlen, line 24: I know, a `for...in` loop would be fine here, but in general they're deprecated. Suggestion:
```suggestion
        const shellHeaders = definitions.veniaAppShell.inline.headers.inline;
        const securityHeaders = definitions[SECURITY_HEADER_DEFINITION].inline;

        for (const name of Object.keys(securityHeaders)) {
            shellHeaders[name] = `${SECURITY_HEADER_DEFINITION}.${name}`;
        }
```

packages/venia-concept/package.json
tjwiebell: Fixed to this constraint in 27e45b1

packages/pwa-buildpack/lib/cli/create-project.js
```diff
@@ -222,10 +222,11 @@ module.exports.handler = async function buildpackCli(argv) {
...
-        const monorepoDir = resolve(pkgDir.sync(__dirname), '../../');
+        const packagesDirectory = resolve(pkgDir.sync(__dirname), '../');
         prettyLogger.info(
-            execa.shellSync('rm -v packages/*/*.tgz', { cwd: monorepoDir })
-                .stdout
+            execa.shellSync('find . -name "*.tgz" -delete -print', {
+                cwd: packagesDirectory
+            }).stdout
         );
     }
```
zetlen, line 231: I don't know why I didn't think of this before, but...what if we just put tarballs in package roots in `.gitignore`? I just tested it and it worked fine. We could just remove this section in `create-project.js` entirely. Which would be nice, since now we're not looking for DEBUG_PROJECT_CREATION in two places.

Gitignore additions:
```gitignore
## May temporarily create tarballs during scaffolding debug
packages/*/*.tgz
packages/extensions/*/*.tgz
```

packages/venia-concept/_buildpack/create.js
```diff
@@ -133,7 +133,14 @@ function setDebugDependencies(fs, pkg) {
...
     const overridden = {};
     const workspaceDir = resolve(__dirname, '../../');
-    fs.readdirSync(workspaceDir).forEach(packageDir => {
+    const extensionDir = resolve(workspaceDir, 'extensions');
+
+    const packageDirs = [
+        ...fs.readdirSync(workspaceDir),
+        ...fs.readdirSync(extensionDir).map(path => `extensions/${path}`)
+    ];
+
```
zetlen, line 142: This works, but I think it would be a little more robust to just ask Yarn what the current packages are.

```js
    // Assume we are in the pwa-studio repo and get its root.
    const monorepoDir = resolve(__dirname, '../../../');

    // The Yarn "workspaces info" command outputs JSON as of v1.22.4.
    // The -s flag suppresses all other non-JSON logging output.
    const yarnWorkspaceInfoCmd = 'yarn -s workspaces info';
    let workspaceInfo = require('child_process').execSync(
        yarnWorkspaceInfoCmd,
        { cwd: monorepoDir }
    );

    let packageDirs;
    try {
        packageDirs = Object.values(JSON.parse(workspaceInfo)).map(
            ({ location }) => resolve(monorepoDir, location)
        );
    } catch (e) {
        // In case Yarn breaks this in the future, throw a more informative
        // error than your typical JSONError.
        throw new Error(
            `Could not parse output of '${yarnWorkspaceInfoCmd}:\n${workspaceInfo}`
        );
    }
```

packages/venia-concept/_buildpack/create.js
```diff
@@ -133,7 +133,14 @@ function setDebugDependencies(fs, pkg) {
...
+    packageDirs.forEach(packageDir => {
```
zetlen, line 143: This works, but I think it would be a little more robust to just ask Yarn what the current packages are.

```js
    // Assume we are in the pwa-studio repo and get its root.
    const monorepoDir = resolve(__dirname, '../../../');

    // The Yarn "workspaces info" command outputs JSON as of v1.22.4.
    // The -s flag suppresses all other non-JSON logging output.
    const yarnWorkspaceInfoCmd = 'yarn -s workspaces info';
    let workspaceInfo = require('child_process').execSync(
        yarnWorkspaceInfoCmd,
        { cwd: monorepoDir }
    );

    let packageDirs;
    try {
        packageDirs = Object.values(JSON.parse(workspaceInfo)).map(
            ({ location }) => resolve(monorepoDir, location)
        );
    } catch (e) {
        // In case Yarn breaks this in the future, throw a more informative
        // error than your typical JSONError.
        throw new Error(
            `Could not parse output of '${yarnWorkspaceInfoCmd}:\n${workspaceInfo}`
        );
    }
```

packages/venia-concept/_buildpack/create.js
```diff
@@ -133,7 +133,14 @@ function setDebugDependencies(fs, pkg) {
...
         const packagePath = resolve(workspaceDir, packageDir);
         if (!fs.statSync(packagePath).isDirectory()) {
             return;
```
zetlen, line 146: If you use the Yarn method in my other comment, you don't need to do this resolve and stat check. `packageDirs` will contain only absolute-path directories.

packages/pwa-buildpack/lib/cli/create-project.js
tjwiebell: Got rid of cleanup and added a gitignore rule in 9a0ce84

packages/venia-concept/_buildpack/create.js
tjwiebell: Much cleaner, implemented in 9a0ce84. This method additionally filters out workspaces without `package.json` files, so I was able to get rid of that check too.

packages/venia-concept/_buildpack/create.js
tjwiebell: Love it, made this change in 9a0ce84.

packages/venia-ui/upward.yml
jimbo: As long as WPT cares, we can include them.

dpatil-magento:
@tjwiebell WPT security score looks great on all configs (img_opt - onboard/backend, upward-php/js). Issue I'm seeing with braintree payment module - Infinite loading indicator with below console error - 

`Refused to frame 'https://assets.braintreegateway.com/' because it violates the following Content Security Policy directive: "child-src 'self'". Note that 'frame-src' was not explicitly set, so 'child-src' is used as a fallback.`

tjwiebell:
> @tjwiebell WPT security score looks great on all configs (img_opt - onboard/backend, upward-php/js). Issue I'm seeing with braintree payment module - Infinite loading indicator with below console error -

@dpatil-magento - Added an additional rule in 092cd5a that allows iframes to load from that braintree domain. Good catch!

dpatil-magento:
QA Approved.

PR #2511, opened 2020-06-24
Mini Cart V2 / Shopping Bag Footer section.

revanth0212:
## Description

Added footer section for the new Mini Cart / Shopping Bag component.

**Should be merged only after #2509** 

## Related Issue
Closes [PWA-601](https://jira.corp.magento.com/browse/PWA-601)

### Verification Stakeholders
@jimbo 
@soumya-ashok 
@schensley 

### Verification Steps

1. run `yarn watch:venia`.
2. Add an item to the cart and click the new cart button.
3. We should be able to see the footer section.

## Screenshots / Screen Captures (if appropriate)

image

## Checklist

- [ ] I have added tests to cover my changes, if necessary.
* **Should be merged only after #2509**

PR #2549, opened 2020-07-10
Mini Cart V2 - Product link for each product in the mini cart.

revanth0212:
## Description

Added product links to the product listings in the new mini cart.

Should be merged only after #2511 is merged.

## Related Issue
Closes [PWA-739](https://jira.corp.magento.com/browse/PWA-739)

### Verification Stakeholders
@jimbo 
@soumya-ashok 
@schensley 

### Verification Steps
1. Go to the deployed app and add an item to the cart.
2. Go to any other page and click on the item in the cart. It should take you to the product's PDP page.

## Checklist


packages/venia-ui/lib/components/MiniCart/ProductList/item.css
```diff
@@ -48,6 +54,7 @@
 }
 
 .editButton {
-    grid-column: 3 / span 1;
+    grid-column: 2 / span 1;
     grid-row: 1 / span 1;
+    min-width: 1rem;
```
revanth0212, line 59: I added this because when the delete icon is clicked, we remove the button because of which the second column in the grid becomes empty and the grid realigns. It looks like things are moving around. To avoid it, I have added a min-width but still, it moves around. Any help is appreciated.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2549.pwa-venia.com : LH Performance Expected 0.85 Actual 0.55, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2549.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2549.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/lib/components/MiniCart/ProductList/item.js
```diff
@@ -20,52 +21,61 @@ const Item = props => {
...
+                        value={prices.price.value}
+                    />
+                    {' ea.'}
+                </span>
+            </Link>
+            <button
+                onClick={removeItem}
+                type="button"
+                className={classes.editButton}
+                disabled={isDeleting}
```
revanth0212, line 71: Disabling delete button instead of removing it from the UI based on UX's review.

soumya-ashok:
@revanth0212 The links from the image and the product title work correctly but

- I would expect that if I clicked on a product listing with a certain configuration that when I went to the PDP, it would reflect that configuration. Currently it does not.

- The page transition from when the link/image is clicked is a bit janky because when the mini-cart opens/closes, the whole right column with the product config on the page shifts to the right/left

- The hover color on the "add to cart" button on the PDP is neither the teal nor the new blue - I know that's not part of this work but wanted to call it out

revanth0212:
> * I would expect that if I clicked on a product listing with a certain configuration that when I went to the PDP, it would reflect that configuration. Currently it does not.

I am not sure if that is possible. We do not have a way to invoke product configurations while loading the PDP page. @jimbo let me know if I am wrong about this.

> * The page transition from when the link/image is clicked is a bit janky because when the mini-cart opens/closes, the whole right column with the product config on the page shifts to the right/left

I am unable to reproduce that @soumya-ashok. I'll try to check with you on a call. Meanwhile, if this is something to do with the PDP page, I am certain it is not related to this PR. I haven't changed anything related to the PDP page here. If that is an issue, I'll create a new ticket to handle it.

> * The hover color on the "add to cart" button on the PDP is neither the teal nor the new blue - I know that's not part of this work but wanted to call it out

I'll create a new ticket and link this comment in there. Thanks for mentioning it.

revanth0212:
> * The page transition from when the link/image is clicked is a bit janky because when the mini-cart opens/closes, the whole right column with the product config on the page shifts to the right/left

I guess this is related to https://jira.corp.magento.com/browse/PWA-735

revanth0212:
> * The hover color on the "add to cart" button on the PDP is neither the teal nor the new blue - I know that's not part of this work but wanted to call it out

Here is the corresponding ticket: https://jira.corp.magento.com/browse/PWA-756

jimbo:
>     * I would expect that if I clicked on a product listing with a certain configuration that when I went to the PDP, it would reflect that configuration. Currently it does not.

I agree, this is important and should land before we ship this feature. We'll have to discuss how to implement it, though, since it could affect routing. I'll talk to Product and create a story.

>     * The page transition from when the link/image is clicked is a bit janky because when the mini-cart opens/closes, the whole right column with the product config on the page shifts to the right/left

True, but it's unrelated to this PR and we're already working on a solution.

>     * The hover color on the "add to cart" button on the PDP is neither the teal nor the new blue - I know that's not part of this work but wanted to call it out

True, but it's unrelated to this PR. I'll roll a fix into the next relevant PR.

soumya-ashok:
Product image and title links work as expected. UX approved.

sirugh:
It feels slightly odd to me that the entire product "box" area is a button that links to the PDP. There is a border applied when you tab to the element though. I know @soumya-ashok already gave her approval but I wanted to just voice my thoughts before it gets merged.

[Image from Gyazo](https://gyazo.com/436644cce0e9fbdca0f019f7eeb050e4)

revanth0212:
The latest merge commits have knocked out @jimbo's review. A re-review is needed.

revanth0212:
> It feels slightly odd to me that the entire product "box" area is a button that links to the PDP. There is a border applied when you tab to the element though. I know @soumya-ashok already gave her approval but I wanted to just voice my thoughts before it gets merged. FWIW I had the same feedback on your other PR, #2557
> 
> [Image from Gyazo](https://gyazo.com/436644cce0e9fbdca0f019f7eeb050e4)

I personally like what @sirugh mentioned, only make the image and the title clickable links. The main reason being, on the cart page we are doing something similar (#2557) but the problem is, the UX for the cart page is different from the mini cart. In the cart page, on the mobile view, the quantity component falls into he options section and if we made the options section clickable, that would make the quantity component a link the PDP page which is not expected. Either we change the UX on the cart page so we can have consistent UX for cart and the mini cart or, only make the image and the title clickable.

This might be a lot to understand, I will be happy to chime into a quick 5 min call to explain the issue and the solutions.

soumya-ashok:
@revanth0212 I'm not sure where this originated, but the original and continuing intent is only to have the image and title link to the PDP, not the entire listing. I have never said it should do that, and it certainly wasn't the case when I approved it.

soumya-ashok:
@revanth0212 - Only the product image and title are linking to the PDP now, with the rest of the text static. UX approved.

packages/venia-ui/lib/components/MiniCart/ProductList/item.js
```diff
@@ -20,38 +21,44 @@ const Item = props => {
...
+                className={classes.thumbnail}
+                to={itemLink}
+                onClick={closeMiniCart}
+            >
+                <Image
+                    alt={product.name}
+                    classes={{ root: classes.thumbnail }}
+                    width={100}
+                    resource={product.thumbnail.url}
+                />
```
sirugh, line 53: For some reason the border around the image got thicker in this PR vs develop:

[Image from Gyazo](https://gyazo.com/72842e45f2b3c2327860eecf18420d61)

packages/venia-ui/lib/components/MiniCart/ProductList/item.js
```diff
@@ -20,38 +21,44 @@ const Item = props => {
...
-            <span className={classes.name}>{product.name}</span>
-            {deleteButton}
+            <Link
+                className={classes.thumbnail}
+                to={itemLink}
+                onClick={closeMiniCart}
+            >
+                <Image
+                    alt={product.name}
+                    classes={{ root: classes.thumbnail }}
```
sirugh, line 50: Oh, it's because you are applying the thumbnail class to both the image and the link. Please fix this :)

packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
```diff
@@ -9,6 +9,8 @@ export const ProductListingFragment = gql`
                 id
                 name
                 sku
+                url_key
+                url_suffix
```
sirugh, line 13: Any specific reason why you're modifying the cart page's product list fragment in this PR? I would assume that would be done as part of #2557

revanth0212: That's because when a product is added to the cart, the mini-cart won't update if those 2 keys are missed in the `productListingFragment`. `productListingFragment` is fetched in return for the add to cart mutation as part of the 
`CartPageFragment` of `cartPageFragments.gql.js`.

We could have added the Mini cart fragment as a requested response as well, but we have decided against it previously in another PR. That is also the reason why we have added `thumbnail { url }` as part of the `productListingFragment` previously (because the mini cart needs it to be updated).

packages/venia-ui/lib/components/MiniCart/ProductList/item.js
revanth0212: Nice catch dude. Fixed in e41d4df. I have created a new CSS class called `thumbnailContainer` to give the developers more control over both the classes in the future.

packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
sirugh: > That's because when a product is added to the cart, the mini-cart won't update if those 2 keys are missed in the productListingFragment.

http://develop.pwa-venia.com works without this addition. I'm able to add an item from the PDP and see it in the mini cart. I can add more items and still see them update in the mini cart. I can also increase/decrement the quantity of cart items within the /cart page and still see the mini cart update.

Regardless, my point is that you are already adding it in #2557 in order to link from the PDP, so you don't need to do this here, right? It's just scope creeping (in a minor way). Not asking you to remove it necessarily.

revanth0212: At first, the mini cart was not updating if I did not have these keys as the return value, which is understandable. Hence I added them as part of this PR. They also happen to be needed as part of #2557 which was a parallel PR I was working on. These 2 happened un-related to each other. 

But now I think we don't need them anymore since 8 days ago, your PR #2505 was merged which requests the `MiniCartFragment` on the `addConfigurableProductToCart`. Since I never modified that file, it did not create a merge conflict. Nice catch dude. Removing them in this commit 53c89966135862476d8945bb4c8f838e02c2aab7.

You will understand why I had those 2 keys initially if you remove the `MiniCartFragment` and those 2 keys from the `productListingFragment` and try to add an item to the cart.

PR #2554, opened 2020-07-14
Hide mini cart on the checkout page.

revanth0212:
## Description

Hide the new mini cart on the checkout page.

## Related Issue
Closes [PWA-737](https://jira.corp.magento.com/browse/PWA-737)

### Verification Stakeholders
@soumya-ashok 
@schensley 
@jimbo 

### Verification Steps
1. Go to the checkout page on the deployed app. Cart trigger should not be visible.

image

image

## Checklist


packages/peregrine/lib/talons/Header/useCartTrigger.js
```diff
@@ -33,6 +33,7 @@ export const useCartTrigger = props => {
     const fetchCartDetails = useAwaitQuery(getCartDetailsQuery);
 
     const itemCount = data ? data.cart.total_quantity : 0;
+    const showCartTrigger = ['/checkout'].includes(history.location.pathname);
```
revanth0212, line 36: I could have gone for a simple checkout page check instead of the array, but after my discussion with UX and @jimbo, I have realized this is can be a configurable option hence made an array to represent routes on which the cart trigger should be hidden.

packages/peregrine/lib/talons/Header/useCartTrigger.js
```diff
@@ -33,6 +33,7 @@ export const useCartTrigger = props => {
     const fetchCartDetails = useAwaitQuery(getCartDetailsQuery);
 
     const itemCount = data ? data.cart.total_quantity : 0;
+    const hideCartTrigger = ['/checkout'].includes(history.location.pathname);
```
revanth0212, line 36: I could have gone for a simple checkout page check instead of the array, but after my discussion with UX and @jimbo, I have realized this is can be a configurable option hence made an array to represent routes on which the cart trigger should be hidden.

packages/venia-ui/lib/components/Header/cartTrigger.js
```diff
@@ -40,10 +41,10 @@ const CartTrigger = props => {
         <span className={classes.counter}>{itemCountDisplay}</span>
     ) : null;
 
-    // Because this button behaves differently on desktop and mobile
-    // we render two buttons that differ only in their click handler
-    // and control which one displays via CSS.
-    return (
+    const cartTrigger = hideCartTrigger ? null : (
```
revanth0212, line 44: Another way to hide the cart trigger is to implement this conditional in the `Header` component instead. But, the if this conditional is placed in the `Header`, the `CartTrigger` would never render in the `checkout` page and the GQL call in the `CartTrigger` will not be fired, due to which the `checkout` page will show an everlasting loading icon. The checkout page relies on the result of the `CartTrigger`'s GQL call to know whether to render an empty state or the `Checkout` component.

The other solution would be to move/add the GQL in the checkout page as well and then conditionally render `CartTrigger` in the `Header` component.

schensley:
@revanth0212 as we discussed, while there is some debate over whether or not to limit shopper access to cart in checkout, it has long been a common/best practice to do so, as having it has proven to have a negative affect on conversion. It'd be ideal in the reference site to follow this practice, allowing individual implementations to show it if they felt it was necessary.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2554.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2554.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2554.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 60

packages/peregrine/lib/talons/Header/useCartTrigger.js
sirugh: I'd probably leave this up to the parent component for reuseability.

packages/venia-ui/lib/components/Header/cartTrigger.js
```diff
@@ -18,7 +18,8 @@ const CartTrigger = props => {
         handleTriggerClick,
         itemCount,
         miniCartRef,
-        miniCartIsOpen
+        miniCartIsOpen,
+        hideCartTrigger
```
sirugh, line 22: What do you think about passing this as a component prop instead of returning it from the talon? Talon responses are less customizable than a prop would be.

Regardless, make sure to update the jsdoc for the component and talon while you're in here.

revanth0212: I don't have an opinion about placing it in a talon or sending it down as a prop.

Either way, the component should render but the trigger should be hidden conditionally. So, I thought it was closer to being a component logic than parent logic, hence added it into the talon.

About the docs, sure, thanks for reminding me, ill update them.

sirugh: For components out of a component library you'd expect a prop but this isn't quite the same. I lean towards that implementation as it feels more future proof but it's up to you (and @jimbo, I suppose).

packages/peregrine/lib/talons/Header/useCartTrigger.js
tjwiebell: Until we decouple cart creation, this would be tricky; this component needs to render. Someone refreshing from this page would never get their cart init'd; or someone signing in from the checkout page would also skip the init of their cart. Both edges, but definitely things Dev would find in QA 😂

sirugh: You're right. I guess until we move cart init out of cart trigger the component must always render (even if hidden).

revanth0212:
> Until we decouple cart creation, this would be tricky; this component needs to render. Someone refreshing from this page would never get their cart init'd; or someone signing in from the checkout page would also skip the init of their cart.

What do you mean by cart initialization? Cart creation you mean?

Cart trigger UI will not be rendered, but the talon itself will be. If you meant cart creation, then the cart creation mutation part of the `useCartTrigger` should be able to create a new cart ID when needed.

tjwiebell:
> What do you mean by cart initialization? Cart creation?
> 
> Cart trigger UI will not be rendered, but the talon itself will be. If you meant cart creation, then the cart creation mutation part of the `useCartTrigger` should be able to create a new cart ID when needed.

I don't believe that is a pattern we want to introduce, we don't have components that only contain logic. **If** we didn't have this coupling, I would expect `Header` to determine to render `CartTrigger` and not need to pass a prop down.

revanth0212:
> > What do you mean by cart initialization? Cart creation?
> > Cart trigger UI will not be rendered, but the talon itself will be. If you meant cart creation, then the cart creation mutation part of the `useCartTrigger` should be able to create a new cart ID when needed.
> 
> I don't believe that is a pattern we want to introduce, we don't have components that only contain logic. **If** we didn't have this coupling, I would expect `Header` to determine to render `CartTrigger` and not need to pass a prop down. eg. `const cartElement = shouldHideMiniCart ? null : <CartTrigger />`

I did mention something like this in a previous [comment](https://github.com/magento/pwa-studio/pull/2554#discussion_r454664479). If this is something we do not want to do, then it will be a bigger refactor. We will have to move the GQL queries that Cart Trigger fires into the components that might get affected due to it not being rendered on certain routes.

tjwiebell:
> I did mention something like this in a previous [comment](https://github.com/magento/pwa-studio/pull/2554#discussion_r454664479). If this is something we do not want to do, then it will be a bigger refactor. We will have to move the GQL queries that Cart Trigger fires into the components that might get affected due to it not being rendered on certain routes.

I think some context was lost here; my comment was a reply to Stephen defending your implementation; we're all good. Nice work dude 👍

sirugh:
Thanks for the discussion re: parent vs child logic. I approved as this does hide the mini cart trigger/icon on the checkout page. I still think it's odd to show it at all on the cart page since on mobile it is just a link to the cart page and on desktop it doesn't provide any additional benefit. But hey, that's not what the ticket called. Good work :)

dpatil-magento:
@revanth0212 On checkout page, every click logging console error, pls take a look.

image

dpatil-magento:
QA Approved.

packages/peregrine/lib/hooks/useDropdown.js
```diff
@@ -15,7 +15,9 @@ export const useDropdown = () => {
 
     // collapse on mousedown outside of this element
     const maybeCollapse = useCallback(({ target }) => {
-        if (!elementRef.current.contains(target)) {
+        const isValidElementRef = elementRef && elementRef.current;
+
+        if (isValidElementRef && !elementRef.current.contains(target)) {
```
tjwiebell, line 20: `useRef` always returns an object shape `{current: <value>}`, you should just need a truthy check of current since your initialValue is null.

```suggestion
        if (elementRef.current && !elementRef.current.contains(target)) {
```

If we wanted to be extra careful, you could even do a type check to ensure contains is in the prototype; `elementRef.current instanceof Node` (HTMLElement might be better?).

revanth0212: I think type check can be an overkill since it is only created and used by react/react components, the type should/will not change. Regarding the `current` check, I am doing that as part of `isValidElementRef`. Is there something that I am missing?

sirugh: > `useRef` always returns an object shape `{current: <value>}`

Tommy is saying that you don't need the `elementRef` part of the condition because it's unnecessary. And if you don't need the `elementRef` truthy check, assigning `elementRef.current` to a variable is just extra. That's his suggested change.

He adds that you can protect the condition against running on invalid ref assignments by verifying that the ref's current value is of the type expected.

PR #2572, opened 2020-07-22
fix: move cart creation out of trigger

sirugh:
## Description

For some reason the cart trigger owns cart creation. When the component mounts it invokes a "getCartDetails" action, which in turn checks for existence of a cart id and if it doesn't find one, calls the "createCart" action, which does another check for cart id and if it also doesn't find one, invokes a mutation to create a cart.

This could be simpler. We can move cart creation to the cart context provider.

## Related Issue

Closes [PWA-776](https://jira.corp.magento.com/browse/PWA-776).

### Verification Steps

1. Clear cache and refresh. A cart id should be fetched (via createCart) from the context provider instead of the trigger.
2. Log in. In the dev tools, modify your `signin_token` to invalidate it (remove a character).
3. Refresh the page. You should see errors logged in console from graphql about unauthorized access, but then you should see those queries be remade with a valid cart id and the user token should be removed.

## Checklist


packages/venia-ui/lib/components/CartPage/GiftCards/giftCardQueries.js
```diff
@@ -4,7 +4,7 @@ import { CartPageFragment } from '../cartPageFragments.gql';
 import { GiftCardFragment } from './giftCardFragments';
 
 export const GET_CART_GIFT_CARDS_QUERY = gql`
-    query getCartDetails($cartId: String!) {
+    query getGiftCardsInCart($cartId: String!) {
```
sirugh, line 7:
```suggestion
    query getGiftCardsOnCart($cartId: String!) {
```

?

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2572.pwa-venia.com : LH Performance Expected 0.85 Actual 0.52, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2572.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.35, LH Best Practices Expected 1 Actual 0.92
https://pr-2572.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/peregrine/lib/context/cart.js
```diff
@@ -72,3 +90,19 @@ export default connect(
...
+    mutation createCart {
+        cartId: createEmptyCart
+    }
+`;
+
+const CART_DETAILS_QUERY = gql`
+    query getCartDetails($cartId: String!) {
+        cart(cart_id: $cartId) @connection(key: "Cart") {
+            # The purpose of this query is to check that the user is authorized
+            # to query on the current cart. Just fetch "id" to keep it small.
```
jimbo, line 104: Good explanation.

packages/peregrine/lib/context/cart.js
```diff
@@ -72,3 +90,19 @@ export default connect(
 )(CartContextProvider);
 
 export const useCartContext = () => useContext(CartContext);
+
+const CREATE_CART_MUTATION = gql`
```
sirugh, line 94: Add comments about the exception to the rule which is that we are including gql in peregrine here.

packages/peregrine/lib/context/cart.js
```diff
@@ -52,6 +55,21 @@ const CartContextProvider = props => {
         derivedCartState
     ]);
 
+    const apolloClient = useApolloClient();
+    const [fetchCartId] = useMutation(CREATE_CART_MUTATION);
+    const fetchCartDetails = useAwaitQuery(CART_DETAILS_QUERY);
+
+    useEffect(() => {
+        // This call initializes the cart if there isn't one. Also, we pass
```
revanth0212, line 63: I was wonder where is the if check for `cartId`. Aparently it is in the `getCartDetails` function. May you can change the wording of the comments to mention that.

```suggestion
        // cartApi.getCartDetails initializes the cart if there isn't one. Also, we pass
```

revanth0212:
I can see that the error message is being logged, kudos to that.

image

Followup question, should we change the `errorLink` function in `venia-concept/src/index.js` to better represent `locations` object instead of [object Object].

packages/peregrine/package.json
```diff
@@ -32,6 +32,7 @@
     "@apollo/react-hooks": "~3.1.2",
     "@babel/runtime": "~7.4.2",
     "apollo-client": "~2.6.4",
+    "graphql-tag": "~2.10.4",
```
jimbo, line 35: Yes, I meant to bring this up when I mentioned that this was the first query in `peregrine`. 💯

PR #2573, opened 2020-07-23
fix: product/category root component prefetch

sirugh:
## Description

The Category/PDP had some "magic" wherein we prefetch the product data for the rendered products up front and then used a fragment matcher to manually pull from the cache on the PDP. To do this we made a "big" query up front for category + product data. 

This PR moves the prefetch logic to the gallery items which makes a prefetch query when mounted. This results in a smaller initial category request but N more requests for product data per gallery item.

I also updated breadcrumbs to follow the new pattern and to use cache-and-network.

## Related Issue

Closes [PWA-778](https://jira.corp.magento.com/browse/PWA-778)
### Verification Steps

1. Go to the category page. Ensure that you can see the product images/prices/titles and that you can click to go to the product page.
2. Go to the product page. Ensure that nothing is broken.
2. Check network requests - you should see one for product details and one for breadcrumbs (this is the "network" part from "cache-and-network").

[Image from Gyazo](https://gyazo.com/42ed6dc88fc185b6103e659e0b773376)

## Checklist


packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
```diff
@@ -28,9 +28,13 @@ const getPath = (path, suffix) => {
...
 export const useBreadcrumbs = props => {
-    const { categoryId, query } = props;
+    const {
+        categoryId,
+        queries: { getBreadcrumbsQuery }
+    } = props;
 
-    const { data, loading, error } = useQuery(query, {
+    const { data, loading, error } = useQuery(getBreadcrumbsQuery, {
+        fetchPolicy: 'cache-and-network',
```
sirugh, line 37: I've been thinking about our use of the default fetch policy lately. `cache-first` is appropriate when your data will rarely change, or when you know that you will be clearing said data sooner than later. I feel like we should change our default policy to `cache-and-network` as we always want to display data as quickly as possible (cache) but we also want to make sure that data is updated incase something changed on the server (network).

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -8,70 +8,27 @@ import { useApolloClient, useQuery } from '@apollo/react-hooks';
...
-     * If we do not have a cached product, we need to go fetch it from the network.
-     * If we do have a cached product, we want to ensure its cache entry doesn't get stale.
-     * In both cases, Apollo will update the cache with the latest data when this returns.
-     */
-    const { loading, error, data } = useQuery(query, {
-        // Once we're able to remove the manual cache lookup,
-        // this fetch policy can change to 'cache-and-network'.
-        fetchPolicy: 'network-only',
+    const { loading, error, data } = useQuery(queries.getProductDetailQuery, {
+        fetchPolicy: 'cache-and-network',
```
sirugh, line 24: To quickly verify that ApolloClient is indeed doing its job, remove this line and load a product page without any cache. You should see a network request for product details. If you refresh the page again you should _not_ see a network request for product details. This is because Apollo defaults to `cache-first` which means if all data requested is in the cache, it doesn't make a network call.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2573.pwa-venia.com : LH Performance Expected 0.85 Actual 0.51, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2573.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.33, LH Best Practices Expected 1 Actual 0.92
https://pr-2573.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/lib/RootComponents/Category/category.gql.js
```diff
@@ -0,0 +1,58 @@
...
+                    label
+                    position
+                    disabled
+                    file
+                }
+                meta_description
+                small_image {
+                    url
+                }
+                url_key
```
sirugh, line 49: This sucks :/ I thought switching to a fragment would prevent the extra fetch that is happening in the PDP (because we pre-fetch?) but it's still occurring even if I remove the `cache-and-network` policy. Probably something to do with the noted bug? But it's been closed and the repo archived so for now we get a comment and some hard coded fields.

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -8,70 +8,27 @@ import { useApolloClient, useQuery } from '@apollo/react-hooks';
  * @kind function
  *
  * @param {object}      props
- * @param {String}      props.cachePrefix - The prefix to apply to the cache key.
- * @param {GraphQLAST}  props.fragment - The GraphQL fragment to match against a cache entry.
  * @param {Function}    props.mapProduct - A function for updating products to the proper shape.
  * @param {GraphQLAST}  props.query - The query to fetch a product.
```
supernova-at, line 12:
```suggestion
 * @param {GraphQLAST}  props.queries - Queries for this component
```

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
supernova-at: I feel like I actually tried that at one point but couldn't get it to take. There is definitely [a configuration option](https://www.apollographql.com/docs/react/api/core/ApolloClient/#example-defaultoptions-object) on the client itself to set this.

packages/venia-ui/lib/RootComponents/Category/category.gql.js
supernova-at: Yeah, agreed - I was going to comment here too. Not sure why we get an extra fetch though, the whole idea of the PR that introduced all this (#1935) was that the category page would pre-fetch all the details the product pages needed.

supernova-at:
Added `Version: Major` label since some talon APIs changed.

packages/venia-ui/lib/RootComponents/Category/category.gql.js
sirugh: My concern is that the data fetched by the category query is not being "seen" by the product query when it checks the cache. This could be related to a misconfiguration or a bug in apollo client, so I'll investigate.

As far as the bug I referenced, it seems like it'll get fixed in 2.4.1 which is good but doesn't help us now :/

sirugh: After discussion with @supernova-at we realized that the original fragment reads were intentional because otherwise apollo client doesn't know how to match the PDP query and the category query. For example, when the category query prefetches products, it stores them in a cache entry keyed with the args to the query. When the PDP loads it checks for a different cache key because all we have at that time is the url_key. What results is a cache miss, so we make the network call.

We have options:

a) See if we can somehow get the category to prefetch a query that is keyed by url_key, so that the PDP will get a cache hit.

b) Remove the prefetch from the category page.

Both are fine, IMO, and I'll see if I can get `a` to work first before removing the category prefetching entirely.

sirugh: "a" did not work but I did move prefetching to the gallery item. May be overstepping the responsibility of the gallery item component tbh.

packages/venia-ui/lib/RootComponents/Category/category.gql.js
```diff
@@ -0,0 +1,51 @@
...
+        products(
+            pageSize: $pageSize
+            currentPage: $currentPage
+            filter: $filters
+            sort: $sort
+        ) {
+            items {
+                # id is always required, even if the fragment includes it
+                id
+                # The following values are used by GalleryItem
```
supernova-at, line 30: Can we have `GalleryItem` export a fragment and plug that in here?

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param {Object} props.queries - the queries for the gallery item
+ */
+export const useGalleryItem = props => {
+    const { item, queries } = props;
+
+    const { prefetchProductQuery } = queries;
+    const { url_key } = item;
+
+    // Prefetch the underlying product for this gallery item.
+    const [runPrefetchQuery] = useLazyQuery(prefetchProductQuery);
```
supernova-at, line 17: 💯 

I like this approach because it re-uses the `GET_PRODUCT_DETAIL_QUERY` and doesn't delay the rendering of the `GalleryItem`.

Another thought is to have the Service Worker do it, but then we have to keep those GraphQL fragments / queries in sync and we don't have a system for that right now.

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
```diff
@@ -0,0 +1,28 @@
...
+    useEffect(() => {
+        if (item) {
+            runPrefetchQuery({
```
supernova-at, line 20: It hurts because we potentially already have this product data elsewhere in the cache but this is more developer-friendly / less magic.

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from 'react';
+import { useLazyQuery } from '@apollo/react-hooks';
+
+/**
+ *
+ * @param {Object} props
+ * @param {Object} props.item - the product item being rendered
+ * @param {Object} props.queries - the queries for the gallery item
+ */
+export const useGalleryItem = props => {
```
supernova-at, line 10: This talon is rare in that it doesn't return anything (returns `undefined`). Consider documenting that in the JSDoc comments.

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
sirugh: We could use `cache-first` here which would return the cached data for the query if it exists and only fall back to network if it doesn't.

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
sirugh: eac357918

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
sirugh: I added cache-first in c51624861 but I'd like to talk through it in team time.

packages/venia-ui/lib/RootComponents/Category/category.gql.js
sirugh: Created a fragment, but unfortunately a bug is preventing us from using it. Added a TODO/comment in eb23ac901

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -8,70 +8,27 @@ import { useApolloClient, useQuery } from '@apollo/react-hooks';
...
+ * @param {GraphQLAST}  props.queries - Queries for this component
  * @param {String}      props.urlKey - The url_key of this product.
  *
  * @returns {object}    result
- * @returns {Bool}      result.error - Indicates a network error occurred.
- * @returns {Bool}      result.loading - Indicates the query is in flight.
- * @returns {Bool}      result.product - The product's details.
+ * @returns {object}    result.error - Indicates a network error occurred.
+ * @returns {boolean}   result.loading - Indicates the query is in flight.
+ * @returns {objct}     result.product - The product's details.
```
sirugh, line 18:
```suggestion
 * @returns {object}     result.product - The product's details.
```

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
```diff
@@ -18,17 +18,48 @@ export const useGalleryItem = props => {
...
-                    urlKey: url_key,
-                    onServer: false
-                },
-                fetchPolicy: 'cache-first'
+    const galleryItemRef = useRef();
+
+    const handleScrollIntoView = useCallback(
+        entries => {
+            entries.forEach(entry => {
+                if (entry.isIntersecting) {
```
sirugh, line 29: Test this by adding items beyond the visible viewport. I did it by using a mobile view in dev tools and increased the gap between the gallery rows.

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
```diff
@@ -0,0 +1,69 @@
...
+        },
+        [runPrefetchQuery, url_key]
+    );
+
+    const observer = useMemo(
+        () =>
+            new IntersectionObserver(handleScrollIntoView, {
+                // `isIntersecting` will be true if 50% of the referenced node
+                // is visible.
+                threshold: 0.5
```
sirugh, line 50: [Image from Gyazo](https://gyazo.com/42ed6dc88fc185b6103e659e0b773376)

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
jimbo: Makes sense to me. Maybe this clause should avoid executing more than once for the same element?

```es6
if (!hasPrefetched && entry.isIntersecting)
```

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
sirugh: Solved through use of IntersectionObserver

packages/venia-ui/lib/RootComponents/Category/category.gql.js
sirugh: Resolved through prefetching with IntersectionObserver in galleryItem

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param {Object} props
+ * @param {Object} props.item - the product item being rendered
+ * @param {Object} props.queries - the queries for the gallery item
+ *
+ * @returns {{
+ *  ref: the ref for the gallery item, used by the intersection observer
+ * }}
+ */
+export const useGalleryItem = props => {
+    const { item, queries } = props;
```
sirugh, line 15: @larsroettig what do you think about a `shouldPrefetch` here?

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
```diff
@@ -0,0 +1,71 @@
...
+                if (entry.isIntersecting && !hasPrefetched) {
+                    setHasPrefetched(true);
+                    runPrefetchQuery({
+                        variables: {
+                            urlKey: url_key,
+                            onServer: false
+                        },
+                        // If we get a cache-hit, the benefit of the prefetch
+                        // is achieved already, so no need to hit network.
+                        fetchPolicy: 'cache-first'
```
sirugh, line 39: As @tjwiebell mentioned, there is possibility to manage this with edge caching to lighten the server load, but that may not be possible for everyone.

I can add a `shouldPrefetch` boolean, check "time-in-viewport", or try other optimizations, but at the end of the day, from a cold cache, this code still has the possibility to emit a large number of queries at once. 

I would love to talk about options outside of reverting the PR, as I believe this code is more explicit and provides a better developer experience.

mtbottens: I'm curious what people think about prefetching on a more specific user interaction. Such as hover/focus/click/tab on a given CTA. 

Pros:
* Still serves the purpose of preloading, but is more specific and should cause less strain on the application server
* For requests that are a cache hit, this shouldn't be too much of a difference than the current approach

Cons:
* For requests that are a cache miss, this would cause a more noticeable lag for the user, when compared with the current approach.
* Won't work as well for mobile users

sirugh: Now that you mention it, I think using on hover/focus might be a good optimization here. I'm also questioning whether mobile prefetch is even necessary. I like to be data conscious for mobile users and disabling prefetch for mobile (since there is no hover/focus) might just be fine.

sirugh:
After reviewing this with @jimbo and hearing the feedback from the community, we've decided to close this fix-turned-spike PR. Thanks to all for the discussions about prefetching - it has been enlightening.

Outcomes: 

1) To render a product page we make two graphql calls: first we resolve the url to get the root component type, and then we fetch the product data.

2) To adequately prefetch this information we have to make individual queries at some point in the app's lifecycle. In this spike we went with making those calls when 50% of a product's gallery item is in the viewport on the category page. When the product image comes into view, two network requests are fired - url resolver and product details. Ideally these would be batched along with other queries made near the same time. All of this happens behind the scenes and without any detriment to the user's experience.

3) Unfortunately, due to server constraints, batching is not currently possible and multiple synchronous requests tend to cause performance issues. For these reasons, this approach to prefetching simply will not work.

4) The current prefetching that occurs in the category graphql call is still extremely bloated - when we load a category page we fetch the product data for _all_ the products visible on the page. On the reference storefront it's only 6 products and about 4.5kb, but you can see how that would not scale.

5) We should have a discussion about whether the current approach to prefetching is acceptable and/or necessary. https://jira.corp.magento.com/browse/PWA-822

6) Leaning on ApolloClient/useQuery to make the url resolver queries seems to be a positive, could potentially do this in the future. https://jira.corp.magento.com/browse/PWA-821

PR #2576, opened 2020-07-24
[PWA-746] [bug]: Can't add product to cart if another out of stock product is in the cart

### LINKED ISSUE #2536: [bug]: Can't add product to cart if another out of stock product is in the cart
**Describe the bug**
If I add a product to the cart, and then this products goes out of stock, I cannot add more products to cart, no error message is shown but the api returns this error `{"errors":[{"message":"Some of the products are out of stock.","extensions":{"category":"graphql-input"}}],"data":{"addSimpleProductsToCart":null}}`

**To reproduce**
1.Add product to cart
2. Set the inventory for this product to 0 (no backorders)
3. Try to add another product to cart
4. No product is added to cart and graphql query returns that error above

**Expected behavior**
Product is added to cart, and error for previous product is shown

 - Magento Version: 2.3.5
 - PWA Studio Version: 6.0.1 (development 0837e212a81312473f365cd17d4ff7530ed2ce15)

tjwiebell:
## Description

See #2536

## Related Issue

* Closes #2536 
* [[PWA-746](https://jira.corp.magento.com/browse/PWA-746)] [bug]: Can't add product to cart if another out of stock product is in the cart

### Verification Steps

You'll need access to a backend that you can set a product to out of stock. The setup to all of these steps is to have two simple products in your cart.

**Initial Bug**
1. In the backend, flip a product to OOS
2. Refresh page, verify you can still interact with MiniCart w/o error
3. Navigate to a product that's in stock, verify you can add that product to your cart without error

**MiniCart**
1. Open MiniCart and verify no stock status message is displayed
2. In the backend, flip a product to OOS
3. Refresh page, open MiniCart, verify that you now see stock status messages

**CartPage**
1. Open CartPage and verify no stock status message is displayed
2. In the backend, flip a product to OOS
3. Refresh page, verify that you now see stock status messages

**CheckoutPage**
1. Open CheckoutPage and verify no stock status message is displayed
2. In the backend, flip a product to OOS
3. Refresh page, verify that you now see stock status messages with link back to CartPage

_note:_ Current guidance doesn't have us preventing user actions right now, so if a user ignores the warnings and proceed to Checkout, we expect them to just get stalled.

**MiniCart**
OOS-miniCart

**CartPage**
OOS-cartPage

**CheckoutPage**
OOS-checkoutPage

## Checklist


packages/venia-concept/src/index.js
```diff
@@ -39,14 +40,38 @@ const authLink = setContext((_, { headers }) => {
...
         );
     if (networkError) console.log(`[Network error]: ${networkError}`);
+
+    const { data, errors } = response;
+
+    errors.forEach(({ message, path }, index) => {
+        // It's within the GraphQL spec to receive data and errors, where errors are merely informational and not
+        // intended to block. Almost all existing components were not built with this in mind, so we build special
+        // handling of this error message so we can deal with it at the time we deem appropriate.
+        if (message === 'Some of the products are out of stock.') {
```
sirugh, line 58: This makes the third file that checks against the error text. I think it's probably time to make a mapping, probably in venia-ui or peregrine, that way we can easily keep track of these strings. It'll make it easier to update in the future too, if/when they change.

https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/store/actions/cart/asyncActions.js#L415-L426

https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.js#L80-L108

packages/venia-concept/src/index.js
```diff
@@ -39,14 +40,38 @@ const authLink = setContext((_, { headers }) => {
...
+            const pathToCartItems = path.slice(0, -1);
+            const cartItems = getWithPath(data, pathToCartItems);
+
+            // Until MC-36092 is resolved, we need to guard against a null product being returned as well.
+            if (cartItems[0] === null) {
```
sirugh, line 63: Is it always the first item or is it possible it could just exist somewhere in the array? May be more future proof to do this in a loop, at least until the noted fix is complete.

packages/venia-concept/src/index.js
```diff
@@ -39,14 +40,38 @@ const authLink = setContext((_, { headers }) => {
...
+                cartItems.splice(0, 1);
+            }
+
+            // Suppress entirely if this is our only error, we don't want to cause operations to throw.
+            if (response.errors.length === 1) {
+                response.errors = undefined;
+            } else {
+                response.errors[index] = undefined;
+            }
```
sirugh, line 72: If I'm reading this correctly it looks like we now delete errors from responses. What affect does this have on components that were previously handling these errors? Do the invokers still throw? Does this just prevent `error` from being a property on the response ie `const { error } = useQuery(...)`?

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2576.pwa-venia.com : LH Performance Expected 0.85 Actual 0.5, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2576.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.34, LH Best Practices Expected 1 Actual 0.92
https://pr-2576.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-concept/src/index.js
tjwiebell: Regardless of the position of the OOS item, or even if there are multiple OOS items, `null` always shows up at the first index. Whatever the bug is, at least it's consistent.

A loop seems overkill, but I'm happy to use the index sliced out of the path that seems to reference the `null` item. GraphQL **is** fixing this in 2.4.1, so future proofing shouldn't really be necessary.

packages/venia-concept/src/index.js
tjwiebell: I feel like @supernova-at did this in a PR, and we heard some arguments for it being unnecessary. I agree referencing `MagentoGraphQLErrorMap.get('OUT_OF_STOCK')` might look cleaner, so happy to implement if you refresh my memory and help decide which package it should go in 🙂

packages/venia-concept/src/index.js
tjwiebell: Both! `response.errors.length > 0` will cause invokers to throw and the population of `error` in responses. PM wants to emulate Luma in this scope, where we don't force resolution of the error until necessary (checkout). There is an Epic planned where we might do a more automated approach to just auto-remove and pop a toast, but keeping it simple for now.

packages/venia-concept/src/index.js
sirugh: Oh I just meant a literal constants file. As peregrine uses strings, and peregrine does not import from venia-ui, it should live in peregrine :)

```js
// peregrine/utils/GraphqlConstants.js
export const SOME_PRODUCTS_OUT_OF_STOCK = 'Some of the products are out of stock.';
```

packages/venia-concept/src/index.js
sirugh: So because we are not slicing it the error still gets populated?

packages/venia-concept/src/index.js
DrewML: There's an [open architecture proposal](https://github.com/magento/architecture/pull/393) to try and make this stuff a bit less painful.

Would it be easier for y'all if you got back codes for various errors, so you could match on those instead of a string? That way the backend could make the message clearer later without breaking pwa-studio.

packages/peregrine/lib/talons/CartPage/useCartPage.js
```diff
@@ -36,9 +36,24 @@ export const useCartPage = props => {
     const hasItems = !!(data && data.cart.total_quantity);
     const shouldShowLoadingIndicator = called && loading && !hasItems;
 
+    const hasOutOfStockItem = useMemo(() => {
```
tjwiebell, line 39: I realize this is duplicated three times and we should probably just move to a `StockStatusMessage` component. Waiting on UX to finalize design, and will re-factor this duplication out before sending to code review.

schensley:
Thank you @tjwiebell 
This looks good, where the top level message has the subtle, but noticeable 'high light', (better than a potentially misunderstood icon).  Additionally, providing a link back to Cart in the message that appears on the Checkout page is a great help for the shopper.
UX Approved.

packages/venia-concept/src/index.js
supernova-at: If there are multiple OOS items are there multiple `null` products, or is there always just one?

tjwiebell: Correct, always just one.

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -27,9 +27,10 @@
     grid-area: details;
     grid-gap: 0.5rem;
     grid-template-areas:
-        'name .'
+        'name quantity'
```
jimbo, line 30: I believe this is changing in #2577 to reflect a bit of overlooked UX intent.

```suggestion
        'name name'
```

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -85,16 +84,21 @@
...
     .details {
-        grid-template-areas:
-            'name'
-            'options'
-            'price'
-            'stock'
-            'quantity';
-        grid-template-columns: auto;
-        grid-template-rows: auto;
+        grid-template: none;
```
jimbo, line 93: This resets `grid-template-areas`, `grid-template-columns`, and `grid-template-rows` in one line. As a result, each grid item creates a new row (when its `grid-area` isn't set).

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -85,16 +84,21 @@
...
+    }
+
+    .name,
+    .options,
+    .price,
+    .stockStatusMessage,
+    .quantity {
+        grid-area: auto;
```
jimbo, line 101: This resets `grid-area` for each grid item, so each one creates a new row.

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -84,13 +87,17 @@
...
+            'quantity';
+        grid-template-columns: auto;
+        grid-template-rows: auto;
     }
 
     .quantity {
-        grid-column: 1;
-        grid-row: 5;
-        justify-self: left;
+        justify-items: start;
```
jimbo, line 105: The value you were looking for here is `start`.

packages/peregrine/lib/talons/CartPage/useCartPage.js
```diff
@@ -36,7 +36,12 @@ export const useCartPage = props => {
     const hasItems = !!(data && data.cart.total_quantity);
     const shouldShowLoadingIndicator = called && loading && !hasItems;
 
+    const cartItems = useMemo(() => {
+        return data && data.cart.items;
```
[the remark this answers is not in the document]
tjwiebell: Changed in 386d55c.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -84,6 +84,10 @@ export const useCheckoutPage = props => {
         }
     });
 
+    const cartItems = useMemo(() => {
+        return checkoutData && checkoutData.cart.items;
```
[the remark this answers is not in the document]
tjwiebell: Changed in 386d55c.

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -81,7 +90,7 @@
 
     .quantity {
         grid-column: 1;
-        grid-row: 4;
+        grid-row: 5;
```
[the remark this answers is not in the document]
tjwiebell: Fixed in 141f6ad.

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
tjwiebell: Reset back to default, I like it. Do we have a strong preference to avoid named template areas?

jimbo: Yes, we prefer sensible implicit grids over explicit grids in general, since they more readily accommodate new (unanticipated) grid items.

dpatil-magento:
@tjwiebell I can still reproduce original issue where I can add product to cart for **partial out of stock product**.  
Scenario - Created a simple product with 20 qty and added 20 qty to cart in venia browser A. Now place an order with say 1 or 2 qty from some other browser B.  Refresh the browser A, toast gets triggered about entire qty not available for partial OOS product in cart. But now try add any other product.
Issue - Product does not get added.

dpatil-magento:
> @tjwiebell I can still reproduce original issue where I can NOT add product to cart for **partial out of stock product**.
> Scenario - Created a simple product with 20 qty and added 20 qty to cart in venia browser A. Now place an order with say 1 or 2 qty from some other browser B. Refresh the browser A, toast gets triggered about entire qty not available for partial OOS product in cart. But now try add any other product.
> Issue - Product does not get added.

Creates issue in backlog to handle scenario's around partial OOS.

@tjwiebell Simple product OOS flow looks good but if cart has OOS configurable product and then user either reloads browser or try adding new product or remove product >> Creating infinite error toasts .

image

tjwiebell:
@dpatil-magento I've solved the infinite looping bug and better handle the additional bug introduced by configurable products that are not completely out of stock. There is a known issue where this configurable product is not triggering the UX treatment, but @awilcoxa would prefer we defer fixing that scope to the upcoming Inventory Epic ([PWA-792](https://jira.corp.magento.com/browse/PWA-792). I've updated the description of that Epic to capture this issue.

PR #2577, opened 2020-07-24
Added CSS to handle multiple configurable options.

revanth0212:
## Description

Added CSS to handle multiple configurable options. 

**_Note: I took the easy route and manipulated the `configurable_options` to add more options at runtime instead of going to the backend and getting a product to test on. For UX and QA, @dpatil-magento do you know if we have an item in the current backend to test against?_**

## Related Issue
Closes [PWA-687](https://jira.corp.magento.com/browse/PWA-687)

### Verification Stakeholders
@soumya-ashok 
@schensley 
@jimbo 

### Verification Steps
1. Go to the deployed app and add an item with more than 2 options.

image

## Checklist


revanth0212:
I had conversations with UX about what is expected. According to them, the expectation is for the listings to display in their entirety with no scroll bar. I did the same, but since we don't have mockups for this use case I am not sure if this is what the UX wants. Let's discuss if changes are needed 👍 .

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2577.pwa-venia.com : LH Performance Expected 0.85 Actual 0.51, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 83.333333333333
https://pr-2577.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.33, LH Best Practices Expected 1 Actual 0.92
https://pr-2577.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/venia-ui/lib/components/LegacyMiniCart/productOptions.js
```diff
@@ -15,11 +15,15 @@ const ProductOptions = props => {
                 const key = `${option_label}${value_label}`;
 
                 return (
-                    <Fragment key={key}>
-                        <dt className={classes.optionLabel}>
-                            {option_label} : {value_label}
-                        </dt>
-                    </Fragment>
+                    <div key={key} className={classes.optionLabel}>
```
revanth0212, line 18: Refactored this piece of code because the correct way to represent a `<dl />` tag is to have a `<dt />` and a `<dd />` tag. Our old implementation only has the `<dt />` tag.

Had to wrap both of them in a `div` instead of a `Fragment` because the fragment component does not honor the CSS given to it. Used `div` because it is a valid child of the `dl` tag.

packages/venia-ui/lib/components/LegacyMiniCart/productOptions.js
```diff
@@ -15,11 +15,15 @@ const ProductOptions = props => {
...
+                        {/**
+                         * Added `\u00a0` to simulate a space character.
+                         * We can't use ` ` because JS string lirerals logic
+                         * does not allow leading or trailing space chaacters.
+                         */}
+                        <dt>{`${option_label} :\u00a0`}</dt>
```
revanth0212, line 24: Not sure if this is the right way to handle it, open for suggestions.

jimbo: Consider using CSS (`margin`, `padding`, `gap`) on the `dd` element?

revanth0212: I did, but the gap has to be exactly `1 space` to match the UX. I even tried `<dt>{option_label + ' : '}</dt>`, it did not work. The trailing space was removed.

jimbo: > I did, but the gap has to be exactly `1 space` to match the UX. I was not able to nail down the exact number of a space character.

We don't need to match designs to the pixel. Just approximate it. 😅 

> Any suggestions on what would be the size of a space character?

For Venia in general, I work in increments of `0.25rem` when estimating spacing. That should work fine here—make sure to add classnames to the `dt` and `dd` elements.

```
// structure
div.optionLabel
  dt.optionName
  dd.optionValue

// css
.optionValue
  margin-left: 0.25rem
```

revanth0212: > We don't need to match designs to the pixel. Just approximate it. 😅

image

I had to in this case because there are other options that don't fall into `ProductOptions` like quantity and if we can't nail the space correctly, they will look misaligned. 

That said, nice suggestion about `0.25rem`, that exactly 4px and works like a gem 👍 .

packages/venia-ui/lib/components/LegacyMiniCart/productOptions.css
```diff
@@ -12,3 +12,10 @@
     grid-auto-flow: column;
     grid-auto-columns: max-content;
 }
+
+.optionName {
+}
```
jimbo, line 17: You can just leave these out of the CSS file if they're empty.

revanth0212: I have seen empty CSS previously, so I thought we allow them. Cool will remove it.

sirugh: In the past we have used certain empty rules to allow for the possibility of customization but I don't think they need to be defined in the CSS file. I think @jimbo was just suggesting you remove these from CSS but keep the `className` prop on the components.

soumya-ashok:
@revanth0212 The listings in the video you attached look good, UX approved.

dpatil-magento:
Verified with real product options. QA Approved.

image


@revanth0212 Adobe CLA check has not passed, can you check?

PR #2588, opened 2020-07-29
PWA-718: FormError

sirugh:
## Description

As a shopper, I would like to see helpful and consistent messaging when I encounter an error in the app. This will allow me to know how to recover from those errors if possible, and ideally lead to fewer abandoned carts and support requests sent to the merchant.

In scope of PWA-336, we identified areas in CartPage and CheckoutPage that were lacking error handling, and consolidated the various way we were handling errors in the app. In reviewing this with UX, Product, and Team Lead, we identified some consistent patterns we would like to start enforcing in the app. In review of this task, we also discovered some bugs in how we were parsing errors from GraphQL that will also need to be resolved (should happen organically with addressing the following).

- ~~Implement FormError throughout the app~~
- ~~Update UX for FormError to new mocks provided by Scott Hensley~~
- ~~Implement utility function/hook that will handle parsing GraphQL Errors~~
- ~~Be sure all components rendering errors not using FormError use this utility (should really only be Toasts)~~
- Given that GraphQL errors are not very user friendly, we need to come up with an ErrorMapper to render better errors (out of scope?)
- ~~Consistently handle/don't handle re-throwing these errors. We have a lot of places where we catch and console.error, which doesn't really accomplish anything, since Apollo has its own error boundary.~~
- ~~When writing tests for FormError, it proved difficult to test the scrollIntoView effect due to jsdom limitations. Stephen Rugh found this article that outlines a reasonable alternative to consider that makes this testable, and giving us another custom hook to add. Worth discussing with team, but its an option so we can have coverage on that features outside of E2E testing.~~ Added in 362e57c0878119202bf5eaabc3462baff768c7fd
 
## Related Issue

Closes [PWA-718](https://jira.corp.magento.com/browse/PWA-718).

### Verification Steps

1. For each form in the app, verify that there is consistent error handling.
2. Verify nothing is broken :D

**Form Error w/o left border, below heading of form**

[Image from Gyazo](https://gyazo.com/a3e4c6d6280cac1f66410df53a37b138)

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2588.pwa-venia.com : LH Performance Expected 0.85 Actual 0.34, LH Accessibility Expected 1 Actual 0.97, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 39.333333333333
https://pr-2588.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.35, LH Best Practices Expected 1 Actual 0.92
https://pr-2588.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/peregrine/lib/util/deriveErrorMessage.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * A function to derive an error string from an array of errors.
+ */
+export const deriveErrorMessage = errors => {
+    const errorCollection = errors.filter(Boolean).map(error => {
+        const { graphQLErrors, message } = error;
+
+        return graphQLErrors && graphQLErrors.length
+            ? graphQLErrors.map(({ message }) => message).join(', ')
+            : message;
```
sirugh, line 10: This would be a great place to map from Magento GQL Error string to a more user friendly string.

schensley:
@sirugh Thanks for the work Stephen. Based on our discussion and the screen shot above, this PR meets UX approval.

packages/venia-ui/lib/components/FormError/formError.css
```diff
@@ -1,3 +1,5 @@
 .errorMessage {
     color: rgb(var(--venia-global-color-error));
+    font-size: var(--venia-typography-body-S-fontSize);
+    font-weight: var(--venia-global-fontWeight-semibold);
```
jimbo, line 4: I have reservations about this but I know the design calls for it.

jimbo:
It's great to see this PR. There's a lot of messy and incomplete code that's cleaned up here, so this will be a big improvement. 👍

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
```diff
@@ -68,18 +74,12 @@ export const useMiniCart = props => {
...
-                await removeItem({
-                    variables: {
-                        cartId,
-                        itemId: id
-                    }
-                });
-            } catch (err) {
-                if (process.env.NODE_ENV !== 'production') {
-                    console.error('Cart Item Removal Error', err);
+            await removeItem({
```
tjwiebell, line 77: Even though our ErrorLink is awesome and auto logs, we still have to handle these mutation rejections. I have just been doing `catch { return; }`, but your comment might be a better solution. I think I've heard feedback that an empty return gives the impression there's a truthy response to expect, which this function does not have.

packages/venia-ui/lib/components/FormError/formError.js
```diff
@@ -26,7 +26,7 @@ const FormError = props => {
 
     return errorMessage ? (
         <div className={classes.root} ref={errorRef}>
-            <span className={classes.errorMessage}>{errorMessage}</span>
+            <div className={classes.errorMessage}>{errorMessage}</div>
```
tjwiebell, line 29: Why was this needed?

jimbo: Good callout. If the content is prose, it should probably be `<p>`. If it's more of a label or snippet, `<span>` was correct.

sirugh: IIRC I switched from `span` to `div` because the left border styling wasn't applying correctly on the root. Since the border no longer applies to the form error (per @schensley's mocks) I can revert back to `span`.

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
sirugh: As you can see I added half of these back. Will add back the ones I missed now :)

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
```diff
@@ -94,30 +94,20 @@ export const useMiniCart = props => {
...
 
     return {
-        loading: miniCartLoading || (removeItemCalled && removeItemLoading),
-        totalQuantity,
-        subTotal,
-        productList,
-        errors,
-        handleRemoveItem,
+        closeMiniCart,
+        derivedErrorMessage,
```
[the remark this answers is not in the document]
sirugh: Certain components such as mini cart are not forms and they don't contain forms. I opted to return the derived error message (variable name debatable).

One thing that this approach raises is the idea to just always return an `errors` array from talons and allow the UI component to handle. In most cases that would be calling `deriveErrorMessage` or passing the errors array to `FormErrors` component. Thoughts?

packages/peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/useCreateAccount.js
```diff
@@ -137,7 +129,7 @@ export const useCreateAccount = props => {
     }, [initialValues]);
 
     return {
-        errors,
+        formErrors: [createAccountError, signInError],
```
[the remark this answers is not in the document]
sirugh: See [this comment](https://github.com/magento/pwa-studio/pull/2588#discussion_r468041233). I think that we should discuss how we handle error responses in talons. In this component they are specifically form errors so I modified the variable name, but I'm starting to feel like maybe just always returning `errors` array is better.

packages/venia-ui/lib/components/FormError/formError.js
sirugh: Resolved in cbd59e3

packages/peregrine/lib/util/deriveErrorMessage.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * A function to derive an error string from an array of errors.
+ */
+export const deriveErrorMessage = errors => {
+    const errorCollection = errors.filter(Boolean).map(error => {
```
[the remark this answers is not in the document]
sirugh: Resolved in c1c0abf

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
sirugh: Resolved in 9e88440

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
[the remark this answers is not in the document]
sirugh: I have renamed the variable to `errorMessage` f4de15405

packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
```diff
@@ -92,16 +91,15 @@ export const useCouponCode = props => {
...
     ]);
 
-    const derivedErrorMessage = useMemo(
-        () => deriveErrorMessage([applyError, removeCouponError]),
-        [applyError, removeCouponError]
-    );
+    const errors = useMemo(() => new Map(), []);
+    errors.set('applyCouponMutation', applyError);
+    errors.set('removeCouponMutation', removeCouponError);
+    errors.set('getAppliedCouponsQuery', fetchError);
```
tjwiebell, line 97: I believe this would create a new Map every loop; would this be better?

```suggestion
    const errors = useMemo(() => new Map([
        ['applyCouponMutation', applyError],
        ['removeCouponMutation', removeCouponError],
        ['getAppliedCouponsQuery', fetchError]
    ]), [applyError, fetchError, removeCouponError]);
```

sirugh: We need to be able to create the map but set and unset each error instance individually. I think an effect for each might be required, as your suggestion would recreate the map every time an error occurred or the query response changed.

sirugh: I also don't think this will create a new error on loop - it should only create once on mount since there are no dependencies.

packages/peregrine/lib/hooks/useScrollIntoView.js
```diff
@@ -0,0 +1,27 @@
...
+    thing,
+    options = {
+        behavior: 'smooth',
+        block: 'center'
+    }
+) => {
+    useEffect(() => {
+        if (
+            ref.current &&
+            typeof ref.current.scrollIntoView !== 'undefined' &&
```
tjwiebell, line 21: Big improvement adding this check, thank you 🙂  Since this hook isn't responsible for ref creation and attachment, I'm curious if we need to be more strict. eg. `ref.current instanceof HTMLElement`. This would error if somehow a truthy ref made it in.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
sirugh: Discussed with @jimbo memoizing a single map on init will not work if folks want to use the map's identity to do some work in the component. Better to recreate the map when the properties change.

packages/peregrine/lib/talons/CartPage/PriceAdjustments/useCouponCode.js
```diff
@@ -10,12 +10,14 @@ export const useCouponCode = props => {
...
 
     const [{ cartId }] = useCartContext();
-    const [fetchAppliedCoupons, { data, error: fetchError }] = useLazyQuery(
-        getAppliedCouponsQuery,
-        {
-            fetchPolicy: 'cache-and-network'
-        }
-    );
+    // Create a memoized error map and toggle individual errors when they change
+    const errors = useMemo(() => new Map(), []);
```
jimbo, line 14: I think the suggestion from @tjwiebell is necessary, actually.

The approach used here is a mutable one: it memoizes the map and then mutates it, so components are vulnerable to missing the update entirely. Instead, we want an immutable approach, where a new map is created and returned whenever the underlying data changes.

sirugh: Handled in c787ebf

packages/peregrine/lib/hooks/useScrollIntoView.js
```diff
@@ -0,0 +1,27 @@
...
+            !!thing
```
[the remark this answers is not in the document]
sirugh: b518f7d

packages/peregrine/lib/hooks/useScrollIntoView.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Object} options https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView
+ */
+export const useScrollIntoView = (
+    ref,
+    thing,
+    options = {
+        behavior: 'smooth',
+        block: 'center'
+    }
+) => {
```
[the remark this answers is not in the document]
sirugh: b518f7d

packages/peregrine/lib/hooks/useScrollIntoView.js
sirugh: b518f7d

sirugh:
After speaking with Dev it's clear I missed a few things in this PR. Here's a TODO list for myself:

* Update price summary error to match mock 
    centered, red text
* Update Gift Card entry - when "applied gift card" error, match mock
    render input form but show error in section for applied gift card.
    "There was an error loading applied gift cards. Please refresh and try again."
    red text, padding above/below
* Update price summary - when fetch error, render red ! triangle icon in place of price that couldn't be rendered.
* cart page product list - render red text top left of product in product listing on error (add/remove)

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -2,10 +2,20 @@
...
     grid-template-columns: 100px 1fr min-content;
 }
 
+.errorText {
+    grid-area: 'error';
+    color: rgb(var(--venia-global-color-error));
+    grid-column-start: 1;
+    grid-column-end: 4;
+    padding-bottom: 0.5rem;
+}
```
tjwiebell, line 17:
```suggestion
.errorText {
    grid-area: error;
    color: rgb(var(--venia-global-color-error));
    padding-bottom: 0.5rem;
}
```

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -2,10 +2,20 @@
     align-items: start;
     display: grid;
     grid-gap: 0.5rem 1rem;
-    grid-template-areas: 'image details kebab';
+    grid-template-areas:
+        'error error error'
```
tjwiebell, line 6: It would be nice if empty rows collapsed the gap too, but this unfortunately leads to some extra padding when no error is rendered.

This is me removing the `error error error` row from the grid template:


To accommodate this I believe we instead always render the error container, but add an `:empty` style to display none, which will correctly collapse that gap.

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
sirugh: Removing the column start/end attributes causes the error to wrap within the first column.

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
sirugh: Ok - we should do the same for out of stock then error then :D

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
sirugh: Oh - I had the wrong syntax for `grid-area` :(

sirugh: https://github.com/magento/pwa-studio/pull/2588/commits/e564fbf6624bfaf678360bc8b580f5b1dec461bf

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
sirugh: One option would be to have an error div outside of the root component so we wouldn't need to even worry about grid.

sirugh: I was able to handle this by adding a wrapping element around the entire component with children being the error span and what was previously the product item `li`. The `display:none` is working as expected!

9a9b6f99.png

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -1,11 +1,25 @@
 .root {
+    display: grid;
+    grid-auto-flow: row;
```
tjwiebell, line 3: This is the default, shouldn't be needed.

```suggestion
```

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
```diff
@@ -1,11 +1,25 @@
...
     align-items: start;
     display: grid;
     grid-gap: 0.5rem 1rem;
     grid-template-areas: 'image details kebab';
     grid-template-columns: 100px 1fr min-content;
 }
 
+.errorText {
+    color: rgb(var(--venia-global-color-error));
+    padding-bottom: 0.5rem;
```
tjwiebell, line 16: Since you have a fancy new grid you're working in, this can be defined by the `.root row-gap` property now. You did go from `0.5rem row-gap + 0.5rem padding-bottom = 1rem total` to just `0.5 padding-bottom`, not sure if that was intentional either.

schensley:
@sirugh things look good from a UX perspective. 
The line-height of the wrapped error message is a little tight which can affect readability. I'd recommend increasing the line-height the equivalent of ~2px.

sirugh:
@jimbo I checked the styleguide and for "body" text it says to use `1.5rem` which seems a bit high.

packages/venia-ui/lib/components/CartPage/ProductListing/product.css
sirugh: Personally I think `0.5rem` helps attribute the error to the product.

sirugh:
@tjwiebell I'm going to be adding `line-height: var(--venia-global-lineHeight-300);` to all the errors fields and to form error in a bit. May want to hold off on review.

PR #2590, opened 2020-07-30
My Account Sign In.

revanth0212:
## Description

Added singin section to the new My account trigger in the header section.

## Related Issue
Closes [PWA-625](https://jira.corp.magento.com/browse/PWA-625)

### Verification Stakeholders
@soumya-ashok 
@schensley 
@jimbo 

### Verification Steps
1. Go to the deployed app and click on the `Sign In` trigger in the header.
2. Sign in should work as expected.

image

image

image

image

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2590.pwa-venia.com : LH Performance Expected 0.85 Actual 0.34, LH Accessibility Expected 1 Actual 0.97, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 39.333333333333
https://pr-2590.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2590.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

schensley:
@revanth0212 this looks really good.  UX approved.

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -33,28 +35,24 @@ export const useSignIn = props => {
         { getUserDetails, setToken }
     ] = useUserContext();
 
-    const [signIn, { error: signInError }] = useMutation(signInMutation, {
-        fetchPolicy: 'no-cache'
+    const [signIn] = useMutation(signInMutation, {
+        fetchPolicy: 'no-cache',
+        onError: signInError => {
+            handleFormErrors([signInError]);
```
sirugh, line 41: The new FormErrors component removes the need to do much of this manual handling. You would just return the errors in a form errors array and then pass that to your `<FormErrors>` component:

https://github.com/magento/pwa-studio/blob/4e65464b1f77e19939ff46ef0f91cdcb11fc3271/packages/peregrine/lib/talons/CheckoutPage/ShippingInformation/AddressForm/useCustomerForm.js#L135

packages/peregrine/lib/talons/SignIn/useSignIn.js
```diff
@@ -63,31 +61,36 @@ export const useSignIn = props => {
...
+                            destinationCartId,
+                            sourceCartId
+                        }
+                    });
+
+                    // Ensure old stores are updated with any new data.
+                    getUserDetails({ fetchUserDetails });
+                    getCartDetails({ fetchCartId, fetchCartDetails });
+                } else {
+                    throw new Error('Unable to Signin, please try again.');
```
sirugh, line 92: You shouldn't need to do this since the mutation will result in an array, and the awaited sign in should throw.

packages/peregrine/lib/talons/Header/useAccountTrigger.js
```diff
@@ -47,24 +56,82 @@ export const useAccountTrigger = props => {
...
+        VIEWS.SIGNIN
+    ]);
 
     const handleTriggerClick = useCallback(() => {
         // Toggle the Account Menu.
         setAccountMenuIsOpen(isOpen => !isOpen);
-    }, [setAccountMenuIsOpen]);
+        /**
+         * When the dropdown closes, the view has to
+         * reset to SINGIN view if the user is not
```
sirugh, line 73:
```suggestion
         * reset to SIGNIN view if the user is not
```

packages/peregrine/lib/talons/SignIn/useSignIn.js
revanth0212: I am doing this instead of a simple `formErrors: [error1, error2]` because these states are being lifted on to the parent. The parent has no idea which one's a duplicate error, all it does it append it the list of errors it maintains. If I keep this login inside a `useEffect` and check against the error from `const [signIn, { error: 'signInError' }]`, I have maintain another variable to not push duplicate values. To avoid doing that, I have used the `onError` structure on `useMutation` which will only fire once on an error occurrence.

packages/peregrine/lib/talons/SignIn/useSignIn.js
revanth0212: Since I am using `onError` method instead,
```
const signInResponse = await signIn({
    variables: { email, password }
});```
is not throwing an error in case of an error situation. I am not sure if that is expected behavior from Apollo or not because it is not documented. Hence I have changed the working of this function. The function had a try-catch block and I needed it to enter that block manually, hence I did that in the else case.

File: packages/peregrine/lib/talons/SignIn/useSignIn.js (hunk 1 above)
Comment (sirugh) [reply packages/peregrine/lib/talons/SignIn/useSignIn.js:41]: What component is the parent and why does it need to know about errors from the sign in form?

File: packages/peregrine/lib/talons/Header/useAccountTrigger.js [hunk 4]
```diff
@@ -47,24 +56,82 @@ export const useAccountTrigger = props => {
...
+        setView(VIEWS.FORGOT_PASSWORD);
+    }, [VIEWS.FORGOT_PASSWORD]);
+
+    const handleCreateAccount = useCallback(() => {
+        setView(VIEWS.CREATE_ACCOUNT);
+    }, [VIEWS.CREATE_ACCOUNT]);
+
+    const handleFormErrors = useCallback(
+        errors => {
+            setFormErrors([...formErrors, ...errors]);
```
sirugh, line 91: Ah here is my confusion - why does the trigger care about form errors for some child component?

revanth0212: Because of the placement of the errors. All errors need to be shown in a single place on the UI. If we don't do this, SignIn errors will show up in a different place and the errors from other views in Signin / Trigger errors in another place. We do not want to show / render multiple `FormError` components.

To avoid this, during the slack conversation couple of days ago, we decided to try the state lift design.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -36,14 +36,21 @@ export const useAuthModal = props => {
...
 
     const apolloClient = useApolloClient();
+
     const [isSigningOut, setIsSigningOut] = useState(false);
     const [username, setUsername] = useState('');
     const [{ currentUser, isSignedIn }, { signOut }] = useUserContext();
-    const [revokeToken] = useMutation(signOutMutation);
+    const [revokeToken] = useMutation(signOutMutation, {
+        onError: signOutMutationErrors => {
+            handleFormErrors([signOutMutationErrors]);
```
sirugh, line 51: This is not a form error, the error should be destructured from the mutation response object and handled in a toast.

packages/peregrine/lib/talons/SignIn/useSignIn.js
sirugh: Oh, right, it won't throw, but it will return an error in the mutation response which can be handled by formerror.

packages/peregrine/lib/talons/Header/useAccountTrigger.js
```diff
@@ -19,9 +19,13 @@ import { clearCustomerDataFromCache } from '@magento/peregrine/lib/Apollo/clearC
  * @returns {Boolean}   talonProps.isUserSignedIn - Whether there is a user currently signed in or not.
  */
 export const useAccountTrigger = props => {
-    const { mutations } = props;
+    const { mutations, VIEWS } = props;
     const { signOut: signOutMutation } = mutations;
 
+    const [formErrors, setFormErrors] = useState([]);
+    const [view, setView] = useState(VIEWS.SIGNIN);
```
sirugh, line 26: I don't think account trigger needs to be the owner of the view state - that can belong in the `AccountMenu` component.

packages/peregrine/lib/talons/Header/useAccountTrigger.js
```diff
@@ -32,10 +36,15 @@ export const useAccountTrigger = props => {
     const apolloClient = useApolloClient();
     const history = useHistory();
     const location = useLocation();
-    const [revokeToken] = useMutation(signOutMutation);
+    const [revokeToken] = useMutation(signOutMutation, {
+        onError: signOutMutationErrors => {
+            setFormErrors([...formErrors, signOutMutationErrors]);
```
sirugh, line 41: Similar to my other feedback - this doesn't need handling here. The error should be destructured from the mutation response object and handled in a toast.

packages/peregrine/lib/talons/Header/useAccountTrigger.js
sirugh: I must have missed the conversation. 

For what it's worth, #2588 will update the sign in component with a new error handling styling. If you're going to re-use the component (I think that's a great idea) then you actually don't need to do any error handling in this PR at all. I think we can back out the error handling changes from this PR and keep the scope smaller.

packages/peregrine/lib/talons/Header/useAccountTrigger.js
```diff
@@ -47,24 +56,82 @@ export const useAccountTrigger = props => {
...
         // Toggle the Account Menu.
         setAccountMenuIsOpen(isOpen => !isOpen);
-    }, [setAccountMenuIsOpen]);
+        /**
+         * When the dropdown closes, the view has to
+         * reset to SIGNIN view if the user is not
+         * signed in.
+         */
+        if (!isUserSignedIn) {
+            setView(VIEWS.SIGNIN);
```
sirugh, line 77: What if I'm creating an account, get an error, and then accidentally close the dropdown by clicking outside of it? When I go to re-open the form through the account trigger all of my input values will be gone! I think this needs some re-thinking. Since state is persistent we could keep whatever was previously opened as the rendered child and provide navigation buttons similar to how the left drawer nav works with account chip.

packages/peregrine/lib/talons/SignIn/useSignIn.js
sirugh: Ah I misunderstood - regardless, you shouldn't need to do this. The work in #2588 will handle signin errors.

packages/venia-ui/lib/components/AccountMenu/accountMenu.js
```diff
@@ -3,23 +3,80 @@ import { bool, func, shape, string } from 'prop-types';
...
-    const dropdownContents = isUserSignedIn ? (
-        <AccountMenuItems handleSignOut={handleSignOut} />
-    ) : (
-        'Sign In TBD (PWA-625)'
-    );
+
+    let dropdownContents = null;
+
+    switch (view) {
+        case VIEWS.ACCOUNT: {
```
sirugh, line 32: Nice work with the view flow. It should behave similarly to the left drawer.

packages/venia-ui/lib/components/AuthModal/authModal.js
```diff
@@ -56,14 +58,24 @@ const AuthModal = props => {
...
             break;
         }
     }
 
     const classes = mergeClasses(defaultClasses, props.classes);
-    return <div className={classes.root}>{child}</div>;
+    return (
+        <div className={classes.root}>
+            {child}
+            <FormErrors
```
sirugh, line 73: You won't need this since #2588 handles it.

packages/venia-ui/lib/components/Header/accountTrigger.js
```diff
@@ -47,11 +63,16 @@ const AccountTrigger = props => {
...
                 ref={accountMenuRef}
+                view={view}
+                username={username}
+                isOpen={accountMenuIsOpen}
+                handleSignOut={handleSignOut}
+                handleForgotPassword={handleForgotPassword}
+                handleCreateAccount={handleCreateAccount}
+                updateUsername={setUsername}
             />
+            <FormErrors errors={formErrors} />
```
sirugh, line 75: I'm not sure the sign out error is considered a form error, but I ran in to similar issues in #2588 where some mutation/query returns an error that we want to parse and render. Since signing out is not done in a form I would ask that you remove the use of it here. I can then make a note to pop a toast in #2588 since I have the `deriveErrorMessage` utility function.

packages/peregrine/lib/talons/Header/useAccountTrigger.js
sirugh: This still feels odd to me, what do you think about moving ownership of the view to AccountMenu?

revanth0212:
@schensley I had to delete the piece of code that resets the error state on dropdown hide. It will be handled as part of https://jira.corp.magento.com/browse/PWA-841.

Can you please perform another UX review and ignore the error state handling for now?

packages/venia-ui/lib/components/Header/accountTrigger.js
revanth0212: That sounds fine to me. Signout certainly feels like a toast error than a form error.

packages/peregrine/lib/talons/Header/useAccountTrigger.js
revanth0212: I agree with you on that. Moved the logic to the new account menu talon. 4e21c97

packages/peregrine/lib/talons/Header/useAccountTrigger.js
revanth0212: It was a UX requirement to reset the view when the dropdown is closed. The point you raised it totally acceptable. @schensley any thoughts on this?

packages/venia-ui/lib/components/SignIn/signIn.js
```diff
@@ -122,7 +122,15 @@ SignIn.propTypes = {
         signInError: string,
         title: string
     }),
-    setDefaultUsername: func.isRequired,
-    showCreateAccount: func.isRequired,
-    showForgotPassword: func.isRequired
+    setDefaultUsername: func,
+    showCreateAccount: func,
+    showForgotPassword: func,
+    showMyAccount: func
```
sirugh, line 128: Can remove `showMyAccount` prop and default.

packages/peregrine/lib/talons/Header/useAccountMenu.js
```diff
@@ -58,14 +58,12 @@ export const useAccountMenu = props => {
...
-    // Set view to sign in / account when the dropdown is closed
+    // Update view based on user status everytime accountMenuIsOpen has changed.
     useEffect(() => {
-        if (!accountMenuIsOpen) {
-            if (isUserSignedIn) {
-                setView('ACCOUNT');
-            } else {
-                setView('SIGNIN');
-            }
+        if (isUserSignedIn) {
```
revanth0212, line 63: Instead of writing 2 if cases, 1 for `accountMenuIsOpen` is true and 1 for false, i took the approach of adding `accountMenuIsOpen` to dependency array even though it is not used in the effect. This works but a new approach we haven't seen before in our code base. Let me know what you guys think.

sirugh: We do this in the affect immediately above this with `location.pathname`. And you've already left a comment to that affect above here so I'm fine with it.

packages/peregrine/lib/talons/Header/useAccountMenu.js
```diff
@@ -0,0 +1,80 @@
...
+    useEffect(() => {
+        if (isUserSignedIn) {
+            setView('ACCOUNT');
+        } else {
+            setView('SIGNIN');
+        }
+    }, [accountMenuIsOpen, isUserSignedIn]);
+
+    return {
+        accountMenuIsOpen,
```
sirugh, line 71: You don't need to return props from the talon that aren't modified within the talon. In fact there seem to be a lot of props being returned here that aren't used in `accountMenu.js`. Can you get rid of the unused ones?

packages/venia-ui/lib/components/AccountMenu/accountMenuItems.css
```diff
@@ -1,3 +1,11 @@
+.root {
+    display: grid;
+    grid-auto-flow: row;
+    justify-content: stretch;
+    row-gap: 1.5rem;
+    padding: 1.5rem 2.5rem;
```
sirugh, line 6: This is affecting the account menu within left nav.

[Image from Gyazo](https://gyazo.com/a205cec22f153e8d2d264a3dd6e5eb8d)

packages/peregrine/lib/talons/Header/useAccountTrigger.js
```diff
@@ -18,53 +13,24 @@ import { clearCustomerDataFromCache } from '@magento/peregrine/lib/Apollo/clearC
  * @returns {Function}  talonProps.handleTriggerClick - A function for handling when the trigger is clicked.
  * @returns {Boolean}   talonProps.isUserSignedIn - Whether there is a user currently signed in or not.
```
sirugh, line 14: Make sure you update the talon props now that the return has changed.

schensley:
@sirugh I take your point generally speaking for clearing a form isn't always ideal (actually it depends on the expectations of the user), in this case **account sign-in** needs only a few bits of information, and the shopper is not required to do so to use the site. Persisting the error once the module is closed and re-opened seems to give the impression that either the shopper has been locked out or sign-in is broken. I think adding navigation elements to the module would introduce clutter and confusion to an otherwise simple and straightforward interaction.

packages/peregrine/lib/talons/Header/useAccountMenu.js
```diff
@@ -0,0 +1,86 @@
...
+import { useApolloClient, useMutation } from '@apollo/react-hooks';
+
+import { useUserContext } from '@magento/peregrine/lib/context/user';
+import { clearCartDataFromCache } from '@magento/peregrine/lib/Apollo/clearCartDataFromCache';
+import { clearCustomerDataFromCache } from '@magento/peregrine/lib/Apollo/clearCustomerDataFromCache';
+
+/**
+ * The useAccountMenu talon complements the AccountMenu component.
+ *
+ * @param {Obejct} props
```
sirugh, line 12:
```suggestion
 * @param {Object} props
```

packages/venia-ui/lib/components/AccountMenu/__tests__/accountMenu.spec.js
```diff
@@ -2,16 +2,58 @@ import React from 'react';
 import { createTestInstance } from '@magento/peregrine';
 
 import AccountMenu from '../accountMenu';
+import { VIEWS } from '../../Header/accountTrigger';
 
 jest.mock('../accountMenuItems', () => 'AccountMenuItems');
+jest.mock('../../SignIn/signIn', () => 'Singin Component');
```
sirugh, line 8:
```suggestion
jest.mock('../../SignIn/signIn', () => 'SignIn Component');
```

dpatil-magento:
@revanth0212  Resolve the conflicts pls.

revanth0212:
> @revanth0212 Resolve the conflicts pls.

Done dude.

PR #2602, opened 2020-08-03
PWA-644: Use GET for GraphQL queries (but not mutations)

DrewML:
## Related Issue

Closes [PWA-644](https://jira.corp.magento.com/browse/PWA-644)

## Description

https://devdocs.magento.com/guides/v2.4/graphql/caching.html

Enables using GET for GraphQL queries. Some work-arounds were used to ensure that clients are unlikely to hit URL length limits in most environments:

- Fastly's WAF currently blocks any requests that include a match on the pattern `include(.+)`. This means that, when using Magento Cloud (Fastly) _and_ the `@include` directive in GraphQL, requests will fail. We need to work with the cloud team on this, but for the time being @sirugh / @zetlen said to just remove `@include` usage entirely (we only used it for an early ssr experiment, afaik).

- This PR makes a time/space trade-off. Apollo Client serializes queries from an AST _at runtime_, and these queries have a ton of unnecessary whitespace and punctuators. To address that, this PR takes a ~16kb hit on client-side JS, in return for significant savings on API request times. The `16kb` comes from the [`stripIgnoredCharacters` utility](https://github.com/graphql/graphql-js/blob/7b3241329e1ff49fb647b043b80568f0cf9e1a7c/src/utilities/stripIgnoredCharacters.js) in the GraphQL JavaScript reference implementation. There are other ways y'all could technically hack around this (webpack-swap the `printer` function from graphql js, and implement a better printer. I have some scratch code that works if anyone wants to take a stab at it).

- Apollo client's [HTTP link uses the typical `encodeURIComponent` function on queries](https://github.com/apollographql/apollo-client/blob/7993a778ef5a9e61bfee72b956f4c179fce691f4/src/link/http/rewriteURIForGET.ts#L11). However, thanks to "don't break the web," we can cheat a bit here since query strings allow `application/x-www-form-urlencoded` encoding. So we (hackily) re-write the URLs HTTP link generates, and use the `URLSearchParams` interface to get more efficient encoding.

## Results

**Before/After Comparisons (Note they're backwards, after is on the left)**
image

I used the `category` query as a test, and used `https://venia.magento.com/graphql` for the API.

### Resolve Route
**Before**: 374ms
**After**: 49ms

### Navigation Query
**Before**: 428ms
**After**: 50ms

### Category Query
**Before**: 1.2s
**After**: 67ms

### Product Filters By Category
**Before**: 539ms
**After**: 76ms

In this test, the total savings was over 2 seconds for page load of a category page.

## Commits

This PR is split up into 3 separate logical commits:

- cf28ebb83fc84ad8ce806f8f8e966221027871e0 (Remove usage of `onServer`)
- 245e8d6279b3c0c272e12cdeca0f0e11d2b55cfc (Switch apollo-client to `GET`)
- e7e6fffebd1cdd78d1226c8305d9577aded83207 (Switch `resolveUnknownRoutes` to `GET`)

### Verification Stakeholders
Don't know!

### Verification Steps

1. Go look at queries in the storefront
2. Some of them are faster!

**Before/After Comparison of GraphQL network requests**
image

## Checklist

- [X] I have added tests to cover my changes, if necessary.


packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -116,8 +117,27 @@ const VeniaAdapter = props => {
...
+        if (options.method === 'GET') {
+            url = shrinkGETQuery(uri);
+        }
+        return fetch(url, options);
+    };
+
     return createHttpLink({
-        uri: apiBase
+        uri: apiBase,
+        fetch: customFetchToShrinkQuery,
```
DrewML, line 135: I just could not find a better way to hook into Apollo. Happy to hear other suggestions, but hard to modify the behavior of a terminating link, from what I saw.

packages/venia-ui/lib/util/__tests__/shrinkGETQuery.spec.js
```diff
@@ -0,0 +1,166 @@
...
+    );
+    expect(apolloClientGETURL.length).toMatchInlineSnapshot(`6223`);
```
DrewML, line 164: These are honestly just here to prove out the savings w/ the new URL encoding. Happy to change this

packages/venia-ui/package.json
```diff
@@ -82,6 +82,7 @@
     "apollo-cache-persist": "~0.1.1",
     "apollo-client": "2.6.4",
     "apollo-link-http": "~1.5.11",
+    "graphql": "~14.3.1",
```
DrewML, line 85: This was a transitive peer dependency before (via apollo-client). Now that we're using it in venia-ui/lib/util, made it explicit.

devops-pwa-codebuild:
**Performance Test Results**

https://pr-2602.pwa-venia.com : LH Performance Expected 0.85 Actual 0.35, LH Accessibility Expected 1 Actual 0.97, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 90 Actual 39.333333333333
https://pr-2602.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.36, LH Best Practices Expected 1 Actual 0.92
https://pr-2602.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.47, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 48

packages/peregrine/lib/Router/resolveUnknownRoute.js
```diff
@@ -89,22 +89,23 @@ function remotelyResolveRoute(opts) {
  * @returns {Promise<{type: "PRODUCT" | "CATEGORY" | "CMS_PAGE"}>}
  */
 function fetchRoute(opts) {
+    const query = `query ResolveURL($url: String!) {
```
DrewML, line 92: I added a wrapping operation around this (ResolveURL) to use a variable and get rid of the string interpolation, which allows query injection.

Example: venia.magento.com/venia-bottoms.html%22)%20%7B%20type%20id%20%7D%20foo:%20urlResolver(url:%20%22isadora-skirt.html

packages/venia-ui/lib/util/shrinkGETQuery.js
```diff
@@ -0,0 +1,32 @@
+import { stripIgnoredCharacters } from 'graphql/utilities/stripIgnoredCharacters';
+
+/**
+ * Shrink a GraphQL query inside of a URL used in a GET request.
+ * There are 2 problems with Apollo-client's encoding of URLs:
```
sirugh, line 5: Did you find any issues or documentation about this? It would be good to link them here for posterity if so - it'd be cool if we didn't have to do this ourselves.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -116,8 +117,27 @@ const VeniaAdapter = props => {
...
 VeniaAdapter.apolloLink = apiBase => {
+    // Intercept and shrink URLs from GET queries. Using
+    // GET makes it possible to use edge caching in Magento
+    // Cloud, but risks exceeding URL limits with default usage
+    // of Apollo's http link. `shrinkGETQuery` encodes the URL
+    // in a more efficient way.
+    const customFetchToShrinkQuery = (uri, options) => {
+        let url = uri;
+        if (options.method === 'GET') {
+            url = shrinkGETQuery(uri);
```
sirugh, line 128: Is there any reason we wouldn't want to strip the ignored characters for POST as well?

packages/venia-ui/lib/drivers/adapter.js
sirugh: I think this is fine - the docs suggest that you do this if you need to target browsers where `fetch` is not natively provided. This seems to be the right place to do this.

packages/peregrine/lib/Router/resolveUnknownRoute.js
```diff
@@ -89,22 +89,23 @@ function remotelyResolveRoute(opts) {
...
+        urlResolver(url: $url) {
+            type
+            id
+        }
+    }`;
+
     const url = new URL('/graphql', opts.apiBase);
+    url.searchParams.set('query', query);
```
sirugh, line 100: Since this query is made outside the apollo client it does not get stripped of ignored characters.

DrewML: Yeah, I kind of figured it didn't matter here, since the total length of this query is based on the URL of the page.

Want me to change it up?

packages/venia-ui/lib/drivers/adapter.js
DrewML: We could! I honestly hadn't given it much thought 😄  Was just critical for GET so we don't exceed limits

packages/venia-ui/lib/util/shrinkGETQuery.js
DrewML: Planning on opening a PR to `apollo-client` to discuss the changes with them. Haven't quite gotten there yet, though!

DrewML: There are some tools today that already use the `stripIgnoreCharacters` utility for this (like [`babel-plugin-graphql-tag`](https://github.com/gajus/babel-plugin-graphql-tag/blob/e76450c2cf667ebdac85dc34a0dc5d3fce64ec8c/src/index.js#L10)), but I haven't seen any solutions that work for apollo client (since it doesn't use the string compiled by `babel-plugin-graphql-tag` or `graphql-tag/loader`)

And re: the querystring encoding, I haven't managed to find any prior discussions about it on the interwebs. Surprising though cause I'd doubt we're the first to notice this.

packages/peregrine/lib/Router/resolveUnknownRoute.js
sirugh: Not really, just pointing it out :) If/When we move url resolver to go through ApolloClient it won't matter.

packages/venia-ui/lib/util/shrinkGETQuery.js
sirugh: > There are some tools today that already use the stripIgnoreCharacters utility for this (like babel-plugin-graphql-tag), but I haven't seen any solutions that work for apollo client (since it doesn't use the string compiled by babel-plugin-graphql-tag or graphql-tag/loader)

Huh, we use the babel plugin. It might shave some bytes on the bundle if we can do the optimization there rather than right before the fetch.

DrewML: > It might shave some bytes on the bundle if we can do the optimization there rather than right before the fetch.

This was actually the first version of my implementation. It turns out, apollo-client can modify the AST of the query for certain things, so they always serialize from the AST, rather than using the string generated at build time.

https://github.com/apollographql/apollo-client/blob/7993a778ef5a9e61bfee72b956f4c179fce691f4/src/link/http/selectHttpOptionsAndBody.ts#L132

So my original implementation saved a tiny bit on the size of queries in the bundle, but it raised a bigger issue: the bundle has _both_ the string query and the AST, but only the AST is used.

sirugh: Sounds like we should still take advantage of it to not have to include the 16kb stripIgnored utility, right? Or am I missing something :P

DrewML: > Sounds like we should still take advantage of it to not have to include the 16kb stripIgnored utility, right? Or am I missing something :P

Still needed. Apollo serializes the AST itself, and that serialization includes spaces/punctuators/etc. The query you write in your editor never actually goes over the wire - it just gets converted to an AST and inlined.

sirugh:
> Fastly's WAF currently blocks any requests that include a match on the pattern include(.+). This means that, when using Magento Cloud (Fastly) and the @include directive in GraphQL, requests will fail. We need to work with the cloud team on this, but for the time being @sirugh / @zetlen said to just remove @include usage entirely (we only used it for an early ssr experiment, afaik).

@jcalcaben we should make sure to document this somewhere visible. We are removing our own usage of the directive but if anyone in the community is using it this will potentially break them.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -116,8 +117,27 @@ const VeniaAdapter = props => {
...
+    };
+
     return createHttpLink({
-        uri: apiBase
+        uri: apiBase,
+        fetch: customFetchToShrinkQuery,
+        // Warning: useGETForQueries risks exceeding URL length limits. These limits
+        // in practice are typically set at or behind where TLS terminates. For Magento
+        // Cloud and Fastly, 8kb is the maximum by default
+        // https://docs.fastly.com/en/guides/resource-limits#request-and-response-limits
```
jimbo, line 139: If the query ends up exceeding this character/byte limit, do we have anything prepared to throw an informative error?

packages/venia-ui/lib/util/__tests__/shrinkGETQuery.spec.js
jimbo: Right, the inline screenshots are probably too specific. A basic comparison asserting that the new approach produces a string of lesser length would suffice here. 👍

packages/venia-ui/lib/drivers/adapter.js
fooman: Any chance to only remove the offending include statement?

Ie
`meta_title @include(if: $onServer)`
To
`meta_title`

DrewML:
FYI @billygilbert is on 🔥 and already is logging a ticket for Fastly

packages/venia-ui/lib/drivers/adapter.js
DrewML: We don't, but I do think it's worth consideration.

My biggest fear was that this can be deployed to non-Magento-owned infrastructure, and there will be different limits (some allow > 8kb).

We could always make it configurable and set a default of 8kb (or ∞)

DrewML: > Any chance to only remove the offending include statement?

To clarify @fooman, you still want the data in the default query, just without the conditional directive, right? We could do that. @jimbo thoughts?

fooman: @DrewML yes - this seems the less intrusive change. Currently I am not sure how the `@include` directive is supposed to work but from the comments so far the idea seems to have been that the query only executes for SSR. However when I browse on a current PR (other than this), meta_title and meta_description do update when navigating the site client side. Keeping the data in the query would not change this behaviour while fixing the issue with `@include`

sirugh: The downside of keeping the fields is that they aren't actually used by anything in pwa studio right now and is both extra data requested and mental overhead when we try to make decisions about new queries or old query maintenance.

I'd rather we provide, for users of pwa-studio, explicit queries that request only the data necessary but that we also allow an easy path to customizing/overriding these queries.

DrewML: > So I'm on board with the less intrusive change of just removing the directives rather than the fields, but I'd like to add a comment block around each field indicating that the data isn't actually used by any of the venia components by default.


Is this something I should push up? Happy to make the change, just want to make sure there's consensus 😃

DrewML:
@jimbo 

> Right, the inline screenshots are probably too specific. A basic comparison asserting that the new approach produces a string of lesser length would suffice here.

Changed to `expect(result.length).toBeLessThan(apolloClientGETURL.length);`

packages/pwa-buildpack/lib/WebpackTools/__tests__/__fixtures__/queries/getProductDetail.graphql
```diff
@@ -19,9 +19,6 @@ query productDetail($urlKey: String, $onServer: Boolean!) {
                 disabled
                 file
             }
-            meta_title @include(if: $onServer)
-            meta_keyword @include(if: $onServer)
-            meta_description @include(if: $onServer)
```
sirugh, line 24: Please add the fields back sans directive.

packages/venia-ui/lib/drivers/adapter.js
DrewML: Fields added back _without_ the `@include` directive

packages/venia-ui/lib/queries/getCategory.graphql
```diff
@@ -11,8 +10,8 @@ query category(
         description
         name
         product_count
-        meta_title @include(if: $onServer)
-        meta_keywords @include(if: $onServer)
+        meta_title
+        meta_keyword
```
sirugh, line 14:
```suggestion
        meta_keywords
```

Oops! My bad.

packages/venia-ui/lib/queries/getCmsPage.graphql
```diff
@@ -1,12 +1,12 @@
...
     cmsPage(id: $id) {
         url_key
         content
         content_heading
         title
         page_layout
-        meta_title @include(if: $onServer)
-        meta_keywords @include(if: $onServer)
+        meta_title
+        meta_keyword
```
sirugh, line 9:
```suggestion
        meta_keywords
```

DrewML:
Update from @billygilbert:

he spoke with Fastly and got some example Varnish config to add exclusions for `@include`. He's going to open up a ticket to propose applying those changes for cloud customers

DrewML:
There are some caveats that pwa-studio developers will want to know about caching and combining queries - opened a devdocs ticket to track: https://github.com/magento/devdocs/issues/7718

PR #2612, opened 2020-08-06
[PWA-767] [Mini-cart] Guest user can't add product to cart if cart is expired.

tjwiebell:
## Bug Description

Steps -
1. Add product to cart as guest.
2. Wait for guest cart to expire (No sure whats the time, need to check)
3. Try adding product again.

*Expected - *
Toast should trigger informing cart expiration (ensure messaging is user friendly).
New Cart should be created.
Product should get added to new cart.

## Solution
The use case of someone leaving their tab open for the default span of thirty days is a bit out there, and it would be best to treat this exactly like any of our other cart initialization recovery. I've instead just added this expired cart message to cart recovery on initial page load, and removed the logic that was popping a Toast on cart query in the new MiniCart component.

This now should just silently recover, and the user should be able to add items to their cart.

## Related Issue

* [[PWA-767](https://jira.corp.magento.com/browse/PWA-767)] [Mini-cart] Guest user can't add product to cart if cart is expired.

### Verification Steps

The easiest way I found to test this is with database access, making this a little difficult 😬 
1. Add products to your cart
2. `SELECT quote_id FROM quote_id_mask WHERE masked_id = '<cart-id>'`
3. `UPDATE quote SET is_active = 0 WHERE entity_id = <quote-id-from-above>`
4. Simulate customer coming back after > 30 days by just reloading
5. Verify your old cart gets wiped and new cart provisioned by adding a item to your cart

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2612.pwa-venia.com : LH Performance Expected 0.85 Actual 0.71, LH Accessibility Expected 1 Actual 0.98, LH SEO Expected 1 Actual 0.91, LH Best Practices Expected 1 Actual 0.85, WPT Cache Expected 90 Actual 39.333333333333
https://pr-2612.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.35, LH Best Practices Expected 1 Actual 0.92
https://pr-2612.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.38, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.92, WPT Cache Expected 65 Actual 50

packages/peregrine/lib/talons/MiniCart/useMiniCart.js
```diff
@@ -25,15 +25,14 @@ export const useMiniCart = props => {
...
-    const {
-        data: miniCartData,
-        loading: miniCartLoading,
-        error: miniCartError
-    } = useQuery(miniCartQuery, {
-        fetchPolicy: 'cache-and-network',
-        variables: { cartId },
-        skip: !cartId
-    });
+    const { data: miniCartData, loading: miniCartLoading } = useQuery(
```
tjwiebell, line 28: I'm not sure if it was intentional that this query was popping a Toast on error, but it resulted in "The cart is inactive" being rendered occasionally, even though CartContext had already handled the recovery. I considered mapping the message to something more friendly, but since it wasn't guaranteed this would pop consistently, I decided to remove it entirely.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -451,6 +451,7 @@ function isInvalidCart(error) {
         error.graphQLErrors.find(
             err =>
                 err.message.includes('Could not find a cart') ||
+                err.message.includes("The cart isn't active") ||
```
sirugh, line 454: Hmm - how does this retry on product add? `useProductFullDetail` invokes the add mutations directly and just returns on error.

https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js#L247-L253

PR #2619, opened 2020-08-10
My Account - Forgot Password

revanth0212:
## Description

Implemented forgot password UX and workflow for the new my account feature.

For this PR I had to add a new route `/customer/account/createPassword` which provides the UX to reset the user's password.

When the user requests a password reset, the backend sends an email with a button to take the user to a new tab. The link that the backend includes in the email, uses the MAGENTO_BACKEND_URL as the domain. Since all the backends don't have the code to handle the new route `/customer/account/createPassword`, @dpatil-magento helped me deploy the new code on this server: https://staging-vdt2zeq-c5v7sxvquxwl4.eu-4.magentosite.cloud/. If someone wants to test this PR locally, please use the above backend URL instead.

Also if you wanna test this PR on a deployed instance please use #2621. @dpatil-magento made sure the backend for that PR is set to https://staging-vdt2zeq-c5v7sxvquxwl4.eu-4.magentosite.cloud/ which has the code to handle the new route.

Another way to test this is to paste the query string part from the email link next to the deployed app URL, for instance: `https://pr-2619.pwa-venia.com/customer/account/createPassword/` + `?email=gooseton%40gmail.com&token=*******`.

Sorry if this is confusing.

**Note: To be merged only after #2590** 

## Related Issue
Closes [PWA-77](https://jira.corp.magento.com/browse/PWA-77)

## Acceptance 
Users should be able to reset their password.

### Verification Stakeholders
@soumya-ashok 
@schensley 
@jimbo 

### Specification
This PR adds a new route `customer/account/createPassword` which takes 2 query params `?email=EMAIL_ID&token=RESET_TOKEN`. 

### Verification Steps
1. Go to the deployed app and open the sign-in dropdown. Click on the forgot password link.
2. Enter the email id and submit.
3. Check for the reset email. Click the reset button in the email.
4. Reset UI should open up in the new tab.
5. Enter new password and click submit.

Demo link: https://jira.corp.magento.com/secure/attachment/367170/Forgot%20password%20demo.mp4

image

image

image

image

image

image

image

image

image

## Checklist


devops-pwa-codebuild:
**Performance Test Results**

https://pr-2619.pwa-venia.com : LH Performance Expected 0.85 Actual 0.34, LH Accessibility Expected 1 Actual 0.97, LH Best Practices Expected 1 Actual 0.93, WPT Cache Expected 90 Actual 28
https://pr-2619.pwa-venia.com/venia-tops.html : LH Performance Expected 0.75 Actual 0.34, LH Best Practices Expected 1 Actual 0.93, WPT Cache Expected 70 Actual 36
https://pr-2619.pwa-venia.com/valeria-two-layer-tank.html : LH Performance Expected 0.8 Actual 0.41, LH Accessibility Expected 0.9 Actual 0.89, LH Best Practices Expected 1 Actual 0.93, WPT Cache Expected 65 Actual 39

packages/venia-ui/lib/targets/venia-ui-intercept.js
```diff
@@ -122,6 +122,12 @@ module.exports = targets => {
             pattern: '/checkout',
             exact: true,
             path: '../CheckoutPage'
+        },
+        {
+            name: 'Reset Password',
+            pattern: '/customer/account/createPassword',
```
revanth0212, line 128: The route can change. It is the route configured in the forgot password email template in the admin panel.

I don't think we have a setting for changing it without modifying the email template manually. Should we have this in the app config?

sirugh: TBH I'm not sure how much we care about the preconfigured routes. I would also probably use hyphen instead of camelcase ie `/customer/account/create-password`.

revanth0212: It is the default route. If we happen to change this and someone forgets to change the template, the whole workflow fails. That is no reason to not change since someone configuring this should know for sure. I prefer your idea as well, hyphen case is correct in this case.

fooman: I believe until such time that there is a M2 module for Magento Core that configures the front-end to be in line with PWA Studio expectations with a click of a button the default route should work.

If there was preference for something else maybe an upward rule could be added to redirect from the default to the PWA Studio expected?

packages/venia-ui/lib/components/MyAccount/ResetPassword/resetPassword.js
```diff
@@ -0,0 +1,80 @@
...
+        loading,
+        email,
+        token,
+        formErrors,
+        handleSubmit
+    } = talonProps;
+
+    const tokenMissing = (
+        <div className={classes.invalidToken}>
+            Token missing. Error message TODO.
```
revanth0212, line 37: @schensley to provide a proper error message to show when the link is not as expected.

packages/venia-ui/lib/targets/venia-ui-intercept.js
revanth0212: Hey @fooman, I don't think UPWARD will be bothered in this case because the link from the email will directly hit the backend server. If we have runtime UPWARD rules, we can perform such kind of mapping. @zetlen any thoughts?

Link from the email would look something like this: 

https://staging-vdt2zeq-c5v7sxvquxwl4.eu-4.magentosite.cloud/customer/account/createPassword/?email=EMAIL_ID&token=TOKEN

fooman: @revanth0212 I do believe you might have stumbled on a bigger discussion here...

This link:
https://staging-vdt2zeq-c5v7sxvquxwl4.eu-4.magentosite.cloud/customer/account/createPassword/?email=EMAIL_ID&token=TOKEN
combined with running the upward php connector would result in this

https://github.com/magento/magento2-upward-connector/blob/develop/Plugin/Magento/Framework/App/AreaList.php#L50

making it an upward response.

The other option I consider likely is that the back-end will be configured to send out an email like
https://venia.magento.com/customer/account/createPassword/?email=EMAIL_ID&token=TOKEN
(the base url can be changed for the complete site, whereas the /customer/account/createPassword part is in the individual email template) with a separate back-end url. In this scenario upward-js would handle it.

revanth0212: I am not sure if the admin panel has an option to store/change the main site URL / public URL for instance `https://venia.magento.com/`. If so, we can check to use that.

This is how the URL is generated in the email template: 

```
<a href="{{var this.getUrl($store,'customer/account/createPassword/',[_query:[token:$customer.rp_token,email:$customer.email],_nosid:1])}}" target="_blank">{{trans "Set a New Password"}}</a>
```

Ill have to look into the implementation of the `this.getUrl` function to know what variables it is considering.

fooman: @revanth0212 it is all configurable under Stores > Configuration > Web.

I guess the point I was trying to make is that upward is part of it one way or the other (upward-php or upward-js).

schensley:
Thank you @revanth0212 this all looks really good. Hopefully, in the future GQL will provide some enhancements such as password criteria and/or more specific messaging around password errors. UX Approved.

packages/venia-ui/lib/components/Password/password.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
```
revanth0212, line 1: Not using this component in the create account component because it will be handled in its own ticket.

packages/venia-ui/lib/components/MyAccount/ResetPassword/resetPassword.gql.js
```diff
@@ -0,0 +1,22 @@
+import gql from 'graphql-tag';
+
+export const RESET_PASSWORD_MUTATION = gql`
+    mutation resetPassword(
```
sirugh, line 4: Same comment here re: using `@connection`.

packages/venia-ui/lib/components/MyAccount/ResetPassword/resetPassword.js
```diff
@@ -0,0 +1,115 @@
...
+
+    useEffect(() => {
+        if (hasCompleted) {
+            addToast({
+                type: 'info',
+                message: 'Your new password has been saved.',
+                timeout: 5000
+            });
+        }
+    }, [addToast, hasCompleted]);
```
sirugh, line 54: Did we want both a toast and for the component to render a success message? Feels redundant.

packages/peregrine/lib/talons/ForgotPassword/useForgotPassword.js
```diff
@@ -1,37 +1,53 @@
...
+ *  forgotPasswordEmail: string,
+ *  inProgress: boolean,
+ *  isResettingPassword: boolean,
+ *  handleCancel: function,
+ *  handleFormSubmit: function,
+ * }}
  */
 export const useForgotPassword = props => {
-    const [{ isResettingPassword }, { resetPassword }] = useUserContext();
+    const [{ isResettingPassword }] = useUserContext();
```
sirugh, line 21: Do we need `isResettingPassword` anymore? This was just to indicate that the submission of the email was in flight - it shouldn't be necessary anymore since we have a GQL mutation.

packages/peregrine/lib/talons/ForgotPassword/useForgotPassword.js
```diff
@@ -1,37 +1,53 @@
...
 
-    const { onClose } = props;
+    const { onCancel, mutations } = props;
 
     const [inProgress, setInProgress] = useState(false);
     const [forgotPasswordEmail, setForgotPasswordEmail] = useState(null);
 
+    const [resetPassword, { error: resetPasswordError }] = useMutation(
+        mutations.resetPasswordMutation
```
sirugh, line 29: I'd probably call this, or the other mutation that actually sets the new password, something else, just to avoid confusion :)

packages/peregrine/lib/talons/ForgotPassword/useForgotPassword.js
```diff
@@ -1,37 +1,53 @@
...
-        forgotPasswordEmail,
-        handleContinue,
-        handleFormSubmit,
+        formErrors: [resetPasswordError],
         inProgress,
-        isResettingPassword
+        isResettingPassword,
+        forgotPasswordEmail,
+        handleCancel,
+        handleFormSubmit
```
sirugh, line 51: Tiny nit pick - can you alphabetize these? They were before.

packages/peregrine/lib/talons/ForgotPassword/useForgotPassword.js
```diff
@@ -1,37 +1,53 @@
...
+ * @param {function} props.mutations.resetPasswordMutation - mutation to call when the user clicks the submit button
+ *
+ * @returns {{
+ *  formErrors: [Error],
+ *  forgotPasswordEmail: string,
+ *  inProgress: boolean,
+ *  isResettingPassword: boolean,
+ *  handleCancel: function,
+ *  handleFormSubmit: function,
+ * }}
```
sirugh, line 18: Now might be a good opportunity to use the doc style @jcalcaben went over today.

packages/venia-ui/lib/components/MyAccount/ResetPassword/resetPassword.js
revanth0212: Yeah the UX demo had both the message and the toast.

supernova-at:
## QA

### Verification Steps

- [x] Go to the deployed app and open the sign-in dropdown. Click on the forgot password link.
- [x] Enter the email id and submit.
- [ ] Check for the reset email. Click the reset button in the email.
- [ ] Reset UI should open up in the new tab.
- [ ] Enter new password and click submit.

### Other

#### Cannot Reset the Customer's Password

I entered an email that does not have an account and got the following error:

Screen Shot 2020-08-24 at 12 03 43 PM

This seems like a reasonable error message, my only concern is that we may be providing valuable information to a potential attacker. The standard message "If there is an account associated with **_EMAIL_** you will receive an email with a link to change your password" covers this error scenario and it might be better to show that message even when we get actual `errors` in the GraphQL response.

I should also note that I am assuming that the reason we got that error was because there was no account associated with that email address. Should probably check with the GraphQL team to make sure.

revanth0212:
> This seems like a reasonable error message, my only concern is that we may be providing valuable information to a potential attacker. The standard message "If there is an account associated with **_EMAIL_** you will receive an email with a link to change your password" covers this error scenario and it might be better to show that message even when we get this error in the `errors` of the GraphQL response.

If I remember correctly, the GQL team is working on updating their mutations to provide more reasonable/understandable error messages. It should be just a matter of time before they come up with a better error message for such use cases. That said, should we wait for them or render our own message? If we do it ourselves, it will be a standard message for any sort of an error from the server.

Also, even if we do that, it might still be subject to security because the attacker can simply check the network tab for the error reported from the server.

schensley:
@supernova-at 
> After resetting my password and signing in via the site header, I was never navigated away from the "Reset Password" page:

This isn't uncommon for reset pages, at least from my experience. I don't think we can predict where the user would like to be. Maybe they will sign in immediately, but do they intend to go to checkout, perform some Account management, or enable a personalized shopping experience? Arbitrarily navigating them to some page (i.e. Home page) could feel imposing or interpreted as presenting some workflow they think they must complete. Keeping them on a page that displays a Success message allows them to make that decision themselves.

supernova-at:
> Also, even if we do that, it might still be subject to security because the attacker can simply check the network tab for the error reported from the server. What do you guys think?

Good point. I'm good keeping this as-is then, with the hopes that GQL error messages improve.

supernova-at:
Are there security concerns with having the customer email and a reset token in the URL?

revanth0212:
> Are there security concerns with having the customer email and a reset token in the URL?

It is usual practice to include email and a token in the reset link. We can get away with not having the email in the link by adding another field on the page where the user can enter their email. But, token needs to be present in the link.

sirugh:
> It is usual practice to include email and a token in the reset link. We can get away with not having the email in the link by adding another field on the page where the user can enter their email. But, token needs to be present in the link.

Is the link and params configurable in the back end? What is the default setting?

revanth0212:
> Is the link and params configurable in the back end? What is the default setting?

Yeah, link and params are configurable. We are using the default link which is `customer/account/createPassword`. The email template by default only has the logic to include the token in the email, but due to the UX requirements, we had to add the email id as well.

tjwiebell:
> > Is the link and params configurable in the back end? What is the default setting?
> 
> ... but due to the UX requirements, we had to add the email id as well.

This is also a technical requirement, the GraphQL mutation requires token, email, and new password.

revanth0212:
> > ... but due to the UX requirements, we had to add the email id as well.
> 
> This is also a technical requirement, the GraphQL mutation requires token, email, and new password.

Yes, you are right but UX wanted the user to click the link, enter new password and submit. Hence we had to change the email template to include email id as well.

sirugh:
> The email template by default only has the logic to include the token in the email, but due to the UX requirements, we had to add the email id as well.

Are you saying that you customized your back end to include `email` in the reset link that is sent to the user? Was the UX requirement _not_ to have an email input field?

I would be cautious about customizing the default link - it would be preferable to leave it as is so pwa studio just works out of the box.

If we absolutely must stick to this UX then can you create a follow up ticket to have @jcalcaben update the "getting started" steps and whatever documentation is necessary?

My instinct is to have an email entry field and a new password field and to not customize the link.

abso-bmarquis:
Hi @supernova-at,

I believe this code was removed accidentally when merging this PR: https://github.com/magento/pwa-studio/pull/2657
I don't see the feature on any of the current tags.

Please advise.
Regards,
Ben

dpatil-magento:
HI @abso-bmarquis,

Latest PR related to forgot password refactor is https://github.com/magento/pwa-studio/pull/2726.  

Thanks,
Dev

abso-bmarquis:
Thanks!

PR #2634, opened 2020-08-17
fix: Cart/Checkout summary uses cache-and-network to fetch latest

sirugh:
## Description

We use the price summary fragment to update it when on the cart or checkout page but we didn't configure the component to update itself when navigating to the page from another page. In this case the fix is pretty simple, we can use `cache-and-network` to indicate that we should use cached data to render but then true-up with the server.

## Related Issue

Closes [PWA-824](https://jira.corp.magento.com/browse/PWA-824).

### Verification Steps
1. Add an item to your cart.
2. Click mini cart and "CHECKOUT" to navigate to /checkout.
3. After page loads, click the back button in your browser, or use the venia logo to navigate to a product page.
4. Add a product to your cart.
5. Click mini cart and "CHECKOUT" to go back to the checkout page. Both total and subtotal should be correct!

## Checklist


packages/peregrine/lib/talons/CartPage/PriceSummary/usePriceSummary.js
```diff
@@ -34,6 +34,7 @@ export const usePriceSummary = props => {
     const isCheckout = !!match;
 
     const { error, loading, data } = useQuery(getPriceSummary, {
+        fetchPolicy: 'cache-and-network',
```
tjwiebell, line 37: Is there a point to doing `cache-and-network` if we never render the cached result? I think this needs some UX treatment (render cache result while we true up with network), or we might as well just do `network-only`. This pattern didn't exist at the time we did `CartPage`, so willing to accept since it does solve the bug.

sirugh: > Is there a point to doing `cache-and-network` if we never render the cached result.

Well, we do render the cached result, and we want to make sure to render changes to it if something else on the page modifies it, especially since we're not using `refetchQueries`.

What results is that the first time we render price summary it will render cache values if it has it and then true-up with the network. If a product is then removed or added the fragment is refetched (that's the pattern we're going with for mutations) updating the cache values which updates the data response here.

> I think this needs some UX treatment (render cache result while we true up with network),

For what it's worth, we already do this, but through a different method - the price summary component takes an `isUpdating` prop and applies a class to the price entries. I don't think we need to do anything.

tjwiebell: https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CartPage/PriceSummary/priceSummary.js#L57L59

sirugh: Oh....ok :)

sirugh: I updated it to render the "updating" state during the network fetch rather than `null`.

[Image from Gyazo](https://gyazo.com/92440932926e39d206a2dfbe486aec2e)

Issue #1784, opened 2019-09-25
[feature]: Revisit certain components to create talons


jimbo:
**Is your feature request related to a problem? Please describe.**

We're extracting the logic from Venia components into Peregrine talons. Some components have logic that needs to be revisited and potentially reworked before a talon can be created, though.

**Describe the solution you'd like**

Revisit the following components:

- Button
- ButtonGroup
- Checkbox
- Checkout/
   - Form
   - Label
   - Section
   - BraintreeDropin (crazy?) I think this is something very custom/confusing -- people will probably roll their own or just use ours wholesale w/ no customization
   - `<CheckoutFormComponent>Summary` (too simple?) These don't have much functionality beyond display
- CmsBlock
- ErrorView
- Field
- `SearchBar` > `SuggestedProduct` (lots of `resourceUrl`)
- `ToastContainer` (refactor: avoid sorting, duplicates, etc.)
- `Head`
- `LoadingIndicator`
- `Logo`
- `Main`
- `Mask`
- `RadioGroup`

**Additional context**

Not all of these components may need talons as they don't have logic warranting the refactor.

**Please let us know what packages this feature is in regards to:**
- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`

PR #2636, opened 2020-08-17
Add access to more talons

jpmii:
## Description

Add access to checkout's useOverview talon

## Related Issue

#1784

## Acceptance 

@awilcoxa @jimbo @zetlen


packages/peregrine/lib/targets/TalonWrapperConfig.js
```diff
@@ -33,6 +33,52 @@ class TalonWrapperConfig {
              */
             useApp: new WrappableTalon(addTransforms, 'App/useApp')
         };
+        /**
+         * Provides access to the talon used in Venia's Checkout UI component.
+         * @instance
+         */
+        this.Checkout = {
```
sirugh, line 40: These are legacy checkout components/talons. Should we indicate such in the property name?

For example, `this.LegacyCheckout`? Or are these property names meant to literally be the name of the path to the talon such as in `Checkout/useFlow`?

@zetlen I remember we had a discussion around naming but I don't remember this specific point.

packages/peregrine/lib/targets/TalonWrapperConfig.js
```diff
@@ -33,6 +33,52 @@ class TalonWrapperConfig {
...
+        /**
+         * Provides access to the talon used in Venia's SignIn UI component.
+         * @instance
+         */
+        this.SignIn = {
+            /**
+             * useSignin {@link WrappableTalon} object for the `useSignin()` talon
+             * @type {WrappableTalon}
+             */
+            useSignin: new WrappableTalon(addTransforms, 'SignIn/useSignin')
```
sirugh, line 80: Match casing - I believe this is `useSignIn`.

```suggestion
            useSignIn: new WrappableTalon(addTransforms, 'SignIn/useSignIn')
```

sirugh (changes requested):
Please run `prettier` and rebase!

packages/peregrine/lib/targets/TalonWrapperConfig.js
jimbo, line 40: In the latest versions of Venia, the drawer checkout and its talons here are essentially deprecated, in favor of `CheckoutPage`. Could we target `CheckoutPage` talons instead?

jimbo (changes requested):
Thanks for the straightforward contribution, and apologies for letting it slip through the cracks for a while. I think we'd prefer to expose the new checkout work rather than the old stuff, but otherwise I think we could move forward with this.

packages/peregrine/lib/targets/TalonWrapperConfig.js
jpmii: In the newest commit, I changed this to CheckoutPage. Thanks!

packages/peregrine/lib/targets/TalonWrapperConfig.js
```diff
@@ -33,6 +33,50 @@ class TalonWrapperConfig {
...
+             * useCheckoutPage {@link WrappableTalon} object for the `useCheckoutPage()` talon
+             * @type {WrappableTalon}
+             */
+            useCheckoutPage: new WrappableTalon(addTransforms, 'CheckoutPage/useCheckoutPage')
+        };
+        /**
+         * Provides access to the talon used in Venia's MiniCart UI component.
+         * @instance
+         */
+        this.MiniCart = {
```
sirugh, line 51: Since we're not wrapping legacy talons you can just remove this object.

jimbo: @sirugh Isn't this the new minicart though?

jimbo (approved):
Thanks for updating it. 🙂

packages/peregrine/lib/targets/TalonWrapperConfig.js
jpmii: > Please run `prettier` and rebase!

The latest commit should have the `prettier` results

packages/peregrine/lib/targets/TalonWrapperConfig.js
sirugh: Doh - you're right @jimbo. My mistake. Forgot we moved the legacy one to `LegacyMiniCart` :P

sirugh (approved):

larsroettig (approved):

tjwiebell:
Tracking internally with [PWA-956](https://jira.corp.magento.com/browse/PWA-956). Sending to QA.

Issue #2120, opened 2020-01-27
[feature]: Proposal Refactoring  for store configuration load


larsroettig:
With the current design of loading store configuration, in some cases, we are forced to have dependency between
loading store configuration and loading product data. With the proposal, I will give a short overview of which configuration we should load from Magento Backend. Also, how we can refactor the current state. 

### Current State 
Currently, we only consider the Magento store configuration for the **footer component** https://github.com/magento/pwa-studio/blob/5542122475b834cd0ae07ee7ccda2aeb4d65d57e/packages/venia-ui/lib/components/Footer/footer.js#L7

Shop owners can change many settings by default in Magento 2, and this helped Magento to grow so fast. From my point of view, we should be taking store configuration into account to change our behaviour in PWA Studio. We already have Tickets in our backlog that are required to loading store configuration see #1940. 
For that reason, it is now a good time to think about what are requirements and how we can implement this.  

## Currently Supported by PWA Studio 

- Copyright
- Media URL

### A good MVP for store configuration load

- Storename 
- Root Category ID
- Count products per page
- Locale options
- Base Currency
- Display  Currency
- Weight Unit 
- Language 

### My requirements to a good  store configuration load

If I think about how often as the customer will change the store name, Copywrite, etc. then we understand this should not so often happen. 

- Central where point we load configuration 
- Extensible with own configuration
- Cachable
- Fast 
- No query dependency

### 1. Load Store Configuration on build time

Currently, we already load only the secure media URL on build time. Loading  Store Configuration on build helps
we don't need to think about cache. On every build, we
load the data from Magento. 

**Advantage**

- No to thing about cache easy to implement 
- Central on the build we can generate a react config provider (env-specific) that helps us with  store config
-  Extensible at the build

**Disadvantages**

- Nees a redeploy the PWA to see store config changes 
- Load on the build will ingress the build time  
- Not scalable for a multi-store setup

#### Implementation Detail

*Methode that provides the config:*
The method `getMediaURL`need to be refactored to `getStoreConfig()`

https://github.com/magento/pwa-studio/blob/develop/packages/pwa-buildpack/lib/Utilities/graphQL.js#L35-L39

**Build Config:**
https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/webpack.config.js#L9

**GraphQL Query:**
https://github.com/magento/pwa-studio/blob/develop/packages/pwa-buildpack/lib/queries/getStoreMediaUrl.graphql

[...]

PR #2643, opened 2020-08-18
[PWA-811] Provide ability to translate Venia's header / footer

anthoula:
## Description

Provide the ability to store translation strings locally, pull correct locale from provided store view code, and translate the header and footer.

## Related Issue
https://jira.corp.magento.com/browse/PWA-811

### Verification Stakeholders
@jimbo 
@dpatil-magento 

### Verification Steps

1. Create new store in Magento with different language (e.g. French)
2. Update env variable to match store view code
3. Add a json file with translation strings in the `i18n` directory
4. Ensure strings in the header and footer are translated

## Checklist


packages/pwa-buildpack/lib/Utilities/graphQL.js
```diff
@@ -61,7 +62,7 @@ const getUnionAndInterfaceTypes = () => {
 };
 
 module.exports = {
-    getMediaURL,
+    getStoreConfigData,
```
davemacaulay, line 65: Is there a way of avoiding this breaking change? Might be better to retain the `getMediaURL` and include the `getStoreConfigData`.

packages/venia-concept/src/index.js
```diff
@@ -20,6 +21,21 @@ import { registerSW } from './registerSW';
...
+let messages;
+
+try {
+    messages = require(`@magento/venia-ui/lib/i18n/${language.replace(
+        '-',
+        '_'
+    )}.json`);
+} catch (err) {
+    console.warn(
+        `Unable to load translation file. Loading 'en_US' instead. \n${err}`
```
davemacaulay, line 34: I would be inclined to include which file path was attempted to be loaded here to help with debugging.

packages/venia-concept/src/index.js
```diff
@@ -110,10 +127,25 @@ const apolloLink = ApolloLink.from([
     Adapter.apolloLink(apiBase)
 ]);
 
+const onIntlError = error => {
+    if (error.code === 'MISSING_TRANSLATION') {
+        console.warn('Missing translation', error.message);
```
davemacaulay, line 132: Does `error.message` include sufficient information here? or do we need to supplement it? :)

packages/venia-concept/src/index.js
```diff
@@ -20,6 +21,21 @@ import { registerSW } from './registerSW';
...
 const apiBase = new URL('/graphql', location.origin).toString();
 
+const language = STORE_VIEW_LOCALE;
+let messages;
+
+try {
+    messages = require(`@magento/venia-ui/lib/i18n/${language.replace(
+        '-',
+        '_'
+    )}.json`);
```
davemacaulay, line 31: What do you think about making 2 utility functions? one to covert from Magento to `react-intl` and one to convert it back?

packages/venia-ui/lib/components/Header/__tests__/__snapshots__/cartTrigger.spec.js.snap
```diff
@@ -4,7 +4,7 @@ exports[`Cart counter displays 99+ when items quantity is more than 99 1`] = `
 Array [
   <div>
     <button
-      aria-label="Toggle mini cart. You have 100 items in your cart."
+      aria-label="Toggle mini cart. You have {count} items in your cart."
```
davemacaulay, line 7: Would we not expect for the test runner to convert the {count} variable into the actual integer here?

packages/venia-ui/lib/i18n/uk_UA.json
```diff
@@ -0,0 +1,40 @@
+{
```
davemacaulay, line 1: I love that you translated this to French and Ukrainian!

packages/pwa-buildpack/lib/Utilities/graphQL.js
larsroettig: I think we should think about https://github.com/magento/pwa-studio/issues/2120 if we need a breaking change also all developer will benefit oft this

larsroettig (changes requested):

packages/venia-concept/src/index.js
anthoula: Error messaging is pretty clear:
`Missing translation [React Intl Error MISSING_TRANSLATION] Missing message: "About Us" for locale "fr-FR", using id as fallback`

packages/pwa-buildpack/lib/Utilities/graphQL.js
anthoula: @davemacaulay Yes, getMediaURL has been restored

anthoula: @larsroettig Thanks for pointing that out

jest.config.js
```diff
@@ -189,7 +189,8 @@ const testReactComponents = inPackage => ({
                 ]
             }
         },
-        STORE_NAME: 'Venia'
+        STORE_NAME: 'Venia',
+        STORE_VIEW_LOCALE: 'en-US'
```
supernova-at, line 193: Should this be `STORE_VIEW_CODE`?

packages/venia-concept/src/index.js
```diff
@@ -16,10 +17,28 @@ import app from '@magento/peregrine/lib/store/actions/app';
...
+try {
+    messages = require(`@magento/venia-ui/lib/i18n/${fromReactIntl(
+        language
+    )}.json`);
+} catch (err) {
+    console.warn(
+        `Unable to load translation file. Loading 'venia-ui/lib/i18n/en_US.json' instead. \n${err}`
+    );
+    messages = require(`@magento/venia-ui/lib/i18n/en_US.json`);
+}
```
jimbo, line 40: I believe we should use [dynamic import](https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import) statements in the browser, rather than `require` statements. Note that these are asynchronous and that we can do some cool things with them, such as preloading.

Since I'm recommending that we switch from a synchronous (blocking) require to an async (non-blocking) one, that means we'll have to re-render `IntlProvider` when the call completes and we have real `messages` to give it. As a result, we should probably move this dynamic import into a component that wraps `IntlProvider`.

Finally, we should do as much of this as possible in `venia-ui` rather than `venia-concept`, I think. Making a component would be a good way to do that.

```es6
// shouldn't even have to provide a default, right?
// https://formatjs.io/docs/react-intl/components#defaultlocale-and-defaultformats
const locale = fromReactIntl(language)

const LocaleProvider = props => {
  const { children } = props
  const [messages, setMessages] = useState(null)

  // try doing the dynamic import as an effect
  // and set the result in state, so we can render before & after
  useEffect(
    () => {
      import(`@magento/venia-ui/lib/i18n/${locale}.json`)
        .then(data => { setMessages(data) })
        .catch(error => { console.error("Unable to load translation file.") })
    },
    [setMessages]
  )

  return (
    <IntlProvider
      key={language}
      {...props}
      defaultLocale="en-US"
      locale={language}
      messages={messages}
    />
  )
}

File: packages/venia-ui/lib/components/AccountMenu/accountMenuItems.js [hunk 9]
```diff
@@ -25,6 +16,25 @@ const AccountMenuItems = props => {
...
+        },
+        { name: intl.formatMessage({ id: 'Favorites Lists' }), url: '' },
+        { name: intl.formatMessage({ id: 'Address Book' }), url: '' },
+        { name: intl.formatMessage({ id: 'Saved Payments' }), url: '' },
+        {
+            name: intl.formatMessage({ id: 'Communications' }),
+            url: '/communications'
+        },
+        { name: intl.formatMessage({ id: 'Account Information' }), url: '' }
+    ];
```
jimbo, line 37: Yes, we do need to translate these, but we should probably continue to define `MENU_ITEMS` statically (before the component) and just call `formatMessage` from the same map operation we're already doing to create elements.

We might also take this opportunity to give these static items keys that more uniquely identify them within the app (e.g., `accountMenu.orderHistoryLink`), since that kinda seems to be what `react-intl` expects us to do. Thoughts?

```es6
const MENU_ITEMS = [/* ... */]

const AccountMenuItems = props => {
  // ...

  const { formatMessage } = useIntl()

  // could call `intl.formatMessage(item.name)` on each
  // but we might as well just use `FormattedMessage`, no?
  const menuItems = MENU_ITEMS.map(item => (
    <Link key={item.name} className={classes.link} to={item.url}>
      <FormattedMessage id={item.name} defaultMessage={item.name} />
    </Link>
  )
}
```

packages/venia-ui/lib/components/AccountMenu/accountMenuItems.js
```diff
@@ -26,9 +27,12 @@ const AccountMenuItems = props => {
     const classes = mergeClasses(defaultClasses, props.classes);
 
     const menuItems = MENU_ITEMS.map(item => {
+        const id = `account_menu.${item.name
+            .toLowerCase()
+            .replace(/ /g, '_')}_link`;
```
jimbo, line 32: This works, but I think I'd rather just add an `id` property to each object in `MENU_ITEMS` rather than generating it here. Also, I think we should probably move `MENU_ITEMS` to the talon, and return it (under the property `menuItems`). Would that work?

Looks much better with `<FormattedMessage id={id} />` though. 👍

packages/venia-ui/lib/components/AccountMenu/__tests__/__snapshots__/accountMenuItems.spec.js.snap
```diff
@@ -2,13 +2,13 @@
 
 exports[`it renders correctly 1`] = `
 <div>
-  &lt;Link&gt;Order History&lt;/Link&gt;
```
davemacaulay, line 5: Is this change intentional? Should we not still have a string represented in the snapshot?

packages/venia-ui/lib/components/AccountMenu/__tests__/__snapshots__/accountMenuItems.spec.js.snap
```diff
@@ -2,13 +2,13 @@
...
-  &lt;Link&gt;Saved Payments&lt;/Link&gt;
-  &lt;Link&gt;Communications&lt;/Link&gt;
-  &lt;Link&gt;Account Information&lt;/Link&gt;
+  &lt;Link&gt;[object Object]&lt;/Link&gt;
+  &lt;Link&gt;[object Object]&lt;/Link&gt;
+  &lt;Link&gt;[object Object]&lt;/Link&gt;
+  &lt;Link&gt;[object Object]&lt;/Link&gt;
+  &lt;Link&gt;[object Object]&lt;/Link&gt;
+  &lt;Link&gt;[object Object]&lt;/Link&gt;
+  &lt;Link&gt;[object Object]&lt;/Link&gt;
```
davemacaulay, line 11: Is this change intentional? Should we not still have a string represented in the snapshot?

packages/venia-ui/lib/components/Header/__tests__/__snapshots__/cartTrigger.spec.js.snap
davemacaulay: I guess this comment got lost, would be good to understand if the snapshots could contain the replacement patterns or if they should be rendered correctly. As we're mainly writing unit tests I would imagine we'd wanna mock our `react-intl` whenever we do not intend for it to translate.

packages/venia-ui/lib/components/Header/accountTrigger.js
```diff
@@ -28,17 +29,20 @@ const AccountTrigger = props => {
 
     const classes = mergeClasses(defaultClasses, props.classes);
     const rootClassName = accountMenuIsOpen ? classes.root_open : classes.root;
+    const intl = useIntl();
```
davemacaulay, line 32: `const { formatMessage } = useIntl();` would be a lot nicer here

davemacaulay: and probably in all other instances unless we're consuming other aspects of `useIntl` frequently.

packages/venia-ui/lib/components/AccountMenu/__tests__/__snapshots__/accountMenuItems.spec.js.snap
jimbo: I noticed this. I do think we need to mock something in order for snapshots to be accurate. The text replacement logic in `react-intl` components seems pretty robust, so this isn't too surprising.

jimbo (dismissed):

dpatil-magento:
@anthoula please check below observations - 

1. Fresh clone repo and checkout pr branch > yarn install > yarn buildpack create-env-file packages/venia-concept >>  .env >> check - STORE_VIEW_CODE=french  instead of default.
2. Sign in to account > Top right Continuous loading spinner, left drawer still shows login screen and account links are not translated.
image

3. Unable to add product to Cart > Graphql Error -  ‘Wrong store code specified for cart <id>’
4. In Admin UI even when French Store view is not default, how its able to fetch French store view?
image

dpatil-magento:
@anthoula @davemacaulay 
> @anthoula please check below observations -
> 
> 1. Fresh clone repo and checkout pr branch > yarn install > yarn buildpack create-env-file packages/venia-concept >>  .env >> check - STORE_VIEW_CODE=french  instead of default.

Looks good. I think now we should be good to remove entry from prod.dockerfile and dev.dockerfile ? Also, nice to update comment `# Default when not set: ` to `# Default when not set: default` ? 

Rest all previous observations are working.

8. There is change in yarn.lock when I install this branch, just need to commit it? pls check.

PR #2649, opened 2020-08-21
[PWA-808] Allow PWA Studio to use a different store to localize CMS Pages

davemacaulay:
## Description
Add ability to localize CMS Pages using the `STORE_VIEW_CODE` env variable

## Related Issue
https://jira.corp.magento.com/browse/PWA-808

### Verification Stakeholders
@jimbo 
@anthoula 
@eug123 
@dpatil-magento 

### Specification
- Developer is able to define the desired store for the application to use
- CMS Page data should be retrieve from the store

### Verification Steps
- Create a CMS Page in the Magento admin associated to a specific store view
- Modify the `STORE_VIEW_CODE` to point to that store
- Verify the CMS Page is loaded from the correct store
- Create a pair of CMS Pages with the same URL key between store views
- Modify `STORE_VIEW_CODE` to both store views and verify store changes after a restart of the `watch` command

## Checklist


packages/venia-ui/lib/RootComponents/CMS/cms.js
```diff
@@ -68,7 +73,7 @@ const CMSPage = props => {
     ) {
         content = (
             <Fragment>
-                <Title>{title}</Title>
+                <Title>{meta_title.length > 0 ? meta_title : title}</Title>
```
fooman, line 76:
```suggestion
                <Meta name="title" content={meta_title} />
```
But it looks like meta title seem to be on the way out https://moz.com/community/q/title-vs-meta-name-title-both-or-just-one-of-them.

davemacaulay: It looks as if Magento will always use the `meta_title` if present in both the `<title />` and `<meta name="title" />`. So I'll include the meta tag but use the same logic to determine which title to include.

image

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -13,7 +13,12 @@ const MESSAGES = new Map()
     .set(INTERNAL_ERROR, 'Something went wrong. Please try again.');
 
 const MagentoRoute = () => {
-    const talonProps = useMagentoRoute();
+    const magentoRouteProps = {};
+    // If we have a specific store view code configured pass it into the url resolver
+    if (process.env.STORE_VIEW_CODE) {
+        magentoRouteProps.store = process.env.STORE_VIEW_CODE;
+    }
```
eug123, line 20: Probably we can use recently added global `STORE_VIEW_CODE` - https://github.com/magento/pwa-studio/pull/2643/files#diff-1acbde853507ec5d5dac59ef5301a593R62

packages/peregrine/lib/talons/MagentoRoute/__tests__/useMagentoRoute.spec.js
```diff
@@ -51,7 +50,8 @@ const Component = props => {
 const routeComponentResults = {
     COMPONENT_FOUND: {
         component: {},
-        id: 1
+        id: 1,
+        store: 'default'
```
anthoula, line 54: Could also use jest global `STORE_VIEW_CODE`

https://github.com/magento/pwa-studio/pull/2643/files#diff-2d0cd5d10b9604941c38c6aac608178aR194

packages/venia-ui/lib/RootComponents/CMS/cms.js
```diff
@@ -49,35 +49,44 @@ const CMSPage = props => {
...
+
+    let pageContent;
     // Only render <RichContent /> if the page isn't empty and doesn't contain the default CMS Page text.
     if (
-        data.cmsPage.content &&
-        data.cmsPage.content.length > 0 &&
-        !data.cmsPage.content.includes('CMS homepage content goes here.')
+        content &&
+        content.length > 0 &&
+        !content.includes('CMS homepage content goes here.')
```
jimbo, line 72: This is probably a good opportunity to create a talon for this component, given that we already have them for the other root components. I don't think that really expands the scope here, just moves the code around. And it strikes me as the right thing to do. Thoughts?

Normally these kinds of derived booleans would be returned by a talon as `hasContent` and `isHomepage`, or some such.

packages/peregrine/lib/Router/magentoRouteHandler.js
```diff
@@ -28,19 +28,6 @@ export default class MagentoRouteHandler extends Component {
...
-            throw new Error(
-                'Current environment does not support CacheStorage at window.caches.'
-            );
-        }
-        const myCache = await window.caches.open(
-            `workbox-runtime-${location.origin}/`
-        );
-        await myCache.addAll(urls);
-    }
-
```
jimbo, line 43: Is this code obsolete now or did it move?

packages/peregrine/lib/Router/resolveUnknownRoute.js
```diff
@@ -5,15 +5,24 @@ import BrowserPersistence from '../util/simplePersistence';
...
-const routeCacheKey = 'urlResolver';
+
+/**
+ * Retrieve the route cache key for a store
+ *
+ * @param store
+ * @returns {*}
+ */
+function getRouteCacheKey(store) {
+    return 'urlResolver' + (store ? `_${store}` : null);
```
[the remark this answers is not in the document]
davemacaulay: Oops, that's a good catch. Thank you!

packages/peregrine/lib/Router/magentoRouteHandler.js
davemacaulay: Yeah I had a discussion with the wider team and they said this was part of some old workbox runtime which was since refactored out, but this legacy code was retained. So we agreed to remove it here.

packages/peregrine/lib/talons/MagentoRoute/__tests__/useMagentoRoute.spec.js
davemacaulay: This test doesn't need to access the global, as it's just a static string that needs to match up for the test to resolve correctly.

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
davemacaulay: Thanks for the heads up!

packages/venia-ui/lib/RootComponents/CMS/cms.js
davemacaulay: This code was always intended as a "hack" to get the Page Builder code live, I think rather than throwing the hack into a talon now we should have a future discussion on how to handle this correctly. As checking the contents of a string to conduct logic is generally a bad time and could easily produce incorrect results.

In my opinion PWA Studio should render "CMS homepage content goes here." rather than replacing this with the `<CategoryList />`, potentially this is an opportunity for us to build out the sample data for the default page (without Page Builder), to be able to remove this logic.

packages/peregrine/lib/Router/resolveUnknownRoute.js
```diff
@@ -103,12 +113,19 @@ function fetchRoute(opts) {
     url.searchParams.set('variables', JSON.stringify({ url: opts.route }));
     url.searchParams.set('operationName', 'ResolveURL');
 
+    const headers = {
+        'Content-Type': 'application/json'
+    };
+
+    // If the store is provided include it as part of the request
+    if (opts.store) {
+        headers['Store'] = opts.store;
```
sirugh, line 122: [This](https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/src/index.js#L29-L41) is how we attach auth token to all GQL requests. We could do the same for the `store` which is probably what we will end up needing to do for any queries made through ApolloClient.

davemacaulay:
> This approach certainly solves the immediate need which is to fetch different store results for the `urlResolver` query. Is there a follow up for the other queries?

@sirugh we're working iteratively, so to meet the A/C of this particular story that's the only query we needed to modify. We have follow on stories for categories & products so as we identify other queries we'll certainly update them.

We should certainly consider adopting Apollo for all GraphQL queries, it seems like there are a few relics using `fetch` which could be updated in future as part of a tech debt story.

sirugh:
Perfect :) I just wanted ya'll to be aware going forward.

packages/venia-ui/lib/RootComponents/CMS/cms.js
davemacaulay: As agreed I moved this logic into a talon.

jimbo:
I'll try this out tomorrow morning, but I think it'll be ready for QA tomorrow.

packages/peregrine/lib/talons/Cms/useCmsPage.js
```diff
@@ -0,0 +1,70 @@
...
+        };
+    }, [setPageLoading]);
+
+    // Ensure we mark the page as loading while we check the network for updates
+    useEffect(() => {
+        if (loading && !isPageLoading) {
+            setPageLoading(true);
+        } else if (!loading && isPageLoading) {
+            setPageLoading(false);
+        }
```
jimbo, line 47: This pattern works, but it's also overkill. If `setPageLoading` is called with the same value that's already in state, it doesn't count as a state change (i.e., doesn't cause an update and re-render). So these five lines can be one, and the dependencies don't need to include `isPageLoading`.

```es6
useEffect(
  () => { setPageLoading(loading) },
  [loading, setPageLoading]
)
```

davemacaulay: That's smart, thanks for explaining that!

PR #2658, opened 2020-08-27
Use region id when possible, otherwise use raw text value

sirugh:
## Description

The `createCustomerAddress` mutation requires different input for `region` depending on the `country`. If a country has `available_regions`, we should send the `region_id`. If it doesn't, we can send the raw text such as "Scotland".

To determine what to send we check the possible regions for the selected country at submission and send the formatted `region` input accordingly.

## Related Issue

Closes [PWA-874](https://jira.corp.magento.com/browse/PWA-874).

### Verification Steps

1. Go to Venia storefront and create an new account or Sign in to account which does not have address.
2. Add product to cart and go to checkout page.
3. Enter shipping address for United Kngdom.
4. Click on Save button.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/CheckoutPage/AddressBook/useAddressCard.js
+++ b/packages/peregrine/lib/talons/CheckoutPage/AddressBook/useAddressCard.js
@@ -29,15 +29,15 @@ export const useAddressCard = props => {
 
     const addressForEdit = useMemo(() => {
         const { country_code: countryCode, region, ...addressRest } = address;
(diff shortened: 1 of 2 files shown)


sirugh:
The current commit solves for submission of a new address, but editing an address that was submitted with a raw `region` field results in the `id` being rendered. In this case, the query returns `0` for the manually entered regions, regardless of how many you have entered.

[Image from Gyazo](https://gyazo.com/300948d743db09c3b6c827ef4d53ebf9)

PR #2660, opened 2020-08-27
[PWA-626] My Account: Order History: Table View (Collapsed)

tjwiebell:
## Description

As a shopper, when logged in and viewing the "order history" page I should see a list of my previous order with relevant details.

In-scope: 

Deliver initial "collapsed" table view for order history for both desktop and mobile 

Requirements: 

In Order History Table, add fields for:
◊ Order #
◊ Date
◊ Item Count
◊ Amount
◊ Status
Status Bar
The initial view of the order history page. View will contain basic information regarding the shoppers previous orders.
Create Empty State View 
Define error states and create appropriate messaging
 
Endstate:
A shopper when authenticated and having navigated to the Order History view, will see the standard (collapsed) view of the table/cards with basic information from the order

## Related Issue

* [[PWA-626](https://jira.corp.magento.com/browse/PWA-626)] My Account: Order History: Table View (Collapsed)

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
+++ b/packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
@@ -15,12 +15,18 @@ export const GET_CUSTOMER_ORDERS = gql`
             orders @client {
                 items {
                     id
+                    invoices {
+                        id
+                    }
                     items {
                         id
                         product_sku
                     }
                     number
                     order_date
+                    shipments {
+                        id
+                    }
                     status
                     total {
                         grand_total {
@@ -50,6 +56,8 @@ export const OrderHistoryTypeDefs = gql`
         number: String!
         items: [OrderItemInterface]
         total: OrderTotal
+        invoices: [Invoice]
+        shipments: [OrderShipment]
     }
 
     interface OrderItemInterface {
@@ -65,6 +73,14 @@ export const OrderHistoryTypeDefs = gql`
     type OrderTotal {
         grand_total: Money!
     }
+
+    type Invoice {
+        id: ID!
+    }
+
+    type OrderShipment {
+        id: ID!
+    }
 `;
 /* eslint-enable graphql/template-strings */
 
@@ -76,6 +92,7 @@ export const OrderHistoryResolvers = {
                 {
                     __typename: 'CustomerOrder',
                     id: 1,
+                    invoices: [{ __typename: 'Invoice', id: 1 }],
                     items: [
                         {
                             __typename: 'OrderItem',
@@ -103,9 +120,10 @@ export const OrderHistoryResolvers = {
                             product_sku: 'dress-5'
                         }
                     ],
-                    number: '167643',
+                    number: '000000002',
                     order_date: '2020-08-26 18:22:35',
-                    status: 'Shipped',
+                    shipments: [],
+                    status: 'Processing',
                     total: {
                         __typename: 'OrderTotal',
                         grand_total: {
@@ -118,6 +136,7 @@ export const OrderHistoryResolvers = {
                 {
                     __typename: 'CustomerOrder',
                     id: 2,
+                    invoices: [{ __typename: 'Invoice', id: 1 }],
                     items: [
                         {
                             __typename: 'OrderItem',
@@ -125,9 +144,10 @@ export const OrderHistoryResolvers = {
                             product_sku: 'pants'
                         }
                     ],
-                    number: '1477634',
+                    number: '000000005',
                     order_date: '2020-05-26 18:22:35',
-                    status: 'Processing',
+                    shipments: [{ __typename: 'OrderShipment', id: 1 }],
+                    status: 'Complete',
                     total: {
                         __typename: 'OrderTotal',
                         grand_total: {
--- a/packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
+++ b/packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
@@ -1,0 +1,150 @@
+import gql from 'graphql-tag';
+
+/*
+    This feature is being built ahead of GraphQL coverage that is landing in 2.4.1 of Magento. We're going to mock
+    the data based on the approved schema to make removing the mocking layer as seamless as possible.
+
+    @see https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/customer-orders.md
+ */
+
+/* eslint-disable graphql/template-strings */
+export const GET_CUSTOMER_ORDERS = gql`
+    query GetCustomerOrders {
+        customer {
+            id
+            orders @client {
+                items {
+                    id
+                    items {
+                        id
+                        product_sku
+                    }
+                    number
+                    order_date
+                    status
+                    total {
+                        grand_total {
+                            currency
+                            value
+                        }
+                    }
+                }
+            }
+        }
+    }
+`;
+
+export const OrderHistoryTypeDefs = gql`
+    extend type Customer {
+        orders: CustomerOrders
+    }
+
+    type CustomerOrders {
+        items: [CustomerOrder]!
+    }
+
+    type CustomerOrder {
+        id: ID!
+        order_date: String!
+        status: String!
+        number: String!
+        items: [OrderItemInterface]
+        total: OrderTotal
+    }
+
+    interface OrderItemInterface {
+        id: ID!
+        product_sku: String!
+    }
+
+    type OrderItem implements OrderItemInterface {
+        id: ID!
+        product_sku: String!
+    }
+
+    type OrderTotal {
+        grand_total: Money!
+    }
+`;
+/* eslint-enable graphql/template-strings */
+
+export const OrderHistoryResolvers = {
+    Customer: {
+        orders: () => ({
+            __typename: 'CustomerOrders',
+            items: [
+                {
+                    __typename: 'CustomerOrder',
+                    id: 1,
+                    items: [
+                        {
+                            __typename: 'OrderItem',
+                            id: '1',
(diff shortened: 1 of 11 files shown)


packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
```diff
@@ -0,0 +1,170 @@
...
+        id: ID!
+    }
+
+    type OrderShipment {
+        id: ID!
+    }
+`;
+/* eslint-enable graphql/template-strings */
+
+export const OrderHistoryResolvers = {
```
sirugh, line 87: I know we merged this in Wishlists but now that the v3 upgrade has been merged have you tried to do this with type policies and `useTypePolicies` hook?

tjwiebell: I have not, but was a partial reason for the draft PR 🙌  I merged mainline and only had to fix a couple imports, was going to ask you this morning if there's anything else I could fix to be more inline with v3. I'll give this a shot (and fix Wishlist if it goes smoothly).

sirugh: If you're only adding `@client` fields then you're probably OK but general advice to anyone working in the repo is to be on the lookout for apollo warnings such as normalization issues. v3 does a much better job of alerting us so hopefully we can catch issues quicker.

Additionally, when querying local fields I would say the main thing to be aware of is that there is a difference between an entirely client side query and a query that has both client and server props. 

```gql
export const GET_CUSTOMER_WISHLIST = gql`
    query GetCustomerWishlist {
        customer {
            id
            wishlists @client {
                id
                items_count
                name
                sharing_code
            }
        }
    }
`;
```

vs

```gql
export const GET_CUSTOMER_WISHLIST = gql`
    query GetCustomerWishlist {
        customer @client {
            id
            wishlists {
                id
                items_count
                name
                sharing_code
            }
        }
    }
`;
```

The former will always make a network request for `id` whereas the latter will not. [As written](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/WishlistPage/wishlistPage.gql.js#L15) you'll see a network request go out when you navigate to the customer wishlist page even though we don't need `id`. Of course, we may eventually need more fields like `name`, etc, so it might not matter.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
```diff
@@ -0,0 +1,170 @@
...
+    This feature is being built ahead of GraphQL coverage that is landing in 2.4.1 of Magento. We're going to mock
+    the data based on the approved schema to make removing the mocking layer as seamless as possible.
+
+    @see https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/customer-orders.md
+ */
+
+/* eslint-disable graphql/template-strings */
+export const GET_CUSTOMER_ORDERS = gql`
+    query GetCustomerOrders {
+        customer {
```
sirugh, line 13:
```suggestion
        customer @client {
```

Could move the directive from `orders` to here and prevent an unnecessary network request.

tjwiebell: Pushed without the tag on accident, fixed in 4c8f95a.

sirugh: Actually I was referencing the top-level `customer` field.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
tjwiebell: I experimented with this change and it introduced some odd behavior where data was only returned on the first request, every request after that was empty. Will need some help with this if this is the preferred pattern from Apollo.

Besides the above issue, I somewhat prefer the mixed usage and throwing `@client` on the field. This allows us to more easily test loading states while it's fetching `id`, giving us a more realistic render.

schensley:
@tjwiebell Looks great!  UX approved.

packages/venia-ui/lib/components/OrderHistoryPage/collapsedImageGallery.js
```diff
@@ -0,0 +1,49 @@
...
+
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const remainderCount = items.length - 4;
+
+    const imageElements = useMemo(() => {
+        if (imageData) {
+            const { products } = imageData;
+            const { items } = products;
+
+            const baseImageElements = items.map(item => {
```
revanth0212, line 26: Shouldn't this be `items.slice(0, 4).map` instead? You don't wanna map over all the items right? Because, after 4, you just want to show the count.

packages/venia-ui/lib/components/OrderHistoryPage/collapsedImageGallery.js
```diff
@@ -0,0 +1,49 @@
...
+    const talonProps = useCollapsedImageGallery({
+        items,
+        ...imageGalleryOperations
+    });
+    const { imageData } = talonProps;
+
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const remainderCount = items.length - 4;
+
+    const imageElements = useMemo(() => {
```
revanth0212, line 21: What do you think about breaking this `useMemo` into 2 parts, 1 for items and other for the count if `remainderCount > 0`?

Not mandatory, you can omit this comment if you wish to.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.js
```diff
@@ -1,32 +1,43 @@
...
+import { fullPageLoadingIndicator } from '../LoadingIndicator';
 import { Title } from '../Head';
 import defaultClasses from './orderHistoryPage.css';
+import orderHistoryOperations from './orderHistoryPage.gql';
+import OrderRow from './orderRow';
 
 const PAGE_TITLE = `Order History`;
 const EMPTY_DATA_MESSAGE = `You don't have any orders yet.`;
 
 const OrderHistoryPage = props => {
```
revanth0212, line 15: Can you update the prop types?

packages/venia-ui/lib/components/OrderHistoryPage/orderRow.js
```diff
@@ -0,0 +1,95 @@
...
+        {
+            year: 'numeric',
+            month: 'short',
+            day: 'numeric'
+        }
+    );
+
+    const hasInvoice = !!invoices.length;
+    const hasShipment = !!shipments.length;
+    const derivedStatus =
```
revanth0212, line 37: Wow, that looks convoluted at first.

packages/peregrine/lib/talons/OrderHistoryPage/useCollapsedImageGallery.js
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'react';
+import { useQuery } from '@apollo/client';
+
+export const useCollapsedImageGallery = props => {
```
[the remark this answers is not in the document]
tjwiebell: Added for this talon in a14f640, but man was it tedious. Is our goal to document everything up front, or did James have a script that generated it? I felt like I was just typing variable names in sentence format 😬

packages/venia-ui/lib/components/OrderHistoryPage/collapsedImageGallery.js
```diff
@@ -0,0 +1,49 @@
...
+    const { items } = props;
+
+    const talonProps = useCollapsedImageGallery({
+        items,
+        ...imageGalleryOperations
+    });
+    const { imageData } = talonProps;
+
+    const classes = mergeClasses(defaultClasses, props.classes);
+    const remainderCount = items.length - 4;
```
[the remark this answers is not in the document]
tjwiebell: Moved to constant in a14f640

packages/venia-ui/lib/components/OrderHistoryPage/collapsedImageGallery.js
tjwiebell: With the constant added in a14f640, you can see I'm cleverly only fetching the first four.

packages/venia-ui/lib/components/OrderHistoryPage/collapsedImageGallery.js
tjwiebell: This shouldn't be needed, we are not over fetching.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.js
tjwiebell: Updated all propTypes in a14f640.

packages/venia-ui/lib/components/OrderHistoryPage/orderRow.js
tjwiebell: Agreed. Would have loved if `status` was more helpful, but it stays in `Processing` pretty much the entire time; had to use these heuristics to get what UX wanted.

packages/peregrine/lib/talons/OrderHistoryPage/useCollapsedImageGallery.js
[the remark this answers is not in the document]
revanth0212: Yeah dude, it is tedious. I am not sure if there is a script to auto-generate or verify if something is missing.

revanth0212 (dismissed):

revanth0212 (dismissed):

revanth0212 (approved):

PR #2664, opened 2020-08-28
fix: Do not render the account menu elements when accountMenuIsOpen is false.

sirugh:
## Description

Because the `aside` element containing the account menu content was never unmounted, even when not "visible", previously executed mutations persisted their state. If you used the wrong email/password you would see an error, but if you closed (clicked out of) the dropdown and reopened it the error would still be visible.

It's possible this solution is too naive, as I don't know why we didn't just do it originally, but we can use the `accountMenuIsOpen` state to render the `aside` or not, which means that when it is false we will unmount the content, thus resetting the mutation/query state.

Note: Originally I unmounted the entire aside but then @jimbo reminded me that we don't want to do that because we lose animations. So I now only unmount the children which has the same outcome (so long as we have a min-height).

## Related Issue

Closes [PWA-841](https://jira.corp.magento.com/browse/PWA-841)

### Verification Steps

1. Try to log in with some random email/pw. See that it fails.
2. Close the dropdown by clicking outside or on the "Sign In" button.
3. Reopen it. The error should be gone.

[Image from Gyazo](https://gyazo.com/21d5a3c5d47bada13ac0ef0ddcac42a1)

## Checklist


revanth0212 (dismissed):
Simplest and best solution 👏

revanth0212:
Can you add a test to cover this change?

revanth0212 (dismissed):

PR #2665, opened 2020-08-29
Support configured URL suffix across entire app

### LINKED ISSUE #2644: [feature]: Support configured URL suffix across entire app
As a developer using PWA Studio, there's too many manual code changes required to support cleaner looking URLs that don't contain a hard-coded `.html` suffix. If all URLs were driven by configuration, I could simply just update my configuration in the backend to get the URL behavior I want. 

- Audit all places in the app that are currently hard-coding `.html` as a suffix (doing a code search for that string should be suitable enough to find all components that need fixed).
- Additionally, components that have already started using this configuration appear to have made a bad assumption about the response from GraphQL. It seems removing the suffix in the backend will result in `null` being returned, instead of an empty string, which is triggering some non-truthy logic and falling back to hard-coded suffix value. eg. [useBreadcrumbs](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js)

**Additional context**
We don't have a great solution for sourcing config once at the app root with mechanisms for busting the cache on value change, so it's reasonable for each component to query for this config as needed for now.

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

huykon:
## Description

As a developer using PWA Studio, there's too many manual code changes required to support cleaner looking URLs that don't contain a hard-coded .html suffix. If all URLs were driven by configuration, I could simply just update my configuration in the backend to get the URL behavior I want.

## Related Issue

Closes #2644.
Closes [PWA-872](https://jira.corp.magento.com/browse/PWA-872)

### Verification Steps

1. Open main left menu.
2. All url of category have suffix url get from storeConfig graphql API.
3. All url of product on site have suffix url from api.

## Checklist


packages/venia-ui/lib/components/Gallery/item.js
```diff
@@ -20,7 +20,6 @@ const IMAGE_WIDTHS = new Map()
     .set(UNCONSTRAINED_SIZE_KEY, 840);
 
 // TODO: get productUrlSuffix from graphql when it is ready
```
sirugh, line 22:
```suggestion
```

Delete this comment as it won't be necessary after your change.

huykon:
> This is really close! Can you please:
> 
> * Use the category url suffix in the [useBreadcrumb talon](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js#L37-L39).
> * Use the category url suffix in the [noProductsFound component](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/RootComponents/Category/NoProductsFound/noProductsFound.js#L12-L13).
> * Use the category url suffix in the [useCategoryTile talon](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/CategoryList/useCategoryTile.js#L3-L4).
> * Update/fix tests.
> 
> Let us know if you need any help with these requests.
> 
> Thanks again!

packages/venia-ui/lib/queries/getCategoryList.graphql
```diff
@@ -20,4 +20,8 @@ query categoryList($id: Int!) {
             }
         }
     }
+    storeConfig {
+        id
+        category_url_suffix
```
sirugh, line 25: We should use the `url_suffix` field off the category query response and not the store config, for now.

packages/venia-ui/lib/queries/getNavigationMenu.graphql
```diff
@@ -1,4 +1,8 @@
 query navigationMenu($id: Int!) {
+    storeConfig {
+        id
+        category_url_suffix
```
sirugh, line 4: We should use the `url_suffix` field off the category query response and not the store config, for now.

packages/venia-ui/lib/RootComponents/Category/NoProductsFound/noProductsFound.js
```diff
@@ -9,17 +9,14 @@ import { mergeClasses } from '../../../classify';
...
-
 const NoProductsFound = props => {
     const { recommendedCategories } = useNoProductsFound(props);
     const classes = mergeClasses(defaultClasses, props.classes);
 
     const categoryItems = useMemo(() => {
         return recommendedCategories.map(category => {
             const uri = resourceUrl(
-                `/${category.url_path}${categoryUrlSuffix}`
+                `/${category.url_path}${category.url_suffix}`
```
sirugh, line 19: Once you switch to using the `url_suffix` field on the category response this should actually work. I don't think this works as is because the catalog reducer does not know what to do with the `storeConfig` section of the response data.

sirugh:
>For useBreadcrumb talon we still have to use get storeConfig because the graphql don't have data for urlSuffix

You can modify the breadcrumb query to ask for the category suffix:

```
query getBreadcrumbData($category_id: Int!) {
    category(id: $category_id) {
        id
        breadcrumbs {
            category_id
            # We may not need level if `breadcrumbs` is sorted.
            category_level
            category_name
            category_url_path
        }
        id
        name
        url_path
        url_suffix
    }
}
```

packages/peregrine/lib/talons/Breadcrumbs/__tests__/__snapshots__/useBreadcrumbs.spec.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`return correct shape while data is loading 1`] = `undefined`;
+
+exports[`returns sorted data 1`] = `undefined`;
```
sirugh, line 5: Where's the sorted data? :P

packages/peregrine/lib/talons/Breadcrumbs/__tests__/__snapshots__/useBreadcrumbs.spec.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`return correct shape while data is loading 1`] = `undefined`;
```
sirugh, line 3: I would expect this to be an array. Can you check this test?

packages/peregrine/lib/talons/Breadcrumbs/__tests__/useBreadcrumbs.spec.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import { createTestInstance } from '@magento/peregrine';
+
+import { act } from 'react-test-renderer';
```
sirugh, line 4: This import isn't used.

```suggestion
```

revanth0212 (approved):

fooman:
I am guessing there may be some implication for the service worker here
https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/src/ServiceWorker/Utilities/routeHandler.js#L19
?

revanth0212:
> I am guessing there may be some implication for the service worker here
> https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/src/ServiceWorker/Utilities/routeHandler.js#L19
> ?

Hey @fooman, nice question, well caught. I have created a new ticket (PWA-912) in our backlog to handle this use case.

Thank you for pointing it out.

packages/venia-ui/lib/components/SearchBar/autocomplete.js
```diff
@@ -28,6 +28,7 @@ const GET_AUTOCOMPLETE_RESULTS = gql`
                     url
```
shahzad-hafeez, line 28: delete this file as we are not using it.

PR #2667, opened 2020-08-31
[PWA-809] Allow PWA Studio to use a different store to localize Catalog Products

eug123:
## Description

Product data should be retrieved from the defined store view and displayed correctly on the product page and category pages.
Product prices should be in default store display currency, and format should be localized according to store locale, 

## Related Issue

https://jira.corp.magento.com/browse/PWA-809
https://jira.corp.magento.com/browse/PWA-810

### Verification Steps

1. Create a Category and Product in the Magento admin
2. Configure Category and Product to have different data (description, meta title, etc ...) associated to a specific store view which has different locale setting
2. Add one more allowed currency for this store view
3. Change default display currency for this store. 
4. Modify the STORE_VIEW_CODE to point to that store
5. Verify the Category listing page is loaded from the correct store with all associated data localized.
6. Verify the Product details page is loaded from the correct store with all associated data localized.
7. Modify STORE_VIEW_CODE to both store views and verify store changes after a restart of the watch command
## Checklist


packages/peregrine/lib/Price/price.js
```diff
@@ -38,10 +38,10 @@ export default class Price extends PureComponent {
     };
 
     render() {
-        const { value, currencyCode, classes } = this.props;
+        const { value, currencyCode, classes, locale } = this.props;
```
davemacaulay, line 41: You need to add locale to `propTypes` above.

It should be required, but that'd be a breaking change due to the way propTypes works. So I think we has it as optional and default it to en-US if not supplied? What do you think?

packages/venia-ui/lib/components/CartPage/PriceSummary/giftCardSummary.ee.js
```diff
@@ -37,16 +37,23 @@ const getGiftCards = (cards = []) => {
...
-        <Fragment>
-            <span className={classes.lineItemLabel}>
-                {'Gift Card(s) applied'}
-            </span>
-            <span className={classes.price}>
-                {'-'}
-                <Price value={cards.value} currencyCode={cards.currency} />
-            </span>
-        </Fragment>
+        <IntlProvider locale="en-US">
```
davemacaulay, line 43: Is this intentional? This is an ee.js file not a spec.js file

eug123: You are right, it was done by accident. Fixed

Thank you

packages/peregrine/lib/Price/price.js
eug123: If it can stay optional, it probably can be `undefined` as before when not provided.

davemacaulay (approved):

package.json
```diff
@@ -68,6 +68,7 @@
     "figures": "~2.0.0",
     "first-run": "~2.0.0",
     "identity-obj-proxy": "~3.0.0",
+    "intl-locales-supported": "~1.8.12",
```
jimbo, line 71: What does this dependency do, and why does it belong in the root `package.json`?

packages/peregrine/lib/Price/price.js
jimbo: It's interesting that `locale` is a prop at all, though.

What was the reasoning behind making `locale` a prop here and calling `useIntl()` in every component that renders `Price`? As opposed to just calling `useIntl()` once within `Price`.

jimbo (approved):
Thanks for making changes. 👍

eug123:
@dpatil-magento this issue is caused by `cart` query error when store view in `Store`  request header is different from store view the cart is associated with. 
MC-37484 created to address this problem, but this issue will be more relevant when the visitor will have ability to switch store views.
At this point PWA website can be configured to use only one store view, so shopping carts will be associtated to this configured store view.

PR #2673, opened 2020-09-02
fix: apollo upgrade path

sirugh:
## Description

@revanth0212 discovered some issues when trying to mimic an upgrade from previous release (v7.0.0) to the latest (develop). The problem appears to be something to do with the way Apollo v3 resolves cached values from v2 cache values. It seems to manifest specifically around the queries/types for the [customer address book](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CheckoutPage/AddressBook/addressBook.gql.js).

Here are steps to reproduce:

1. `git checkout v7.0.0`
2. `yarn clean:all`
3. `yarn`
4. `yarn watch:all`
5. Log in as user with a saved address.
6. With a product in cart, visit `/checkout`.
7. Kill the watcher from step 4.
8. `git checkout develop`
9. `yarn clean:all`
10. `yarn`
11. `yarn watch:all`
12. Log in as user with a saved address.
13. With a product in cart, visit `/checkout`.

If you do the above steps, step 13 will throw errors. I don't believe it resolves after any number of refreshes unless you clear your local storage or the cache-persist entry and refresh. We can't expect users to do that, so we need to figure this out.

I've started making some changes to attempt to resolve it, but this is a work in progress.

You can also repro by copying the content of `apollo-cache-persist` to an editor after step 6. From then on you can reset your cache on develop, log in and go to /checkout, paste over the old cache in local storage, and refresh. This is probably easier so that you don't need to keep checking out the old branch and cleaning/reinstalling.

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Follow above steps but check out this PR instead of develop. When you refresh the page, it should not error.

You may see warnings related to merging of legacy references with new references - you can ignore these unless they continue to appear on subsequent renders:

[Image from Gyazo](https://gyazo.com/5844665c4ad90691b7f5002f46621487)

## Checklist


packages/venia-ui/lib/policies/index.js
```diff
@@ -59,7 +65,34 @@ const typePolicies = {
...
+        keyFields: () => 'Customer',
+        fields: {
+            addresses: {
+                merge(existing, incoming) {
+                    return incoming;
+                },
+                read(cachedAddresses, { toReference }) {
+                    if (cachedAddresses) {
+                        return cachedAddresses.map(address => {
+                            // Update v2 identifiers to new references. Previous
```
revanth0212, line 77: Nice comments. Good work buddy.

revanth0212 (approved):

supernova-at (approved):
Code looks good!

jimbo (approved):

dpatil-magento:
QA Approved.

PR #2680, opened 2020-09-05
[Fix bug]: Zip does not get cleared on switching country

### LINKED ISSUE #2678: [bug]: Zip does not get cleared on switching country
**Describe the bug**

Steps -
1. Add product to Cart and go to /cart page
2. Enter shipping details for USA and get available shipping methods.
3. Now change the country

Expected - State/Region and Zip should get cleared And Get Shipping Options button should displayed.
Actual - Zip does not get cleared.

Note -
1. This needs UX review. Attached screen has actual and expected UI.
2. Also Check if say user goes back to already selected country then show results instead of again showing "Get Shipping Options" button.

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

huykon:
## Description

Clear zip/Post code if re-select country at checkout shipping information

## Related Issue

Closes #2678.

### Verification Steps

1. Add product to Cart and go to /cart page
2. Enter shipping details for USA and get available shipping methods.
3. Now change the country.
4. Zip/Post code cleared

Repeat the steps above on the Checkout page Shipping Address section.

## Checklist


packages/venia-ui/lib/components/Postcode/postcode.css
```diff
@@ -0,0 +1,3 @@
+.root {
+    grid-area: state;
```
supernova-at, line 2: We don't want connections like this in standalone components.

This would be something that the calling component would overwrite for its instance of `Postcode` using the `classes` prop.

huykon: I updated to use `grid-area: zip` . Is that right to working?

jimbo (changes requested):
Hi @huykon , thanks for submitting another fix. 🙂  I have a few concerns right now.

#### Assumptions

As @supernova-at pointed out, this approach would build some assumptions into the `Postcode` component:

1. It will be rendered in a form that has also a field named `country`
2. It will be rendered in a grid container that defines a grid area named `zip`

Unfortunately, these are not safe assumptions. They may be true for one or more forms, but not necessarily for every form. A form might have a different name for its country field (or no such field at all), in which case this component would be unusable. Similarly, a form might use an implicit grid layout without named areas (or no grid layout at all).

#### Alternatives

Perhaps the form should be responsible for clearing this field instead. A form is responsible for knowing what fields it contains, and for maintaining any cross-field validation or side effects, so we could probably just attach a change handler to the form. See the pseudo-code below for an example.

```es6
const handleValueChange = useCallback(
  values => {
    if (countryHasChanged) setValue("postcode", "")
  },
  []
)

return (
  <Form onValueChange={handleValueChange}>
    {fields}
  </Form>
)
```

As for the CSS, we should be able to use the `postcode` classname in `CustomerForm` and `GuestForm`, right?

supernova-at:
Okay, I was able to spend some time messing around with this on the `/cart` page.

> It will be rendered in a form that has also a field named country

I see that this PR borrows patterns from other existing components (such as `Region`) that also bake in the assumption that they will be rendered alongside a form field named `country`.

> It will be rendered in a grid container that defines a grid area named zip

I know you said you made this change but it wasn't working. It should, so I will look into it deeper.

> Perhaps the form should be responsible for clearing this field instead.

I experimented with this and I think we can make `Form.onValueChange` work. There are a handful of gotchas and infinite loop potholes to avoid, but I think it makes a lot more sense for the Form to control these field resets instead of other fields - especially if we're refactoring fields out into their own standalone components as we're doing here.

<hr />

So all that said I think I'm good with this and I'll create a follow up ticket for the Forms themselves to handle this logic instead of the form fields (out of scope here, in my opinion).

I'll swing through and try to fix up the CSS for you 👍

supernova-at:
I have created an internal ticket to come back through and apply the changes requested previously.

supernova-at:
Okay so there are a few forms that are relevant here:

| Page | Form | Fix Applied |
| --- | --- | --- |
| Checkout | `GuestForm` (guest user) | ✅  |
| Checkout | `CustomerForm` (authenticated users. First time through and address book use the same form) | ✅  |
| Cart | All | ❌  |

huykon:
@supernova-at so will we don't need to use `Postcode` component on Cart page?

supernova-at:
> @supernova-at so will we don't need to use Postcode component on Cart page?

It would be nice if we did 😁 😉

schensley:
I've check the PR instance and it does in fact clear any data in State and ZIP field from the Shipping & Tax Estimator when another Country is selected from the Country selector drop down.  UX approved.

huykon:
Please recheck @jimbo @supernova-at

packages/venia-ui/lib/components/Postcode/postcode.js
```diff
@@ -0,0 +1,73 @@
...
+
+    useEffect(() => {
+        if (country) {
+            if (hasInitialized.current) {
+                postcodeInputFieldApi.reset();
+            } else {
+                hasInitialized.current = true;
+            }
+        }
+    }, [country, postcodeInputFieldApi]);
```
tjwiebell, line 43: Consider extracting to a talon

packages/venia-ui/lib/components/Postcode/postcode.js
```diff
@@ -0,0 +1,73 @@
...
+
+Postcode.propTypes = {
+    classes: shape({
+        root: string
+    }),
+    fieldInput: string,
+    label: string,
+    translationId: string,
+    validate: func,
+    initialValue: string
```
tjwiebell, line 72: I don't think we should double up on adding propTypes for Informed input props. These are just two that we happen to use, but this component supports them all.

packages/venia-ui/lib/components/Postcode/postcode.js
```diff
@@ -0,0 +1,73 @@
...
+        </Field>
+    );
+};
+
+export default Postcode;
+
+Postcode.defaultProps = {
+    fieldInput: 'postcode',
+    label: 'ZIP / Postal Code',
+    translationId: 'postcode.label'
```
tjwiebell, line 61: This API seems odd; `label` indicates to me it would override what's rendered, but it would actually still render the entry for that id in the language file. I think it's fine to expect a translated string here for overrides, and don't expose the translation api in the component.

tjwiebell (changes requested):
_Talonize_ this thing and cleanup the API, and we're good to go 👍

tjwiebell:
@supernova-at - I'm running into `Invariant Violation` exceptions on the deployed instance when switching country to France. Not sure if it's related to this change, but please look into that as well.

packages/venia-ui/lib/components/Postcode/postcode.js
huykon: should I remove that propType?

supernova-at:
> @supernova-at - I'm running into Invariant Violation exceptions on the deployed instance when switching country to France. Not sure if it's related to this change, but please look into that as well.

I can confirm the Invariant Violation exceptions. Let's fix these merge conflicts and confirm that it is this PR.

huykon:
Conflict resolved

packages/peregrine/lib/talons/Postcode/usePostcode.js
```diff
@@ -0,0 +1,50 @@
...
+ * @typedef {Object} PostcodeTalonProps
+ */
+
+/**
+ * @typedef {Object} Postcode
+ *
+ * @property {number} key the id of the region
+ * @property {String} label the label of the region
+ * @property {String} value the value of the region
+ */
```
tjwiebell, line 50: Are these docs outdated? I don't think I see this type used anywhere.

packages/venia-ui/lib/components/Postcode/postcode.js
```diff
@@ -25,27 +19,18 @@ const Postcode = props => {
...
-    const hasInitialized = useRef(false);
-    const countryCodeField = 'country';
-    const countryFieldState = useFieldState(countryCodeField);
-    const { value: country } = countryFieldState;
-
-    const postcodeInputFieldApi = useFieldApi(fieldInput);
+    const fieldLabel =
+        label ||
+        formatMessage({
+            id: 'postcode.label'
```
tjwiebell, line 25: I think we still want defaultMessage here.

```suggestion
            id: 'postcode.label',
            defaultMessage: 'ZIP / Postal Code'
```

tjwiebell (changes requested):
So close! Sending to QA, just some minor tweaks before delivery.

tjwiebell (dismissed):

dpatil-magento:
@supernova-at  Cart and Shipping form look good, but billing form still has the issue.

dpatil-magento:
Thanks @huykon Now Zip is being cleared in billing form also, but on clicking Review order, user is beingg blocked with error - 

image

huykon:
@dpatil-magento Thank for your report. I updated my code. Please check again

packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/useCreditCard.js
```diff
@@ -13,11 +13,12 @@ import { useCartContext } from '../../../context/cart';
  */
 export const mapAddressData = rawAddressData => {
     if (rawAddressData) {
+        console.log(rawAddressData);
```
sirugh, line 16: Remove log

huykon: Removed

dpatil-magento:
@huykon When billing is same as shipping :( , please check.

image

huykon:
@dpatil-magento updated

supernova-at (dismissed):
I'm curious why we were aliasing `postcode` to `postalCode` in the first place. I think is simpler for sure 👍

huykon:
@supernova-at `postalCode` is using at summary now. It's very weird, why we need to use that variable

dpatil-magento:
QA Approved

supernova-at (approved):

PR #2683, opened 2020-09-07
Update index.md

rubenRP:
## Description

Replaces old components reference with new HomePage component reference.

### Verification Steps
1. Check that lib/Routes/routes.js components references are different and are not updated.


larsroettig:
@jcalcaben I think we need to close we already updated

jcalcaben:
This topic has already been updated

PR #2685, opened 2020-09-08
Header Search Dropdowns

supernova-at:
## Description

This PR updates the Search trigger button in the site header to behave like my account and cart trigger buttons for consistency.
This also solves the issue of two dialogs overlapping each other.

It also:
  * Removes `searchOpen` and its corresponding `toggleSearch` action from global app state (Redux). The search trigger controls this piece of state entirely now.
  * Treats the search input (`SearchBar` component) as a `dropdown`, triggered by `SearchTrigger`.
  * Preserves search input behavior such as focusing it automatically when search is opened
  * No longer turns the Search Trigger text blue when the search dropdown is open (consistency).
  * Adds a "Showing results for {search term}" message on the search page

## Related Issue

Closes PWA-786.

### Verification Stakeholders

@schensley 

### Verification Steps

1. Verify that you cannot have both the Search and My Account (or MiniCart) triggers open at the same time

#### No Loss of Functionality

For this section, compare this PR deployment with https://develop.pwa-venia.com/.

1. Verify that opening search works the same
1. Verify that closing search works the same except that now, clicking outside of the search area or the search trigger will close search as well
1. Verify that autocomplete works
1. Verify that you can execute a search

On the Search Results page (`/search.html`):

1. Verify that opening the search text field pre-populates it with the last executed search term (if any)
1. Verify that the "Showing results for..." message is accurate, both with and without a category (get a category by clicking a suggested category in the autocomplete dropdown)

## Checklist


packages/peregrine/lib/talons/Header/useHeader.js
```diff
@@ -1,21 +1,39 @@
...
     const handleSearchTriggerClick = useCallback(() => {
-        toggleSearch();
-    }, [toggleSearch]);
+        // Toggle the Search input form.
+        setIsSearchOpen(isOpen => !isOpen);
+    }, [setIsSearchOpen]);
+
+    // If we're on the search page, ensure the search textbox is open to begin with.
+    useEffect(() => {
+        if (location.pathname === '/search.html') {
```
supernova-at, line 25: I'm not a fan of `Header` caring about what page it's on but this is what enables us to remove `searchOpen` and `toggleSearch` from Redux.

Previously, this effect was running on the search page. Because it needs to control whether search is open or not, we had to hoist that state into global app state. With this effect being here instead, we no longer have to share that state.

schensley:
Thank you @supernova-at this all looks great. Problem solved. UX approved.

packages/peregrine/lib/talons/Header/useHeader.js
tjwiebell: I think we wanted to spike on this before committing, but this could be an Apollo @client variable. You could additionally query for this in AppContext and distribute that way, while also providing an action for toggling; we could safely delete the Redux state/action this way, since we've provided a backwards compatible replacement.

tjwiebell (dismissed):
Dropdown behavior seems odd on the search page. Provided some suggestions, but I'm going to request a second reviewer on this since it did pass UX review.

packages/peregrine/lib/talons/Header/useHeader.js
jimbo: I'd be fine with removing this, it's a pretty arcane requirement in the first place. We can probably negotiate it with UX.

If it has to stay, the location check is fine. Eventually we'll want a utility function to abstract the whole "what page are we on?" inference using location and some other data/heuristics.

supernova-at:
@schensley @tjwiebell I tried something. Not sure if I like it yet, so I submit it to you for UX approval:

I added a dedicated search input to the search page. Always there, don't have to mess with the header. In fact, I disabled the one from the header when you're on the search page so you can't have two.

Let me know how it feels to you.

supernova-at:
I messed up the commit history trying to remove some commits but the diffs look good so I think this is everything.

packages/peregrine/lib/talons/Header/useHeader.js
supernova-at: Ended up removing this entirely as Jimmy suggested.

packages/peregrine/lib/talons/Header/useHeader.js
```diff
@@ -1,21 +1,39 @@
...
-        { hasBeenOffline, isOnline, isPageLoading, searchOpen },
-        { toggleSearch }
-    ] = useAppContext();
+    const [{ hasBeenOffline, isOnline, isPageLoading }] = useAppContext();
+    const {
+        elementRef: searchRef,
+        expanded: isSearchOpen,
+        setExpanded: setIsSearchOpen,
+        triggerRef: searchTriggerRef
+    } = useDropdown();
```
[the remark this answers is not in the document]
supernova-at: Great feedback. The search input now pre-populates with the search term and we've added a "Showing results for..." heading on the search page to drive the point home.

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -103,6 +104,10 @@ const SearchPage = props => {
                 </div>
                 {maybeSortContainer}
             </div>
+            <div className={classes.heading}>
+                {'Showing results for '}
+                <span className={classes.headingHighlight}>{searchTerm}</span>:
```
supernova-at, line 109: Taking a value out of the URL and displaying it on the page would normally open us up to XSS but [React / JSX takes care of this for us](https://reactjs.org/docs/introducing-jsx.html#jsx-prevents-injection-attacks).

packages/venia-ui/lib/i18n/en_US.json
```diff
@@ -34,8 +34,7 @@
     "Product Suggestions": "Product Suggestions",
     "Register": "Register",
     "Returns": "Returns",
-    "searchPage.searchTermLabel": "Showing results for",
-    "searchPage.searchTermValue": "{term} in {category}",
+    "searchPage.searchTerm": "Showing results for <highlight>{term}</highlight>{category, select, null {} other {<highlight> in {category}</highlight>}}:",
```
supernova-at, line 37: Well this one was fun 😅 .

It's either displaying "Showing results for {searchTerm}" or "Showing results for {searchTerm} in {category}".

Additionally, the "{searchTerm}" / "{searchTerm} in {category}" text should be emphasized (bolded) on the page.  See this definition and the accompanying `searchPage.searchTerm` `FormattedMessage` in `searchPage.js` on how to accomplish both of these at the same time.

Note that some languages might have the emphasized text / search term elsewhere in the text, so we can't just translate "Search results for" separately.

sirugh: Also, nitpicking, but should `in` be highlighted?



[Image from Gyazo](https://gyazo.com/8851d6e9b3bd6cc6060169e7da21a645)

packages/venia-ui/lib/components/Header/searchTrigger.js
```diff
@@ -27,14 +29,15 @@ const SearchTrigger = props => {
                 defaultMessage: 'Search'
             })}
             onClick={handleClick}
+            ref={ref}
         >
             <Icon src={SearchIcon} />
             <span className={classes.label}>
                 <FormattedMessage id={'Search'} />
```
sirugh, line 36: You didn't add this, but while you're in here can you fix the warning:

[Image from Gyazo](https://gyazo.com/f534ea206066ba7b3a932825053650a9)

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -25,46 +26,51 @@ export const useSearchPage = props => {
...
+        if (inputFilters.size === 0) {
+            return null;
+        }
+
+        const targetCategoriesSet = inputFilters.get('category_id');
+        if (!targetCategoriesSet) {
+            return null;
+        }
+
+        console.log('the set', targetCategoriesSet);
```
sirugh, line 62: Whoops!

```suggestion
```

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -25,46 +26,51 @@ export const useSearchPage = props => {
...
+
+        // The set looks like ["Bottoms,11", "Skirts,12"].
```
sirugh, line 64:
```suggestion
        // The set looks like ["Bottoms,11", "Skirts,12"]. Split and return "Bottoms, Skirts", etc.
```

sirugh (changes requested):
Nice work! I like the discovery done with formatted message and styling :)

Also I think the "X in Y" string is a good replacement for keeping the search field open after searching.

packages/venia-ui/lib/i18n/en_US.json
supernova-at: > wouldn't it have been easier to just conditionally render the strings based on category truthiness?

I actually did this initially, lol. I had two separate translations, one for "Showing results for" and then one for "{term} in {category}".

But I don't think that is the right way to do it because some languages will structure their sentences differently, where the {term} and {category} might be mid-sentence. I don't have any concrete examples but I do remember running into this problem the last time I worked on a localized app (at BV).

supernova-at: More interesting might be the idea of two separate translations, one for "Showing results for {term}" and another for "Showing results for {term} in {category}".

Then the component would decide which _**`id`**_ to use based on the truthiness of `category`.

packages/venia-ui/lib/i18n/en_US.json
```diff
@@ -1,6 +1,14 @@
 {
+    " in {label}": " in {label}",
```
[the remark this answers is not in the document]
supernova-at: Lol yeah that's the autocomplete translation. I didn't mess with it for scope reasons but it is kind of related.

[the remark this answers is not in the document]
supernova-at: Screen Shot 2020-09-17 at 10 24 29 AM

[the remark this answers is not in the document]
supernova-at: I should also note I didn't add this, I just moved it (alphabetized this object)

packages/venia-ui/lib/i18n/en_US.json
supernova-at: > should in be highlighted?

Ooh, I like it! https://github.com/magento/pwa-studio/pull/2685/commits/ccf898cc9e403981db502b41a747c39b2b0dd3a5

packages/peregrine/lib/store/actions/app/asyncActions.js
```diff
@@ -6,5 +6,6 @@ export const toggleDrawer = name => async dispatch =>
 export const closeDrawer = () => async dispatch =>
     dispatch(actions.toggleDrawer(null));
 
+/** @deprecated in PWA Studio v8.0.0 (Magento 2.4.1). */
```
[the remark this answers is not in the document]
supernova-at: Good call, https://github.com/magento/pwa-studio/pull/2685/commits/e989d46ad97bd8c4ef2e6db1e27472c0e0e0a4a3.

I was trying for some way for us to know when we can go back and actually delete these things. I'll create an audit ticket or something instead.

packages/venia-ui/lib/components/Header/searchTrigger.js
supernova-at: https://github.com/magento/pwa-studio/pull/2685/commits/a2e1055bce4061c684505e4fd13403d1a6180ff6

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
supernova-at: 😅  I really need to write myself a lint rule for this.

https://github.com/magento/pwa-studio/pull/2685/commits/ce200bfdd87aa9a9a4ff98dfb631f2a8a64d8298

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
supernova-at: https://github.com/magento/pwa-studio/pull/2685/commits/ce200bfdd87aa9a9a4ff98dfb631f2a8a64d8298

sirugh (dismissed):

jimbo (commented):
> This PR updates the Search trigger button in the site header to behave like my account and cart trigger buttons for consistency. This also solves the issue of two dialogs overlapping each other.

Consistency is good, so this is a good effort.

> Removes searchOpen and its corresponding toggleSearch action from global app state (Redux). The search trigger controls this piece of state entirely now.

Great. If we decide at any point that it does need to belong in global state, we can use Apollo.

> Treats the search input (SearchBar component) as a dropdown, triggered by SearchTrigger.

I'm not wild about the difference between clicking away from the autocomplete by clicking the body (collapses everything, discards input and results) versus the header (collapses autocomplete only, discards nothing). This is basically still a dropdown within a dropdown. But it's an improvement over live.

> Preserves search input behavior such as focusing it automatically when search is opened
> No longer turns the Search Trigger text blue when the search dropdown is open (consistency).
> Adds a "Showing results for {search term}" message on the search page

These are all good improvements.

---

After seeing this in action, I agree with @tjwiebell about the experience here; it could be better. The two-row header is still not a great design choice, and I'd prefer the search trigger to simply be replaced by the search input (so the autocomplete results would be the only dropdown part), but that's far from consistent with the other pieces and likely out of scope here.

I could approve this as is.

dpatil-magento:
@supernova-at Please check below observations - 
1. If you hit https://pr-2685.pwa-venia.com/search.html?  directly then  text says `Showing results for :` ,  does sound odd.
2. On mobile (s10e)  , texts are wrapping.
image


3. After search if user click `sort` `filter` button then search text box is closing but respective drop down does not open for first time, user has to click one more time to open sort/filter.  Also I didn't like search text field closing automatically like after clicking outside (just my opinion :) )

supernova-at:
> If you hit https://pr-2685.pwa-venia.com/search.html? directly then text says Showing results for : , does sound odd.

Nice catch. Resolved in https://github.com/magento/pwa-studio/pull/2685/commits/c2d9951044cc0e11ab4a86ad2373ad77198cb816.

> On mobile (s10e) , texts are wrapping.

In progress / TBD

> After search if user click sort filter button then search text box is closing but respective drop down does not open for first time, user has to click one more time to open sort/filter.

This one is actually pretty funny: it happens because the Filter and Sort buttons move on the page (because the search disappears). So you mousedown on them, but by the time you mouseup the button has moved and you don't get a click event (click = mousedown + mouseup).

I think we've all agreed over the course of this PR that we don't like the search "dropping down" like this, and that it would be better if it were inline. UX already has preliminary designs for this, but we deemed it out of scope here.

Two possible solutions:
1. Implement UX's preliminary designs now (increase scope of this PR or create fast follow ticket)
1. Attach event handler to Filter / Sort `mousedown` instead of `click`

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -94,6 +94,24 @@ const SearchPage = props => {
...
+                    <span className={classes.headingHighlight}>{chunks}</span>
+                ),
+                category: searchCategory,
+                term: searchTerm
+            }}
+        />
+    ) : (
+        <FormattedMessage
+            id={'searchPage.searchTermEmpty'}
+            defaultMessage={'Showing all results:'}
```
sirugh, line 111: Shouldn't need the default here since the `id` directly correlates to the string translation. You need defaults when there is a template string/replacement in the json.

supernova-at: https://github.com/magento/pwa-studio/pull/2685/commits/7a888d7fab8024edb7344f26f40776677e1ed921

packages/venia-ui/lib/components/ProductSort/productSort.js
```diff
@@ -76,6 +76,7 @@ const ProductSort = props => {
                 priority={'low'}
                 classes={{ root_lowPriority: classes.sortButton }}
                 onMouseDown={handleSortClick}
+                onTouchStart={handleSortClick}
```
sirugh, line 79: I believe this is the first and only time we've done something like this in the app. You should add a comment above the return explaining that this is temporary and will be replaced by `onClick={handleSortClick}` once the search bar stops shifting layout on blur.

supernova-at: I created a ticket and called out this specific line to be changed but I can add something in code too so the pattern isn't repeated elsewhere

supernova-at: Nevermind! 🚨 

Instead of moving these handlers "forward" (`click` -> `mousedown`), in https://github.com/magento/pwa-studio/pull/2685/commits/7900de96f68dc90559ef720935424aef7d91adb0 I instead moved the `useDropdown` handler "back" (`mousedown` -> `click`).

I think this is preferable because it has fewer side effects throughout the rest of the codebase (like the lines commented here).

I'd like to verify with @jimbo any potential negative side effects of this change but clicking around in the five places we use `useDropdown` right now, it works great.

I should also note that I had to set the `capture` option of `useDropdown`'s click handler to `true` otherwise it was possible to get the search and my account dropdowns open at the same time (the original bug). Unfortunately I can't _really_ explain why that would be the case. It seems like it should work even if `capture` was `false`.

revanth0212 (dismissed):

supernova-at:
@dpatil-magento I have reverted the change to `useDropdown` following discussion in group code review.

This means that

> After search if user click sort filter button then search text box is closing but respective drop down does not open for first time, user has to click one more time to open sort/filter.

has returned.

We have a ticket and design for the search bar not re-laying out the entire page, which is the actual fix for that ^ problem.
We didn't want to introduce a temporary fix that we knew we were going to have to rip out later once the new design lands.

dpatil-magento:
QA Approved, @supernova-at please take care of conflict :)

revanth0212 (approved):

PR #2686, opened 2020-09-08
[PWA-814] Enable shoppers to change locale using a store view switcher

eug123:
## Description

Adding store switcher component to allow visitor to switch store views and get content translated to respective store view locale.

## Related Issue

https://jira.corp.magento.com/browse/PWA-814

### Verification Stakeholders

@jimbo
@sirugh 
@dpatil-magento 
### Verification Steps

1. Configure Magento to have more then one store view
2. See store switcher appeared in the PWA header for desktop / Side navigation for mobile screen
3. Use switcher to change store view/locale
4. Verify that data been retrieved according to selected store view

## Checklist


packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -0,0 +1,90 @@
...
+
+        return filteredData;
+    }, [availableStoresData]);
+
+    const handleSwitchStore = useCallback(
+        // Refresh shopping cart
+        async storeCode => {
+            const locale = availableStores[storeCode].locale;
+
+            await storage.setItem('store_view', {
```
sirugh, line 65: I would replace this with a `writeQuery`

packages/venia-concept/src/index.js
```diff
@@ -32,11 +32,20 @@ const authLink = setContext((_, { headers }) => {
...
     const token = storage.getItem('signin_token');
 
+    if (!storage.getItem('store_view')) {
+        storage.setItem('store_view', {
+            code: STORE_VIEW_CODE,
+            locale: STORE_VIEW_LOCALE
+        });
+    }
+
+    const storeCode = storage.getItem('store_view').code;
```
sirugh, line 42: Remove the default state above, preferably.

```suggestion
    const storeCode = storage.getItem('store_view').code || STORE_VIEW_CODE;
```

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -5,18 +5,24 @@ import {
...
 const MagentoRoute = () => {
+    const { formatMessage } = useIntl();
     const magentoRouteProps = {};
     // If we have a specific store view code configured pass it into the url resolver
-    if (STORE_VIEW_CODE) {
-        magentoRouteProps.store = STORE_VIEW_CODE;
+    const storage = new BrowserPersistence();
+
+    if (storage.getItem('store_view')) {
+        magentoRouteProps.store = storage.getItem('store_view').code;
```
sirugh, line 25: `useQuery(getStoreViewCode)`

packages/venia-ui/lib/components/Header/storeSwitcher.js
```diff
@@ -0,0 +1,124 @@
...
+                    <ul>{stores}</ul>
+                </div>
+            </div>
+        ) : null;
+
+        if (hasMultipleStores) {
+            window.document.documentElement.style.setProperty(
+                '--header-height',
+                '7.5rem'
+            );
```
sirugh, line 100: There's gotta be a better way. Have you run the implementation by UX first and foremost? If this UX is what they want, then that means we need some way to conditionally use either `5rem` or `7.5rem`. Feels odd to do this in the component though.

Also, if we're going to keep this, we should use an effect, on mount, so that it does not continue to call the setter every render.

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -7,17 +7,18 @@ import {
 } from '@magento/peregrine/lib/talons/MagentoRoute';
 
 import { fullPageLoadingIndicator } from '../LoadingIndicator';
+import { useIntl } from 'react-intl';
```
sirugh, line 10: Third party deps should be near the top of the imports.

Can you make sure this change is applied throughout the changeset?

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -34,14 +35,14 @@ const MagentoRoute = () => {
     } else if (routeError === NOT_FOUND) {
         return (
             <ErrorView>
-                <h1>{MESSAGES.get(routeError)}</h1>
+                <h1>{formatMessage({ id: MESSAGES.get(routeError) })}</h1>
```
sirugh, line 38: This will work because `formatMessage` falls back to the `id` if there is no defined message in the json. That said, I think this should actually be something like:

```js
const errorMessage = formatMessage({ 
  id: routeError,
  defaultMessage: MESSAGES.get(routeError);
});

...
<h1>
  {errorMessage}
</h1>
```

And you'd define the mapping of route error within in the translation json. You'd probably need to modify the keys to be something like `magentoRoute.notFoundError`. etc.

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -34,14 +35,14 @@ const MagentoRoute = () => {
...
             </ErrorView>
         );
     }
 
     return (
         <ErrorView>
-            <h1>{MESSAGES.get(INTERNAL_ERROR)}</h1>
+            <h1>{formatMessage({ id: MESSAGES.get(INTERNAL_ERROR) })}</h1>
```
sirugh, line 45: Same idea here. We can use the map strings as a default message but should also make sure they are in the json.

packages/venia-ui/lib/components/Navigation/navigation.js
```diff
@@ -63,13 +63,13 @@ const Navigation = props => {
...
                 <CategoryTree
                     categoryId={categoryId}
-                    categories={categories}
                     onNavigate={handleClose}
                     setCategoryId={setCategoryId}
                     updateCategories={catalogActions.updateCategories}
                 />
             </div>
             <div className={classes.footer}>
+                <StoreSwitcher mobileView={true} />
```
sirugh, line 72: Usually we indicate mobile using media queries in css. I think it makes more sense to do it that way anyways. The component doesn't seem too complex, just a horizontal bar with an option/select.

packages/venia-ui/lib/components/Header/header.css
```diff
@@ -1,3 +1,7 @@
+:root {
+    --header-height: 5rem;
```
sirugh, line 2: Do we define other global variables in such a way? I'd like to talk about this in dev time.

cc @jimbo

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -37,8 +40,9 @@ const SearchPage = props => {
     } = talonProps;
 
     const [currentSort] = sortProps;
+    useScrollTopOnChange(pageControl.currentPage);
```
sirugh, line 43: This should occur in the talon.

packages/venia-ui/lib/i18n/en_US.json
```diff
@@ -37,5 +37,7 @@
     "accountMenu.addressBookLink": "Address Book",
     "accountMenu.savedPaymentsLink": "Saved Payments",
     "accountMenu.communicationsLink": "Communications",
-    "accountMenu.accountInfoLink": "Account Information"
+    "accountMenu.accountInfoLink": "Account Information",
+    "That page could not be found. Please try again.": "That page could not be found. Please try again.",
+    "Something went wrong. Please try again.": "Something went wrong. Please try again."
```
sirugh, line 42: Ah, yes, here I would use id format like the above `component.thing` rather than a string.

sirugh (changes requested):

packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
```diff
@@ -27,9 +23,11 @@ export const useCategoryContent = props => {
     const {
         categoryId,
         data,
+        pageSize,
```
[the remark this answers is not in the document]
eug123: Done, 
Thank you

packages/peregrine/lib/talons/CategoryTree/useCategoryTree.js
```diff
@@ -1,52 +1,45 @@
 import { useEffect, useMemo } from 'react';
 import { useLazyQuery } from '@apollo/client';
-
```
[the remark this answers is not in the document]
eug123: Fixed, 
Thank you

packages/peregrine/lib/talons/CategoryTree/useCategoryTree.js
```diff
@@ -58,16 +51,14 @@ export const useCategoryTree = props => {
...
-        for (const id of children || '') {
-            const category = categories[id];
+        children.map(category => {
             const isLeaf = !parseInt(category.children_count);
-
-            childCategories.set(id, { category, isLeaf });
-        }
-
+            childCategories.set(`${category.id}-${category.name}`, {
```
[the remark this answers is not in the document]
eug123: Changed,
Thank you

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -0,0 +1,90 @@
...
+
+    const handleSwitchStore = useCallback(
+        // Change store view code and currency to be used in Appollo link request headers
+        async storeCode => {
+            await storage.setItem('store_view', {
+                code: storeCode,
+                currency: availableStores[storeCode].currency
+            });
+
+            history.go(0);
```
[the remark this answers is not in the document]
eug123: Comment added,
Thank you

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
eug123: Done

packages/venia-ui/lib/components/Navigation/navigation.js
eug123: This allows not having two switchers in the same time in header and navigation panel.

packages/venia-ui/lib/components/Header/header.css
eug123: Change to be header local variable and value is set in header component and depends on header inner layout.

packages/venia-ui/lib/components/SearchPage/searchPage.js
eug123: Done

packages/venia-ui/lib/i18n/en_US.json
eug123: Done,
Thank you

sirugh: Thanks for updating it - seems like we're still going to be figuring this out for a while, but I don't think you need to change it again in your PR.

packages/venia-ui/lib/components/Header/header.css
sirugh: Ah I see. You did it like this:

```
const style = {
        ...(switchers && { '--header-height': '7.5rem' })
    };

    return (
        <header className={rootClass} style={style}>
        ...
```

packages/venia-ui/lib/components/Navigation/navigation.js
sirugh: Is having two switchers rendered a concern?

packages/venia-ui/lib/components/Header/switcherItem.js
```diff
@@ -0,0 +1,39 @@
...
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    const handleClick = useCallback(() => {
+        onClick(switcherItem.code);
+    }, [switcherItem, onClick]);
+
+    const activeIcon = active ? <Icon size={20} src={Check} /> : null;
+
+    return (
+        <button className={classes.root} onClick={active ? null : handleClick}>
```
sirugh, line 20: Rather than change the click handler you can just set `disabled={active}`

packages/venia-ui/lib/components/Navigation/navigation.js
```diff
@@ -51,6 +51,12 @@ const Navigation = props => {
         </Suspense>
     ) : null;
 
+    const switchers = StoreSwitcher ? (
+        <div className={classes.switchers}>
+            <StoreSwitcher mobileView={true} />
+        </div>
+    ) : null;
```
sirugh, line 58: Will `StoreSwitcher` ever be falsy? I don't think you need this conditional.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -63,7 +69,7 @@ export const useMagentoRoute = props => {
             return;
         }
 
-        if (shouldFetch(routeData, store)) {
+        if (store && shouldFetch(routeData, store)) {
```
sirugh, line 72: Can move this `store` truthiness check into the `shouldFetch` utility.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -0,0 +1,89 @@
...
+ * @returns {Ref}       talonProps.storeMenuRef - A React ref to the menu that this trigger toggles.
+ * @returns {Ref}       talonProps.storeMenuTriggerRef - A React ref to the trigger element itself.
+ * @returns {Function}  talonProps.handleTriggerClick - A function for handling when the trigger is clicked.
+ * @returns {Function}  talonProps.handleSwitchStore - A function for handling when the menu item is clicked.
+ */
+
+export const useStoreSwitcher = props => {
+    const { getStoreConfig } = props;
+    const history = useHistory();
+    const storage = new BrowserPersistence();
```
sirugh, line 24: This should be initialized outside of the talon.

sirugh (commented):
More feedback :) I have concerns about the mobile views but this is mostly ready to go IMO.

packages/venia-ui/lib/components/Navigation/navigation.js
eug123: It may be when there are only one store view so nothing to switch between, currency switcher may go to `<div class="switchers"/>` too, and they are both independently can be available/unavailable

sirugh: But the export itself will always return the function component. This conditional rendering should probably occur within the switcher component itself.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
eug123: Done, thank you

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
eug123: Done

packages/venia-ui/lib/components/Header/switcherItem.js
eug123: Changed this.

Thank you

packages/venia-ui/lib/components/Navigation/navigation.js
eug123: You are right, I need this to have control over `--header-height` custom property, but it's never worked. I have changed to `useRef` to get switchers height.

packages/venia-ui/lib/components/Navigation/navigation.js
eug123: Resolved this on `header` and `navigation` components level with media queries for `.switchers` container

packages/venia-ui/lib/queries/getAvailableStoresConfigData.graphql
```diff
@@ -0,0 +1,19 @@
+query availableStoresConfigData {
```
sirugh, line 1: Instead of generic `queries` we are moving to colocating queries next to their components that utilize them.

So this would become:

```
import { GET_CONFIG_DATA_QUERY } from './storeSwitcher.gql.js';
```

And you would modify it to use `gql` and export the query.

I'd like to see this new query rewritten in this manner, but the other queries you are just modifying in this change set I leave to your discretion.

sirugh (changes requested):
I've been through this PR a lot with you and I think I've given all the feedback I have to give. I would like to see someone else like @jimbo review it before we merge.

There are also some small changes I'd like before approving, and there is also a comment from brabender [here](https://github.com/magento/pwa-studio/pull/2696#issuecomment-694329608) about some issues he had with this and dave's PR combined. Please work with Dave to resolve that.

packages/venia-ui/lib/queries/getAvailableStoresConfigData.graphql
eug123: Moved query next to storeSwitcher component.
Regarding this [issue](https://github.com/magento/pwa-studio/pull/2696#issuecomment-694329608), I've tried to merge Dave's branch into mine before, there were some conflicts, but at the first look it seems to work fine, I'm starting more thorough testing both branches merged now, and will push it, when Dave's PR will be merged.

packages/venia-concept/.graphqlconfig
```diff
@@ -14,7 +14,8 @@
                     "filesGlob": "../{venia-ui,venia-concept}/{lib,src}/**/*.{js,graphql,gql}",
                     "ignore": [
                         "../venia-ui/lib/**/giftOptions.js",
-                        "../venia-ui/lib/**/wishlistPage.gql.js"
+                        "../venia-ui/lib/**/wishlistPage.gql.js",
+                        "../venia-ui/lib/**/storeSwitcher.gql.js"
```
sirugh, line 18: What is the reason for ignoring the store switcher queries?

packages/venia-ui/lib/queries/getStoreConfigData.graphql
```diff
@@ -2,5 +2,10 @@ query storeConfigData {
     storeConfig {
         id
         copyright
+        code
+        locale
+        base_currency_code
+        default_display_currency_code
+        grid_per_page
```
sirugh, line 7: The category and searchPage components should each have their own queries for `storeConfig` data and it should be pared down to _just_ the data those components need. Infact it looks like they both only use `grid_per_page` so you should be able to create a much smaller query next to each:

```gql
const getPageSize = gql`
  query getPageSize {
    storeConfig {
      grid_per_page
    }
  }
}
```

packages/venia-ui/lib/queries/getStoreConfigData.graphql
```diff
@@ -2,5 +2,10 @@ query storeConfigData {
     storeConfig {
         id
         copyright
+        code
+        locale
+        base_currency_code
+        default_display_currency_code
```
sirugh: You can remove the other properties from this query that aren't used.

packages/venia-ui/lib/queries/getStoreConfigData.graphql
sirugh: And you can remove `grid_per_page` from this query.

sirugh (changes requested):
Can we get an approval from someone else besides me? This is a pretty big change and I want to get more than just my eyes on the whole thing.

packages/venia-concept/.graphqlconfig
eug123: It contains query only available since Magento 2.4.1

packages/venia-ui/lib/queries/getStoreConfigData.graphql
eug123: Moved queries for `storeConfig` to respective components

packages/venia-concept/.graphqlconfig
sirugh: Hmm is there a way we can indicate a comment in this page or do we just have to remember to remove these when 2.4.1 is released?

packages/venia-ui/lib/components/App/localeProvider.js
```diff
@@ -39,6 +53,8 @@ const LocaleProvider = props => {
         }
     };
 
+    if (loading) return fullPageLoadingIndicator;
```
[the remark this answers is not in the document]
sirugh: I know it was causing a warning but I'd like to try and figure out why before we just merge this line. If you change it to 

```js
if (loading && !data) {
   debugger;
  return fullPageLoadingIndicator
}
```

does it pause after you change the locales?

If so, it means that for some reason data is getting deleted even after you refresh the page,

You might also try changing this to `if (!messages)` which should be the same thing.

packages/venia-ui/lib/queries/getStoreConfigData.graphql
eug123: Done

packages/venia-ui/lib/components/App/localeProvider.js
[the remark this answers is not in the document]
eug123: It doesn't step into condition `if (loading && !data)` and stop on debugger when I switch locales. Since we already have `data` cached `storeConfig` for all stores in `storeSwitcher`

It does stop on `if (loading && !messages)` and just `if (!messages)`, but still got warning.

Leaving just `if (loading)` works without warning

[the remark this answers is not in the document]
sirugh: OK. That's fine. No need to debug the deeper issue with the warning right now.

packages/venia-concept/.graphqlconfig
eug123: Add comment to `storeSwitcher.gql.js`

sirugh (dismissed):

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -0,0 +1,101 @@
...
+import { BrowserPersistence } from '@magento/peregrine/lib/util';
+
+const storage = new BrowserPersistence();
+
+const mapAvailableOptions = (rawConfigData, rawAvailableStoresData) => {
+    const { code } = rawConfigData;
+    const availableOptions = {};
+
+    rawAvailableStoresData.forEach(store => {
+        availableOptions[store.code] = {
```
jimbo, line 14: If you're going to reduce an iterable (`rawAvailableStoresData` here) to a keyed object, you should consider whether the consumer will

- **know** or **not know** the keys
- **look up** values or **iterate over** entries

In this case, the consumer (component) will **not know** the keys and will **iterate over** entries to create elements, so it'd be better to reduce to a `Map` than a plain object.

```es6
const getAvailableOptions = (config, stores) => {
  const { code: configCode } = config

  const options = stores.reduce(
    (map, store) => {
      const {
        code,
        default_display_currency_code: currency,
        locale,
        store_name: storeName
      } = store

      const isCurrent = code === configCode
      const option = { currency, isCurrent, locale, storeName }

      return map.set(code, option)
    },
    new Map()
  )
}
```

Also, the option object here should probably use `isCurrent` instead of `is_current`, to match the casing of `storeName`.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -0,0 +1,101 @@
...
+            availableStoresData &&
+            mapAvailableOptions(
+                storeConfigData.storeConfig,
+                availableStoresData.availableStores
+            )
+        );
+    }, [storeConfigData, availableStoresData]);
+
+    const handleSwitchStore = useCallback(
+        // Change store view code and currency to be used in Appollo link request headers
```
jimbo, line 71:
```suggestion
        // Change store view code and currency to be used in Apollo link request headers
```

jimbo (changes requested):
It would be nice if we could just use the mobile store switcher UI, where the switcher is in the main drawer, at all resolutions. I think it's much better than the header toolbar, comes with way fewer complications, and works perfectly in the drawer.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
eug123: Done, thank you

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
eug123: Corrected

packages/peregrine/lib/talons/CategoryTree/useCategoryTree.js
```diff
@@ -41,8 +43,8 @@ export const useCategoryTree = props => {
         }
     }, [data, updateCategories]);
 
-    const rootCategory = categories[categoryId];
-    const { children } = rootCategory || {};
+    const rootCategory = data && data.category;
```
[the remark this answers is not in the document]
sirugh: The catalog data is still necessary even after this change as it controls the rendering near the top of this tree in `useNavigation`. All that @eug123 did here is change where the category tree (this component) gets the data - instead of using the prop data it just uses the query data that it was making anyways. Functionally identical without the round trip of updating the store and then re-rendering when the `catalog` prop makes its way back here.

I do think we should try to disentangle some of the catalog reducer logic from the navigation drawer but I think that scope is beyond this PR.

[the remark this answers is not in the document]
eug123: Other components still use `categories` from Redux. Seems removing `updateCategories()` from here, will require rebuild all of them to use Apollo cache, and eventually deprecate `useCatalogContext`

jimbo (dismissed):

fooman:
One thing which I am not sure is handled yet is that different store views can very likely have different root_category_ids. When switching this should also update.

dpatil-magento:
@eug123 Please check below observations - 
1.Safari Icon and text are too close.
image
2.Even though #STORE_VIEW_CODE=fr is commented, store switcher still displays when backend have multiple store views. Expected?  I was thinking store switcher would only be shown when  STORE_VIEW_CODE is enabled in .env.
3. Respective store categories are not maintained. i.e. Default Store specific categories displayed for French store also.

dpatil-magento:
1 is fixed, 2 as designed, 3 backend data fixed.

QA Approved.

eug123:
@fooman, updating root category is not handled by this PR, which covers basic scenario when all store views are in the one store group. Since root_category is store group level setting, it will be addressed in the upcoming Multi-Store Support epic.

packages/venia-ui/lib/RootComponents/Category/category.js
```diff
@@ -1,177 +1,63 @@
...
+    if (error && pageControl.currentPage === 1 && !loading) {
         if (process.env.NODE_ENV !== 'production') {
             console.error(error);
         }
         return <div>Data Fetch Error</div>;
     }
 
     // Show the loading indicator until data has been fetched.
-    if (totalPagesFromData === null) {
+    if (loading) {
```
sirugh, line 47: This is a good change but I would move this `if` block above the error case block and then remove the `&& !loading`.

packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
```diff
@@ -0,0 +1,216 @@
...
+    const categoryData = categoryLoading && !data ? null : data;
+    const metaDescription =
+        data && data.category && data.category.meta_description
+            ? data.category.meta_description
+            : '';
+
+    const loading =
+        (introspectionCalled && !categoryCalled) ||
+        (categoryLoading && !data) ||
+        introspectionLoading;
```
sirugh, line 203: From the perspective of the UI component `loading` should be true if and only if we have do not have data and there are in flight requests.

You can simplify this:

```
const loading = !data && categoryLoading;
```

Reference [this example](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js#L29-L49) to see how we use data and loading states to determine when to render the full page loading indicator versus the background loading indicator.

sirugh (changes requested):
Hopefully last round of feedback!

packages/venia-ui/lib/policies/index.js
```diff
@@ -127,6 +127,18 @@ const typePolicies = {
...
+        fields: {
+            children: {
+                merge(existing, incoming) {
+                    return incoming;
+                }
+            }
+        }
+    },
+    StoreConfig: {
+        keyFields: ['id']
```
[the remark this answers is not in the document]
eug123: Removed, thank you

packages/venia-ui/lib/components/MagentoRoute/magentoRoute.js
```diff
@@ -7,17 +8,17 @@ import {
...
+import { GET_STORE_CODE } from './magentoRoute.gql';
 
 const MESSAGES = new Map()
     .set(NOT_FOUND, 'That page could not be found. Please try again.')
     .set(INTERNAL_ERROR, 'Something went wrong. Please try again.');
 
 const MagentoRoute = () => {
-    const magentoRouteProps = {};
+    const { formatMessage } = useIntl();
+    const magentoRouteProps = { getStoreCode: GET_STORE_CODE };
```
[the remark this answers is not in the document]
eug123: There are more queries added with this PR, probably it is better to align them with new [pattern](https://github.com/magento/pwa-studio/pull/2712) in one separate PR

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -119,21 +138,16 @@ export const useSearchPage = props => {
                 currentPage: Number(currentPage),
                 filters: newFilters,
                 inputText,
-                pageSize: Number(PAGE_SIZE),
+                pageSize: Number(pageSize),
```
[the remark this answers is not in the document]
eug123: Done

packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
```diff
@@ -0,0 +1,216 @@
...
+        // Use the category id for the current category page regardless of the
+        // applied filters. Follow-up in PWA-404.
+        newFilters['category_id'] = { eq: String(id) };
+
+        runQuery({
+            variables: {
+                currentPage: Number(currentPage),
+                id: Number(id),
+                filters: newFilters,
+                pageSize: Number(pageSize),
```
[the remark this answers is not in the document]
eug123: Done

packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
eug123: It is done similar to [useSearchPage](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/SearchPage/useSearchPage.js#L198), when only categoryLoading is involved, `noProductsFound` component flashes for a moment.

packages/venia-ui/lib/components/App/localeProvider.js
```diff
@@ -20,14 +37,19 @@ const LocaleProvider = props => {
...
-                console.error(`Unable to load translation file. \n${error}`);
-            });
-    }, [fetchLocale, setMessages]);
+        if (language) {
+            const locale = fromReactIntl(language);
+            fetchLocale(locale)
+                .then(data => {
+                    setMessages(data.default);
+                })
+                .catch(error => {
```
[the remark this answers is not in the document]
eug123: Fixed blocking error [here](https://github.com/magento/pwa-studio/pull/2686/files?file-filters%5B%5D=.js#diff-837aa57217caa0b1c16e1ca9e5461009R94)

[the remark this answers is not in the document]
sirugh: Awesome, thanks.

packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
sirugh: Ok, please leave a comment indicating it at least, since it is an outlier.

eug123: Done

sirugh (approved):
APPROVED! Hopefully for the last time :D

dpatil-magento:
QA re-approved.

PR #2690, opened 2020-09-09
[feature]: Replace ProductQuantity on Product Page with QuantityFields stepper

### LINKED ISSUE #2679: [feature]: Replace ProductQuantity on Product Page with QuantityFields stepper
### Description
The quantity selector on PDP and cart don't currently match either in the way they look or in the way they behave.

The cart stepper is the latest design, and the quantity selector on PDP should be updated to match this.

### Acceptance Criteria
- ProductQuantity is deprecated and replaced with quantity stepper component currently on CartPage (you will want to use `QuantityFields`, much like the product edit usage)
- Unit tests are updated to reflect expected view changes
- (Optional) The quantity stepper is probably general use enough that it can be moved to the components directory so it becomes API. Discuss with @tjwiebell if interested in this re-factor. 

### Internal JIRA Links
* [[PWA-591](https://jira.corp.magento.com/browse/PWA-591)] Replace quantity selector on PDP to match cart stepper

- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [x] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

huykon:
## Description

The quantity selector on PDP and cart don't currently match either in the way they look or in the way they behave.

The cart stepper is the latest design, and the quantity selector on PDP should be updated to match this.

## Related Issue

Closes #2679.

### Verification Stakeholders

@tjwiebell 
### Verification Steps

1. Go to the PDP.
2. Verify the quantity field is QuantityFields.

## Checklist


schensley:
@huykon while this work does make the "stepper" quantity control more consistent with the control as it exists in cart, that its size (width) seems to only be restricted by the with of the area it exists in can present a bit of a problem in terms of recognition of the control or user expectations (see screen shot). 
At some sizes it appears that that quantity control may expand to such a degree that 
1. it's hard to recognize it as a means for setting the quantity (recognition) and 
2. separates the "increase" and "decrease" buttons to a degree that can make for a less efficient UI. 
I would recommend placing some restrictions to the width of this such that it will appear to accommodate a typical and reasonable quantity (if a shopper is likely to purchase quantities in the hundreds then it should look like a 3-digit number would fit). The "right" size might be a matter of opinion or aesthetics, but generally speaking there is a "shape" that most users recognize.
Also, just a question about the values that appear in the drop down when you click into the quantity field, I am assuming these are meant to be configurable, but some thought might be given to the height of the drop down container that it would fit within the viewport should it provide an long (infinite?) list of numbers.
Thanks.
Screen Shot 2020-09-21 at 10 53 50 AM

huykon:
@schensley Thank for your comment so what should I do now:
- Reduce the width of quantity box
- limit the max number to quantity input
Please give me your feedback!

packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.css
```diff
@@ -166,3 +166,9 @@
 .relatedTitle {
     composes: sectionTitle;
 }
+
+.quantityRoot {
+    composes: root from '../CartPage/ProductListing/quantity.css';
+    grid-template-columns: auto 1fr auto;
+    padding: 0 1rem;
```
tjwiebell, line 173: @huykon - I would suggest this change given @schensley's feedback on sizing. This is no longer a dropdown, so I don't think anything else needs changed other than the width.
```suggestion
    grid-template-columns: auto 4rem auto;
    justify-content: start;
```

huykon: I have just updated code as your recommend.

schensley:
HI @huykon sorry I didn't get back to you, but it looks like you've done an excellent job.  Thanks. UX approved.

huykon:
@schensley thank you

tjwiebell (dismissed):
Change looks good 👍  `ProductQuantity` can safely be deprecated now, but I'll create a followup to make `QuantityStepper` @api and deprecate that old guy. Nice work!

dpatil-magento:
@huykon @tjwiebell Let me know if we need to create follow up tickets or these needs to be handled here - 
1. When user enter qty greater than available, no error is shown to user even though graphql returns it 


2. Might just be a QA flow :) if user enters 0 in qty field then good to enable + stepper and disable Add to Cart button?

huykon:
@dpatil-magento I will wait the confirmation from @tjwiebell to decide whether we should update or not. Thanks for your reporting

tjwiebell:
> @dpatil-magento I will wait the confirmation from @tjwiebell to decide whether we should update or not. Thanks for your reporting

@huykon 

1. Great catch @dpatil-magento - these errors are not getting passed to the QuantityStepper, which doesn't appear to have error handling even if we were. This should be resolved in this scope.

@huykon - Let's add an optional `message` prop to `QuantityFields` and dependently render a message element if truthy. It should auto flow to the next row given the grid, so you should just need `grid-column-end: span 3;` (and maybe `justify-self: start` so it's left aligned). Use previous rendering as a guide.

Screen Shot 2020-09-30 at 8 34 32 AM

2. This is simple and should be fixed, just pass `min={1}` prop to `QuantityFields`

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.js
```diff
@@ -30,40 +31,43 @@ export const QuantityFields = props => {
...
+                    aria-label={'Increase Quantity'}
+                    className={classes.button_increment}
+                    disabled={isIncrementDisabled}
+                    onClick={handleIncrement}
+                    type="button"
+                >
+                    <Icon classes={iconClasses} src={PlusIcon} size={20} />
+                </button>
+            </div>
+            {message && <Message>{message}</Message>}
```
tjwiebell, line 69: We have a preference to avoid inline expressions like this; consider extracting this to a variable.

huykon:
I have just updated. Please check again @tjwiebell @dpatil-magento

tjwiebell (changes requested):
Tiny change to make this more consistent with the code base then this looks good to go 🚀

tjwiebell (dismissed):

dpatil-magento:
Thanks @huykon @tjwiebell , there are still some breaking flows - 

2 (continuing) - Make qty blank and click outside (I think at this step only good to make qty display back to 1) and enter `01` which turns to `11` and now click add to cart button, user see the error.
<a href="https://gyazo.com/08f38d8034185c6d1ea8f7f9b08e68c7">Image from Gyazo</a>
 
3 -  Add an item to cart and then increase the qty and directly (keeping cursor in qty field only) hit Add to cart again - Previous qty being added agian and not newly entered one 
<a href="https://gyazo.com/69f0ef160523a7433ebbc137c8d4a564">Image from Gyazo</a>

dpatil-magento:
QA Approved.

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.js
```diff
@@ -29,41 +30,46 @@ export const QuantityFields = props => {
         maskInput
     } = talonProps;
 
+    const errorMessage = message ? <Message>{message}</Message> : null;
```
sirugh, line 33: This looks good - my only concern is that the message is a bit close to the field and not red.

sirugh (approved):
One comment but I won't hold the PR up on it as I don't have context around discussions with UX on this change.

packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.js
```diff
@@ -12,7 +12,7 @@ import Button from '../Button';
 import Carousel from '../ProductImageCarousel';
 import FormError from '../FormError';
 import { fullPageLoadingIndicator } from '../LoadingIndicator';
-import Quantity from '../ProductQuantity';
```
sirugh, line 15: Oh also, can we throw a `@deprecated` label on the old `ProductQuantity` component? After this PR it will only be used in the legacy mini cart.

tjwiebell: There were two things I wanted to do as a followup, prefer to just merge this:

https://github.com/magento/pwa-studio/pull/2690#pullrequestreview-497817541

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.js
tjwiebell: We stuck with the same UX from the dropdown treatment, which you recently implemented 😉  it looks pretty identical to me.

https://github.com/magento/pwa-studio/pull/2690#issuecomment-701398370

packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.js
tjwiebell: Created [PWA-960](https://jira.corp.magento.com/browse/PWA-960), thanks for the reminder; would have missed actually creating that in the backlog.

huykon: @tjwiebell please mention to me when you create that issue. I hope I can resolve that :D

sirugh (approved):

tjwiebell (approved):
Do I have write access to approve?

PR #2696, opened 2020-09-11
[PWA-812] Enable PWA Studio packages and extensions to provide translations

davemacaulay:
## Description

Enable PWA Studio packages and extensions to provide translations

- Plugin is handed an array of directories which declare an `i18n` `specialFlag` along with the `context` which is `Venia-concept`.
- We scan the `i18n` directory of every provided module and locate any translation files matching the pattern `/([a-z]{2}_[A-Z]{2})\.json$/` (basically this is anything en_US.json, fr_FR.json).
- We generate an object with locales as keys, then an array of all matched files, ordered by the `BuildBus` with context being the final entry, and thus the final override position.
- We iterate through each locale, overwrite merge each file over the last
- We produce a virtual module into the Webpack file system as to not produce files on the actual filesystem
- We codegen a dynamic import of the virtual module and expose the `__fetchLocaleData__` function to the window to be consumed within the `localeProvider`

## Related Issue
https://jira.corp.magento.com/browse/PWA-812

### Specification
- Language packs / translations are dynamically pulled from NPM packages / extensions and from the scaffolded `i18n` directory
- The scaffolded app has the highest priority and any string present in the directory will override others
- Translation components should be usable in other packages throughout the app
- Translations should be stored within an i18n directory in the root of the package 

### Verification Steps
**Build `venia-concept`:**
1. Create `i18n` directory with `en_US.json` file in another npm package (`venia-concept`) overriding one of the existing translations (like "Search")
2. Build application using `yarn run build`
3. Run application `yarn stage:venia` 
4. View application in browser and ensure that "Search" string is overridden by the new translated string

**Watch `venia-concept`:**
1. Create `i18n` directory with `en_US.json` file in another npm package (`venia-concept`) overriding one of the existing translations (like "Search")
2. Run `yarn watch:venia`
3. Open the application in the browser
4. Verify the new translation is displayed in place
5. Modify translated string and verify app is hot reloaded and new string is present

**Watch `peregrine`:**
1. Create `i18n` directory with `en_US.json` file in another npm package (`peregrine `) overriding one of the existing translations (like "Search")
2. Modify the `peregrine-intercept.js` file and include the `i18n: true` `specialFeatures` flag
3. Run `yarn watch:venia`
4. Open the application in the browser
5. Verify the new translation is displayed in place
6. Modify translated string and verify app is hot reloaded and new string is present

**Custom Module:**
1. Create a custom npm module, with the required `i18n` folder, specialFeatures flag etc
2. Add an override translation
3. Run `yarn watch:venia` and verify translation is show.
4. Run `yarn run build` then `yarn stage:venia` and verify translation is shown

**Scaffolded Project:**
1. Scaffold project from this branch
2. Add translation overrides in root of scaffolded project (`i18n/en_US.json`)
3. Build and run app
4. Verify new translations are present

## Checklist


chris-brabender:
Been testing this and 814 out the last couple of days and its looking good so far!

Small issue when the two are combined (cannot replicate on 814 on its own) so may be more in this side of the setup.

When selecting a new store view from the header component, the page is reloaded (expected behaviour).

The graphql data is fetch properly on the new page load it seems but the translated components are not updated with the new locale. It takes a full load of that page (triggering some update internally) and then a full refresh of the page to see the updated component text.

sirugh:
> The graphql data is fetch properly on the new page load it seems but the translated components are not updated with the new locale. It takes a full load of that page (triggering some update internally) and then a full refresh of the page to see the updated component text.

This is due to the fetch policies for the queries. Default is `cache-first` which means if cache data exists it will never true-up with backend. This is not the optimal solution for us since we want to make sure to a) render ASAP and b) render fresh data. We can accomplish with `fetchPolicy: 'cache-and-network'` which does just that :)

@eug123 has been working to update the fetch policies in his PR so that on refresh the data is refetched with the new store header, and thus the cache data is updated.

chris-brabender:
I think I worded it wrong and may have confused you, the graphql isnt so much an issue here from what I see.

It's the rest of the components (non-graphql data) that aren't correctly translated until a second full refresh.

sirugh:
> It's the rest of the components (non-graphql data) that aren't correctly translated until a second full refresh.

Oh, hm. I guess that would up to whenever we switch to the correct locale i18n json file? I haven't tried any of this locally so I'm just speculating.

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
```diff
@@ -0,0 +1,236 @@
...
+                                    localePath
+                                );
+                            });
+                        }
+                    });
+                }
+            });
+        });
+
+        return `;window.fetchLocaleData = (${importerFactory})()`;
```
sirugh, line 135: In my experience, using `window.globalFunction` can lead to issues if any js on the page also happens to redefine this property. Consider namespacing, for example, `window[${storeName}].fetchLocaleData`.

sirugh (commented):
I don't have too many comments. It reads well to me, so long as you understand what is happening. Could do with a more descriptive block comment in the plugin explaining what the plugin does.

Good job :)

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
```diff
@@ -0,0 +1,236 @@
...
+                    locales = this.mergeLocales(locales, packageTranslations);
+                } else {
+                    throw new Error('Path is not directory.');
+                }
+            } catch (e) {
+                debug(e);
+
+                if (dir !== context) {
+                    throw new Error(
+                        `${dir} module has i18n special flag, but i18n directory does not exist at ${localeDir}.`
```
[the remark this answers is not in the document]
davemacaulay: The feature is opt in, thus if they set the flag the directory should exist. If they don't intend to provide translations within their module they shouldn't set the flag.

This enforces correct usage of the API as the build will fail if they do it wrong.

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
```diff
@@ -0,0 +1,236 @@
...
+
+        const combinedLocale = {};
+
+        for (const locale of Object.keys(locales)) {
+            debug(`Combining locale for ${locale}`);
+            const files = locales[locale];
+            let combined = {};
+            for (const file of files) {
+                const data = inputFileSystem.readFileSync(file, 'utf8');
+                combined = merge.recursive(combined, JSON.parse(data));
```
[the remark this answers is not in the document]
davemacaulay: Not particularly easy, we provide the debug output of the loaded module order which can help people debug issues with overrides. 

If we see consistent bug reports / issues for people not understanding the merging strategy we can investigate it at a future point.

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
davemacaulay: I think this just hides the problem, someone can still overwrite `window[${storeName}]` all the same, if someone messes with `fetchLocaleData` they're gonna see the whole app error as it's consumed early on in the `localeProvider`.

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
```diff
@@ -0,0 +1,236 @@
...
+        return new Promise(resolve => {
+            const translations = {};
+            const done = () => resolve(translations);
+            walk(dir, { fs: inputFileSystem })
+                .on('readable', function() {
+                    let item;
+                    while ((item = this.read())) {
+                        if (
+                            item.stats.isFile() &&
+                            localeFileNameRegex.test(item.path)
```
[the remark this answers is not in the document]
davemacaulay: I've added a `debug` statement for invalid files in these directories, just to help those if they've named something slightly wrong.

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
sirugh: I guess it's less a problem in the scope of Magento-land where we own the page.

revanth0212: I do agree with both of you. How about using `window.__fetchLocaleData__` instead? It is an accepted norm that anything that has a `__` prefix and/or suffix is internal and not messed around with.

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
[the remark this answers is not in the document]
revanth0212: Again I agree with both of you, but what do you think about a stop-gap solution for now where before merging let's take the length of keys of both the objects and compare that with the length after merging, if they are not same, leave a warning saying the following 2 packages have conflicting translations. We don't need to enumerate those as of now and like Dave mentioned if the community feels like they may need that info, we can consider upgrading the function in the future releases.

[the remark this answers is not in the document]
davemacaulay: @revanth0212 that's a pretty awesome suggestion, I'll certainly do that!

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
davemacaulay: @revanth0212 good suggestion, I've implemented the change. Maybe we should consider updating the root components version of this at some point, but that'd be a breaking change so maybe not the best idea.

zetlen (commented):
(Messed up this review)

zetlen (approved):
LGTM :+1: Cut it, print it. 🎬

sirugh (dismissed):
Nice suggestions @revanth0212 :)

revanth0212 (dismissed):

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
```diff
@@ -0,0 +1,262 @@
...
+                // If the directory declares i18n support, but doesn't contain the directory error the build
+                if (dir !== context) {
+                    throw new Error(
+                        `${dir} module has i18n special flag, but i18n directory does not exist at ${localeDir}.`
+                    );
+                }
+            }
+        }
+
+        if (!locales) {
```
thierry2015, line 70: This condition will never be met since `locales` is an object. Recommending `Object.keys(locales).length === 0`

packages/venia-ui/lib/components/App/localeProvider.js
```diff
@@ -8,15 +8,26 @@ const locale = fromReactIntl(language);
...
-        import(`../../i18n/${locale}.json`)
+        fetchLocale(locale)
             .then(data => {
                 setMessages(data.default);
             })
             .catch(error => {
                 console.error(`Unable to load translation file. \n${error}`);
             });
-    }, [setMessages]);
+    }, [fetchLocale, setMessages]);
```
thierry2015, line 30: Do we want this effect to be called when `fetchLocale` changes? It seems like a one-shot operation to me, and should not have any dependency.

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
```diff
@@ -0,0 +1,262 @@
...
+            const localeDir = path.join(dir, i18nDir);
+            try {
+                const stats = inputFileSystem.statSync(localeDir);
+
+                if (stats.isDirectory()) {
+                    const packageTranslations = await this.findTranslationFiles(
+                        inputFileSystem,
+                        localeDir
+                    );
+                    locales = this.mergeLocales(locales, packageTranslations);
```
sirugh, line 54: Does `mergeLocales` merge into the passed `current` ie `locales` object? Might not need the return here or in the function.

packages/venia-ui/lib/components/App/localeProvider.js
davemacaulay: You're right `fetchLocale` currently won't change, however if we get hot reload support working for translations this could be the trigger which updates the translations.

Also our linter cares about this dependency so rather than forcing the linter not to care I vote we just leave it in as it won't have any negative side effects I believe.

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
davemacaulay: Oops, writing PHP for so long makes you forget JS is pass by reference land. Good catch thank you :)

packages/pwa-buildpack/lib/WebpackTools/plugins/LocalizationPlugin.js
```diff
@@ -0,0 +1,262 @@
...
+                }
+            }
+        }
+
+        if (Object.keys(locales).length === 0) {
+            debug(
+                'No locales found while traversing all modules with i18n flag.'
+            );
+
+            return '';
```
sirugh, line 75: Can we just `return;`? Empty string is return is new to me :D

sirugh (approved):
Approved minus that one return statement.

sirugh (approved):

dpatil-magento:
QA Approved.

PR #2697, opened 2020-09-11
PDPs show intended Product only

supernova-at:
## Description

Venia does not support Grouped, Bundled, or Virtual products yet, but the Venia sample data does include them.
This PR adds a bit of filtering logic into the `useProduct` root component talon to only return the `product` we requested (by `url_key`).

## Related Issue

Closes PWA-895.

### Verification Steps

1. Set your `MAGENTO_BACKEND_URL` to a `2.4.1+` backend (this PR deployment may be that already)
1. Go to the `/gold-veritas-cuff-set.html` page
1. See that you are presented with the simple product, not the bundled product set

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
+++ b/packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
@@ -1,11 +1,6 @@
 import { useQuery } from '@apollo/client';
 import { useMemo } from 'react';
 
-const VENIA_SUPPORTED_PRODUCT_TYPES = ['SimpleProduct', 'ConfigurableProduct'];
-const isSupportedProductType = product => {
-    return VENIA_SUPPORTED_PRODUCT_TYPES.includes(product.__typename);
-};
-
 /**
  * A [React Hook]{@link https://reactjs.org/docs/hooks-intro.html} that
  * controls the logic for the Product Root Component.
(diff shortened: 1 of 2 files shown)


packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -29,14 +34,23 @@ export const useProduct = props => {
     });
 
     const product = useMemo(() => {
+        if (!data) {
+            // The product isn't in the cache and we don't have a response from GraphQL yet.
+            return null;
+        }
+
+        const supportedProducts = data.products.items.filter(
```
tjwiebell, line 42: To future proof this fix (in a distant future where we support grouped products), we can instead just search for the `url_key` we intended to query for. `const product = data.product.items.find(item => item.url_key === urlKey)`

packages/peregrine/lib/talons/RootComponents/Product/useProduct.js
```diff
@@ -29,14 +34,23 @@ export const useProduct = props => {
...
-            return mapProduct(data.products.items[0]);
+        // Additionally, we may have an empty array after filtering out unsupported product types.
+        if (!supportedProducts.length) {
+            return null;
         }
 
-        // The product isn't in the cache and we don't have a response from GraphQL yet.
-        return null;
+        // Pick the first supported product.
+        return mapProduct(supportedProducts[0]);
```
tjwiebell, line 53: I can't believe `mapProduct` is still around, it was a compatibility fix for `2.3.1` which is long out of support. I'll create a tech debt task to remove it, which should clean this talon up a little bit.

tjwiebell (changes requested):
Minor suggestion to future proof this fix a little more. Test coverage to awesome 🙌  thanks for adding that.

tjwiebell (approved):

PR #2703, opened 2020-09-14
My Account - Order History Table View (Expanded) 

revanth0212:
## Description

In this PR I have added the expanded view in the order history table.

The buttons `Buy Again`, `Return This`, and `Print Receipt` are dummy for now since they are not part of the current Epics's scope. The same goes for the tracking number link.

GQL coverage for this feature is missing hence I have used mock data based on the schema mentioned [here](https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/customer/customer-orders.md).

## Related Issue
Closes [PWA-627](https://jira.corp.magento.com/browse/PWA-627)

### Verification Stakeholders
@schensley 

### Verification Steps
1. Go to the deployed app's order history page.
2. Click on the order row to expand it.

Desktop
image

Mobile
image

## Checklist


packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/item.js
```diff
@@ -0,0 +1,70 @@
...
+            />
+            <span
+                className={classes.quantity}
+            >{`Qty : ${quantity_ordered}`}</span>
+            <span className={classes.price}>{product_sale_price}</span>
+            <Button onClick={() => {}} className={classes.buyAgainButton}>
+                {'Buy Again'}
+            </Button>
+            <Button onClick={() => {}} className={classes.returnThisButton}>
+                {'Return This'}
```
sirugh, line 64: Per conversation in dev time, lets try returning `strings` from the talons for all of these components.

Something like:

```
strings: {
  returnThisButtonText: formatMessage({ id: orderHistoryPage.orderDetails.item.returnThisButtonText })
  ...
}
```

This does raise a point of how to properly identify the strings. If we just use `component.thing` we will most likely have overlap.

packages/venia-ui/lib/components/OrderHistoryPage/orderRow.js
```diff
@@ -46,16 +52,26 @@ const OrderRow = props => {
...
+
     const contentToggleIconSrc = isOpen ? ChevronUp : ChevronDown;
+
     const contentToggleIcon = <Icon src={contentToggleIconSrc} size={24} />;
+
     const collapsedImageGalleryElement = isOpen ? null : (
-        <CollapsedImageGallery items={items} />
+        <CollapsedImageGallery
+            totalItemsCount={items.length}
+            displayCount={DISPLAY_COUNT}
```
sirugh, line 72: Why does this constant need to be a prop?

packages/peregrine/lib/talons/OrderHistoryPage/useOrderRow.js
```diff
@@ -1,14 +1,72 @@
...
+ * @param {OrderRowQueries} props.queries GraphQL queries for the Order Row Component
+ *
+ * @returns {OrderRowTalonProps}
+ */
+export const useOrderRow = props => {
+    const { items, queries } = props;
+    const { getProductThumbnailsQuery } = queries;
+
+    const skus = useMemo(() => {
+        return items.map(item => item.product_sku);
```
sirugh, line 18: Because apollo client caches by `toString` of query name + variables, and because these `skus` are used in the query below, you should sort them.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/item.js
```diff
@@ -0,0 +1,70 @@
...
+    const {
+        product_name,
+        product_sale_price,
+        quantity_ordered,
+        selected_options,
+        thumbnail,
+        url_key,
+        url_suffix
+    } = props;
+    const { url } = thumbnail;
```
sirugh, line 22: We saw an error from this LOC when removing `apollo-cache-persist` and refreshing the page when on the order history page.

Worth investigating.

packages/venia-ui/lib/components/OrderHistoryPage/collapsedImageGallery.js
```diff
@@ -1,33 +1,20 @@
...
-        imageCount: DISPLAY_COUNT,
-        items,
-        ...imageGalleryOperations
-    });
-    const { imageData } = talonProps;
+    const { displayCount, items, totalItemsCount } = props;
 
     const classes = mergeClasses(defaultClasses, props.classes);
-    const remainderCount = items.length - DISPLAY_COUNT;
+    const remainderCount = totalItemsCount - displayCount;
```
sirugh, line 13: `totalItemsCount` can just be derived from the `items.length`. Is there a reason it is now a prop? Similar question for the passing of `displayCount`.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
```diff
@@ -14,25 +14,86 @@ export const GET_CUSTOMER_ORDERS = gql`
...
+                    payment_methods {
+                        name
+                        type
+                        additional_data {
+                            # additional data is not a constant documented in the schema architecture doc because
+                            # it changes based on the type of payment method. This schema I have here is temp based
+                            # on what braintree returns in checkout step. Once orders GQL coverage is ready, we have to revist this.
+                            card_type
+                            last_four
+                        }
```
sirugh, line 53: Are you saying that this is not actually going to exist? Or are you saying that it _could_ exist if you have braintree enabled? I'm a little confused about the comment.

sirugh (commented):
I know this is still in UX review but I left a preliminary review from discussion earlier.

packages/peregrine/lib/talons/OrderHistoryPage/useOrderRow.js
revanth0212: So, even if the order of SKU's change, apollo cache hit will be a success?

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/item.js
revanth0212: What if a component does not have a talon but has constant strings? For instance, [shippingInformation.js](https://github.com/magento/pwa-studio/blob/revanth/order-history-expanded/packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/shippingInformation.js#L23) does not have a talon but has a heading called `Shipping Information`.

packages/peregrine/lib/talons/OrderHistoryPage/useOrderRow.js
sirugh: No, if the order of SKUs change, the cache will not "hit". The results will reference the same cached object after the network call but a network call will be made.

```
"query([sku1, sku3, sku2])": CACHED_DATA
"query([sku1, sku2, sku3])": CACHED_DATA
```

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/item.js
sirugh: There was discussion in chat yesterday about not doing this - and your point is valid. I think we need to rediscuss in dev time, if at all. I lean towards just doing formatting in the view.

revanth0212: Me too, let discuss about this today 👍 .

packages/peregrine/lib/talons/OrderHistoryPage/useOrderRow.js
revanth0212: Correct me if I am wrong. Because of sorting, the order of SKUs in the `skus` array passed to the query will stay the same even if their order changes in the `items` array. This will lead to cache hits every time after the first time.

sirugh: > Because of sorting, the order of SKUs in the skus array passed to the query will stay the same even if their order changes in the items array.

Yes, this is what I'm saying. You should sort the `skus` because array order is not guaranteed. The `items` array could be a different order.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.gql.js
revanth0212: Sorry, this is an old comment I forgot to remove. It is not needed anymore.

schensley:
UX Approved.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/item.js
```diff
@@ -0,0 +1,99 @@
...
+    const {
+        product_name,
+        product_sale_price,
+        quantity_ordered,
+        selected_options,
+        thumbnail,
+        url_key,
+        url_suffix
+    } = props;
+    const thumbnailUrl = thumbnail ? thumbnail.url : '';
```
sirugh, line 23: This certainly solves the issue when `thumbnail` is `undefined` but did you figure out why it was undefined and how we ended up here at all? I'd rather fix that than the side effect :)

sirugh (commented):
Only one thought, really, related to the error we saw earlier. Would rather try to solve that now than to kick it down the road.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/item.js
revanth0212: I presume that is because parts of the data is mock data which is resolved immediately in one query and the thumbnail is part of another query which is taking long.

If you look at how the `items` component renders the `item` component, you would see that I am spreading the `item` object and the `mappedImageData` object. `item` object is readily available while the `mappedImageData` takes a while to be resolved. `thumbnail` is part the `mappedImageData` hence it is undefined at first but on subsequent renders it is available.

sirugh: So what happens the first time we render before the image data query resolves? Does the UI attempt to render an invalid image using `url=""` or something? Can you check that out? Do we skip render of the thumbnail until we have it?

revanth0212: If `url=''` no image will be rendered also no error will be thrown.

As I was writing this comment, I realized it is semantically not correct to render an image tag with an empty source so I will make the change to not render the component till the src is available. I can use the loading state to show a loading indicator till the src is available.

revanth0212: I have pushed d03d1a2 which will not render the order details if the data is loading. Let me know what you think.

sirugh: Ok - now you don't need this line, and can just reference `thumbnail` directly. I'd also make the prop type required.

```suggestion
```

revanth0212: Sorry forgot to remove that line of code.

sirugh (approved):

revanth0212:
image

I don't see any failures while running the tests locally.

dpatil-magento:
@revanth0212 Please check following when compared to Mocks in ticket - 
1. `Order Details` label is missing.
2. `Buy Again` is button but implemented as link.
3. Credit card number is `**** 1234` but implemented as just `1234`

dpatil-magento:
> @revanth0212 Please check following when compared to Mocks in ticket -
> 
> 1. `Order Details` label is missing.
> 2. `Buy Again` is button but implemented as link.
> 3. Credit card number is `**** 1234` but implemented as just `1234`

1 and 2 are UX recent changes and 3 will work with real data.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/shippingMethod.js
```diff
@@ -0,0 +1,66 @@
...
+            <div className={classes.heading}>
+                <FormattedMessage
+                    id="orderDetails.shippingMethodLabel"
+                    defaultMessage="Shipping Method"
+                />
+            </div>
+            <div className={classes.method}>{shippingMethod}</div>
+            <div className={classes.tracking}>
+                <FormattedMessage
+                    id="{carrier} Tracking: {number}"
```
sirugh, line 32: Use the dot notation id format please! 
```suggestion
                    id="shippingMethod.trackingInformation"
```

packages/venia-ui/i18n/en_US.json
```diff
@@ -109,9 +109,16 @@
...
     "navigationTrigger.ariaLabel": "Toggle navigation panel",
     "Order Status": "Order Status",
+    "orderDetails.billingInformationLabel": "Billing Information",
+    "orderDetails.buyAgain": "Buy Again",
+    "orderDetails.paymentMethodLabel": "Payment Method",
+    "orderDetails.printLabel": "Print Receipt",
+    "orderDetails.returnThis": "Return This",
+    "orderDetails.shippingInformationLabel": "Shipping Information",
+    "orderDetails.shippingMethodLabel": "Shipping Method",
+    "orderProgressBar.deliveredText": "Delivered",
```
sirugh, line 119: Weird merge? This was previously in alphabetical order but got moved above `orderHistoryPage` strings.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/item.js
```diff
@@ -0,0 +1,110 @@
...
+            </Link>
+            <ProductOptions
+                options={mappedOptions}
+                classes={{
+                    options: classes.options
+                }}
+            />
+            <span className={classes.quantity}>
+                <FormattedMessage
+                    id="Qty : {quantity}"
```
sirugh, line 59: Same here, `id={"orderDetails.quantity"}`

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/orderDetails.js
```diff
@@ -0,0 +1,180 @@
...
+            <div className={classes.paymentMethodContainer}>
+                <PaymentMethod data={payment_methods} />
+            </div>
+            <div className={classes.itemsContainer}>
+                <Items data={{ imagesData, items }} />
+            </div>
+            <div className={classes.orderTotalContainer}>
+                <OrderTotal data={total} />
+            </div>
+            <Button className={classes.printButton} onClick={() => {}}>
```
sirugh, line 57: Is there a ticket to implement printing? I'd probably throw a comment in the empty callback indicating. If not...then make one :D

sirugh (changes requested):
Almost there!

packages/venia-ui/i18n/en_US.json
revanth0212: Yeah, addressed the order.

dpatil-magento:
QA Approved.

sirugh (approved):
Bueno!

PR #2705, opened 2020-09-14
fix: storybook i18n

sirugh:
## Description

We added i18n stuff to the [venia webpack config](https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/webpack.config.js#L57-L60) but did not add it to the venia-ui storybook config.

We also did not wrap storybook entry in the venia-ui provider, which is where we connected the `LocaleProvider` among others.

## Related Issue

Closes [PWA-907](https://jira.corp.magento.com/browse/PWA-907).

### Verification Steps

1. `yarn storybook:venia`
2. Verify each story loads and renders. You may see errors in console related to GQL but that is because the dummy backend doesn't return data. 
3. Verify that scaffolded storybook works.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/Header/__stories__/header.js
+++ b/packages/venia-ui/lib/components/Header/__stories__/header.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { storiesOf } from '@storybook/react';
 
 import { useAppContext } from '@magento/peregrine/lib/context/app';
@@ -17,7 +17,9 @@ stories.add('Search Bar Open', () => {
         const { toggleSearch } = api;
 
         // Open the search
-        toggleSearch();
+        useEffect(() => {
+            toggleSearch();
+        }, [toggleSearch]);
 
         return <Header classes={defaultClasses} />;
     };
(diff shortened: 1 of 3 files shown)


packages/venia-ui/lib/components/Header/__stories__/header.js
```diff
@@ -17,7 +17,9 @@ stories.add('Search Bar Open', () => {
         const { toggleSearch } = api;
 
         // Open the search
-        toggleSearch();
+        useEffect(() => {
+            toggleSearch();
+        }, [toggleSearch]);
```
sirugh, line 22: Without the effect we had an infinite loop in this story. Not sure if changing out the provider is what caused this to start happening, but wrapping in an effect fixes it.

supernova-at (approved):

packages/venia-ui/lib/components/Header/__stories__/header.js
jimbo: Good catch. [Side effects definitely belong in `useEffect`](https://reactjs.org/docs/hooks-reference.html#useeffect).

> Mutations, subscriptions, timers, logging, and other side effects are not allowed inside the main body of a function component (referred to as React’s render phase). Doing so will lead to confusing bugs and inconsistencies in the UI.

We've seen with `console.log()` that it often works, but only coincidentally.

jimbo (approved):

PR #2706, opened 2020-09-14
Remove venia-styleguide

jimbo:
## Description
Remove the `venia-styleguide` package, since it interferes with file navigation and confuses people.

### Verification Stakeholders

- @dpatil-magento 

### Verification Steps
1. Basic regression due to lockfile changes.


brendanfalkowski:
@jimbo Any more explanation on this? Is team axing the design docs implementation again?

jimbo:
> Any more explanation on this? Is team axing the design docs implementation again?

We're just going to use Storybook.

Originally, I had intended to use `venia-styleguide` to get out ahead of the redesign we were bringing to `venia-ui`, but I personally never got the time to actually build out the styleguide. We also got the chance to apply the redesign a lot earlier than anticipated, so the app itself (and thus Storybook) is now more accurate than the styleguide.

sirugh (approved):

supernova-at (approved):

dpatil-magento (approved):

PR #2712, opened 2020-09-17
[PWA-851] Migrate gql related files and folders to peregrine

tjwiebell:
## Description

Goal:

Simplify DX when working with venia-ui/peregrine by reducing the amount of prop imports/prop threading for UI components.

Steps:

Create a new gql/ folder in peregrine.
Move queries/, fragments/, resolvers/, typeDefs/ and policies/ into the folder from venia-ui
Move all *.gql.js files to peregrine next to corresponding talons.
Update talons to still accept operations as arguments but to default to the built in queries.
 

Task: 

Create github issue documenting plan/risks/benefits for community to review and provide feedback.

## Notes
These two examples differ slightly. The first one maintains the mutations/queries grouping, where the latter instead starts using an `operations` prop and relies on operation naming to signify query/mutation. Given we don't enforce returning raw data, I'm not sure the override use case makes a lot of sense here, but would be possible with this setup.

* Example of moving out of queries directory: a8e4337
* Example of moving out of components directory: a9942a2

## Related Issue

* [[PWA-851](https://jira.corp.magento.com/browse/PWA-851)] Migrate gql related files and folders to peregrine

### Verification Steps

This PR touched on a large set of components, so this almost needs a regression level amount of testing. I personally feel since the unit test suite passes without a single snap needing updated should give confidence that all is working as expected, but a smoke test of critical path scenarios makes sense here.

## Checklist


packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -35,14 +36,15 @@ const UNAUTHED_ONLY = ['CREATE_ACCOUNT', 'FORGOT_PASSWORD', 'SIGN_IN'];
 export const useAuthModal = props => {
     const {
         closeDrawer,
+        mutations = DEFAULT_MUTATIONS,
```
sirugh, line 39: I like namespacing the queries and mutations. IMO it could be either:

```js
const { mutations, queries } = props;
const { operations } = props; // where operations is { mutations, queries }
```

Both would allow users to customize specific named operations through the passed props by importing the default and spreading over a new object.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
```diff
@@ -2,9 +2,10 @@ import { useCallback, useEffect, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import { useApolloClient, useMutation } from '@apollo/client';
 
-import { useUserContext } from '../../context/user';
 import { clearCartDataFromCache } from '../../Apollo/clearCartDataFromCache';
 import { clearCustomerDataFromCache } from '../../Apollo/clearCustomerDataFromCache';
+import { useUserContext } from '../../context/user';
+import SIGN_OUT_MUTATION from '../../gql/operations/signOut.graphql';
```
sirugh, line 8: I want to get away from generic mutations/queries if possible but I also know we have some, like sign out, that are global.

I think the goal of this was to move _most_ of them next to the talons where they are used.

packages/peregrine/lib/talons/CommunicationsPage/communicationsPage.gql.js
```diff
@@ -21,10 +21,6 @@ export const GET_CUSTOMER_SUBSCRIPTION = gql`
...
 
 export default {
-    mutations: {
-        setNewsletterSubscriptionMutation: SET_NEWSLETTER_SUBSCRIPTION
-    },
-    queries: {
-        getCustomerSubscriptionQuery: GET_CUSTOMER_SUBSCRIPTION
-    }
+    getCustomerSubscriptionQuery: GET_CUSTOMER_SUBSCRIPTION,
+    setNewsletterSubscriptionMutation: SET_NEWSLETTER_SUBSCRIPTION
```
sirugh, line 25: I kind of like this better - we're already naming things `fooQuery` and `barMutation` anyways. No need to put another layer around them to indicate the type.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
tjwiebell: Any operations I come across that have a single use I'll definitely do this. If we can come to an agreement about what to do with shared queries, I would love to get rid of this directory entirely. Simple solution to just duplicate operations would be fine by me. We can colocate fragments to a central directory, since it makes more sense to share those.

packages/peregrine/lib/talons/CommunicationsPage/communicationsPage.gql.js
tjwiebell: @supernova-at commented about this recently, and I fully agree, it became redundant since we wanted to preserve the non-appended version for executing mutations (or lazy queries), eg. `setNewsletterSubscription` will execute the mutation.

packages/peregrine/lib/talons/AuthModal/useAuthModal.js
sirugh: Yea, I think duplicating operations is fine for now and we can optimize later.

tjwiebell: DevTime Discussion: Willing to accept duplication to delete queries out of this central location.

packages/peregrine/lib/talons/CommunicationsPage/useCommunicationsPage.js
```diff
@@ -2,13 +2,14 @@ import { useCallback, useMemo } from 'react';
 import { useMutation, useQuery } from '@apollo/client';
 
 import { useUserContext } from '../../context/user';
+import DEFAULT_OPERATIONS from './communicationsPage.gql';
 
 export const useCommunicationsPage = props => {
+    const { afterSubmit, operations = DEFAULT_OPERATIONS } = props;
```
supernova-at, line 8: Other considerations for this name: `gqlOperations`, `networkOperations`, `apolloOperations`

packages/peregrine/lib/talons/CommunicationsPage/communicationsPage.gql.js
tjwiebell: DevTime Discussion: Flatten out to `operations` and enforce operations append what type they are to the end of their name.

packages/peregrine/lib/talons/CommunicationsPage/useCommunicationsPage.js
tjwiebell: We haven't decided how we're going to return our raw data, but I would want it to match this input; eg. `dataOperations` returns `data`. I hate naming things, but this should be a simple find/replace. I'll poll the team today during peer review 👍

sirugh: `gqlOperations` ✔️

tjwiebell:
@magento import pr to magento-commerce/pwa-studio

magento-engcom-team:
@tjwiebell the pull request successfully imported.

packages/pagebuilder/lib/ContentTypes/Products/products.js
```diff
@@ -257,3 +256,38 @@ Products.propTypes = {
 };
 
 export default Products;
+
+export const GET_PRODUCTS_BY_SKU = gql`
```
tjwiebell, line 260: I opted not to apply this pattern to pagebuilder and treated this the same as deprecated components.

packages/peregrine/lib/talons/CreateAccount/createAccount.gql.js
```diff
@@ -0,0 +1,126 @@
...
+                grand_total {
+                    value
+                    currency
+                }
+            }
+        }
+    }
+`;
+
+export const MERGE_CARTS = gql`
```
tjwiebell, line 102: This mutation previously injected multiple component fragments into the response which I removed, under the assumption I would re-add when we tackle migrating component colocated queries to peregrine. To my surprise, I didn't spot any stale cache issues after merge with just this response shape; I suspect Apollo v3 now correctly re-runs those queries for us.

We can still re-add those fragments once components come over, but just wanted to point out this seems to be working as-is.

packages/peregrine/lib/talons/RootComponents/Category/category.gql.js
```diff
@@ -0,0 +1,71 @@
...
+            page_info {
+                total_pages
+            }
+            total_count
+        }
+    }
+`;
+
+export const GET_FILTER_INPUTS = gql`
+    query GetFilterInputsForCategory {
```
tjwiebell, line 56: I attempted along the way to make sure query names were unique, as I thought there was a lint rule about this. I've found a few duplicate names that I missed, but the linter doesn't care. Is this something we need to be mindful of?

packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
```diff
@@ -0,0 +1,143 @@
+import { useEffect, useMemo, useRef } from 'react';
```
tjwiebell, line 1: This is a bit of scope creep, but this RootComponent didn't have a talon. I broke it up without changing anything about the logic.

packages/peregrine/lib/talons/SignIn/signIn.gql.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+`;
+
+export const CREATE_CART = gql`
+    mutation CreateCartAfterSignIn {
+        cartId: createEmptyCart
+    }
+`;
+
+export const MERGE_CARTS = gql`
```
tjwiebell, line 29: Same treatment as `createAccount.gql.js`; fragments were removed but components still appear to be re-rendering correctly. Still worth focusing on in testing.

packages/peregrine/lib/talons/RootComponents/Category/category.gql.js
sirugh: Hmm - I think this might be related to caching, but also readability in network tab/query browsers. You might try naming two queries identically but having them request different data and see what happens in the cache. Would need to have identical inputs as well.

packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
sirugh: Careful - I'm fairly sure @eug123 is doing this already https://github.com/magento/pwa-studio/pull/2686/files#diff-c08d9f0ea0b6c1b26fb3111ffffeadd8

packages/peregrine/lib/targets/peregrine-intercept.js
```diff
@@ -7,7 +7,7 @@ module.exports = targets => {
         featuresByModule['@magento/peregrine'] = {
             cssModules: true,
             esModules: true,
-            graphQlQueries: true
+            graphqlQueries: true
```
tjwiebell, line 10: @zetlen 

zetlen-typo

packages/pwa-buildpack/lib/WebpackTools/PWADevServer.js
```diff
@@ -170,51 +167,12 @@ const PWADevServer = {
...
-                    );
-                });
                 /* istanbul ignore next: dummy next() function not testable */
                 const noop = () => {};
                 app.get('/graphiql', async (req, res) => {
                     if (!middleware) {
                         middleware = playgroundMiddleware({
-                            endpoint,
-                            tabs: await gatheringQueryTabs
+                            endpoint
```
tjwiebell, line 175: This can be discussed; we don't have any `.graphql` files anymore, so I didn't know what to render here, so I just removed the tabs entirely. I'm sure with some effort, we could generate tabs from our new pattern, but didn't seem worth it in this scope.

packages/peregrine/lib/talons/Footer/useFooter.js
```diff
@@ -1,12 +1,15 @@
...
+
 /**
  *
- * @param {*} props.query the footer data query
+ * @param {*} props.operations GraphQL operations used by talons
  */
-export const useFooter = props => {
-    const { query } = props;
-    const { data } = useQuery(query);
+export const useFooter = (props = {}) => {
```
tjwiebell, line 9: Discussion point: what is the preferred way to handle talons where `operations` is the **only** prop? We can either:

1. Do it this way and default to an empty object so operation destructuring below works
2. Don't defaults props to empty object and force that usage of talon with no operation override do `const talonProps = useFooter({})`

I have a preference for option 1.

packages/venia-ui/lib/components/Checkout/addressForm.js
```diff
@@ -179,3 +178,61 @@ AddressForm.propTypes = {
 };
 
 export default AddressForm;
+
+export const SET_GUEST_EMAIL_MUTATION = gql`
```
tjwiebell, line 182: For deprecated components like this one (old drawer checkout), I just inlined the operations. It didn't make sense to me to decompose into talon and migrate operations.

packages/venia-ui/lib/components/Checkout/index.js
```diff
@@ -1 +1,6 @@
+/**
+ * @deprecated and will be deleted in next release
```
tjwiebell, line 2: Looks like this never got deprecated, so we'll have to remove in next available release. Would have much preferred to just delete this component in this scope.

packages/venia-ui/lib/components/CmsBlock/cmsBlock.js
```diff
@@ -58,3 +57,14 @@ CmsBlockGroup.propTypes = {
 };
 
 export default CmsBlockGroup;
+
+export const GET_CMS_BLOCKS = gql`
```
tjwiebell, line 61: I couldn't find any usage of this component...so I inlined the operation.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -57,8 +56,7 @@ const VeniaAdapter = props => {
     } else {
         apolloClient = new ApolloClient({
             cache,
-            link,
-            resolvers
+            link
```
tjwiebell, line 59: We deprecated resolvers in last release, so safe to delete in this scope.

packages/peregrine/lib/talons/Footer/useFooter.js
sirugh: Option 1 :D

packages/venia-ui/lib/components/Checkout/index.js
sirugh: I think we need to figure out a better deprecation story. For now we should just mark components as deprecated but I would avoid comments indicating deletion time frame going forward.

We don't want to alienate anyone who has done a lot of customization on their legacy checkout overrides.

packages/peregrine/lib/talons/CreateAccount/createAccount.gql.js
tjwiebell: Add a TODO so we don't forget to re-add

packages/venia-ui/lib/components/CmsBlock/cmsBlock.js
tjwiebell: Consider _talonizing_ this

packages/venia-ui/lib/components/CategoryList/categoryList.js
```diff
@@ -47,13 +43,16 @@ const CategoryList = props => {
     ) : null;
 
     let child;
+
+    // TODO: Actually handle errors; the logic below will never allow this to render
```
supernova-at, line 47: Nice catch - do you mind creating a ticket for this?

packages/venia-ui/lib/components/CommunicationsPage/__tests__/communicationsPage.spec.js
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 import { createTestInstance } from '@magento/peregrine';
-import { useCommunicationsPage } from '@magento/peregrine/lib/talons/MyAccount/useCommunicationsPage';
+import { useCommunicationsPage } from '@magento/peregrine/lib/talons/CommunicationsPage/useCommunicationsPage';
```
supernova-at, line 3: Nice catch

packages/venia-ui/lib/components/Footer/__tests__/footer.spec.js
```diff
@@ -42,5 +29,4 @@ test('footer renders copyright', () => {
     );
 
     expect(instance.toJSON()).toMatchSnapshot();
-    expect(true).toBe(true);
```
supernova-at, line 45: You're bringing down our numbers! 🤣

supernova-at (approved):
If you want to duplicate `getPageSize` and make it part of the default operations in the two talons that use it, otherwise 👍

packages/peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/createAccount.gql.js
```diff
@@ -0,0 +1,108 @@
+import { gql } from '@apollo/client';
+
+export const CREATE_ACCOUNT = gql`
+    mutation CreateAccountAfterCheckout(
```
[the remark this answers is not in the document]
sirugh: The cool thing is it's not part of the api so we can change them easily whereas the export cannot be changed :)

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -2,29 +2,32 @@ import { useCallback, useEffect, useMemo, useRef } from 'react';
...
+export const useSearchPage = (props = {}) => {
     const {
-        queries: {
-            filterIntrospection,
-            getProductFiltersBySearch,
-            productSearch,
-            getPageSize
-        }
+        operations = DEFAULT_OPERATIONS,
+        queries: { getPageSize }
```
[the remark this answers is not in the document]
sirugh: Merge conflict, I'm guessing.

packages/peregrine/lib/talons/FilterModal/useFilterModal.js
```diff
@@ -2,10 +2,10 @@ import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useQuery } from '@apollo/client';
 import { useHistory, useLocation } from 'react-router-dom';
 
-import { useAppContext } from '@magento/peregrine/lib/context/app';
-
+import { useAppContext } from '../../context/app';
```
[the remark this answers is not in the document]
tjwiebell: Definitely. I'm not sure we got confirmation on scoped or non-scoped absolute paths, but it sounds like definitely _not_ relative, so I agree. I'll undo the import "fixes" I did 🤦

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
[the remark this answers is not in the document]
tjwiebell: Bingo 🐶  This came in from a recent merge, and since we've decided to split up the rest of component work, left this alone. In the next grooming I wanted to talk about [PWA-983](https://jira.corp.magento.com/browse/PWA-983) and [PWA-984](https://jira.corp.magento.com/browse/PWA-984), and how we want to break up/prioritize the rest.

Once this is merged too, we can start enforcing this prop structure and colocation going forward, so that migration scope doesn't continue to expand.

packages/venia-ui/lib/components/CategoryList/categoryList.js
tjwiebell: [PWA-990](https://jira.corp.magento.com/browse/PWA-990) 👍

packages/peregrine/lib/talons/CreateAccount/createAccount.gql.js
tjwiebell: This PR got cut off at just migration of operations in `venia-ui/lib/queries`. This operation used four very large fragments that we'll gradually add back as migrated. Good reminder though to add this detail to those followup tickets.

packages/peregrine/lib/talons/FilterModal/useFilterModal.js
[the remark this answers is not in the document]
tjwiebell: Reverted import change in 5f790c8. Next DevTime we should finalize this preference and begin cleaning up the codebase to be more consistent with that choice.

supernova-at (approved):
Amazing work, this is such a big improvement!

dpatil-magento:
QA Approved.

PR #2716, opened 2020-09-18
Remove max-height on code blocks in dev docs

brendanfalkowski:
## Description

In dev docs, code blocks have a max-height of 700px. I think this comes from Spectrum.

It's a bad default for dev docs, because it introduces multiple scrolling regions in long detailed technical documents. It causes code sample to "cut off" abruptly before the end of the sample. Depending on the OS there's no indication that content is missing due the inner scroll region.

For technical docs, it's better to just let code blocks have intrinsic height. Let the window scroll with no inner regions. It doesn't matter if the document is long or individual code blocks are long. Readability is most important.

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

### Before
Note: code is cut off with no visual indication that's happening.

Screen Shot 2020-09-18 at 3 53 33 PM

### After 

Screen Shot 2020-09-18 at 3 53 18 PM

## Checklist

### DIFF
--- a/pwa-devdocs/src/_scss/src/_code-block-style-overrides.scss
+++ b/pwa-devdocs/src/_scss/src/_code-block-style-overrides.scss
@@ -1,3 +1,7 @@
+.highlight {
+    max-height: none;
+}
+
 .highlight .gd,
 .highlight .gi {
     border: none;


jimbo (approved):

tjwiebell:
Tracking internally with [PWA-952](https://jira.corp.magento.com/browse/PWA-952).

jcalcaben:
@belbiy can you take a look at this PR?

belbiy:
I believe the max-height was added to eliminate the infinite scroll on some of the devdocs pages... The benefits and drawbacks of both limited and unlimited code blocks can be debated (i personally prefer unlimited as well).

One thing of note is that the change has to be made on a theme level, not on this project. Unless of course we want to support both scenarios. In that case we need to do following:
- Decide on a default code block behavior - is it going to be limited in height or not by default?
- Implement a class name for a special behavior (`.limited` class name? any suggestions?).

All this has to be done in a theme https://github.com/magento-devdocs/devdocs-theme

brendanfalkowski:
When your PR gets closed and you weren't even doing Hacktoberfest :)

https://youtu.be/Lx7NwPreJpQ?t=25

PR #2718, opened 2020-09-21
chore: Allow for async taps of targets

sirugh:
## Description

It may come to be that we need to perform some async calls to determine available routes. This change makes that possible, in a backwards compatible way that doesn't break anyone already tapping into the available targets.

```js
targets.own.routes.tapPromise(async routes => {
    const builtinRoutes = [
        {
            name: 'SavedPaymentsPage',
            pattern: '/saved-payments',
            exact: true,
            path: '../SavedPaymentsPage'
        },
        // ... other built in routes
    ];

    // Returns ['SavedPaymentsPage', ...] based on GQL introspection.
    const disabledRoutes = await getDisabledRoutes();

    return [...routes, ...builtinRoutes].filter(route =>
        disabledRoutes.contains(route.name)
    );
});
```

## Related Issue

Closes [PWA-933](https://github.com/magento/pwa-studio/pull/2718).

### Verification Steps

1. Run `yarn build` and make sure it completes without error.
2. Change line in venia-ui-intercept.js to `targets.own.routes.tapPromise(async routes => {` and run `yarn build` and it should complete without error.
3. `yarn watch:venia` and make sure you can hit the UI without error.
4. `yarn stage:venia` and make sure you can hit the UI without error.

## Checklist


packages/pwa-buildpack/lib/WebpackTools/configureWebpack/configureWebpack.js
```diff
@@ -166,7 +166,7 @@ async function configureWebpack(options) {
 
     /** @typedef {import('../../BuildBus/declare-base)} BuiltinTargets */
 
-    bus.getTargetsOf('@magento/pwa-buildpack').transformModules.call(x =>
+    bus.getTargetsOf('@magento/pwa-buildpack').transformModules.promise(x =>
```
zetlen, line 169:
```suggestion
    await bus.getTargetsOf('@magento/pwa-buildpack').transformModules.promise(x =>
```

packages/pwa-buildpack/lib/TestHelpers/testFullBuild.js
```diff
@@ -64,7 +64,7 @@ async function buildModuleWith(
     const resolver = new MagentoResolver({ paths, alias });
 
     const transforms = new ModuleTransformConfig(resolver);
-    bus.getTargetsOf('@magento/pwa-buildpack').transformModules.call(x =>
+    bus.getTargetsOf('@magento/pwa-buildpack').transformModules.promise(x =>
```
zetlen, line 67:
```suggestion
    await bus.getTargetsOf('@magento/pwa-buildpack').transformModules.promise(x =>
```

zetlen (changes requested):
This oughta do it.

zetlen (approved):
LGTM 👍 Thanks for boldly maintaining the targets framework. This change is backward-compatible!

tjwiebell (approved):
Made sense after your explanation and with some refresher of `webpack/tapable`. Would like to see an example of how position in the waterfall works, but that's probably easier to understand when a route makes use of this change (Wishlist maybe?).

revanth0212:
QA Done 👍

PR #2723, opened 2020-09-23
fix: Continue to Payment Information button gets enabled even before Payment section starts loading.

sirugh:
## Description

There is a bug due to either:

a) A bug in @apollo/client that makes refetches due to cache invalidations have `loading: false` **or**
b) Our inability to use `id` in our GQL `shipping_address` type 

When you click on the "Proceed to payment info" button, the button temporarily disables while the mutation for "selecting payment method" is in flight, but then re-enables before navigating to the next step.

To solve this we simply prevent the refetches from occurring, trusting that the data necessary will come back from the mutation.

If/when we finally get an `id` for each shipping address we can revert this change (and probably other changes related to type policy/merging).

## Related Issue

Closes [PWA-915](https://jira.corp.magento.com/browse/PWA-915).

### Verification Steps

1. Clear cache
2. Go to first item in tops, and add to cart.
3. Proceed to checkout and enter address. Click to go to method.
4. Select a non free method and click "proceed to payment info".
5. The button should disable immediately and never re-enable before you are navigated to the payment information view.
6. Verify that editing/changing the method once submitted also works.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
+++ b/packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
@@ -128,8 +128,6 @@ export const useShippingMethod = props => {
                 value.shipping_method
             );
 
-            setPageIsUpdating(true);
-
             try {
                 await setShippingMethodCall({
                     variables: {
@@ -142,13 +140,11 @@ export const useShippingMethod = props => {
                 });
             } catch {
                 return;
-            } finally {
-                setPageIsUpdating(false);
             }
 
             setIsUpdateMode(false);
         },
-        [cartId, setIsUpdateMode, setPageIsUpdating, setShippingMethodCall]
+        [cartId, setIsUpdateMode, setShippingMethodCall]
     );
 
     const handleCancelUpdate = useCallback(() => {
@@ -170,6 +166,10 @@ export const useShippingMethod = props => {
             onSave();
         }
     }, [hasData, onSave]);
+
+    useEffect(() => {
+        setPageIsUpdating(isSettingShippingMethod);
+    }, [isLoadingShippingMethods, isSettingShippingMethod, setPageIsUpdating]);
 
     // If an authenticated user does not have a preferred shipping method,
     // auto-select the least expensive one for them.
@@ -218,7 +218,7 @@ export const useShippingMethod = props => {
         errors,
         handleCancelUpdate,
         handleSubmit,
-        isLoading: isLoadingShippingMethods,
+        isLoading: isLoadingShippingMethods || isSettingShippingMethod,
         isUpdateMode,
         selectedShippingMethod: derivedSelectedShippingMethod,
         shippingMethods: derivedShippingMethods,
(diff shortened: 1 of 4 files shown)


packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.gql.js
```diff
@@ -36,14 +43,19 @@ export const SET_SHIPPING_METHOD = gql`
...
                 ...PriceSummaryFragment
-                # Intentionally do not re-fetch available shipping methods because
-                #  a) they are wrong in the mutation response
-                #  b) it is expensive to recalculate.
+                # We include the following fragments to avoid extra requeries
+                # after this mutation completes. This all comes down to not
+                # having ids for shipping address objects. With ids we could
+                # merge results.
+                ...ShippingInformationFragment
+                ...AvailableShippingMethodsCheckoutFragment
```
revanth0212, line 51: Looking at the previous comments in this fragment, we did not fetch available shipping methods on purpose because they can be faulty or are costly to calculate. Has something changed on the mutation level to make them accurate and fast?

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
```diff
@@ -171,6 +167,10 @@ export const useShippingMethod = props => {
         }
     }, [hasData, onSave]);
 
+    useEffect(() => {
+        setPageIsUpdating(isSettingShippingMethod);
```
revanth0212, line 171: Shouldn't this be `isLoadingShippingMethods || isSettingShippingMethod`?

revanth0212 (commented):
A couple of questions regarding the choice of code.

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.gql.js
sirugh: I was not able to replicate the "faulty" part of that comment. As for the slowness, sadly it is necessary in order to fix the bug where people get stuck in a state where they cannot continue to the next step. This is primarily because we don't get `id` back from `shipping_addresses` which means we cannot gracefully merge the array entries.

The TLDR is that we were blowing away old data when any query/mutation returned part of `shipping_addresses`. Which means it would re-query anyways. So this just gets it done up front.

I urge you to experiment with this. Remove the lines, see what's happening in the network tab.

packages/peregrine/lib/talons/CheckoutPage/useShippingMethod.js
sirugh: No, this is a functionally identical change to what was happening before in the submit callback.

revanth0212 (approved):
Other than the failing tests, it looks fine. Approved 👍

dpatil-magento:
QA Approved.

revanth0212 (approved):

supernova-at (approved):

PR #2725, opened 2020-09-24
Import from 'react-router-dom', not 'react-router'

jvaughan:
## Description

I needed to import from react-router-dom in order to use parameters in my static route.


jimbo (approved):

jimbo:
Thanks for submitting a PR to correct this. I also noticed this when I demonstrated this the other day. 😅

PR #2727, opened 2020-09-25
fix: offline homepage error

sirugh:
## Description

Network errors are always returned by `useQuery`, even if we use `cache-and-network`. Going offline and then navigating to a cached cms page results in the `data` from the query returning, but the network call being made. Since we're offline, a network error is returned by the hook, which was causing the `Page Fetch Error` to be rendered in the UI.

To be honest the error was really a bad UX anyways. We have better methods for handling page level errors such as toasts.

## Related Issue

Closes [PWA-708](https://jira.corp.magento.com/browse/PWA-708).

### Verification Steps

1. Open PR/deployed home page.
2. Now got to category or products page.
3. Turn network to offline in devtools.
4. click Venia logo to go back to home page.
5. Wait for 3-5 secs.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/Cms/useCmsPage.js
+++ b/packages/peregrine/lib/talons/Cms/useCmsPage.js
@@ -21,7 +21,8 @@ export const useCmsPage = props => {
         variables: {
             id: Number(id)
         },
-        fetchPolicy: 'cache-and-network'
+        fetchPolicy: 'cache-and-network',
+        nextFetchPolicy: 'cache-first'
     });
 
     const [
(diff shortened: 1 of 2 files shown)


revanth0212 (approved):
Any tests that need to be updated?

packages/peregrine/lib/talons/Cms/useCmsPage.js
```diff
@@ -60,7 +61,6 @@ export const useCmsPage = props => {
     return {
         cmsPage,
         hasContent,
-        error,
```
jimbo, line 63: I think I'd prefer to keep returning this value from the error, even if we don't use it in the component.

jimbo (commented):
Makes sense. One change noted.

jimbo (approved):

packages/peregrine/lib/talons/Cms/useCmsPage.js
sirugh: Reverted

dpatil-magento:
QA Approved.

PR #2735, opened 2020-09-28
[PWA-816] Enable store codes to be present in the URL path if configured

davemacaulay:
## Description

As a merchant I want the ability to enable having store codes present in the URL so that I can separate out my various content with a URL path. This PR introduces a new configurable env variable called `USE_STORE_CODE_IN_URL` which when enabled will place the current store code into the URL.

There are a few different pieces of functionality included with this change:

- Include a `basename` in the router to always include the store code when clicking on a link in the app
- Modify store switcher to ensure store code is correctly updated or added if not present in the URL
- Add routing logic to detect if the URL is manually changed and switch to the correct store (e.g. modify the URL and change the `default` store to the `french` store).
- Include globally available `AVAILABLE_STORE_VIEWS` for the app to consume for performance reasons

## Related Issue
https://jira.corp.magento.com/browse/PWA-816

### Verification Steps
**Store Switcher from Home Page:**
- Configure store with multiple store views with different locales / data, enable `USE_STORE_CODE_IN_URL=true` in `.env`
- Navigate to the PWA Studio store front and use the store switcher and verify the store code is present in the URL
- Validate store code is now included in the URL and the homepage is displayed
- Navigate around the site and ensure the store code stays in the URL

**Direct Visit of Store Code**
- Configure store with multiple store views with different locales / data, enable `USE_STORE_CODE_IN_URL=true` in `.env`
- Directly visit a store code which is different from the default store view (e.g. `/French`)
- Verify the store is loaded in the French store view on the home page

**Manual Change of Store Code in URL**
- Configure store with multiple store views with different locales / data, enable `USE_STORE_CODE_IN_URL=true` in `.env`
- Navigate to a product on the default store view
- Manually change the URL and replace the `default` store with a different store code (e.g. `/French`)
- Verify the store is changed to the new store code and the product is displayed with the correct data

## Checklist


larsroettig:
@davemacaulay this is cool I was short before to created this PR as well :)

packages/peregrine/lib/Router/resolveUnknownRoute.js
```diff
@@ -99,6 +99,17 @@ function remotelyResolveRoute(opts) {
...
+    const storeCodes = AVAILABLE_STORE_VIEWS.map(store => store.code);
+    let route = opts.route;
+
+    storeCodes.forEach(function(storeCode) {
+        const path = '/' + storeCode.toLowerCase();
+        if (route.startsWith(path) || route === path) {
+            route = route.substring(path.length, route.length);
+        }
+    });
+    route = route === '' ? '/' : route;
```
larsroettig, line 111: this is so magic and complex and relay readable

packages/pwa-buildpack/lib/Utilities/__tests__/__snapshots__/createDotEnvFile.spec.js.snap
```diff
@@ -54,6 +54,10 @@ MAGENTO_BACKEND_EDITION=CE
 #   - Default when not set: 
 STORE_VIEW_CODE=
 #
+#   Specify whether or not the store code URLs should be appended to the URLs of the store.
+#   - Default when not set: false
+USE_STORE_CODE_IN_URL=false
```
larsroettig, line 59: Needs added in docs as well

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -82,14 +82,46 @@ export const useStoreSwitcher = props => {
...
+            if (process.env.USE_STORE_CODE_IN_URL === 'true') {
+                const pathName = window.location.pathname;
+
+                // Check to see if we're on a page outside of the homepage
+                if (pathName !== '' && pathName !== '/') {
+                    const pathNameSplit = pathName.split('/');
+                    const pathStoreCode = pathNameSplit[1];
+
+                    // Is the current store code in the URL? If so replace it with the new one
+                    if (previousStoreCode === pathStoreCode) {
```
eug123, line 103: I think you can use `availableStores.get(pathStoreCode).isCurrent` here

packages/pwa-buildpack/lib/queries/getAvailableStoresConfigData.graphql
```diff
@@ -0,0 +1,19 @@
...
+        grid_per_page
+    }
+    availableStores {
+        base_currency_code
+        code
+        default_display_currency_code
+        id
+        locale
+        store_name
+    }
```
eug123, line 18: I think we need only `id`, `code` and `default_display_currency_code` from `availableStores` in this request for `AVAILABLE_STORE_VIEWS`

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
davemacaulay: While you could, the required code is actually much longer: `pathStoreCode && availableStores.get(pathStoreCode) && availableStores.get(pathStoreCode).isCurrent`. So I think it's best to keep the shorter version for readability.

davemacaulay: Okay updated it with the `has` then `get` API, thanks @eug123

packages/pwa-buildpack/lib/queries/getAvailableStoresConfigData.graphql
davemacaulay: Updated!

packages/peregrine/lib/Router/resolveUnknownRoute.js
davemacaulay: I've updated this with a much simpler expression which results in the same functionality 😄

packages/pwa-buildpack/lib/Utilities/__tests__/__snapshots__/createDotEnvFile.spec.js.snap
davemacaulay: @jcalcaben tells me the docs for the .env file is automatically generated.

packages/pwa-buildpack/envVarDefinitions.json
```diff
@@ -31,6 +31,12 @@
                     "type": "str",
                     "desc": "Specify the Magento store view code set in Stores > Settings > All Stores > Store View name",
                     "default": ""
+                },
+                {
+                    "name": "USE_STORE_CODE_IN_URL",
+                    "type": "bool",
+                    "desc": "Specify whether or not the store code URLs should be appended to the URLs of the store.",
```
jcalcaben, line 38:
```suggestion
                    "desc": "Appends the store code to the store URLs if value is true.",
```

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,96 @@ const VeniaAdapter = props => {
...
 
+    let storeCodeRouteHandler = '';
+    const browserRouterProps = {};
+    if (process.env.USE_STORE_CODE_IN_URL === 'true') {
+        const storeCode = storage.getItem('store_view_code') || STORE_VIEW_CODE;
+        browserRouterProps.basename = `/${storeCode}`;
+        const storeCodes = [];
+        const storeCurrencies = {};
+        if (
+            process.env.USE_STORE_CODE_IN_URL === 'true' &&
```
eug123, line 101: Do we need to check it again?

davemacaulay: Good call, updated

packages/peregrine/lib/talons/Header/__tests__/useStoreSwitcher.spec.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { useHistory } from 'react-router-dom';
 import createTestInstance from '@magento/peregrine/lib/util/createTestInstance';
 import { useStoreSwitcher } from '../useStoreSwitcher';
-import { mockSetItem } from '../../../util/simplePersistence';
+import { mockSetItem, mockGetItem } from '../../../util/simplePersistence';
```
eug123, line 5: We don't need `mockGetItem` anymore

eug123 (dismissed):

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,93 @@ const VeniaAdapter = props => {
...
+                    {children}
+                </BrowserRouter>
             </ReduxProvider>
         </ApolloProvider>
     );
 };
 
+// Create a new store link to include store codes and currency in the request
+VeniaAdapter.storeLink = setContext((_, { headers }) => {
+    // get the authentication token from local storage if it exists.
```
sirugh, line 166: Update this comment.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,93 @@ const VeniaAdapter = props => {
...
+                             * Only execute if one store code is present in the URL, multiple store codes will cause
+                             * the store state to break and cause weird side effects for the user
+                             */
+                            const regex = new RegExp(
+                                `(${storeCodes.join('|')})\/`,
+                                'g'
+                            );
+                            const storeCodesInUrl = window.location.pathname.match(
+                                regex
+                            ).length;
```
sirugh, line 130: This throws an error if I navigate to `<url>/fr` with a cold cache. For what it's worth, if you first navigate to `<url>/fr/` and then remove the slash it does not throw an error. Seems to conflict with your second set of verification steps.

packages/venia-concept/webpack.config.js
```diff
@@ -56,7 +63,8 @@ module.exports = async env => {
             STORE_NAME: JSON.stringify('Venia'),
             STORE_VIEW_CODE: process.env.STORE_VIEW_CODE
                 ? JSON.stringify(process.env.STORE_VIEW_CODE)
-                : JSON.stringify(storeConfigData.code)
+                : JSON.stringify(storeConfigData.code),
+            AVAILABLE_STORE_VIEWS: JSON.stringify(availableStores)
```
sirugh, line 67: To ensure storybook continues to function, changes to the webpack config such as global props or new queries must also be applied to the [storybook webpack config](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/.storybook/webpack.config.js)

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,93 @@ const VeniaAdapter = props => {
...
+                                storage.setItem(
+                                    'store_view_currency',
+                                    storeCurrencies[match.params.storeCode]
+                                );
+                            } else {
+                                console.warn(
+                                    'Multiple store codes present in URL.'
+                                );
+                            }
+                        }
```
sirugh, line 145: I'm seeing a warning when navigating. Looks like you need to return something from this block.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,93 @@ const VeniaAdapter = props => {
         return null;
     }
 
+    let storeCodeRouteHandler = '';
+    const browserRouterProps = {};
+    if (process.env.USE_STORE_CODE_IN_URL === 'true') {
+        const storeCode = storage.getItem('store_view_code') || STORE_VIEW_CODE;
+        browserRouterProps.basename = `/${storeCode}`;
```
sirugh, line 97: When I navigate to the root url without a locale such as https://magento-venia-concept-mhkx8.local.pwadev:8389/ I get a warning:

[Image from Gyazo](https://gyazo.com/eda94887070c51137bda5f0cc26596cc)

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,93 @@ const VeniaAdapter = props => {
...
+                                    'store_view_code',
+                                    match.params.storeCode
+                                );
+                                storage.setItem(
+                                    'store_view_currency',
+                                    storeCurrencies[match.params.storeCode]
+                                );
+                            } else {
+                                console.warn(
+                                    'Multiple store codes present in URL.'
```
sirugh, line 142: Should we/can we trim out extra store codes? Having multiple doesn't prevent us from rendering the page using the default (or first?) in the url.

sirugh (changes requested):
Verification step 3 mostly worked but I noticed that when on the `/locale/cart` page, hard navigating to a different store via url (`/otherlocale/cart`) caused an error:

[Image from Gyazo](https://gyazo.com/a114dface20eb0df73f2927d2ebd9bbe)

packages/venia-ui/lib/drivers/adapter.js
davemacaulay: Multiple store codes in the URL causes a very interesting redirect loop. The regex match pattern for the `<Route />` will read the last match, rather than the first which is annoying.

As this isn't a valid use case, and people shouldn't be trying to navigate in this manner in an actual e-commerce store (it's incredibly rare for someone to go change the URL rather than use the switcher). I opted to just `warn` during development incase someone encounters this issue while debugging some weird use case they've got.  I also don't like doing URL rewrites as it gets even messier making sure we don't break the entire URL.

packages/venia-ui/lib/drivers/adapter.js
davemacaulay: This is expected, the warning is a bit annoying but it mimics the Luma functionality here. You can still visit pages without the store code, but every link will get the store code included when you navigate. 

I believe this warning won't occur on a production system, I don't know if we should try suppress this during development?

packages/venia-ui/lib/drivers/adapter.js
davemacaulay: Done!

packages/venia-ui/lib/drivers/adapter.js
davemacaulay: Updated the regex pattern to search for a slash in front of the code, as that's guaranteed.

packages/venia-concept/webpack.config.js
davemacaulay: Updated :)

packages/venia-ui/lib/drivers/adapter.js
davemacaulay: Good catch, fixed.

davemacaulay:
@sirugh good catch, we have to do a full reload to make the static pages work properly. As we need to properly swap the `basename` and it doesn't really support doing so at runtime.

packages/venia-ui/.storybook/webpack.config.js
```diff
@@ -39,7 +44,8 @@ module.exports = async ({ config: storybookBaseConfig, mode }) => {
             STORE_VIEW_LOCALE: JSON.stringify(global.LOCALE),
             STORE_VIEW_CODE: process.env.STORE_VIEW_CODE
                 ? JSON.stringify(process.env.STORE_VIEW_CODE)
-                : JSON.stringify(storeConfigData.code)
+                : JSON.stringify(storeConfigData.code),
+            AVAILABLE_STORE_VIEWS: JSON.stringify(availableStores)
```
sirugh, line 48: There is also the `venia-concept` storybook config which I think we have so folks can use storybook after scaffolding.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,103 @@ const VeniaAdapter = props => {
...
+
+    // return the headers to the context so httpLink can read them
+    return {
+        headers: {
+            ...headers,
+            store: storeCode,
+            ...(storeCurrency && { 'Content-Currency': storeCurrency })
+        }
+    };
+});
```
sirugh, line 188: Specialized links. A+.

packages/venia-ui/lib/drivers/adapter.js
sirugh: Shouldn't warn in production. Worth double checking @dpatil-magento

sirugh (dismissed):
One last change re: scaffold storybook!

packages/pwa-buildpack/lib/queries/getAvailableStoresConfigData.graphql
```diff
@@ -0,0 +1,8 @@
+query availableStoresConfigData {
```
sirugh, line 1: Pointing out here that @tjwiebell is in the process of moving most/all of the queries next to their respective talons. Obviously this and other build time queries are unique and I don't know the plan for them. I assume they will just become `.gql` files next to the files that use them.

packages/venia-ui/lib/drivers/adapter.js
davemacaulay: This was @eug123's idea 😄

packages/peregrine/lib/Router/resolveUnknownRoute.js
```diff
@@ -99,6 +99,21 @@ function remotelyResolveRoute(opts) {
  * @returns {Promise<{type: "PRODUCT" | "CATEGORY" | "CMS_PAGE"}>}
  */
 function fetchRoute(opts) {
+    const storeCodes = AVAILABLE_STORE_VIEWS.map(store => store.code);
+    let route = opts.route;
+
+    storeCodes.forEach(function(storeCode) {
```
sirugh, line 105:
```suggestion
    // Trim store code off the route before querying.
    storeCodes.forEach(function(storeCode) {
```

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -80,16 +80,49 @@ export const useStoreSwitcher = props => {
...
-            // Refresh the page to re-trigger the queries once code/currency are saved in local storage.
-            history.go(0);
+            // Handle updating the URL if the store code should be present
+            if (process.env.USE_STORE_CODE_IN_URL === 'true') {
+                const pathName = window.location.pathname;
+
+                // Check to see if we're on a page outside of the homepage
+                if (pathName !== '' && pathName !== '/') {
+                    const pathNameSplit = pathName.split('/');
+                    const pathStoreCode = pathNameSplit[1];
```
sirugh, line 98:
```suggestion
                    const [, pathStoreCode] = pathName.split('/');
```

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -80,16 +80,49 @@ export const useStoreSwitcher = props => {
...
+                            )
+                        );
+                    } else {
+                        // If there is no store code present, include it and reload
+                        window.location.assign(
+                            `/${storeCode}${
+                                pathName && pathName !== '/' ? pathName : ''
+                            }`
+                        );
+                    }
```
sirugh, line 118: I feel like this could be simplified. Strip out any existing store code in the url and add in the selected one plus all the rest of the path and params.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -80,16 +80,49 @@ export const useStoreSwitcher = props => {
...
+                                `/${storeCode}`
+                            )
+                        );
+                    } else {
+                        // If there is no store code present, include it and reload
+                        window.location.assign(
+                            `/${storeCode}${
+                                pathName && pathName !== '/' ? pathName : ''
+                            }`
+                        );
```
sirugh, line 117: Search and filter query params are dropped when switching store code - make sure to add them back in these `assign`s.

packages/peregrine/lib/Router/resolveUnknownRoute.js
```diff
@@ -99,6 +99,21 @@ function remotelyResolveRoute(opts) {
  * @returns {Promise<{type: "PRODUCT" | "CATEGORY" | "CMS_PAGE"}>}
  */
 function fetchRoute(opts) {
+    const storeCodes = AVAILABLE_STORE_VIEWS.map(store => store.code);
```
sirugh, line 102: This needs to always match the result from the `availableStores` query used to generate the store switcher options otherwise it'll be possible that we miss a new route + new store.

I'm betting you can repro this if you:

1) Build/deploy/visit a page.
2) Add a new store code to you backend that will result in a new entry of `availableStores`
3) Clear cache but don't "rebuild" your pwa studio bundle.
4) Refresh the browser, use the store switcher to change to the new store.
5) I'm betting that it will try to resolve that full route here since it won't get trimmed below.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,103 @@ const VeniaAdapter = props => {
...
+        /**
+         * The current store code won't be matched as it's included as the basename, if we match another store
+         * we need to change the current store.
+         */
+        storeCodeRouteHandler = (
+            <Switch>
+                <Route path={`/:storeCode(${storeCodes.join('|')})?`}>
+                    {({ match }) => {
+                        if (
+                            match.params.storeCode &&
```
sirugh, line 117: When I clear my cache and hit https://magento-venia-concept-mhkx8.local.pwadev:8389/fr/ I am seeing `match.params.storeCode === 'fr'` but when I refresh after it loads it's undefined. Why?

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -86,15 +90,103 @@ const VeniaAdapter = props => {
...
+        }
+
+        /**
+         * The current store code won't be matched as it's included as the basename, if we match another store
+         * we need to change the current store.
+         */
+        storeCodeRouteHandler = (
+            <Switch>
+                <Route path={`/:storeCode(${storeCodes.join('|')})?`}>
+                    {({ match }) => {
```
jimbo, line 115: Now that we have hooks, we should avoid function-as-a-child wherever possible. Not only are components more readable and performant, they can call hooks—whereas function children can't.

It's even more relevant in this case because the function child here is performing side effects, such as writing to local storage and reloading `window.location`. These effects [must be encapsulated in `useEffect`](https://reactjs.org/docs/hooks-reference.html#useeffect):

> Mutations, subscriptions, timers, logging, and other side effects are not allowed inside the main body of a function component (referred to as React’s render phase). Doing so will lead to confusing bugs and inconsistencies in the UI. 

Consider extracting this logic to a real component.

```es6
import { useRouteMatch } from "react-router-dom"

const StoreCodeRoute => () => {
  const pattern = `/:storeCode(${storeCodes.join()})?`
  const match = useRouteMatch(pattern)

  // ...

  useEffect(() => {
    // storage.setItem()
  }, [])
}
```

jimbo (changes requested):

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -80,16 +80,49 @@ export const useStoreSwitcher = props => {
...
+                if (pathName !== '' && pathName !== '/') {
+                    const pathNameSplit = pathName.split('/');
+                    const pathStoreCode = pathNameSplit[1];
+
+                    // Is the current store code in the URL? If so replace it with the new one
+                    if (
+                        availableStores.has(pathStoreCode) &&
+                        availableStores.get(pathStoreCode).isCurrent
+                    ) {
+                        window.location.assign(
```
jimbo, line 105: Generally, `react-router` (and its `history` dependency) are our abstraction for interacting with the location. This helps with testing and will eventually be important for SSR. I think we should be using [its API](https://reactrouter.com/core/api/history) rather than `window.location` directly.

Rather than assigning to `window.location`, can we `history.push()` or `history.replace()` as necessary?

jimbo (dismissed):

packages/venia-ui/lib/drivers/adapter.js
davemacaulay: Because the base name is including `fr`, meaning fr is no longer considered in the `<Route />` match.

sirugh: I see - it is because the basename has changed, so once `fr` is the basename it doesn't match because it is part of the base.

packages/peregrine/lib/Router/resolveUnknownRoute.js
sirugh: Discussed with @davemacaulay. This is a trade off where we pre-optimize by generating this global at build time using GQL. 

The upside is that we can render routes without having to first make an `availableStores` query.

The downside is that implementers must re-build/deploy their PWA when they create a new store and want it accessible in the UI.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
sirugh: I believe this was used to work around the basename being updated in `BrowserRouter`. If your current basename is `/en`, `history` will append after the basename but not replace it, even if you use `replace`.

If there is a way to re-assign basename from within the `BrowserRouter` we would need to do that here.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
sirugh: Done.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
sirugh: I see now that this was all necessary to ensure we don't mistake a path for a store code. ie

`https://example.com/search.html` where `search.html` would be `pathName` but not a valid store code. In that case we need to insert the store code before the path.

packages/venia-ui/lib/drivers/adapter.js
sirugh: Updated.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -80,16 +80,53 @@ export const useStoreSwitcher = props => {
...
+                    ) {
+                        const newPath = `${pathName.replace(
+                            `/${pathStoreCode}`,
+                            `/${storeCode}`
+                        )}${params}`;
+
+                        window.location.assign(newPath);
+                    } else {
+                        // Otherwise include it and reload.
+                        const newPath = `/${storeCode}${
```
revanth0212, line 116: I would suggest you move the `pathName && pathName !== '/' ? pathName : ''` check outside of the string template for ease of reading.

packages/pwa-buildpack/envVarDefinitions.json
```diff
@@ -31,6 +31,12 @@
                     "type": "str",
                     "desc": "Specify the Magento store view code set in Stores > Settings > All Stores > Store View name",
                     "default": ""
+                },
+                {
+                    "name": "USE_STORE_CODE_IN_URL",
+                    "type": "bool",
+                    "desc": "Appends the store code to the store URLs if value is true.",
```
revanth0212, line 38:
```suggestion
                    "desc": "Includes the store code in the store URLs if value is true.",
```

I suggest that because we are technically not appending to the URLs, we are appending to the store's domain.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
sirugh: Removed the conditional entirely since it is assumed from the wrapping condition anyways.

packages/venia-ui/lib/components/StoreCodeRoute/storeCodeRoute.js
```diff
@@ -0,0 +1,55 @@
...
+    useEffect(() => {
+        // The current store code won't be matched as it's included as the
+        // basename, if we match another store we need to change the current
+        // store.
+        if (match.params.storeCode) {
+            // Only execute if one store code is present in the URL, multiple
+            // store codes will cause the store state to break and cause weird
+            // side effects for the user.
+            const regex = new RegExp(`\/(${storeCodes.join('|')})`, 'g');
+            const storeCodesInUrl = window.location.pathname.match(regex);
```
revanth0212, line 35: I am not sure if this is working as expected.

Try on this JS snippet

```es6
const storeCodes = ['en', 'fr', 'en-uk', 'frs'];
const pattern = `/:storeCode(${storeCodes.join('|')})?`;
const regex = new RegExp(`\/(${storeCodes.join('|')})`, 'g');

'/frs/home.html'.match(regex) // should be ['/frs'] but returns ['/fr']
'/en-uk/home.html'.match(regex) // should be ['en-uk'] but returns ['/en']
```

Let me know if that is expected behavior.

revanth0212 (changes requested):
Nice work guys. I have a couple questions.

1. How should the app react if a wrong store view code is used in the URL? I was thinking it should render using the default value and also change the store code in the URL to the default value. Right now the app fails to render the page and the user will not have an idea of what is happening. They would refresh the page but nothing's going to change.

image

2. Should we still consider this a `minor` change (judging by the GitHub labels)? This looks like a breaking change because if the backend does not support multiple views the build process seems to fail while fetching the `availableStores` query. Also, for a smooth upgrade, I would suggest we employ a nondestructive way of handling this. Instead of failing may be a warning during the build step.

image

packages/venia-ui/lib/components/StoreCodeRoute/storeCodeRoute.js
sirugh: Good catch!

sirugh:
@revanth0212 thanks for the review! To answer your questions:

> How should the app react if a wrong store view code is used in the URL? I was thinking it should render using the default value and also change the store code in the URL to the default value. Right now the app fails to render the page and the user will not have an idea of what is happening. They would refresh the page but nothing's going to change.

The app is reacting correctly - an invalid store code is treated the same as an invalid url. More specifically, all routes are checked to see if the first path after the url is a valid code, and uses it if so. If it isn't a valid store code the app treats the entire path as an attempt to use the default or configured route, which will always return the invalid route error you see in your screenshot.

> Should we still consider this a minor change (judging by the GitHub labels)?

Good point. Technically a breaking change in that you have to update your backend to 2.4.1 to be able to use it.

packages/venia-ui/lib/components/StoreCodeRoute/storeCodeRoute.js
sirugh: @davemacaulay suggested ordering `storeCodes` by length to solve this issue for now.

revanth0212 (approved):

revanth0212:
There are merge conflicts. Can you please take care of those? Thanks.

revanth0212 (approved):

dpatil-magento:
@davemacaulay @sirugh  Store specific categories are not being loaded.

<a href="https://gyazo.com/af8f0b5569129c9af30af8a4a23ff818">Image from Gyazo</a>

sirugh:
[Image from Gyazo](https://gyazo.com/cc8550ad54fd6df91fe138e802d37fb7)

Oops

eug123 (dismissed):

dpatil-magento:
QA Approved. @sirugh Please check on failed unit tests.

sirugh:
> QA Approved. @sirugh Please check on failed unit tests.

@dpatil-magento done!

supernova-at (approved):

PR #2736, opened 2020-09-28
[bug] fix data sort error

jimbo:
## Description

Fix a bug where data from GQL was being sorted in place. Apollo doesn't allow data arrays to be mutated.

## Related Issue
PWA-946.

### Verification Stakeholders
- @dpatil-magento 
- @sirugh 
- @tjwiebell 

### Verification Steps
1. Visit a category page, then a product page.
2. Verify the error is no longer thrown.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
+++ b/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
@@ -44,10 +44,16 @@ export const useBreadcrumbs = props => {
 
             return (
                 breadcrumbData &&
-                breadcrumbData.sort(sortCrumbs).map(category => ({
-                    text: category.category_name,
-                    path: getPath(category.category_url_path, categoryUrlSuffix)
-                }))
+                breadcrumbData
+                    .map(category => ({
+                        category_level: category.category_level,
+                        text: category.category_name,
+                        path: getPath(
+                            category.category_url_path,
+                            categoryUrlSuffix
+                        )
+                    }))
+                    .sort(sortCrumbs)
             );
         }
     }, [categoryUrlSuffix, data, loading]);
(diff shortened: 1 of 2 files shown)


packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
```diff
@@ -44,10 +44,16 @@ export const useBreadcrumbs = props => {
 
             return (
                 breadcrumbData &&
-                breadcrumbData.sort(sortCrumbs).map(category => ({
-                    text: category.category_name,
-                    path: getPath(category.category_url_path, categoryUrlSuffix)
-                }))
+                breadcrumbData
+                    .map(category => ({
+                        category_level: category.category_level,
```
sirugh, line 49: This will work but we will return `category_level` now which isn't necessary for rendering.

jimbo: Meh. Good data for someone to have anyway.

supernova-at (approved):

packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
supernova-at: Was going to comment here too. It's needed for `sortCrumbs`:

```js
const sortCrumbs = (a, b) => a.category_level > b.category_level;
```

We could `map` it again at the end to take it out but yeah, a similar thing happened in `useMagentoRoute` recently where a developer wanted information that we already had but weren't returning because we weren't using it to render.

sirugh (approved):
I'd prefer to not return extra data but I won't fight you on it :P

dpatil-magento:
QA Approved.

PR #2756, opened 2020-10-06
[Fix bug]: Correct Message on Search Page When Searching for Less Than 3 Characters

### LINKED ISSUE #2753: [bug]: Incorrect Message on Search Page When Searching for Less Than 3 Characters
### Describe the bug ###
If a user submits a search for less than 3 characters, then the resulting page shows "No results found!". This is incorrect messaging & should be worded differently with something like "Must search for at least 3 characters". See attachments

 ### To reproduce ###
1. Create a product with name "simple1"
2. Go to PWA storefront & search for "s"
3. Note issue
4. Search for "si"
5. Note issue
6. Search for "sim"
7. Note that product is returned

### Expected behavior ###
A message is displayed that your search term didn't meant the minimum length requirement. If the message from GraphQL is user friendly, just display that, otherwise come up with one to display.

**Screenshots**
1-char
2-char
3-char

### JIRA Link ###
[PWA-683](https://jira.corp.magento.com/browse/PWA-683)

huykon:
## Description

If a user submits a search for less than 3 characters, then the resulting page shows "Must search for at least 3 characters" instead of "No results found!". 

## Related Issue

Closes #2753.

### Verification Steps

Follow the steps of that issue
1. Create a product with name "simple1"
2. Go to PWA storefront & search for "s"
3. The search bar show correctly message "Must search for at least 3 characters"

## Checklist


schensley:
Hi just had a look at the PR instance and it appears that suggested search results aren't displayed until at least 5 characters are entered. 
Screen Shot_A_2020-10-07 at 12 23 47 PM
Screen Shot_B_2020-10-07 at 12 23 40 PM

huykon:
@schensley the first you search **simp**, it will not found any product because we have not add any product name is **simple** on backend, I don't make any effect to search query so this is no problem.

schensley:
Ah, yes, a little bit of weirdness. I see that generally it does display suggested results based on as few as 3 characters. I did see that I had to type 5 for the category (?) of SHIRTS to fetch something (see below). Otherwise, this is working as expected.  UX approved.

Screen Shot 2020-10-08 at 11 08 22 AM
Screen Shot 2020-10-08 at 11 08 28 AM

packages/venia-ui/lib/components/SearchBar/autocomplete.js
```diff
@@ -105,6 +105,13 @@ const Autocomplete = props => {
                 },
                 { resultCount: resultCount }
             )
+        )
+        .set(
+            'CHARACTER_NOT_ENOUGH',
+            formatMessage({
+                id: 'autocomplete.characterNotEnough',
+                defaultMessage: 'Must search for at least 3 characters'
```
tjwiebell, line 113: First, my apologies, because I know this is the message suggested in the ticket, which is an improvement from Luma; but after reviewing this code this isn't quite accurate. Technically you **can** search by hitting enter, GraphQL is not imposing this search limit, our implementation is.

@schensley @jcalcaben - may need some help on this copy 😄 
- `Enter at least three characters to load results`
- `Search term must be at least three characters`

packages/venia-ui/lib/components/SearchBar/autocomplete.js
```diff
@@ -105,6 +105,13 @@ const Autocomplete = props => {
                 },
                 { resultCount: resultCount }
             )
+        )
+        .set(
+            'CHARACTER_NOT_ENOUGH',
+            formatMessage({
+                id: 'autocomplete.characterNotEnough',
```
tjwiebell, line 112:
```suggestion
                id: 'autocomplete.invalidCharacterLength',
```

tjwiebell (changes requested):
Looks good 👍  just need to tweak some naming and get some feedback on preferred messaging. Nice work!

packages/venia-ui/lib/components/SearchBar/autocomplete.js
jcalcaben: Maybe something like:

> Search term must be at least 3 characters

huykon: @jcalcaben I updated as your recommend

tjwiebell (dismissed):
Looks good, thanks for making those changes. Please add a test, kicking to QA.

dpatil-magento:
QA looks good, waiting for conflict to resolve and test to be added.

tjwiebell (approved):

PR #2759, opened 2020-10-06
[PWA-835] Localize Checkout

anthoula:
## Description

Localize all strings within the Checkout area of the application. Include localized strings in default en_US language pack

## Related Issue

https://jira.corp.magento.com/browse/PWA-835

### Verification Stakeholders

@sirugh 
@dpatil-magento 
### Verification Steps

1. Add a product to cart
2. Go to Checkout page
3. Go through each step of the checkout process (with guest and with authorized user)
3. Verify strings are translated where applicable

## Checklist


sirugh (changes requested):
I still need to actually verify using a non-english translation pack but this looks pretty great! There's a few things in here I do want to discuss but mostly are just my opinions about the id format. Apologies for the pedantry :P

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/shippingMethod.js
```diff
@@ -19,7 +20,12 @@ import defaultClasses from './shippingMethod.css';
 import shippingMethodOperations from './shippingMethod.gql';
 
 const initializingContents = (
-    <LoadingIndicator>{'Loading shipping methods...'}</LoadingIndicator>
+    <LoadingIndicator>
+        <FormattedMessage
```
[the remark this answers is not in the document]
anthoula: Yes, it still translates, and renders the French translation in my French store.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -56,16 +58,34 @@ const OrderConfirmationPage = props => {
...
+                        defaultMessage={'Thank you for your order!'}
+                    />
                 </h2>
-                <div
-                    className={classes.orderNumber}
-                >{`Order Number: ${orderNumber}`}</div>
+                <div className={classes.orderNumber}>
+                    <FormattedMessage
+                        id={'checkoutPage.orderNumber'}
+                        defaultMessage={'Order Number'}
```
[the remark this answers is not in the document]
anthoula: The defaultMessage should be able to stand by itself, until the translation completes, so opted to remove extra artifacts like colons that could imply info is missing.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -267,7 +319,15 @@ const CheckoutPage = props => {
 
     return (
         <div className={classes.root}>
-            <Title>{`Checkout - ${STORE_NAME}`}</Title>
+            <Title>
+                {formatMessage(
+                    {
+                        id: 'checkoutPage.titleCheckout',
```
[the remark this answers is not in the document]
anthoula: There is a collision with `checkoutPage.titleReceipt`, so needed greater specificity.

packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
```diff
@@ -204,11 +237,24 @@ const CheckoutPage = props => {
             </div>
         ) : null;
 
-        const guestCheckoutHeaderText = isGuestCheckout
-            ? 'Guest Checkout'
-            : customer.default_shipping
-            ? 'Review and Place Order'
-            : `Welcome ${customer.firstname}!`;
+        let guestCheckoutHeaderText;
```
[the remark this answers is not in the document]
anthoula: Renamed the var per our discussion, thank you

sirugh (approved):
Good work!

dpatil-magento:
1. Shipping Method Edit overlay buttons not translated. 
2. Guest checkout order confirmation page > `Create Account` button does not have right UI.

sirugh:
> Guest checkout order confirmation page > Create Account button does not have right UI.

@dpatil-magento can you clarify?

dpatil-magento:
> > Guest checkout order confirmation page > Create Account button does not have right UI.
> 
> @dpatil-magento can you clarify?

@sirugh 

image

dpatil-magento:
@anthoula  2 is fixed, 1 translation is fixed but button name is changed to `CONFIRM`, it was `UPDATE`, expected?

image

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/createAccount.js
```diff
@@ -120,12 +151,22 @@ const CreateAccount = props => {
...
+                        })}
                     />
                 </div>
                 <div className={classes.actions}>
-                    <Button disabled={isDisabled} type="submit">
-                        {'Create Account'}
+                    <Button
+                        disabled={isDisabled}
+                        type="submit"
+                        className={classes.create_account_button}
```
sirugh, line 164: Will probably want to add `composes ... from Button` to the custom css class. Forgot to tell you! My bad.

dpatil-magento:
@anthoula Last one :)  In mobile mode button text flows out in Shipping Info and in shipping method button extends beyond available screen size.  Not sure what expected or these are designed?

image

packages/venia-concept/.envbackup2
```diff
@@ -0,0 +1,194 @@
+######## PWA Studio Environment Variables ######################################
```
sirugh, line 1: I don't think you meant to commit this.

packages/venia-ui/lib/components/CheckoutPage/ShippingMethod/updateModal.js
```diff
@@ -31,13 +33,17 @@ const UpdateModal = props => {
     return (
         <Dialog
             confirmText={'Update'}
+            confirmTranslationId={'global.updateButton'}
```
sirugh, line 36: I think it makes sense to have the parent component do the translation and simply provide the text prop instead of text and id. Similar to the `title` prop below.

anthoula: There are some strings specified in the `defaultProps` of the Dialog component, and needed a way to translate the defaults while also supporting the ability to replace the default text

packages/venia-ui/lib/components/clickable.css
```diff
@@ -6,5 +6,4 @@
     line-height: 1;
     pointer-events: auto;
     text-align: center;
-    white-space: nowrap;
```
sirugh, line 9: This might be a bit too aggressive as `clickable` is the base dependency upon which the [Button](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/Button/button.css#L2) class composes itself. If the team is good with it, I would just reset the `white-space` rule on buttons.

I think @jimbo would have an opinion here and I remember @supernova-at doing work in this area, specifically around wrapping text in dialogs. I think in Andy's case he was avoiding a situation where you have two buttons but one wraps which presents an odd UX.

supernova-at: Yes exactly.

The direction from UX is that we don't want text in buttons to be multi-line. If a client / merchant / partner has a need for it (translated text is too long, for example) then yes, they should treat that as an exception and either use a different prompt or apply a class override to wrap that specific button's text.

I agree that it is too aggressive here, and likely too aggressive in the button component as well.

jimbo: You two are correct. Not only do we not necessarily want `white-space: nowrap` on every clickable element, we don't want it on _any_ clickable element.

In early designs, we attempted to define a rule prohibiting clickable elements from wrapping. It didn't work. In some languages, authors simply can't create a meaningful call to action when they only have one line to work with; the right call to action is simply too verbose to fit, so it must wrap.

dpatil-magento:
QA Approved.

PR #2767, opened 2020-10-08
Feature to unmount child components in the Dialog component.

revanth0212:
## Description

I have added a new prop that will decide whether to unmount or just hide child components when the dialog is hidden.

## Related Issue
Closes none

### Verification Stakeholders
@jimbo 

### Verification Steps
none

## Checklist


packages/venia-ui/lib/components/Dialog/dialog.js
```diff
@@ -94,7 +108,7 @@ const Dialog = props => {
                             {maybeCloseXButton}
                         </div>
                         <div className={classes.body}>
-                            <div className={classes.contents}>{children}</div>
+                            <div className={classes.contents}>{contents}</div>
```
sirugh, line 111: The reason we did not just render `null` when `!isOpen` was because of the desire for a smooth transition to the closed state. Does this maintain that transition or is there a jarring squish of the dialog when the content goes `null`?

revanth0212: It does. The smooth transition is done if the whole dialog is unmounted. We are only unmounting the child components.

sirugh (dismissed):
So long as the transition continues to work correctly this is a good solution :)

sirugh (approved):

dpatil-magento:
Ran regression, QA approved.

PR #2778, opened 2020-10-13
fix: fix unsupported webp for safari #2551

### LINKED ISSUE #2551: [bug]: responses with 404 code should not be cached

mercs600:
## Description

resolves #2551 and #2707. The main issue for us was that image requests with 404 response were cached. The second issue is when you open images on chrome which supports webp and after that you  try to open the same image request on safari - where webp is not suported  - you will get broken request. This tweak is working for us on production. 

## Related Issue
#2551 and #2707.

### Verification Steps
Go with scenario from #2551 and #2707.

@zetlen


packages/pwa-buildpack/lib/Utilities/addImgOptMiddleware.js
```diff
@@ -29,9 +37,20 @@ function addImgOptMiddleware(app, config) {
...
-            debug: cacheDebug
-        });
+        cacheMiddleware = cache(
+            cacheExpires,
+            (req, res) =>
+                hastily.hasSupportedExtension(req) && res.statusCode === 200,
+            {
+                debug: cacheDebug,
+                appendKey: req =>
+                    IsWebpSupported(req) ? 'supported' : 'regular',
```
tjwiebell, line 47: This definitely solves the problem, but I'm concerned why it is needed. The `Vary` header in the response contains the `accept` header, and this lib doesn't seem to be respecting it. If we proceed with this workaround, I would probably just want `appendKey: req => req.headers.accept` while we investigate why this lib isn't respecting the `Vary` header.

tjwiebell (dismissed):
Great work! This solves both problems and is a huge improvement for anyone using this middleware in production. I have some suggestions to make the cache key less opinionated, but I'm open to discussing.

revanth0212 (approved):
Basically, we will be checking for the status code before caching so we don't cache invalid responses. Nice work.

dpatil-magento:
QA Approved.

packages/pwa-buildpack/lib/Utilities/addImgOptMiddleware.js
```diff
@@ -29,9 +29,15 @@ function addImgOptMiddleware(app, config) {
...
-        cacheMiddleware = cache(cacheExpires, hastily.hasSupportedExtension, {
-            debug: cacheDebug
-        });
+        cacheMiddleware = cache(
+            cacheExpires,
+            (req, res) =>
+                hastily.hasSupportedExtension(req) && res.statusCode === 200,
+            {
+                debug: cacheDebug,
+                appendKey: req => req.get('accept')
```
mercs600, line 38: @tjwiebell it will be enough to support or not webp images for safari ?

PR #2781, opened 2020-10-14
[PWA-1009] Localize Form Validators

anthoula:
## Description

Localize all strings within form validation of the application

## Related Issue

https://jira.corp.magento.com/browse/PWA-1009

### Verification Stakeholders

@dpatil-magento 
@sirugh 
### Verification Steps

1. Go to the forms in the app such as Create Account form.
2. Enter all supported invalid scenarios for the form fields.
3. Make sure form error validation messages are translated.

## Checklist


packages/venia-ui/lib/util/__tests__/formValidators.spec.js
```diff
@@ -64,25 +64,25 @@ describe('isRequired', () => {
     test('it returns a string for an invalid string', () => {
```
sirugh, line 64: Can you update all these test titles?

packages/venia-ui/lib/components/Field/message.js
```diff
@@ -1,18 +1,31 @@
...
+    const { formatMessage } = useIntl();
     const { asyncError, error } = fieldState;
     const errorMessage = error || asyncError;
 
     const classes = mergeClasses(defaultClasses, propClasses);
     const className = errorMessage ? classes.root_error : classes.root;
+    let translatedErrorMessage;
 
-    return <p className={className}>{errorMessage || children}</p>;
+    if (errorMessage) {
```
sirugh, line 18: As written, this assumes `asyncError` is the same format, but I think @jimbo said that the fieldState no longer returns an `asyncError` prop.

Can you validate that? If it no longer has the prop, you can remove the supporting code here for `asyncError` and then your current conditional here works. However, If we have to continue supporting `asyncError` we should just use `if(error)` here and use its props within this conditional.

packages/venia-ui/i18n/en_US.json
```diff
@@ -318,6 +317,17 @@
     "stockStatusMessage.message": "An item in your cart is currently out-of-stock and must be removed in order to Checkout.",
     "taxSummary.estimatedTax": "Estimated Tax",
     "taxSummary.tax": "Tax",
+    "validation.hasLengthAtLeast": "Must contain at least {value} character(s).",
+    "validation.hasLengthAtMost": "Must not exceed {value} character(s).",
+    "validation.hasLengthExactly": "Must contain exactly {value} character(s).",
+    "validation.invalidAbbrev": "State \"{value}\" is not an valid state abbreviation.",
```
sirugh, line 323:
```suggestion
    "validation.invalidAbbreviation": "State \"{value}\" is not a valid state abbreviation.",
```

packages/venia-ui/lib/util/formValidators.js
```diff
@@ -61,17 +83,31 @@ export const validateRegionCode = (value, values, countries) => {
...
+                'Country "US" does not contain any available regions.'
+        };
+        return invalidRegions;
     }
 
     const region = regions.find(({ code }) => code === value);
     if (!region) {
-        return `State "${value}" is not an valid state abbreviation.`;
+        const invalidAbbrev = {
+            id: 'validation.invalidAbbrev',
```
sirugh, line 106:
```suggestion
            id: 'validation.invalidAbbreviation',
```

sirugh (approved):

dpatil-magento:
QA Approved.

PR #2786, opened 2020-10-16
[PWA-893] My Account: Order History - Remove @client directive

tjwiebell:
## Description

This feature is being included in PWA v8, and to support 2.4.0, it was developed with mock data and uses the @client directive so it doesn't hit the actual backend for data. Once all of our backing instances have been upgraded to 2.4.1, our support matrix for the next release (v9) will shift to 2.4.1-2.4.2, and this @client directive can safely be removed.

This should go smoothly, but a full round of QA will need to happen now that Devagouda Patil can test against live data.

Acceptance Criteria
@client directive removed
MOCK_ORDERS and CUSTOM_TYPES are removed, and useTypePolices call in useOrderHistoryPage is removed
Configurable product images should load with real data.

## Related Issue

* [[PWA-893](https://jira.corp.magento.com/browse/PWA-893)] My Account: Order History - Remove @client directive
### Verification Steps

1. Login as Customer with Orders and go to Order History Page
2. Verify that everything continues rendering as it did before with mock data

## Checklist


packages/peregrine/lib/talons/OrderHistoryPage/orderRow.gql.js
```diff
@@ -1,8 +1,8 @@
 import { gql } from '@apollo/client';
 
-export const GET_PRODUCT_THUMBNAILS_BY_SKU = gql`
-    query GetProductThumbnailsBySku($skus: [String!]!) {
-        products(filter: { sku: { in: $skus } }) {
+export const GET_PRODUCT_THUMBNAILS_BY_URL_KEY = gql`
```
tjwiebell, line 3: Had to change this over to query by URL Key, since the product_sku returned by orders was variant skus which will return no results from the backend. Issue was escalated to GraphQL team, and they will be adding new fields in 2.4.2 so we shouldn't need to make a secondary query to get this data.

packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
```diff
@@ -4,13 +4,11 @@ import { useQuery } from '@apollo/client';
...
-import { useTypePolicies } from '../../hooks/useTypePolicies';
+import DEFAULT_OPERATIONS from './orderHistoryPage.gql';
 
-export const useOrderHistoryPage = props => {
-    const { queries, types } = props;
-    const { getCustomerOrdersQuery } = queries;
-
-    useTypePolicies(types);
+export const useOrderHistoryPage = (props = {}) => {
+    const { operations = DEFAULT_OPERATIONS } = props;
```
tjwiebell, line 10: @revanth0212 - There was overwhelming support for your proposed pattern of merging operations; do you want to propose a solution in this PR? I don't recall what we said needed to happen with classify and generalizing the shallow merge utility.

packages/venia-ui/__mocks__/react-intl.js
```diff
@@ -7,6 +7,6 @@ const intl = reactIntl.createIntl({
 
 module.exports = {
     ...reactIntl,
-    FormattedMessage: jest.fn(({ defaultMessage }) => defaultMessage),
+    FormattedMessage: jest.fn(({ defaultMessage, id }) => defaultMessage || id),
```
tjwiebell, line 10: This should be discussed; @revanth0212 solved this by providing his own mock in each test file, which I think we should replacing this one for all tests. Added this simple solution if we're okay just swapping components for one of these props (fallback to id when defaultMessage isn't provided).

packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
revanth0212: Honestly, I don't remember either if we wanted to create a utility or just do the shallow merge in every PR, but I can add a utility and surrounding code in another branch carved out of this. If the team is ok with it, we can merge it into this PR. How does that sound?

packages/venia-ui/__mocks__/react-intl.js
revanth0212:
```jsx
jest.mock('react-intl', () => ({
    FormattedMessage: props => (
        <div componentName="Formatted Message Component" {...props} />
    )
}));
```
I usually do it like this because while adding snapshot tests it is important to capture all the props that a component is provided.

If we did this instead,
```jsx
FormattedMessage: jest.fn(({ defaultMessage, id }) => defaultMessage || id)
```
There can be false positives because this component returns the same even if the value of `defaultMessage` changes from `''` to `null` or `false` for instance.

revanth0212: Checkout https://github.com/magento/pwa-studio/pull/2792 

I have updated the mock. The snapshots now can store more information about the formatted message component including child components if any.

tjwiebell:
@magento import pr to magento-commerce/pwa-studio

magento-engcom-team:
@tjwiebell the pull request successfully imported.

schensley:
This looks good @tjwiebell .  Thanks.  UX Approved.

packages/peregrine/lib/talons/OrderHistoryPage/orderHistoryContext.js
```diff
@@ -0,0 +1,31 @@
+import React, { createContext, useContext, useMemo } from 'react';
+
+import DEFAULT_OPERATIONS from './orderHistoryContext.gql';
+import { useQuery } from '@apollo/client';
+
+const OrderHistoryContext = createContext();
+
+export const OrderHistoryContextProvider = props => {
+    const operations = Object.assign({}, DEFAULT_OPERATIONS, props.operations);
```
tjwiebell, line 9: Will fix this once #2794 lands

fooman: @tjwiebell it's merged

packages/peregrine/lib/talons/OrderHistoryPage/orderRow.gql.js
sirugh: > which will return no results from the backend.

Is this because of our sample data or is it by design of the API?

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/shippingInformation.js
```diff
@@ -30,14 +31,12 @@ const ShippingInformation = props => {
                 />
             </div>
             <div className={classes.name}>
-                <span>{firstname}</span>
-                <span>{lastname}</span>
+                <span>{fullName}</span>
```
sirugh, line 34: Same comment here re: needing `span`.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/shippingInformation.js
```diff
@@ -30,14 +31,12 @@ const ShippingInformation = props => {
...
             </div>
             <div className={classes.addressLine1}>{street}</div>
             <div className={classes.addressLine2}>
                 {additionalAddressString}
             </div>
-            <div className={classes.country}>{country_code}</div>
```
sirugh, line 40: Same comment here re: class.

packages/peregrine/lib/talons/OrderHistoryPage/orderHistoryContext.js
```diff
@@ -0,0 +1,31 @@
...
+    const storeConfig = useMemo(() => {
+        return {
+            productURLSuffix: data
+                ? data.storeConfig.product_url_suffix
+                : '.html'
+        };
+    }, [data]);
+
+    return (
+        <OrderHistoryContext.Provider value={storeConfig}>
```
sirugh, line 25: Upon further review I was wondering why this component was even necessary. Were there plans to extend it beyond just providing access to the `storeConfig`? Was the creation of this provider meant to avoid the list of N products each attempting to query for the store config?

packages/peregrine/lib/talons/OrderHistoryPage/orderRow.gql.js
tjwiebell: That is by design; configurable variants all have Visibility = Not Visible Individually, so GraphQL is correct in not returning them. Couple JIRA tickets were created to address this though; [this one](https://jira.corp.magento.com/browse/MC-38534) so we don't even need to do a second query for images, and [this one](https://jira.corp.magento.com/browse/MC-38536) so parent sku is returned (which is also a blocker to doing re-orders).

sirugh (changes requested):
It's so nice that we can, with minimal changes, just pull out `@client` directives and it works!

packages/peregrine/lib/talons/OrderHistoryPage/orderHistoryContext.js
tjwiebell: Exactly; I wanted to avoid the potential thundering herd of all products making this query, and the common parent was so many layers up it didn't seem ideal to prop drill. This can also lift nicely to an app-wide context when we get around to the Admin Config epic, didn't feel appropriate to take that on in this scope, but I would expect to see a solution similar to this one, just higher up in the tree.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/paymentMethod.js
```diff
@@ -35,8 +24,7 @@ const PaymentMethod = props => {
                     defaultMessage="Payment Method"
                 />
             </div>
-            <div className={classes.payment_type}>{typeString}</div>
-            <div className={classes.payment_last_four_digits}>{last_four}</div>
```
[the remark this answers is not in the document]
tjwiebell: Lack of data in the response.

packages/peregrine/lib/talons/OrderHistoryPage/orderHistoryContext.js
```diff
@@ -0,0 +1,31 @@
...
+import { useQuery } from '@apollo/client';
+
+const OrderHistoryContext = createContext();
+
+export const OrderHistoryContextProvider = props => {
+    const operations = Object.assign({}, DEFAULT_OPERATIONS, props.operations);
+    const { getProductURLSuffixQuery } = operations;
+
+    const { data } = useQuery(getProductURLSuffixQuery, {
+        fetchPolicy: 'cache-and-network'
```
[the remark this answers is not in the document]
tjwiebell: Since there are no `storeConfig` mutations, I didn't think there would be a scenario where this would change, so felt the real `cache-and-network` was appropriate here and wouldn't lead to unnecessary re-fetches. I'm not 100% on that though, and would like clarification on our desired fetch policies; I had assumed we were only leveraging `nextFetchPolicy: 'cache-first'` to be backwards compatible during the v3 upgrade, and intended to use `cache-and-network` as designed at some point.

packages/venia-ui/i18n/en_US.json
```diff
@@ -362,5 +362,7 @@
     "wishlistPage.disabledMessage": "Sorry, this feature has been disabled.",
     "wishlistPage.fetchErrorMessage": "Something went wrong. Please refresh and try again.",
     "wishlistPage.headingText": "Favorites Lists",
-    "wishlistPage.wishlistDisabledMessage": "The wishlist is not currently available."
+    "wishlistPage.wishlistDisabledMessage": "The wishlist is not currently available.",
+    "orderDetails.waitingOnTracking": "Waiting for tracking information",
+    "orderItems.itemsHeading": "Items"
```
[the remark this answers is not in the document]
tjwiebell: Ideally we auto-generate as much as possible; relying on checkboxes is fragile, but fine for en_US for now. I do not think developers should be required to maintain the french pack, since it's entirely for testing purposes. Since we have the linting rule in place, I don't think that requirement would catch missed translations or add any value.

packages/peregrine/lib/talons/OrderHistoryPage/orderHistoryContext.js
tjwiebell: Updated in 6d3da3c. I really wish I could figure out jsdoc generics/templates so Intellisense could understand this merge utility 😭

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/billingInformation.js
```diff
@@ -30,14 +31,12 @@ const BillingInformation = props => {
...
             <div className={classes.name}>
-                <span>{firstname}</span>
-                <span>{lastname}</span>
+                <span>{fullName}</span>
             </div>
             <div className={classes.addressLine1}>{street}</div>
             <div className={classes.addressLine2}>
                 {additionalAddressString}
             </div>
-            <div className={classes.country}>{country_code}</div>
```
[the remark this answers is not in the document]
tjwiebell: Removed in 6d3da3c

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/billingInformation.js
```diff
@@ -30,14 +31,12 @@ const BillingInformation = props => {
                 />
             </div>
             <div className={classes.name}>
-                <span>{firstname}</span>
-                <span>{lastname}</span>
+                <span>{fullName}</span>
```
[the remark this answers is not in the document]
tjwiebell: No we do not, removed the `div` in 6d3da3c. I'm not really a fan of the overuse of divs here either, but I left the rest alone.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/item.js
```diff
@@ -45,9 +45,10 @@ const Item = props => {
                     resource={thumbnail.url}
                 />
             </Link>
-            <Link className={classes.name} to={itemLink}>
-                {product_name}
-            </Link>
+            <div className={classes.nameContainer}>
+                <Link to={itemLink}>{product_name}</Link>
+            </div>
+
```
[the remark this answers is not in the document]
tjwiebell: Weird that this didn't fail linting, but removed in 6d3da3c.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/shippingInformation.js
tjwiebell: Removed in 6d3da3c.

packages/venia-ui/lib/components/OrderHistoryPage/OrderDetails/shippingInformation.js
tjwiebell: Removed in 6d3da3c.

packages/peregrine/lib/talons/OrderHistoryPage/orderHistoryContext.js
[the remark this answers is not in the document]
sirugh: You're correct that we don't expect mutations to affect `storeConfig` however I believe the `cache-and-network` policy alone will re-query the network if the cached object changes at all.

One example might be this query just gets `storeConfig.product_url_suffix` while another gets `storeConfig.store_name`. I believe the addition of `store_name` to the cached object would trigger a requery unless you have `nextFetchPolicy: cache-first`. Would need to double check.

This is really only an issue while the component is mounted, so if you're not on the order history page it won't matter. Likely not a big deal and I wouldn't spend too much time on it.

sirugh (approved):

dpatil-magento:
I have already updated about these finding to @tjwiebell , posting here just for reference. - 

1. Simple product qty/price field misplaced when compared to configurable products.
2. Bundle product thumbnail also shown with simple product when the simple product is part of bundle product.
3. Latest order displayed at bottom instead of Top.
4. Gift card row is missing in Order Total
5. Bug Again/Print Receipt links are dead, can we keep them or remove?
6. Tracking number is not a link and needs some updates around upper/lower cases.
8. john.doe@example.com/ 123123^q  has error toast on mcstaging backend.
9. Order with 5 products display +1 instead of showing the 5th order in place of +1.
10. Street 1 and 2 don't have spacing.

tjwiebell:
> Simple product qty/price field misplaced when compared to configurable products.
> Order with 5 products display +1 instead of showing the 5th order in place of +1.

Initially reviewed with UX, @schensley approved. Might be worth double checking if we'd like to move config options to the third column to keep things in consistent columns, but not a blocker.

> Bundle product thumbnail also shown with simple product when the simple product is part of bundle product.
> Bug Again/Print Receipt links are dead, can we keep them or remove?
> Street 1 and 2 don't have spacing.
> If a order has multiple shipping packages then only first shipping info is displayed.

Fixed in 1c1a478

> Latest order displayed at bottom instead of Top.
> Gift card row is missing in Order Total

Created [MC-38927](https://jira.corp.magento.com/browse/MC-38927) and [MC-38928](https://jira.corp.magento.com/browse/MC-38928); neither were possible with existing coverage.

> Tracking number is not a link and needs some updates around upper/lower cases.

Left as plain strings and fixed up label to adhere to suggestion from UX in JIRA comment about expected rendering. We can address linkage in the ticket you created about the Luma view that we've broken for tracking shipments. 

> john.doe@example.com/ 123123^q has error toast on mcstaging backend.

Somehow there is an order without a shipping address on this account. Since this is something we do not support currently, I think we can ignore.

Screen Shot 2020-11-05 at 3 50 46 PM

fooman:
@tjwiebell what would a solution to fix the order with no shipping address look like? This is something we will need rather soon (I'd imagine PWA Studio for Commerce too as otherwise one couldn't for example order just a virtual gift card).

sirugh:
> @tjwiebell what would a solution to fix the order with no shipping address look like?

We have an upcoming epic to work on additional product types and I imagine that work would get handled there or sooner. To start, I'm curious if the UI has any way to "know" that there won't be a shipping address. It's not like we query for "required checkout steps" and I'm not sure there's even an API for that. Might have to do some UI logic and check if all items are virtual/don't need shipping.

dpatil-magento:
@tjwiebell Some more observations - 
10 (continued). Multiple Shipment info is displayed correctly now but single shipment with multiple packages only first package tracking is displayed, order ending in 575 is below screen
image

11. Discount row is displayed always, expected? I was hoping to see only when order has discount applied.
12. Only first 20 orders are returned by Graphql even when more orders are available. 
13. When an order is placed making it as an out of stock product, on accessing order history `TypeError: Cannot read property ‘url’ of undefined`

tjwiebell:
> @tjwiebell what would a solution to fix the order with no shipping address look like? This is something we will need rather soon (I'd imagine PWA Studio for Commerce too as otherwise one couldn't for example order just a virtual gift card).

This is a fair point, and a solution so simple we might as well handle now. While it has grown quite large, the intention here wasn't to mess with the existing implementation of components, but given some changes were necessary with live data, I can take on this scope. A bigger point to address is that we're all following our schemas correctly; myself included, I expect data even though non-required fields are nullable, and our designs and implementations don't account for this.

tjwiebell:
@dpatil-magento 

- Fixed 10, 13, and null shipping address in 5604e84. Just used a placeholder label for null shipping until the Virtual Product epic addresses the rest.
- 11 has been UX approved, leaving for now.
- 12 is expected, we have a Story to add pagination and search to this view

dpatil-magento:
Thanks @tjwiebell almost there, some minor ones - 
13 (continued) - Its fixed, but Thumbnail image is missing, expected? 
14 - Localization - When Discount price is 0 then it has $ sign instead of Euro.

15 - On switching store, previous store data is shown and then data is updated for new store.
<a href="https://gyazo.com/89e3200eec088586484d001463f1bea1">Image from Gyazo</a>

tjwiebell:
@dpatil-magento 

13. Missing thumbnail is expected given we have to do a second query to get images. Since the item is now OOS, GraphQL won't return any data about it anymore. There are fixes landing in 2.4.2 to will remove the need for this second query, and this should resolve itself.

14. Tripled checked with UX, decided to dependently render this row like you suggested. Fixed in b8eb620.

15. I'm not gonna get near store code switching and cache invalidation, but we can create a bug for the Owls to see if this is something that needs addressed. I would expect this view not to care about store and render orders placed on both stores anyway, since carts/orders are shared across website scope.

sirugh:
> I'm not gonna get near store code switching and cache invalidation, but we can create a bug for the Owls to see if this is something that needs addressed. I would expect this view not to care about store and render orders placed on both stores anyway, since carts/orders are shared across website scope.

You're seeing stale data because we render from cache. Once the store code is changed using the switcher, we refresh the page and re-query with the new store code in the appropriate header. After refresh, but before the new query returns, you will see cached data. That's just how it works with any query on reload when configured to use `cache-and-network`.

dpatil-magento:
QA Approved.

supernova-at (approved):

PR #2787, opened 2020-10-17
a11y fixes for filter modal

### LINKED ISSUE #1222: [feature]: Filters (Layered Navigation) - Add keyboard focus and the way tab order should work
As part of [#797](https://github.com/magento/pwa-studio/pull/797) - Layered navigation UI feature is added. But the keyboard focus and tab order does not work as expected.

Per meeting with [@soumya-ashok](https://github.com/soumya-ashok), these are the changes we've identified that need to be made:
* Focus on close button first
* “Escape” should always close modal no matter what depth/element is focused, should remember filter state
* Add outline/border to focused elements
* Add better aria labels to actionable for voiceover (get jimothy’s input)
* Kill tooltip or remove a11y arrow key from tooltip ie aria-ignore.
* Using space to select/unselect a color moves focus back to first color. Should remain on selected color/element. Apply treatment to other filter items
* Shift-tab on close should move focus to apply filters/resetfilters/ or last filter category. “Tab/focus trapping”
* On “apply filters” focus should move back to filter button
* On “reset filters” focus should move to first filter category button *not* the close button.

Additional:
* Make width 360px.
* Make expand from left instead of bottom.

Helpful Links:
* <https://webaim.org/techniques/keyboard/>

dani97:
## Description

Resolves A11y for filter modal

## Related Issue

Closes #1222

### Verification Stakeholders

@dpatil-magento 
@soumya-ashok 

### Verification Steps

1. Go to the Category  page.
2. Open the layered navigation, verify tab order, key controls
3. Check for other a11y features

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/FilterModal/CurrentFilters/currentFilters.js
+++ b/packages/venia-ui/lib/components/FilterModal/CurrentFilters/currentFilters.js
@@ -33,7 +33,11 @@ const CurrentFilters = props => {
         return elements;
     }, [classes.item, filterState, removeItem]);
 
-    return <ul className={classes.root}>{filterElements}</ul>;
+    return (
+        <ul className={classes.root} aria-label="Current Filters">
+            {filterElements}
+        </ul>
+    );
 };
 
 CurrentFilters.propTypes = {
--- a/packages/venia-ui/lib/components/FilterModal/filterModal.js
+++ b/packages/venia-ui/lib/components/FilterModal/filterModal.js
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { array, arrayOf, shape, string } from 'prop-types';
 import { X as CloseIcon } from 'react-feather';
@@ -12,6 +12,8 @@ import CurrentFilters from './CurrentFilters';
 import FilterBlock from './filterBlock';
 import FilterFooter from './filterFooter';
 import defaultClasses from './filterModal.css';
+
+import { tabbable } from 'tabbable';
 
 /**
  * A view that displays applicable and applied filters.
@@ -34,6 +36,7 @@ const FilterModal = props => {
 
     const classes = mergeClasses(defaultClasses, props.classes);
     const modalClass = isOpen ? classes.root_open : classes.root;
+    const modalRef = useRef(null);
 
     const filtersList = useMemo(
         () =>
@@ -66,9 +69,49 @@ const FilterModal = props => {
         </div>
     ) : null;
 
+    function changeFocusOnModal(event) {
+        if (isOpen) {
+            event.target.focus();
+        }
+    }
+
+    function handleKeyActions(event) {
+        if (isOpen) {
+            if (event.keyCode === 27) {
+                handleClose(event);
+            }
+            const element = modalRef.current;
+            const focusableEls = tabbable(element);
+            const firstFocusableEl = focusableEls[0];
+            const lastFocusableEl = focusableEls[focusableEls.length - 1];
+            if (event.keyCode === 9 || event.key === 'Tab') {
+                if (event.shiftKey) {
+                    /* shift + tab */
+                    if (document.activeElement === firstFocusableEl) {
+                        lastFocusableEl.focus();
+                        event.preventDefault();
+                    }
+                } /* tab */ else {
+                    if (document.activeElement === lastFocusableEl) {
+                        firstFocusableEl.focus();
+                        event.preventDefault();
+                    }
+                }
+            }
+        }
+    }
+
     return (
         <Portal>
-            <aside className={modalClass}>
+            {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
+            <aside
+                className={modalClass}
+                tabIndex="-1"
+                ref={modalRef}
+                onKeyDown={handleKeyActions}
+                onTransitionEnd={changeFocusOnModal}
+                role="dialog"
+            >
                 <div className={classes.body}>
                     <div className={classes.header}>
                         <h2 className={classes.headerTitle}>
@@ -77,7 +120,7 @@ const FilterModal = props => {
                                 defaultMessage={'Filters'}
                             />
                         </h2>
-                        <button onClick={handleClose}>
+                        <button onClick={handleClose} aria-label="close">
                             <Icon src={CloseIcon} />
                         </button>
                     </div>
@@ -87,7 +130,9 @@ const FilterModal = props => {
                         filterState={filterState}
                     />
                     {clearAll}
-                    <ul className={classes.blocks}>{filtersList}</ul>
+                    <ul className={classes.blocks} aria-label="Filters">
+                        {filtersList}
+                    </ul>
                 </div>
                 <FilterFooter
                     applyFilters={handleApply}
--- a/packages/venia-ui/package.json
+++ b/packages/venia-ui/package.json
@@ -47,6 +47,7 @@
     "react-slick": "~0.25.2",
     "redux-actions": "~2.6.4",
     "redux-thunk": "~2.3.0",
+    "tabbable": "~5.1.2",
     "uuid": "~3.3.2"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -15586,6 +15586,11 @@ symbol.prototype.description@^1.0.0:
     es-abstract "^1.17.0-next.1"
     has-symbols "^1.0.1"
 
+tabbable@~5.1.2:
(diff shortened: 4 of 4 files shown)


dani97:
Focus order corrected
Focus trap implemented
filter closes on esc

dani97:
> Focus order corrected
> Focus trap implemented
> filter closes on esc

Focus trap not working as expected

packages/venia-ui/lib/components/FilterModal/filterModal.js
```diff
@@ -66,9 +77,42 @@ const FilterModal = props => {
...
+                        lastFocusableEl.focus();
+                        event.preventDefault();
+                    }
+                } /* tab */ else {
+                    if (document.activeElement === lastFocusableEl) {
+                        firstFocusableEl.focus();
+                        event.preventDefault();
+                    }
+                }
+            }
```
dani97, line 102: is this logic can be converted to an util for other modals and dialogs

schensley:
@dani97 The tab order seems generally good and correct. If possible, it would be ideal that when the shopper opens the layered navigation menu that it takes precedence for the tab order, so that the shopper tabs through the menu items rather than tabbing through the underlying page first.

dani97:
> @dani97 The tab order seems generally good and correct. If possible, it would be ideal that when the shopper opens the layered navigation menu that it takes precedence for the tab order, so that the shopper tabs through the menu items rather than tabbing through the underlying page first.

@schensley Is focus trap implemented on modal not working for you? I tested in firefox and chrome in linux. It works for me

tjwiebell:
@dani97 - The focus trap is also not working for me, using Chrome v86 and the deployed PR instance. I see that you increased the timer, but it might be better to look into animation events to determine when to fire those actions. Please investigate, and let me know if you need any assistance.

dani97:
thanks for the idea @tjwiebell. this works without timeout workaround 😄

packages/venia-ui/lib/components/FilterModal/filterModal.js
dani97: @tjwiebell

schensley:
@dani97 Sorry, my misunderstanding I was checking the site navigation.  Sorry, I'm looking at the "filter drawer" now.  This looks good. UX approved.

packages/venia-ui/lib/components/FilterModal/CurrentFilters/currentFilters.js
```diff
@@ -33,7 +33,11 @@ const CurrentFilters = props => {
         return elements;
     }, [classes.item, filterState, removeItem]);
 
-    return <ul className={classes.root}>{filterElements}</ul>;
+    return (
+        <ul className={classes.root} aria-label="Current Filters">
```
tjwiebell, line 37: I believe we would want to open this for translation; please use `formatMessage` for this label. Similar to this in [AddressBookPage](https://github.com/magento/pwa-studio/blob/1494c40366538ead7cbb41c873666352986198ba/packages/venia-ui/lib/components/AddressBookPage/addressBookPage.js#L26:L29).

packages/venia-ui/lib/components/FilterModal/filterModal.js
```diff
@@ -66,9 +69,49 @@ const FilterModal = props => {
         </div>
     ) : null;
 
+    function changeFocusOnModal(event) {
```
tjwiebell, line 72: Both of these should be moved to `useFilterModal` and follow standard conventions.

`const handleFocusOnOpen = useCallback(event => {...})`

packages/venia-ui/lib/components/FilterModal/filterModal.js
```diff
@@ -66,9 +69,49 @@ const FilterModal = props => {
...
+
+    function handleKeyActions(event) {
+        if (isOpen) {
+            if (event.keyCode === 27) {
+                handleClose(event);
+            }
+            const element = modalRef.current;
+            const focusableEls = tabbable(element);
+            const firstFocusableEl = focusableEls[0];
+            const lastFocusableEl = focusableEls[focusableEls.length - 1];
```
tjwiebell, line 86: This seems like something you could initialize on open and not need to calculate again on every key press; since we just need to measure once the ref is attached, I'm thinking a callback ref is the best solution here.

```jsx
const filterModalRef = useCallback(node => {
  if (node !== null) {
    // do your calc here and set first/last into local state variable
  }
}, []);
```

packages/venia-ui/lib/components/FilterModal/filterModal.js
tjwiebell: Let's not over optimize yet, but I do agree that this could be decomp'd into a general utility at some point.

packages/venia-ui/lib/components/FilterModal/filterModal.js
```diff
@@ -77,7 +120,7 @@ const FilterModal = props => {
                                 defaultMessage={'Filters'}
                             />
                         </h2>
-                        <button onClick={handleClose}>
+                        <button onClick={handleClose} aria-label="close">
```
tjwiebell, line 123: Please use `formatMessage`

packages/venia-ui/lib/components/FilterModal/filterModal.js
```diff
@@ -87,7 +130,9 @@ const FilterModal = props => {
                         filterState={filterState}
                     />
                     {clearAll}
-                    <ul className={classes.blocks}>{filtersList}</ul>
+                    <ul className={classes.blocks} aria-label="Filters">
```
tjwiebell, line 133: Please use `formatMessage`

packages/venia-ui/package.json
```diff
@@ -47,6 +47,7 @@
     "react-slick": "~0.25.2",
     "redux-actions": "~2.6.4",
     "redux-thunk": "~2.3.0",
+    "tabbable": "~5.1.2",
```
tjwiebell, line 50: We're willing to accept the size of this lib if you can guarantee it doesn't end up in the main vendor bundle; this might require that `FilterModal` is lazy loaded if it already didn't shake out that way on its own. You can run `yarn stats:venia` to compare bundle sizes before and after this change.

tjwiebell (changes requested):
This functions perfectly, nice work 👍  some suggestions for making this more consistent with the rest of the code base, and some areas for optimization. Please let me know if you have any questions, or need any help finishing this up.

packages/venia-ui/package.json
dani97: https://github.com/magento/pwa-studio/pull/2843 @tjwiebell this pr introduces `react-aria` for focus trap, if that is merged I will rewrite this pr.

tjwiebell: Good eye; this issue did get a little outdated and we very recently started experimenting with the Adobe maintained `react-aria` library, which we probably should have included as implementation guidance. Since this solution with the tabbable lib worked, and we didn't provide that guidance, we were prepared to accept as-is, and re-factor later. I think we're confident enough in the results of that Spike that `react-aria` would be our preferred solution, using the `FocusScope` HOC here.

I don't know the timeline of that Spike getting delivered, but if you would like to preemptively switch this implementation to `react-aria`, it would be very much appreciated; but again, very much optional.

tjwiebell:
@dani97 - Just wanted to make sure you saw [this comment](https://github.com/magento/pwa-studio/pull/2787#discussion_r516028449); I didn't @ reply so very likely you didn't get notified. Did you want to try a re-factor to `react-aria`? If you're spread thin on availability we're happy to finish this up internally, just let us know 🙂

dani97:
> @dani97 - Just wanted to make sure you saw [this comment](https://github.com/magento/pwa-studio/pull/2787#discussion_r516028449); I didn't @ reply so very likely you didn't get notified. Did you want to try a re-factor to `react-aria`? If you're spread thin on availability we're happy to finish this up internally, just let us know 🙂 

@tjwiebell I will refactor it with react-aria  before next week

dani97:
@sirugh @tjwiebell I tried with react aria focusScope component but the tab order is not working properly in accordion, the tab focus is not shifting to next block instead it is contained within the element instead of modal. Any suggestions how to fix it

sirugh:
> @sirugh @tjwiebell I tried with react aria focusScope component but the tab order is not working properly in accordion, the tab focus is not shifting to next block instead it is contained within the element instead of modal. Any suggestions how to fix it

Can you push to a new branch or show a gif of the issue? Hard to know what's going on without seeing code or an example.

dani97:
@sirugh I have committed what I have tried here  1222-test-react-aria

sirugh:
Closing this - we have an internal ticket to track this work and have referenced this draft PR in the github issue for our records. Should whomever creates this work build upon this PR I'll make sure to give credit :)

PR #2789, opened 2020-10-19
PWA-973: Use jsx-no-literals rule 

sirugh:
## Description

Do not merge until PWA-1033 is merged.

This PR 
* adds the jsx-no-literals eslint rule with a configuration that meets our needs. 
* Updates eslint-plugin-react to 7.21.5 (latest).

It will prevent direct string usage in jsx but allow strings to be passed as `id`s. It cannot catch strings passed as `const`s but that's not something we should/can handle anyways

The following will error:

```
<div>{'oops!'}</div>
```

The following will not error:

```
const noError = 'No error!';
<div>{noError}</div>
```

```
<FormattedMessage id={'im.allowed'} />
```

## Related Issue

Closes [PWA-973](https://jira.corp.magento.com/browse/PWA-973).

### Verification Steps

1. Check out PR.
2. `yarn lint` should not error
3. Add a string literal to some jsx in a component, like `<span>{'literal'}</span>`.
4. `yarn lint` should error

## Checklist


.eslintrc.js
```diff
@@ -1,9 +1,18 @@
...
-        'no-useless-escape': 'off'
+        'no-useless-escape': 'off',
+        'react/jsx-no-literals': [
+            'error',
+            {
+                allowedStrings: [],
+                ignoreProps: true,
+                noStrings: true
+            }
+        ]
```
sirugh, line 15: I decided to add the rule to our eslint file as I don't want to introduce an error rule to the generic `magento-eslint`.

I'm not against it, but I think it's a bit opinionated and only applies to us because we are using i18n/format message and want to use this rule to not forget strings.

anthoula: The props that we would like it to pick up are mostly for `label` and `title`. Doesn't seem to be a way to allow those, sadly.

packages/venia-ui/lib/components/Checkout/shippingForm.js
```diff
@@ -35,16 +35,21 @@ const ShippingForm = props => {
 
     const classes = mergeClasses(defaultClasses, props.classes);
 
+    const formHeading = 'Shipping Information';
+    const shippingMethodLabel = 'ShippingMethod';
```
anthoula, line 39: Looks like this string should be `Shipping Method`

sirugh: Oops! Fixed.

packages/venia-ui/lib/components/LegacyMiniCart/product.js
```diff
@@ -14,6 +14,8 @@ import ProductOptions from './productOptions';
 import Section from './section';
 import { gql } from '@apollo/client';
 
+const QUANTITY_OPERATOR = 'x';
```
anthoula, line 17: Looks like this may be a special character `×` instead of the letter `x`

sirugh: Oops! You're good :D

anthoula (dismissed):

anthoula (dismissed):

anthoula (dismissed):

anthoula (approved):

dpatil-magento:
QA Approved.

PR #2792, opened 2020-10-20
Updated Intl FormattedMessage component mock.

revanth0212:
## Description

Related to #2786

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist

### DIFF
--- a/packages/peregrine/lib/Price/__tests__/__snapshots__/price.spec.js.snap
+++ b/packages/peregrine/lib/Price/__tests__/__snapshots__/price.spec.js.snap
@@ -48,7 +48,7 @@ Array [
     1
   </span>,
   <span>
-     
+     
   </span>,
   <span>
     000
--- a/packages/peregrine/lib/Price/__tests__/__snapshots__/price.spec.js.snap
+++ b/packages/peregrine/lib/Price/__tests__/__snapshots__/price.spec.js.snap
@@ -48,7 +48,7 @@ Array [
     1
   </span>,
   <span>
-     
+     
   </span>,
   <span>
     000
--- a/packages/venia-ui/__mocks__/react-intl.js
+++ b/packages/venia-ui/__mocks__/react-intl.js
@@ -8,6 +8,6 @@ const intl = reactIntl.createIntl({
 
 module.exports = {
     ...reactIntl,
-    FormattedMessage: props => <mock-formatted-message {...props} />,
+    FormattedMessage: props => <mock-FormattedMessage {...props} />,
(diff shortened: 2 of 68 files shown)


packages/venia-ui/__mocks__/react-intl.js
```diff
@@ -7,6 +8,8 @@ const intl = reactIntl.createIntl({
 
 module.exports = {
     ...reactIntl,
-    FormattedMessage: jest.fn(({ defaultMessage, id }) => defaultMessage || id),
+    FormattedMessage: props => (
+        <div componentName="Formatted Message Component" {...props} />
```
tjwiebell, line 12: Jest suggests mock components using this convention, which I think I'd prefer so we're not adding semantic markup.

https://jestjs.io/docs/en/tutorial-react#snapshot-testing-with-mocks-enzyme-and-react-16

```suggestion
        <mock-FormattedMessage>" {...props} />
```

packages/peregrine/lib/Price/__tests__/__snapshots__/price.spec.js.snap
```diff
@@ -48,7 +48,7 @@ Array [
     1
   </span>,
   <span>
-     
+
```
tjwiebell, line 51: Why do I feel like I've seen this several times now; any idea why the whitespace in this snap keeps changing?

revanth0212: It keeps sneaking in every time I update the mocks. Thanks for catching it, Tommy. I'll have to research and fix this once and for all. Even @jcalcaben has this issue.

tjwiebell (changes requested):
I support this change, let's figure out how to get into the queue; get rid of the `div` usage and this looks ready.

tjwiebell (approved):
Love it 👍  let's chat about it during team time and get this merged.

packages/venia-ui/__mocks__/react-intl.js
```diff
@@ -7,6 +8,6 @@ const intl = reactIntl.createIntl({
 
 module.exports = {
     ...reactIntl,
-    FormattedMessage: jest.fn(({ defaultMessage, id }) => defaultMessage || id),
+    FormattedMessage: props => <mock-FormattedMessage {...props} />,
```
sirugh, line 11: I really like this pattern for mocking things! Going to start using it.

PR #2794, opened 2020-10-20
Added shallowMerge util.

revanth0212:
## Description

Create a shallow merge utility that can be used to merge classes in UI components and merge operations in talons.

## Related Issue
Closes [PWA-1042](https://jira.corp.magento.com/browse/PWA-1042)

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
This PR makes a code change that does not affect the current functionality in any way. A simple sanity/smoke test should be good.

## Checklist


packages/peregrine/lib/util/shallowMerge.js
```diff
@@ -0,0 +1,3 @@
+const shallowMerge = (...args) => Object.assign({}, ...args);
```
tjwiebell, line 1: My only criticism is that this broke intellisense being able to know the shape of `DEFAULT_OPERATIONS` and providing type hints. That was a huge improvement in this move to peregrine, and I'm hoping it can be retained. Any thoughts on a jsdoc return type that might allow this to work again?

tjwiebell (approved):
This change looks fine if this is what the rest of the team wants, but I have to go on the record that I hate it, because it broke type hinting. If you can think of a way to keep operations hinting working, I'll double approve, but I currently don't think ease of merging is better than operations being self-documenting.

PR #2800, opened 2020-10-21
chore: add translation checkbox line for pr template

sirugh:
Going forward we want to ensure no raw/non-translated strings are used in the app on accident. #2789 will start failing CI if there are raw strings found, but @anthoula had a good idea which was to add a line to the PR template as an additional check.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -22,7 +22,7 @@ TODO: Describe your changes in detail here.
 <!--- Using the above wording causes Github to automatically close the issue on merge. -->
 Closes #ISSUE_NUMBER.
 
-## Acceptance 
+## Acceptance
 <!-- The people and processes this pull request needs before it is merged. -->
 <!-- These fields are not required when opening the pull request, but they -->
 <!-- should be populated after code review. -->
@@ -44,4 +44,5 @@ Closes #ISSUE_NUMBER.
 ## Checklist
 <!--- Go over all the following points, and make sure you've done anything necessary -->
 - [ ] I have added tests to cover my changes, if necessary.
+- [ ] I have added translations for new strings, if necessary.
 * I have updated the documentation accordingly, if necessary.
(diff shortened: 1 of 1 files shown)


anthoula (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -44,4 +44,5 @@ Closes #ISSUE_NUMBER.
 ## Checklist
 <!--- Go over all the following points, and make sure you've done anything necessary -->
 - [ ] I have added tests to cover my changes, if necessary.
+- [ ] I have added translations for new strings, if necessary.
 * I have updated the documentation accordingly, if necessary.
```
davemacaulay, line 48: Shall we update this bullet point to be consistent while we're here?

revanth0212: Was gonna ask the same 🤣

revanth0212, line 48:
```suggestion
- [ ] I have updated the documentation accordingly, if necessary.
```

sirugh: Was there a reason it was a bullet vs checkbox? I was hoping someone would remember :P

sirugh: Going the opposite direction, I think?

anthoula: I would advocate to use checkboxes for all items under that category. It would require action to take to ensure items are performed.

sirugh: @anthoula the team intentionally changed to use bullets in the past, but then swapped back only the test line checkbox as a way to ensure folks paid attention to it. That said, you'll find that most people ignore it, which means that we end up with github telling us that folks have `0 out of 1` checkboxes checked, as in the image I linked above.

My preference would be to not use checkboxes unless we're going to start holding folks to actually checking them instead of ignoring them. I will leave it to the team.

revanth0212: I personally want them to address the checkboxes because they seem to be important. But how to handle cases where a PR has nothing to do with Docs for instance? Honestly, I don't remember if we had this conversation before about checkboxes vs bullet points.

revanth0212 (approved):

PR #2819, opened 2020-10-23
[doc] env variable tutorial

jcalcaben:
## Description

Creates a new tutorial on how to use environment variables in front end code.

## Related Issue

Closes PWA-798

## Acceptance

Any developer

### Verification Stakeholders

Any developer

### Verification Steps

1. Navigate to the `pwa-devdocs` directory: `cd pwa-devdocs`
1. Start the html preview server: `yarn develop`
1. Navigate to the tutorials section
1. Verify **Add a new environment variable** is listed on the left nav
1. Verify links leads to expected page
1. Verify tutorial is accurate

## Checklist


pwa-devdocs/src/tutorials/intercept-a-target/add-new-environment-variable/index.md
```diff
@@ -0,0 +1,259 @@
...
+import React from "react";
+
+const PlaceholderImage = (props) => {
+  const { width = 300, height = 300 } = props;
+
+  const urlTemplate = process.env.IMAGE_PLACEHOLDER_SERVICE_URL;
+
+  if (!urlTemplate) {
+    console.error("Image placeholder service not defined!");
+    return undefined;
```
tjwiebell, line 65: React prefers `null` if you're not going to render anything, causes this error: `Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.`
```suggestion
    return null;
```

pwa-devdocs/src/tutorials/intercept-a-target/add-new-environment-variable/index.md
```diff
@@ -0,0 +1,259 @@
...
+mkdir -p src/components/PlaceholderImageDemo && touch src/component/PlaceholderImageDemo/placeholderImageDemo.js
+```
+
+Inside `placeholderImageDemo.js`, add the following content:
+
+```jsx
+import React from 'react'
+import PlaceholderImage from 'PlaceholderImage'
+
+const PlaceholderImageDemo = props => {
```
tjwiebell, line 207: Alright, I have to ask. The code style here matches the rules in the monorepo, but the examples above don't appear to have been run through the same ruleset. Not sure if you bounce back and forth depending if it's JSX or vanilla, just curious what the expectation is for consistency.

pwa-devdocs/src/tutorials/intercept-a-target/add-new-environment-variable/index.md
```diff
@@ -0,0 +1,259 @@
...
+
+Now, when you start your project, you can navigate to `/placeholder-image-demo` and see the PlaceholderImage component in action.
+
+Click to see the full solution for this tutorial:
+
+[![Edit environment-variable-tutorial](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/environment-variable-tutorial-9z0rb?fontsize=14&hidenavigation=1&module=%2Fextensions%2FPlaceholderImage%2Fsrc%2Fcomponents%2FPlaceholderImage%2FplaceholderImage.js&theme=dark)
+
+### A note about CodeSandbox
+
+CodeSandbox uses an iframe to display a preview of the site.
```
tjwiebell, line 250: I think it would be reasonable to remove the `@magento/upward-security-headers` package in your sandbox template, and you can get rid of this section.

tjwiebell (changes requested):
Content looks good, really cool example. Just a couple suggestions, and a question about code style.

pwa-devdocs/src/tutorials/intercept-a-target/add-new-environment-variable/index.md
jcalcaben: I didn't really consider consistency or code style when I created this example. What needs to be changed to achieve consistency?

tjwiebell: Quotes and the types of things eslint removes that are unneeded jump out. eg. `const PlaceholderImageDemo = (props) => {`, parens aren't needed there. I have no preference if all code samples follow the monorepo rules, but do think they should all be formatted the same, whatever style that may be.

tjwiebell (dismissed):

pwa-devdocs/src/peregrine/reference/targets/wrappable-talons/index.md
```diff
@@ -62,7 +60,5 @@ The reference doc content is generated automatically from the source code.
 To update this section, update the doc blocks in the source code
 -->
 
-{% include auto-generated/peregrine/lib/targets/TalonWrapperConfig.md %}
```
jcalcaben, line 65: This file no longer exists since the recent extensibility update. So I need to remove this line so the site will build properly. I could make this change in a separate PR, but it's very minor.

tjwiebell (approved):
Good catch 👍

pwa-devdocs/src/tutorials/intercept-a-target/add-new-environment-variable/index.md
jcalcaben: got it. I probably copied from CodeSandbox and a local copy which would explain some of the inconsistencies. CodeSandbox formats every time I save, and I never ran any sort of linter on my local files.

pwa-devdocs/src/tutorials/intercept-a-target/add-new-environment-variable/index.md
jcalcaben: that worked. thanks!

PR #2820, opened 2020-10-23
Modified SW to ignore Admin panel routes.

revanth0212:
## Description

Ignoring `admin` panel routes in the SW. This new rule utilizes the `NetworkOnly` strategy which makes sure that the SW directly relays the call to the network.

https://developers.google.com/web/tools/workbox/modules/workbox-strategies#network_only

This code works theoretically because it is looking for the `/admin` path in the URL but I couldn't test because the backend instance I use has a different domain all together. @dpatil-magento make sure you test this on an instance that has the same domain for the app and the admin panel.

## Related Issue
Closes [PWA-462](https://jira.corp.magento.com/browse/PWA-462)

### Verification Stakeholders
@dpatil-magento 
@jimbo 

### Specification
Since we added this check in the SW, it will ignore all the `/admin/` routes. This can be a problem if a store has a route in its app that has `/admin/` in its pathname. Store developers have to be vary of this.

### Verification Steps
1. Go to an admin page and see that it works.
2. Another way to verify is to check the cache storage. You should not find any assets related to the admin panel.

## Checklist

### DIFF
--- a/packages/venia-concept/src/ServiceWorker/registerRoutes.js
+++ b/packages/venia-concept/src/ServiceWorker/registerRoutes.js
@@ -1,7 +1,11 @@
 import { cacheNames } from 'workbox-core';
 import { ExpirationPlugin } from 'workbox-expiration';
 import { registerRoute } from 'workbox-routing';
-import { CacheFirst, StaleWhileRevalidate } from 'workbox-strategies';
+import {
+    CacheFirst,
+    NetworkOnly,
+    StaleWhileRevalidate
+} from 'workbox-strategies';
 import {
     isResizedCatalogImage,
     findSameOrLargerImage,
@@ -18,6 +22,25 @@ import { THIRTY_DAYS, MAX_NUM_OF_IMAGES_TO_CACHE } from './defaults';
  */
 export default function() {
     const catalogCacheHandler = createCatalogCacheHandler();
+
+    /**
+     * Never cache admin URL assets. Always go to the network.
+     *
+     * Will match URLs like:
+     * 1. venia.com/admin
+     * 2. venia.com/admin/admin
+     * 3. venia.com/backend/admin
+     * 4. venia.com/admin/store/settings
+     *
+     * Will not match URLs like:
+     * 1. venia.com/store_admin
+     * 2. venia.com/store/store_admin/settings
+     */
+    registerRoute(
+        ({ url }) =>
+            url.pathname.endsWith('/admin') || url.pathname.includes('/admin/'),
+        new NetworkOnly()
+    );
 
     registerRoute(
         new RegExp('(robots.txt|favicon.ico|manifest.json)'),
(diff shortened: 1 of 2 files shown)


chris-brabender:
Just a note, 99.9% of the time the /admin route is not used (and definitely should not be used). https://devdocs.magento.com/guides/v2.4/config-guide/secy/secy.html#use-of-non-default-magento-admin-url

You may want to move that to a .env config var per install/environment? As long as that variable is never exposed publicly, you don't want to tell anyone what your admin path or let them discover it somehow.

packages/venia-concept/src/ServiceWorker/registerRoutes.js
```diff
@@ -19,6 +23,25 @@ import { THIRTY_DAYS, MAX_NUM_OF_IMAGES_TO_CACHE } from './defaults';
...
+     * 3. venia.com/backend/admin
+     * 4. venia.com/admin/store/settings
+     *
+     * Will not match URLs like:
+     * 1. venia.com/store_admin
+     * 2. venia.com/store/store_admin/settings
+     */
+    registerRoute(
+        ({ url }) =>
+            url.pathname.endsWith('/admin') || url.pathname.includes('/admin/'),
```
davemacaulay, line 41: The admin URL for Magento is configurable and typically randomly generated or changed due to security issues.

We also would want to avoid exposing the admin path within client side code, which may make this fix harder.

davemacaulay (changes requested):

packages/venia-concept/src/ServiceWorker/registerRoutes.js
revanth0212: I was under the impression that the admin URL will have `/admin` all the time but can be masked around with other paths for instance `/******/admin` or `store_backend/admin` to name a couple of examples.

@chris-brabender the problem with `.env` or other configs is, they might be hidden from the public but at the end of the day the SW has to reference those values and it will be doing it publicly. There is no way we can keep that value hidden if we choose to use it.

revanth0212: Can an admin panel URL not have `/admin` in it?

fooman: Yes https://example.com/custom323/sales/order/ could be an admin route. Does the SW have access to cookies? The backend will set an "admin" cookie that has as its path the custom admin url
image

If that works I guess the issue would only present itself if someone went to the frontend first. A workaround for this if we don't want to solve this without changes to the graphql url resolver could be to add a special query to force the SW to go network only. 

Would you mean admin users need to use an adopted url (https://example.com/custom323/?sw_network_only=true
) to get to the login but that may be easier than explaining how to unregister the SW.

revanth0212:
Not needed any more.

PR #2843, opened 2020-10-28
[spike] Add focus trapping with react-aria

jimbo:
## Description

Use Adobe's `react-aria` library to implement focus trapping in dialogs. The issue calls for the `useDialog` hook, but this spike currently uses the `FocusScope` component instead.

#### New libraries

- `react-aria`

Module | Size (gzip) | Size (parse)
:---|---:|---:
`SSRProvider` | `0.35 kB` | `0.62 kB`
`FocusScope` | `5.38 kB` | `19.05 kB`

## Related Issue
Closes PWA-619.

### Verification Stakeholders

- @schensley 
- @supernova-at 

### Verification Steps

1. Go to the checkout page.
2. Hit `Tab` a dozen times to focus different elements. Focus should cycle between children of the dialog.

Only valid focusable elements should receive focus.

## Checklist

### DIFF
--- a/packages/venia-ui/__mocks__/react-aria.js
+++ b/packages/venia-ui/__mocks__/react-aria.js
@@ -1,9 +1,9 @@
-import { Fragment } from "react"
+import { Fragment } from 'react';
 
-const reactAria = jest.requireActual("react-aria")
-const FocusScope = Fragment
+const reactAria = jest.requireActual('react-aria');
+const FocusScope = Fragment;
 
 module.exports = {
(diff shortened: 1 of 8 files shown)


sirugh:
I noticed that after selecting a shipping method and then clicking edit the focus would not always go to the dialog, but sometimes to the next thing ie the credit card checkbox next to braintree. Curious why that occurs.

sirugh:
Closing - Referenced internally in PWA-174

PR #2848, opened 2020-11-04
spike: Example of custom button through intercept

sirugh:
Just a spike to try out one approach to themes through the scaffolded intercept file.

The intercept file replaces all instances of `Button` with the `CustomButton` component which is identical except for a dynamic load of the css used. When `store_code` is `fr`, it loads `frenchButton.css` and uses those classes instead of the default.

Try it out by switching the store to "French Store View".

[Image from Gyazo](https://gyazo.com/9b2bcdc88e62ce505751c82057a2b987)


packages/venia-concept/src/CustomButton/CustomButton.js
```diff
@@ -0,0 +1,73 @@
...
+const CustomButton = props => {
+    const {
+        children,
+        classes: propClasses,
+        priority,
+        type,
+        negative,
+        disabled,
+        ...restProps
+    } = props;
```
sirugh, line 22: Still need to figure out why these props aren't being passed. I imagine it is something to do with the `replaceJSX` call not passing props. I wasn't sure how to actually forward them with that API.

packages/venia-concept/local-intercept.js
```diff
@@ -16,7 +16,24 @@
...
+    const button = myTheme.reactComponent(
+        '@magento/venia-ui/lib/components/Button/button.js'
+    );
+
+    const CustomButton = button.addImport(
+        "CustomButton from '@magento/venia-concept/src/CustomButton/CustomButton.js'"
+    );
+    button.replaceJSX(
+        'button',
+        `<${CustomButton} className={rootClassName} type={type} disabled={disabled} {...restProps}>{children}</${CustomButton}>`
```
sirugh, line 33: Preferably we would just overwrite the export of the `Button` component with the `CustomButton` rather than replacing the JSX which requires knowledge of the props/children structure.

packages/venia-concept/src/CustomButton/CustomButton.js
```diff
@@ -0,0 +1,74 @@
...
+        className,
+        ...restProps
+    } = props;
+
+    const storeCode = storage.getItem('store_view_code');
+    const [css, setCss] = useState(defaultClasses);
+    useEffect(() => {
+        if (storeCode === 'fr') {
+            loadCss();
+        }
```
davemacaulay, line 30: I'm not sure how I feel about components having an understanding of the store view code. I wonder if there'd be some mechanism of doing this at a framework level so we don't end up with store view codes in components. Just food for thought :)

sirugh: How would you know which css to load for which store code? Something will have to know, right? Also, this "CustomButton" is a custom component from the client, not built into PWA Studio. You can only do such extension through scaffolding so it's in venia-concept as an example.

davemacaulay: It really depends what we refer to when we say theming, this could mean changing one or two components, changing some CSS, or having a completely different tree of components with inheritance. 

Would be good to validate legitimate business cases of theme customization and see where they align.

packages/venia-concept/src/ThemedCss/frenchButton.css
```diff
@@ -0,0 +1,76 @@
+.root {
+    color: red;
+    border-color: red;
+}
```
sirugh, line 4: My intent was just for this to be the css that was needed. Should be possible to provide _just_ those overrides without having to know all the rest of the css classes.

packages/venia-concept/src/CustomButton/CustomButton.js
```diff
@@ -0,0 +1,66 @@
...
+        useThemedCss('Button', defaultClasses),
+        propClasses
+    );
+    const rootClassName = classes[getRootClassName(priority, negative)];
+
+    return (
+        <button
+            // The only other custom thing here...But there might be a better
+            // way to ensure that incoming classes are used.
+            className={`${className} ${rootClassName}`}
```
sirugh, line 35: I think it would be preferable to just modify the `Button` component itself with this and the changes from line 25. No copy of `Button` would be necessary, just need to import the hook and pass it the component name + default classes.

packages/venia-concept/src/CustomButton/CustomButton.js
sirugh: Yea, the point of this spike PR was just to explore dynamic CSS loading. My take is that customization and theming are going to continue to fall into the realm of our developer users. The extension framework allows for so much customization and granularity that it seems counter productive to implement some rigid structure around what _we_ consider theming.

Anyways, I agree that we need to understand the business cases before implementing anything concrete. I've written up my thoughts in the JIRA ticket.

sirugh:
Closing this as I have explored the idea enough.

PR #2850, opened 2020-11-06
Work around order details issue

jimbo:
## Description

Users have encountered an issue in production where the `useOrderConfirmationPage` talon throws Javascript errors. This happens because we render the confirmation page as soon as we have an order number, but the order number and order details come from separate queries, so the data the confirmation page needs may not have arrived yet.

Just need to wait for that data.

## Related Issue

PWA-1106

### Verification Stakeholders

- @dpatil-magento 
- @davemacaulay 
- @revanth0212 

### Verification Steps
1. Complete checkout.
2. Verify nothing's changed.

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#2885: Work around order details issue

### DIFF
--- a/packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
+++ b/packages/venia-ui/lib/components/CheckoutPage/checkoutPage.js
@@ -111,7 +111,7 @@ const CheckoutPage = props => {
               defaultMessage: 'Checkout'
           });
 
-    if (orderNumber) {
+    if (orderNumber && orderDetailsData) {
         return (
             <OrderConfirmationPage
                 data={orderDetailsData}


packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -13,12 +13,57 @@ const OrderConfirmationPage = props => {
...
+                    <div className={classes.orderNumber}>
+                        <FormattedMessage
+                            id={'checkoutPage.orderNumber'}
+                            defaultMessage={'Order Number'}
+                            values={{ orderNumber }}
+                        />
+                    </div>
+                    <div className={classes.additionalText}>
+                        <FormattedMessage
+                            id={'checkoutPage.additionalText'}
```
revanth0212, line 56: Do we need to add this to the `en_US.json`?

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -13,12 +13,57 @@ const OrderConfirmationPage = props => {
...
+                            id: 'checkoutPage.titleReceipt',
+                            defaultMessage: 'Receipt'
+                        },
+                        { name: STORE_NAME }
+                    )}
+                </Title>
+                <div className={classes.mainContainer}>
+                    <h2 className={classes.heading}>
+                        <FormattedMessage
+                            id={'checkoutPage.thankYou'}
```
revanth0212, line 43: Similarly here.

revanth0212 (dismissed):

revanth0212:
Also, you might need to add some tests to cover the changes. Other than that this is good to go.

I really wished we know what the error is really about.

dpatil-magento:
Might be test region issue given we use test credit cards but I can reproduce this when order total is greater than $2000. In this scenario also error toast is triggered (not captured in screen)

sirugh:
@magento create issue from PR

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
```diff
@@ -13,12 +13,57 @@ const OrderConfirmationPage = props => {
...
+                            defaultMessage={
+                                'You will also receive an email with the details and we will let you know when your order has shipped.'
+                            }
+                        />
+                    </div>
+                </div>
+                <div className={classes.sidebarContainer} />
+            </div>
+        );
+    }
```
supernova-at, line 66: Since there's a lot of duplication with the "full confirmation" below I'm thinking that this can get out of sync and we wouldn't know it, especially since it occurs infrequently.

If we could re-use this shell and have a `maybeFullOrder` or similar that we could interpolate that'd simplify readability and maintainability I think.

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
supernova-at: It's in there already 👍  https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/i18n/en_US.json

packages/venia-ui/lib/components/CheckoutPage/OrderConfirmationPage/orderConfirmationPage.js
supernova-at: It's in there already 👍  https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/i18n/en_US.json

supernova-at (dismissed):

sirugh (approved):

dpatil-magento:
QA Approved.

PR #2852, opened 2020-11-09
Add Adobe client data layer

jimbo:
## Description

Add the Adobe Client Data Layer as an extension, enabling product recommendations and others to do their work. As an extension, the script is optional, and it should be loaded with the [lowest priority possible](https://addyosmani.com/blog/script-priorities/). Webpack's [dynamic imports with magic comments](https://webpack.js.org/api/module-methods/#magic-comments) make this pretty straightforward.

### Verification Steps

1. Verify that `window.adobeDataLayer` eventually evaluates to an object.
2. Verify that the library isn't included in any of our bundles and doesn't delay rendering.

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#2886: Add Adobe client data layer


packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -85,6 +85,15 @@ const VeniaAdapter = props => {
         }
     }, [initialized, persistor]);
 
+    useEffect(() => {
```
deloreyj, line 88: Did you mean to import it here in addition to the extension?

packages/venia-concept/package.json
```diff
@@ -39,6 +39,7 @@
     "@magento/pwa-buildpack": "~7.0.0"
   },
   "devDependencies": {
+    "@adobe/adobe-client-data-layer": "~1.1.3",
```
deloreyj, line 42: Should this dependency be in `venia-concept-data-layer/package.json`?

packages/venia-concept/template.html
```diff
@@ -126,6 +126,7 @@
               unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
             }
         </style>
+        <script>window.adobeDataLayer = window.adobeDataLayer || []</script>
```
deloreyj, line 129: Would this line be better in `venia-concept-data-layer/index.js` to avoid adding this array unless the extension is used? Like
```js
useEffect(() => {
        window.adobeDataLayer = window.adobeDataLayer || [];
        import(/* webpackChunkName: "acdl" */
        /* webpackMode: "lazy" */
        /* webpackPrefetch: true */
        '@adobe/adobe-client-data-layer');
    }, []);
```

packages/extensions/venia-adobe-data-layer/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@magento/venia-adobe-data-layer",
+  "version": "0.0.1",
+  "publishConfig": {
+    "access": "public"
+  },
+  "description": "Provides demo translations for PWA Studio.",
```
deloreyj, line 7:
```suggestion
  "description": "Provides Adobe Client Data Layer support in PWA Studio",
```

packages/venia-concept/template.html
jimbo: That might work, but I'm not really familiar with how things are going to use this data layer object, so for now I put the property definition as early as possible—prior to the React render, even. The only really important thing is that the library itself gets deferred as long as possible and the browser is completely idle.

supernova-at: > [The BuildBus] runs the intercept files in dependency order. If your module declares another module with Targets in peerDependencies, the other module will intercept first.

If another extension relies on this object to be present, it should list `adobe-client-data-layer` as a `peerDependency` to ensure the `adobe-client-data-layer` extension intercepts first (and this array is added).

jimbo: @supernova-at @deloreyj Good points. Done.

sirugh:
@magento create issue from PR

supernova-at (dismissed):

dpatil-magento:
QA Approved.

supernova-at (dismissed):

packages/extensions/venia-adobe-data-layer/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "scripts": {
+    "clean": " "
+  },
+  "repository": "github:magento/pwa-studio",
+  "license": "(OSL-3.0 OR AFL-3.0)",
+  "dependencies": {
+    "@adobe/adobe-client-data-layer": "~1.1.3"
+  },
+  "peerDependencies": {
+    "@magento/peregrine": "~8.0.0"
```
revanth0212, line 18: Is there a reason why we have this peer dep?

packages/extensions/venia-adobe-data-layer/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "main": "./index.js",
+  "scripts": {
+    "clean": " "
+  },
+  "repository": "github:magento/pwa-studio",
+  "license": "(OSL-3.0 OR AFL-3.0)",
+  "dependencies": {
+    "@adobe/adobe-client-data-layer": "~1.1.3"
+  },
+  "peerDependencies": {
```
revanth0212, line 17: I saw you were importing `react in the `index.js` file. Shouldn't you add that as a peer dep?

packages/extensions/venia-adobe-data-layer/intercept.js
```diff
@@ -0,0 +1,7 @@
+module.exports = targets => {
+    const { talons } = targets.of('@magento/peregrine');
```
revanth0212, line 2: I am assuming this is the reason why you have peregrine as a peer dep, but here were are searching for peregrine targets not importing them. Is that accurate?

revanth0212 (changes requested):
A couple of questions regarding imports.

packages/extensions/venia-adobe-data-layer/package.json
jimbo: Good catch. Will add it.

packages/extensions/venia-adobe-data-layer/intercept.js
jimbo: I assume it's good practice to list anything you expect to be present as a peer dependency. This package won't work if `@magento/peregrine` isn't present, but it also doesn't want to be the one to import it. Is that not a peer dependency?

Open to changing it if my assumptions aren't correct here.

packages/extensions/venia-adobe-data-layer/package.json
jimbo, line 18:
```suggestion
    "@magento/peregrine": "~8.0.0",
    "react": "~16.9.0"
```

packages/extensions/venia-adobe-data-layer/intercept.js
revanth0212: You are right about the premise. This extension is not useful if peregrine isn't part of the main package.

But I am not sure if adding it as part of peer dep, for that reason, is justifiable. On top of it, versioning is involved, and if the base app uses a diff version of peregrine this package might complain about a missing version in the console causing more doubts.

dpatil-magento:
QA re-approved.

revanth0212 (dismissed):

revanth0212 (approved):

PR #2853, opened 2020-11-09
Added feature to pick from multiple sample backends.

revanth0212:
## Description

This PR introduces a new extension and a new feature that allows developers to choose which Magento Backend they want to use while developing. 

While creating a new app, developers will be prompted to choose a Magento backend from the sample backends provided by the PWA Team or enter their own backend URL. The sample backends are fetched at runtime from a secure location hosted and maintained by the PWA Team.

This PR also introduces a new Target that accepts env validator functions. These functions will receive the project env in the JSON format and they can run validations and report errors to the buildpack's build process. The extension added in the PR uses this target to add a validator that checks the backend URL and provides options if needed.

## Related Issue
Closes [PWA-96](https://jira.corp.magento.com/browse/PWA-96)

### Verification Stakeholders
@zetlen 
@jimbo 
@dpatil-magento 

### Specification
We need to update the docs about this change. @jcalcaben can you advise where I need to update these docs.

### Verification Steps

Scaffold test
1. Checkout this branch.
2. `DEBUG_PROJECT_CREATION=1 <Path to PWA Studio Package>/packages/create-pwa/bin/create-pwa`
3. You should be able to see options from the new API.

image

`Magento 2.3.3 with Venia sample data installed` from the above image is fetched from the API.

Build process
1. Checkout this branch.
2. Change the `MAGENTO_BACKEND_URL` to an unreachable URL.
3. Run `yarn build`. The build process should fail and suggest some sample backends.

image

If the backend is fine, nothing should change from how `develop` works.

## Checklist


packages/create-pwa/lib/index.js
```diff
@@ -1,17 +1,38 @@
...
+    sampleBackends: defaultSampleBackends
 } = require('@magento/pwa-buildpack/lib/cli/create-project');
 
+const removeDuplicateBackends = backendEnvironments =>
+    uniqBy(backendEnvironments, 'url');
+
+const fetchSampleBackends = async () => {
+    try {
+        const res = await fetch(
+            'https://fvp0esmt8f.execute-api.us-east-1.amazonaws.com/default/getSampleBackends'
```
revanth0212, line 24: I am looking at AWS docs to use a DNS name instead of this raw path. It will be changed in the future (before this PR is merged).

sirugh: Is this a lambda? Does it read configuration somewhere? I'm a bit surprised we don't just maintain a list in the extension.

revanth0212: The plan is to pull these backend URLs dynamically so we don't need to publish every single time a backend changes, which we will have to if these are part of the package. The URL above is an AWS Gateway URL. The plan is to use a domain-specific URL so in the future if we move between cloud providers (Azure, Ethos, etc), we can keep using the same URL but the implementation in the back can change.

packages/create-pwa/lib/index.js
```diff
@@ -1,17 +1,38 @@
...
+        );
+        const { sampleBackends } = await res.json();
+
+        return sampleBackends.environments;
+    } catch (err) {
+        console.error(err);
```
tjwiebell, line 30: I think this failure needs to be silent outside of debug environments. This would be surfaced to someone trying to scaffold an app for the first time, and they aren't able to resolve this failure.

packages/extensions/venia-sample-backends/intercept.js
```diff
@@ -0,0 +1,80 @@
+const fetch = require('node-fetch');
+
+const isBackendActive = async env => {
+    try {
+        const magentoBackend = env.MAGENTO_BACKEND_URL;
+        const res = await fetch(magentoBackend);
+
+        return res.ok;
+    } catch (err) {
+        console.error(err);
```
tjwiebell, line 10: Similarly, I'm not sure it makes sense to log this to the user. Can we make these debug level messages too?

tjwiebell (changes requested):
Some output needs to be suppressed, but overall works as described. I also ran into this error when running locally, appears to be related if you can investigate.

```
(node:58492) UnhandledPromiseRejectionWarning: Error: scaffold4000: Cannot getTargetsOf("@magento/pwa-buildpack"): @magento/pwa-buildpack has not yet declared
```

Repro: do `DEBUG_PROJECT_CREATION` create-pwa, but I answered don't install dependencies.

packages/create-pwa/lib/index.js
```diff
@@ -1,17 +1,38 @@
...
 const os = require('os');
+const fetch = require('node-fetch');
 const changeCase = require('change-case');
 const inquirer = require('inquirer');
 const execa = require('execa');
 const chalk = require('chalk');
 const gitUserInfo = require('git-user-info');
 const isInvalidPath = require('is-invalid-path');
 const isValidNpmName = require('is-valid-npm-name');
+const { uniqBy } = require('lodash');
```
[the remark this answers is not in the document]
revanth0212: I am a fan of the method imports myself. But I realized we import the whole lodash library as part of buildpack anyway. Hence I tried to add it as a peer dependency. Also, Lodash one of the libraries which are really good for tree shaking. Webpack should be able to remove all unnecessary code.

packages/extensions/venia-sample-backends/intercept.js
revanth0212: You are right, these need to be debug messages. Nice suggestion @tjwiebell 👍 .

packages/create-pwa/lib/index.js
sirugh: > so we don't need to publish every single time a backend changes

_Something_ will have to change _somewhere_ whether that's a file in s3, a lambda script, or a local file in a pwa studio extension/package (like create-pwa).

I personally prefer keeping this file locally within `create-pwa` and just publishing a new version if the backend changes. It's much more straightforward and requires no knowledge of lambdas, aws routing, etc.

If the team want to keep this solution as is, I'd really like to see some documentation with steps to update this file.

revanth0212: The points you have mentioned are perfectly reasonable. I did have a thought of making it a package to avoid the cost of hosting and also avoid potential DoS security issues. But the biggest problem with publishing on NPM is that developers don't generally run `yarn install` every time they build or might be using an older version of the package causing older environments to be used instead. The cost of using AWS Lambda and AWS S3 is minimal and we have enabled throttling as an initial security check.

sirugh: > But the biggest problem with publishing on NPM is that developers don't generally run yarn install every time they build or might be using an older version of the package causing older environments to be used instead.

Two questions: 

1) How often does the `getSampleBackends` service actually need to be used? It should just be used when someone spins up their project for the first time, right?

2) Is the `getSampleBackends` API used during the _build_ or do an up-check to see if the backend is accessible during build?

packages/pwa-buildpack/lib/Utilities/runEnvValidators.js
```diff
@@ -0,0 +1,49 @@
...
+
+    const validationContext = { env, onFail, debug };
+
+    await bus
+        .getTargetsOf('@magento/pwa-buildpack')
+        .validateEnv.promise(validationContext);
+
+    if (errorMessages.length) {
+        debug('Found validation errors in ENV, stopping the build process');
+        throw new Error(errorMessages.join('\n'));
```
zetlen, line 41:
```suggestion
    /**
     * A friendly error with enumerable sub-messages for node-to-node usage.
     */
    const removeErrorPrefix = (msg) => msg.replace(/^Error:\s*/, "");
    const printValidationMsg = (error, index) =>
      `(${index + 1}) ${removeErrorPrefix(error.message)}`;
    const prettyErrorList = errorMessages.map(printValidationMsg).join("\n    ");
    const validationError = new Error(`Environment has ${errorMessages.length} validation errors: 
            ${prettyErrorList}
        `);
    validationError.errorMessages = errorMessages;
    throw validationError;

```

zetlen (commented):
Looks great, Revanth! This is really fine work. One question and one suggestion, no demanded changes.

packages/create-pwa/lib/index.js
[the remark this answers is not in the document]
tjwiebell: Works for me 👍  @revanth0212 `lodash` is currently in the `dependencies` list, did you intend to put it under `peerDependencies`? As long as it's not grabbing two different versions it shouldn't matter, just wanted to point it out.

packages/pwa-buildpack/lib/Utilities/__tests__/loadEnvironment.spec.js
```diff
@@ -69,7 +73,7 @@ afterAll(() => {
 
 const loadEnvironment = require('../loadEnvironment');
 
-test('throws on load if variable defs are invalid', () => {
+test.skip('throws on load if variable defs are invalid', () => {
```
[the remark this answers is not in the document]
revanth0212: I was updating each test individually. Must have forgotten to update this guy. Nice catch JZ.

packages/create-pwa/lib/index.js
revanth0212: > 1. How often does the `getSampleBackends` service actually need to be used? It should just be used when someone spins up their project for the first time, right?

Along with bootstrapping, It will be used every time someone adds the `venia-sample-backends` extension and builds their project.

> 2. Is the `getSampleBackends` API used during the _build_ or do an up-check to see if the backend is accessible during build?

It is used during the build stage if the extension is installed. It will be used to fetch the sample backends, if the current backend is inactive. We won't be calling the service if the backend mentioned in the ENV file is active.

packages/pwa-buildpack/lib/Utilities/__tests__/loadEnvironment.spec.js
```diff
@@ -87,9 +87,14 @@ test.skip('throws on load if variable defs are invalid', () => {
...
-    const fn = () => loadEnvironment('./');
 
-    expect(fn).toThrow('Bad environment variable definition');
+    try {
+        await loadEnvironment('./');
+    } catch (e) {
+        expect(e.message.replace(/\n/g, '')).toBe(
+            'Bad environment variable definition. Section inscrutable variable { "type": "ineffable"} declares an unknown type ineffable'
+        );
+    }
```
tjwiebell, line 97: This test wouldn't fail with the expect inside the catch if this method ever stopped throwing. Async rejections are tough in jest, but I think this would work (you can swap out the inline snapshot if that's not your jam).

```suggestion
await expect(loadEnvironment('./')).rejects.toThrowErrorMatchingInlineSnapshot()
```

packages/create-pwa/lib/index.js
tjwiebell: I got a 👍  on this suggestion but it hasn't been resolved. I still see errors output to the user if you could please resolve.

tjwiebell (changes requested):
Really close! Error handling of `create-pwa` still needs addressed, and a suggestion to improve a test.

packages/pwa-buildpack/lib/Utilities/__tests__/loadEnvironment.spec.js
revanth0212: Nice to know. I had no idea about `toThrowErrorMatchingInlineSnapshot`. Thanks, @tjwiebell.

revanth0212:
> I got a 👍 on this suggestion but it hasn't been resolved. I still see errors output to the user if you could please resolve.

Sorry forgot to update. Removed the `console.error`.

tjwiebell (approved):
Code-wise this is a 👍  for me, nice work! It does appear you picked up a test failure, and I'm a bit surprised the linter didn't complain about the unused `err` variable [here](https://github.com/magento/pwa-studio/pull/2853/files#diff-723d49b48a5e36cdcb5211585f738467825bd9528a320eb72ddd3e9fa3c02ae3R29). Fix those up and you can send this to QA.

revanth0212:
> Code-wise this is a 👍 for me, nice work! It does appear you picked up a test failure, and I'm a bit surprised the linter didn't complain about the unused `err` variable [here](https://github.com/magento/pwa-studio/pull/2853/files#diff-723d49b48a5e36cdcb5211585f738467825bd9528a320eb72ddd3e9fa3c02ae3R29). Fix those up and you can send this to QA.

The `catch` statement should always have an argument. If I remove it, the linter will complain.

image

tjwiebell:
> > Code-wise this is a 👍 for me, nice work! It does appear you picked up a test failure, and I'm a bit surprised the linter didn't complain about the unused `err` variable [here](https://github.com/magento/pwa-studio/pull/2853/files#diff-723d49b48a5e36cdcb5211585f738467825bd9528a320eb72ddd3e9fa3c02ae3R29). Fix those up and you can send this to QA.
> 
> The `catch` statement should always have an argument. If I remove it, the linter will complain.
> 
> image

Drop your parens and it should pass linting.

tjwiebell (approved):

revanth0212:
Finally, I got the tests to pass in AWS. All the commits you see in the last 24-30hrs are test related stuff. No functionality has been changed.

packages/create-pwa/lib/index.js
davemacaulay: What happens if this URL goes down? Will the yarn install hang for X seconds? 

Also if we're maintaining this micro server we need to setup monitoring to ensure we can be aware of uptime issues. We should also consider any security implications of having this service running.

dpatil-magento:
@revanth0212 Please check following :
1.  root .env is copied to scaffold instead considering build env vars. 
2. When backend and API are down  better to avoid this message in `yarn build` as we see null response -   `......is inactive. Please consider using one of these other backends:
[]`

packages/create-pwa/lib/index.js
revanth0212: Nice question @davemacaulay. I have used a wrong URL to simulate the service failure and you can see here, if the service is down, the script will use the internal URLs shipped as part of the `sampleBackends.json` file.

image

You are right about monitoring. I'll create a ticket to investigate the monitoring capabilities of our team/AWS.

Coming to security, we are not giving the public access to the Lamda function or the S3 object. It is an AWS Gateway Service that we are using. Also, the Lamda function that we are using only has read access to that particular object.

packages/pwa-buildpack/lib/Utilities/runEnvValidators.js
```diff
@@ -32,9 +32,21 @@ async function validateEnv(context, env) {
 
     const validationContext = { env, onFail, debug };
 
-    await bus
-        .getTargetsOf('@magento/pwa-buildpack')
-        .validateEnv.promise(validationContext);
+    try {
```
revanth0212, line 35: @zetlen I am using a try-catch to avoid `Buildpack targets not available error` while using the CLI tools. I am assuming the targets are not available because we are not going through buildpack/webpack while using the CLI tools. To avoid failures, I am wrapping this in the try-catch. Is that fine/expected?

packages/create-pwa/lib/index.js
dpatil-magento: @revanth0212 It would nice if we can move the lambda setup to aws prod account and if any Audits fail. Rest all looks good.

revanth0212:
> @revanth0212 Please check following :
> 
> 1. root .env is copied to scaffold instead considering build env vars.
> 2. When backend and API are down  better to avoid this message in `yarn build` as we see null response -   `......is inactive. Please consider using one of these other backends: []`

Fixed 1 and added a different error message in the case of 2. If backends is empty for any reason (failure of API etc), the error message will be `<MAGENTO_BACKEND> is inactive. Please consider using an active backend.`

dpatil-magento:
@revanth0212 There is prettier failure and last open question, other than these its good to go.

During `yarn build`  say when customer real backend is down (worst case), suggesting to use demo backend instance would not be good idea?
```
Error: Environment has 1 validation errors:
 (1) https://production-test.test.com/ is inactive. Please consider using one of these other backends:

 [{"name":"2.4.1-cloud-meta-package","description":"Magento 2.4.1 with Venia sample data installed","url":"https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/"},{"name":"2.4-develop","description":"Magento 2.4-develop with Venia sample data installed","url":"https://venia-develop-ccewfjy-mfwmkrjfqvbjk.us-4.magentosite.cloud/"}]
```

revanth0212:
> @revanth0212 There is prettier failure and last open question, other than these its good to go.
> 
> During `yarn build` say when customer real backend is down (worst case), suggesting to use demo backend instance would not be good idea?
> 
> ```
> Error: Environment has 1 validation errors:
>  (1) https://production-test.test.com/ is inactive. Please consider using one of these other backends:
> 
>  [{"name":"2.4.1-cloud-meta-package","description":"Magento 2.4.1 with Venia sample data installed","url":"https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/"},{"name":"2.4-develop","description":"Magento 2.4-develop with Venia sample data installed","url":"https://venia-develop-ccewfjy-mfwmkrjfqvbjk.us-4.magentosite.cloud/"}]
> ```

Good question. This extension is for developers who are trying out PWA Studio, not for production instances. It is expected that they will remove the extension from their `package.json`. Once it is removed, they won't be prompted with options in case of backend failure.

dpatil-magento:
QA Approved.

sirugh:
> It is expected that they will remove the extension from their package.json.

@revanth0212 is this documented?

revanth0212:
> > It is expected that they will remove the extension from their package.json.
> 
> @revanth0212 is this documented?

I haven't included this in the docs. @jcalcaben any idea where can this go if docs are the right option to place this info?

I say expected because the package itself is named as `sample backends` so people would know. As I type this, I can imagine someone missing to remove this from `package.json`. Ill do one thing, ill post a warning in the console that they are using a non-prod package/extension and they should consider removing it before going to prod as we do with the default Braintree token. How does that sound?

revanth0212:
I have added a console log to indicate that the package being used is a Dev only package and one has to remove it before going to Prod.

image

supernova-at (approved):

dpatil-magento:
QA re-approved.

oefterdal:
I get this error when I remove the package `venia-sample-backends` as instructed. 
image

PR #2856, opened 2020-11-10
fix: Service Worker caching for home page routes w/ store code

sirugh:
## Description

Normally, the home page route is like `http://develop.pwa-venia.com` but if store codes are configured to display in the url route it could be `http://develop.pwa-venia.com/fr/` or `http://develop.pwa-venia.com/fr`. The Service Worker cached root pathnames but not those with the code in them. This PR updates that code to use the build-time configuration to check if the current pathname could be considered a home page given the available store codes.

## Related Issue

Closes [PWA-1083](https://jira.corp.magento.com/browse/PWA-1083).

### Verification Steps

1. Run the app with `USE_STORE_CODE_IN_URL=true` and in stage mode.
2. Load /default/.
3. Go offline.
4. Refresh the page. It should render.

## Checklist


revanth0212 (dismissed):
PR looks perfect. Are there any tests that you need to update?

anthoula (approved):

PR #2859, opened 2020-11-11
Use ApolloClient for MagentoRoute

jimbo:
## Description

`MagentoRoute` and its talon execute a GQL query and fetch a RootComponent on route change. The code for doing these things is old, running queries and caching the results in local storage manually, which makes it difficult to clear cached results when admins change routes in the backend.

This PR attempts to update these calls to our standards.

## Related Issue

[PWA-821](https://jira.corp.magento.com/browse/PWA-821)

### Verification Stakeholders

- @dpatil-magento 
- @sirugh 

### Verification Steps

1. Clear site data.
2. Click links to navigate from the home page to a category, then to a product page.
3. Add a product to your cart, then click a link to the cart page.
4. Click the link in your cart items (product name) to go back to the product page.
5. Click the link in the breadcrumbs (category name) to go back to the category page.

- Navigation should be quick.
- There should be only one `ResolveURL` call in the Network tab for each navigation.
- Revisiting a page (steps 4 and 5) should be instant, as a cache hit.
- There should be no regressions when changing store code, whether in the URL or not.

## Checklist


packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -55,49 +36,69 @@ export const useMagentoRoute = props => {
...
+
+    const { data: storeCodeData } = storeCodeResult;
+    const store = storeCodeData && storeCodeData.storeConfig.code;
+    const routeKey = getRouteKey(pathname, store);
+    const routeData = responses.get(routeKey);
+
+    const urlResult = useQuery(RESOLVE_URL, {
+        fetchPolicy: 'cache-and-network',
+        nextFetchPolicy: 'cache-first',
+        variables: { url: pathname }
```
sirugh, line 52: Will this work when using store code in url? You may have a pathname like `/fr/shop-the-look.html` in this case.

Or I could be wrong as we should be setting the base properly when updating store code, so maybe it removes the prefix.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -55,49 +36,69 @@ export const useMagentoRoute = props => {
...
+        if (!isMountedRef.current || loading || !data || !store) return;
+
+        // the result may be null or incomplete, amounting to a 404
+        const { urlResolver } = data;
+        const { id, redirectCode, relative_url, type } = urlResolver || {};
+        const isNotFound = !urlResolver || !id || !type || id === -1;
+        const isRedirect = REDIRECT_CODES.has(redirectCode);
+
+        getRouteComponent(type, id)
+            .then(component => {
```
sirugh, line 69: Might be easier to read if you convert to `await`.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -55,49 +36,69 @@ export const useMagentoRoute = props => {
...
+                // throw Apollo errors to the catch block
+                if (error) throw error;
```
sirugh, line 71: So we always want to call `getRouteComponent` even if urlResolver returns an error or `null`?

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
jimbo: Yes; it should, since the store code is a header and we apply with Apollo Link.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
jimbo: Refactored to avoid extra calls. Seems a lot cleaner, and visibly faster.

packages/peregrine/lib/talons/MagentoRoute/helpers.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const RESPONSES = {
+    ERROR: routeError => ({ hasError: true, routeError }),
+    FOUND: component => component,
+    LOADING: { isLoading: true },
+    NOT_FOUND: { isNotFound: true },
+    REDIRECT: relativeUrl => ({ isRedirect: true, relativeUrl })
+};
+
+export const getRouteKey = (pathname, store) => `[${store}]--${pathname}`;
```
sirugh, line 15: Is prepending the store view code just to provide protection in the event the page is not reloaded/refreshed when you switch store views? In practice I found that we always refresh the page so there was never an instance where the local state map would contain values from `default` _and_ `fr`, for example.

My point is essentially that we might be able to do away with the store query in the talon and just use the path as the route key, which would save us a network call.

packages/peregrine/lib/talons/MagentoRoute/helpers.js
```diff
@@ -0,0 +1,20 @@
...
+
+const warning = () => new Error('fetchRootComponent is not defined');
+const { fetchRootComponent = warning } = window || {};
+export const getRootComponent =
+    fetchRootComponent.default || fetchRootComponent;
```
supernova-at, line 20: Mind adding a comment above L17 indicating how / when `fetchRootComponent` gets added to `window`?

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -55,49 +26,80 @@ export const useMagentoRoute = props => {
...
 
-    // If the entry for this pathname is a redirect, perform the redirect.
-    useEffect(() => {
-        if (routeData && routeData.isRedirect) {
-            history.replace(routeData.relativeUrl);
-        }
-    }, [componentMap, history, pathname, routeData]);
+    const setComponent = useCallback(
+        (key, value) => {
+            if (isMountedRef.current) {
```
supernova-at, line 31: I'm wondering if we should queue these if not mounted yet.

Once we mount, we can iterate over the queue to ensure we process all of these.

I don't have a particular bug in mind that this would solve but this smells like potential for a "silent error" scenario.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -55,49 +26,80 @@ export const useMagentoRoute = props => {
...
+    const { storeConfig } = storeCodeData || {};
+    const { code: store } = storeConfig || {};
+    const routeKey = getRouteKey(pathname, store);
+
+    // destructure url result
+    const { data: urlData, error: urlError } = urlResult;
+    const { urlResolver } = urlData || {};
+    const { id, redirectCode, relative_url, type } = urlResolver || {};
+
+    // calculate response
```
supernova-at, line 60: 💯  separation of concerns + comments

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -55,49 +26,80 @@ export const useMagentoRoute = props => {
...
+    const routeData =
+        component && !fetchError
+            ? RESPONSES.FOUND(component)
+            : queryError
+            ? RESPONSES.ERROR(queryError)
+            : isEmpty && !isPending
+            ? RESPONSES.NOT_FOUND
+            : isRedirect
+            ? RESPONSES.REDIRECT(relative_url)
+            : RESPONSES.LOADING;
```
sirugh, line 77: Is it just me or are these really hard to read?

supernova-at (commented):
Much needed changes and great tests!

Need some verification steps (especially around the edges) and this should be 👍

packages/peregrine/lib/talons/MagentoRoute/helpers.js
jimbo: The idea is that when the store view changes, we don't want to assume that the component we have in state is correct, since it could be different.

If you're saying that `MagentoRoute` unmounts and local state is lost when we change stores, then you're right, the logic & call for store code here would be unnecessary. I could remove them, in that case.

sirugh: > If you're saying that MagentoRoute unmounts and local state is lost when we change stores, then you're right, the logic & call for store code here would be unnecessary. I could remove them, in that case.

I am saying just that :) We refresh the page to trigger all the queries again using the new store view code or currency header. There may be a point in the future where we use Apollo's API to trigger requery (`clearStore`?) but that's a future state and we don't need to pre-optimize here, especially when it could get in the way of a render.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
jimbo: Actually, this brings up a very good point. Once we visit a client-only page that isn't rendered with `MagentoRoute`, such as `/cart` or `/checkout`, `MagentoRoute` unmounts and we lose its local state. That's why this safeguard is here, but that's also a sign that we should be saving this map elsewhere—somewhere that won't unmount.

We've kept these values in local state because they're functions, so they can't go in local storage. But we could lift them up, at which point we'd no longer lose the map when we navigate (nor need this safeguard). Maybe an Apollo reactive variable?

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -1,52 +1,23 @@
...
+import {
+    REDIRECT_CODES,
+    RESPONSES,
+    getRootComponent,
+    getRouteKey
+} from './helpers';
+import DEFAULT_OPERATIONS from './magentoRoute.gql';
+
+export const useMagentoRoute = (props = {}) => {
+    const { operations = DEFAULT_OPERATIONS } = props;
```
supernova-at, line 14: ⚠️  New pattern: 

```js
import mergeOperations from '@magento/peregrine/lib/util/shallowMerge';

const operations = mergeOperations(DEFAULT_OPERATIONS, props.operations);
```

sirugh: Oh right

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
brendanfalkowski: I always do a double-take on tenaries that aren't simply two values. We usually rewrite them as overblown if/elseif/else trees, so it's super obvious in spite of syntax characters making it "cluttered".

packages/peregrine/lib/talons/MagentoRoute/helpers.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const RESPONSES = {
+    ERROR: routeError => ({ hasError: true, routeError }),
+    FOUND: component => component,
+    LOADING: { isLoading: true },
+    NOT_FOUND: { isNotFound: true },
+    REDIRECT: relativeUrl => ({ isRedirect: true, relativeUrl })
+};
```
sirugh, line 18: I know you're just moving this code but I really don't like this map. My type senses are tingling.

jimbo: Okay, they're inline now.

supernova-at: Lol yeah I wrote it and I don't like it either. All of these things used to be in that other giant chained ternary though so at least that's less of a mess.

Strikes me as a possible use case for `useReducer` since we keep track of `hasError`, `isLoading`, `isNotFound`, `isRedirect`.

jimbo: I think it's fine, honestly. We don't need to keep the response type in state; the return object is just derived from the data we get back from the two requests. This should be solid. 👍

supernova-at (commented):
Code looks great, simple navigation around the deployed site seems to work as expected.

Approval pending verification steps, including around store code.

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
jimbo: Confirmed. This works because we set the `basename` on the router, so `pathname` would be `/shop-the-look.html` in that case.

PDP

jimbo:
> Code looks great, simple navigation around the deployed site seems to work as expected.
> 
> Approval pending verification steps, including around store code (discussion

Cool, I've added these.

supernova-at (approved):
Verification steps ✅ .

@jimbo there are some unit test failures though.

dpatil-magento:
QA Approved.

packages/venia-ui/lib/RootComponents/Category/__tests__/__snapshots__/category.spec.js.snap
```diff
@@ -1,76 +1,40 @@
...
-    </svg>
-  </span>
-  <span>
-    <mock-FormattedMessage
-      defaultMessage="Fetching Data..."
-      id="loadingIndicator.message"
-    />
-  </span>
-</div>
+Array [
```
sirugh, line 4: I'm not sure about this test - it says "renders a loading indicator when appropriate" but I'm not sure when appropriate _is_. The snap before seems to have the indicator whereas the snap after does not.

jimbo: Clarified by adding another test and updating the wording. The snapshot in question had changed because we're no longer rendering a full-page indicator when there's cached data.

sirugh: Ok that makes sense. Thanks for the update :)

sirugh (approved):

PR #2867, opened 2020-11-18
[bug]: Pagebuilder links should use store code

sirugh:
## Description

When `USE_STORE_CODE_IN_URL=true` we update the basename property so links created within venia-ui contain the proper store code.

Pagebuilder needs a similar treatment, as it just injects raw html into the content types.

## Related Issue

Closes [PWA-1063](https://jira.corp.magento.com/browse/PWA-1063).

### Verification Steps

1. Go to PR deploy page. Check links within PageBuilder content ("shop now button", etc).
2. Links should contain store code prefix ie `/default/shop-the-look.html`.

## Checklist

### DIFF
--- a/packages/pagebuilder/lib/utils.js
+++ b/packages/pagebuilder/lib/utils.js
@@ -177,6 +177,13 @@ export function getIsHidden(node) {
     };
 }
 
+/**
+ * Modifies the node and children to include the current store view code in the
+ * link.
+ *
+ * @param node
+ * @returns node
+ */
 export function injectStoreCodeHref(node) {
     if (process.env.USE_STORE_CODE_IN_URL === 'true') {
         const storeViewCode =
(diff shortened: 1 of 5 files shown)


packages/pagebuilder/lib/utils.js
```diff
@@ -173,3 +176,25 @@ export function getIsHidden(node) {
...
+ */
+export function injectStoreCodeHref(node) {
+    if (process.env.USE_STORE_CODE_IN_URL === 'true') {
+        const storeViewCode =
+            storage.getItem('store_view_code') || STORE_VIEW_CODE;
+
+        // For each link update the href to include the active store code.
+        node.querySelectorAll('a').forEach(oldNode => {
+            const newNode = oldNode.cloneNode(true);
+            newNode.href = `/${storeViewCode}${oldNode.getAttribute('href')}`;
```
sirugh, line 195: @davemacaulay pointed out that the `href` may be relative or absolute so we need to make this a bit more resilient to the possibility of either.

sirugh:
Closing - deemed unnecessary as PB generated links should already have store code in header if "Use Store Code In URL = Yes" in the admin. The links that won't have store code, even in that case, will be raw links entered by a user. These could be relative paths, but could also exist on a sub-page, not the home route, which means we cannot know how or when to inject the store code in those cases. Better to just leave this to the backend.

PR #2868, opened 2020-11-19
Move product detail GraphQl fields to fragment

huykon:
## Description

Base on [#1027](https://github.com/magento/graphql-ce/issues/1027) Use ProductInterface fragment for product detail query.

### Verification Steps

1. Go to the product detail page.
2. Verify the product data shows up.
3. Make sure everything working well.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/RootComponents/Product/product.gql.js
+++ b/packages/venia-ui/lib/RootComponents/Product/product.gql.js
@@ -1,92 +1,15 @@
 import { gql } from '@apollo/client';
 
+import { ProductDetailsFragment } from './productDetailFragment.gql';
+
 export const GET_PRODUCT_DETAIL_QUERY = gql`
     query getProductDetailForProductPage($urlKey: String!) {
         products(filter: { url_key: { eq: $urlKey } }) {
             items {
-                # Once graphql-ce/1027 is resolved, use a ProductDetails fragment
-                # here instead.
-                __typename
-                categories {
-                    id
-                    breadcrumbs {
-                        category_id
-                    }
-                }
-                description {
-                    html
-                }
                 id
-                media_gallery_entries {
-                    id
-                    label
-                    position
-                    disabled
-                    file
-                }
-                meta_description
-                name
-                price {
-                    regularPrice {
-                        amount {
-                            currency
-                            value
-                        }
-                    }
-                }
-                sku
-                small_image {
-                    url
-                }
-                url_key
-                ... on ConfigurableProduct {
-                    configurable_options {
-                        attribute_code
-                        attribute_id
-                        id
-                        label
-                        values {
-                            default_label
-                            label
-                            store_label
-                            use_default_value
-                            value_index
-                            swatch_data {
-                                ... on ImageSwatchData {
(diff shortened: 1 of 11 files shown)


packages/venia-ui/lib/RootComponents/Product/product.gql.js
```diff
@@ -1,92 +1,15 @@
...
 
+import { ProductDetailsFragment } from './productDetailFragment.gql';
+
 export const GET_PRODUCT_DETAIL_QUERY = gql`
     query getProductDetailForProductPage($urlKey: String!) {
         products(filter: { url_key: { eq: $urlKey } }) {
             items {
-                # Once graphql-ce/1027 is resolved, use a ProductDetails fragment
-                # here instead.
-                __typename
```
sirugh, line 9: I believe `__typename` got lost in the copy to the fragment. Can you add it back?

packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/productForm.js
```diff
@@ -120,49 +121,11 @@ export const GET_CONFIGURABLE_OPTIONS = gql`
...
-                                        currency
-                                        value
-                                    }
-                                }
-                            }
-                            sku
-                        }
-                    }
-                }
+                ...ProductDetailsFragment
```
sirugh, line 124: Unfortunately you cannot just re-use the fragment here and the content needed by the `RootComponent/Product` and the content needed by this component are different. Re-using the fragment will result in over-fetching of data. 

Please revert this change, or create a `productFormFragment`.

packages/venia-ui/lib/components/LegacyMiniCart/editItem.js
```diff
@@ -55,51 +56,10 @@ export const PRODUCT_DETAILS = gql`
...
-                                file
-                                label
-                                position
-                            }
-                            sku
-                            stock_status
-                        }
-                    }
-                }
+                ...ProductDetailsFragment
```
sirugh, line 60: Same comment here. Unfortunately you cannot just re-use the fragment here and the content needed by the `RootComponent/Product` and the content needed by this component are different. Re-using the fragment will result in over-fetching of data. 

Please revert this change, or create an `editFormFragment`.

sirugh (changes requested):
I checked the fragment use and it does seem like graphql-ce/1027 was resolved and the fragment returns the correct data. I would still want to do a regression on the PDP.

While you're here making changes to the GQL files, would you mind moving the Product RootComponent query and fragment to peregrine? You can import and use like the [`useBreadcrumbs` talon](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js#L37).

sirugh (approved):
I think I see why you didn't move the other fragments to talons as those components define their gql in the same file. I think that is fine for now, so please don't feel like it's necessary to do any further.

Thanks for this!

huykon:
Thank you! I hope this done :)

dpatil-magento:
QA Approved.

PR #2879, opened 2020-12-01
Address Book - Add & Edit

supernova-at:
## Description

This PR adds the ability to add and edit addresses in the address book.

## Related Issue

Closes PWA-634.

### Verification Stakeholders

@schensley @dpatil-magento 

### Specification

n/a

### Verification Steps

1. Sign in
1. Go to My Account > Address Book

#### Add New Address

1. Click on "Add an Address" button 
1. Complete the form
1. Click "Save"
1. Verify that the newly created address appears on the page

##### Default

1. Click on "Add an Address" button 
1. Complete the form
1. Check the "make this my default address" checkbox
1. Click "Save"
1. Verify that the newly created address appears on the page and has the `default` icon on it

#### Edit Existing Address

1. Click on "Edit" button of an existing address
1. Complete the form
1. Click "Save"
1. Verify that the address is correctly updated on the page

## Checklist

- [x] Make this my default works
- [x] Errors need to be cleared out between showings of the dialog
- [x] Check and disable "make default" when the first address is being added
- [x] Check and disable "make default" when the only address is being edited


packages/peregrine/lib/talons/AddressBookPage/addressBookPage.gql.js
```diff
@@ -22,14 +23,54 @@ export const GET_CUSTOMER_ADDRESSES = gql`
...
+            firstname
+            lastname
+            postcode
+            region {
+                region
+                region_code
+                region_id
+            }
+            street
+            telephone
```
supernova-at, line 55: We may be able to remove these fields from the query result since we do a `refetchQueries` to get the updated data.

packages/venia-ui/lib/components/AddressBookPage/__tests__/addEditDialog.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        isEditMode: true,
+        activeEditAddress: {
+            region: {
+                region: 'Arizona',
+                region_code: 'AZ'
+            },
+            country_code: 'US',
+            street: ['123 Main Street'],
+            telephone: '7777777777',
+            postcode: '77777',
```
jimbo, line 60: This sure doesn't look like an Arizona zipcode to me. 😉

jimbo (dismissed):
Works pretty well for me.

packages/peregrine/lib/talons/AddressBookPage/useAddressBookPage.js
```diff
@@ -56,14 +90,101 @@ export const useAddressBookPage = (props = {}) => {
...
-        alert('TODO!');
+        // Used to seed the form when adding a new address.
+        // country_code is needed for the Region component to work properly.
+        // default_shipping set to true when this is the first address being added.
+        const formAddress = {
+            country_code: 'US',
+            default_shipping: customerAddresses.length === 0
+        };
+        formApi.reset();
+        formApi.setValues(formAddress);
```
supernova-at, line 101: We send these as `initialValues` to the form further down in this file but unfortunately we also have to repeat this each time the dialog is opened in `Add` mode.

`initialValues` is only set once and can't be updated; we need to change `default_shipping` based on the current number of customer addresses, which might change as the user interacts with the page (adding / removing addresses).

packages/venia-ui/i18n/en_US.json
```diff
@@ -23,6 +23,9 @@
     "addressBook.headerText": "Change Shipping Information",
     "addressBookPage.addAddressText": "Add an Address",
     "addressBookPage.addressBookText": "Address Book",
+    "addressBookPage.addDialogTitle": "New Address",
+    "addressBookPage.editDialogTitle": "Edit Address",
+    "addressBookPage.makeDefaultAddress": "Make this my default address",
```
supernova-at, line 28: Do these things need to be added to the french language pack too?

packages/venia-ui/lib/components/AddressBookPage/__tests__/__snapshots__/addressBookPage.spec.js.snap
```diff
@@ -69,6 +69,9 @@ exports[`renders correctly when there are existing addresses 1`] = `
       </span>
     </button>
   </div>
+  <AddEditDialog
+    numAddresses={3}
```
supernova-at, line 73: I honestly don't know why `numAddresses` is the only prop that shows up in the snapshot 🤔

packages/venia-ui/lib/components/AddressBookPage/addEditDialog.js
```diff
@@ -0,0 +1,178 @@
...
+
+    const { formatMessage } = useIntl();
+
+    const classes = mergeClasses(defaultClasses, props.classes);
+    // Disable the checkbox if this is the first address being added
+    // or the only address being edited.
+    // The box is auto-checked in these cases by useAddressBookPage.js.
+    const isCheckboxDisabled =
+        (!isEditMode && numAddresses === 0) ||
+        (isEditMode && numAddresses === 1);
```
supernova-at, line 38: So this is worth calling out: If this is the first address being created, the talon will check the "make this my default address" checkbox and this code will disable it.

The same holds true when editing the only address.

The backend does allow a single address that is _not_ the default, which is a little funky.
In fact, you can't delete an address that is marked as the default, so this may complicate the implementation of removing an address (you'd have to set it as not the default first).

I'm open to suggestions on which way is better:
1. This, where the only address is assumed to be the default but removing it would require two calls (one to mark it as not the default, the second to actually delete it)
2. Don't automatically mark addresses as default. Always let the user control this.

packages/venia-ui/lib/components/AddressBookPage/addEditDialog.js
```diff
@@ -0,0 +1,178 @@
...
+                </div>
+            </div>
+        </Dialog>
+    );
+};
+
+export default AddEditDialog;
+
+AddEditDialog.propTypes = {
+    classes: shape({}),
```
supernova-at, line 172: Need to actually fill this out.

packages/venia-ui/lib/components/AddressBookPage/addressBookPage.js
```diff
@@ -38,15 +54,11 @@ const AddressBookPage = props => {
...
                     countryName={countryName}
+                    onEdit={handleEditAddress}
                 />
             );
         });
-
-        // sort the collection so the default is first
-        return addresses.sort(address =>
-            address.props.address.default_shipping ? -1 : 1
-        );
```
supernova-at, line 48: This code would sort the default address to the front but it made no guarantees about the order of the rest of the list.

While that's ultimately what the storefront wants, it caused issues during snapshot testing (the order of the rest of the list would sometimes change).

packages/venia-ui/lib/components/AddressBookPage/addressCard.css
```diff
@@ -49,8 +49,6 @@
 
 .editButton {
     composes: linkButton;
-    /** hidden until PWA-634 is completed */
-    visibility: hidden;
```
supernova-at, line 53: This PR completes PWA-634 😉

packages/venia-ui/lib/components/AddressBookPage/addressCard.js
```diff
@@ -25,6 +25,10 @@ const AddressCard = props => {
 
     const classes = mergeClasses(defaultClasses, propClasses);
 
+    const handleEdit = useCallback(() => {
+        onEdit(address);
+    }, [address, onEdit]);
```
supernova-at, line 30: Should I create a talon just for this?

Honestly I'd much prefer to pass a bound function here for this instead.

packages/venia-ui/lib/components/Checkbox/checkbox.css
```diff
@@ -44,12 +44,17 @@
...
+}
+
+.input:checked:enabled + .icon {
     --stroke: var(--venia-brand-color-1-700);
 }
 
-.input:active,
-.input:focus {
+.input:active:enabled,
+.input:focus:enabled {
```
supernova-at, line 57: I added these `:enabled` and `:disabled` pseudo-classes here to more distinctly visually differentiate an enabled checkbox from a disabled one.

Previously, a disabled checkbox still looked very much like an enabled one, you just couldn't click it. It felt more like a bug.

packages/venia-ui/lib/components/Region/region.js
```diff
@@ -79,6 +82,7 @@ Region.propTypes = {
     classes: shape({
         root: string
     }),
+    countryCodeField: string,
```
supernova-at, line 85: Also within this scope I added `countryCodeField` as a prop to `Region`.

`useRegion` actually already accepted this parameter, but the `Region` component never passed it in. This is basically just connecting the two ends. 

This allows `Region` components to define the `countryCodeField` for their sibling `Country` components within a form. `Region` needs this information so that it can populate itself based on the `Country` selection.

packages/venia-ui/i18n/en_US.json
jimbo: Good idea to do so, so it doesn't get out of date.

packages/venia-ui/lib/components/AddressBookPage/addEditDialog.js
jimbo: I think the former is probably more in line with UX expectations. It's not a big deal if the delete callback takes slightly more time to complete.

supernova-at: We talked about this in team time today. Resolved in https://github.com/magento/pwa-studio/pull/2879/commits/03880a99fbf40aed7e8cff6eaa7bc2c2a136fd06.

packages/venia-ui/lib/components/AddressBookPage/addressCard.js
tjwiebell: I support moving it up to `addressBookPage` in the creation loop; just pass bound functions instead of doing it here. I don't mind this as-is, but there may be a good argument to keep this purely presentational.

packages/venia-ui/lib/components/Checkbox/checkbox.css
tjwiebell: Awesome, this was definitely needed. Cursor only treatment doesn't mean much on mobile.

packages/venia-ui/lib/components/AddressBookPage/addEditDialog.js
```diff
@@ -0,0 +1,170 @@
...
+import Region from '@magento/venia-ui/lib/components/Region';
+import TextInput from '@magento/venia-ui/lib/components/TextInput';
+import defaultClasses from './addEditDialog.css';
+
+const AddEditDialog = props => {
+    const {
+        formErrors,
+        formProps,
+        handleCancel,
+        handleConfirm,
```
tjwiebell, line 23: Should these be `onCancel` and `onConfirm` as prop names?

packages/peregrine/lib/talons/AddressBookPage/useAddressBookPage.js
```diff
@@ -56,14 +90,93 @@ export const useAddressBookPage = (props = {}) => {
...
+                try {
+                    await updateCustomerAddress({
+                        variables: {
+                            addressId: activeEditAddress.id,
+                            updated_address: formValues
+                        },
+                        refetchQueries: [{ query: getCustomerAddressesQuery }]
+                    });
+
+                    setIsDialogOpen(false);
```
tjwiebell, line 134: I suppose this means refetch isn't blocking, which is why we get a flash of that stale data. It also appears to behave this way in the Checkout experience, not sure if I missed that or it's just amplified by a slow backend. This is fine as-is, I don't think we should dive into optimistic UI just yet.

tjwiebell (changes requested):
This looks really close, nice work 👍  added some comments about fragment usage, which seems to be an existing problem you didn't introduce.

packages/venia-ui/lib/components/AddressBookPage/__tests__/__snapshots__/addressBookPage.spec.js.snap
tjwiebell: React filters out `undefined`, but that's still odd given how many props should be truthy at the point of this snap. I'll play around with this on a local, might actually be exposing a bug; or might be catching an early render cycle. Did @jimbo fix something like this recently by waiting for a full render? `await act(...)` I think.

supernova-at: `numAddresses` was needed in order to determine whether or not we should disable the "make this my default" checkbox.

Since we decided not to do that on the front end, this prop has been removed in the end.

Still weird but I didn't dig into it further 😜

packages/peregrine/lib/talons/AddressBookPage/addressBookPage.gql.js
```diff
@@ -22,14 +23,54 @@ export const GET_CUSTOMER_ADDRESSES = gql`
...
+`;
+
+export const UPDATE_CUSTOMER_ADDRESS = gql`
+    mutation UpdateCustomerAddressInAddressBook(
+        $addressId: Int!
+        $updated_address: CustomerAddressInput!
+    ) {
+        updateCustomerAddress(id: $addressId, input: $updated_address)
+            @connection(key: "updateCustomerAddress") {
+            id
```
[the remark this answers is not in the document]
supernova-at: > my simple proposal is to wait to close the dialog until the refetch completes.

Man, I thought this was going to be very convoluted to accomplish but then I found [`awaitRefetchQueries`](https://www.apollographql.com/docs/react/data/mutations/):

> Queries refetched as part of `refetchQueries` are handled asynchronously, and are not waited on before the mutation is completed (resolved). Setting this to `true` will make sure refetched queries are completed before the mutation is considered done. `false` by default.

packages/peregrine/lib/talons/AddressBookPage/addressBookPage.gql.js
supernova-at: Removed from the `Add` mutation in https://github.com/magento/pwa-studio/pull/2879/commits/a10123de9d740b962336ba7496bce14cd78f6c93.

packages/peregrine/lib/talons/AddressBookPage/addressBookPage.gql.js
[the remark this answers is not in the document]
supernova-at: Added these query result fields to the `Update` mutation in https://github.com/magento/pwa-studio/pull/2879/commits/63796cd33180562a80a10504503d06778f1dc747.

packages/venia-ui/lib/components/AddressBookPage/addEditDialog.js
supernova-at: https://github.com/magento/pwa-studio/pull/2879/commits/340c38688f60dcbbed7bc46548969a0e96d0cd92

packages/venia-ui/lib/components/AddressBookPage/addressCard.js
supernova-at: Great call, done in https://github.com/magento/pwa-studio/pull/2879/commits/180b344b7bf2d373c662ec2cc168e0fc6a3098b4.

packages/peregrine/lib/talons/AddressBookPage/addressBookPage.gql.js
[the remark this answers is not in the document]
supernova-at: Adds fragments in https://github.com/magento/pwa-studio/pull/2879/commits/a87d3659f80e8cfea4ed0d59279ac4ec279dfd19.

packages/venia-ui/lib/components/AddressBookPage/addEditDialog.js
supernova-at: Nice catch, renamed in https://github.com/magento/pwa-studio/pull/2879/commits/8083d2a681f94d963c38306b8dd7a50ecbcc2c91 and unit tests updated in https://github.com/magento/pwa-studio/pull/2879/commits/cbc51f8fed92a17749b70ff6b35108c7d8e30492.

packages/peregrine/lib/talons/AddressBookPage/useAddressBookPage.js
supernova-at: > I don't think we should dive into optimistic UI just yet.

My thoughts exactly. `awaitRefetchQueries` solved my needs for now 👍

tjwiebell: Incredible, nice find!

tjwiebell (approved):

dpatil-magento:
@supernova-at Please check following observations - 

1.  While adding first Address and/or during editing Default address, `Make this my default address` should be disabled but currently its enabled.  
2. If user has item in cart and then user updates default address in account address book and go to checkout > User still sees stale address.
3. When user add Middle name in address, the middle name is not displayed in read only mode of address book.

supernova-at:
> While adding first Address and/or during editing Default address, Make this my default address should be disabled but currently its enabled.

This was actually implemented but we talked about this at length during a group code review with UX and decided to revert it. It is possible to not have a default address.

> If user has item in cart and then user updates default address in account address book and go to checkout > User still sees stale address.

After discussion we decided this is acceptable for now. Ideally GraphQL keeps the cart address in sync.

> When user add Middle name in address, the middle name is not displayed in read only mode of address book.

Fixed in https://github.com/magento/pwa-studio/pull/2879/commits/fc73ce7f64eb0c7e5315eeb5691581058b275c89.

supernova-at:
Issue found in QA:

1. Add or Update an address and set the country as "United Kingdom"
1. Type a value into the Region field
1. Save
1. Edit the address
1. Change the country dropdown to "United States"
1. Cancel the edit
1. Edit the address
1. See that the value is empty / blank / not filled with the proper value

<hr />

**Investigation**:

This happens because the form is seeded with `initialValues`, including an empty `region`.
When the edit dialog is opened, the form gets reset to its initial state and then the target address is filled in.
When the `Region` component in the form sees that the `country` value has changed, its `useEffect` fires and resets itself back to its initial value, which is empty.

Note that his happens with any country that doesn't populate the region dropdown, not just UK.
In fact, you can repeat the bug just by repeatedly editing an address in a country without a region dropdown.

<hr />

**Solution**:

The solution was actually fixing a bug in the `Dialog` component so that it would properly destroy the form when `shouldUnmountOnHide` was `true`. Now that `Dialog` properly unmounts the form we can use `initialValues` instead of `formApi.setValues` to set the edit form values. This avoids the "field got reset to an empty initial state" bug described above.

supernova-at:
Other places that use `shouldUnmountOnHide`:

1. Checkout Page > Edit Payment Information
1. Wishlist Page > Confirm Remove Product 

I should note that animation in on the dialog still works but the animation out appears instant. We can open a separate ticket for that if we'd like.

packages/peregrine/lib/talons/AddressBookPage/useAddressBookPage.js
```diff
@@ -70,7 +70,9 @@ export const useAddressBookPage = (props = {}) => {
     const [isDialogOpen, setIsDialogOpen] = useState(false);
     const [activeEditAddress, setActiveEditAddress] = useState();
     const isDialogEditMode = !!activeEditAddress;
-    const [formApi, setFormApi] = useState();
+    const [formInitialValues, setFormInitialValues] = useState({
+        country_code: 'US'
+    });
```
tjwiebell, line 75: You may not need a duplicate piece of state here; pass `activeEditAddress` as initialValues and update `handleAddAddress` to pass this default value when it calls `setActiveEditAddress`, eg. `setActiveEditAddress({ country_code: 'US' });`.

Regardless of this duplication, I think there is a bug here where `initialValues` is not being set correctly on add, and the previous edited address is being displayed. Either go with my proposal above, or add `setFormInitialValues({ country_code: 'US' });` to that handler if there's a necessity to have two copies of this state.

tjwiebell (changes requested):
Tiny bug with add address flow. I'll also look into a solution to retaining the animation, it is a bit jarring.

packages/peregrine/lib/talons/AddressBookPage/useAddressBookPage.js
supernova-at: Dang, you're totally right - great catch

tjwiebell:
@supernova-at - I did work out an `onTransitionEnd` solution, but **every** transition inside of the Dialog bubbles up to this handler, so I can see it becoming fragile unless we add some target/ref validation. Definitely understand @jimbo's concerns with this approach a little better now.

```javascript
const [readyToUnmount, setReadyToUnmount] = useState(true);
// handles onTransitionEnd prop on element - fires both ways (open/close)
const handleTransitionEnd = useCallback(e => {
    if (e.propertyName === 'opacity') {
        setReadyToUnmount(current => !current);
    }
}, []);
```

tjwiebell (approved):
Fixes look good, nice work 👍  I do believe we should solve the animation breakage before we release, so please get that in the backlog.

dpatil-magento:
@supernova-at Just a minor observation >> When user uncheck `Make this my default address`  checkbox > Address card is moved to second position, which is not needed I believe.  Rest all looks good.

PR #2838, opened 2020-10-27
Payment extendable

larsroettig:
## Description
As developer, we want to build our own Payment methods as Extension for that reasons.
Payments should be extendable via Targetables

## Related Issue
Closes #2816

### Verification Steps
1. Checkout should work as before 
2.  Scaffolding
```
DEBUG_PROJECT_CREATION=true ./pwa-studio/packages/pwa-buildpack/bin/buildpack create-project payment-test-1 --template "venia-concept" --name "payment-test-1" --author "larsroettig" --backend-url "https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/" --braintree-token "sandbox_8yrzsvtm_s2bg8fs563crhqzk" --npm-client "yarn"
```
3. Payment addable via 
payment-test-1/local-intercept.js
```js 
function localIntercept(targets) {
    targets.of("@magento/venia-ui").payments.tap((payments) =>
        payments.add({
            paymentCode: "custome",
            importPath: "src/components/MyPayment/payment.js",
        })
    );
}
```

## Screenshots / Screen Captures (if appropriate)

Screenshot at 11-50-24

## Checklist

* I have added tests to cover my changes, if necessary.
* I have added translations for new strings, if necessary.
* I have updated the documentation accordingly, if necessary.

PR #2882, opened 2020-12-01
Saved payment methods main view.

revanth0212:
## Description

Saved Payments page, main view.

This PR also introduces a new Venia UI target called `savedPaymentTypes` which the extension packages can use to add a newly saved payment type renderer.

Also, the target that @larsroettig worked on as part of #2838 has been renamed to `checkoutPagePaymentTypes` to represent its purpose a bit more accurately.

## Related Issue
Closes [PWA-636](https://jira.corp.magento.com/browse/PWA-636)

### Verification Stakeholders
@schensley 
@dpatil-magento 

### Verification Steps
1. Go to the deployed app's `/saved-payments` page.
2. If there are any saved payment methods on the account, they should show up on the page.

image

image

## Checklist


packages/venia-ui/lib/components/SavedPaymentsPage/creditCard.js
```diff
@@ -0,0 +1,85 @@
...
+    DN: 'Diners',
+    ELO: 'Elo',
+    HC: 'Hipercard',
+    JCB: 'JCB',
+    MC: 'MasterCard',
+    MD: 'Maestro Domestic',
+    MI: 'Maestro International',
+    UN: 'UnionPay',
+    VI: 'Visa'
+};
```
sirugh, line 29: Is this something we can get from the backend? We would have to do a release of PWA Studio just to account for new card types added or if codes change. I don't know how often that happens, so I guess it really just depends.

revanth0212: You are right about it. I have created https://jira.corp.magento.com/browse/MC-39592 on the GQL team's board to get these details from the GQL query.

schensley:
@revanth0212 This looks great.  One tiny point, ideally in the mobile view the "delete" tool would collapse to show the icon only.

revanth0212:
> @revanth0212 This looks great. One tiny point, ideally in the mobile view the "delete" tool would collapse to show the icon only.

Got it, nice point. Let me fix that.

schensley:
@revanth0212 Thanks. UX Approved

packages/venia-ui/lib/components/SavedPaymentsPage/creditCard.js
```diff
@@ -0,0 +1,85 @@
...
+
+const CreditCard = props => {
+    const { classes: propClasses, details, public_hash } = props;
+    const classes = mergeClasses(defaultClasses, propClasses);
+
+    const number = `**** ${details.maskedCC} \u00A0\u00A0 ${
+        cardTypeMapper[details.type]
```
tjwiebell, line 36: Just in case let's not render `undefined` to the screen. Another reason we should also consider not using unicode here to get the spacing you want; consider using css.
```suggestion
        cardTypeMapper[details.type] || ''
```

packages/venia-ui/lib/components/SavedPaymentsPage/creditCard.js
```diff
@@ -0,0 +1,85 @@
...
+    }`;
+    const cardExpiryDate = useMemo(() => {
+        const [month, year] = details.expirationDate.split('/');
+        const longMonth = new Date(+year, +month - 1).toLocaleString(
```
tjwiebell, line 40: Looks like this became a `shortMonth` at some point. Also minor, but it's also missing the punctuation from the mocks, though I'm not certain how i18n friendly that would be.

packages/venia-ui/lib/components/SavedPaymentsPage/savedPaymentsPage.js
```diff
@@ -1,15 +1,14 @@
...
 
 import { useSavedPaymentsPage } from '@magento/peregrine/lib/talons/SavedPaymentsPage/useSavedPaymentsPage';
 import { mergeClasses } from '@magento/venia-ui/lib/classify';
 
-import Icon from '../Icon';
-import LinkButton from '../LinkButton';
 import { Title } from '../Head';
+import PaymentCard from './paymentCard';
 
 import { GET_SAVED_PAYMENTS_QUERY } from './savedPaymentsPage.gql';
```
tjwiebell, line 10: Consider moving this to peregrine.

packages/venia-ui/lib/components/SavedPaymentsPage/creditCard.js
```diff
@@ -0,0 +1,85 @@
...
+                        id={'storedPayments.delete'}
+                        defaultMessage={'Delete'}
+                    />
+                </span>
+            </span>
+        </button>
+    );
+
+    return (
+        <div className={classes.root} key={public_hash}>
```
tjwiebell, line 71: I don't think you need to define this key twice.

packages/venia-ui/lib/components/SavedPaymentsPage/savedPaymentsPage.js
```diff
@@ -19,54 +18,32 @@ const SavedPaymentsPage = props => {
...
-                </div>
-            )
-        );
-    }, [savedPayments]);
+    const savedPaymentElements = useMemo(
+        () =>
+            savedPayments.map(paymentDetails => (
+                <PaymentCard
+                    key={paymentDetails.public_hash}
+                    {...paymentDetails}
```
tjwiebell, line 30: We need to be explicit about the interface for this target; I would use a named prop here.

packages/venia-ui/lib/components/SavedPaymentsPage/paymentCard.js
```diff
@@ -0,0 +1,11 @@
...
+
+import paymentCardMapper from './savedPaymentTypes';
+
+const PaymentCard = props => {
+    const PaymentComponent = paymentCardMapper[props.payment_method_code];
+
+    return <PaymentComponent {...props} />;
+};
+
+export default PaymentCard;
```
tjwiebell, line 11: You can define that expected interface here via propType definitions.

tjwiebell (changes requested):
This is mostly there, and I'll address my own feedback and get another reviewer on this. Nice work! Really like the forward thinking of extensibility; this will likely be re-usable to some extent in the Checkout experience.

packages/peregrine/lib/talons/AccountMenu/useAccountMenuItems.js
```diff
@@ -23,7 +23,7 @@ export const useAccountMenuItems = props => {
         {
             name: 'Store Credit & Gift Cards',
             id: 'accountMenu.storeCreditLink',
-            url: ''
+            url: '/saved-payments'
```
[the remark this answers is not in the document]
revanth0212: My bad

packages/venia-ui/lib/components/SavedPaymentsPage/savedPaymentsPage.js
revanth0212: Can you elaborate on this? Sorry, I dint get your point.

tjwiebell: I initially was thinking this wasn't very extensible if someone changed the query to return other fields, thus changing the interface for this component. Expected to see `data={paymentDetails}` here but seemed like overkill after seeing how simple the response was, and that any extra data would be serialized in `details` anyway.

I instead just added propTypes for the expected query response [here](https://github.com/magento/pwa-studio/pull/2882/commits/d3ea89e0455481f40e05356f2b3dfadb14ebd6f0#diff-f6533ffec4057589b73942585c5f19e161a2fc4353dde8a1c0d70565c1275a3dR14-R17).

packages/venia-ui/lib/components/SavedPaymentsPage/creditCard.js
```diff
@@ -0,0 +1,81 @@
...
+        >
+            <Icon classes={{ icon: undefined }} size={16} src={DeleteIcon} />
+            <span className={classes.deleteText}>
+                <FormattedMessage
+                    id={'storedPayments.delete'}
+                    defaultMessage={'Delete'}
+                />
+            </span>
+        </LinkButton>
+    );
```
davemacaulay, line 64: If the delete functionality is not included as part of this functionality I would not include the button in the UI.

davemacaulay (dismissed):

packages/venia-ui/lib/components/SavedPaymentsPage/creditCard.js
revanth0212: We do have the mutation to delete a payment method, `DeletePaymentTokenOutput`. This functionality will be handled in https://jira.corp.magento.com/browse/PWA-638

tjwiebell (dismissed):

packages/venia-ui/lib/components/SavedPaymentsPage/creditCard.js
```diff
@@ -1,13 +1,13 @@
 import React, { useCallback, useMemo } from 'react';
+import { shape, string } from 'prop-types';
 import { FormattedMessage } from 'react-intl';
 import { Trash2 as DeleteIcon } from 'react-feather';
 
-import { mergeClasses } from '@magento/venia-ui/lib/classify';
-
+import LinkButton from '../LinkButton';
```
tjwiebell, line 6: Did your auto-sorter do this? Import order isn't worth fixing, but watch your tooling if this wasn't intentional.

tjwiebell (approved):

packages/venia-ui/lib/components/SavedPaymentsPage/creditCard.js
revanth0212: This was intentional. I have a habit of moving package imports to the top, above the relative imports, and below the npm imports.

revanth0212: It was kind of intentional.

tjwiebell (approved):

dpatil-magento:
@revanth0212 Please check following - 
1. When No card is saved, related message is not displayed.


2. Saved card info is saved in Apollo cache even after logout, Should we treat it same as email which gets removed after logout?( when I login to another account for a fraction of time this info is displayed in Saved Payment page.)
[{"__typename":"PaymentToken","details":"{\"type\":\"VI\",\"maskedCC\":\"1111\",\"expirationDate\":\"02\\/2022\

3. When Braintree Vault is disabled, these saved payment options still displayed,  expected? 
(Don't see any requirement about this)

revanth0212:
> @revanth0212 Please check following -
> 
> 1. When No card is saved, related message is not displayed.
> 
> 
> 
> 1. Saved card info is saved in Apollo cache even after logout, Should we treat it same as email which gets removed after logout?( when I login to another account for a fraction of time this info is displayed in Saved Payment page.)
>    [{"__typename":"PaymentToken","details":"{"type":"VI","maskedCC":"1111","expirationDate":"02\/2022\
> 2. When Braintree Vault is disabled, these saved payment options still displayed,  expected?
>    (Don't see any requirement about this)

Good points @dpatil-magento. I have addressed the first 2 points. Regarding vault, I am not totally sure. We are rendering what the GQL service is returning. It would be a good question for the GQL team.

dpatil-magento:
Thanks @revanth0212  for fix and creating ticket on GQL for 3rd issue. QA Approved.

packages/peregrine/lib/Apollo/policies/index.js
```diff
@@ -143,6 +143,9 @@ const typePolicies = {
             }
         }
     },
+    CustomerPaymentTokens: {
+        keyFields: () => 'CustomerPaymentTokens'
```
tjwiebell, line 147: This is a clever solution, just give the collection an identity. I would normally suggest that the `PaymentToken` type needs an identity, but that may not be necessary since there's no way to mutate it. I also don't like adding these hidden requirements to queries, not sure how Apollo reacts when the field for the key identity is missing.

@sirugh - is there a technical reason we should do something like:
```javascript
PaymentToken: {
    keyFields: ({ public_hash }) => `CustomerPaymentToken:${public_hash}`
},
```

tjwiebell (changes requested):
Remove some code and this is good to go. Have a question about `keyFields`, but not a blocker in this scope.

packages/peregrine/lib/Apollo/policies/index.js
revanth0212: The solution you are proposing would work on individual tokens, the one I have here is for the whole collection. I feel like adding `keyFields` to each token would be overkill unless it provides some use. I am not equipped enough to make the call about the apollo types we are using here, I am fine with what the team thinks is the right approach.

packages/peregrine/lib/Apollo/clearCartDataFromCache.js
```diff
@@ -16,4 +16,7 @@ export const clearCartDataFromCache = async client => {
     await deleteCacheEntry(client, key =>
         key.match(/^\$?AvailableShippingMethod/)
     );
+    await deleteCacheEntry(client, key =>
+        key.match(/^\$?customerPaymentTokens/)
+    );
```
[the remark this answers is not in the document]
revanth0212: True, I forgot to remove it from this file.

tjwiebell (approved):

dpatil-magento:
QA Re-approved.

packages/peregrine/lib/Apollo/policies/index.js
sirugh: > If one of your types defines its primary key with a field besides id or _id, you can customize how the InMemoryCache generates unique identifiers for that type. To do so, you define TypePolicy for the type. You specify all of your cache's typePolicies in the options object you provide to the InMemoryCache constructor.

This is from the [docs](https://www.apollographql.com/docs/react/caching/cache-configuration/#data-normalization). I wouldn't try to get much more creative that our use case. We only use this because we need a handle on the cache key in order to delete it. If we find that other properties are still cached and not garbage collected after the deletion of these other objects then we may need to identify PaymentTokens with keyFields and delete them too.

PR #2888, opened 2020-12-03
Delete Address from Address Book

supernova-at:
## Description

This PR adds the ability to delete addresses from the Address Book.

## Related Issue

Closes PWA-635.

### Verification Stakeholders

@schensley @dpatil-magento 

### Verification Steps

1. Sign in
1. Go to the Address Book page
1. Verify that a "Delete" button appears for addresses that are not the default address
1. Verify that clicking the "Delete" button presents a confirmation overlay on the target address
1. Click "Cancel"
1. Verify that the confirmation overlay disappears and nothing happens
1. Click the "Delete" button again
1. Click "Delete" in the confirmation overlay
1. Verify that the buttons are disabled for a second while the request is in flight
1. Verify that the target address is deleted and the page refreshes<sup>1</sup>

<sup>1</sup>: There is a slight delay after the overlay disappears and the address actually gets removed. This is due to a workaround due to a graphQL bug and will be resolved in another scope.

Screen Shot 2020-12-03 at 12 15 53 PM

Screen Shot 2020-12-03 at 12 16 01 PM

Mobile:
Screen Shot 2020-12-03 at 12 16 21 PM

## Checklist

- [x] The confirmation buttons extend past their container sometimes


packages/venia-ui/lib/components/AddressBookPage/addressCard.css
```diff
@@ -5,13 +5,43 @@
...
     animation: flash var(--venia-global-anim-bounce) 640ms 2;
 }
 
+.confirmDeleteContainer {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    background-color: rgba(255, 255, 255, 0.9);
```
supernova-at, line 22: See if we can use a token for this.

packages/venia-ui/lib/components/AddressBookPage/addressCard.css
```diff
@@ -5,13 +5,43 @@
...
+    grid-template-columns: 1fr 1fr;
+    gap: 0.5rem;
+    padding: 1rem;
+}
+
+.confirmDeleteButton {
+    composes: root_normalPriorityNegative from '../Button/button.css';
+
+    background-color: white;
+    max-width: 6rem;
```
supernova-at, line 35: It'd be great to use a token for the color. 
As for the `max-width`, this is very specific to how the English words look. We may not want this in the end.

supernova-at:
The deletion confirm buttons no longer extend past their container but the mocks call for the "Delete" button to be on the right (second) when they're in columns, and on top (first) when they're stacked in rows.

Unfortunately there's no way to do that cleanly.

tjwiebell:
> Unfortunately there's no way to do that cleanly.

I solved for this recently in some [createForm re-factoring](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CreateAccount/createAccount.css), should be possible. Grid position does not apply to keyboard nav, but that's actually exactly the behavior W3C suggests (submit/confirm should be the first focusable action).

schensley:
@supernova-at things look pretty good. UX approved.

packages/venia-ui/lib/components/AddressBookPage/addressCard.css
revanth0212: I'll leave this for a future use case. Let us slide it for this PR though.

packages/venia-ui/lib/components/AddressBookPage/addressCard.js
```diff
@@ -25,6 +36,12 @@ const AddressCard = props => {
     } = address;
 
     const classes = mergeClasses(defaultClasses, propClasses);
+    const confirmDeleteButtonClasses = {
```
revanth0212, line 39: Interesting choice of code. Any particular reason why you moved this out of JSX?

packages/venia-ui/lib/components/AddressBookPage/addressCard.js
```diff
@@ -63,6 +77,36 @@ const AddressCard = props => {
         </LinkButton>
     ) : null;
 
+    const maybeConfirmingDeleteOverlay = isConfirmingDelete ? (
```
revanth0212, line 80: Looks like there are tests missing for this change.

revanth0212 (dismissed):
Looks perfect. Almost ready to be merged. A couple of tests need to be added, will do that myself and approve it once done.

revanth0212:
Can someone review the following commits: 0afe723 58b2424

tjwiebell (approved):
New tests look good 👍

dpatil-magento:
QA Approved.

PR #2891, opened 2020-12-03
Allow self signed certs for backend validation and only attach to scaffold sample

sirugh:
## Description

The validator was running against every back end, even your own custom backend which you most likely will have configured when developing your storefront.

If you used a local backend with some self signed cert, it fails to validate:

```
~/code/pwa-studio develop
λ node
Welcome to Node.js v14.15.1.
Type ".help" for more information.
> const fetch = require('node-fetch');
undefined
> fetch('https://latest.test')
Promise { <pending> }
> (node:64113) UnhandledPromiseRejectionWarning: FetchError: request to https://latest.test/ failed, reason: unable to verify the first certificate
    at ClientRequest.<anonymous> (/Users/stephen/code/pwa-studio/node_modules/node-fetch/lib/index.js:1444:11)
    at ClientRequest.emit (events.js:315:20)
    at ClientRequest.EventEmitter.emit (domain.js:486:12)
    at TLSSocket.socketErrorListener (_http_client.js:469:9)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket.EventEmitter.emit (domain.js:486:12)
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:64113) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:64113) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

One solution would be to allow self signed certs through the `https` agent.

## Related Issue

Closes [PWA-1182](https://jira.corp.magento.com/browse/PWA-1182).

### Verification Steps

In all cases, I have removed the `console.warn` that was displaying on every `yarn` command.

1. Check out PR.
2. Run `yarn`. It shouldn't say anything because the extension isn't installed in the core repo.
3. Using the PR, scaffold a new app: `DEBUG_PROJECT_CREATION=1 pwa-studio/packages/create-pwa/bin/create-pwa`.
4. in the new app dir, run `yarn build` and you should see the console warning.
5. In the new app, swap the backend url to an invalid url and run `yarn build`. The error should display with backends to try.

Also, navigate to the `validate-sample-backends` extension directory and run `yarn test` to ensure the tests pass. I don't think our test runner validates extensions.

Try this with 

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#2894: Allow self signed certs for backend validation


sirugh:
@magento create issue from PR

packages/extensions/venia-sample-backends/intercept.js
```diff
@@ -29,56 +35,61 @@ const fetchBackends = async debug => {
...
+ * @param {Object} config.env - The ENV provided to the app, usually available through process.ENV
  * @param {Function} config.onFail - callback function to call on validation fail
  * @param {Function} config.debug - function to log debug messages in console in debug mode
  *
- * To watch the debug messages, run the command with DEBUG=*runEnvValidators*
+ * To watch the debug messages, run the command with DEBUG=pwa-buildpack:runEnvValidators
  */
 const validateSampleBackend = async config => {
-    console.warn(
-        "\n venia-sample-backends is a Dev only extension, please remove it from the project's package.json before going to production. \n"
```
sirugh, line 45: I removed this message. The original intent of [my request](https://github.com/magento/pwa-studio/pull/2853#issuecomment-736123153) was that we document in the `Getting Started` or `FAQ` that this extension should be removed from your package.json dependencies outside of development. Now that we only validate sample backends, it shouldn't be as big of a deal, but every person who forgets to remove it will be inadvertently hitting the `getSampleBackends` endpoint for no reason.

packages/extensions/venia-sample-backends/intercept.js
```diff
@@ -29,56 +35,61 @@ const fetchBackends = async debug => {
...
-
-        debug('Fetching other backends');
-
-        const sampleBackends = await fetchBackends(debug);
-        const otherBackends = sampleBackends.filter(
-            ({ url }) => url !== env.MAGENTO_BACKEND_URL
+    if (!sampleBackendUrls.includes(env.MAGENTO_BACKEND_URL)) {
+        // We don't need to validate anyone's custom backend. The validator is
+        // only for the sample backends we provide.
+        debug(`Not validating custom backend: ${env.MAGENTO_BACKEND_URL}`);
```
sirugh, line 57: This is important - if someone is playing around with PWA Studio using a sample backend, we want to inform them of alternatives if it goes down. But if someone has already swapped to their custom backend, we don't need to validate it.

packages/extensions/venia-sample-backends/intercept.js
```diff
@@ -29,56 +27,61 @@ const fetchBackends = async debug => {
...
-    );
-
     const { env, onFail, debug } = config;
 
-    const backendIsActive = await isBackendActive(env, debug);
-
-    if (!backendIsActive) {
-        debug(`${env.MAGENTO_BACKEND_URL} is inactive`);
+    const sampleBackends = await fetchBackends(debug);
+    const sampleBackendUrls = sampleBackends.map(({ url }) => url);
```
jimbo, line 44: Do you think we could use a few precautions to make sure these identities are actually arrays and not undefined or null?

jimbo (approved):

packages/extensions/venia-sample-backends/intercept.js
revanth0212: I feel we should have something yelling at them to remove the package, like a warning. We do the same with Braintree token but that's at a different level than this validation. What does the team think?

packages/extensions/venia-sample-backends/intercept.js
revanth0212: We are not checking for null or undefined but we do leave them in a try catch. https://github.com/magento/pwa-studio/pull/2891/files#diff-12c9882122c89016650304307999b91ea76326b0514c9bd254372e69eaefa698R21

packages/extensions/venia-sample-backends/intercept.js
revanth0212: We dint do this on purpose. If we do this, irrespective of whether they are using a custom or sample backend, we will be hitting the Sample Backends URL to fetch the list, which will be a burden on the server. That's the reason why we only fetch the list if the given backend is down.

packages/extensions/venia-sample-backends/intercept.js
```diff
@@ -29,56 +27,61 @@ const fetchBackends = async debug => {
...
-        "\n venia-sample-backends is a Dev only extension, please remove it from the project's package.json before going to production. \n"
-    );
-
     const { env, onFail, debug } = config;
 
-    const backendIsActive = await isBackendActive(env, debug);
-
-    if (!backendIsActive) {
-        debug(`${env.MAGENTO_BACKEND_URL} is inactive`);
+    const sampleBackends = await fetchBackends(debug);
```
tjwiebell, line 43: This swaps one network call on every yarn command for another, we don't want this either. I haven't pulled this down, but does this still run on every command still?

packages/extensions/venia-sample-backends/intercept.js
sirugh: The `fetchBackends` function returns the `environments` prop which we hardcode to be an array. I can add a safety check there.

revanth0212 (changes requested):

packages/extensions/venia-sample-backends/__tests__/__snapshots__/intercept.spec.js.snap
```diff
@@ -13,9 +13,3 @@ exports[`should call onFail with a different error message if environments is em
  [{\\"name\\":\\"2.3.3-venia-cloud\\",\\"description\\":\\"Magento 2.3.3 with Venia sample data installed\\",\\"url\\":\\"https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/\\"}]
```
sirugh, line 13: This seems incorrect. If the environments is empty, where is this "other backend" data coming from? I'll have to fix the test.

packages/extensions/venia-sample-backends/intercept.js
revanth0212: You are right @tjwiebell. We should not be doing this. It will be running on every yarn command and that will be huge burden on the server.

sirugh: @tjwiebell oh jeez you are right. My intention was to only do an up-check against sample backends. But to obtain a list of sample backends to see if we're using one, I have to make a fetch.

packages/extensions/venia-sample-backends/intercept.js
sirugh: Ah dang. Right. Hmm.

packages/extensions/venia-sample-backends/__tests__/__snapshots__/intercept.spec.js.snap
sirugh: Fixed. fetch wasn't updating mock for some reason. Had to move test order.

packages/extensions/venia-sample-backends/__tests__/__snapshots__/intercept.spec.js.snap
```diff
@@ -1,21 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`should call onFail if backend is inactive 1`] = `
-"https://www.magento-backend-2.3.4.com/ is inactive. Please consider using one of these other backends: 
-
- [{\\"name\\":\\"2.3.3-venia-cloud\\",\\"description\\":\\"Magento 2.3.3 with Venia sample data installed\\",\\"url\\":\\"https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/\\"}] 
+"https://www.magento-backend-2.3.4.com/ is inactive. Please consider using an active backend.
```
sirugh, line 4: Huh. Now this one is wrong. lol

packages/extensions/venia-sample-backends/intercept.js
sirugh: Personally I don't want to be reminded that we have this extension installed on the core repo.

A solution that only installs the extension after scaffolding would suffice.

packages/venia-concept/_buildpack/create.js
```diff
@@ -34,6 +34,11 @@ function createProjectFromVenia({ fs, tasks, options }) {
         'storybook:build': 'build-storybook -c src/.storybook -o storybook-dist'
     };
 
+    // Dependencies we want to ensure are added to the scaffolded app.
+    const devDependenciesToInsert = {
+        '@magento/venia-sample-backends': '~0.0.1'
```
sirugh, line 39: Does this seem OK to folks? I believe the extension was meant to be used for scaffolded apps anyways.

revanth0212: Yeah the extension is for new apps created using PWA Studio. This is ok with me.

tjwiebell: I'm going to say no, unless there is strong evidence that there is a large portion of scaffolded apps using our sample backend. This just seems like unnecessary spam, and we shouldn't annoy developers with a necessary step they have to take to go live that they didn't opt-in to.

tjwiebell (changes requested):
I would love a second opinion on this, but the team already had reservations about this being overly complex, and this ended up being very different from what we discussed at team time. With the additional bugs it's causing in other CLI commands, I think I'm at the point that I would prefer reverting this entirely. We should keep the microservice and integration with initial scaffolding, but I think that is where this should end.

packages/venia-concept/_buildpack/create.js
sirugh: Compromise: 

Only attach dependency if user has selected one of the sample backends.

tjwiebell (dismissed):
We share reservations about the double fetch, but this immediately solves the problem. I would also argue the second fetch enables support for anyone using `buildpack create-project` without the creator, so there's the justification 😄

dpatil-magento:
@sirugh Please take care of test failure and `lodash `  lock file change needs to be committed.

tjwiebell (approved):

revanth0212 (approved):

PR #2895, opened 2020-12-04
[PWA-1099] Add support for Product URL and Category URL Settings

anthoula:
## Description

As a admin user, I would like the ability to change settings in the admin panel to control the product URL and category URL, and view it in my Venia storefront.

**Admin Config Settings:**

- Stores > Configuration > Catalog > Catalog > Search Engine Optimization > Product URL Suffix
- Stores > Configuration > Catalog > Catalog > Search Engine Optimization > Category URL Suffix

## Related Issue

https://jira.corp.magento.com/browse/PWA-1099

### Verification Steps

1. Switch store view after directly accessing Category via Home Page, Search Suggestions, Bread Crumbs, Left Nav Category Menu, Bookmarks and verify url is being updated per settings.
2. Switch store view after directly accessing Product via Home/Search Page, Search Suggestions, Bookmarks, Mini Cart/Cart and verify url is being updated per settings.
3. Verify on updating invalid url suffix user is displayed with 404 error.

## Checklist


packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -40,8 +49,12 @@ const mapAvailableOptions = (config, stores) => {
 
 export const useStoreSwitcher = props => {
     const { queries } = props;
-    const { getStoreConfigData, getAvailableStoresData } = queries;
-    const history = useHistory();
+    const {
+        getStoreConfigData,
+        getUrlResolverData,
+        getAvailableStoresData
+    } = queries;
```
sirugh, line 56: Could you modify this code to use the new DEFAULT_OPERATIONS pattern?

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -76,12 +100,41 @@ export const useStoreSwitcher = props => {
...
+                suffix = availableStores.get(storeCode).productSuffix || '';
+            }
+
+            return suffix;
+        },
+        [availableStores, pageType]
+    );
+
+    // Handle updating the URL with configured product or catalog url suffix
+    const setNewUrl = useCallback(
```
sirugh, line 120: I'd name this something else - it's not actually setting anything, just making a string.

In fact you could probably do away with this function and just put lines 122-125 in the `else` of the switch callback.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -102,38 +102,36 @@ export const useStoreSwitcher = props => {
...
         storeCode => {
             // Do nothing when store view is not present in available stores
             if (!availableStores.has(storeCode)) return;
 
-            const suffix = getSuffix(storeCode);
+            // Use window.location.pathname to get the path with the store view code
+            // pathname from useLocation() does not include the store view code
+            const pathName = window.location.pathname.split('.')[0];
+            const index = window.location.pathname.split('.')[1];
+            const currentSuffix = index ? '.' + index : '';
```
sirugh, line 132:
```suggestion
            const [pathName, index] = window.location.pathname.split('.');
            const currentSuffix = index ? '.' + index : '';
```

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -76,25 +102,49 @@ export const useStoreSwitcher = props => {
...
+
     const handleSwitchStore = useCallback(
         // Change store view code and currency to be used in Apollo link request headers
         storeCode => {
             // Do nothing when store view is not present in available stores
             if (!availableStores.has(storeCode)) return;
 
+            // Use window.location.pathname to get the path with the store view code
+            // pathname from useLocation() does not include the store view code
+            const [pathName, index] = window.location.pathname.split('.');
```
eug123, line 132: I think we can use current suffix for page type from config of the current store.

anthoula: For pages that are not category/product, such as `search.html` we may need to derive the suffix from window.location.pathname. That case is a client-side route, and pageType returned is `undefined`.

eug123: I don't think we need to derive suffix from pathname from the page types other that category/product, just use the whole pathname as is.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -76,25 +102,49 @@ export const useStoreSwitcher = props => {
...
+            }
+            if (pageType === 'PRODUCT') {
+                suffix =
+                    availableStores.get(storeCode).product_url_suffix || '';
+            }
+
+            return suffix;
+        },
+        [availableStores, pageType]
+    );
```
sirugh, line 122: I see what @eug123 is saying.

This function should return two things - the suffix and the pathname. Because configured suffix delimiter could be a random char, and not `.`, we can't always assume to split on `.`.

In this case, you'll have to use different logic to determine the suffix/pathnames for pdp/category.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -76,25 +102,61 @@ export const useStoreSwitcher = props => {
...
+            const params = window.location.search || '';
+
             storage.setItem('store_view_code', storeCode);
             storage.setItem(
                 'store_view_currency',
                 availableStores.get(storeCode).currency
             );
 
             // Handle updating the URL if the store code should be present.
+            // Handle updating the URL with configured catalog or product URL suffix
```
sirugh, line 156: Can remove this comment, I think.

eug123 (approved):
Looks good to me, thank you

packages/peregrine/lib/talons/Header/__tests__/useStoreSwitcher.spec.js
```diff
@@ -233,31 +255,125 @@ describe('handleSwitchStore updates url with configured store code and url suffi
             assign: jest.fn()
         };
 
-        handleSwitchStore('store1');
+        handleSwitchStore('store2');
 
+        // .html => null
```
sirugh, line 260: Is this comment wrong or the assertion innaccurate?

packages/peregrine/lib/talons/OrderHistoryPage/__tests__/__snapshots__/orderHistoryContext.spec.js.snap
```diff
@@ -4,7 +4,7 @@ exports[`renders children 1`] = `"Context Provider Children"`;
 
 exports[`returns default value for suffix 1`] = `
```
sirugh, line 5: The statement for this test could be updated, though I suppose empty string could be the default. I would just say `returns empty string if no suffix found in gql` or something like that.

packages/venia-ui/lib/components/Header/storeSwitcher.js
```diff
@@ -7,7 +7,7 @@ import { useStoreSwitcher } from '@magento/peregrine/lib/talons/Header/useStoreS
 import { mergeClasses } from '../../classify';
 import defaultClasses from './storeSwitcher.css';
 import SwitcherItem from './switcherItem';
-import storeSwitcherOperations from './storeSwitcher.gql';
+import storeSwitcherOperations from '@magento/peregrine/lib/talons/Header/storeSwitcher.gql';
```
sirugh, line 10: You no longer need to pass these operations to the talon since the talon imports and defaults to them when nothing is passed. You can remove this import and the passed arg on line 14-16.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -39,9 +50,13 @@ const mapAvailableOptions = (config, stores) => {
  */
 
 export const useStoreSwitcher = props => {
```
sirugh, line 52: You'll want to update this to `(props = {})`

sirugh (changes requested):
A few small changes, but nothing that should affect tests.

packages/peregrine/lib/talons/Header/__tests__/useStoreSwitcher.spec.js
anthoula: Good catch, comment is wrong. This test is to check for store code, not the suffix

sirugh (dismissed):

dpatil-magento:
@anthoula Unable to proceed as Home page is not displaying, tried multiple backends but same issue.

image

anthoula:
> @anthoula Unable to proceed as Home page is not displaying, tried multiple backends but same issue.

Thanks @dpatil-magento, should be resolved now

dpatil-magento:
Thanks @anthoula , as discussed product URL suffix is being added to product url key var in request causing product not being returned. In following var `test` is url suffix.

image

anthoula:
Decided to remove the unused util function `getUrlKey()`, instead of deprecating, as it may surface future bugs if left around.

eug123 (dismissed):

dpatil-magento:
@anthoula Pls take a look on following observations - 

1. After Hitting an error (be on category A….and switch to store 2 which does not have this category …you see error)....switch store view back  > category/product url not updating.
https://gyazo.com/08f7a3b8ef041222dd0a77791d3af6b3

2. Navigating fast between store and clicking product in category page….hits error as it still has old url link
https://gyazo.com/0d22f4651ebeee75a7b608433b9110a8  Note - I had already accessed that product so it was cached.  After switching store view I didn't give much time to update product url so its still trying to hit old url

anthoula:
> Thanks @anthoula , as discussed product URL suffix is being added to product url key var in request causing product not being returned. In following var `test` is url suffix.

Thanks @dpatil-magento, should be resolved now

sirugh:
> Navigating fast between store and clicking product in category page….hits error as it still has old url link

Let's say you have two store views as follows:

* View A uses page suffixes `.html` and View B uses no page suffixes.
* Both views share category/product data.

If you are on a category page in a view and use the store switcher to change to another view, you'd expect to see the same category page, and we would expect it to render this fairly quickly after the page refresh from the store code switch since all that data is already cached. The problem here is that part of that data is _stale_, specifically the `url_suffix` values for the category/product. 

So if you're on a category page for View A and switch to View B and quickly click a product image to go to the product page, you'll be navigated, using an invalid link, to a product page with `.html` suffixes.

It seems we're at odds with our cache-and-network strategy now.

Options: 

1) Refactor Product/Category page links to be disabled until `storeConfig.category/product_url_suffix` has resolved. (slower TTR for links)
2) Do nothing (bad for slow network)
3) ?

anthoula:
> 1. After Hitting an error (be on category A….and switch to store 2 which does not have this category …you see error)....switch store view back  > category/product url not updating.

@dpatil-magento We have determined that this is working as expected. In order to get the suffix from the GQL response, we need to know the page type, either a category, or a product type. This comes from the urlResolver. When the page doesn’t exist, the urlResolver doesn’t know if it’s a category or a product type, so we won’t be able to provide the correct suffix. It works with the back button, however.

anthoula:
> > Navigating fast between store and clicking product in category page….hits error as it still has old url link
> So if you're on a category page for View A and switch to View B and quickly click a product image to go to the product page, you'll be navigated, using an invalid link, to a product page with `.html` suffixes.

@sirugh @dpatil-magento - A third solution proposed is to create a top-level `apollo-cache-persist` for each store code, so when switching stores, it will use the cache for that specific store code. This should resolve the stale data issue when clicking quickly after a store switch.

sirugh (dismissed):
Great work!

dpatil-magento:
@anthoula Changes look good. Just found an edge case and leave call to @tkacheva if this needs to be fixed - 

when a url suffix string  is also part of category/product/store_url_code and if user switch store view which has different url suffix …then it replace first string from url.

Example -
User on URL -` /test-cat1/cat-level2/cat-level-3test?page=1`  and switch store view whose URL suffix is ` .html
`
URL Becomes - `/.html-cat1/cat-level2/cat-level-3test?page=1` and user gets 404 page not of error

Instead of - /test-cat1/cat-level2/cat-level-3.html?page=1

sirugh (approved):

PR #2902, opened 2020-12-07
[feature] Use root category id from gql

sirugh:
## Description

Venia Concept was hardcoded to use root category "2" which is the id of the root category that Magento configures by default but not necessarily the category that a client will have configured. Obtaining the actual root category id for a store from the `storeConfig` query is just one more step in the right direction of a seamless implementation of PWA Studio.

## Related Issue

Closes [PWA_1067](https://jira.corp.magento.com/browse/PWA-1067).

### Verification Steps

1. Configure two 'Stores' with different root categories
2. Using the store switcher, switch between the stores and verify the left nav shows the proper categories.

## Checklist

### DIFF
--- a/packages/peregrine/lib/store/reducers/catalog.js
+++ b/packages/peregrine/lib/store/reducers/catalog.js
@@ -18,8 +18,7 @@ const initialState = {
     categories: {},
     currentPage: 1,
     pageSize: 6,
-    prevPageTotal: null,
-    rootCategoryId: 2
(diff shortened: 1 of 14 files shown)


eug123 (approved):

dpatil-magento:
QA Approved

eug123 (approved):

dpatil-magento (approved):

dpatil-magento (approved):

PR #2904, opened 2020-12-08
Add middleware to schedule sign-out

jimbo:
## Description
When a user signs in, we get a token that we store in both app state (Redux store's `user` slice) and local storage (`signin_token`, namespaced). This token has a limited lifetime before it becomes invalid (currently hardcoded to **one hour**), and an invalid token is removed from storage the first time it is accessed after expiration.

Unfortunately, the storage interface can't notify app state when the token expires or when it removes an expired token, since it doesn't have access to the `dispatch` function. This leads app state and storage to fall out of sync, such that components start to throw errors when app state tells them the user is signed in but the token has actually been invalidated.

This PR adds a middleware that schedules the app state's `signOut` workflow to run whenever a token is set (sign-in or re-auth). Just as the token expires, the app will sign the user out, clear the token and all user state, and refresh the page. In the future, we can remove the page refresh, but that will require some amount of component rework and regression.

## Related Issue
PWA-1105

### Verification Stakeholders
- @dpatil-magento 
- @sirugh 

### Verification Steps
1. Sign in to a user account.
2. Open developer tools and edit the `signin_token` in storage to have a `ttl` of `30000` (30 seconds).
3. Refresh your browser tab (to simulate having waited approximately one hour).
4. Wait 30 seconds.
5. Verify that you are automatically signed out and the page automatically refreshes.
6. Verify that there are no JS errors in the console.
7. Repeat this process on the checkout page.

## Checklist


packages/peregrine/lib/store/middleware/auth.js
```diff
@@ -0,0 +1,60 @@
+import BrowserPersistence from '../../util/simplePersistence';
+import userActions, { signOut } from '../actions/user';
+
+const timers = new Map();
+const storage = new BrowserPersistence();
+const SET_TOKEN = userActions.setToken.toString();
+const CLEAR_TOKEN = userActions.clearToken.toString();
+const GET_DETAILS = userActions.getDetails.request.toString();
+
+const isSigningIn = type => type === SET_TOKEN || type === GET_DETAILS;
```
sirugh, line 10: Why `|| GET_DETAILS`? Is this a safeguard against tokens invalidating between the token set and the details fetch?

packages/peregrine/lib/store/middleware/auth.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * This function adheres to Redux's middleware pattern.
+ *
+ * @param {Store} store The store to augment.
+ * @returns {Function}
+ */
+const scheduleSignOut = store => next => action => {
+    const { dispatch } = store;
+
+    if (isSigningIn(action.type)) {
```
sirugh, line 22: So when we sign in, we create a timer to remove the token based on the TTL of the token just set? So this is similar to what [getItem](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/util/simplePersistence.js#L33-L45) does but is also able to dispatch the `signOut` action and clear the token from redux (persisted as `M2_VENIA_BROWSER_PERSISTENCE__signin_token`).

The current iteration of the auth link also uses the [`getItem` mechanism](https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/src/index.js#L33) which should delete the token if expired.

Both seem to handle similar situations related to client-side invalidation of the token, but I still don't know if we gracefully handle a user token that invalidates server-side before the client-side expiry.

packages/peregrine/lib/store/middleware/auth.js
jimbo: When a user starts up the application, if they have a valid token in storage, we sign them in. Since they already have a token, a `SET_TOKEN` action is never dispatched. Instead, the first user action is `GET_DETAILS`.

packages/peregrine/lib/store/middleware/auth.js
jimbo: Correct, `getItem` can't dispatch `signOut`, so the Redux store contains stale state (`isSignedIn: true`) when there's an expired token or no token in storage.

This PR doesn't handle a server-side invalidation, but I believe those are going to be much less frequent. Expiration should be the 99% case.

packages/peregrine/lib/store/middleware/auth.js
sirugh: So it's really "is signing in _or_ is getting user details". And in both cases we set a timer based on the time remaining on the TTL. Got it.

packages/peregrine/lib/store/middleware/auth.js
sirugh: >  This PR doesn't handle a server-side invalidation, but I believe those are going to be much less frequent. Scheduled expiration should be the 99% case.

Right, Tommy's solution gets us there, assuming the error messages are correct. That said, I don't know enough about auth invalidation so I'll trust that you when you say that a timeout expiry is the most common scenario.

sirugh (changes requested):
We need to solve for the cache clearing just like we do in the other sign out handlers. :(

jimbo:
> We need to solve for the cache clearing just like we do in the other sign out handlers. :(

The `signOut` async action now calls the other cache clearing functions. 👍

packages/peregrine/lib/context/cart.js
```diff
@@ -55,20 +55,16 @@ const CartContextProvider = props => {
...
     const [fetchCartId] = useMutation(CREATE_CART_MUTATION);
     const fetchCartDetails = useAwaitQuery(CART_DETAILS_QUERY);
 
     useEffect(() => {
-        // cartApi.getCartDetails initializes the cart if there isn't one. Also, we pass
-        // apolloClient to wipe the store in event of auth token expiry which
-        // will only happen if the user refreshes.
+        // cartApi.getCartDetails initializes the cart if there isn't one.
         cartApi.getCartDetails({
-            apolloClient,
```
jimbo, line 67: Passing things that can't be serialized, such as functions or objects containing functions, in action payloads is an anti-pattern in Redux. I understand the thought behind putting `apolloClient` in the payload, but we shouldn't have been doing this.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -317,9 +317,9 @@ export const removeItemFromCart = payload => {
 };
 
 export const getCartDetails = payload => {
-    const { apolloClient, fetchCartId, fetchCartDetails } = payload;
+    const { fetchCartId, fetchCartDetails } = payload;
 
-    return async function thunk(dispatch, getState) {
+    return async function thunk(dispatch, getState, { apolloClient }) {
```
jimbo, line 322: Now that our thunk middleware includes an `extraArgument` object with the Apollo client, we can access it from any thunk.

packages/peregrine/lib/store/actions/user/asyncActions.js
```diff
@@ -23,6 +25,8 @@ export const signOut = (payload = {}) =>
         await dispatch(clearToken());
         await dispatch(actions.reset());
         await clearCheckoutDataFromStorage();
+        await clearCartDataFromCache(apolloClient);
+        await clearCustomerDataFromCache(apolloClient);
```
jimbo, line 29: Here, we clear the rest of the cache data.

packages/peregrine/lib/store/middleware/thunk.js
```diff
@@ -0,0 +1,4 @@
+import thunk from 'redux-thunk';
+
+export const extraArgument = {};
+export default thunk.withExtraArgument(extraArgument);
```
jimbo, line 4: Fortunately for us, `redux-thunk` has already considered our exact situation and [provides a method](https://github.com/reduxjs/redux-thunk#injecting-a-custom-argument) that makes it easy to provide our API client to every thunk.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -46,44 +47,37 @@ const storage = new BrowserPersistence();
...
-        storage: window.localStorage,
-        debug: process.env.NODE_ENV === 'development'
-    });
-
-    let apolloClient;
-    if (apollo.client) {
-        apolloClient = apollo.client;
-    } else {
-        apolloClient = new ApolloClient({
+    const apolloClient = useMemo(() => {
```
jimbo, line 52: **This should always have been memoized.** Performing side effects in the body of a render function (component) has always been an anti-pattern; here, the original code was creating multiple Apollo clients as `Adapter` is executed several times during initial rendering.

An easy mistake to make, but a very important one to avoid.

packages/venia-ui/lib/drivers/adapter.js
```diff
@@ -46,44 +47,37 @@ const storage = new BrowserPersistence();
...
-            }
+        if (initialized) return;
+
+        // immediately invoke this async function
+        (async () => {
+            // restore persisted data to the Apollo cache
+            await apolloClient.persistor.restore();
+
+            // attach the Apollo client to the Redux store
+            attachClient(apolloClient);
```
jimbo, line 75: This is the only real change to this file. Ideally, we would create the Apollo client at the same time we create the Redux store and provide it directly to the middleware, but since we've delegated client creation to `Adapter`, we have to wait and add it here.

packages/peregrine/lib/store/actions/user/asyncActions.js
sirugh: Beautiful - plan on applying this to the other actions like [cart](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/store/actions/cart/asyncActions.js#L373)?

jimbo: It looks like maybe we have to add it to the `signIn` async action as well in order to remove it from other handlers.

packages/peregrine/lib/store/actions/cart/asyncActions.js
```diff
@@ -367,13 +366,7 @@ export const getCartDetails = payload => {
                     await dispatch(removeCart());
                 }
 
-                // Clear the cart data from apollo client if we get here and
-                // have an apolloClient.
-                if (apolloClient) {
-                    await clearCartDataFromCache(apolloClient);
```
sirugh, line 373: Why did we remove this?

sirugh (dismissed):
Excellent work on this solution. And nice find on the adapter bug.

I do have concerns about alerting the user that they are being signed out, but I believe we are going to handle that detail in another ticket. The main point of this effort was to not leave the user in a stuck state where their token was invalid but not cleared.

dpatil-magento:
@jimbo Pls check 2 observations - 
1. Even if user is actively working is session, they would be automatically logged out after 60mins? (I tested it by changing ttl to 100 secs).
2. On Mobile I kept session open for 60mins and dont see automatic signout but on refresh it did signout.

dpatil-magento:
@jimbo Looks issue is happening on desktop browser as well, if you leave idle for an hour then auto sign-out does not happen unless new call is made.

<a href="https://gyazo.com/7c64b74b874120f9928c52a28c8c0285">Image from Gyazo</a>

sirugh (approved):

dpatil-magento:
QA Approved.

PR #2910, opened 2020-12-17
[feature]: Debugging Reporter

sirugh:
## Description

This feature/hackathon project should make it easier for folks to provide debugging info about their Magento/PWA Studio system.

I was not able to access extension or target usage info. I wasn't sure what the optimal way to do this was. Perhaps there is some tooling we have already written that exposes it.

## Related Issue

Closes [PWA-1091](https://jira.corp.magento.com/browse/PWA-1091).

### Verification Steps

1. Scaffold a new app using this branch
    `DEBUG_PROJECT_CREATION=1 ../pwa-studio/packages/create-pwa/bin/create-pwa`
2. In the scaffold, `yarn install` and `yarn build:report`
3. See output!

[Image from Gyazo](https://gyazo.com/960b390007953778d67d298bd27b4886)

## Checklist


packages/venia-ui/upward.yml
```diff
@@ -12,7 +12,7 @@ veniaResponse:
         # handled by the top-level 'veniaProxy' object, which is a ProxyResolver
         # that passes the request through to the backing Magento server.
         - matches: request.url.pathname
-          pattern: '^/(graphql|rest|media)(/|$)'
+          pattern: '^/(magento_version|graphql|rest|media)(/|$)'
```
sirugh, line 15: @tjwiebell suggested that this may not be a good idea: https://magento.slack.com/archives/G8N21AQTV/p1608225831014200?thread_ts=1608224666.012800&cid=G8N21AQTV

revanth0212:
Love it ❤️ 

Can you also add a feature where the whole output is copied to the clipboard so people can simply paste it into the issue?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -19,44 +19,54 @@ Feel free to remove this section before creating this issue.
...
- - Device [e.g. iPhone6, PC, Mac, Pixel3]:
- - OS [e.g. iOS8.1, Windows 10]:
- - Browser [e.g. Chrome, Safari]:
- - Browser Version [e.g. 22]:
- - Magento Version:
- - PWA Studio Version:
- - NPM version `npm -v`:
- - Node Version `node -v`:
+
+-   Device [e.g. iPhone6, PC, Mac, Pixel3]:
```
revanth0212, line 52: Instead of asking them to fill out all this info, it makes their life easier if they can simply paste the output of the `yarn build:report` command. Might increase adoption.

sirugh: I added a bit of text to the "additional context" section. The rest of this was just auto-formatting.

sirugh:
> Can you also add a feature where the whole output is copied to the clipboard so people can simply paste it into the issue?

Copy commands differ across operating systems. I figured for now it's easy enough to just ask folks to copy the output of the command into the issue when creating one.

.github/ISSUE_TEMPLATE/bug_report.md
revanth0212: True, I understand this is what we had before, I am trying to see how we can increase the adoption of this amazing tool. I don't think we often see issues with all this info provided. If we can simply make the tool copy the output to the clipboard and they can paste it here, it will make ours and their life easier.

It is a suggestion, I am fine with the current approach. Ill leave it to you.

sirugh: I'll remove the stuff that the generator should include.

revanth0212:
> > Can you also add a feature where the whole output is copied to the clipboard so people can simply paste it into the issue?
> 
> Copy commands differ across operating systems. I figured for now it's easy enough to just ask folks to copy the output of the command into the issue when creating one.

Ahhh got it. Makes sense.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -19,44 +19,54 @@ Feel free to remove this section before creating this issue.
...
 A clear and concise description of what you expected to happen.
 
 **Screenshots**
+
 If applicable, add screenshots to help explain your problem.
 
 **Additional context**
-Add any other context about the problem here.
+
+Paste the output of `yarn build:report`, and any other context about the problem, here.
```
revanth0212: ❤️

packages/pwa-buildpack/lib/cli/generate-build-report.js
```diff
@@ -0,0 +1,213 @@
...
+    // } catch (e) {
+    //     prettyLogger.warn(
+    //         'Unable to determine Magento version - ensure that the Magento_Version route is enabled when generating this report.'
+    //     );
+    //     prettyLogger.error('Reason:', e);
+    // }
+}
+
+function inspectBuildEnv() {
+    const osVersion = os.version();
```
revanth0212, line 173: We need to be careful of this. `os.version` is added in node 13.

Failed in my local machine since I was running node 10. Do we recommend a minimum version of node for using PWA Studio?

image

revanth0212 (dismissed):

packages/pwa-buildpack/lib/cli/generate-build-report.js
```diff
@@ -101,19 +101,10 @@ function getYarnDependencies(dependencies) {
...
-            input: listbuffer.stdout
-        });
-
-        // TODO: handle multiple versions installed when not dupe. ie "npm ls core-js"
-        const [, , version] = grepBuffer.stdout
-            .toString()
-            .split('\n') // ["│ └── @magento/upward-js@5.0.0  deduped", "└── @magento/upward-js@5.0.0"]
-            .filter(text => !text.includes('deduped'))[0] // ["└── @magento/upward-js@5.0.0"]
-            .split('@'); // [ '└── ', 'magento/upward-js', '5.0.0' ]
+        const version = lockFile.dependencies[packageName].version;
```
sirugh, line 107: Ha this is so much simpler. Does this handle multiple versions of a dependency such as `core-js`? I know we don't necessarily need to worry about that since we're just checking Magento packages right now, but it was something I didn't figure out before with the TODO.

packages/pwa-buildpack/lib/cli/generate-build-report.js
sirugh: Oops, you're right! Our monorepo [package.json](https://github.com/magento/pwa-studio/blob/develop/package.json#L88-L91) requires at least 10, but if this doesn't work in 10 we need a fall back (upgrading to minimum 12 is definitely out of scope).

packages/pwa-buildpack/lib/cli/generate-build-report.js
```diff
@@ -0,0 +1,213 @@
...
+    //     prettyLogger.warn(
+    //         'Unable to determine Magento version - ensure that the Magento_Version route is enabled when generating this report.'
+    //     );
+    //     prettyLogger.error('Reason:', e);
+    // }
+}
+
+function inspectBuildEnv() {
+    const osVersion = os.version();
+    const nodeVersion = process.version();
```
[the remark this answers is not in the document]
sirugh: Another good catch. I wonder how I got this to work. Maybe I had it written when I ran it for the screenshot and forgot to push? 🤷 Thanks for fixing :)

revanth0212:
Using `yarnpkg/lockfile` package to parse `yarn.lock`. This significantly reduced the time taken to generate the list of dependencies and versions. Done in 2 seconds compared to 14 seconds.

image

revanth0212:
Changed the NPM and YARN dependency generation logic. Speeds have improved quite a bit.

Left is YARN and right is NPM.

image

packages/pwa-buildpack/lib/cli/generate-build-report.js
revanth0212: Nice question dude. I did check for that and added a loop construct in the YARN version of this function to handle multiple versions. Coming to NPM it is handled a little differently. package-lock.json is only holding on to 1 version of the package.

sirugh: > package-lock.json is only holding on to 1 version of the package.

Are you sure? If you do `npm ls core-js` it'll show two resolved versions.

sirugh:
Awesome! Normally wouldn't want to add dependencies unless necessary or vetted but in this case it is

a) used in a build tool so won't bloat the bundle and
b) published by the yarn maintainers and not some random third party

Good find :)

revanth0212:
> Awesome! Normally wouldn't want to add dependencies unless necessary or vetted but in this case it is
> 
> a) used in a build tool so won't bloat the bundle and
> b) published by the yarn maintainers and not some random third party
> 
> Good find :)

Well put 👍 .

packages/pwa-buildpack/lib/cli/generate-build-report.js
sirugh: @revanth0212 this is from a scaffolded app running `npm ls core-js`.

```
~/code/v9test
λ npm ls core-js
v9test@0.0.1 /Users/stephen/code/v9test
├─┬ @storybook/react@5.2.8
│ ├─┬ @storybook/addons@5.2.8
│ │ ├─┬ @storybook/api@5.2.8
│ │ │ └── core-js@3.8.2  deduped
│ │ ├─┬ @storybook/channels@5.2.8
│ │ │ └── core-js@3.8.2  deduped
│ │ ├─┬ @storybook/client-logger@5.2.8
│ │ │ └── core-js@3.8.2  deduped
│ │ ├─┬ @storybook/core-events@5.2.8
│ │ │ └── core-js@3.8.2  deduped
│ │ └── core-js@3.8.2  deduped
│ ├─┬ @storybook/core@5.2.8
│ │ ├─┬ @storybook/channel-postmessage@5.2.8
│ │ │ └── core-js@3.8.2  deduped
│ │ ├─┬ @storybook/client-api@5.2.8
│ │ │ └── core-js@3.8.2  deduped
│ │ ├─┬ @storybook/router@5.2.8
│ │ │ └── core-js@3.8.2  deduped
│ │ ├─┬ @storybook/theming@5.2.8
│ │ │ └── core-js@3.8.2  deduped
│ │ ├─┬ @storybook/ui@5.2.8
│ │ │ ├─┬ @storybook/components@5.2.8
│ │ │ │ ├── core-js@3.8.2  deduped
│ │ │ │ └─┬ simplebar-react@1.2.3
│ │ │ │   └─┬ simplebar@4.2.3
│ │ │ │     └── core-js@3.8.2  deduped
│ │ │ └── core-js@3.8.2  deduped
│ │ ├── core-js@3.8.2  deduped
│ │ └─┬ lazy-universal-dotenv@3.0.1
│ │   └── core-js@3.8.2  deduped
│ ├─┬ @storybook/node-logger@5.2.8
│ │ └── core-js@3.8.2  deduped
│ └── core-js@3.8.2
└─┬ graphql-cli@3.0.14
  ├─┬ apollo-codegen@0.20.2
  │ ├─┬ apollo-codegen-core@0.20.1
  │ │ └── core-js@2.6.12
  │ └─┬ apollo-codegen-flow@0.20.0
  │   └─┬ apollo-codegen-core@0.20.1
  │     └── core-js@2.6.12
  └─┬ tmp-graphql-config-extension-openapi@1.0.7
    └─┬ @kbrandwijk/swagger-to-graphql@2.4.3
      ├─┬ babel-runtime@6.26.0
      │ └── core-js@2.6.12
      └─┬ json-schema-ref-parser@3.3.1
        └─┬ z-schema@3.25.1
          └── core-js@2.6.12
```

revanth0212: You are right about `npm ls`. But there is a small caviate, It is listing all the packages that `require` the package, not the list of installed versions.

For instance:
image

At the moment there is only 1 version of it present in the `lock` file.
image

On the other hand, there can be packages that require a diff version of the package, for those, the package will be listed in the dependencies section of those packages.
image

After combing through the lock file for `core-js` I realized we have to look for all the packages that have our package as a dependency, which means we have to dig deeper than we are doing at the moment. I will work on that push the changes in the next commits.

Thanks for pointing this out. Good work buddy.

packages/pwa-buildpack/lib/cli/generate-build-report.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+};
+
+const cwd = process.cwd();
+
+function inspectDependencies(package) {
+    prettyLogger.info('Inspecting Dependencies');
+
+    // Inspect all '@magento/...' dependencies.
+    const magentoDependencies = [
```
revanth0212, line 47: @sirugh while writing tests for this file, it hit me, we should be capturing all `adobe` and `magento` packages right, not just `magento`. What do you think?

sirugh: Depends on what you really want to glean from this information. I think we have at least one adobe dependency so might as well throw that in :)

revanth0212: We actually use two:

@adobe/adobe-client-data-layer
@adobe/apollo-link-mutation-queue

sirugh:
👍 on Revanth's commits, but as this is my PR, github won't allow me to approve.

tjwiebell (dismissed):
Approving on behalf of @sirugh 🙂

jimbo (dismissed):

revanth0212:
Not sure why the test job failed on AWS. Works fine in my local. Gonna investigate the logs for more info.

image

revanth0212:
> Not sure why the test job failed on AWS. Works fine in my local. Gonna investigate the logs for more info.
> 
> image

Found it

image

Our CICD must be on a version of Node that does not support `os.version` API. We have documented this in the code but not implemented for it in tests. My bad. Ill update the tests to handle this.

sirugh (commented):
Approved again - we skip the `os.version` check if we aren't on >= node 13.

jimbo (approved):

dpatil-magento:
@sirugh @revanth0212 Have some queries with respect this report tool code dependency with rest of pwa-studio.
1. Any reason it can't be an extension? Was expecting this to work both on cloned/forked repo and scaffolded project  on need basis and not by default, something like  `yarn venia add -D @magento/venia-sample-language-packs@0.0.1` and then `yarn build:report`
2. Hope  it does not interfere in default `yarn install` and `yarn build`.

sirugh:
>Any reason it can't be an extension?

Haha no, but it is easier to make it a utility of buildpack than a random extension. Also the initial ticket/request suggested it be a buildpack utility, which is why I went that direction.

Stop trying to throw in scope @dpatil-magento!

dpatil-magento:
QA Approved.

PR #2911, opened 2020-12-18
Add props to nav query to render root category links

sirugh:
## Description

We lost the `All {TopCategory}` links at the top of the left nav, after clicking into a sub category that had child categories.
This is because we missed a few props from the original query when we [migrated them to peregrine](https://github.com/magento/pwa-studio/pull/2712/files#diff-a2d1a2e334760de1342ca77d4c3129673dc3382e370203990046e4fe1b050912).

Easy fix :)

The props added are used by [category tree](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/CategoryTree/useCategoryTree.js#L62-L64) and [category leaf](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CategoryTree/categoryLeaf.js#L12).

## Related Issue

Closes [PWA_1025](https://jira.corp.magento.com/browse/PWA-1025).

### Verification Steps

1. Open app
2. Open left nav
3. Click bottoms
4. See "All Bottoms" at the top of left nav.

## Checklist


tjwiebell (approved):
The dangers of that migration PR staying open so long, thought something like this might slip by 😿  nice work on the fix!

dpatil-magento:
QA Approved.

PR #2912, opened 2020-12-18
PWA-1074: Fix relative import for local custom loader

sirugh:
## Description

Storybook broke...again. This time it was because we introduced a [custom loader](https://github.com/magento/pwa-studio/blob/develop/packages/pwa-buildpack/lib/WebpackTools/configureWebpack/getModuleRules.js#L59), which seemed to be auto-loaded things like watch/build but not for things like storybook. This PR fixes that by making the loader import a relative one, using the path within buildpack.

## Related Issue

Closes [PWA_1074](https://jira.corp.magento.com/browse/PWA-1074).

### Verification Steps

1. `yarn && yarn storybook:venia`. Should work.
2. Scaffold an app locally and in the new app `yarn && yarn storybook`. Should work. FWIW I had to copy the `getModuleRules` file to the node_modules of the scaffold as the local file wasn't used when I scaffolded for some reason. Maybe an issue with publishing? idk.

## Checklist


tjwiebell (approved):
Fix works as described 👍  I additionally checked it on a scaffolded app with success (was a bit nervous about the relative path, but works fine).

dpatil-magento:
QA Approved

sirugh:
@tjwiebell glad you checked it :) I put it in the verification steps too.

PR #2916, opened 2020-12-21
Order History Page Search

revanth0212:
## Description

Added feature to search by order number on the order history page.

## Related Issue
Closes [PWA-888](https://jira.corp.magento.com/browse/PWA-888)

## Acceptance
Should be able to search by order number on the order history page.

### Verification Stakeholders
@schensley 
@dpatil-magento 

### Verification Steps
1. Go to the deployed PR's order history page.
2. Enter a valid order number in the search field. You should be able to see the order details of the given order number.
3. Enter an invalid number and you should see a message stating there are no orders yet.

image

image

image

## Checklist


packages/venia-ui/lib/components/OrderHistoryPage/index.js
```diff
@@ -1 +1,9 @@
-export { default } from './orderHistoryPage';
+import React from 'react';
+import { Form } from 'informed';
+import OrderHistoryPage from './orderHistoryPage';
+
+export default props => (
```
revanth0212, line 5: I wish we had a `wrapWithForm` HOC for such cases. A simple HOC that just wraps the given component inside an `informed form`.

schensley:
@revanth0212 this looks good. thanks for working with me on the details. UX approved.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.js
```diff
@@ -30,31 +64,96 @@ const OrderHistoryPage = props => {
...
+    if (isLoadingWithoutData) {
+        return fullPageLoadingIndicator;
+    }
+
     return (
         <OrderHistoryContextProvider>
             <div className={classes.root}>
                 <Title>{title}</Title>
                 <h1 className={classes.heading}>{PAGE_TITLE}</h1>
+                <div className={classes.search}>
```
tjwiebell, line 139: Something like this:
```suggestion
                <Form onSubmit={handleSubmit} className={classes.search}>
```

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.js
```diff
@@ -30,31 +64,96 @@ const OrderHistoryPage = props => {
...
+                    <TextInput
+                        after={resetButton}
+                        before={searchIcon}
+                        field="search"
+                        id={classes.search}
+                        placeholder={SEARCH_PLACE_HOLDER}
+                        onKeyPress={handleKeyPress}
```
tjwiebell, line 146:
```suggestion
```

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.js
```diff
@@ -30,31 +64,96 @@ const OrderHistoryPage = props => {
...
+                    />
+                    <Button
+                        disabled={isBackgroundLoading || isLoadingWithoutData}
+                        priority={'high'}
+                        onClick={getOrderDetails}
```
tjwiebell, line 151:
```suggestion
```

tjwiebell (changes requested):
Clarified what we discussed in Slack a bit more, I think this can be simplified a bit further. Ping me if you have questions 🙂

packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
```diff
@@ -19,21 +24,75 @@ export const useOrderHistoryPage = (props = {}) => {
...
+    const formApi = useFormApi();
+
+    const searchText = useMemo(() => {
+        return formState.values.search;
+    }, [formState]);
+
+    const [
+        getOrdersData,
+        { loading: ordersLoading, error: getOrdersError }
+    ] = useLazyQuery(getCustomerOrdersQuery, {
```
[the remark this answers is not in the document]
revanth0212: Right.. good suggestion @tjwiebell. I have pushed the changes to use a single query with fuzzy search. cc1b52c

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.js
revanth0212: Nice idea @tjwiebell, Unfortunately, the problem with this solution is the talon won't have access to the `searchText`. The talon needs `searchText` and so does the component. We do some if-else checking in the component that needs the `searchText` from the talon.

tjwiebell: I've encountered the chicken/egg issue with form context and using those hooks from the parent talon, so understand what you mean. I'll try my suggestion in scope of the pagination ticket, and won't hold this one up.

tjwiebell (dismissed):
Looks good, nice work 👍  would prefer to get rid of those event handlers if possible, but understand the limitation, so won't hold this up.

dpatil-magento:
@revanth0212 Please check following - 
1. When there are no (zero) orders, search field is still displaying > Dont think we need it there. Not a blocker, check with UX if update needed.

image


2. When I clear order number using keyboard then there is no way for user to load all orders. Actually all order should load if user clear the field and does not start typing new search string.

<a href="https://gyazo.com/30e6692b9128e54888965b26b026e63c">Image from Gyazo</a>

revanth0212:
> @revanth0212 Please check following -
> 
> 1. When there are no (zero) orders, search field is still displaying > Dont think we need it there. Not a blocker, check with UX if update needed.
> 
> image
> 
> 1. When I clear order number using keyboard then there is no way for user to load all orders. Actually all order should load if user clear the field and does not start typing new search string.
> 
> [Image from Gyazo](https://gyazo.com/30e6692b9128e54888965b26b026e63c)

Very good point @dpatil-magento. In my opinion, it makes sense to leave the search field in place even if the orders list is empty. @schensley what do you think is the right approach in this case.

Regarding 2, I have handled it in the latest commits. f229636 0d71cc4

dpatil-magento:
QA Approved.

schensley:
@revanth0212 @dpatil-magento I think it's good to leave the search box even when there are no orders. It would be ideal to have a message along the lines of "You have no orders" (or the like) where the orders would/will appear.

tjwiebell (approved):

PR #2922, opened 2020-12-27
Move category params query to handle by CategoryTree fragment

huykon:
## Description

Based on [28584](https://github.com/magento/magento2/issues/28584). I move category params query to handle by CategoryTree fragment

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Go to a category page.
2. Check GraphQl category query.
3. Make sure CategoryTree fragment called by that query and check the data response.

## Checklist


sirugh:
Hi @huykon, could you explain to me the benefit of the changes in this PR? It looks like the [linked fix](https://github.com/magento/magento2/issues/28584) allows you to use fragments _if you want_ but doesn't require you to use them.

huykon:
@sirugh this will help the code more neatly. I think so

sirugh:
I'm sorry but I disagree. There are times when using an inline fragment may be necessary or helpful but I don't think there is any benefit here since the type you are using for the fragment **is** the type [returned by the query](https://devdocs.magento.com/guides/v2.4/graphql/queries/category.html#Categories). 

An example of where this sort of optimization might be useful would be if the `Breadcrumbs` component simply took `data`, then any page that wants to display breadcrumbs would make a use of the fragment below, so that they fetch data from [CategoryInterface](https://devdocs.magento.com/guides/v2.4/graphql/interfaces/category-interface.html)

```gql
fragment BreadcrumbData on CategoryInterface {
  category_id
  category_level	
  category_name	
  category_url_path
}
```

sirugh:
I'm going to close this PR but I think there are some places we could apply a similar treatment to the app where it might be helpful. Ping me on slack and we can start up a thread or github issue with the details :)

PR #2928, opened 2021-01-05
[PWA-887] My Account: Order History - Pagination

tjwiebell:
## Description

As a customer with lots of Orders (20+), I would like pagination (struggling to come up with a story here, sorry!). This will allow me to view my most recent Orders without waiting for a large network request to come back with all of my orders.

Technical Notes:
pageSize defaults to 20 and can be adjusted
Response fields include total_count and page info (current page, page size, total pages)
I don't see any ability to sort the orders response, appears to be by order number/date

After some thinking on this, I propose a simple UI for the Shoppers where if there are >10 orders in their Order History the page will display the first 10 orders (most recent to oldest) and a "Load More" button will appear below the list of orders. The Shopper can then tap this to load the next 10 orders. This will be the the behavior until all orders have been loaded.  

## Related Issue

* [[PWA-887](https://jira.corp.magento.com/browse/PWA-887)] My Account: Order History - Pagination

### Verification Steps

Test Plan : (In below considered 10 as pagination point)

**Verify Pagination**
1. Does not display if there are 0 orders
2. Does not display if there are 10 orders
3. Does display if there are 11 orders.

**Verify if user hit Load More**
1. When there are than 11 orders then next 1 order should display and Load More button should no more exist.
2. When there are 20 orders next 10 order should display and Load More button should no more exist.
3. When there are 21 orders next 10 order should display and Load More button should exists.

1. Verify Load more button functionality for page with 50 or more orders.
2. Verify above scenarios for Desktop Chrome, FF, Safari and Android/iOS devices.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
+++ b/packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
@@ -1,7 +1,6 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useHistory } from 'react-router-dom';
-import { useFormState, useFormApi } from 'informed';
-import { useLazyQuery } from '@apollo/client';
+import { useQuery } from '@apollo/client';
 
 import mergeOperations from '@magento/peregrine/lib/util/shallowMerge';
 
@@ -15,7 +14,6 @@ export const useOrderHistoryPage = (props = {}) => {
     const operations = mergeOperations(DEFAULT_OPERATIONS, props.operations);
     const { getCustomerOrderQuery } = operations;
 
-    const [orders, setOrders] = useState([]);
     const [
         ,
         {
@@ -24,35 +22,25 @@ export const useOrderHistoryPage = (props = {}) => {
     ] = useAppContext();
     const history = useHistory();
     const [{ isSignedIn }] = useUserContext();
-    const formState = useFormState();
-    const formApi = useFormApi();
 
-    const searchText = useMemo(() => {
-        return formState.values.search;
-    }, [formState]);
+    const [searchText, setSearchText] = useState('');
 
-    const [
-        getOrderData,
-        { loading: orderLoading, error: getOrderError }
-    ] = useLazyQuery(getCustomerOrderQuery, {
+    const {
+        data: orderData,
+        loading: orderLoading,
+        error: getOrderError
+    } = useQuery(getCustomerOrderQuery, {
         fetchPolicy: 'cache-and-network',
-        nextFetchPolicy: 'cache-first',
-        onCompleted: orderData => {
-            setOrders(orderData.customer.orders.items);
+        variables: {
+            orderNumber: {
+                number: {
+                    match: searchText
+                }
+            }
         }
     });
 
-    const getOrdersData = useCallback(() => {
-        getOrderData({
-            variables: {
-                orderNumber: {
-                    number: {
-                        match: ''
-                    }
-                }
-            }
-        });
-    }, [getOrderData]);
+    const orders = orderData ? orderData.customer.orders.items : [];
 
     const isLoadingWithoutData = !orders && orderLoading;
     const isBackgroundLoading = !!orders && orderLoading;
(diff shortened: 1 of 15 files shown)


schensley:
@tjwiebell this behaves as expected and looks great.  thanks for adding in the "showing" text. UX approved

packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
```diff
@@ -24,76 +24,69 @@ export const useOrderHistoryPage = (props = {}) => {
...
+    }, []);
 
-            formApi.reset();
-            getOrdersData();
-        },
-        [formApi, getOrdersData]
-    );
+    const loadMoreOrders = useMemo(() => {
+        if (orderData) {
+            const { page_info } = orderData.customer.orders;
```
supernova-at, line 80: I'm a little fuzzy on the difference between this `page_info` and the one locally defined (L52).

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.css
```diff
@@ -20,13 +20,21 @@
...
 }
 
+.filterRow {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+}
+
+.pageInfo {
+    font-size: 0.875rem;
```
supernova-at, line 30: Is there a token we can use for this? Seems like we'd want consistency across the app for these.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.js
```diff
@@ -115,6 +118,28 @@ const OrderHistoryPage = props => {
...
 
+    const pageInfoLabel = pageInfo ? (
+        <FormattedMessage
+            defaultMessage={'Showing {current} of {total}'}
+            id={'orderHistoryPage.pageInfo'}
+            values={pageInfo}
+        />
+    ) : null;
+
+    const loadMoreButton = loadMoreOrders ? (
```
supernova-at, line 129: Interesting. I was expecting to see a flag `canLoadMoreOrders` or something but this function will be `null` if you can't load more. I'm okay with this, I like that it allows for fewer props between the talon and the component and that we don't have to define a function that we know won't be called.

packages/venia-ui/lib/components/OrderHistoryPage/resetButton.js
```diff
@@ -0,0 +1,22 @@
...
+
+const clearIcon = <Icon src={ClearIcon} size={24} />;
+
+const ResetButton = props => {
+    const { onReset } = props;
+    const formApi = useFormApi();
+
+    const handleReset = () => {
+        formApi.reset();
+        onReset();
```
supernova-at, line 16: Can we add `onReset` to `propTypes` and make it optional? Would need to add a null guard around this invocation then too.

supernova-at (changes requested):
I'd like to make the `ResetButton`'s `onReset` prop be optional. Other than that, this looks great!

packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
tjwiebell: The local defined feeds the count indicator `Showing {x} of {y}` and the `page_info` data response is just used to determine if we're on the last page or not. The order this was developed in isn't obvious, but the count indicator came in last minute, and I suppose this `page_info` isn't really needed since we can derive this by `pageSize < total_count`. Multiple ways to derive, but happy to clean this up if you think it will make it easier to understand.

packages/venia-ui/lib/components/OrderHistoryPage/orderHistoryPage.css
tjwiebell: Good call, switched to token in 8c923bf.

packages/venia-ui/lib/components/OrderHistoryPage/resetButton.js
tjwiebell: Added in 8c923bf.

supernova-at (approved):

PR #2929, opened 2021-01-06
[fix] offline cached search and category pages

sirugh:
## Description

Category and Search pages were not rendering cached data when offline. This is because we rendered errors even when we had data to render otherwise.

A good example of the "correct" implementation for offline functionality would be the [Product Root Component](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/RootComponents/Product/product.js#L28) which only rendered the error when there was no product data to render.

I applied this logic to the category and search pages.

## Related Issue

Closes [PWA-1237](https://jira.corp.magento.com/browse/PWA-1237).

### Verification Steps

1. Check out this PR.
2. In your `.env`, set `DEV_SERVER_SERVICE_WORKER_ENABLED=true`.
3. Run `yarn watch:all`.
4. Load the Home Page, then a Category Page, then a Product Page. Then search for that product and remember the term you use.
5. Switch to offline mode. Click the home page. It should render.
6. Now click the Category. It should render.
7. Now click the Product. It should render.
8. Now search using the term from the prior step. It should render the search results.
9. 
## Checklist


revanth0212:
I haven't reviewed the code but looks like the issue isn't fixed totally. I can still see the error banner flicker before the cached data takes its place. Check out the demo.


https://user-images.githubusercontent.com/35203638/103809598-59754d00-501f-11eb-9522-6b5f41f95c75.mov

sirugh:
I'm not able to reproduce this issue using the [PR deployment](https://pr-2929.pwa-venia.com/). Can you recheck?

sirugh:
OK I know what the problem is here @revanth0212. The "Shop the Look" that you click on the home page is a "hard" link in the sample CMS content. It is not a `Link` component that can utilize our internal routing. 

So a click on "Shop the Look" or any other link in the home page will result in a hard navigation which causes the entire app to re-mount. When it re-mounts, we lose the [root component map](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js#L15) that was previously created in state. This map is then re-filled as you navigate (in offline mode) to pages whose data is cached in apollo. If you use the browser's forward/back buttons you should not see the flash of the message, because the root component map is there.

Try this - Use other links throughout the app, but not CMS links. Then go offline and click through the same way. You should not see the error message because you haven't hard navigated.

TL;DR - The issue you are experiencing is an artifact of us not rendering CMS links using the `Link` component. If this is even considered a bug, it is for another time.

revanth0212:
> OK I know what the problem is here @revanth0212. The "Shop the Look" that you click on the home page is a "hard" link in the sample CMS content. It is not a `Link` component that can utilize our internal routing.
> 
> So a click on "Shop the Look" or any other link in the home page will result in a hard navigation which causes the entire app to re-mount. When it re-mounts, we lose the [root component map](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js#L15) that was previously created in state. This map is then re-filled as you navigate (in offline mode) to pages whose data is cached in apollo. If you use the browser's forward/back buttons you should not see the flash of the message, because the root component map is there.
> 
> Try this - Use other links throughout the app, but not CMS links. Then go offline and click through the same way. You should not see the error message because you haven't hard navigated.
> 
> TL;DR - The issue you are experiencing is an artifact of us not rendering CMS links using the `Link` component. If this is even considered a bug, it is for another time.

Nice read. Well done with the explanation. I will continue with the PR review, meanwhile lets check with the team if they think this is a bug or if we can postpone this fix for another day.

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -75,7 +73,7 @@ const SearchPage = props => {
         );
     }
 
-    const totalCount = data ? data.products.total_count : 0;
+    const totalCount = data.products.total_count || 0;
```
revanth0212, line 76: In a previous team discussion, I remember we wanted the ternary operator instead of the logical operators in such cases. I might be wrong, ignore if so.

revanth0212 (dismissed):
Wonderful work buddy. Thanks for adding the tests as well. Other than a minor query about ternary operator vs logical operator, the code looks ready to be merged.

revanth0212:
I tried the steps you mentioned while testing and it works. Unfortunately, when I tried to refresh the page, the issue is back presumably because the whole app is being re-mounted on refresh. So your hypothesis stands.

packages/venia-ui/lib/components/SearchPage/searchPage.js
sirugh: The ternary is unnecessary because `data` must exist for execution to reach here.

supernova-at (approved):

dpatil-magento:
@sirugh Category Pagination does not work in offline mode. Rest areas look good though.

supernova-at (approved):

dpatil-magento:
QA Approved

PR #2936, opened 2021-01-11
[feature]: ErrorView

sirugh:
## Description

Goal: A generic error view component that handles some basic styling. Should take props allowing for customization, and provide defaults.

Also in this PR:
* Cleanup of unused components
* Refactor of ErrorView and NotFound into single component.
* Fixed some Storybook issues, namely global styles being applied and fonts not being imported.

## Related Issue

Closes [PWA-1032](https://jira.corp.magento.com/browse/PWA-1032).

### Verification Steps

1. Check each of the updated page types and ensure functionality in non-error state.
2. On each of the page types, cause error to be thrown. Simplest way is to block specific data GQL request for that component. Observe that the error view is shown as expected, for that page. Some may have custom messages, others will show default messages.
3. Go to a non-existent route ie https://pr-2936.pwa-venia.com/foo. It should render the error view.

## Checklist


schensley:
@sirugh thanks for the work on this.  UX approved.

packages/venia-ui/lib/RootComponents/Product/product.js
```diff
@@ -7,6 +7,7 @@ import { fullPageLoadingIndicator } from '../../components/LoadingIndicator';
 import ProductFullDetail from '../../components/ProductFullDetail';
 import getUrlKey from '../../util/getUrlKey';
 import mapProduct from '../../util/mapProduct';
+import ErrorView from '@magento/venia-ui/lib/components/ErrorView';
```
revanth0212, line 10: Minor thing, I am not sure if we have this documented somewhere or if this is just an accepted thing in our code base but package imports should go above relative imports. I might be wrong, use your best guess.

revanth0212 (dismissed):
Nice work buddy. Much needed feature.

packages/venia-ui/lib/RootComponents/Product/product.js
supernova-at: Yeah, that's a convention we have. I think we've said we don't want to reject PRs on it unless there's tooling for it though.

It's a "try your best to keep it consistent" kind of thing. If we _really_ care about it we'll write and enforce some lint rules.

supernova-at:
### QA

The `UseCmsPage` talon [does return an error](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/Cms/useCmsPage.js#L68) when the GraphQL query fails, but the `CMS` component itself [never looks at that error](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/RootComponents/CMS/cms.js#L17-L22) when determining how to render.

As a result, if the `cmsPage` query fails, the homepage forever shows the page loading indicator instead of this error message.

Should this logic be fixed in this PR?

packages/venia-ui/lib/RootComponents/Product/product.js
sirugh: I'll update all these imports to use the absolute and then order based on distance from the file, as we've discussed prior.

sirugh:
> As a result, if the cmsPage query fails, the homepage forever shows the page loading indicator instead of this error message.

This is now remedied by changing `shouldShowLoadingIndicator` to be `loading && !data`, so it only shows when the loading boolean from the query hook is true. Once _something_ is returned by the network call we either render the content (`hasContent`) or we pass rendering responsibility to the child, `categoryList` component, which will now render the error view if necessary.

I opted not to handle the error state for cmsPage as that's a little out of scope of this work.

packages/venia-ui/lib/RootComponents/Product/product.js
sirugh: Updated.

packages/venia-ui/lib/components/ErrorView/errorView.js
```diff
@@ -1,17 +1,67 @@
...
-    return <div className={classes.root}>{props.children}</div>;
+    const history = useHistory();
+
+    const handleGoHome = useCallback(() => {
+        history.push('/');
+    }, [history]);
+
+    const {
+        header = DEFAULT_HEADER,
+        message = DEFAULT_MESSAGE,
```
sirugh, line 24: Add defaults to i18n

revanth0212 (dismissed):

revanth0212 (approved):

PR #2939, opened 2021-01-12
[feature]: Adding a product to Wishlist from the PDP

### LINKED ISSUE #2934: [feature]: Adding a product to Wishlist from the PDP
Issue is re-opened to unify experience across all storefront areas: catalog/CMS/PDP

### Story

**Preconditions**
* Wishlist is enabled in Magento Admin Panel>Stores > Settings > Configuration > Customers > Wish List
* User is on Venia storefront PDP
* User is not logged in

**Precondition 1 - single wishlist**
* Magento doesn't have multiple wish-list (CE) or multi-wishlist is disabled in Stores > Settings > Configuration > Customers > Wish List>Enable Multiple Wish Lists (EE)

**Precondition 2 - multiple wishlist**
* Multi-wishlist is enabled in Stores > Settings > Configuration > Customers > Wish List>Enable Multiple Wish Lists (EE)

**Acceptance Criteria**
* User sees an indication to add product to the wishlist 
* User initiates adding product to the wishlist and gets prompt to log in/create account (as adding to wish list requires authentication).
* If user is logged in, no prompt is shown, see next AC
* In case of **Precondition 1**:
  * After user successfully logged in, user sees the indication of product added to the wishlist.
* In case of **Precondition 2**:
  * After user successfully logged in, user sees a dialogue to choose a wishlist to add a product to.
  * User creates a wishlist or selects one created before
  * Once wishlist selected/created, User sees the indication of product added to the wishlist
* User sees the indication of the product to be part of the wishlist when they reload a PDP. If a configurable variation is wish listed, we consider the configurable product to be wish listed.

As a user I want to be able to:
* add to a wishlist from the product page, optionally needing to
* create a wishlist from the product page

#### Test Plan
1. Navigate to PDP
1. Favorite the product
1. Select an existing wishlist
1. Verify that the product is added to the wishlist (verify success and error handling in error scenario)
1. Navigate to separate PDP
1. Favorite the product
1. Create a new wishlist (verify error scenario)
1. Verify that the product is added to the newly created wishlist (verify error scenarios)
1. Repeat above scenario's for Configurable and Simple products.
1. Repeat above for CE and verify that when you favorite the product it is immediately added to the only wishlist (no dialog is shown)
1. Verify UI of overlay for Minimum (no saved lists) and Maximum (Many saved lists)
1. Verify Functionality of Cancel and X icon.
1. Run MFTF
1. Verify UI in different browsers and devices.

 

Question - 
1. Cart page Kebab menu also gives option for adding to Favourite lists, should that flow also be covered?
  1. Good question, I'll create a separate story for this: PWA-1495.

huykon:
## Description

Add a product to my favorites products list from the Product Details Page.

## Related Issue

Closes #2934.

### Verification Steps

1. Login as a customer on store.
2. Go to the product detail page.
3. Make sure the response successfully.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
+++ b/packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
@@ -1,6 +1,7 @@
 import { useCallback, useState, useMemo } from 'react';
 import { useMutation } from '@apollo/client';
 import { useCartContext } from '@magento/peregrine/lib/context/cart';
+import { useUserContext } from '@magento/peregrine/lib/context/user';
 
 import { appendOptionsToPayload } from '@magento/peregrine/lib/util/appendOptionsToPayload';
 import { findMatchingVariant } from '@magento/peregrine/lib/util/findMatchingProductVariant';
@@ -168,7 +169,9 @@ export const useProductFullDetail = props => {
     const {
         addConfigurableProductToCartMutation,
         addSimpleProductToCartMutation,
-        product
+        addProductToWishlistMutation,
+        product,
+        afterSubmit
     } = props;
 
     const productType = product.__typename;
@@ -178,6 +181,12 @@ export const useProductFullDetail = props => {
     );
 
     const [{ cartId }] = useCartContext();
+    const [{ isSignedIn, currentUser }] = useUserContext();
+
+    const [formApi, setFormApi] = useState();
+
+    const wishlistId =
+        isSignedIn && currentUser.wishlist && currentUser.wishlist.id;
 
     const [
         addConfigurableProductToCart,
@@ -191,6 +200,11 @@ export const useProductFullDetail = props => {
         addSimpleProductToCart,
         { error: errorAddingSimpleProduct, loading: isAddSimpleLoading }
     ] = useMutation(addSimpleProductToCartMutation);
+
+    const [
+        addProductToWishlist,
+        { error: errorAddingWishlist, loading: isAddWishlistLoading }
+    ] = useMutation(addProductToWishlistMutation);
 
     const breadcrumbCategoryId = useMemo(
         () => getBreadcrumbCategoryId(product.categories),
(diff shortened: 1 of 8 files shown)


packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -178,6 +181,12 @@ export const useProductFullDetail = props => {
     );
 
     const [{ cartId }] = useCartContext();
+    const [{ isSignedIn, currentUser }] = useUserContext();
+
+    const [formApi, setFormApi] = useState();
+
+    const wishlistId =
+        isSignedIn && currentUser.wishlist && currentUser.wishlist.id;
```
sirugh, line 189: Looking into this a bit, `wishlist` is deprecated, and you should instead use `wishlists` to get a list of available wishlist ids, and then we probably want to present the choice to the user in some form of UX.

huykon: that mean 1 user have many wishlist ids? Then how can I know exactly id to add product to wishlist?

sirugh: > that mean 1 user have many wishlist ids?

Yep! You can have more than one, at least that's how the schema reads to me.

>  Then how can I know exactly id to add product to wishlist?

This is why we would need some UX direction. I imagine "add to wishlist" would be a button that, when clicked, would turn into a dropdown/select box.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
```diff
@@ -178,6 +181,12 @@ export const useProductFullDetail = props => {
     );
 
     const [{ cartId }] = useCartContext();
+    const [{ isSignedIn, currentUser }] = useUserContext();
```
sirugh, line 184: Rather than using `currentUser` to access data, you should make queries and mutations.

In this case, you would likely use `useLazyQuery` to fetch user data if the user is signed in.

huykon: that mean each time we come to another product then call that query 1 time? why we don't use id as my code? I think it's better

sirugh: The way ApolloClient works with the cache is configurable. We have opted, in Venia, to use `cache-and-network` for first render, and `cache-first` for subsequent renders (assuming no refresh).

What this means is that the first time a user visits a page, they see data from the cache, if present. Then ApolloClient makes a network call to the GQL server to fetch up-to-date data, and replaces the older data in the cache. The result is an immediately visible page (based on cached data) which is then visually updated with any new information that may come from the back end.

Applying that example to the wishlist UX: Let's say we have a component `AddToWishlistButton` on the product page. This button/input contains a dropdown of wishlists. The first time the page loads, we make a query to get a list of the customer's wishlists to render in the button. Then if we go to a different product page, the data for that button is already cached, so the button is immediately visible. In a similar vein, if a user visits their account wishlist page we would have their wishlists already in cache, so when they visit the product page we can immediately render them (and make a true-up network call behind the scenes).

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
huykon: @sirugh will we have design for this dropdown/select box?

sirugh: Yes! I will add mocks to the linked issue asap.

packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js
huykon: That mean should I create new component `AddToWishlistButton` & that talon then how can I get quantity of form?

sirugh:
Closing draft - we would like to decompose/groom this work a bit more before we ask anyone to work on it. Please be on the lookout for that! Once we do this I will ping you, or you can just watch for the "ready for dev" label on the issue. Thanks!

PR #2941, opened 2021-01-12
[PWA-1065] Media Url generation should use store code header but have fallback to default like other store config queries

anthoula:
## Description

As a admin user, I would like the ability to change settings in the admin panel to control the Media Url per store view so that I can store media assets for different locales in different media folders, and get that respected on Venia storefront.

## Related Issue

https://jira.corp.magento.com/browse/PWA-1065

### Verification Stakeholders

@dpatil-magento 
@supernova-at 
### Verification Steps

1. User goes to Stores > Configuration  and change the scope to  2 store view
2. User navigates to Store> Configuration >General> Web> Base URL (secure) for User Media Files and sees the default setting
3. User updates the setting to use different media folder for store view 2. E.g. cdn.domain.com/store2/pub/media/
4. User creates content (page/block) with image stored in the specified media location 
5. User navigates to the storefront and sees image displayed correctly 

## Checklist


packages/venia-ui/lib/util/makeUrl.js
```diff
@@ -1,8 +1,14 @@
...
+
+const storage = new BrowserPersistence();
+
 // If the root template supplies the backend URL at runtime, use it directly
 const htmlDataset = document.querySelector('html').dataset;
 const { imageOptimizingOrigin } = htmlDataset;
 // Protect against potential falsy values for `mediaBackend`.
-let mediaBackend = htmlDataset.mediaBackend;
+let mediaBackend =
+    storage.getItem('store_view_secure_base_media_url') ||
```
sirugh, line 10: This will only fetch the stored value on initial eval. You'd need to call `getItem` in the functions that reference `mediaBackend` so that the value is always fresh.

That said, I'm not sure what the affect will be on performance if we start doing storage reads for EVERY image/url.

packages/venia-ui/lib/util/makeUrl.js
```diff
@@ -1,8 +1,21 @@
...
+
 // If the root template supplies the backend URL at runtime, use it directly
 const htmlDataset = document.querySelector('html').dataset;
 const { imageOptimizingOrigin } = htmlDataset;
 // Protect against potential falsy values for `mediaBackend`.
-let mediaBackend = htmlDataset.mediaBackend;
+const storeCode = storage.getItem('store_view_code') || STORE_VIEW_CODE;
+const storeSecureBaseMediaUrl = {};
+
+AVAILABLE_STORE_VIEWS.forEach(store => {
```
supernova-at, line 12: Where does `AVAILABLE_STORE_VIEWS` come from? A comment here would be helpful for the future.

packages/venia-ui/lib/util/makeUrl.js
anthoula: Decided together that we would do a storage read, and fallback to global data from `AVAILABLE_STORE_VIEWS`, to build the `mediaBackend` url.

supernova-at (dismissed):

dpatil-magento:
QA Approved.

supernova-at (approved):

PR #2942, opened 2021-01-12
Set fixed min-height on main page

jimbo:
## Description

The `Main` component, which renders the header, page, and footer, sets `min-height: 100vh` on the page. As intended, this style prevents the footer from ever appearing above the fold. Unfortunately, setting a height with relative units causes issues when the page is rendered in an `<iframe>`, so we'll have to remove it.

This PR changes to `min-height: 40rem` (`640px`) and addresses any fallout from that change.

## Related Issue
PWA-1243.

### Verification Stakeholders
- @sirugh 
- @dpatil-magento

### Verification Steps
1. Visit the home page.
2. Verify that the footer is at least `640px` below the header.
3. Repeat for `/cart`, `/checkout`, and other pages.
4. On an empty page or 404 page, ensure the footer is only `640px` below the header.

Cart page on mobile

## Checklist


packages/venia-ui/lib/components/Main/main.css
```diff
@@ -12,7 +12,8 @@
 .page {
     margin: 0 auto;
     max-width: var(--venia-global-maxWidth);
-    min-height: 100vh;
+    min-height: 40rem;
```
sirugh, line 15: This is much more palatable than 100vh :) Nice and simple.

packages/venia-ui/lib/components/Main/main.css
```diff
@@ -12,7 +12,8 @@
...
+    padding: 1rem 0 1.5rem;
```
sirugh, line 16: I don't think this will work - some components are meant to fit snug against the header ie error page and the home page sample content. Notice the gap:

sirugh (changes requested):
Going to need to re-think the global padding. Also, something seems off with the page width and the left nav height:

[Image from Gyazo](https://gyazo.com/27a08edcadc3c58bd84c9c223fa198de)

fooman:
Bringing back the footer into view does cause quite a bit of jumping around on any pages that load content async like product and categories - https://pr-2942.pwa-venia.com/venia-tops/venia-blouses.html?page=1 which will affect Cumulative Layout Shift negatively https://web.dev/cls/

jimbo:
> Bringing back the footer into view does cause quite a bit of jumping around on any pages that load content async like product and categories - https://pr-2942.pwa-venia.com/venia-tops/venia-blouses.html?page=1 which will affect Cumulative Layout Shift negatively https://web.dev/cls/

Yes, very good point.  This highlights how problematic it is to async-load content without placeholders that allow layout to pre-allocate space for that content. Lighthouse does and should call this out, and we do need to address it.

brendanfalkowski:
Devil's advocate though — loading skeletons almost always have CLS too. They just never seem to be content-shaped enough to perfectly avoid it. Yeah Google wants you to avoid it, but can anyone without SSR? This just feels like one of those "a real project would never allocate resources to fixing it when the backlog has 200 tickets" areas.

packages/venia-ui/lib/components/Main/main.css
```diff
@@ -10,9 +10,10 @@
 }
 
 .page {
+    display: grid;
```
sirugh, line 13: Causes issues on mobile for the CMS page

[Image from Gyazo](https://gyazo.com/481d090a5de66b0f1cc891e35559ebb6)

jimbo:
> Devil's advocate though — loading skeletons almost always have CLS too. They just never seem to be content-shaped enough to perfectly avoid it. Yeah Google wants you to avoid it, but can anyone without SSR? This just feels like one of those "a real project would never allocate resources to fixing it when the backlog has 200 tickets" areas.

That's true, but CLS measures _how much_ layout shift, so even an imperfect skeleton is better than none.

That said, it's also true that layout shift will continue to be a problem until we land SSR. Because routing information is stored on the server rather than the client, we can't even render a meaningful skeleton until `useMagentoRoute` finishes its query, at which point rendering a skeleton is questionable.

Accordingly, I've stripped this PR down to the essential bugfix and nothing more.

supernova-at (approved):

sirugh (approved):

dpatil-magento:
@jimbo Looks good. Just a query when page content is very less, should there be max height?

image

jimbo:
> @jimbo Looks good. Just a query when page content is very less, should there be max height?

Should be fine as it is.

dpatil-magento:
QA Approved.

PR #2948, opened 2021-01-15
[SPIKE] Infinite Loading w/o over fetching

tjwiebell:
## Description

Example of how we might handle more optimized infinite loading. Our `cache-and-network` fetchPolicies make this a bit difficult, where we likely don't want to render everything in the cache back to the user when they return to this view, because it would re-fetch every single order in the cache; because of this I've implemented a merge function that resets the cache result if you return to the first page.

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
+++ b/packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
@@ -10,7 +10,7 @@ import { deriveErrorMessage } from '../../util/deriveErrorMessage';
 
 import DEFAULT_OPERATIONS from './orderHistoryPage.gql';
 
-const PAGE_SIZE = 10;
+const PAGE_SIZE = 2;
 
 export const useOrderHistoryPage = (props = {}) => {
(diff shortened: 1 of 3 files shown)


packages/peregrine/lib/talons/OrderHistoryPage/useOrderHistoryPage.js
```diff
@@ -10,7 +10,7 @@ import { deriveErrorMessage } from '../../util/deriveErrorMessage';
 
 import DEFAULT_OPERATIONS from './orderHistoryPage.gql';
 
-const PAGE_SIZE = 10;
+const PAGE_SIZE = 2;
```
sirugh, line 13: Reason for smaller page size? Just to limit the amount of data?

tjwiebell: Just so it's easier to demo the behavior; would back this out before delivery.

supernova-at:
Seems like we lost track of this one, @tjwiebell . Feel free to re-open if you'd like.

PR #2960, opened 2021-01-21
Correct dialog form values after submission

supernova-at:
## Description

This ticket originally called for the submit / confirm handler to clear out form values like the reset / cancel button does.

Unfortunately, that didn't end up being viable. Take the following code addition to the `Dialog` component itself:

```js
    const onSubmit = useCallback(
        async args => {
            if (onConfirm) await onConfirm(args);
            formApiRef.current.reset();
        },
        [onConfirm]
    );
```

We actually only want to reset the form when `onConfirm` is successful. When this function results in an error, we want to keep the form fields in their current state so that the user can more readily fix the problem.

Unfortunately, there's no established pattern or enforcement of a return value from the `onConfirm` callback. In fact, most of the ones in the Venia storefront return `undefined`, explicitly or implicitly.

It feels very heavy handed and error prone for the `Dialog` component to suggest or require such a return value from its callers.

As such, I think the best approach for now is a targeted one where callers set `shouldUnmountOnHide` to `true` on their `Dialog`s when they want the reset functionality. The form will be re-mounted when the `Dialog` reappears on screen, and the (new, updated)`initialValues` will be used.

To facilitate this, this PR changes the default value for the `Dialog` `shouldUnmountOnHide` prop to `true`.
Any instance of `Dialog` that was explicitly setting `shouldUnmountOnHide` to `true` has been updated to omit that prop, and any instance of `Dialog` that was missing `shouldUnmountOnHide` / relying on the default value has been updated to explicitly set `shouldUnmountOnHide` to `false`.

## Related Issue

Closes PWA-961.

### Verification Steps

1. Sign In
2. Navigate to the `/account-information` page (`My Account` > `Account Information`)
3. Click the "Edit" button
4. Change the "Last Name" field and enter the correct password
5. Keep a close eye on the "Last Name" field and click "Submit"
6. Verify that there was no flash of old content in the "Last Name" field
7. Re-open the "Edit" Dialog by clicking on the "Edit" button
8. Verify that the "Last Name" field is correctly updated to the new value and that the password field has been cleared out

Screen Shot 2021-01-21 at 12 39 51 PM

## Checklist

## Risks

Do we need to make this change anywhere else?


schensley:
Reviewed via EDIT Account Information and EDIT Shipping Address in Checkout. Behaves as expected. UX Approved.

packages/venia-ui/lib/components/Dialog/dialog.js
```diff
@@ -180,5 +180,5 @@ Dialog.defaultProps = {
     confirmText: 'Confirm',
     confirmTranslationId: 'global.confirmButton',
     isModal: false,
-    shouldUnmountOnHide: false
+    shouldUnmountOnHide: true
```
sirugh, line 183: This solution unmounts/destroys the component meaning we lose animations.

sirugh (approved):
There are several files that did not use this prop (meaning they used the default of `false`). Should we add `shouldUnmountOnHide={false}` to those `<Dialog>` implementations?

I also have concerns about losing the transitions. I'm not sure it was intentional otherwise I'd expect you to have removed related CSS.

sirugh (changes requested):
There are several files that did not use this prop (meaning they used the default of `false`). Should we add `shouldUnmountOnHide={false}` to those `<Dialog>` implementations?

I also have concerns about losing the transitions. I'm not sure it was intentional otherwise I'd expect you to have removed related CSS.

supernova-at:
> There are several files that did not use this prop (meaning they used the default of false). Should we add shouldUnmountOnHide={false} to those <Dialog> implementations?

Yes, great idea. I can definitely do that.

> This solution unmounts/destroys the component meaning we lose animations.

Very valid concern but we have a separate ticket (PWA-1194) to restore the close animations when `shouldUnmountOnHide` is `true` so I didn't do any of that in this scope.

sirugh (approved):

dpatil-magento:
@supernova-at Tried verification steps and Password field still not cleared - 

<a href="https://gyazo.com/80110cb46fa3c6f83725b4a912791ab3">Image from Gyazo</a>

supernova-at:
> @supernova-at Tried verification steps and Password field still not cleared -

Lol oops, accidentally flipped that one twice.

Good to go in https://github.com/magento/pwa-studio/pull/2960/commits/053b97710b9ff0eb3287353c64efad02536d7c21.

dpatil-magento (approved):

dpatil-magento:
QA Approved

PR #2967, opened 2021-01-23
Eliminate cyclic dependencies

jimbo:
## Description
There should not be cyclic dependencies between `venia-ui` and `peregrine`. The former should depend on the latter as necessary, but the latter should never depend on the former.

This PR eliminates the few cases where `peregrine` was importing from `venia-ui`.

- Where possible, files move from `venia-ui` to `peregrine`.
- Otherwise, files are duplicated in `peregrine` pending a future refactor.

## Related Issue
PWA-432.

### Verification Stakeholders
- @sirugh 
- @dpatil-magento 

### Verification Steps
1. Verify images load on product pages.
2. Verify tests pass.

## Checklist


sirugh (approved):
LGTM!

sirugh (approved):

dpatil-magento:
QA Approved.

PR #2980, opened 2021-01-29
Support for express server compression middleware.

revanth0212:
## Description

The `express` server that serves the app provides a middleware that applies GZIP compression if not applied already on the responses requested by the storefront. Compression is a very important but sensitive feature to manage because it can be applied at different layers of a system. One can apply it at the proxy layer or application layer and overdoing it can cause performance degradation. Hence the express server compression will be controlled by an environment variable. If someone wants to apply compression at the express layer, they have to set the value of `ENABLE_EXPRESS_SERVER_COMPRESSION` to `true`. By default, it will be set to `false`.

`ENABLE_EXPRESS_SERVER_COMPRESSION` in a sample project's `.env` file created using the `create-pwa` scafolding tool:

image

## Related Issue
Closes PWA-1108.

## Acceptance
Should apply GZIP compression on responses if `ENABLE_EXPRESS_SERVER_COMPRESSION` to `true`.

### Verification Stakeholders
@jimbo 
@dpatil-magento 

### Verification Steps
1. Change/create env variable in the .env file `ENABLE_EXPRESS_SERVER_COMPRESSION=true`
2. Run the app. Check both `yarn watch:venia` and `yarn build && yarn stage:venia`
3. Open the app and check the network tab for responses **with** a header `Content-Encoding: gzip`.
4. Change the value of `ENABLE_EXPRESS_SERVER_COMPRESSION` to `false`.
5. Restart the app.
6. Open the app and check the network tab for responses **without** a header `Content-Encoding: gzip`.

When `ENABLE_EXPRESS_SERVER_COMPRESSION=true`

image

When `ENABLE_EXPRESS_SERVER_COMPRESSION=false`

image

## Checklist


fooman:
This project used shrink-ray-current in the past
https://github.com/magento/pwa-studio/pull/2618/files#diff-2d63256fca103d1f1f8c9cc0a8c4adc56738f8a0397261b1240f824c87ba599e
to provide compression which I guess was favoured as it included brotli compression https://github.com/expressjs/compression/issues/71. Was dropping brotli support on purpose?

tjwiebell:
> This project used shrink-ray-current in the past
> ...
> Was dropping brotli support on purpose?

@fooman 
Digging through old PRs I can't find the justification from @zetlen, but I recall two big reasons
1. Ideally a different layer of the application would handle compression, and an Express middleware should be the very last option. We wanted it in place so local development would still give close to production performance scores, but was not intended as a go live solution.
2. The `iltorb` Python dependency of `shrink-ray` was troublesome. Between security alerts and weekly bug reports, the maintenance began to pile up.

This PR ideally solves both of those problems; use a very simple gzip compression middleware and make it opt-in. Paired with this we will also be working on documentation around compression best practices at the web server layer, focusing on AWS/nginx and Magento Cloud (Fastly on Pro/Stage branches, nginx config on Integration branches).

packages/venia-concept/package.json
```diff
@@ -12,7 +12,7 @@
   "scripts": {
     "build": "yarn run build:prod",
     "build:analyze": "yarn run clean && mkdir dist && webpack -p --profile --no-progress --env.mode production --json > dist/build-stats.json && webpack-bundle-analyzer dist/build-stats.json",
-    "build:debug": "node --inspect-brk ./node_modules/.bin/webpack --no-progress --env.mode",
+    "build:debug": "node --inspect-brk ./node_modules/.bin/webpack --no-progress --env.mode development",
```
tjwiebell, line 15: Any reason this went from `none` to `development`?

tjwiebell (changes requested):
Logic is sound, I think it just needs moved. Potential to remove unnecessary middleware too. Nice work!

packages/venia-concept/package.json
revanth0212: This was broken long ago. We never tried the script so we never found it. I had to use this script to debug the code and that's how I found the issue.

packages/upward-js/lib/gzipValidationMiddleware.js
```diff
@@ -0,0 +1,26 @@
...
+
+        try {
+            const acceptEncoding = req.headers['accept-encoding'];
+
+            if (acceptEncoding.split(',').includes('gzip')) {
+                const contentEncoding = res.getHeader('Content-Encoding');
+
+                if (contentEncoding !== 'gzip') {
+                    log(
+                        '\nGzip compression is supported by the Client. For better performance please enable gzip compression.\n'
```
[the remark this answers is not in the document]
revanth0212: It should not if you haven't set the ENV variable. Can you check? 

Also, I have added a check to not render this warning in a production environment but there is a known bug in the serve module that does not let it work if `NODE_ENV` is production. I have created a ticket in our backlog to handle this.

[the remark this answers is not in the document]
tjwiebell: Checked again, always compresses with `yarn watch:venia`, while `yarn stage:venia` seems to work as intended. Regardless, I would still get rid of the warning. If someone is purposely not using compression or using a different algorithm, there's no reason to annoy them.

packages/pwa-buildpack/envVarDefinitions.json
```diff
@@ -199,6 +195,17 @@
...
             ]
+        },
+        {
+            "name": "Express compression settings",
+            "variables": [
+                {
+                    "name": "ENABLE_EXPRESS_SERVER_COMPRESSION",
+                    "type": "bool",
+                    "desc": "Specify if express server compression needs to be enabled. Defaults to false if not provided.",
+                    "default": "false"
```
tjwiebell, line 206: Maybe this line? The type isn't correct and might be casting to true.
```suggestion
                    "default": false
```

packages/upward-js/lib/createUpwardServer.js
```diff
@@ -49,6 +51,16 @@ async function createUpwardServer({
...
     const upward = await middleware(resolve(upwardPath), env);
+
+    if (env.ENABLE_EXPRESS_SERVER_COMPRESSION === 'true') {
+        app.use(
+            compression({
+                threshold: 0
+            })
+        );
+        app.use(gzipValidationMiddleware(env));
+    }
```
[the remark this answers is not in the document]
revanth0212: Having it in the `PWADevServer` will only run this as part of the Dev mode. If someone wants to have compression in other modes/environments as well, that wouldn't be possible with it being in `PWADevServer`. That's the reason why I have left this in the `serve` module.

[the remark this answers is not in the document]
tjwiebell: Let me rephrase; we cannot add API to `upward-js` without also updating the `upward-spec` and `upward-php` implementations. **If** I had confidence UPWARD was going to remain in the stack, this would be an option, but it would likely be better to use `upward.yml` config as opposed to an environment variable.

Given this, please move the middleware to `PWADevServer` and `utilities/serve` in buildpack. Sorry I didn't specify to also put it in the `serve` utility as well, was just pointing out where to find an example of the IOC pattern.

packages/pwa-buildpack/lib/Utilities/serve.js
```diff
@@ -29,12 +30,23 @@ module.exports = async function serve(dirname) {
...
+                            threshold: 0
+                        })
+                    );
+                }
             }
         }
     );
 
     let envPort;
-    if (process.env.PORT) {
```
revanth0212, line 37: This is a false positive all along. Since we dint have tests, we never found it. I have added covering tests for this change.

packages/pwa-buildpack/lib/Utilities/serve.js
```diff
@@ -29,12 +30,23 @@ module.exports = async function serve(dirname) {
...
+    /**
+     * null and undefined are represented as strings in the env
+     * so we have to match using strings instead.
+     */
+    if (process.env.PORT !== 'null' && process.env.PORT !== 'undefined') {
```
tjwiebell, line 49: I was going to propose `parseInt` here, but I guess 0 is a valid port number. I'm fine with this, but a little type checking wouldn't hurt.

tjwiebell (dismissed):
Suggestion for test improvement but not a blocker to send this to QA. Awesome work covering this with tests too 👌

packages/pwa-buildpack/lib/Utilities/serve.js
revanth0212: Yeah you are right, 0 is a valid port number.

tjwiebell (dismissed):

dpatil-magento:
@revanth0212 Condition update to port validation is failing. When you run `yarn stage:venia` with STAGING_SERVER_PORT set in .env. Either its ignored or entering if condition before this and ends up running server on random port. Same issue on AWS so PR deploy instance is not accessible.

```
yarn run v1.22.4
$ yarn venia start
$ yarn workspace @magento/venia-concept start
$ buildpack serve .
  ℹ  PORT is set in environment: undefined
  ℹ  Launching UPWARD server
https://magento-venia-concept-ub-0n.local.pwadev:9536/
```

dpatil-magento:
QA Approved.

docker/.env.docker.prod
```diff
@@ -16,4 +16,4 @@ UPWARD_JS_LOG_URL=1
 CHECKOUT_BRAINTREE_TOKEN=redacted
 GOOGLE_MAPS_API_KEY=redacted
 MAGENTO_BACKEND_EDITION=EE
-ENABLE_EXPRESS_SERVER_COMPRESSION=false
+ENABLE_EXPRESS_SERVER_COMPRESSION=true
```
tjwiebell, line 19: Our official guidance should be to do compression at the web server or proxy layer, not the application layer. Any reason you opted for this config instead of adding compression to nginx?

tjwiebell (changes requested):

docker/.env.docker.prod
revanth0212: Dev wanted to test it out and have it enabled on our AWS instances which are not served by Nginx (I presume).

revanth0212: You are right @tjwiebell. Our app is served using nginx but the compression is not enabled.

For instance, this is a PR instance without 
image


And here is the same network req from this PR instance with compression enabled,
image

@dpatil-magento do you know if we can enable compression at nginx level, if so, how long does it take. Till then this can be a stop-gap solution.

tjwiebell: Given this change I'm assuming we just deploy these Docker images directly to EBS, so nginx is involved. It sounds like this isn't as straight forward as it should be with the `jwilder/nginx-proxy` container, but is possible. Not going to block this PR, but we should consider doing this at the correct layer if our goal is performance.

Where to add config: https://github.com/nginx-proxy/nginx-proxy#proxy-wide
Sample of config needed: https://github.com/h5bp/server-configs-nginx/blob/master/h5bp/web_performance/compression.conf

tjwiebell (approved):

PR #2982, opened 2021-01-30
Fix #2981: let DEFAULT_TIMEOUT for Toasts be used

### LINKED ISSUE #2981: [bug]: Toasts don't fallback to DEFAULT_TIMEOUT
**Describe the bug**

The `useToasts` hook appears to support a DEFAULT_TIMEOUT but the way it's implemented it can't be used.

You shouldn't need to specify the timeout every time a Toast is added.

**To reproduce**

1. Add a toast anywhere
2. But don't pass the `timeout` prop

**Expected behavior**

The Toast should timeout after the DEFAULT_TIMEOUT value specified in the hook.

**What Actually Happens**

Not passing `timeout` results in a TypeError. This means that every instance adding a Toast must specify the value which is bad for consistency.

**Possible solutions**

I will PR a fix shortly.

-   [ ] `venia-concept`
-   [ ] `venia-ui`
-   [ ] `pwa-buildpack`
-   [x] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

brendanfalkowski:
## Description

Toasts don't use the DEFAULT_TIMEOUT which is defined, so developers must specify it in every instance.

## Related Issue

Closes #2981.

### Verification Steps

1. Find an instance of `addToast()`
2. Remove the `timeout` prop.
3. It should fallback to the default specified in the hook.

## Checklist


packages/peregrine/lib/Toasts/useToasts.js
```diff
@@ -113,7 +113,7 @@ export const useToasts = () => {
             const {
                 dismissable,
                 message,
-                timeout,
+                timeout = DEFAULT_TIMEOUT,
```
sirugh, line 116: We specifically did not do this because passing a timeout of `0` or `false` is allowable, and actually means "do not time out".  We actually do use the default timeout later in the file:

https://github.com/magento/pwa-studio/blob/9edff927cfc5488fc74504b02a5a8d1097a2f0aa/packages/peregrine/lib/Toasts/useToasts.js#L149-L158

If you pass a `timeout` of `0` or `false`, it never times out (meaning the user has to close it manually).
If you pass a `timeout` of a number > 0, it uses times out after that.
If you do not pass a `timeout`, it times out after the `DEFAULT_TIMEOUT`.

sirugh (changes requested):
I want to say this is working as intended, but I'd happily concede the point if you can show me an example where this is not working properly.

packages/peregrine/lib/Toasts/useToasts.js
jimbo: > We specifically did not do this because passing a timeout of 0 or false is allowable

Sure, but application of the default param here only happens if `props.timeout` is [`undefined`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters). Seems like the right change to make to me, unless I'm missing something else.

jimbo (approved):

packages/peregrine/lib/Toasts/useToasts.js
brendanfalkowski: My DevEx case was not wanting to pass any props that just reiterate what the logic determines to be default behavior. So `{ message, type }` is better than having to remember `{ message, timeout, type }` or `{ dismissable, message, type }` every time.

sirugh: > Sure, but application of the default param here only happens if props.timeout is undefined.

Huh, for some reason I thought it would not apply but I think I'm confusing `defaultProps` with destructured defaults.

PR #2983, opened 2021-02-01
[PWA-1342] Upgrade workbox to v6

tjwiebell:
## Description

Story
As a developer, I would like to upgrade workbox to the latest version so that we can take advantage of improvements to how precache assets requests are fired from the ServiceWorker. Switching from v5 Promise.all flood to v6 one request at a time means that the main app does not have to compete for bandwidth with these precache requests that aren't even needed for initial render. High level testing has shown dramatic improvements in simulated 3G mobile Lighthouse tests.

The workbox upgrade also requires an upgrade to webpack to at least version 4.40.0, but I have found no issue going to the latest 4.46.0.

Acceptance Criteria
- workbox-* dependencies upgraded to 6.0.2 (or latest)
- webpack upgraded to 4.46.0
- Lighthouse Mobile performance metrics should improve

## Related Issue

* [[PWA-1342](https://jira.corp.magento.com/browse/PWA-1342)] Upgrade workbox to v6

### Verification Steps

1. `yarn watch:venia` with `DEV_SERVER_SERVICE_WORKER_ENABLED=true`
2. Verify no errors are in the console related to ServiceWorker registration
3. Do a full build (`yarn build && yarn stage:venia`)
4. Verify no errors during build compilation and app is served
5. Verify the upgrade worked by observing precache asset requests in the Performance tab of DevTools (requests should fire one at a time, instead of a big flood). Precache assets should follow the `<number>.<hash>.js` pattern.

## Checklist


package.json
```diff
@@ -50,6 +50,7 @@
     "@babel/plugin-transform-runtime": "~7.4.4",
     "@magento/eslint-config": "~1.5.0",
     "@types/jest": "~25.2.1",
+    "babel-plugin-dynamic-import-node": "~2.2.0",
```
tjwiebell, line 53: Another dependency that was being hoisted before that no longer is after this upgrade. Similar to the other deps we added here in the React17 upgrade, instead of doing a full audit was just added here.

packages/pwa-buildpack/lib/WebpackTools/__tests__/PWADevServer.spec.js
```diff
@@ -1,12 +1,8 @@
 jest.mock('errorhandler');
-jest.mock('fs');
```
tjwiebell, line 2: Not sure why this was mocked, but was causing an error with an upstream workbox dep trying to destructure a utility from `fs.promises`. Since nothing was asserting on this spy, seemed safe to remove the mock entirely.

packages/venia-concept/src/ServiceWorker/setupWorkbox.js
```diff
@@ -29,7 +29,7 @@ export default function() {
     clientsClaim();
```
tjwiebell, line 29: Doesn't have to be in this scope, but I'd like to audit the purpose of this file. `skipWaiting` is going to be removed in the future release, and from my quick review of docs, I'm not sure this is needed since we're using the [webpack plugin](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-webpack-plugin.GenerateSW#GenerateSW), which has flags for both `skipWaiting` and `clientsClaim`.

I _almost_ missed bumping this script version too, seems fragile to have this hidden dependency in here. The webpack plugin also seems to have an option for `importScripts`, so this file might not be needed anymore if this can all be accomplished at buildtime with the plugin.

supernova-at: Would you mind creating a ticket for this so we don't forget?

tjwiebell: For sure. I was hoping to bring this up to @sirugh and @revanth0212 and see how they felt about it. If it seems simple and straight forward, might make sense to do now so we don't have to fully regress SW functionality twice.

I'll create the ticket now, and we can just close it if we include it here.

supernova-at:
### Verification Steps

`yarn watch:venia` with `DEV_SERVER_SERVICE_WORKER_ENABLED=true`
✅  Verify no errors are in the console related to ServiceWorker registration 
⛔  I did get an "Failed to register SW" error on Chrome but I believe it was related to it not trusting my certificate. When I regenerated one using `create-custom-origin` I couldn't get Chrome to use the new cert but I verified that the SW is registered in Firefox now.

Do a full build (`yarn build && yarn stage:venia`)
✅ Verify no errors during build compilation and app is served.
❓ Again, this worked in Firefox but Chrome still failed to register SW.

Verify the upgrade worked by observing precache asset requests in the Performance tab of DevTools (requests should fire one at a time, instead of a big flood). Precache assets should follow the <number>.<hash>.js pattern.

❓ I'm not sure how to verify this one. I see the requests for a handful of <number>.<hash>.js files. They all say transferred by "service worker" and initiated by `runtime.<hash>.js`. `runtime.<hash>.js` was the first thing loaded.

They do appear all at once though, even though they take "no time" to resolve.

supernova-at (approved):

dpatil-magento:
@tjwiebell As discussed take a look if these console errors also needs to be fixed in scope of this pr.

image

dpatil-magento:
Above observation will be addressed separately. 
QA Approved .

PR #2986, opened 2021-02-03
[FEATURE] Added the xfwd option to the proxyMiddleware so HTTP_X_FORW…

lewisvoncken:
…ARDED_HOST can be used to determine the store

## Description

When you allow a Magento Route through the Venia proxy there is no possible way to determine the current store.
By enabling the xfwd option in the proxyMiddleware it is possible to allow Magento to determine the current store based on HTTP_X_FORWARDED_HOST.

For example when you have a Magento Module which generates a PDF file with product data it should also have the store information to display the correct language and storeview values of the product.

```
  'HTTP_X_FORWARDED_FOR' => '127.0.0.1',
  'HTTP_X_FORWARDED_PORT' => '8227',
  'HTTP_X_FORWARDED_PROTO' => 'https',
  'HTTP_X_FORWARDED_HOST' => 'pwa-studio-xelbq.local.pwadev:8227',
```

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#3031: [FEATURE] Added the xfwd option to the proxyMiddleware so HTTP_X_FORW…


sirugh:
@lewisvoncken this slipped past me - I'll get it on the board so we can review it :)

lewisvoncken:
@sirugh 

- create 2 storeviews (one en_US and en_UK)
- open a controller by adding a venia Proxy (for example /contacts )
- change the layout for en_US to set the Page Title including en_US and for en_US include en_UK in the name
- you will always see the default store because there is no proxy reference given
- pwastudio.com/contacts > returns Contacts - en_US
- pwastudio.co.uk/contacts > returns Contacts - en_US > expected result was Contacts - en_UK
- switch based on proxy reference data can be added by this change and be used to determine the actual country

lewisvoncken:
@tjwiebell can you please proceed with the review?

tjwiebell (approved):
Makes sense to me, proxies generally apply this header 👍

dpatil-magento:
@lewisvoncken Can you please update the branch with latest `develop` ? I don't have access to do that and cant test fully until it's upto date.

lewisvoncken:
@dpatil-magento The PR has been updated

dpatil-magento:
QA Approved.

PR #2988, opened 2021-02-03
[feature]: Lazy Load/trim unused bytes in main routes

sirugh:
## Description

Our main `client` and `vendor` bundles are pretty large. There are savings (**130k split from this PR alone!**) to be had by splitting code that is unnecessary for the first render. That would include things like modals, drawers, drop downs, etc. The majority of changes in this PR are just adding `React.lazy` to components that aren't necessary on first render, and wrapping their implementations in `<Suspense>`.

Note: Much of the remaining unused bytes are from dependencies that we must load at start, such as Apollo, but for whatever reason do not tree-shake. Other bytes are from our implementation of ContextProviders which expose a lot of unnecessary actions in the main bundle. Further work is necessary to trim out that deprecated code. Definitely more work than is in the scope of this specific ask.

Testing these changes proved more difficult than I would have liked. Kent Dodds has [a way](https://awesomereact.com/playlists/dev-tips-with-kent/lfb5jvHq9c4) around testing Suspense/Lazy but we don't have react-testing-library at our disposal right now. The current approach (mocking `React.lazy`) is a stop-gap.

## Related Issue

Closes [PWA-1109](https://jira.corp.magento.com/browse/PWA-1109).

### Verification Steps

1. Go to the home page/route.
2. "Clear site data" in dev tools.
3. Run the [coverage](https://developers.google.com/web/tools/chrome-devtools/coverage) tool.
4. Compare numbers to those on the same route/page on develop.pwa-venia.com. They should be **better**.
5. Repeat step 1-4 for all of the routes including: cms, category, product page, cart, checkout and my account.

**Develop Category Page**

**"Improved" Category Page**
[Image from Gyazo](https://gyazo.com/1c198242e88b58c6658ea1f06de5fe52)

## Checklist


docker-compose.yml
```diff
@@ -33,7 +33,6 @@ services:
       - ./packages/venia-concept/static:/usr/src/app/packages/venia-concept/static:rw
       - ./packages/venia-ui/.storybook:/usr/src/app/packages/venia-ui/.storybook:rw
       - ./packages/venia-ui/lib:/usr/src/app/packages/venia-ui/lib:rw
-      - ./packages/venia-ui/templates:/usr/src/app/packages/venia-ui/templates:rw
```
sirugh, line 36: For starters, I might have scope creeped this a bit but as far as I can tell we no longer use the `templates`, opting instead to use `venia-concept/template.html`. @zetlen @revanth0212 can ya'll confirm this?

revanth0212: You are right. We got rid of mustache templates in favor of Webpack's HTML Plugin which utilized template.html to render the main HTML file.

jimbo (commented):
Makes sense to me. Good savings. 👍

packages/venia-ui/lib/components/App/__tests__/__snapshots__/app.spec.js.snap
```diff
@@ -1,5 +1,29 @@
...
+  >
+    <Routes />
+  </Main>
+  <button
+    onClick={[Function]}
+  />
+  <() => {
+  /* istanbul ignore next */
+  cov_101e9trynu().f[0]++;
+  cov_101e9trynu().s[1]++;
```
sirugh, line 17: I don't know why code coverage started showing up in _some_ of the snapshots but not others. It has something to do with the jest mock. When the mock is local to the component being tested the coverage comments do not appear.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/__tests__/__snapshots__/priceAdjustments.spec.js.snap
```diff
@@ -32,7 +32,7 @@ exports[`it renders Venia price adjustments 1`] = `
         </span>
       </button>
       <div>
-        <ShippingMethods />
+        <() => Promise.resolve().then(() => _interopRequireWildcard(require('./ShippingMethods'))) />
```
sirugh, line 35: And I don't understand why the coverage comments do not show up here but they do in app.js...

packages/venia-ui/lib/components/App/__tests__/app.spec.js
```diff
@@ -160,7 +159,8 @@ test('renders a full page with onlineIndicator and routes', () => {
     };
     const { root } = createTestInstance(<App {...appProps} />);
 
-    getAndConfirmProps(root, Navigation);
+    // TODO: Figure out how to mock the React.lazy call to export the component
```
revanth0212, line 162: Hey dude, is this still valid given you have added a react mock in `venia-ui` mocks?

sirugh: Yes. There is no `Navigation` component, just the stringified version of the React.lazy callback.

packages/venia-ui/lib/components/App/__tests__/__snapshots__/app.spec.js.snap
sirugh: ok it looks like this was happening because I was running with some jest filter like "pattern" or "failed test" which caused the reporter to attach comments for that run. When I run the entire suite, they are removed.

revanth0212 (dismissed):
Nice work buddy. Couple of minor questions but mostly it is ready for QA.

packages/venia-ui/upward.yml
```diff
@@ -132,19 +132,11 @@ veniaStatic:
...
-# elsewhere in this file or any other upward.yml file in the build, the
-# UpwardIncludePlugin would fail to copy them if they were not mentioned here.
-
-# The templates directory includes partials which are only required by other
-# templates, and the UpwardIncludePlugin cannot parse templates to see them.
-veniaTemplateIncludes:
-    resolver: directory
-    directory:
-        resolver: inline
-        inline: './templates'
```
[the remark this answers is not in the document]
sirugh: These have not been used in PWA Studio since September 2019, after we swapped to the webpack `template.html` file. Anyone who was on the version from before then and updates since then, will have had to do this upgrade. This is just removing the files that aren't hooked up to anything.

packages/venia-ui/lib/components/CheckoutPage/PaymentInformation/__tests__/paymentInformation.spec.js
```diff
@@ -105,7 +105,8 @@ test('Should render EditModal component only if doneEditing is true', () => {
 
     const tree = createTestInstance(<PaymentInformation />);
 
-    expect(tree.root.findByType(EditModal)).not.toBeNull();
+    expect(tree.toJSON()).toMatchSnapshot();
+    // expect(tree.root.findByType(EditModal)).not.toBeNull();
```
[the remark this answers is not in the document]
sirugh: Added a comment.

packages/venia-ui/lib/components/CartPage/PriceAdjustments/giftCardSection.ee.js
```diff
@@ -15,7 +16,9 @@ const GiftCardSection = props => {
                 defaultMessage: 'Apply Gift Card'
             })}
         >
-            <GiftCards setIsCartUpdating={setIsCartUpdating} />
+            <Suspense fallback={null}>
```
tjwiebell, line 19: Why the lack of fallbacks in all these added usages? Happy path fast network would mean you'd likely never see it, but perhaps this is still reachable? Tiny example of toggling Suspense on/off with React DevTools:

sirugh: I can add loading spinners. I just didn't think it was necessary since all the sections are behind user interaction. You're right though, we should probably include _something_ to indicate loading is happening.

tjwiebell: Intended to be a conversation starter, don't implement anything yet. I honestly haven't been able to catch an unloaded suspense component, even on Slow 3G, so if the use case isn't realistic, the `null`s are fine.

sirugh: It's easy enough to add them to these price adjustment accordions. Everything else though, I will not touch, including modals, or dropdowns, or the left nav.

revanth0212 (approved):

dpatil-magento:
@sirugh Lighthouse score on desktop looks good but mobile still shows unused js warning, pls check - 

image

sirugh:
@dpatil-magento, are you saying that desktop and mobile have different lighthouse scores? That's odd. Either way, in the PR description I did try to clarify why there is still unused js:

> Note: Much of the remaining unused bytes are from dependencies that we must load at start, such as Apollo, but for whatever reason do not tree-shake. Other bytes are from our implementation of ContextProviders which expose a lot of unnecessary actions in the main bundle. Further work is necessary to trim out that deprecated code. Definitely more work than is in the scope of this specific ask.

fooman:
Mobile would nearly always be lower as you have the same time targets with less bandwidth and less cpu. Still anything improving desktop should also improve mobile. Would love to see this merged with further efforts coming in additional prs.

sirugh:
@fooman yes, in reality, but does lighthouse throttle automatically? If so that would explain it :)

fooman:
Depends on how you run it. But for example via Google Chrome 
image

the result will show what throttling was applied
image


or via [webpagetest.org](https://webpagetest.org/easy.php)
image

fooman:
So looking at the results via https://webpagetest.org/ and comparing it to PR3000 this PR looks to be an improvement

https://webpagetest.org/result/210220_DiG7_13e1bbb50b7351e69bc60f1c6a16853b/
vs
https://webpagetest.org/result/210220_DiS0_f09ec649ab7eee75b1cabd9b573fe1ff/

First Contentful Paint
4.862s
vs
6.947s

Largest Contentful Paint
10.876s
vs
12.758s

Lighthouse score
22
vs
8

As always these are to be taken with a grain of salt as when you compare the time to first byte there is big difference too which I believe is the time to serve index.html (which should be very much the same between these two PRs):
1.193s
vs
2.848s

fooman:
I reran the above


PR 3000
https://webpagetest.org/result/210221_DiQP_73de239e4eb7cee73cc4816eebaab02e/
First Byte
1.062s

First Contentful Paint
4.982s

Largest Contentful Paint
11.801s

Lighthouse 
12

showing the variability of results (I believe a lot may be driven by how the PRs are hosted) and this time not a clear cut improvement. (lighthouse score up but Largest Contentful Paint 6 seconds slower!)

This PR
https://webpagetest.org/result/210221_Di9R_5a284d8d6785c73d84ac613a235c0bd9/

First Byte
0.824s

First Contentful Paint
6.034s

Largest Contentful Paint
17.380s

Lighthouse 
17

dpatil-magento:
@sirugh @fooman WPT results are might depend on backend performance also. Here I compared 3 instances and this PR instance looks better.  But @sirugh any reason js byte size is more on this pr https://www.webpagetest.org/video/compare.php?tests=210223_DiAX_2d42c6d682b41cde18176b657f3ae906,210223_DiXG_8ffb29c506015a3c670612e84235d0d2,210223_DiTF_b9a81f142cab77a940def48ca0fdfaf9 ?

sirugh:
> @sirugh any reason js byte size is more on this pr

@dpatil-magento without any investigation, my gut reaction is that this is because we are now creating more chunks which means more boilerplate JS.

fooman:
With more chunks the JS minification will not be able to create as many savings.

revanth0212 (dismissed):

dpatil-magento:
Given its a balanced approach going ahead with this. QA Approved

revanth0212 (approved):

PR #2991, opened 2021-02-04
Remove references to window

jimbo:
## Description

In preparation to support SSR in a Node environment, we'll need to remove references to `window` and other browser-specific global objects that don't exist in Node.

Additionally it was attempted to move to an SSR friendly webpack loader that would work with CSS Modules on the server, but was discovered injection of CSS by effects could cause a FOUC in some areas. Until SSR is ready, we'll continue using `style-loader`, but begin the process of migrating to a `useStyle` hook that will eventually be hooked up to this new loader.

## Related Issue
PWA-1392

### Verification Steps
1. General smoke test regression for client
2. Verify all form elements look and behave the same (several were re-factored to functional components)
3. `yarn build` and verify no errors
4. `cd packages/venia-concept/dist && node client.<hash>.SERVER.js` - verify no exceptions or errors thrown

## Checklist


pwa-devdocs/src/technologies/basic-concepts/app-shell/index.md
```diff
@@ -44,9 +44,12 @@ The following is a simple example of an application shell:
...
 
     <script>
-        if (process.env.SERVICE_WORKER && 'serviceWorker' in navigator) {
+        const { navigator } = globalThis;
+        const { serviceWorker } = navigator || {};
+
+        if (process.env.SERVICE_WORKER && serviceWorker) {
             window.addEventListener('load', () => {
-                navigator.serviceWorker
+                serviceWorker
```
supernova-at, line 52: Should the line above this change too?

pwa-devdocs/_drafts/project-setup/create-simple-peregrine-app/index.md
```diff
@@ -44,9 +44,12 @@ const container = document.getElementById('root');
...
 app.mount(container);
 
-if ('serviceWorker' in navigator) {
+const { navigator } = globalThis;
+const { serviceWorker } = navigator || {};
+
+if (serviceWorker) {
     window.addEventListener('load', () => {
-        navigator.serviceWorker
+        serviceWorker
```
supernova-at, line 52: Should the line above this change too?

supernova-at:
#2992 was just merged and references `localStorage` directly 😬 . 

I think we should update `deleteCacheEntry` to use `globalThis.localStorage`.

https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/Apollo/deleteCacheEntry.js#L57
https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/Apollo/deleteCacheEntry.js#L71

packages/peregrine/lib/hooks/useWindowSize.js
```diff
@@ -22,9 +23,9 @@ const getSize = () => {
 const useWindowSizeListener = () => {
     const [windowSize, setWindowSize] = useState(getSize());
 
-    const handleResize = () => {
+    const handleResize = useCallback(() => {
         setWindowSize(getSize());
-    };
+    }, [setWindowSize]);
 
     useEventListener(window, 'resize', handleResize);
```
tjwiebell, line 30: Is this change needed?
```suggestion
    useEventListener(globalThis, 'resize', handleResize);
```

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
```diff
@@ -218,20 +218,20 @@ export const useStoreSwitcher = (props = {}) => {
                             `/${storeCode}`
                         )}${params}`;
 
-                        window.location.assign(newPath);
+                        globalThis.location.assign(newPath);
```
tjwiebell, line 221: I'd be nervous about taking on this scope, but should react-router handle this? I'm positive I missed some justification in the demo why this was needed, but just wanted to check. Some added benefits would also being able to derive path with `useLocation` and not have to craft our own paths.

```js
const history = useHistory()
history.replace(newPath)
```

packages/peregrine/lib/util/imageUtils.js
```diff
@@ -33,7 +33,7 @@ export const generateUrlFromContainerWidth = (
     type = 'image-product',
     ratio = DEFAULT_WIDTH_TO_HEIGHT_RATIO
 ) => {
-    const intrinsicWidth = window.devicePixelRatio * containerWidth;
+    const intrinsicWidth = globalThis.devicePixelRatio || 1 * containerWidth;
```
tjwiebell, line 36:
```suggestion
    const intrinsicWidth = (globalThis.devicePixelRatio || 1) * containerWidth;
```

tjwiebell (changes requested):
Tedious but straight forward, nice work. A few suggestions + what Andy commented on. I also did an search of the codebase and found a few more that may need attention:

- packages/peregrine/lib/store/enhancers/errorHandler.js:114
- packages/peregrine/lib/talons/MagentoRoute/helpers.js:8
- packages/venia-concept/src/registerSW.js:18 (I'm guessing SW might be okay and won't be rendered in SSR scenarios)

pwa-devdocs/src/technologies/basic-concepts/app-shell/index.md
jimbo: No, this whole block of code only makes sense in a browser environment anyway. The server won't have a service worker, nor a window to attach listeners to (see L48).

pwa-devdocs/_drafts/project-setup/create-simple-peregrine-app/index.md
jimbo: No, for the same reason.

packages/peregrine/lib/talons/Header/useStoreSwitcher.js
jimbo: In general, you're correct: where possible, we should use React Router's `history` and `location` APIs to read from and write to the URL. The problem with store switching is that we designate a leading store code (e.g., `fr`) as a `basename`, and those APIs (e.g., `history.push()`) presume we aren't changing the `basename`. Since this code's purpose is to change the `basename`, it needs to use the browser's `location.assign()` instead.

That said, I ensured that code in this file only accesses `window.location` from within a user-interaction callback, `handleSwitchStore`. The server won't run this code, I believe.

tjwiebell (changes requested):
Changes look good 👍  community may benefit from some supporting docs here, or at least updating our [coding standards](https://github.com/magento/pwa-studio/wiki/Project-coding-standards-and-conventions); not a blocker though.

Did want to double check you saw Andy's comment about updating code that was recently merged: https://github.com/magento/pwa-studio/pull/2991#issuecomment-784373526. If that change isn't needed LMK and I'll approve and send this to QA.

sirugh:
Jumping in without context, but is there a lint rule we can add that will yell at us for `window` usage? Might make it easy to catch all these merged usages.

jimbo:
> Jumping in without context, but is there a lint rule we can add that will yell at us for `window` usage? Might make it easy to catch all these merged usages.

It's tricky, because there are two different cases when it comes to global objects throwing a `ReferenceError` in Node:

1. Code that should **also** run in Node
2. Code that should **not** run in Node

For the former, we simply need to change `window` to `globalThis`. But for the latter, if the code doesn't make sense in a Node environment, we need to add guardrails that prevent it from running there. Blindly switching to `globalThis` isn't enough; the lint rule would have to be pretty sophisticated.

```es6
// browser: works as intended
// server: ReferenceError: window is not defined
window.localStorage.setItem("foo", "bar")

// browser: works as intended
// server: TypeError: Cannot read property 'setItem' of undefined
globalThis.localStorage.setItem("foo", "bar")
```

In this way, it might actually be safer to _retain_ references to `window` for browser-only code and only use `globalThis` for isomorphic code that we want to run in both environments. That might make linting even more difficult, though.

jimbo:
> #2992 was just merged and references `localStorage` directly 😬 .
> 
> I think we should update `deleteCacheEntry` to use `globalThis.localStorage`.
> 
> https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/Apollo/deleteCacheEntry.js#L57
> https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/Apollo/deleteCacheEntry.js#L71

Thanks for letting me know. I've rebased on `develop` and added a guard. 👍

tjwiebell (approved):

tjwiebell (approved):
I think that does it, great improvements that pave the way for SSR. I'm not holding this up for linting, but we'll have to be mindful in reviews not to introduce any `window` or `mergeClasses` references going forward. Nice work 👍

dpatil-magento:
1. Filter model UI is broken - 
  -    Apply filter and reload page >> The check mark indication is not visible.
  -    Apply filter and click on Clear all button >> Filters UI shows grayed out check marks
  -    **On Mobile** - Select a filter >> It gets applied as desktop but it should wait for user to take next action.

<a href="https://gyazo.com/6422184f6b6c129100d54e02363bfb4e">Image from Gyazo</a>

WIP.......

jimbo:
> @jimbo Pls check following observations -
> 
>     1. Filter model UI is broken
>     2. SERVER.js size is in MB's, `yarn build` throws warning
>     3. `yarn storybook:venia` fails
>     4. Localization/Multistore broken when config set to non-default store- See blank page with console error

1. This is due to a bug with Informed. I'll try to find a workaround.
2. This is fine. The size of the server bundle doesn't matter.
3. Good catch, fixed.
4. Good catch, fixed.

dpatil-magento:
@jimbo check mark issue solved.   But mobile issue I can still reproduce (MFTF test fails so not stale cache related and all) and there are conflicts which you need to take care of. 

<a href="https://gyazo.com/a42a1038c1cb8bf92169b578fca7f30b">Image from Gyazo</a>

I will retest app once all these are taken care.

dpatil-magento:
@jimbo Fixes look good.   I believe you missed to apply storybook change in https://github.com/magento/pwa-studio/blob/8e5dfeae386e96ce4fca8c2af50700f656afdd57/packages/venia-concept/src/.storybook/webpack.config.js#L32 and scaffold app storybook is failing. I did apply the change and tested locally, it worked, pls check and commit. 

Also there are test failures to be fixed. 

 **Once these are taken care and no other app code touched part on new commits then this PR is good to merge.**  In case any other app code gets updated I can just do quick QA on that area and then merge.

packages/venia-ui/lib/components/RadioGroup/radioGroup.js
```diff
@@ -33,15 +32,15 @@ const RadioGroup = props => {
                     label: classes.radioLabel,
                     root: classes.radioContainer
                 }}
-                id={`${domId}--${value}`}
+                id={`${id}--${value}`}
```
tjwiebell, line 35: Reflected in the snaps, all usages of this component don't pass `id`, so this generates ids like `undefined--value`. That id also gets passed to Informed, but React removes it since it's `undefined`. I propose we either require `id` or remove its usage entirely.

tjwiebell (changes requested):
Clean up `undefined` leakage and this is good to go 👍

tjwiebell (approved):

dpatil-magento:
Reverified radio buttons, QA approved.

PR #2992, opened 2021-02-04
Remove data from all persisted Apollo caches

supernova-at:
## Description

PWA Studio uses [apollo-cache-persist](https://github.com/apollographql/apollo-cache-persist) to persist (save) the Apollo client cache to the browser's local storage.

Unfortunately our `CachePersitor` can only delete from the "active" store view's cache representation in local storage. Meaning when we were deleting sensitive data (after sign out, for example), we were potentially leaving some behind.

This PR targets and deletes sensitive data in what I'm calling these "inactive" caches in addition to the removal from the active cache, which is already handled.

## Related Issue

Closes PWA-1339.

#### Sign In

1. Add an item to the guest cart
1. Switch store views
1. Sign in
1. See that the `Cart` and `Customer` entries in all of the `apollo-cache-persist*` entries in local storage are removed
1. Note that the active store view will have the `Cart` and `Customer` entries replaced with the newly signed-in user's information soon as those network requests complete. That's fine.

#### Sign Out

1. Sign in and add an item to your cart
1. Switch store views
1. Sign out
1. See that the `Cart` and `Customer` entries in all of the `apollo-cache-persist*` entries in local storage are removed

#### Create Account

1. Add an item to the guest cart
1. Switch store views
1. Create a new account
1. See that the `Cart` entry in all of the `apollo-cache-persist*` entries in local storage is removed

#### Place Order

1. Sign in
2. Add an item to the cart
3. Switch store views
4. Complete the checkout flow and place an order
5. See that the `Cart` entry in all of the `apollo-cache-persist*` entries in local storage is removed

## Checklist


packages/peregrine/lib/Apollo/deleteCacheEntry.js
```diff
@@ -21,14 +26,48 @@ export const deleteCacheEntry = async (client, predicate) => {
...
         await client.persistor.persist();
     }
 };
+
+const deleteInactiveCachesEntry = async (client, predicate) => {
+    const activeApolloCacheLocalStorageKey =
+        client.persistor.persistor.storage.key;
+
+    const isAnInactiveApolloCache = ([key]) => {
+        // TODO: how do we keep this in sync with the storefront?
```
sirugh, line 47: What do you mean?

sirugh (dismissed):
This should work well :) I had a question about your TODO -  would probably leave guidance rather than an open question.

packages/peregrine/lib/Apollo/deleteCacheEntry.js
```diff
@@ -21,14 +26,48 @@ export const deleteCacheEntry = async (client, predicate) => {
...
 };
+
+const deleteInactiveCachesEntry = async (client, predicate) => {
+    const activeApolloCacheLocalStorageKey =
+        client.persistor.persistor.storage.key;
+
+    const isAnInactiveApolloCache = ([key]) => {
+        // Keep this in sync with the cache names used by your storefront.
+        return (
+            key.startsWith('apollo-cache-persist') &&
```
sirugh, line 49: I do think this deserves to be an export of a constants file like `peregrine/lib/Apollo/constants.js`. Then we could import here and in the main app where we define the cache key.

sirugh (changes requested):
You noted that it needs tests on the jira ticket. Aside from that, I would only ask that we find a way to keep the key prefix, ie `apollo-cache-persist` in sync.

packages/peregrine/lib/Apollo/deleteCacheEntry.js
supernova-at: Done: https://github.com/magento/pwa-studio/pull/2992/commits/0d8cc98dc0a773c55b1114585e4e110ab89f83d4

sirugh (approved):
Beautiful

dpatil-magento:
QA Approved.

PR #2996, opened 2021-02-05
[PWA-1030] Merge Cart does not update mini cart if auth user has same simple product already added

tjwiebell:
## Description

URL - [https://develop.pwa-venia.com/]

Steps - 
1. Login as Auth user
2. add a simple product or config product to cart
3. Sign out from account.
4. Now add same simple or config product with same configs to gues cart
5. Now login to as Auth user again

Expected - Mini cart count should update and if user open mini cart it should show updated counts.

Actual - Mini cart count does not update. 

Note - 
1. Cart page works fine
2. If user does browser refresh then mini cart updates.
3. Works fine if guest cart has any item other than auth cart has.

## Related Issue

* [[PWA-1030](https://jira.corp.magento.com/browse/PWA-1030)] Merge Cart does not update mini cart if auth user has same simple product already added

### Verification Steps

1. Run through reproduction steps, verify that counts are now accurate (compare to counts after refresh)

## Checklist


sirugh (changes requested):
As discussed this isn't working right now. I followed verification steps on a simple product, https://pr-2996.pwa-venia.com/silver-amor-bangle-set.html, and the cart trigger did not update to "2" nor does the "qty" in the cart. What is odd is that the cart query for fetching the post-merge details shows the correct info, so it seems the cache is not updating.

[Image from Gyazo](https://gyazo.com/65cb5b1762229fdf969a47c28847e47c)

sirugh (approved):
Looks like it's working!

dpatil-magento:
QA Approved

PR #2997, opened 2021-02-08
Typescript spike

sirugh:
TS Spike, see various commits.


packages/extensions/venia-sample-language-packs/src/i18n-intercept.ts
```diff
@@ -0,0 +1,13 @@
...
+    of: (arg0: string) => any;
+    name: string | number;
+}) => {
+    const builtins = targets.of('@magento/pwa-buildpack');
+
+    builtins.specialFeatures.tap(
+        (features: { [x: string]: { i18n: boolean } }) => {
+            features[targets.name] = { i18n: true };
+        }
+    );
```
sirugh, line 12: Something to investigate: wrapping existing files ie useApp.ts/js

PR #2998, opened 2021-02-08
[PWA-998] Increase Test Coverage in peregrine/lib/talons/SignIn

tjwiebell:
## Description

As part of our initiative to increase test coverage, stability and confidence in our project we've identified this folder as having a lower than acceptable level of test coverage.

Acceptance Criteria:

Increase test coverage to an acceptable level, ideally 100%

## New Library Added
This PR introduces two changes to how we approach unit testing of talons

- Introduction of `@testing-library/react-hooks` - gives us a consistent pattern to align towards when testing talons and custom hooks. Seems proven enough that it's safe to introduce, at least in scope of peregrine. [Their docs](https://react-hooks-testing-library.com/usage/basic-hooks) have lots of good information on usage, but the big patterns are all covered in this PR.
- Usage of Apollo's `MockProvider` that simulates actual operations, instead of us mocking those hooks. The only odd piece here is that operations wait for the next tick of the event loop to flush, but RLRH simplifies this by just awaiting those triggers if you're not concerned with loading state. [Their docs](https://www.apollographql.com/docs/react/development-testing/testing/) also do a great job at explaining usage, but the major patterns are covered in this PR.

## TODO?
I haven't made any of what I've done in this PR a utility just yet. I'm thinking the whole team should iterate on these new patterns a bit before deciding if usage is consistent enough to add utilities. Advice from the maintainers to handle complex context nesting in the app is to add a `test/test-utils` utility that export a `renderHook` method that does this automatically, eg. similar to what `renderHookWithProviders` does here, but every context so we no longer need to mock `useCartContext` or `useUserContext`. This could prove to be difficult with redux still hanging around, but is the ideal solution going forward.

## Related Issue

* [[PWA-998](https://jira.corp.magento.com/browse/PWA-998)] Increase Test Coverage in peregrine/lib/talons/SignIn
### Verification Steps

1. Look at coveralls report, verify useSignIn is now covered

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/SignIn/__tests__/useSignIn.spec.js
+++ b/packages/peregrine/lib/talons/SignIn/__tests__/useSignIn.spec.js
@@ -1,0 +1,174 @@
+import React from 'react';
+import { MockedProvider } from '@apollo/client/testing';
+import { renderHook, act } from '@testing-library/react-hooks';
+
+import { useCartContext } from '../../../context/cart';
+import { useUserContext } from '../../../context/user';
+import defaultOperations from '../signIn.gql';
+import { useSignIn } from '../useSignIn';
+
+jest.mock('../../../Apollo/clearCartDataFromCache');
+jest.mock('../../../Apollo/clearCustomerDataFromCache');
+jest.mock('../../../hooks/useAwaitQuery');
+jest.mock('../../../store/actions/cart', () => ({
+    retrieveCartId: jest.fn().mockReturnValue('new-cart-id')
+}));
+
+jest.mock('../../../context/cart', () => ({
+    useCartContext: jest.fn().mockReturnValue([
+        { cartId: 'old-cart-id' },
+        {
+            createCart: jest.fn(),
+            removeCart: jest.fn(),
+            getCartDetails: jest.fn()
+        }
+    ])
+}));
+
+jest.mock('../../../context/user', () => ({
+    useUserContext: jest.fn().mockReturnValue([
+        {
+            isGettingDetails: false,
+            getDetailsError: 'getDetails error from redux'
+        },
+        { getUserDetails: jest.fn(), setToken: jest.fn() }
+    ])
+}));
+
+const signInVariables = {
+    email: 'fry@planetexpress.com',
+    password: 'slurm is the best'
+};
+const authToken = 'auth-token-123';
+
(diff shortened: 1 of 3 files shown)


packages/peregrine/lib/talons/SignIn/__tests__/useSignIn.spec.js
```diff
@@ -0,0 +1,174 @@
...
+    const [, { getUserDetails, setToken }] = useUserContext();
+
+    const { result } = renderHookWithProviders();
+
+    await act(() => result.current.handleSubmit(signInVariables));
+
+    expect(result.current.isBusy).toBe(true);
+    expect(setToken).toHaveBeenCalledWith(authToken);
+    expect(getCartDetails).toHaveBeenCalled();
+    expect(getUserDetails).toHaveBeenCalled();
```
tjwiebell, line 116: I would like to avoid implementation details from bleeding into tests, but included these two checks since we're mocking both of these contexts. Ideally we wouldn't mock these contexts and include them in the helper `renderHookWithProviders`, but that proved to be a bit difficult with redux in place. I hope once redux is deprecated from these contexts, this will be easier going forward, and we can assert on global state results, instead of checking that methods were called.

larsroettig (approved):
Looks clean :heart:

packages/peregrine/lib/talons/SignIn/__tests__/useSignIn.spec.js
```diff
@@ -0,0 +1,174 @@
...
+    act(() => result.current.handleCreateAccount());
+
+    expect(initialProps.setDefaultUsername).toHaveBeenCalledWith(mockUsername);
+    expect(initialProps.showCreateAccount).toHaveBeenCalled();
+});
+
+test('mutation error is returned by talon', async () => {
+    const signInErrorMock = {
+        request: signInMock.request,
+        error: new Error('Uh oh! There was an error signing in :(')
```
revanth0212, line 165: Do we seriously have a sad smiley in the error message 🤣 ?

larsroettig: He why not is funny :smile_cat:

revanth0212 (approved):
Wonderful work @tjwiebell.

PR #3004, opened 2021-02-09
PWA-1005: Increase Test Coverage in venia-ui/lib/components/Main

anthoula:
## Description

Increase test coverage for the `Main` component.

## Related Issue

https://jira.corp.magento.com/browse/PWA-1005

## Acceptance

Any developer

### Verification Stakeholders

Any developer

### Verification Steps

1. Run `yarn test --collectCoverageFrom=**/main.js packages/venia-ui/lib/components/Main/`
2. Verify tests succeed
3. Verify acceptable test coverage

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/Main/__tests__/main.spec.js
+++ b/packages/venia-ui/lib/components/Main/__tests__/main.spec.js
@@ -7,15 +7,27 @@ jest.mock('../../Footer/footer', () => 'Footer');
 jest.mock('../../../classify');
 
 test('it renders correctly when masked', () => {
-    const { root } = createTestInstance(<Main isMasked={true} />);
+    const props = {
+        children: [],
+        isMasked: true
+    };
+    const { root } = createTestInstance(<Main {...props} />);
+    const page = root.findByProps({ className: 'page_masked' });
 
     expect(root.findByProps({ className: 'root_masked' })).toBeTruthy();
-    expect(root.findByProps({ className: 'page_masked' })).toBeTruthy();
+    expect(page).toBeTruthy();
+    expect(page.props.children).toEqual(props.children);
 });
 
 test('sets different classnames when not masked', () => {
-    const { root } = createTestInstance(<Main isMasked={false} />);
+    const props = {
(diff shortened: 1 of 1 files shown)


packages/venia-ui/lib/components/Main/__tests__/main.spec.js
```diff
@@ -0,0 +1,21 @@
...
+    expect(root.findByProps({ className: 'root_masked' })).toBeTruthy();
+    expect(root.findByProps({ className: 'page_masked' })).toBeTruthy();
+});
+
+test('sets different classnames when not masked', () => {
+    const { root } = createTestInstance(<Main isMasked={false} />);
+
+    expect(root.findByProps({ className: 'root' })).toBeTruthy();
+    expect(root.findByProps({ className: 'page' })).toBeTruthy();
+});
```
tjwiebell, line 21: I know there's some debate about the value of snapshot tests, but I feel like we currently use them to detect breaking changes. I'm not locked into that mindset so I would consider either a snapshot test here, or a simple test to verify the `children` prop to rendered.

tjwiebell (changes requested):
Tiny suggestion for an additional test.

packages/venia-ui/lib/components/Main/__tests__/main.spec.js
sirugh: For what it's worth I suggested that @anthoula use these direct props assertions rather than snapshots. Your request about rendering the children is still valid though!

tjwiebell: Nice! Worth talking about at next peer review/team time then. Snapshot tests can serve a purpose, but we should probably not call them unit tests.

anthoula: Yep, originally had snapshot assertions, as that seems to be a current convention widely in use. Added an additional assertion for rendered children

tjwiebell (approved):

PR #2969, opened 2021-01-25
[FEATURE]  Implement Checkmo #2927

larsroettig:
## Description

- ImplementsCheckmo

## Related Issue

Closes [PWA-1242](https://jira.corp.magento.com/browse/PWA-1242).
Closes #2927 

## Acceptance

- Check/Money order is available during checkout if it's enabled for the website, available in the customer's country and the order total threshold is met. (Admin Panel Stores>configuration> Sales> Payment Methods> Check/Money Order)
- Check/Money order appears in the list of available payment methods according to the sorting order set in the Admin Configuration
- Billing address fields are shown for the check/money order payment if the Admin Panel configuration (Store>Configuration>Sales>Checkout> Checkout options> Display billing address on) set to Payment Method
- Make Check payable to and Send Check to values are store view specific and taken from Admin Panel Configuration. (Admin Panel Stores>configuration> Sales> Payment Methods> Check/Money Order)
- Payment method title _Pay by Check or Money order" is configurable in Admin Panel
- When Customer selects Check/Money order order, they complete checkout without providing any additional information related to payment method
- Customer sees payment information is displayed on order confirmation page and order history

### Verification Steps

1. Checkout Branch 
2. Config Checkmo in Magento backend
3. Link Extension in venia-concept `package.json` via 
```json 
"dependencies": {
   "@magento/venia-sample-payments-checkmo": "link:../extensions/venia-sample-payments-checkmo"
},
```
4.  run `yarn install`
5. run `yarn watch:venia`

## Checklist

* I have added tests to cover my changes, if necessary.
* I have added translations for new strings, if necessary.
* I have updated the documentation accordingly, if necessary.

PR #3018, opened 2021-02-18
[PWA-1002] Increase Test Coverage in venia-ui/lib/components/CheckoutPage

tjwiebell:
## Description

As part of our initiative to increase test coverage, stability and confidence in our project we've identified this folder as having a lower than acceptable level of test coverage.

Acceptance Criteria:

Increase test coverage to an acceptable level, ideally 100%

## Implementation Note
There was a component in here `brainTreeDropIn` that I didn't feel comfortable trying to cover. It relied on a lot of waterfall async work to render that wasn't straight forward, so might be better to tap on the original develop for covering that file. Could also be an opportunity to re-visit and see if patterns we've started coalescing on could be added to this component to help simplify.

## Related Issue

* [[PWA-1002](https://jira.corp.magento.com/browse/PWA-1002)] Increase Test Coverage in venia-ui/lib/components/CheckoutPage

### Verification Steps

1. Run tests locally or look at Coveralls reports
2. Verify that CheckoutPage, OrderSummary, and PriceAdjustments coverage improved

## Checklist


packages/venia-ui/lib/components/Accordion/__mocks__/index.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const MockAccordion = jest.fn(({ children, ...props }) => (
```
tjwiebell, line 3: I did this after discussions from peer review; that our components would benefit from more manual mocks. Would like to discuss this further, but provided manual mocks like these is pretty simple, but not necessary.

sirugh:
I haven't looked, but I really hope there aren't too many conflicts with https://github.com/magento/pwa-studio/pull/2969

packages/venia-ui/lib/components/Accordion/__mocks__/index.js
tjwiebell: I may have stumbled onto why we haven't embraced this more. It seems some very old Jest bug causes these duplicate mock warnings, because `jest-haste-map` stores these mocks by name and not by path. Given we're a project where we use index files to export public API, I can see how this warning may have scared us away from manual mocks, since all of these module names would be index.

I'm digging for a solution, but nothing seems to be working. As you can see, the tests function fine, there's just this very in your face warning you get the first time you run the suite. We could mock underlying components instead of index exports, but expect we'd run into a similar issue eventually.

Original Bug: https://github.com/facebook/jest/issues/2070
The warning:

```
jest-haste-map: duplicate manual mock found: index
  The following files share their name; please delete one of them:
    * <rootDir>/packages/peregrine/lib/RestApi/__mocks__/index.js
    * <rootDir>/packages/venia-ui/lib/components/Accordion/__mocks__/index.js
```

tjwiebell: Went ahead and just pulled this out, can disregard.

sirugh (approved):

PR #3019, opened 2021-02-19
#1617 - Fetch store name from the GraphQL server at compile time

### LINKED ISSUE #1617: Fetch store name from the GraphQL server at compile time.

vasilii-b:
## Description

With the changes from this PR, the `STORE_NAME` is not hardcoded anymore from the webpack's DefinePlugin 

https://github.com/magento/pwa-studio/blob/70cf1e0d1d656487f8e03d6aaaec21e78f09fb36/packages/venia-concept/webpack.config.js#L66

As suggested in #1617, if `STORE_VIEW_CODE` environment variable is used, the store_name value from the corresponding available store is used. If not, use the default value returned by the `getStoreConfigData` query is used for the `STORE_NAME`.

## Related Issue
Closes #1617.

### Verification Stakeholders

@revanth0212 

### Specification

Changes to the `venia-concept` and `pwa-buildback` packages must be reviewed.

### Verification Steps

Setup and run the webserver.

1. Go to the homepage.
2. View page's source code.
3. Find the `title` tag
4. Make sure the `title` tag value contains the store name as specified by the `STORE_VIEW_CODE` (or the default store).

image

## Checklist


packages/venia-concept/webpack.config.js
```diff
@@ -45,8 +45,18 @@ module.exports = async env => {
...
+     * in the .env file, because should set the store name from the
+     * given store code instead of the default one.
+     */
+    const availableStore = process.env.STORE_VIEW_CODE
+        ? availableStores.find(store => store.code === process.env.STORE_VIEW_CODE)
+        : undefined;
+
     global.MAGENTO_MEDIA_BACKEND_URL = mediaUrl;
     global.LOCALE = storeConfigData.locale.replace('_', '-');
+    global.STORE_NAME = storeConfigData.store_name;
```
sirugh, line 59: Do we need the global variable here? `global.MAGENTO_MEDIA_BACKEND_URL` and `global.LOCALE` are used in the `venia-concept/template.html` but I don't think we use `global.STORE_NAME` or `global.AVAILABLE_STORE_VIEWS`, so this just might be leaking detail unnecessarily.

sirugh (changes requested):
Thanks for the contribution! I had a small bit of feedback and please run `prettier` :)

packages/venia-concept/webpack.config.js
vasilii-b: Hi @sirugh,
I've added the `global.STORE_NAME` as per the suggestion in #1617.
Please confirm I should delete it.
Thanks!

sirugh: I think the suggestion was indicating that `STORE_NAME` should be used in the `DefinePlugin`, which it already is. Please delete it.

packages/venia-concept/webpack.config.js
```diff
@@ -63,7 +75,10 @@ module.exports = async env => {
              * the globals object in jest.config.js.
              */
             POSSIBLE_TYPES: JSON.stringify(possibleTypes),
-            STORE_NAME: JSON.stringify('Venia'),
+            STORE_NAME:
+                process.env.STORE_VIEW_CODE && availableStore
+                    ? JSON.stringify(availableStore.store_name)
+                    : JSON.stringify(global.STORE_NAME),
```
revanth0212, line 81:
```suggestion
                    : JSON.stringify(storeConfigData.store_name),
```

revanth0212 (changes requested):
Thanks for your contribution. Please take care of the change @sirugh requested and also handle the suggestion I have provided below. With those handled, this PR is good to go.

packages/venia-concept/webpack.config.js
```diff
@@ -45,6 +45,17 @@ module.exports = async env => {
...
+    /**
+     * Loop the available stores when there is provided STORE_VIEW_CODE
+     * in the .env file, because should set the store name from the
+     * given store code instead of the default one.
+     */
+    const availableStore = process.env.STORE_VIEW_CODE
+        ? availableStores.find(
+              store => store.code === process.env.STORE_VIEW_CODE
+          )
+        : undefined;
```
sirugh, line 57: We can also simplify this, since `Array.find` will return `undefined` if it does not find an entry that matches the predicate.

```suggestion
    const availableStore = availableStore.find({ code } => code === process.env.STORE_VIEW_CODE);
```

packages/venia-concept/webpack.config.js
```diff
@@ -63,7 +74,10 @@ module.exports = async env => {
              * the globals object in jest.config.js.
              */
             POSSIBLE_TYPES: JSON.stringify(possibleTypes),
-            STORE_NAME: JSON.stringify('Venia'),
+            STORE_NAME:
+                process.env.STORE_VIEW_CODE && availableStore
+                    ? JSON.stringify(availableStore.store_name)
+                    : JSON.stringify(storeConfigData.store_name),
```
sirugh, line 80: This can also be simplified. If we have a matching `availableStore` we should just use that value, otherwise fall back to the default store name.

```suggestion
            STORE_NAME: availableStore
                    ? JSON.stringify(availableStore.store_name)
                    : JSON.stringify(storeConfigData.store_name),
```

sirugh (changes requested):
A few more thoughts I had last second.

sirugh:
I fixed the lint issue from my suggestion - sorry :)

packages/venia-concept/webpack.config.js
```diff
@@ -45,6 +45,15 @@ module.exports = async env => {
     const storeConfigData = await getStoreConfigData();
     const { availableStores } = await getAvailableStoresConfigData();
 
+    /**
+     * Loop the available stores when there is provided STORE_VIEW_CODE
+     * in the .env file, because should set the store name from the
+     * given store code instead of the default one.
+     */
+    const availableStore = availableStore.find(
```
revanth0212, line 53:
```suggestion
    const availableStore = availableStores.find(
```

revanth0212 (changes requested):
There is a build error due to illegal usage of `availableStore`. Please fix it.

image

sirugh (dismissed):

revanth0212 (approved):

sirugh:
@vasilii-b I'm so sorry but I think I gave bad direction on this original work. Since the app hard codes the store name into the bundle, it'll always be wrong when you switch to store views that use different names. You can see that here: https://pr-3019.pwa-venia.com/venia-tops.html?page=1 by switching to a different view from default. `Default Store View` is still shown as the name. I'm going to spend some time thinking about how to handle this, and probably will push a change to remedy it. I just wanted to apologize for wasting your time!

sirugh:
@vasilii-b I just pushed a commit with an _idea_ that I had, not something we're positive that we want just yet. I will work with the internal team to make a decision about how we want to proceed.

brendanfalkowski:
Two cents — we went back/forth on this. Some things like our manifest + static icons have to be overwritten when our Docker containers are built for our multi-site frontends. But the title ended up being simple enough to hardcode a conditional with a reasonable default (for us).

Screen Shot 2021-02-25 at 1 06 16 PM

I know that's not a config-only solution, but it saves an extra request at render, so we've ended up just hardcoding a lot of Magento config in frontend code that's only set once and never changed. We've started doing this as a rule now (fetching config should never delay render) for performance, but I know it only solves this at the implementation-level not the framework.

Still curious to see what you come up with @sirugh

sirugh:
> Still curious to see what you come up with @sirugh

@brendanfalkowski I pushed my approach which hopes to solve two problems:

1) Build time store config is not valuable since it is store specific (indicated by store code in the header).
2) Many components pass `STORE_NAME`.

I wanted to prevent developers from needing to make a store config query in every component, so the reasonable take was to do a single fetch _somewhere_. Ended up making a new Title component that just appends the fetched name.

packages/venia-concept/template.html
```diff
@@ -7,8 +7,6 @@
         <meta name="theme-color" content="#ff6334">
         <link rel="manifest" href="/manifest.json">
 
-        <title><%= STORE_NAME %></title>
```
sirugh, line 10: Is this OK? The app renders the title, not sure if we need the service worker template to do this. We wouldn't be able to ensure this is valid anyways.

dpatil-magento:
@sirugh Browser title looks good but offline mode on Category and Search is broken.

Steps - 
1. Visit Home, category, product and load some search results.
2. Now go offline mode and try to load above

Issue - Category and search does not load.
<a href="https://gyazo.com/4536ade8af1c7beb2c06226761e6aa84">Image from Gyazo</a>

sirugh:
Sort of figured it out. The category page uses a conditional to render an `ErrorView` based on whether an error is returned. For some reason there is an error being returned now in the `category` query, so it renders even though there is cached data to render.

dpatil-magento:
QA Approved.

PR #3024, opened 2021-02-25
Increase test coverage for `packages/peregrine/lib/talons/CheckoutPage`

jcalcaben:
## Description

Increase coverage for files under `packages/peregrine/lib/talons/CheckoutPage`

**NOTE:**
_The uncovered lines in the report represent branching logic paths that appear to be unreachable._

Line 12 in `CheckoutPage/AddressBook/useAddressCard.js`: Address cannot be undefined because the destructuring in Line 5 would cause an error before it could reach Line 12.

Line 132 in `CheckoutPage/PaymentInformation/useCreditCard.js`: Cannot test the ( `false` || `true` ) path for this line since the only time `dropInLoading` is set to `false`, the `stepNumber` is always set to 0.

## Related Issue

Closes PWA-996

## Acceptance

any developer

### Verification Stakeholders

any developer

### Verification Steps

1. Run tests for this specific directory: `yarn test --collectCoverageFrom="packages/peregrine/lib/talons/CheckoutPage/**/*.js" packages/peregrine/lib/talons/CheckoutPage/`
2. Verify acceptable test coverage

## Checklist


sirugh:
I'm seeing some test failures when running `yarn test:dev` but also the command you gave in your verification steps. Can you double check these please :)

sirugh:
The untestable places seem like refactor targets if they're easy enough to fix up and the logic is identical. What do you thnink?

larsroettig (dismissed):

sirugh (approved):
Would be good to do a quick merge from dev to ensure the test coverage remains. Also had a thought about mocking queries that generally applies to any places you used that style in these tests. Regardless, I'll approve now and let you decide how to proceed.

sirugh (approved):
Bueno

PR #3025, opened 2021-02-25
Added ability to use custom template for scaffolding.

revanth0212:
## Description

This PR adds the ability to request the `scaffolding` program to use a custom template/version while creating a new project.

image

image

## Related Issue
Closes PWA-483

## Acceptance
Should be able to scaffold a new project using any compatible template on NPM.

### Verification Stakeholders
@jimbo @dpatil-magento 

### Specification
Need to update docs the mention the same.

### Verification Steps

Note: This won't work yet with `yarn create @magento/pwa` because we did not publish the new `@magento/create-pwa` yet. So let's test this using the new unpublished version locally.

1. Go to a folder where you would like to scaffold the new project.
2. In terminal run `node PATH_TO_PWA_STUDIO/packages/create-pwa/bin/create-pwa`
3. Answer all the questions.
4. For the question `Which template would you like to use to bootstrap test? Defaults to venia-concept.` enter any version from [NPM](https://www.npmjs.com/package/@magento/venia-concept). For instance: `@magento/venia-concept@9.0.1-alpha.3`. If you don't provide anything, it would check in the file system for `venia-concept`. If it can not find that, then it would search for a `venia-concept` in the local machine cache and use that instead.

## Checklist


larsroettig (approved):
- :heavy_check_mark: CodeRevbiew
- :heavy_check_mark: TestCoverage
- :heavy_check_mark: Mannual QA

packages/create-pwa/lib/index.js
```diff
@@ -96,6 +96,12 @@ module.exports = async () => {
                 return author;
             }
         },
+        {
+            name: 'template',
+            message: ({ name }) =>
+                `Which template would you like to use to bootstrap ${name}? Defaults to venia-concept.`,
+            default: 'venia-concept'
```
sirugh, line 103: Is there a difference between `@magento/venia-concept` and `venia-concept`? If not maybe we can use the namespaced version since it is similar to how a user would define by version

sirugh (changes requested):
I'd like to see the default be `@magento/venia-concept` if possible. But this is great and allows people to target a specific template version! So simple...why didn't we do this before 🤣

packages/create-pwa/lib/index.js
revanth0212: We are maintaining an [array](https://github.com/magento/pwa-studio/blob/develop/packages/pwa-buildpack/lib/cli/create-project.js#L16) which will be [matched against first](https://github.com/magento/pwa-studio/blob/develop/packages/pwa-buildpack/lib/cli/create-project.js#L77). If that doesn't work, that's when the code looks to pull the package from NPM. The object that we have in that array is looking for `venia-concept`. You are right, what you mentioned is what the code does but right now the way the code is laid out works this way:

```
if template === 'venia-concept`
    // use local mapping to find the local venia-concept package
else if template === anything other than `venia-concept`
   // check inside mac cache and if it does not find that package, fetch from NPM
   // if template is `@magento/venia-concept` use the latest stable version
   // if template has a version string attached, use the package with that particular version.
```

If we change the default to `@magento/venia-concept` instead, we won't be able to use local packages while developing. There definitely will be a way to do that, but do we need to spend more time on that? I did think about what you mentioned but opted out personally, but wanted to check with the team. Thanks for bringing this up buddy. Good question.

sirugh: We discussed in dev time and I think we want to be pretty explicit here to avoid the "magic" that helped obscure the release bug. 

Basically the target template input text should be the deployed name of the package, as if you were doing a `yarn add` ie https://classic.yarnpkg.com/en/docs/cli/add/. So in this case, defaulting to `@magento/venia-concept` is fine, but it should default to fetch the latest release version from remote, _not_ local. To use local we would be explicit such as `template: file:/path/to/local/folder`.

I think the team had input so if you want to start a thread on slack about it, let's do it!

packages/create-pwa/lib/index.js
```diff
@@ -147,40 +153,44 @@ module.exports = async () => {
...
-        stdio: 'inherit'
-    });
+        const buildpackBinLoc = resolve(
+            require.resolve('@magento/pwa-buildpack'),
+            '../../bin/buildpack'
+        );
+        await execa.shell(`${buildpackBinLoc} ${argsString}`, {
+            stdio: 'inherit'
+        });
+    } catch (e) {
```
revanth0212, line 193: This is just a small change. Moved catch statement to the end of the function.

packages/create-pwa/lib/index.js
revanth0212: Done, handled.

packages/create-pwa/lib/index.js
sirugh: Thanks for the call out! You can see this by appending `?w=1` :) https://github.com/magento/pwa-studio/pull/3025/files?w=1

revanth0212: Good to know 👍

packages/pwa-buildpack/lib/cli/create-project.js
```diff
@@ -111,8 +100,7 @@ module.exports.builder = yargs =>
         .options({
             template: {
                 describe:
-                    'Name of a "template" to clone and customize. Currently only the "venia-concept" template is supported: `buildpack create-project --template venia-concept`',
-                choices: ['venia-concept']
+                    'Name of a "template" to clone and customize. Currently only the "venia-concept" template is supported: `buildpack create-project --template venia-concept`'
```
sirugh, line 103: Should we change this to the following? I don't know if `buildpack create-project --template venia-concept` would work anymore now that we resolve relative path or npm deployment.
```suggestion
                    'Name of a "template" to clone and customize. Currently only the "venia-concept" template is supported: `buildpack create-project --template "@magento/venia-concept"`'
```

sirugh (dismissed):
Preliminary approval so we can get this in QA. The last of my comments have really been minor string changes which I'd like to see updated but no sense in delaying things.

sirugh (dismissed):

dpatil-magento:
@revanth0212 Project built with local (packages/venia-concept) repo has errors after launching Venia and also storybook build fails.

dpatil-magento:
QA Approved.

sirugh (dismissed):

dpatil-magento (approved):

sirugh (approved):

PR #3026, opened 2021-02-26
Update asyncActions.js

indranil-m:
The cart context should get the details if any error occurs

## Description

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist

### DIFF
--- a/packages/peregrine/lib/store/actions/cart/asyncActions.js
+++ b/packages/peregrine/lib/store/actions/cart/asyncActions.js
@@ -347,10 +347,11 @@ export const getCartDetails = payload => {
         try {
             const { data } = await fetchCartDetails({
                 variables: { cartId },
-                fetchPolicy: 'network-only'
+                fetchPolicy: 'no-cache',
+                errorPolicy: 'all'
             });
             const { cart: details } = data;
-
+            details.items = details.items.filter(function (e) {return e != null;});
             dispatch(actions.getDetails.receive({ details }));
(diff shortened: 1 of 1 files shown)


sirugh:
Hello @indranil-m, would you please open an issue describing the problem and link it here as the PR template suggests? We will also need repro steps for the bug (what is wrong and how did you get it to occur) as well as steps we can use to check that this code fixes that problem.

indranil-m:
Hi,

The issue is whenever some cart related error occurs like qty out of max
limit or out of scope etc at that time cartdetails will raise an exception
and it will go to error block.but if we do errorPolicy all then it will
throw the data in details object.otherwise details object will contain cart
id and in pwa we cant show the cart data

sirugh (commented):
Sorry, I don't understand the issue. We currently handle errors [here](https://github.com/magento/pwa-studio/blob/df8720144f6e0e275c30d4adc3b8feb79e776394/packages/venia-concept/src/index.js#L54-L90) and intentionally log and do not throw. I'd suggest opening an issue with a bug report so we can discuss what the problem is :)

sirugh:
Closing this. In the future please use the BUG issue template to report any issues you might have found. This helps us assess and prioritize issues that arise. In this case, because there are no repro steps or an "expected vs actual", I can't verify.

PR #3030, opened 2021-03-02
Bugfix/breadcrumbs

### LINKED ISSUE #3029: [bug]: Breadcrumbs always return .html suffix
Because an empty string "" is a falsy in javascript it will always return .html as the suffix.

**See:**
https://github.com/magento/pwa-studio/blob/0573cfca9e6c6d3b9295b61032c217e537f3530d/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js#L47

**We can just change it to:**
```
const categoryUrlSuffix = data?data.category.url_suffix:'.html';
```

Jordaneisenburger:
## Description

Just changed the short circuit to a ternary operator

## Related Issue
Closes #3029.

### Verification Stakeholders
@sirugh 
### Verification Steps

1. Go to pdp or category page and check breadcrumbs still work

## Checklist

* I have ran tests and passed

### DIFF
--- a/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
+++ b/packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
@@ -43,7 +43,7 @@ export const useBreadcrumbs = props => {
         nextFetchPolicy: 'cache-first'
     });
 
-    const categoryUrlSuffix = data?data.category.url_suffix:'.html';
+    const categoryUrlSuffix = data ? data.category.url_suffix : '.html';
 
     // When we have breadcrumb data sort and normalize it for easy rendering.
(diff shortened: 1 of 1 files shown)


packages/peregrine/lib/talons/Breadcrumbs/useBreadcrumbs.js
```diff
@@ -43,7 +43,7 @@ export const useBreadcrumbs = props => {
         nextFetchPolicy: 'cache-first'
     });
 
-    const categoryUrlSuffix = (data && data.category.url_suffix) || '';
```
sirugh, line 46: Wait, did we already fix this? `|| ''` means it'll use an empty string even if the configured suffix is `''`...

PR #3041, opened 2021-03-03
PWA-989: Real Data Wishlists and Wishlist Create

eug123:
## Description
- Hook up the Wishlist page to pull actual real data from GraphQL (remove all the mock data)
- Hook up the Create Wishlist button on the Favorites / Wishlist page so that it actually fires the `createWishlist` GQL mutation (and updates the page)

## Related Issue

Closes PWA-989

## Acceptance
- Wishlist page should use real data from backend
- Should be possible to create new wishlist, when new wishlist created - it appears on the page

### Verification Stakeholders
@dpatil-magento
### Verification Steps

 1. Create a Private wish list and check the UI.
 2. Create a Public wish list and check the UI.
 3. Verify existing list are shown when user log-in.
 4. Verify Private and Public lists with Max and Min characters and check the UI.
 5. Verify functionality of Cancel Button
 6. Verify wishlists are not accessible for guest users.
 7. Verify Auth user can access wish list of their account only.
 8. Verify UI on other browsers and devices (only if any UI changes)
 9. Run MFTF

## Checklist


packages/peregrine/lib/talons/WishlistPage/createWishlist.gql.js
```diff
@@ -0,0 +1,38 @@
+import { gql } from '@apollo/client';
+
+export const CREATE_WISHLIST = gql`
+    mutation createWishlist(
+        $name: String!
+        $visibility: WishlistVisibilityEnum!
+    ) {
+        createWishlist(input: { name: $name, visibility: $visibility }) {
```
tjwiebell, line 8: Minor preference that makes changes easier later, model your form after the mutation input, and just pass it as-is.

```suggestion
    mutation CreateWishlist(
        $input: CreateWishlistInput!
    ) {
        createWishlist(input: $input) {
```

packages/peregrine/lib/talons/WishlistPage/useCreateWishlist.js
```diff
@@ -1,12 +1,45 @@
...
+            update(
+                cache,
+                {
+                    data: { createWishlist }
+                }
+            ) {
+                const { customer } = apolloClient.readQuery({
+                    query: getCustomerWishlistsQuery
+                });
+                apolloClient.writeQuery({
```
tjwiebell, line 29: This manual cache write ends up just triggering the wishlistPage query to refetch, which is a lot more straight forward code wise, and you have the ability to stall completion of the mutation until the refetch comes back. Consider it optional since this works and that refetch doesn't appear to cause any re-render, but you can find an example of that [in the addressBook component](https://github.com/magento/pwa-studio/blob/870cb896919f67effe21ef37f2f498867b2193cf/packages/peregrine/lib/talons/AddressBookPage/useAddressBookPage.js#L121).

packages/peregrine/lib/talons/WishlistPage/useCreateWishlist.js
```diff
@@ -16,18 +49,36 @@ export const useCreateWishlist = () => {
...
-        setIsModalOpen(false);
-    }, []);
+    const handleCreateList = useCallback(
+        async data => {
+            try {
+                await createWishlist({
+                    variables: {
+                        name: data.name,
+                        visibility: data.visibility
+                    }
```
tjwiebell, line 59: Paired with mutation input change:

```suggestion
                    variables: {
                        input: data
                    }
```

tjwiebell (changes requested):
Some minor signature improvements to future proof, but overall looks great. I did run into an issue now that this is hooked up to live data.

If you create a brand new Customer, it appears the first wishlist is not automatically created via GraphQL, so the `/wishlist` page hangs. Likely an existing bug where some assumptions were made with mocked data that aren't correct. It does appear Luma will automatically create your first list for you, so we'll need to check with product if that's something Venia should do as well, or just render the page with no lists.

packages/peregrine/lib/talons/WishlistPage/createWishlist.gql.js
eug123: Changed.
Thank you

packages/peregrine/lib/talons/WishlistPage/useCreateWishlist.js
eug123: Thank you for advice! Changed

packages/peregrine/lib/talons/WishlistPage/useCreateWishlist.js
eug123: Done

packages/peregrine/lib/talons/WishlistPage/useCreateWishlist.js
```diff
@@ -16,18 +26,38 @@ export const useCreateWishlist = () => {
...
-        setIsModalOpen(false);
-    }, []);
+    const handleCreateList = useCallback(
+        async data => {
+            try {
+                await createWishlist({
+                    variables: {
+                        input: data
+                    },
+                    refetchQueries: [{ query: getCustomerWishlistsQuery }],
```
tjwiebell, line 36: Right implementation, wrong query 😄  this ended up adding another network round trip. Here's a screenshot of network requests with that mutation, which additionally showcases why we should be giving operations unique names as well.

Screen Shot 2021-03-12 at 11 33 31 AM

This will be a little more involved, but totally worth it.

1. Migrate `wishlistPage.gql.js` to Peregrine. Update `useWishlistPage` talon to use the `defaultOperations` merge pattern.
2. `import { GET_CUSTOMER_WISHLIST } from './wishlistPage.gql'`
3. Use that query here
4. Remove `GET_CUSTOMER_WISHLISTS` from `createWishlist.gql.js`
5. Additionally, you can scale down the response of `CREATE_WISHLIST` to just `id`, since we expect to refetch here anyway

packages/venia-ui/lib/components/WishlistPage/wishlist.js
```diff
@@ -28,14 +28,18 @@ const Wishlist = props => {
...
+    const visibilityLabel =
+        visibility === 'PUBLIC'
+            ? formatMessage({
+                  id: 'wishlist.publicText',
+                  defaultMessage: 'Public'
+              })
+            : formatMessage({
+                  id: 'wishlist.privateText',
+                  defaultMessage: 'Private'
+              });
```
tjwiebell, line 40: Should we use the globals here instead?

```suggestion
    const visibilityLabel =
        visibility === 'PUBLIC'
            ? formatMessage({
                  id: 'global.public',
                  defaultMessage: 'Public'
              })
            : formatMessage({
                  id: 'global.private',
                  defaultMessage: 'Private'
              });
```

tjwiebell (changes requested):
Clarified some points about refetch logic, this is real close. Ping me on Slack if you have any questions.

packages/peregrine/lib/talons/WishlistPage/useCreateWishlist.js
eug123: Have not checked network(. Corrected it.

Thank you

packages/venia-ui/lib/components/WishlistPage/wishlist.js
eug123: Yes it makes sense for this phrase "global", and these global phrases already used in other wishlist component)
Fixed.
Thank you

tjwiebell (dismissed):

packages/venia-ui/lib/components/WishlistPage/createWishlist.ee.js
```diff
@@ -63,6 +63,51 @@ const CreateWishlist = props => {
         }
     ];
 
+    const errors = useMemo(() => {
+        const translatedErrors = [];
+        Array.from(formErrors.values()).forEach(error => {
```
sirugh, line 68: I thought we were leaving error translation to the back end.

supernova-at:
## QA

> If you create a brand new Customer, it appears the first wishlist is not automatically created via GraphQL, so the /wishlist page hangs.

This is happening for me with an existing customer (who I assume does not have a wishlist).

sirugh:
Is this coded to assume at least one exists? May not work since there is that bug where customers don't always have one.

eug123:
@supernova-at, can not reproduce this, just created new customer here https://pr-3041.pwa-venia.com/, 
@sirugh, no it will just show page without any list with "Create list button"

dpatil-magento:
@eug123 Please check following - 

1. Once I create  wishlist and hit create list again then previous list selections are pre-populated.

<a href="https://gyazo.com/f12e49e246291249a7ed58911239d88f">Image from Gyazo</a>

2. Created wishlist for customer does not exists in Admin UI > Customer > Wishlist panel.

3. On CE backend, still error exists, was expecting this to work -

image

eug123:
@dpatil-magento,
1. Fixed.
2. Wishlist names appears in the grid on the admin customer page when there are products have been added to those lists
3. Fixed errors, but now if there was no items added previously to wishlist - it is just an empty page on CE

Thank you

packages/peregrine/lib/talons/WishlistPage/__tests__/useCreateWishlist.spec.js
```diff
@@ -1,57 +1,121 @@
...
+                    ]
+                }
+            }
+        };
+    }
+};
 
-        return root.findByType('i').props.talonProps;
-    };
+const createWishlistMock = {
```
sirugh, line 52: How do you test the loading state of this mutation? I'm trying to do this in my PR but I'm having trouble...

sirugh: Nvm, found it :D Just don't `await` the act'ed click.

dpatil-magento:
Thanks @eug123 those 2 fixes looks good.   Please check following - 
1. On Sign-Out Apollo cache still has wishlist info saved. I think wish list data should be cleared on sign-out given its customer specific. 



2. When `Enable Multiple Wish Lists` is not enabled then customer is not allowed to create one wishlist also. 



3. The `Enable Multiple Wish Lists` can be store specific but currently once I create on enabled store the same lists can be viewed on non-enabled store view also.

4. CE default has wishlist and not Create List button.

Don't think we need to extend the scope of this PR,  First issue is something I believe needs a fix but on other 3 observations lets Demo to @tkacheva  and create a follow up story around admin configs if needed.

eug123:
@dpatil-magento,
Issues 1 and 2 are fixed, please check.

Thank you

packages/peregrine/lib/Apollo/clearWishlistDataFromCache.js
```diff
@@ -0,0 +1,13 @@
+import { deleteCacheEntry } from './deleteCacheEntry';
+
+/**
+ * Deletes references to Customer Wishlists from the apollo cache including *WishlistItem entries
+ *
+ * @param {ApolloClient} client
+ */
+export const clearWishlistDataFromCache = async client => {
```
tjwiebell, line 8: I'm not sure about another utility function here, since I would consider wishlist part of customer data. Fine to keep this, but would be better to have `clearCustomerDataFromCache` call it.

Alternatively, there is a data security pattern we might want to establish. The `clearCustomerDataFromCache` will clear entities that start with `Customer` and I think we've just gotten lucky so far that they followed this pattern. You could instead define `keyFields` for Wishlist and WishlistItem, and this extra utility function wouldn't be needed.

```js
Wishlist: {
    keyFields: ({ id }) => `CustomerWishlist:${id}`
},
WishlistItem: {
    keyFields: ({ id }) => `CustomerWishlistItem:${id}`
},
```

tjwiebell (changes requested):
Let's simplify the new utility as much we can, but otherwise the rest looks good 👍  ping me if you have questions.

packages/peregrine/lib/Apollo/clearWishlistDataFromCache.js
eug123: Changed according your recommendations.
Thank you!

tjwiebell (dismissed):
Changes look perfect, nice catch covering all the WishlistItem implementations. It does look like [Coveralls is dinging us](https://coveralls.io/builds/38177312/source?filename=packages%2Fperegrine%2Flib%2FApollo%2Fpolicies%2Findex.js#L248) a tiny bit for not testing the params of these new keyField functions, but not a blocker for approval.

tjwiebell (approved):

dpatil-magento:
QA Approved.  Currently unable to check when multiple wishlist is disabled how the wishlist gonna look but will test it in pr 3048

packages/extensions/venia-sample-language-packs/i18n/fr_FR.json
```diff
@@ -350,8 +350,6 @@
     "validation.mustBeChecked": "Doit être vérifié.",
     "validation.validatePassword": "Un mot de passe doit contenir au moins 3 des éléments suivants: minuscules, majuscules, chiffres, caractères spéciaux.",
     "wishlist.emptyListText": "Il n'y a actuellement aucun élément dans cette liste",
-    "wishlist.privateText": "Privée",
-    "wishlist.publicText": "Publique",
```
evanc420, line 354: packages/extensions/venia-sample-language-packs/i18n/fr_FR.json

PR #3043, opened 2021-03-04
[DO NOT MERGE] Cypress POC

dpatil-magento:
## Description

POC for writing Functional tests with cypress.

## Related Issue

Closes PWA-1434.

## Acceptance

Team should agree with using Cypress for Venia Functional Tests. Also add any improvements/best practices/Suggestions to below TO-DO list.

## TO-DO - 
1. Create alias for Dom elements in one place and call in scripts as needed.
2. Create utility file for commonly used methods like UID generation.
3. Need to built utility function which can wait until all elements of the page loads before going to next step.
4. Remove Cypress Example tests
4. Create a wiki page with information.

### Verification Stakeholders
PWA Team
### Verification Steps

1. Check instructions in README.md under /venia-e2e-tests folder.
2. Verify able to Venia run demo tests /cypress/integration/real-data-tests/ and  /cypress/integration/mock-data-tests/

## Checklist


davemacaulay:
@dpatil-magento this is so awesome! I've had a play locally and it works super well.

Out of interest why do we include the `examples` folder? Is this to verify the browser isn't being silly or just part of the PoC?

venia-e2e-tests/cypress/integration/examples/aliasing.spec.js
```diff
@@ -0,0 +1,39 @@
+/// <reference types="cypress" />
```
revanth0212, line 1: We can avoid pushing the examples folder all together.

revanth0212 (approved):
Nice work dude. Let's clean this up and create a new PR with all the setup work and 1 sample test included.

venia-e2e-tests/cypress.json
```diff
@@ -0,0 +1,8 @@
+{
+  "baseUrl": "https://develop.pwa-venia.com/",
```
davemacaulay, line 2: Is there a way we can make this react to the current instances URL? For instance when running the tests in CI against a deployed PR we'll need to modify this URL.

revanth0212: It has to be configurable. Definitely doable.

PR #3047, opened 2021-03-05
Ignoring CE and EE gql files based on Mangeto edition.

revanth0212:
## Description

Ignore CE and EE files based on `env.MAGENTO_BACKEND_EDITION`.

## Related Issue
Closes PWA-843

## Acceptance
`yarn run build:dev` should work fine.

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Checkout this branch.
2. Use a valid CE Magento backend and set MAGENTO_BACKEND_EDITION to CE. Run `yarn build:dev`, it should work fine.
3. Use a valid EE Magento backend and set MAGENTO_BACKEND_EDITION to EE. Run `yarn build:dev`, it should work fine.

image

## Checklist


tjwiebell (approved):
Looks good 👍  could use a test but not blocking approval.

dpatil-magento:
@revanth0212 I still see error after setting `MAGENTO_BACKEND_EDITION=CE` and using CE backend then running `yarn build:dev` 

```
  18:9  error  Cannot query field "giftCardAccount" on type "Query"                                                                          graphql/template-strings
  32:9  error  Cannot query field "applyGiftCardToCart" on type "Mutation". Did you mean "applyCouponToCart"?                                graphql/template-strings
  51:9  error  Cannot query field "removeGiftCardFromCart" on type "Mutation". Did you mean "removeItemFromCart" or "removeCouponFromCart"?  graphql/template-strings
```

packages/venia-concept/.graphqlconfig
```diff
@@ -11,11 +11,15 @@
...
+                    "magentoBackendEdition": "${env:MAGENTO_BACKEND_EDITION}",
                     "filesGlob": "../{peregrine,venia-ui,venia-concept}/{lib,src}/**/*.{js,graphql,gql}",
+                    "ceFilesGlob": "../{peregrine,venia-ui,venia-concept}/{lib,src}/**/*.ce.js",
+                    "eeFilesGlob": "../{peregrine,venia-ui,venia-concept}/{lib,src}/**/*.ee.js",
                     "ignore": [
                         "../venia-ui/lib/**/giftOptions.js",
                         "../venia-ui/lib/**/wishlistPage.gql.js",
-                        "../venia-ui/lib/**/wishlistItem.gql.js"
+                        "../venia-ui/lib/**/wishlistItem.gql.js",
+                        "../venia-ui/lib/**/giftCardQueries.js"
```
revanth0212, line 22: Not sure why the GraphQL validator is failing on this file. The queries and mutations used in this file are valid ones according to the schema docs but for some reason, the validator doesn't like them. So ignoring them for now.

tjwiebell: GiftCards are an EE only feature. It looks like what we did here is selectively render `CartPage/PriceAdjustments/giftCardSection`, so the that queries file didn't need to have the flavor extension. I think it would be preferable here to maybe rename `giftCardQueries.js => giftCardQueries.ee.js` to keep validation working for EE.

tjwiebell (changes requested):

packages/venia-concept/.graphqlconfig
tjwiebell: Didn't consider that rename is going to make editors think this file doesn't exist now. Do we have a pattern to solve this? Do we require that the non flavored file must also exist?

Screen Shot 2021-03-16 at 2 06 57 PM

revanth0212: I think it is editor-specific. I don't see it happening on VS Code. Right now we just create files with flavors attached (CE or EE) but we don't have non-flavored ones to accommodate for editors. If we want to solve this, it has to be its own ticket.

image

tjwiebell (dismissed):
LGTM 👍  just needs tests and this is ready to go.

tjwiebell (approved):

dpatil-magento:
QA Approved.

PR #3049, opened 2021-03-10
[PWA-632] My Account: Wishlist: Actions Menu and Edit Flow

anthoula:
## Description

Add ability to edit a wishlist's name and visibility (public or private). Add additional action menu panel.

## Related Issue

 
[PWA-632](https://jira.corp.magento.com/browse/PWA-632)

### Verification Stakeholders

@dpatil-magento 

### Verification Steps

1. Sign in to account
2. Go to Favorites List
3. Click on action menu kebob for a wishlist
4. See new action menu options (Edit)
5. Click Edit and edit name/visibility
6. Name and visibility of wishlist updates

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/WishlistPage/__tests__/useWishlist.spec.js
+++ b/packages/peregrine/lib/talons/WishlistPage/__tests__/useWishlist.spec.js
@@ -3,25 +3,6 @@ import { act } from 'react-test-renderer';
 
 import createTestInstance from '../../../util/createTestInstance';
 import { useWishlist } from '../useWishlist';
-import { useMutation } from '@apollo/client';
-
-jest.mock('@apollo/client', () => {
-    return {
-        ...jest.requireActual('@apollo/client'),
-        useMutation: jest.fn(() => [
-            jest.fn(),
-            {
-                error: false,
-                loading: false
-            }
-        ])
-    };
-});
-
-jest.mock('../wishlist.gql', () => ({
-    getCustomerWishlistQuery: jest.fn().mockName('getCustomerWishlistQuery'),
-    updateWishlistMutation: jest.fn().mockName('updateWishlistMutation')
-}));
 
 const Component = () => {
     const talonProps = useWishlist();


packages/venia-ui/lib/components/WishlistPage/__tests__/wishlistEditFavoritesListDialog.spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+jest.mock('../../../classify');
+jest.mock('../../Dialog', () => 'Dialog');
+jest.mock('../../FormError', () => 'FormError');
+
+const props = {
+    formErrors: new Map([]),
+    isOpen: false,
+    isEditInProgress: false,
+    onCancel: jest.fn(),
```
tjwiebell, line 14: We can squeeze a little extra value out of snapshots by naming these mocks.

```suggestion
    onCancel: jest.fn().mockName('onCancel'),
```

tjwiebell (changes requested):
Posting what I have so far until #3041 gets us hooked up to live data. This is mostly there, really nice work 👍

packages/venia-ui/lib/components/WishlistPage/wishlist.js
```diff
@@ -62,6 +83,19 @@ const Wishlist = props => {
                         {contentToggleIcon}
                     </button>
                 </div>
+                <WishlistListActionsDialog
```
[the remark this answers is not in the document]
tjwiebell: Team just had [a conversation about this](https://magento.slack.com/archives/G8N21AQTV/p1615564627083800), and it sounds like Informed DynamicFields could be a solution here.

packages/peregrine/lib/talons/WishlistPage/wishlist.gql.js
```diff
@@ -0,0 +1,23 @@
...
+        $name: String!
+        $visibility: WishlistVisibilityEnum!
+        $wishlistId: ID!
+    ) {
+        updateWishlist(
+            name: $name
+            visibility: $visibility
+            wishlistId: $wishlistId
+        ) {
+            name
```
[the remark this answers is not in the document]
tjwiebell: #3041 is in QA, so hopefully is merged soon. Consider creating a `WishlistItemInterface` fragment out of [this query](https://github.com/magento/pwa-studio/pull/3041/files#diff-5959ac5cbfcf50ed26cc295c80fe0db6da23e18b60e273df9e3652a733ba124f), and re-using that fragment in this response. The goal here being to remove any unnecessary network calls after this mutation.

packages/venia-ui/lib/components/WishlistPage/wishlist.js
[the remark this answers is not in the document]
anthoula: I have looked into DynamicFields to be able to turn this into a single Dialog, but it looks as though that option will not allow some flexibility needed to customize the Dialog header title, header actions, and mask actions between the two Dialogs currently in place. Considering this, and parity with `wishlistItem`, I will leave this pattern as is for now, but can re-visit if team decides otherwise.

packages/peregrine/lib/talons/WishlistPage/wishlist.gql.js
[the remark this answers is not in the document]
anthoula: Discussed and deduced that doing a refetch of wishlist data is necessary in this use case, to update the UI.

packages/peregrine/lib/talons/WishlistPage/__tests__/__snapshots__/useWishlist.spec.js.snap
```diff
@@ -1,9 +1,137 @@
...
+exports[`it handles editing a wishlist handleEditWishlist() runs the mutation with a thrown error 1`] = `
+Array [
+  [Error: Wish list "Favorites" already exists.],
+]
+`;
+
+exports[`it handles editing a wishlist handleEditWishlist() runs the mutation with the expected values 1`] = `
+Object {
+  "awaitRefetchQueries": true,
+  "refetchQueries": Array [
```
tjwiebell, line 12: I ran into something similar in a snapshot recently, and wasn't a fan of including the entire GraphQLDocument AST in the snap. It would be possible to mock here since this test doesn't make use of the `MockedProvider`. This may be personal preference so consider this suggestion optional.

```js
jest.mock('../wishlist.gql', () => ({
    getCustomerWishlistQuery: jest.fn().mockName('getCustomerWishlistQuery'), 
    updateWishlistMutation: jest.fn().mockName('updateWishlistMutation')
}));
```

tjwiebell (approved):
Looks great, nice work 👍  One tiny suggestion for test improvement, but not blocking sending this to QA.

tjwiebell (dismissed):

dpatil-magento:
@anthoula EE looks good but CE wishlist is broken.

image

tjwiebell (approved):

dpatil-magento:
@anthoula Given CE support only one list there should not be Edit option available?   No specific requirement, so please check with @tkacheva.  Might be added scope/Bug or a follow up ticket.

image

anthoula:
> @anthoula Given CE support only one list there should not be Edit option available? No specific requirement, so please check with @tkacheva. Might be added scope/Bug or a follow up ticket.

Checking with @tkacheva on possible solution for CE.

sirugh:
Would it be sufficient to just not display the `Edit` call to action/button in CE?

anthoula:
> Would it be sufficient to just not display the `Edit` call to action/button in CE?

We have proposed to @tkacheva to hide the entire actions menu in CE, as there's limited functionality, and "Edit" is the only action we have right now.

tkacheva:
@anthoula, @sirugh and @schensley I agree on not showing edit option for CE customers. The only wishlist they have as a default name that we never show on the storefront.

packages/venia-ui/lib/components/WishlistPage/wishlist.js
```diff
@@ -71,38 +71,43 @@ const Wishlist = props => {
...
     );
 
-    const formProps = {
-        initialValues: {
-            name: name,
-            visibility: visibility
-        }
-    };
+    // wishlist actions are limited in CE
+    const actionMenu = name ? (
```
sirugh, line 75: This comment and conditional don't quite explain _why_ we check for `name` before showing the action menu.

I might suggest following the pattern [here](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CartPage/PriceAdjustments/giftCardSection.js) where we have two components - one renders `null` if CE and the other renders a component as normal if EE. You could do the same for an `ActionMenu` component, for example.

You could then just omit the conditional and always include the `<ActionMenu />` component passing props as necessary.

Obviously that is a lot more than just using this conditional though, so I don't know if I'd say you _must_ do this. At a minimum, perhaps you could update the comment to clarify why the existence of `name` means the backend is EE :)

sirugh: But now that I think a bit more on it - using presence of some value like this seems fragile. What if we end up adding `name` to the schema for CE?

packages/peregrine/lib/talons/WishlistPage/useActionMenu.js
```diff
@@ -0,0 +1,93 @@
...
+                await updateWishlist({
+                    variables: {
+                        name: data.name,
+                        visibility: data.visibility,
+                        wishlistId: id
+                    },
+                    refetchQueries: [{ query: getCustomerWishlistQuery }],
+                    awaitRefetchQueries: true
+                });
+                await setCurrentDialog(dialogs.NONE);
```
sirugh, line 59: As far as I know this is async but awaiting doesn't do anything. May as well remove the `await`.

sirugh (dismissed):
This is much cleaner now. The `wishlist.js` component no longer has a bunch of extra code in the talon. Instead, that responsibility is passed down to the backend-specific component, so the CE wishlist page will be much lighter in that respect.

Good job! I thought it would be more difficult :)

sirugh (approved):

dpatil-magento:
QA Approved.

Issue #3050, opened 2021-03-10
[feature]: Extension's ability manipulate Venia-Ui JSX


abrhim:
**Is your feature request related to a problem? Please describe.**
Extensions have limited access to extending the JSX markup in components. In Product Recommendations' use case, we would like to add a `onClick` handler to <Link> tags inside of the `GalleryItem` component to enable rendered recommendation product tracking. 

**Describe the solution you'd like**
Short term: merge this pr  https://github.com/magento/pwa-studio/pull/3053
Long term: The issue with enabling extensions to target JSX in components in the same way that Storefronts can target and manipulate JSX in components that if any JSX is changed in a PWA-Studio release, it becomes a breaking change. To solve this make the targeting system class based. Inside of each component localized CSS classes are being used, these are essentially unique identifiers in the DOM tree and could be reused as a method to target JSX inside of components. Because the extension (and i would propose considering using this method for storefronts as well) doesn't rely upon the DOM tree inside of JSX, there will only be breaking changes in the codebase if a className is removed or changed, not when a new one is added. 

**Describe alternatives you've considered**
Alternative for adding props to Venia-ui components: Adding props catch alls to peregrine talons that are then spread across Venia-UI components. Extensions could then intercept the Talons and then change the prop catch alls to their desired props. This is limited and would not solve the issue that extensions need to extend all of the markup in Venia-UI components, not just the props. 

**Please let us know what packages this feature is in regards to:**
- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

PR #3053, opened 2021-03-10
Gallery item handle link click

abrhim:
## Description

Add link click handler to <Link> inside of GalleryItem. This is necessary for recommendation tracking for product recommendations. 

## Related Issue

Related to https://github.com/magento/pwa-studio/issues/3050. We need a long term solution for this problem, but this PR addresses the issue we need solved immediately. 

closes JIRA-2857

### Verification Steps

1. in an example extension wrap useGalleryItem and override handleLinkClick with another function.
2. Verify function runs.


packages/peregrine/lib/talons/Gallery/__tests__/useGalleryItem.spec.js
```diff
@@ -0,0 +1,8 @@
+import { useGalleryItem } from '../useGalleryItem';
+test('returns object handelLinkClick as a function', () => {
+    const expectedReturnValue = {
+        handleLinkClick: () => {}
+    };
+    const props = useGalleryItem({});
+    expect(JSON.stringify(props)).toBe(JSON.stringify(expectedReturnValue));
```
supernova-at, line 7: Since this test is specifically looking for `handleLinkClick` and wants it to be a `function`, you could use [objectContaining](https://jestjs.io/docs/expect#expectobjectcontainingobject) combined with [expect.any](https://jestjs.io/docs/expect#expectobjectcontainingobject) for this.

```js
expect(props).toEqual(
  expect.objectContaining({
    handleLinkClick: expect.any(Function)
  })
)
```

Doesn't block a 👍 , just more of an fyi :)

supernova-at (dismissed):

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
```diff
@@ -0,0 +1,4 @@
+export const useGalleryItem = props => {
+    const handleLinkClick = () => {};
```
supernova-at, line 2: I mean this is a little strange of course. I'm assuming your extension overrides this?

abrhim: Correct :)

packages/peregrine/lib/talons/Gallery/__tests__/useGalleryItem.spec.js
abrhim: Ah ha! i was struggling for some way other than `JSON.stringify` to verify `handleLinkClick` was a function. Thanks :)

davemacaulay:
@magento export issue to JIRA project PWA as Story

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
sirugh: If we're going to attach a click handler we may as well attempt to invoke some sort of callback that may be passed, incase someone wanted to implement `GalleryItem` and pass an `onClick` handler.

```suggestion
    const { onClick = () => {} } = props;
    const handleLinkClick = useCallback((e) => {
        if (onClick) onClick(e);
    }, [onClick]);
```

sirugh: I'm also interested in what @jimbo has to say about this - we haven't yet implemented a function for the sake of having a target to wrap, so this would be the first.

packages/peregrine/lib/talons/Gallery/__tests__/useGalleryItem.spec.js
abrhim: Change has been made.

abrhim:
@jimbo  after pulling in latest from `develop` there seems to be a failing test: https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/AddressBookPage/__tests__/addEditDialog.spec.js

The test fails because it is expecting a string but receiving a number in props for a SVG tag. Not sure why this is happening, but it outside the scope of the PR.

supernova-at (approved):

packages/peregrine/lib/talons/Gallery/useGalleryItem.js
supernova-at: > Since an extension that wraps a talon may include new properties and not just override existing properties, and since undefined may be passed to a React element's onClick, we really don't even need to define a noop function as the default value. This talon could literally return an empty object.

I'd much prefer this over returning a no-op function here.

supernova-at (changes requested):

abrhim:
@supernova-at your ask has been resolved.

jimbo (approved):

supernova-at (dismissed):

supernova-at (approved):

PR #3055, opened 2021-03-10
Scroll completed section to the top after every checkout step.

revanth0212:
## Description

Scroll the view to each completed section after every checkout step to allow the user to check the progress and the data they have entered.

After the Shipping Information step is done

image

image

After the Shipping Method step is done

image

image

## Related Issue
Closes PWA-723

### Verification Stakeholders
@schensley 
@dpatil-magento 

### Verification Steps
1. Go to the deployed app's checkout page.
2. After each step is completed, you should notice that the window should scroll to the top of the view.

## Checklist


packages/peregrine/lib/talons/CheckoutPage/__tests__/__snapshots__/useCheckoutPage.spec.js.snap
```diff
@@ -29,6 +29,12 @@ Object {
   "setPaymentInformationDone": [Function],
   "setShippingInformationDone": [Function],
   "setShippingMethodDone": [Function],
+  "shippingInformationRef": Object {
+    "current": undefined,
```
revanth0212, line 33: These are expected because these are refs that haven't been assigned to any components.

packages/peregrine/lib/context/__tests__/__snapshots__/header.spec.js.snap
```diff
@@ -0,0 +1,22 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`provides state and actions via context 1`] = `
+Array [
+  Array [
+    Object {
+      "headerRef": Object {
+        "current": null,
```
revanth0212, line 8: Expected because we have assigned the ref to a temp component that is not mounted.

packages/venia-ui/lib/components/ScrollAnchor/scrollAnchor.css
```diff
@@ -0,0 +1,15 @@
+.anchor {
+    position: absolute;
+    top: calc(var(--venia-global-header-minHeight) * -1px - 5px);
```
revanth0212, line 3: For some reason, this is not working. Calc and var don't seem to like each other. If I directly using `var(--venia-global-header-minHeight)` it seems to work but since our intention is use a negative offset, it isn't gonna work.

packages/venia-ui/lib/tokens.css
```diff
@@ -85,6 +85,8 @@
 
     /* dimensions */
     --venia-global-maxWidth: 1440px;
+    --venia-global-header-minHeight: 80px;
+    --venia-mobile-global-header-minHeight: 56px;
```
revanth0212, line 89: I am not sure how we name mobile-specific tokens. Team, what do your guys think?

schensley:
@revanth0212  demo'd this to me yesterday (03/22/2021) and it behaved as requested. - UX Approved. 
A new backlog story will need to be added to remove the "Sign in" module after the shopper proceeds as Guest (after step one).

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -146,6 +165,8 @@ export const useCheckoutPage = (props = {}) => {
 
     const setPaymentInformationDone = useCallback(() => {
         if (checkoutStep === CHECKOUT_STEP.PAYMENT) {
+            console.log('PM scrolling');
```
sirugh, line 168: oops

packages/venia-ui/lib/tokens.css
sirugh: We discussed this in dev time, and if I recall correctly you could use a single variable, `--venia-global-header-minHeight` and change it here in `tokens.css` using a media query. so that you don't need a media query in `scrollAnchor.css` or two variable names.

```
@media (max-width: 960px) {
    :global(:root) {
        --venia-global-header-minHeight: 56px;   
    }
}  
```

packages/venia-ui/lib/components/ScrollAnchor/scrollAnchor.css
```diff
@@ -0,0 +1,15 @@
+.anchor {
+    position: absolute;
+    top: calc(var(--venia-global-header-minHeight) * -1 - 5px);
+}
+
+/*
+ * Mobile-specific styles.
+ */
+
+@media (max-width: 960px) {
```
sirugh, line 10: See requested change in `tokens.css` :)

packages/venia-ui/lib/components/Header/header.css
```diff
@@ -61,7 +61,7 @@
     .toolbar {
         grid-template-areas: 'primary title secondary';
         grid-template-columns: auto auto 1fr;
-        min-height: 5rem;
+        min-height: var(--venia-global-header-minHeight);
```
sirugh, line 64: Should we switch this variable to `rem`? I believe it uses `px` after your change, but I didn't realize it was going to be used here.

packages/venia-ui/lib/tokens.css
sirugh: Also it looks like we might need three media queries as there are three different heights of the header, at least from my brief inspection:

[Image from Gyazo](https://gyazo.com/378804d88abf29d49368473e94d7847b)

packages/venia-ui/lib/components/Header/header.css
```diff
@@ -34,7 +34,7 @@
     grid-template-rows: 3rem;
     justify-items: center;
     max-width: var(--venia-global-maxWidth);
-    min-height: 3.5rem;
+    min-height: var(--venia-mobile-global-header-minHeight);
```
sirugh, line 37: Same comment re: single token/media queries.

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
sirugh: Remove!

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
```diff
@@ -132,12 +135,28 @@ export const useCheckoutPage = (props = {}) => {
         setReviewOrderButtonClicked(false);
     }, [setReviewOrderButtonClicked]);
 
+    const scrollShippingInformationIntoView = useCallback(() => {
+        if (shippingInformationRef.current) {
+            shippingInformationRef.current.scrollIntoView({
+                behavior: 'smooth'
+            });
+        }
+    }, [shippingInformationRef]);
```
sirugh, line 144: This callback is basically duplicated and will be for each step that gets this new treatment. Don't spend too much time optimizing, but I wonder if there is a way to make this more generic. We already do have a `done` callback but it is used in a slightly different way (refresh of page vs "submit" of mutation). Hmmm...

sirugh (changes requested):
This feels much better than the current experience!

Feedback left about the discussed changes to the token.

I added the `needs-tests` label to the issue - we will need to figure out how to test calling `onCompleted`.

Also, please update all the default props/propTypes!

packages/peregrine/lib/talons/CheckoutPage/useCheckoutPage.js
revanth0212: Yeah, the problem with the done callbacks is, they get called whenever the section is done, which includes page refresh as well. So I couldn't use them.

revanth0212:
> Also it looks like we might need three media queries as there are three different heights of the header, at least from my brief inspection:

Yep, that's the problem. We don't have a proper definition of mobile width, that's the reason why I dint do the `@media` part. Wanted to check with the team.

packages/venia-ui/lib/components/Header/header.css
revanth0212: All values in the tokens.css were px so I converted these to px as well. But after discussing with the team, I realized there is no standard that we follow at the moment for tokens so changed it back to `rem`. Nice question @sirugh 😄

sirugh:
> Yep, that's the problem. We don't have a proper definition of mobile width, that's the reason why I dint do the @media part. Wanted to check with the team.

I'd say, for purposes of THIS work, only the changes in header height matter. My vote is for using the three breakpoints as seen in the gif, along with only a single variable/token.

packages/venia-ui/lib/tokens.css
```diff
@@ -85,8 +85,13 @@
 
     /* dimensions */
     --venia-global-maxWidth: 1440px;
-    --venia-global-header-minHeight: 80px;
-    --venia-mobile-global-header-minHeight: 56px;
+    --venia-global-header-minHeight: 5rem;
+}
+
+@media (max-width: 960px) {
```
sirugh, line 91: So we want an additional one at 641px I believe.

revanth0212: I don't understand why we might need another one at 640. Wouldnt `960` cover for `640` as well? I didn't want to introduce new UI changes as part of this PR. I only wanted to move the existing hardcoded values to tokens.

sirugh: Ok I think I was confused - I thought the store switcher would affect the height, since it hides below 640px. So my concern was how we would handle the scrolling for widths between 640 and 960.

What I did not know was that we auto-hide the switcher when you scroll down, and only the header is sticky. Because of that, only the two possible header sizes matter, so you've covered those cases :)

sirugh (approved):
This is good to go, but needs `propTypes` and tests. Not holding up QA for it.

dpatil-magento:
@revanth0212 MFTF for edit checkout flow is falling, can reproduce manually too -

steps - 
1. Submit shipping and payment info and hot review
2. Try edit shipping address and save/update 

image

sirugh (approved):

dpatil-magento:
QA Approved.

PR #3061, opened 2021-03-16
Checkout page region field fix.

revanth0212:
## Description

The shipping information form in the checkout page has a bug where it does not pick up the region code provided in the cart page if the user does not have a default address assigned to their account. This PR addresses that issue.

## Related Issue
Closes PWA-671

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Create an account and get to the cart page after adding an item to the cart.
2. Estimate the shipping costs by giving the country, state, and zip code information.
3. Get to the checkout page, see if the region code is being carried over.
4. Perform the above steps on a country that does not support regions as well. For instance, United Kingdom.

United States

image

United Kingdom

image

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/Region/useRegion.js
+++ b/packages/peregrine/lib/talons/Region/useRegion.js
@@ -48,11 +48,12 @@ export const useRegion = props => {
     }, [country, regionInputFieldApi, regionSelectFieldApi]);
 
     const { data, error, loading } = useQuery(getRegionsQuery, {
-        variables: { countryCode: country }
+        variables: { countryCode: country },
+        skip: !country
     });
 
     let formattedRegionsData = [{ label: 'Loading Regions...', value: '' }];
-    if (!loading && !error) {
+    if (data && !loading && !error) {
         const { country } = data;
         const { available_regions: availableRegions } = country;
         if (availableRegions) {
--- a/packages/peregrine/lib/talons/Region/useRegion.js
+++ b/packages/peregrine/lib/talons/Region/useRegion.js
@@ -22,7 +22,7 @@ export const useRegion = props => {
         countryCodeField = 'country',
         fieldInput = 'region',
         fieldSelect = 'region',
-        optionValueKey = 'code',
+        optionValueKey = 'id',
         queries: { getRegionsQuery }
     } = props;
 
@@ -57,7 +57,7 @@ export const useRegion = props => {
         const { available_regions: availableRegions } = country;
         if (availableRegions) {
             formattedRegionsData = availableRegions.map(region => ({
-                key: region.id,
+                key: region.key,
                 label: region.name,
                 value: region[optionValueKey]
             }));
--- a/packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.js
+++ b/packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.js
@@ -221,7 +221,6 @@ const CustomerForm = props => {
                         validate={isRequired}
                         fieldInput={'region[region]'}
                         fieldSelect={'region[region_id]'}
-                        optionValueKey="id"
                     />
                 </div>
                 <div className={classes.postcode}>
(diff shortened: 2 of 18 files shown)


packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.js
```diff
@@ -219,9 +219,9 @@ const CustomerForm = props => {
                 <div className={classes.region}>
                     <Region
                         validate={isRequired}
-                        fieldInput={'region[region]'}
-                        fieldSelect={'region[region_id]'}
-                        optionValueKey="id"
+                        fieldInput={'region[code]'}
+                        fieldSelect={'region[code]'}
+                        optionValueKey="code"
```
sirugh, line 224: `code` is the default prop for `optionValueKey` but I can't remember why we switched to `id` here in the first place. Seems like we also use `id` and `region[region]`/`region[region_id]` in [addEditDialog.js](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/AddressBookPage/addEditDialog.js#L138) as well.

revanth0212: Good point but unfortunately [addEditDialog.js](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/AddressBookPage/addEditDialog.js#L138) needs to be changed as well. Everywhere in graphql the returned response has `code` as the unique key to identify the region. 

For instance here is a sample region object when the `optionValueKey="id"`
image

And here is the same object with `optionValueKey="code"`
image

We have to use `code` instead of `id` for `optionValueKey` everywhere a region is used.

Shall I push the change for `addEditDialog.js` as part of this PR?

sirugh: Let's talk about this in dev time and update back here with a comment because I don't recall enough context to make this decision.

packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -48,16 +48,17 @@ export const useRegion = props => {
     }, [country, regionInputFieldApi, regionSelectFieldApi]);
 
     const { data, error, loading } = useQuery(getRegionsQuery, {
-        variables: { countryCode: country }
+        variables: { countryCode: country },
+        skip: !country
```
revanth0212, line 52: Without this change, the `getRegions` query keeps failing when the app is loading at which point the `country` is `undefined`.

image

packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -48,16 +48,17 @@ export const useRegion = props => {
...
 
     let formattedRegionsData = [{ label: 'Loading Regions...', value: '' }];
-    if (!loading && !error) {
+    if (data && !loading && !error) {
         const { country } = data;
         const { available_regions: availableRegions } = country;
         if (availableRegions) {
             formattedRegionsData = availableRegions.map(region => ({
-                key: region.id,
+                key: region.key,
```
tjwiebell, line 61: I don't believe `key` exists on this type, should probably change back.

```suggestion
                key: region.id,
```

packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -22,7 +22,7 @@ export const useRegion = props => {
         countryCodeField = 'country',
         fieldInput = 'region',
         fieldSelect = 'region',
-        optionValueKey = 'code',
+        optionValueKey = 'id',
```
tjwiebell, line 25: With the usage of aliases above, I assumed we were aiming for a backwards compatible fix, but I believe this would be breaking. I would change back and just make sure our component usage passes `id` here. When we get to a breaking change release line we can likely remove this logic entirely.

tjwiebell (changes requested):
Let's keep this backward compatible if possible, but overall looks good. Nice work 👍

revanth0212:
Looks like the GQL team have fixed the issue. IDs are no longer repetitive. So the issue has been fixed automatically.

image

revanth0212:
> Looks like the GQL team have fixed the issue. IDs are no longer repetitive. So the issue has been fixed automatically.
> 
> image

The issue seems to be fixed. So closing this PR and taking care of the `getRegions` query issue in [another PR](https://github.com/magento/pwa-studio/pull/3093).

sirugh:
> Looks like the GQL team have fixed the issue. IDs are no longer repetitive. So the issue has been fixed automatically.

@revanth0212 in what backend version is the fix? Would be good to document, at least here in the github thread, for historical reasons at least :)

PR #3087, opened 2021-03-26
Change style-loader to something SSR-ready

jimbo:
## Description

In the course of removing (or guarding) `window` and other browser-only global references from the codebase, it became clear that a few references existed in dependency code. One such dependency is `style-loader`, a standard Webpack loader that inserts CSS into `<style>` elements. The maintainers of `style-loader` appear to be aware of the issue, [but have closed it as `won't-fix`](https://github.com/webpack-contrib/style-loader/issues/439), so we need to replace it.

The closest server-compatible loader I can find is [`isomorphic-style-loader`](https://www.npmjs.com/package/isomorphic-style-loader); while it has some minor flaws (hard dependency on `react`) and doesn't appear to be actively maintained anymore, it is MIT licensed, so we should be able to incorporate it with proper attribution. I've tried to do so in this PR.

The largest change proposed here is a switch from `mergeClasses` to a new hook (`useStyle`) with context, which allows style accumulation and insertion to work differently between browser and server.

### Verification Stakeholders
- @dpatil-magento 

### Verification Steps
1. Verify the app builds without errors (`yarn watch:venia`).
2. Verify there are no console errors.
3. Verify the site doesn't render differently from `develop`.

## Checklist


packages/venia-concept/src/index.js
```diff
@@ -11,15 +11,18 @@ import setWithPath from 'lodash.set';
...
 import app from '@magento/peregrine/lib/store/actions/app';
 import App, { AppContextProvider } from '@magento/venia-ui/lib/components/App';
+import StyleContextProvider from '@magento/peregrine/lib/context/style';
 
 import { registerSW } from './registerSW';
 
+const isServer = !globalThis.document;
 const { BrowserPersistence } = Util;
-const apiBase = new URL('/graphql', location.origin).toString();
+const origin = 'https://magento-venia-concept-e1r4b.local.pwadev:8048';
```
tjwiebell, line 24: This hard-coded value likely only works on your local, this will need some SSR treatment as well since `location.origin` will not be available. Pretty sure you won't have access to `useLocation` outside of the router context either, so some sort of communication from UPWARD would need to inform Apollo of its origin. React Router docs also seem to indicate the need to use the stateless `StaticRouter` instead of `BrowserRouter`, should that be in this scope?

packages/venia-concept/webpack.config.js
```diff
@@ -91,5 +93,28 @@ module.exports = async env => {
...
+        },
+        devtool: false,
+        optimization: {
+            minimize: false
+        },
+        plugins: [...config.plugins]
+    });
+
+    // remove HtmlWebpackPlugin
+    serverConfig.plugins.pop();
```
tjwiebell, line 111: Is this position guaranteed? Seems a little fragile, but comfortable with this if we control both sides.

tjwiebell (changes requested):
Spotted some hard-coded values that need resolved, and a few TODO's, but overall the re-factor to `useStyle` looks good. We've picked up additional `mergeClasses` calls from mainline that'll need updated, but probably best to wait until this is ready for delivery, and update once more.

packages/venia-concept/webpack.config.js
jimbo: No; it's not guaranteed, and it is fragile. I've updated the code to filter by plugin name rather than index.

packages/venia-concept/src/index.js
jimbo: True, I should have reworked this. The server will have to provide the origin when it executes the script, so we'll probably want to pull parts of `src/index.js` into a separate function that takes `origin` as an argument.

jimbo: @tjwiebell Okay. I've refactored `Adapter` to follow our talon-component pattern, slimmed down the `index.js` entrypoint, and [set `origin` to `MAGENTO_BACKEND_URL` on the server for now](https://github.com/magento/pwa-studio/pull/3087/commits/7c779ec989f34569faa19150bc32b80f53b8b7e6#diff-dfb401e2af69cac95df4a319d657ddb8771d97bf5a35fd619fc722dec6258e08R12-R15). When we dig into the real SSR work and have an express handler, we can find a way to retrieve `origin` from the HTTP request and get it to our script.

tjwiebell (approved):
This looks good to merge into the source branch, love the adapter abstraction and how clean the component looks with prop collections. There are some failures related to scaffolding, but I'm comfortable merging this and we'll get everything passing in #2991.

PR #2769, opened 2020-10-10
[feature]: Add to Cart support for Virtual products

shikhamis11:
## Description

Add to Cart support for Virtual products

## Related Issue

Closes #2768

### Verification Steps
go to virtual product product page and add to cart 

## Checklist

- [ ] I have added tests to cover my changes, if necessary.
* I have updated the documentation accordingly, if necessary.

PR #3092, opened 2021-03-29
[PWA-1016] Refactor Add To Cart to use generic AddProductsToCart mutation

tjwiebell:
## Description

In Magento 2.4.1 there is a new AddProductsToCart mutation which is more generic and type agnostic than the older, type specific mutations such as "AddConfigurableProductsToCart".

1) Investigate whether we can replace the Simple and Configurable add mutations with the generic one. If so, please do the work as part of this ticket.

2) Investigate whether it can support future types such as Bundled/Virtual/etc. If so, we don't need the "Supported Product Types" logic anymore, so please remove that too.

Links:

https://i.gyazo.com/5e19a2e43a1b46654d053197a280c1ca.png
https://github.com/magento/pwa-studio/pull/2769#issuecomment-708097397
https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/ProductFullDetail/useProductFullDetail.js#L237-L262

## Related Issue

* [[PWA-1016](https://jira.corp.magento.com/browse/PWA-1016)] Refactor Add To Cart to use generic AddProductsToCart mutation

### Verification Steps

1. Verify guest shoppers can add simple and configurable products to their cart
2. Verify logged in shoppers can add simple and configurable products to their cart

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/Header/cartTriggerFragments.gql.js
+++ b/packages/peregrine/lib/talons/Header/cartTriggerFragments.gql.js
@@ -1,0 +1,8 @@
+import { gql } from '@apollo/client';
+
+export const CartTriggerFragment = gql`
+    fragment CartTriggerFragment on Cart {
+        id
+        total_quantity
+    }
+`;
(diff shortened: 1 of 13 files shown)


sirugh (dismissed):
Great job with the implementation allowing for backwards compat. I think we could even reuse this pattern where we've got un-migrated gql files.

Hopefully we actually get around to removing the deprecated code.

Approving though it seems there are tests to update and a node failure.

sirugh (approved):

dpatil-magento:
QA Approved.

PR #3093, opened 2021-03-29
Skipping getRegion call when countryCode is falsy.

revanth0212:
## Description

The shipping information form in the checkout page has a bug where it does not pick up the region code provided in the cart page if the user does not have a default address assigned to their account. They fixed the region IDs to be unique and the problem was solved automatically in the un-authenticated workflow. The problem still exists in the authenticated workflow, and this PR attempts to fix it. Also, the UI has an issue where the `getRegions` query keeps getting called even if `countryCode` is `falsy`. Fixed it in this PR.

## Related Issue
Closes PWA-671

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Create an account and get to the cart page after adding an item to the cart.
2. Estimate the shipping costs by giving the country, state, and zip code information.
3. Get to the checkout page, see if the region code is being carried over.
4. Perform the above steps on a country that does not support regions as well. For instance, United Kingdom.

United States

image

United Kingdom

image

## Checklist


packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -48,11 +48,12 @@ export const useRegion = props => {
...
 
     const { data, error, loading } = useQuery(getRegionsQuery, {
-        variables: { countryCode: country }
+        variables: { countryCode: country },
+        skip: !country
     });
 
     let formattedRegionsData = [{ label: 'Loading Regions...', value: '' }];
-    if (!loading && !error) {
+    if (data && !loading && !error) {
```
tjwiebell, line 56: Some older code of mine before realizing we only care about the truthy-ness of `data`; this can probably be simplified:

```suggestion
    if (data) {
```

tjwiebell (changes requested):
Nice and simple, and much less scope, nicely done. Let's simplify some logic while we're in here, cover in tests, and this is ready to go.

tjwiebell (approved):

dpatil-magento:
@revanth0212 I can still reproduce the issue, can you please check?

<a href="https://gyazo.com/60deeb60e34063f7cd345073968e82f7">Image from Gyazo</a>

revanth0212:
Worked for me, dude. Would chrome version matter? I am on Version 89.0.4389.90 .

https://user-images.githubusercontent.com/35203638/113316714-44355500-92d4-11eb-87de-a2220f5793d2.mov

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.js
```diff
@@ -227,8 +227,7 @@ const CustomerForm = props => {
                     <Region
                         validate={isRequired}
                         fieldInput={'region[region]'}
-                        fieldSelect={'region[region_id]'}
-                        optionValueKey="id"
+                        fieldSelect={'region[code]'}
```
revanth0212, line 230: @tjwiebell I know this isn't the solution you suggested and your other thoughts but I believe those are becoming complex given it involves changing a fragment that is used by multiple components. So I resorted to this solution which is the simple one for time being till graphql team makes the change and at that point, we will have to change the default value of `optionValueKey` in the `region talon` and everything will start working fine.

Hope this explanation makes a valid point to stick with this solution.

tjwiebell: The CustomerAddress field is `region[region_code]`, so this fix ends up breaking a lot of other functionality (can't update or create new addresses in the Checkout AddressBook). I think even switching that here would bring back the bug this intends to solve, since the data from the Cart wouldn't line up. This might just require that breaking change to the fragment.

Screen Shot 2021-04-08 at 9 15 21 AM

tjwiebell (approved):

dpatil-magento:
QA Approved.

Issue #3094, opened 2021-03-31
[feature]: Email field validation


Karlasa:
**Is your feature request related to a problem? Please describe.**
Currently Email field validation is done on backend side.

**Describe the solution you'd like**
Improve form validators with email specific validation in frontend

**Describe alternatives you've considered**

**Additional context**

**Please let us know what packages this feature is in regards to:**
- [ ] `venia-concept`
- [x] `venia-ui`
- [ ] `pwa-buildpack`
- [ ] `peregrine`
- [ ] `pwa-devdocs`
- [ ] `upward-js`
- [ ] `upward-spec`
- [ ] `create-pwa`

Karlasa:
@magento I am working on this

sirugh:
@magento export issue to JIRA project PWA as Story

sirugh:
@Karlasa thanks for opening this issue, but we've actually gone in circles around this and have decided not to implement email validation in the front end. You can see [this PR](https://github.com/magento/pwa-studio/pull/2157) where we removed it for context:

> Reviewed this task with team lead, and it turns out this feature should have never been implemented in the first place. There is little value in doing client-side validation, and the most we should do in our app is validate the email address isn't in use, which isn't currently possible with the form library we're using. He requested we strip this out, which we're doing in the scope of this task now. Article for reference: https://medium.com/hackernoon/the-100-correct-way-to-validate-email-addresses-7c4818f24643

As such, I am closing this issue and related PR. At least now we have an issue that is easy to find :)

PR #3095, opened 2021-03-31
add email field validation

Karlasa:
## Description
Adds frontend side email field validation

## Related Issue
#3094 Email field validation

### Verification Steps
email field on Account creation, account login, forgot password and checkout form has email specific validation and validates it in frontend side.

## Checklist

### DIFF
--- a/packages/venia-ui/i18n/en_US.json
+++ b/packages/venia-ui/i18n/en_US.json
@@ -375,6 +375,7 @@
     "validation.isNotEqualToField": "{value} must be different",
     "validation.isRequired": "Is required.",
     "validation.mustBeChecked": "Must be checked.",
+    "validation.validateEmail": "Please enter a valid email address",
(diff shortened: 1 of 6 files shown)


fooman:
I suggest removing the regex in favour of a simple check for the presence of the @ symbol. A well laid out argument coming from here https://medium.com/hackernoon/the-100-correct-way-to-validate-email-addresses-7c4818f24643.

The regex seems to match what is currently in the back-end but the below does show that the back-end regex has needed more work over time where valid email addresses are still/were blocked.
https://github.com/magento/magento2/issues/12075
https://github.com/magento/magento2/pull/11576/files

In summary I would be in favour of keeping the frontend validation intent only and then if at a future time the back-end switches to email validation the frontend doesn't require any changes (and also wouldn't need to be kept in sync).

sirugh:
See issue comment https://github.com/magento/pwa-studio/issues/3094#issuecomment-815879415

PR #3105, opened 2021-04-06
[PWA-1268] Add products to Wishlist from Category Page

tjwiebell:
## Description

## Related Issue

* [[PWA-1268](https://jira.corp.magento.com/browse/PWA-1268)] Add products to Wishlist from Category Page

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/Gallery/useGallery.js
+++ b/packages/peregrine/lib/talons/Gallery/useGallery.js
@@ -1,0 +1,21 @@
+import { useQuery } from '@apollo/client';
+
+import mergeOperations from '../../util/shallowMerge';
+import defaultOperations from './gallery.gql';
+
+export const useGallery = (props = {}) => {
+    const operations = mergeOperations(defaultOperations, props.operations);
+
+    const { data: wishlistConfigData } = useQuery(
+        operations.getWishlistConfigQuery,
+        { fetchPolicy: 'cache-and-network' }
+    );
+
+    const storeConfig = wishlistConfigData
+        ? wishlistConfigData.storeConfig
+        : null;
+
+    return {
+        storeConfig
+    };
+};
--- a/packages/venia-ui/lib/components/Icon/icon.css
+++ b/packages/venia-ui/lib/components/Icon/icon.css
@@ -1,5 +1,4 @@
 .root {
-    --stroke: rgb(var(--venia-global-color-text));
     align-items: center;
     display: inline-flex;
     justify-content: center;
@@ -8,7 +7,7 @@
 
 .icon {
     fill: var(--fill, none);
-    stroke: var(--stroke);
+    stroke: var(--stroke, rgb(var(--venia-global-color-text)));
 }
 
 @media (min-width: 641px) {
--- a/packages/venia-ui/lib/components/ToastContainer/toast.css
+++ b/packages/venia-ui/lib/components/ToastContainer/toast.css
@@ -1,6 +1,7 @@
 @value info: rgb(0, 104, 108);
 @value warning: rgb(var(--venia-global-color-orange));
 @value error: rgb(220, 20, 60);
+@value success: rgb(var(--venia-global-color-green-500));
 
 .root {
     align-items: start;
@@ -48,7 +49,7 @@
 }
 
 .infoToast > .icon {
-    color: info;
+    --stroke: info;
 }
 
 .warningToast {
@@ -57,7 +58,7 @@
 }
 
 .warningToast > .icon {
-    color: warning;
+    --stroke: warning;
 }
 
 .errorToast {
@@ -66,7 +67,16 @@
 }
 
 .errorToast > .icon {
-    color: error;
+    --stroke: error;
+}
+
+.successToast {
+    composes: root;
+    border-bottom: 4px solid success;
+}
+
+.successToast > .icon {
+    --stroke: success;
 }
 
 .message {
--- a/packages/venia-ui/lib/components/Wishlist/GalleryButton/galleryButton.js
+++ b/packages/venia-ui/lib/components/Wishlist/GalleryButton/galleryButton.js
@@ -1,0 +1,75 @@
+import React, { useEffect } from 'react';
+import { Check, Heart, Info } from 'react-feather';
+import { useIntl } from 'react-intl';
+import { useToasts } from '@magento/peregrine';
+import mergeClasses from '@magento/peregrine/lib/util/shallowMerge';
+import { useGalleryButton } from '@magento/peregrine/lib/talons/Wishlist/GalleryButton/useGalleryButton';
+
+import Icon from '../../Icon';
+import defaultClasses from './galleryButton.css';
+
+const CheckIcon = <Icon size={20} src={Check} />;
+const HeartIcon = <Icon size={20} src={Heart} />;
+const InfoIcon = <Icon size={20} src={Info} />;
+
+const GalleryButton = props => {
+    const talonProps = useGalleryButton(props);
+    const {
+        handleClick,
+        isLoading,
+        isSelected,
+        showLoginToast,
+        showSuccessToast
+    } = talonProps;
+
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    const { formatMessage } = useIntl();
+    const [, { addToast }] = useToasts();
+
+    useEffect(() => {
+        if (showLoginToast) {
+            addToast({
+                type: 'info',
+                icon: InfoIcon,
+                message: formatMessage({
+                    id: 'wishlist.galleryButton.loginMessage',
+                    defaultMessage:
+                        'Please sign-in to your Account to save items for later.'
+                }),
+                timeout: 5000
+            });
+        }
+    }, [addToast, formatMessage, showLoginToast]);
+
+    useEffect(() => {
+        if (showSuccessToast) {
+            addToast({
+                type: 'success',
+                icon: CheckIcon,
+                message: formatMessage({
+                    id: 'wishlist.galleryButton.successMessage',
+                    defaultMessage:
+                        'Item successfully added to the "My Favorites" list.'
+                }),
+                timeout: 5000
+            });
+        }
+    }, [addToast, formatMessage, showLoginToast, showSuccessToast]);
+
+    const buttonClass =
+        isLoading || isSelected ? classes.root_selected : classes.root;
+
+    return (
+        <button
+            className={buttonClass}
+            disabled={isLoading}
+            onClick={handleClick}
+            type="button"
+        >
+            {HeartIcon}
+        </button>
+    );
+};
+
+export default GalleryButton;
--- a/packages/peregrine/lib/Apollo/policies/index.js
+++ b/packages/peregrine/lib/Apollo/policies/index.js
@@ -5,6 +5,7 @@
  * https://www.apollographql.com/docs/react/caching/cache-configuration/#typepolicy-fields
  * https://www.apollographql.com/docs/react/caching/cache-field-behavior/
  */
+
 const typePolicies = {
     // Query/Mutation are "types" just like "Cart".
     Query: {
@@ -18,6 +19,9 @@ const typePolicies = {
             },
             customerCart: {
                 keyArgs: () => 'Cart'
+            },
+            wishlistProducts: {
+                read: existing => existing || []
             }
         }
     },
--- a/packages/peregrine/lib/talons/Gallery/gallery.gql.js
+++ b/packages/peregrine/lib/talons/Gallery/gallery.gql.js
@@ -1,0 +1,17 @@
+import { gql } from '@apollo/client';
+
+export const GET_WISHLIST_CONFIG = gql`
+    query GetWishlistConfigForGallery($isEE:Boolean = ${process.env
+        .MAGENTO_BACKEND_EDITION === 'EE'}) {
+        storeConfig {
+            id
+            magento_wishlist_general_is_enabled
+            enable_multiple_wishlists @include(if: $isEE)
+            maximum_number_of_wishlists @include(if: $isEE)
+        }
+    }
+`;
+
+export default {
+    getWishlistConfigQuery: GET_WISHLIST_CONFIG
+};
--- a/packages/peregrine/lib/talons/RootComponents/Category/categoryFragments.gql.js
+++ b/packages/peregrine/lib/talons/RootComponents/Category/categoryFragments.gql.js
@@ -25,6 +25,7 @@ export const ProductsFragment = gql`
                     }
                 }
             }
+            sku
             small_image {
                 url
             }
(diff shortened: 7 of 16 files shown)


packages/venia-ui/lib/components/Icon/icon.css
```diff
@@ -8,7 +7,7 @@
 
 .icon {
     fill: var(--fill, none);
-    stroke: var(--stroke);
+    stroke: var(--stroke, rgb(var(--venia-global-color-text)));
```
tjwiebell, line 10: Moving this here and falling back to this default value allows the parent to provide a stroke color w/o having to override the `classes` prop of Icon.

packages/venia-ui/lib/components/ToastContainer/toast.css
```diff
@@ -48,7 +49,7 @@
 }
 
 .infoToast > .icon {
-    color: info;
+    --stroke: info;
```
tjwiebell, line 52: I think this broke when the Icon component was re-factored, or I broke it. Either way, this should work as expected now.

packages/peregrine/lib/talons/Wishlist/GalleryButton/useGalleryButton.ee.js
```diff
@@ -0,0 +1,95 @@
...
+
+export const useGalleryButton = props => {
+    const { item, storeConfig } = props;
+
+    const [isModalOpen, setIsModalOpen] = useState(false);
+    const [successToastName, setSuccessToastName] = useState();
+    const [{ isSignedIn }] = useUserContext();
+    const { formatMessage } = useIntl();
+
+    const ceTalonProps = useGalleryButtonCE(props);
```
sirugh, line 17: 🤔

tjwiebell: Ha, I warned you 😂  wanted to chat with the team today about this. I know this violates our rule, but would love to discuss alternatives to this that don't duplicate a bunch of code.

packages/peregrine/lib/talons/Gallery/useGallery.js
```diff
@@ -0,0 +1,82 @@
...
+            for (const wishlist of wishlists) {
+                const pageInfo = wishlist.items_v2.page_info;
+
+                if (pageInfo.total_pages > pageInfo.current_page) {
+                    setCurrentPage(current => ++current);
+                    break;
+                }
+            }
+        }
+    }, [wishlistItemsData]);
```
sirugh, line 73: Can you just do this within the `onCompleted` callback?

packages/venia-ui/lib/components/Wishlist/GalleryButton/galleryButton.js
```diff
@@ -0,0 +1,83 @@
...
+
+    useEffect(() => {
+        if (getSuccessToastProps) {
+            addToast(getSuccessToastProps({ icon: CheckIcon }));
+        }
+    }, [addToast, formatMessage, getSuccessToastProps]);
+
+    useEffect(() => {
+        if (getErrorToastProps) {
+            addToast(getErrorToastProps({ icon: ErrorIcon }));
```
sirugh, line 57: Feels like the first time I'm seeing this pattern with toast/props. Specific reason for not just passing them here in the component?

packages/peregrine/lib/talons/Wishlist/GalleryButton/useGalleryButton.ce.js
```diff
@@ -0,0 +1,105 @@
...
+        isSignedIn,
+        item.sku,
+        operations.getProductsInWishlistsQuery
+    ]);
+
+    const getSuccessToastProps = useMemo(() => {
+        if (addProductData) {
+            return (additionalProps = {}) => ({
+                type: 'success',
+                message: formatMessage({
```
sirugh, line 69: Along the lines of toast props in a talon, this is also the first time we'd call `formatMessage` within a talon.

sirugh (commented):
Left you some initial feedback before UX review, and I also wanted to have my concerns noted here that I would have preferred to remedy the UX discrepancy between the PDP and PLP page functionality around this button. Originally intended for the work I did on the PDP to be easily used here, but it seems like there is enough difference in UX to warrant unique component/talons.

I know we're going to discuss this stuff in dev time later, but I wanted to get in early ;) 

Good work so far!

packages/venia-ui/lib/components/Wishlist/GalleryButton/galleryButton.js
tjwiebell: This is a pattern widely used in Spectrum and nicely documented by Kent Dodds (who also claims react core uses it a ton). This example is a bit contrived and likely isn't super useful here, but `getModalProps` might show off usage a little better. Can help limit talon response size and provide inversion of control to the component to pass and override whatever props it wants. There's some additional extensibility I haven't added here since I was waiting to discuss with the team, but we can also make event handlers behave as a composite and allow our handler and handlers passed by the component to fire together.

packages/peregrine/lib/talons/Gallery/useGallery.js
tjwiebell: I'll have to look into this, was concerned about doing a local state update inside of a callback, but I'll give it a try.

packages/venia-ui/lib/components/Wishlist/GalleryButton/galleryButton.js
sirugh: Found his blog: https://kentcdodds.com/blog/how-to-give-rendering-control-to-users-with-prop-getters

packages/venia-ui/lib/components/Wishlist/GalleryButton/galleryButton.js
```diff
@@ -0,0 +1,67 @@
...
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    const { formatMessage } = useIntl();
+    const [, { addToast }] = useToasts();
+
+    useEffect(() => {
+        if (loginToastProps) {
+            addToast({ ...loginToastProps, icon: InfoIcon });
+        }
+    }, [addToast, formatMessage, loginToastProps]);
```
revanth0212, line 37:
```suggestion
    }, [addToast, loginToastProps]);
```

Do we need `formatMessage` in the dependency array?

packages/venia-ui/lib/components/Wishlist/GalleryButton/galleryButton.js
```diff
@@ -0,0 +1,67 @@
...
+
+    useEffect(() => {
+        if (successToastProps) {
+            addToast({ ...successToastProps, icon: CheckIcon });
+        }
+    }, [addToast, formatMessage, successToastProps]);
```
revanth0212, line 43:
```suggestion
    }, [addToast, successToastProps]);
```
Same here, do we need `formatMessage`?

packages/peregrine/lib/talons/Gallery/useGallery.js
```diff
@@ -0,0 +1,76 @@
+import { useState } from 'react';
+import { useQuery } from '@apollo/client';
+
+import { useUserContext } from '../../context/user';
+import mergeOperations from '../../util/shallowMerge';
+import defaultOperations from './gallery.gql';
```
revanth0212, line 6: Not a huge deal and not sure if we wanted it as a standard but to be consistent with other talons can we use `DEFAULT_OPERATIONS` instead.

packages/peregrine/lib/talons/Gallery/gallery.gql.ce.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+`;
+
+export const GET_WISHLIST_ITEMS = gql`
+    query GetWishlistItemsForLocalField($currentPage: Int!) {
+        customer {
+            id
+            wishlists {
+                id
+                items_v2(currentPage: $currentPage, pageSize: 10) {
```
revanth0212, line 20: Do you think it would be cleaner to use `items` instead here? In an effort to make code look cleaner while using this in a talon?

packages/peregrine/lib/talons/Wishlist/GalleryButton/helpers/useSingleWishlist.js
```diff
@@ -0,0 +1,130 @@
+import { useCallback, useMemo, useState } from 'react';
```
revanth0212, line 1: Just a note: This pattern gels well with the refactor described in #3141 with some refactoring.

Basically in that refactor, there is no single and multiple difference. The difference is only CE and EE and the logic single and multiple in case EE is contained in that talon itself.

packages/venia-ui/package.json
```diff
@@ -60,6 +60,7 @@
     "@babel/preset-env": "~7.3.4",
     "@pmmmwh/react-refresh-webpack-plugin": "0.4.1",
     "@storybook/react": "~6.1.14",
+    "@testing-library/react-hooks": "~5.0.3",
```
revanth0212, line 63: Wouldn't it make sense to have this in main repo instead of `venia-ui`? `peregrine` might use it one day?

packages/peregrine/lib/talons/Gallery/useGallery.js
```diff
@@ -0,0 +1,76 @@
...
+
+    const { data: wishlistConfigData } = useQuery(
+        operations.getWishlistConfigQuery,
+        { fetchPolicy: 'cache-and-network' }
+    );
+
+    useQuery(operations.getWishlistItemsQuery, {
+        fetchPolicy: 'cache-and-network',
+        onCompleted: data => {
+            const itemsToAdd = new Set();
```
revanth0212, line 28: Might be a dumb question, are you fetching page by page and not all of them at the same time, in the interest of avoiding lag if there are too many items in the list?

revanth0212 (changes requested):
Nice work @tjwiebell. Excellent work with tests as well.

I just have a couple of questions and some minor suggestions which you can omit if you deem they are not needed.

packages/peregrine/lib/Apollo/clearCustomerDataFromCache.js
```diff
@@ -10,5 +10,5 @@ import { deleteCacheEntry } from './deleteCacheEntry';
  * @param {ApolloClient} client
  */
 export const clearCustomerDataFromCache = async client => {
-    await deleteCacheEntry(client, key => key.match(/^\$?Customer/));
+    await deleteCacheEntry(client, key => key.match(/^\$?Customer/i));
```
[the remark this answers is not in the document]
tjwiebell: Local-only fields wind up in the cache under `ROOT_QUERY` with the name of the field, which is `customerWishlistProducts`. Making this insensitive, paired with the additional loop through `ROOT_QUERY` on logout ensures this data is not persisted.

packages/peregrine/lib/talons/Gallery/useGallery.js
tjwiebell: Agreed on consistency, but I shifted here to being consistent with the other usage of `shallowMerge` as with `defaultClasses`. Let's align and I'll fix this up 👍

packages/peregrine/lib/talons/Gallery/gallery.gql.ce.js
tjwiebell: Since `items` is an actual field of `Wishlist`, I would avoid that alias, as not to cause confusion. When you see property references that don't follow our styleguide, it's also a good indicator you're working with GraphQL data. Developers can still alias on destructure, which we commonly do, eg. `const {items_v2: items} = data.wishlists`.

packages/venia-ui/package.json
tjwiebell: It is already in peregrine 😄  I'm still waiting on my flow chart on how to determine where deps go, but happy to take your lead on this. I should add to the root package and remove from these two?

packages/peregrine/lib/talons/Gallery/useGallery.js
tjwiebell: Discussed this in peer review; largely this is personal preference, but for consistency we prefer to align on `defaultOperations`, sticking with the `defaultClasses` convention established in venia-ui. All imports are read-only so the constant syntax was a bit redundant, and probably introduced by me during the operation colocation re-factor. This syntax can be useful to indicate named export should be treated as immutable, but is less indicative on a default export.

packages/peregrine/lib/talons/Gallery/gallery.gql.ee.js
```diff
@@ -0,0 +1,20 @@
+import { gql } from '@apollo/client';
+
+import { GET_PRODUCTS_IN_WISHLISTS } from '../Wishlist/GalleryButton/galleryButton.gql';
+import { GET_WISHLIST_ITEMS } from './gallery.gql.ce';
```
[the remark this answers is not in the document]
tjwiebell: Didn't get a clear answer on this one, it's a new paradigm we just need to decide on. Current approach doesn't cause any bloat, but if we want to establish a rule that an `ee.js` file SHOULD NOT import from a `ce.js`, we can do that. I'm comfortable with this approach until we nail down how EE extensibility will shift from file naming to Targetables, when this become a non-issue.

packages/venia-ui/package.json
revanth0212: Now I am confused 😓 . If we add it to the main repo and don't add in `venia-ui` and `peregrine`, for someone using these packages individually, they won't have the package in their `node_modules` which will cause problems while running tests.

So I guess, we have to add it in `venia-ui` and `peregrine`'s dev dependencies list along with the main repo so for us it will be a single installation and the individual packages will pickup from higher `node_modules` and for someone using these packages individually, it will from the `node_modules` of the package itself.

Does that make sense? 😅

revanth0212 (changes requested):
All most there. The code looks fine but I noticed one discrepancy. When I ran the code against an EE background with multiple backends enabled the category page is not showing the dialog but the PDP page does. Is that expected?

https://user-images.githubusercontent.com/35203638/116315966-e848d980-a776-11eb-882c-b175c8a59fc6.mov

tjwiebell:
> All most there. The code looks fine but I noticed one discrepancy. When I ran the code against an EE background with multiple backends enabled the category page is not showing the dialog but the PDP page does. Is that expected?

Good observation, but I believe my code is behaving as intended and the PDP is still awaiting a re-factor to behave more like the Category page. My EE implementation will behave like CE when `enable_multiple_wishlists === '0'`, whereas the PDP implementation always pops the multiple wishlist Dialog, regardless if multiple wishlists is enabled.

revanth0212 (dismissed):

dpatil-magento:
@tjwiebell Please check following - 
1. Lighthouse accessibility score on category page is decreased.
2. Selected Heart icon does not go back to default (selectable) state  when user removes product from wishlist and goes back to category page.

dpatil-magento:
QA Approved.

revanth0212 (approved):

PR #3106, opened 2021-04-06
Moving npm-is into pwa-studio.

revanth0212:
## Description

**npm-is** a utility package that validates if a given script is executed using the given package manager. For instance, `pwa-studio` can only be installed using `yarn` and if you try to install using `npm` it has to fail. @zetlen created this wonderful utility package a while ago which we have been using in the **preinstall** script to check validate if yarn is being used to install the package. Unfortunately in npm 7 update a critical piece of code that the package relies on has changed. Specifically the value of `process.env.npm_execpath` when it is invoked using `npx`. The script extensively relies on the value of `process.env.npm_execpath` to make the decision about what package manager is used to initiating the script.

Previously (npm 6 or less) if you initiated a script using `yarn` which internally uses `npx` the value of `process.env.npm_execpath` was `PATH_TO_THE_YARN_EXECUTABLE` but after the update, the value of `process.env.npm_execpath` has changed to `PATH_TO_THE_NPX_EXECUTABLE` instead. Due to this the validation has failed and `pwa-studio` installation fails with the following error:

image

This PR addresses this issue by removing all references of `npm-is` and adding a note in the documentation to avoid using `npm`.

## Related Issue
Closes PWA-1660

## Acceptance
Should be able to validate if a script is initiated using a given package manager.

### Verification Stakeholders
@zetlen 
@dpatil-magento 

### Verification Steps
1. Checkout the branch.
2. Run `yarn install` and it should work fine.
3. Run `npm install` and it should fail.

## Checklist


packages/npm-is/README.md
```diff
@@ -0,0 +1,102 @@
...
+Error: This project can only be used with the "yarn" package manager, but it was
+invoked by "npm", which is not supported.
+```
+
+## Install
+
+**Don't install this as a dependency if you're going to use it in the**
+**[recommended way described above](#Recommended_usage).**
+
+**Instead, run it with `npx` so that it works before dependencies are installed!**
```
revanth0212, line 51: This does not make sense anymore because that is the problem at hand 😅 . Leaving it for now till I check with Zetlen.

revanth0212: Instead how about changing the usage of `npx` to `yarn add -D @magento/npm-is && yarn @magento/npm-is run validate yarn` that way we are only installing the required package and checking for the package manager before installing the whole package.json file?

revanth0212: I even tried `npm exec` and it fails with a similar error because the value of `process.env.npm_execpath` is the path to the NPM CLI.

sirugh:
@revanth0212 what's up with this PR?

package.json
```diff
@@ -30,7 +30,7 @@
     "danger": "danger-ci",
     "lint": "eslint '@(packages|venia-integration-tests|scripts)/**/{*.js,package.json}' --ignore-pattern node_modules --ignore-pattern storybook-dist",
     "postbuild": "rimraf \"./packages/*/dist/{,**/}__*__\"",
-    "preinstall": "node -e 'process.env.CI||process.exit(1)' || npx npm-is yarn",
+    "preinstall": "echo \"Please make sure to use yarn while installing the packages. NPM is not supported.\"",
```
sirugh, line 33: This will echo every time you run install - I don't think we want that.

sirugh (approved):

pwa-devdocs/src/venia-pwa-concept/setup/index.md
```diff
@@ -46,6 +46,8 @@ In the PWA Studio project's root directory, run the following command to install
 yarn install
 ```
 
+**Note:** Please be aware that the project uses `yarn workspaces` and does not support `npm install`. Please use `yarn install` instead.
```
supernova-at, line 49: @jcalcaben This is the correct place to put this, right?

@revanth0212 We may want to add a note to this repo's `README` as well.

jcalcaben: Good callout on the README

PR #3108, opened 2021-04-07
🪄 Product Item style fixes

fnhipster:
#### Gallery Item Inset Shadow

Products in Venia (and most products) have a white background. Here I added a subtle inset shadow to separate product items from each other.

image
image

#### Product Items per Row (Responsive)

image
image
image
image

### Verification Steps
- Go to any Category Page
- View any PageBuilder content that is rendering the Products (Carousel) Content-Type

### Resolved issues:
1. [x] resolves magento/pwa-studio#3109: 🪄 Product Item style fixes

### DIFF
--- a/packages/pagebuilder/lib/ContentTypes/Products/products.css
+++ b/packages/pagebuilder/lib/ContentTypes/Products/products.css
@@ -56,3 +56,7 @@
     composes: root;
     composes: root from '../Slider/slider.css';
 }
+
+.carousel :global .slick-track .slick-slide:not(:first-child) > div {
+    margin-left: 4px;
+}
--- a/packages/pagebuilder/lib/ContentTypes/Products/products.js
+++ b/packages/pagebuilder/lib/ContentTypes/Products/products.js
@@ -71,8 +71,8 @@ const Products = props => {
         paddingBottom,
         paddingLeft,
         cssClasses = [],
-        slidesToShow = 5,
-        slideToShowSmall = 2,
+        slidesToShow = 4,
+        slideToShowSmall = 1,
         slideToShowSmallCenterMode = 1
     } = props;
 
--- a/packages/pagebuilder/lib/ContentTypes/Slider/slider.css
+++ b/packages/pagebuilder/lib/ContentTypes/Slider/slider.css
@@ -59,7 +59,6 @@
 
 .root :global .slick-track .slick-slide > div {
     min-height: inherit;
-    margin: 3px;
 }
 
 .root :global .slick-track:before,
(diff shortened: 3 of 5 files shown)


sirugh:
@magento create issue from PR

brendanfalkowski:
@fnhipster This seems like a bad default for fashion store (or any) where product color matters to the buyer.

It's hard to tell without a before/after but from the code change it looks like the shadow would cast a nearly flat grade of 7% black box over the entire image, which I don't think anyone would want. It's hard enough getting colors true-to-life on screen, and this just cooks them to mimic what's normally chosen done on-set or in post.

fnhipster:
@brendanfalkowski but I do see your point.

brendanfalkowski:
@fnhipster Wouldn't there be zero visual change then? If the image is opaque any inset shadow on its container would be behind the image. I assumed this was being added above the image to darken its white background to very light gray. Otherwise you'd never see it.

In my screenshot, I just copied the same box shadow into an unrelated project to show that's it's a nearly flat gradient with a similarly sized container because the shadow's spread is so wide (10rem). It didn't look like the intent was to add vignetting but more of a flat overlay, which is what I think most people would object to.

Please correct if I'm understanding it wrong. I don't have a clone of this repo running right now, so I couldn't pull to see a before/after in context.

sirugh:
We always deploy demo instances with each PR: https://pr-3108.pwa-venia.com/ which you can compare against develop.pwa-venia.com if you like :)

fooman:
Did a comparison here
image

and the color does change
image

image

I agree with @brendanfalkowski if this is really something super important to do for this particular merchant (ie Venia), maybe best to implement it as an extension so make this optional.

brendanfalkowski:
@sirugh Thanks, didn't know about that.

This is what I was getting at. Same 7% alpha but full green channel to show what is being tinted clearer. When each title is large, the 10rem spread doesn't interfere too much:

Screen Shot 2021-04-07 at 2 53 40 PM

But as the tile size changes, the spread doesn't change length so you get increasingly flatter overlay as the container shrinks:

Screen Shot 2021-04-07 at 2 54 01 PM

If you boost the alpha and keep the 10rem spread, you can see clearer how the overlay would affect the image colors.

Screen Shot 2021-04-07 at 2 54 14 PM

Obviously, a merchant isn't going to cast a ghoul green haze over their models. But it's just to illustrate that this technique does perceptibly affect the appearance of colors (even at 7% alpha) due to box-shadow spread using a fixed length for all tile sizes.

Ultimately these "defaults" go unquestioned in most Magento builds, and it just doesn't make sense to me to introduce a default that changes the appearance of photography as it was shot/approved before reaching the site.

fnhipster:
@fooman / @brendanfalkowski, you are right about the vignette. This should be a style update to the Venia Concept Store instead of Venia UI. I appreciate the time an effort communicating your point-of-views. I'll close this PR, and create a new one with just the responsive products grid updates.

PR #3113, opened 2021-04-09
[bug] Avoid extra category query

### LINKED ISSUE #3112: Queries fire multiple times
**Describe the bug**

Whilst I was updating my project to PWA Studio 8.0 I noticed that some queries run twice. Couldn't directly see why this was happening so decided to check https://venia.magento.com/ where the same issue occurred.

**To reproduce**

1. Go to https://venia.magento.com/
2. Open the chrome network tab and set the filter to XHR + graphql
3. Click in the menu on 'Accessories'
4. You should see the following results:
Screenshot 2021-04-09 at 15 29 03

As you can tell the `GetCategories` query is running twice. I've checked to see if there is a difference and but request (including params) are exactly the same.

I'm not sure if this is actually a bug or intended behaviour, hope someone can clarify?

**Expected behavior**
-

**Screenshots**
-

**Possible solutions**
-

**Debug Report**
-

-

-   [ ] `venia-concept`
-   [x] `venia-ui`
-   [ ] `pwa-buildpack`
-   [ ] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

sirugh:
## Description

The introspection query in the category page is used to construct a map of filters and types so as to properly construct the final `filters` variable when making the category query.

Using `cache-and-network` causes the introspection query to resolve with (likely identical) data. This causes the downstream category query to run twice because the identity of the filters object is not the same.

This PR uses a ref to do a deep equality check on the filters during each render so when the introspection query resolves from the `network` part of `cache-and-network` it does not cause a the category query to refetch unless the filters change.

## Related Issue

Closes #3112 and PWA-1666

### Verification Steps

1. Go to the a category page like `/default/venia-bottoms.html?page=1`
2. View network tab. Ensure that the `GetCategories` query is only invoked once.
3. Try this with a clear cache or with a prepopulated cache. Should never query more than once.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
+++ b/packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
@@ -1,6 +1,7 @@
 import { useEffect, useMemo, useRef } from 'react';
 import { useLocation } from 'react-router-dom';
 import { useLazyQuery, useQuery } from '@apollo/client';
+import { isEqual } from 'lodash';
 
 import mergeOperations from '../../../util/shallowMerge';
 import { useAppContext } from '../../../context/app';
@@ -11,7 +12,7 @@ import {
     getFiltersFromSearch,
     getFilterInput
 } from '../../../talons/FilterModal/helpers';
-import { isEqual } from 'lodash';
+
 import DEFAULT_OPERATIONS from './category.gql';
 
 /**
--- a/packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
+++ b/packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
@@ -11,7 +11,7 @@ import {
     getFiltersFromSearch,
     getFilterInput
 } from '../../../talons/FilterModal/helpers';
-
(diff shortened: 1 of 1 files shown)


packages/peregrine/lib/talons/RootComponents/Category/useCategory.js
```diff
@@ -137,24 +139,43 @@ export const useCategory = props => {
...
-        currentSort,
-        filterTypeMap,
-        id,
-        pageSize,
-        runQuery,
-        search
-    ]);
+        },
+        // exhaustive-deps complains about refs...
+        /* eslint-disable react-hooks/exhaustive-deps */
```
sirugh, line 167: Is there a better way to solve this without a ref so we avoid having to disable the lint rule?

sirugh:
Closing as this was just my attempt at a solution and I'm out next week. Also the ticket needs to be groomed/triaged according to process first ;)

PR #3122, opened 2021-04-15
[BUGFIX] Replace the mediaBackendUrl when the feature onboard is used

lewisvoncken:
## Description

if IMAGE_OPTIMIZING_ORIGIN is set to onboard it should replace the mediaBackendUrl with the current window.location
When the media url is different from the admin url it won't be replaced and could result in CORS errors.

### Verification Steps

1. Setup an environment with multiple stores
2. use a different media url compared to the base url
3. when cors is enabled this will result in CORS erros

### Resolved issues:
1. [x] resolves magento/pwa-studio#3167: [BUGFIX] Replace the mediaBackendUrl when the feature onboard is used

### DIFF
--- a/packages/peregrine/lib/util/makeUrl.js
+++ b/packages/peregrine/lib/util/makeUrl.js
@@ -78,7 +78,6 @@ const makeOptimizedUrl = (path, { type, ...opts } = {}) => {
 
     const { origin } = window.location;
     const isAbsolute = absoluteUrl.test(path);
-    const magentoBackendURL = process.env.MAGENTO_BACKEND_URL;
     let baseURL = new URL(path, mediaBackend);
 
     // If URL is relative and has a supported type, prepend media base onto path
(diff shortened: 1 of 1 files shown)


fooman:
@lewisvoncken in your steps to recreate you have 

>2. use a different media url compared to the base url

which to me means you will need to configure CORS. If you didn't want this to be an issue wouldn't you only need to configure the media url to be under the same base url.

I do believe the code you changed is not yet fully multi store capable. Looking at the test case here
https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/util/__tests__/makeUrl.spec.js#L16
my thinking would be that we somehow need to make sure that we get secure_base_media_url for each store view and then construct the url accordingly for media urls.

Might be worth a broader discussion to remove one configuration dimension, for example
if base_url + /media == secure_base_media_url this is essentially onboard

sirugh:
Not sure what to do with this guys :) Could you open an issue first and we can discuss the problem and potential solutions?

sirugh:
@magento create issue from PR

revanth0212:
@lewisvoncken can you please add/update tests to cover your changes?

Also, it would be great if you can elaborate a bit more on the reason behind the PR and the solution. Thank you.

revanth0212:
Hi @lewisvoncken, we are closing this ticket at the moment due to inactivity but feel free to reopen it back if you see the need. Please update the issue description while reopening in case.

Thank you.

PR #3130, opened 2021-04-16
Add to Wishlist from Cart

revanth0212:
## Description

This PR adds a new feature that lets the user move an item from the cart to the wishlist. This PR only targets single wishlist use cases, for instance, CE backends or EE backends with multiple wishlists turned off.

## Related Issue
Closes PWA-1495

### Verification Stakeholders
@schensley 
@tkacheva 
@dpatil-magento 

### Verification Steps
1. Go to the deployed version of the app.
2. Login and add an item to the cart.
3. Go to the cart page and click on the kabab menu.
4. Click the `Save for later` button.
5. The item should be removed from the cart and should be available in the favorites list.

## Checklist


revanth0212:
UX and PM review passed.

packages/peregrine/lib/talons/CartPage/ProductListing/useWishlist.ce.js
```diff
@@ -0,0 +1,85 @@
...
+                    type: 'info',
+                    message: formatMessage({
+                        id: 'cartPage.wishlist.ce.successMessage',
+                        defaultMessage:
+                            'Item successfully added to your favorites list.'
+                    }),
+                    timeout: 5000
+                });
+
+            await removeItemFromCart({
```
revanth0212, line 56: We have discussed this in the Dev time and decided will not be rolling back if remove from the cart fails.

packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
```diff
@@ -42,6 +42,7 @@ export const ProductListingFragment = gql`
             ... on ConfigurableCartItem {
                 configurable_options {
                     id
+                    configurable_product_option_value_uid
```
revanth0212, line 45:
```suggestion
```

I guess this is a change I forgot to revert after reverting the rollback code.

packages/peregrine/lib/talons/CartPage/ProductListing/wishlistFragment.ee.js
```diff
@@ -0,0 +1,24 @@
+import { gql } from '@apollo/client';
+
+export const WishlistFragment = gql`
+    fragment WishlistFragment on Wishlist {
+        id
+        name
+        items: items_v2 {
```
sirugh, line 7: Generally I would avoid renaming or aliasing responses from gql as it is just another layer of abstraction that could make it more difficult to debug. If something went wrong with `items` in the response we would have to know that it was aliased from `items_v2`. Obviously this is a pretty minor thing, and I won't ask you to change it, but I think it's just good to be aware.

In this case I might actually leave it as `items_v2` is...rough.

packages/peregrine/lib/talons/CartPage/ProductListing/useWishlist.ce.js
```diff
@@ -0,0 +1,85 @@
...
+                    wishlistId: '0',
+                    itemOptions: {
+                        sku,
+                        quantity,
+                        selected_options
+                    }
+                }
+            });
+
+            wishlistData &&
```
sirugh, line 45: Prefer `if (wishlistData) {`?

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -1,9 +1,14 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useMutation, useQuery } from '@apollo/client';
+import { useIntl } from 'react-intl';
 import { useCartContext } from '@magento/peregrine/lib/context/cart';
-import { deriveErrorMessage } from '../../../util/deriveErrorMessage';
+import { useUserContext } from '@magento/peregrine/lib/context/user';
 import configuredVariant from '@magento/peregrine/lib/util/configuredVariant';
+
+import { useWishlist } from './useWishlist';
```
sirugh, line 8: I feel like having this hook live as a sibling to the talon within the `talons` directory could be confusing. I know we spoke about this regarding @tjwiebell's use case. Did we end up with a decision on where to place these utility hooks?

sirugh (changes requested):
Nice work :) I was a little confused why we have a `.ee` file in here when you're just implementing the single wishlist, but I suppose this is going to help be a jumping off point for the next ticket (add to multiple).

packages/venia-ui/lib/components/CartPage/ProductListing/productListingFragments.js
revanth0212: Yes, my bad. Wrong call.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -72,42 +82,89 @@ export const useProduct = props => {
...
-        if (updateItemCalled || removeItemCalled) {
-            // If a product mutation is in flight, tell the cart.
-            setIsCartUpdating(updateItemLoading || removeItemLoading);
+    const [{ cartId }] = useCartContext();
+
+    // Use local state to determine whether to display errors or not.
+    // Could be replaced by a "reset mutation" function from apollo client.
+    // https://github.com/apollographql/apollo-feature-requests/issues/170
+    const [displayError, setDisplayError] = useState(false);
+    const [showLoginToast, setShowLoginToast] = useState(0);
```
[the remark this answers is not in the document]
revanth0212: I picked this logic from Tommy's #3105

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
```diff
@@ -91,10 +101,16 @@ const Product = props => {
               })
             : '';
 
+    useEffect(() => {
+        if (loginToastProps) {
+            addToast({ ...loginToastProps, icon: InfoIcon });
```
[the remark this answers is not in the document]
revanth0212: Yeah I kind of like this idea of sending toast props from the talon and let the component only worry about when to show the toast itself.

packages/venia-ui/lib/components/CartPage/cartPage.js
```diff
@@ -40,19 +45,32 @@ const CartPage = props => {
...
 
+    const [, { addToast }] = useToasts();
+
     const classes = mergeClasses(defaultClasses, props.classes);
 
+    useEffect(() => {
+        if (wishlistSuccessProps) {
+            addToast({ ...wishlistSuccessProps, icon: CheckIcon });
+        }
+    }, [addToast, wishlistSuccessProps]);
```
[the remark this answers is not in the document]
revanth0212: The reason for this is when the item is added to the wishlist, it is immediately removed from the cart so there won't be a component to display the toast. So I had to lift this logic to the CartPage component instead which is always rendered irrespective of the cart being empty or full.

packages/venia-ui/lib/components/CartPage/ProductListing/product.js
```diff
@@ -33,11 +47,13 @@ const Product = props => {
         errorMessage,
         handleEditItem,
         handleRemoveFromCart,
-        handleToggleFavorites,
+        handleSaveForLater,
         handleUpdateItemQuantity,
         isEditable,
         isFavorite,
```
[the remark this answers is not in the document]
revanth0212: Remove `handleSaveForLater`? It is a function we are using.

packages/venia-ui/i18n/en_US.json
```diff
@@ -376,10 +378,10 @@
     "validation.isRequired": "Is required.",
     "validation.mustBeChecked": "Must be checked.",
     "validation.validatePassword": "A password must contain at least 3 of the following: lowercase, uppercase, digits, special characters.",
+    "wishlist.emptyListText": "There are currently no items in this list",
+    "wishlistButton.addedText": "Added to Favorites",
```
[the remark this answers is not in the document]
revanth0212: Alphabetically they are correct arent they?
image

packages/peregrine/lib/talons/CartPage/ProductListing/wishlistFragment.ee.js
revanth0212: Yeah dude you are right. I am not a fan of renaming usually but I felt `items_v2` is not readable in the code.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
revanth0212: Not quite completely though. I am trying to see if I can generalize wishlist logic into a talon of its own so all the components/talons that use it can just import. But unfortunately, that can be tricky because the use cases can be different and Jimmy suggested moving over it if it's getting complex to generalize.

sirugh: Yea I agree. At a minimum, throw the file into a `helpers` directory.

packages/venia-ui/i18n/en_US.json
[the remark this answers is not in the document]
sirugh: I don't think so :P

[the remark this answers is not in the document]
sirugh: Honestly we should just set up a lint rule to auto sort this file for us :D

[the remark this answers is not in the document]
jimbo: Prefer standard string-value sorting, please, as it's the least arbitrary.

```
addError
addText
addedText
```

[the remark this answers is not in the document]
revanth0212: Yeah, the plugin I used is not considering the length of the keys, it is only considering based on the character in the position. Since `addEr` comes after `added` it sorted it that way.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
revanth0212: If I am able to generalize this, yes I will move it into a helper's folder but if it is bound to stay like this, it's not a helper anymore so we can leave it as it is. I'll update soon about the generalization work and let's make the call then.

packages/peregrine/lib/talons/CartPage/ProductListing/wishlistFragment.ee.js
sirugh: After discussion in the other PR I think we should move the alias to the destructure so we don't run into a conflict with the existing `items` schema value.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -5,7 +5,7 @@ import { useCartContext } from '@magento/peregrine/lib/context/cart';
 import { useUserContext } from '@magento/peregrine/lib/context/user';
 import configuredVariant from '@magento/peregrine/lib/util/configuredVariant';
 
-import { useWishlist } from '../../Wishlist/Wishlist/useWishlist';
+import { useWishlist } from '../../Wishlist/useWishlist';
```
revanth0212, line 8: I am still not sure where I am should put these hooks. So I just flattened them into the Wishlist talon in peregrine. Any help?

revanth0212:
I still have to add tests but the functional code is up for re-review.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
sirugh: I think this is fine for now - we have a followup PWA-1709 to consolidate.

sirugh (dismissed):
Good work! I appreciate the effort you spent spiking on the alternative pattern. This is good to go to QA :)

packages/venia-ui/lib/components/CartPage/ProductListing/__tests__/product.spec.js
```diff
@@ -68,7 +68,7 @@ test('renders simple product correctly', () => {
         errorMessage: undefined,
         handleEditItem: jest.fn(),
         handleRemoveFromCart: jest.fn(),
-        handleToggleFavorites: jest.fn(),
+        handleSaveForLater: jest.fn(),
         handleUpdateItemQuantity: jest.fn(),
         isEditable: false,
         isFavorite: false,
```
sirugh, line 74: Could remove `isFavorite` right?

packages/venia-ui/lib/components/CartPage/ProductListing/__tests__/product.spec.js
```diff
@@ -119,7 +123,7 @@ test('renders configurable product with options', () => {
         errorMessage: undefined,
         handleEditItem: jest.fn(),
         handleRemoveFromCart: jest.fn(),
-        handleToggleFavorites: jest.fn(),
+        handleSaveForLater: jest.fn(),
         handleUpdateItemQuantity: jest.fn(),
         isEditable: true,
         isFavorite: false,
```
sirugh, line 129: Same here

packages/venia-ui/lib/components/CartPage/ProductListing/__tests__/product.spec.js
```diff
@@ -141,10 +145,55 @@ test('renders configurable product with options', () => {
...
+
+test('renders toast if wishlistSuccessProps is not falsy', () => {
+    useProduct.mockReturnValueOnce({
+        errorMessage: undefined,
+        handleEditItem: jest.fn(),
+        handleRemoveFromCart: jest.fn(),
+        handleSaveForLater: jest.fn(),
+        handleUpdateItemQuantity: jest.fn(),
+        isEditable: false,
+        isFavorite: false,
```
sirugh, line 166: And here :)

supernova-at:
### QA

- [ ] MFTF
- [x] Run Verification Steps
- [x] Chrome
- [x] Firefox
- [ ] Safari. I have accidentally upgraded to 14.1 and nothing works. May be related to https://bugs.webkit.org/show_bug.cgi?id=225344.
- [x] Mobile device

<hr />

I'll note that it would be cool if clicking on the toast redirected the user to the favorites page 🥇 
Screen Shot 2021-05-06 at 3 46 43 PM

revanth0212:
Hey @supernova-at, thanks for doing the QA on this. I have handled the simple product's scenario. Unfortunately, I could not understand what the error scenario was. Can you expand on that?

supernova-at:
> Unfortunately, I could not understand what the error scenario was. Can you expand on that?

Sure, it's this one (in the JIRA ticket):

> Verify error flow. One way to simulate this scenario is to have 2 tabs open for same customer > Move or remove product in tab 1 and try to move the product on second tab.

supernova-at:
@revanth0212 I updated the QA comment after merging in `develop`. There is still one MFTF test failure.

dpatil-magento:
> @revanth0212 I updated the QA comment after merging in `develop`. There is still one MFTF test failure.

@supernova-at Those are valid failures as display position in selectors has been changed. You can try running using this branch https://github.com/magento-commerce/pwa-tests/pull/15

revanth0212:
> Verify error flow. One way to simulate this scenario is to have 2 tabs open for same customer > Move or remove product in tab 1 and try to move the product on second tab.

Unfortunately, this is a problem with the GQL mutation. Since there is no single mutation to move an item from cart to wishlist, we are using 2 mutations in tandem.

1. First we add the item to the wishlist, which in this case is working fine.
image

2. Second we remove the item from the cart, which in this case is failing because the item is not part of the cart anymore.
image

I am not sure if we need to solve for this case, because firstly, its an edge case and secondly this wont happen if GQL can provide us with a unified GQL mutation for moving an item from cart to wishlist.

Also, this wont be a problem if the user refreshes the page because the cart will update and the item wont be visible anymore.

supernova-at:
MFTF tests are failing but not because of this PR. Running using a PR in that repo, they all pass.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
```diff
@@ -109,10 +110,18 @@ export const useProduct = props => {
         return null;
     }, [formatMessage, showLoginToast]);
 
+    const handleWishlistUpdateError = useCallback(() => {
+        setIsCartUpdating(true);
+
+        fetchCartDetails({ variables: { cartId } });
```
sirugh, line 116: Threading this prop all the way down from the cart page feels odd. Could you tell me why this is necessary in the first place and how to repro or get this function to be called? If we do have to go this route, perhaps we can go with a callback like `onError` that is passed down.

sirugh (commented):
Not sure how to evaluate the recent push - could you help me understand how to repro the error state?

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
revanth0212: The react state update error I showed earlier today was present even before this change. That said, you can reproduce an error that can call this function by opening the cart in 2 tabs, move an item in 1 cart to the wishlist and do the same on the other tab. Removing items from the cart will fail because the item is not in the cart anymore.

sirugh: Ok thanks for providing repro.

I do think the handling here is a bit off still - the item greys out, then the success toast pops up, then you see an error on the item for a brief moment and then the item is deleted from the cart page UI.

[Image from Gyazo](https://gyazo.com/1a4ded3a8088a9d19680006f5d0f0c4f)

revanth0212: Yeah, it is doggy for sure, but that's the best at the moment till GQL comes up with a unified query. And this is a very very edge case. UX approves this as a valid solution given the circumstances.

revanth0212: > If this is the desired approach (refetching the cart data on error) then I think a callback such as `onError` defined by the cart page and passed to any children who care, is preferable to a `fetchCartDetails` prop.

Considering `useWishlist` as a helper talon, I preferred to name it `onErrorX` instead because later when we move to a unified `useWishlist` helper solution it will be a straight swap.

sirugh:
Maybe the state error, at least the one from product listing, is coming from here: https://github.com/magento/pwa-studio/blob/revanth/add-cart-to-wishlist/packages/peregrine/lib/talons/CartPage/ProductListing/useQuantity.js#L94

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
revanth0212: > I also think we can remove the `setDisplayError(true)` call for the product causing the error since it's going to be deleted anyways.

Wouldn't that hide legitimate errors as well?

sirugh: > Wouldn't that hide legitimate errors as well?

Would it? Normally you would call `setDisplayError(true)` in the [catch block](https://github.com/magento/pwa-studio/blob/revanth/add-cart-to-wishlist/packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js#L197) related to the mutation that fails. Is this what we use this function for?

Perhaps it would hide legitimate errors but I still think the UX is weird. Look at the gif!

supernova-at:
> I still think the UX is weird. Look at the gif!

It is definitely weird but since we're getting a mutation that moves to wishlist from cart all at once, this is an interim weirdness that we made sure to get UX and Dev's approval on.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
sirugh: I think I understand my confusion about this. In these files we have called `setDisplayError` in the `catch` block of the related item call, such as [here](https://github.com/magento/pwa-studio/pull/3130/files#diff-935be6164790da836b0c844633f2a602ae4ced096ff8bfc17c4427d2a13d885bR197). In fact, I believe that is the code that is causing the error to display, not what you added here on line 118.

So please disregard my suggestion to remove the call to `setDisplayError`. We do need it, in the event that the wishlist mutation throws an error.

I still would like to see my other concern addressed :)

sirugh:
@supernova-at thanks, I didn't have that context! I also realized that Revanth's use of `setDisplayError` was not actually the cause of the weirdness in the gif - as you said it is due to the dual mutations.

When are we getting a single mutation?

tjwiebell:
@revanth0212 - you mentioned during Office Hour that you were using a single operation for this, and just wanted to clarify what I meant in case it can help with future use cases. No need to re-factor if you've worked through the issues of these mutations being done is separate transactions, but there is one example of this with [Configurable Product edits](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/productForm.js#L186:L220) (which happen to be an add + removal).

revanth0212:
> No need to re-factor if you've worked through the issues of these mutations being done is separate transactions, but there is one example of this with [Configurable Product edits](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/productForm.js#L186:L220) (which happen to be an add + removal).

This is a wonderful idea @tjwiebell but unfortunately, it did not also solve the problem and also this will be a huge change because the cart fragment resides in `venia-ui` and the talon is in `peregrine`. We can move the cart fragments over but I would prefer that to be done in a different PR.

revanth0212:
@sirugh can you do a code review on the recent changes and @supernova-at can you please do a QA re-run on this?

supernova-at:
> When are we getting a single mutation?

Likely not until 2.4.3 in August. I created PWA-1736 to track.

sirugh:
@revanth0212 did you see my previous comment? It may have gotten lost in the discussion about the other issue.

> If this is the desired approach (refetching the cart data on error) then I think a callback such as onError defined by the cart page and passed to any children who care, is preferable to a fetchCartDetails prop.

packages/peregrine/lib/talons/CartPage/ProductListing/useProduct.js
sirugh: @revanth0212 had reasoning for not making my suggested change, I wanted to post it here so it was captured:

> I would prefer it called fetchCartDetails because form the cart page POV we are not using that function for all errors. We are only using that when the wishlist move errors out.

This makes sense to me.

sirugh (approved):

supernova-at:
### QA

⏳ In Progress ⏳ 

- [x] MFTF
- [ ]  JIRA ticket test plan
- [ ] Run Verification Steps
- [ ] Chrome
- [ ] Firefox
- [ ] Safari. I have accidentally upgraded to 14.1 and the whole app fails to load. Also fails on develop and demo instance, so not specific to this PR. May be related to https://bugs.webkit.org/show_bug.cgi?id=225344.
- [ ] Mobile device

PR #3131, opened 2021-04-17
#3114 Set fetch policy and next fetch policy for cms block query

### LINKED ISSUE #3114: [bug]: useCmsPage nextFetchPolicy: 'cache-first'? (also cmsBlock)
**Describe the bug**

You load a CMS page/block once and it stays like that forever.

**To reproduce**

1. Load a CMS page
2. Change content of CMS page
3. Reload CMS page
4. Old content

**Expected behavior**

Fresh content

**Possible solutions**

Change policy to something that makes sense (e.g., cache-and-network). CMS content changes. It's true it happens infrequently, but anyway, I am not aware of a way to force clients to clear their application cache (the only way to get fresh data now).

With cmsBlock we can switch policy, but given the inability to use a custom routing component (see https://github.com/magento/pwa-studio/issues/3081#issuecomment-812389729), there's no proper way to change the useCmsPage talon right now. I have to edit directly in node_modules.

-   [ ] `venia-concept`
-   [x] `venia-ui`
-   [ ] `pwa-buildpack`
-   [x] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

vasilii-b:
## Description

This PR sets the fetch policy for CMS Block.

## Related Issue

Closes #3114.

## Acceptance

N/A

N/A

## Checklist


packages/venia-ui/lib/components/CmsBlock/cmsBlock.js
```diff
@@ -14,7 +14,9 @@ const CmsBlockGroup = props => {
     const classes = mergeClasses(defaultClasses, props.classes);
 
     const { loading, error, data } = useQuery(GET_CMS_BLOCKS, {
-        variables: { identifiers }
+        variables: { identifiers },
+        fetchPolicy: 'cache-and-network',
+        nextFetchPolicy: 'cache-first'
```
fooman, line 19: wouldn't cache-first mean that updated content would not show if it was edited again?
https://medium.com/@galen.corey/understanding-apollo-fetch-policies-705b5ad71980?

sirugh: > wouldn't cache-first mean that updated content would not show if it was edited again?

On first load, cache-and-network _always_ updates with network (server) data. But if that data ever changes on the client, the query makes _another_ call to the server. We applied the `nextFetchPolicy` to avoid excessive server calls as it allows the initial network fetch but then always trusts the cache from then on (for this page load).

supernova-at (approved):

dpatil-magento:
QA Approved.

PR #3136, opened 2021-04-21
Cypress visual testing integration

revanth0212:
## Description

Added Cypress Visual Testing plugin. With this, we can write tests that can capture snapshots of the page visually while testing.

## Related Issue
Closes PWA-1661

### Verification Stakeholders
@dpatil-magento 

### Verification Steps
1. Checkout the branch locally.
2. Navigate to the integration tests package.
3. Run `yarn install`
3. Run `yarn test` to start the cypress UI.
4. Run the sample visual test (pageBuilder/homePage.spec.js). Other tests you might see are dummy ones, ignore them for now.

## Checklist


venia-integration-tests/src/tests/pageBuilder/homePage.spec.js
```diff
@@ -0,0 +1,13 @@
+/// <reference types="cypress" />
```
supernova-at, line 1: Is this needed?

venia-integration-tests/src/tests/pageBuilder/homePage.spec.js
```diff
@@ -0,0 +1,13 @@
...
+
+context('Assert Venia Home Page pagebuilder content', () => {
+    describe('Assert Page Builder content', () => {
+        it('should render page builder content correctly', () => {
+            cy.visit('/');
+
+            cy.wait(3000);
```
supernova-at, line 8: I feel like this is too short. A no-cache, hard refresh page load on https://develop.pwa-venia.com/ took 6000ms for me.

Maybe 10000 here?

davemacaulay: https://docs.cypress.io/faq/questions/using-cypress-faq#How-can-I-make-Cypress-wait-until-something-is-visible-in-the-DOM

supernova-at:
I got a couple errors while validating:

```
Error: The plugins file is missing or invalid.

Your `pluginsFile` is set to `path/to/pwa-studio/venia-integration-tests/src/plugins/index.js`, but either the file is missing, it contains a syntax error, or threw an error when required. The `pluginsFile` must be a `.js`, `.ts`, or `.coffee` file.

Or you might have renamed the extension of your `pluginsFile`. If that's the case, restart the test runner.

Please fix this, or set `pluginsFile` to `false` if a plugins file is not necessary for your project.
```

The file is definitely there and doesn't contain a syntax error. I'm not sure why / how it would throw an error when required.

<hr />

I also got an "update" message immediately from Cypress saying `7.1.0` is out (we're on `6.8.0`). Since we're just introducing Cypress I feel like we should be on the latest major version, at least. Do we have a reason we're not?

supernova-at (changes requested):
Verification errors

revanth0212:
> I also got an "update" message immediately from Cypress saying `7.1.0` is out (we're on `6.8.0`). Since we're just introducing Cypress I feel like we should be on the latest major version, at least. Do we have a reason we're not?

I have upgraded Cypress in this PR. Can you try `yarn install` in the folder instead of the outside which is `pwa-studio`? This package is not hooked up into yarn workspaces so doing a `yarn install` on the mono repo is not gonna install or update this package. 

This was done on purpose because this package is just like docs, has nothing to do with pwa-studio from the outside world's perspective.

venia-integration-tests/src/tests/pageBuilder/homePage.spec.js
revanth0212: Yeah, you guys are right, but in the scope of this PR, I was working on introducing the plugin and integrating it not dint quite get into the test itself. In the following tickets, we can delegate this to the developer working on the ticket about what they think is the right choice of wait or use a helper as Dave suggested.

The test in this PR is a dummy one, just there to prove that this plugin works after the integration.

supernova-at (dismissed):
Worked great once I installed everything properly 😅

anthoula:
@revanth0212 Observed that the sample visual test (pageBuilder/homePage.spec.js) is not passing, and there is a failure at `dangerfile.js`
Screen Shot 2021-04-28 at 3 02 37 PM

revanth0212:
I have made some changes to the comparison logic and hopefully, it will work across machines this time.
 
image

venia-integration-tests/src/tests/pageBuilder/homePage.spec.js
revanth0212: Yeah, it is needed for IntelliSense to suggest cypress auto completions.

dpatil-magento:
@revanth0212 It still fails when screen shot captured window size differs current running window size.

venia-integration-tests/src/plugins/resizeSnapshotPlugin.js
```diff
@@ -0,0 +1,65 @@
+const jimp = require('jimp');
+const fs = require('fs');
+const { matchImageSnapshotPlugin } = require('cypress-image-snapshot/plugin');
+
+const resizeSnapshot = imageConfig => {
```
jimbo, line 5: Is this the function that resolves differences in terms of resolution and pixel density?

venia-integration-tests/src/support/__tests__/loadFullPageCommand.spec.js
```diff
@@ -0,0 +1,17 @@
...
+            add: mockAdd
+        }
+    };
+});
+
+afterEach(() => {
+    delete global.Cypress;
+});
+
+test('sup', () => {
```
jimbo, line 15: Can this test be skipped?

venia-integration-tests/src/plugins/resizeSnapshotPlugin.js
revanth0212: Yes, that's the one.

anthoula:
Tests pass. QA approved. Back @revanth0212 to confirm if there will be additional commits

dpatil-magento (approved):

PR #3146, opened 2021-04-28
Wishlist cypress test

dpatil-magento:
## Description

This is E2E test for single wishlist. Also included Category feature which is in progress.

## Related Issue

Closes [PWA-1677](https://jira.corp.magento.com/browse/PWA-1677).

### Verification Steps

1. Checkout this branch locally and update test url to https://develop.pwa-venia.com/
2. `cd venia-integration-test`
2. `yarn install && yarn test`
## Checklist


venia-integration-tests/src/fixtures/accountAccess/index.js
```diff
@@ -0,0 +1,10 @@
+export const firstName = 'John';
+
+export const lastName = 'Doe';
+
+export const accountPassword = '123123^q';
+
+// Need unique email for test account creation
+const uuid = () => Cypress._.random(0, 1e6)
+const id = uuid()
+export const accountEmail = `${id}test@tester.com`
```
fooman, line 10:
```suggestion
export const accountEmail = `${id}test@example.com`
```
Best to take one of these https://tools.ietf.org/html/rfc2606

dpatil-magento: thanks for the feedback @fooman , updated.

venia-integration-tests/src/tests/wishList/e2eTests/singleWishlistAddRemoveProduct.spec.js
```diff
@@ -0,0 +1,54 @@
...
+import { addProductToWishlistFromProductPage } from '../../../actions/productPage/index';
+
+
+// TODO add tags CE, EE to test to filter and run tests as needed
+describe('verify single wishlist basic features', () => {
+    it('user should be able to add and remove products from wishlist', () => {
+        visitPage(homePage);
+        openLoginDialog();
+        createAccount(firstName, lastName, accountEmail, accountPassword);
+        assertCreateAccount(firstName);
```
revanth0212, line 34: I understand why you went for the re-usable function but I feel we should assert in the test instead of using it as a common utility. That's how we do it in the jest tests as well. The reason being, when someone is reading the test, the ultimate thing a test is gonna prove is the assertion and they would like to read the assertion in the test. It gets hard if we don't know what the assertion does.

Thats my view. Thoughts guys?

venia-integration-tests/src/support/visitPageCommand.js
```diff
@@ -0,0 +1,6 @@
+const visitPage = (routeUrl, wait = 5000) => {
```
fooman, line 1: I am not convinced that this approach is going to match end user behaviour. An end user will click themselves through the react app to navigate. This has a different behaviour to typing the url in the address bar.  So each test should at max have 1 visitPage use at the very start as the entrypoint.

Additionally I suggest removing the hard coded wait. From experience I can say that some will require a lot more than 5000, depending on back-end. It would be best to wait for the appropriate spinning loader to no longer be present or the next element required for the next interaction being there.

dpatil-magento: Yeah agree on that point @fooman however these direct page visit would be used to avoid retesting already tested user flow part of other tests which makes test more lighter, stable and decoupled, examples -
1. Add product to cart - Actual user navigation to PDP would be part of category listing test so when need to add product to cart we would direct hit pdp.
2. Order placement flows - we will use user navigation to `cart` `checkout` pages in 1 or 2 tests but when not needed to check link functionality `Edit Shipping Bag`  `checkout` or `go to checkout` but still need to test something on checkout flows then will hit direct checkout.
3. Actual user category visits would be part of Mega menu tests but rest of the times we will directly go to needed category.

Regrading 5 sec wait - currently we dont have better way to handle full page load so might be temporary until we find better solution. One example - When we visit home page and try to create account >> cypress automatically waits and clicks on `Sign In`  button but at this time the page is still loading. Now when page loads the signin dropdown is collapsed, cypress does not find `Create an Account` button and fail the test.  In this case we did try to wait for home page banner image to load but its not same banner image always, it keeps changing and test fails randomly.

venia-integration-tests/src/actions/wishlist/index.js
```diff
@@ -0,0 +1,74 @@
...
+    removeProduct,
+    deleteProduct
+} from '../../fields/wishlist';
+import { wishlistRemove, removeProductMessage } from '../../fixtures/wishlist';
+
+/**
+ * Utility function to assert empty wishlist
+ * @param {String} wishlistHeaderText wishlist page header text 
+ */
+export const assertWishlistHeading = (wishlistHeaderText) => {
```
revanth0212, line 17: I am thinking about moving these assertions to a folder of their own. Ill make the changes and push the commit.

revanth0212 (approved):

revanth0212 (approved):

venia-integration-tests/src/tests/wishList/e2eTests/singleWishlistAddRemoveProduct.spec.js
```diff
@@ -0,0 +1,88 @@
...
+const { addProductToWishlistFromProductPage } = productPageActions;
+
+const { assertCreateAccount } = myAccountMenuAssertions;
+const {
+    assertWishlistHeading,
+    assertEmptyWishlist,
+    assertProductInWishlist
+} = wishlistAssertions;
+
+// TODO add tags CE, EE to test to filter and run tests as needed
```
sirugh, line 44: Is this something that should get done before we merge this? If not, can you open a ticket and add to the comment?

venia-integration-tests/src/tests/wishList/e2eTests/singleWishlistAddRemoveProduct.spec.js
sirugh: We discussed in chat but I don't know if that was ever applied here - I think we said to go with functions for non assertions, such as `navigateToCart` and `addItemToCart` but to use assertions for things like `didAddItemToCart`

venia-integration-tests/src/assertions/wishlist/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+    // assert Product Price exists
+    cy.contains('div', productName)
+        .children(wishlistItemPrice)
+        .should('exist');
+
+    // assert Add To Cart Section  exists
+    cy.contains('div', productName)
+        .children(wishlistItemAddToCartSection)
+        .should('exist');
```
sirugh, line 46: [Image from Gyazo](https://gyazo.com/7551fbd0aa603a6cbee2b1652615ea9d)

sirugh (changes requested):
Test failed, also some comments.

venia-integration-tests/src/tests/wishList/e2eTests/singleWishlistAddRemoveProduct.spec.js
dpatil-magento: We have decided to have assertions also part of functions which makes it easy to maintain in long run, examples can be found here https://github.com/magento/pwa-studio/tree/dev/wishlist-cypress/venia-integration-tests/src/assertions

revanth0212 (dismissed):

sirugh (approved):
QA Approved, though there is an outstanding TODO that should get addressed :)

venia-integration-tests/src/tests/wishList/e2eTests/singleWishlistAddRemoveProduct.spec.js
revanth0212: We had a couple of discussions about this in slack and in the Dev time. The decision was to try and omit having assertions as functions but if the assertions are being used too many times and are too many in number, its ok to co-locate them into a function if the developer feels so. 

The important thing is to not overdo it because with multiple developers contributing its hard to keep track of these functions and quickly we can end up with too many assertions and some duplicates.

revanth0212 (approved):

PR #3182, opened 2021-05-19
[feature] Add products to wishlist from peregrine gallery

sirugh:
## Description

In #3105 the `Gallery` and `GalleryItem` components were augmented with logic and functionality to:

* initialize a local, cached list of skus within a customer's wishlists and
* render a button (the heart) for adding an item to a list
* render that button as filled or not, based on whether the product sku is in the local array

This PR takes the majority of that logic, specifically the first bullet above, and moves it into a re-useable hook, `useCustomerWishlistSkus` that can be run in a component to "prep" the cache array for comparison by some child component such as the [GalleryButton](https://github.com/magento/pwa-studio/blob/pwa-1672/add-to-wishlist-from-cms/packages/peregrine/lib/talons/Wishlist/GalleryButton/helpers/useSingleWishlist.js#L25-L27).

Because of how Pagebuilder's `Products` component was implemented (it renders either a "SlickSlider" or venia's `Gallery` component) I didn't want to call the shared hook directly as that could potentially result in duplicate calls (once in `Products` and once in `Gallery`). Instead, there is a new `Carousel` component which calls this new shared hook and does a minimal amount of work to fetch the `storeConfig` to pass to the `GalleryItems` it renders.

## Potential breaking changes

If this PR is released with #3105 then it is not breaking, but if it goes out in a subsequent release it would be breaking as the operations have moved to a new hook from `useGallery`.

## Related Issue

Closes [PWA-1672](https://jira.corp.magento.com/browse/PWA-1672).

### Verification Steps

1. In CE and EE, check that the gallery view on the home page renders the wishlist button.
2. Validate that the workflows continue to function when adding an item in all situations (only single wishlist enabled/multiple wishlist enabled, etc)
3. Verify functionality from #3105 works on the category pages.

## Checklist


packages/pagebuilder/lib/ContentTypes/Products/products.js
```diff
@@ -95,10 +96,6 @@ const Products = props => {
     const items = restoreSortOrder(skus, data.products.items);
 
     if (appearance === 'carousel') {
-        const galleryItems = items.map((item, index) => {
-            return <GalleryItem key={index} item={item} />;
```
sirugh, line 99: `GalleryItem` takes a store config object from the parent, telling it whether to render the wishlist button or not. In this case I did not want to add duplicative logic to this `product.js` file since it potentially already renders a `Gallery` component which contains that logic. Instead I made a new `Carousel` component that contains the same logic as the `useGallery` component and passes the store config to the children.

packages/pagebuilder/lib/ContentTypes/Products/products.js
```diff
@@ -152,7 +149,7 @@ const Products = props => {
                     centerModeSmallClass
                 ].join(' ')}
             >
-                <SlickSlider {...carouselSettings}>{galleryItems}</SlickSlider>
+                <Carousel settings={carouselSettings} items={items} />
```
sirugh, line 152: Multiple Pagebuilder Products components could be rendered on a page. Potential race condition.

packages/peregrine/lib/hooks/useCustomerWishlistSkus/useCustomerWishlistSkus.js
```diff
@@ -0,0 +1,69 @@
...
+    const [{ isSignedIn }] = useUserContext();
+
+    const [currentPage, setCurrentPage] = useState(1);
+
+    const {
+        client,
+        data: { customerWishlistProducts }
+    } = useQuery(operations.getProductsInWishlistsQuery);
+
+    useQuery(operations.getWishlistItemsQuery, {
```
sirugh, line 26: It's possible that this hook is called multiple times on a single page by different components, for example if a carousel is rendered on a category page.

In that event, we _should_ be safe to assume that the `onCompleted` callbacks are invoked synchronously, and would result in the same list of skus regardless of order.

eug123: I have checked on the page with multiple instances of PageBuilder Carousels and it seems no problems

eug123 (dismissed):

dpatil-magento:
QA Approved

PR #3184, opened 2021-05-20
[feature] Always show a wishlist name and do not show visibility toggle if only one list

sirugh:
## Description

* Added placeholder text for display if there is no wishlist name configured.
* Show or hide visibility toggle based on number of wishlists belonging to a user.

## Related Issue

Closes [PWA-1742](https://jira.corp.magento.com/browse/PWA-1742).

### Verification Steps

1. Verify a wishlist without a name at least has the default "Wish List" name text.
2. In the user's wishlist page, verify that the visibility toggle does not appear if there is only one list.

## Checklist


packages/venia-ui/lib/components/WishlistPage/wishlistPage.js
```diff
@@ -24,7 +24,11 @@ const WishlistPage = props => {
     const wishlistElements = useMemo(() => {
         if (wishlists.length === 0) return null;
         return wishlists.map(wishlist => (
-            <Wishlist key={wishlist.id} data={wishlist} />
+            <Wishlist
+                key={wishlist.id}
+                data={wishlist}
+                shouldRenderVisibilityToggle={wishlists.length > 1}
```
revanth0212, line 30: Would it merit if we create and send it over from the talon instead? By doing so we are opening up the component to have custom definitions for `shouldRenderVisibilityToggle`. In the default case, it will be `wishlists.length > 1`. What do you think?

revanth0212 (approved):
Other than a question I had about using a talon instead, the code looks fine. Nice work buddy.

revanth0212:
There are some test failures. Check it out dude.

packages/venia-ui/lib/components/WishlistPage/wishlistPage.js
sirugh: Sure, sounds good.

revanth0212 (approved):

PR #3186, opened 2021-05-23
#3185 fix - Duplicate react component key props

### LINKED ISSUE #3185: [bug]: Duplicate react component key props produced when selecting variants in product detail carousel
**Describe the bug**

This is my brief understanding but it seems at least in the sample store data, for each product it's default media entry is also a variant. So the variant that is the same as the default also has the same file and label attributes. This in turn causes the Thumbnail components for the default and the variant matching the default to have the same same key so react loses track of the list of thumbnails.
You can see that in carousel.js at #L55.

**To reproduce**

1. In a venia store with venia sample data, go to any product with colour variants
2. Click on the default colour variant (matches the colour of the selected product's default carousel thumbnails).
3. Click on an alternative colour, then click on the default colour again, repeat this
4. Duplicate thumbnails for the default colour variant will be produced

In the console react will warn that it has detected duplicate key props in the thumbnail list

**Expected behavior**

Should be no duplicate keys in the carousel's thumbnail list.

**Screenshots**

[here's a recording of this behaviour](https://i.imgur.com/GkpTSuo.mp4)

**Possible solutions**

I found that using the item's id as in ```key={item.id}``` instead of ```key={`${item.file}--${item.label}`}``` fixed the issue.

-   Device: Debian
-   Browser: Firefox
-   Browser Version: 88.0.1
-   Magento Version: 2.4

-   [ ] `venia-concept`
-   [x] `venia-ui`
-   [ ] `pwa-buildpack`
-   [ ] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

jh0l:
Fix for #3185

## Description

> Steps to reproduce the behavior:
> 
>     1. In a venia store with venia sample data, go to any product with colour variants
> 
>     2. Click on the default colour variant (matches the colour of the selected product's default carousel thumbnails).
> 
>     3. Click on an alternative colour, then click on the default colour again, repeat this
> 
>     4. Duplicate thumbnails for the default colour variant will be produced
> 
> 
> In the console react will warn that it has detected duplicate key props in the thumbnail list
> 
> **Expected behavior**
> 
> Should be no duplicate keys in the carousel's thumbnail list.

* carousel thumbnail keys are not currently unique based on item file + label

* item IDs are unique (within a products variants) so should be a good fix

## Changes

* Change ProductImageCarousel Thumbnail key to unique id

## Related Issue
Closes #3185

### Verification Steps
1. Go to any product page that has colour swatch variants.
2. Verify that selecting between the default variant and alternative variants does not produce duplicate react component keys.


tjwiebell (dismissed):
Simple fix, nice work 👍

sirugh:
@jh0l thanks for the contribution - would you please sign the Adobe CLA? Without that we cannot use/merge this PR :(

jh0l:
Okay I've signed it,

sirugh:
Ok thanks for that - I also was curious how to repro this issue - I'm not able to on https://develop.pwa-venia.com/default/athena-tank-dress.html Do I need to change something in the backend?

sirugh:
Though I wasn't able to repro the reported bug, I do believe this solution is better than what we had previously. `uid` is exactly the field we should be using for key, as it is meant to be unique per item within a list. I don't mind approving this and moving it forward.

sirugh (dismissed):
Approving though I did make some changes (id -> uid).

jh0l:
> 
> 
> Ok thanks for that - I also was curious how to repro this issue - I'm not able to on https://develop.pwa-venia.com/default/athena-tank-dress.html Do I need to change something in the backend?

Yeah there must be some difference in the way the pwa-studio tutorial sets up venia versus how your CI sets up venia. I'm not sure how to see that though.

packages/peregrine/lib/talons/RootComponents/Product/productDetailFragment.gql.js
```diff
@@ -14,7 +14,10 @@ export const ProductDetailsFragment = gql`
         }
         id
         media_gallery_entries {
```
larsroettig, line 16: `media_gallery_entries`  is DEPRECATED Use product's `media_gallery` instead
https://devdocs.magento.com/guides/v2.4/graphql/interfaces/product-interface.html?itm_source=devdocs&itm_medium=quick_search&itm_campaign=federated_search&itm_term=

sirugh (approved):
Approving this although I technically made a few changes (id -> uid).

larsroettig (approved):

PR #3199, opened 2021-05-28
[chore] do not check coverage on .gql javascript files

sirugh:
## Description

Our gql javascript files are being considered in coverage, which makes our coverage numbers look lower than they are.

For example: https://coveralls.io/builds/40065394

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Ensure coveralls does not report on .gql.js files.

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -326,7 +326,10 @@ const jestConfig = {
         // Not this file itself
         '!jest.config.js',
         // Exclude deprecated components from coverage report
-        '!**/venia-ui/lib/components/Checkout/**'
+        '!**/venia-ui/lib/components/Checkout/**',
+        // Not GraphQL files, gql.ce.js or gql.ee.js
+        '!**/*.gql.js',
+        '!**/*.gql.**.js'
     ],
     // Don't look for test files in these directories.
     testPathIgnorePatterns: [


jest.config.js
```diff
@@ -326,7 +326,10 @@ const jestConfig = {
         // Not this file itself
         '!jest.config.js',
         // Exclude deprecated components from coverage report
-        '!**/venia-ui/lib/components/Checkout/**'
+        '!**/venia-ui/lib/components/Checkout/**',
+        // Not GraphQL files, gql.ce.js or gql.ee.js
+        '!**/*.gql.js',
+        '!**/*.gql.**.js'
```
sirugh, line 332: I noticed that the coverage has decreased in this PR, and I think it might be due to `.gql.ee.js` or `.gql.js` files being imported by default during our CI which makes them "covered". You'll notice on [the failure](https://coveralls.io/builds/40065394) for this other job that it is `.gql.ce.js` files that are reported as missing coverage.

Should we _only_ ignore `.gql.ce.js` files then?

sirugh:
Oops - I think the coverage loss is actually informative ;)

PR #3210, opened 2021-06-02
Ticket #3209. Adds check by include_in_menu attribute before appendin…

### LINKED ISSUE #3209: Sidebar menu not filtering by include_in_menu attribute.
**Describe the bug**

Sidebar menu not filtering by include_in_menu attribute.

**To reproduce**

1. Go to Categories.
2. Make some of the categories **include_in_menu** false and some true.
3. Check your PWA
4. See how all of categories are visible instead of only the ones with **include_in_menu** seet to true.

**Expected behavior**

PWA sidebar menu should only show the **include_in_menu** categories.

**Possible solutions**

- A workaround could be early exiting on the `CategoryTree` component:
image

Or not return the children that doesn't apply in the **useCategoryTree** talon

- mobile devices.

-   [***** ] `venia-concept`
-   [ ] `venia-ui`
-   [ ] `pwa-buildpack`
-   [ ] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

aagasi:
…g children categories to the returned array.

## Description

Adds a check to avoid returning children categories when doesn't apply.

Closes #3209 

## Breaking Changes

Breaking because it now does not display content if `include_in_menu` is false.

### Verification Steps

1. Go to the Categories edit form in the admin and make some of them **include_in_menu** true, and some false.
2. Go to your pwa and open the mobile devices view.
3. Make sure you only see the categories that have **include_in_menu** true.


aagasi:
@sirugh any idea why the coverage decreased?

sirugh:
> @sirugh any idea why the coverage decreased?

You can click "Details" and it should tell you :)

packages/peregrine/lib/talons/CategoryTree/useCategoryTree.js
```diff
@@ -70,9 +70,10 @@ export const useCategoryTree = props => {
         }
 
         children.map(category => {
-            const isLeaf = !parseInt(category.children_count);
-
-            childCategories.set(category.id, { category, isLeaf });
+            if (category.include_in_menu) {
```
sirugh, line 73: I'd guess that you need to now test this case.

aagasi: @sirugh  I added the test, but still getting the same result. Apparently Coverrals is referring to a miss branch? 
image

And the other issue a missing label in the PR apparently. But I don't see a button to add label on this PR...

aagasi: @sirugh is there anything I could do? I've seen you added a version label, but that error didn't go away.

sirugh: Don't worry about the label warning, as for the coveralls error I'll have to check but it should highlight the missing coverage if you click the file.

packages/peregrine/lib/talons/CategoryTree/__tests__/useCategoryTree.spec.js
```diff
@@ -266,4 +266,38 @@ describe('child categories', () => {
...
+                getNavigationMenu,
+                {
+                    data: {
+                        category: {
+                            ...result.data.category,
+                            children: [
+                                ...result.data.category.children,
+                                {
+                                    id: categoryId,
+                                    include_in_menu: 0,
```
sirugh, line 283: Because the default data `result` all has `include_in_menu: 1`, we have to augment/add a case that will check the false branch/path.

aagasi:
@sirugh I believe we need a reviewer now?

sirugh:
Correct :) It's in the queue.

supernova-at (approved):

supernova-at (changes requested):
Code looks and works great, thank you!

I would like to request this same fix applied to the [Mega Menu](https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/MegaMenu/megaMenu.js#L35-L46) as well.

supernova-at (approved):

tjwiebell:
@aagasi - During QA I observed this behaving a little different than expected, but will need to confirm with the PM if Luma behavior is expected here. In Luma today the children of a not-in-menu node become siblings of that nodes parent, so that setting shouldn't make the node terminating. Will review with @tkacheva and let you know if this needs any adjustments.

Category Structure:
Screen Shot 2021-06-14 at 9 43 53 AM

Luma:
Screen Shot 2021-06-14 at 9 45 36 AM

Venia:
Screen Shot 2021-06-14 at 9 47 52 AM

tjwiebell:
Current behavior is acceptable, will create followup task for UX to review. QA Approved ✅

PR #3211, opened 2021-06-03
PWA-1790: Spike: Page Builder support - How to enable components to m…

eug123:
## Description

When scroll bar is configured in system to "always show" - it's width is not considered in full viewport width (100vw), so added calculation to subtract it from viewport width, also `ResizeObserver` is used to track if scrollbar appeared after page height was increased.

## Related Issue

Closes #[PWA-1790](https://jira.corp.magento.com/browse/PWA-1790).

### Verification Steps

1. Add 3 rows using PageBuilder to the CMS page.
2. Configure each row to have different backgrounds and appearance settings:
   - Contained
   - Full width
   - Full bleed
3. See rows content types looks correct on the Venia storefront:
    - Full Bleed and Full Width rows background should take full width of the viewport.
    - Full Bleed row inner content should take full width of the viewport
4. Configure system to have a different scroll bar behaviour
    - (on MacOS - System Preferences -> General -> Show scroll bars: Automatically/Always)
5. See how scroll bars influence on rows appearance in the viewport (no horizontal scrollbar should appear)
6. Check other possible row content type settings: Paddings, Margins, Video Background, Parallax, etc. 
7. Run MFTF
8. Verify new/impacted UI elements in supported browsers (chrome, firefox, safari) and devices (Safari- iOS, Chrome- Android)

image

## Checklist


larsroettig (dismissed):
@eug123 looks very nice would move this QA but we merge conflict here may you can solved them before we do QA

eug123:
@larsroettig, conflicts resolved. Thank you!

larsroettig (approved):

packages/pagebuilder/lib/ContentTypes/Row/__tests__/__snapshots__/row.spec.js.snap
```diff
@@ -25,34 +27,28 @@ exports[`render full-bleed row 1`] = `
 />
 `;
 
-exports[`render row with all props configured 1`] = `
+exports[`render full-width row 1`] = `
 <div
-  className="root test-class"
+  className="fullBleed root"
```
supernova-at, line 32: What's the difference between the full bleed and the full width snapshots here?

supernova-at:
## QA

- [x] Test Plan / Verification Steps
- [ ] MFTF
- [ ] Cypress
- [ ] Lighthouse
- [x] Alternate Browsers

PR #3215, opened 2021-06-03
fix: Remove the extension from venia concept by default

sirugh:
## Description

ACDL may break if it is double loaded. Some downstream dependencies might auto load it so we should remove the default inclusion of the extension. I have also provided guidance for future use.

## Related Issue

Closes PWA-1815.

### Verification Steps

1. Scaffold an app from this branch.
2. Ensure that the ACDL extension is not auto included.

## Checklist


jimbo (approved):

eug123:
- [ ] Scaffold an app from this branch.
- [ ] Ensure that the ACDL extension is not auto included.
Verification steps passed

PR #3223, opened 2021-06-08
[DRAFT] Restore Style Loader

tjwiebell:
## Description

Cleanup work to get this target branch into QA. The FOUC introduced by iso-style-loader would have been a bad state to leave `develop` in, and SSR isn't likely to land for a few more months. We'll bring it back in once SSR is done, and that FOUC is solved by serving pre-rendered content.

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Go to the FOO page.
2. Verify the BAR shows up.

## Checklist


packages/peregrine/lib/talons/Adapter/useAdapter.js
```diff
@@ -162,12 +162,12 @@ export const useAdapter = props => {
...
-                authLink,
-                errorLink,
-                httpLink,
                 mutationQueueLink,
                 retryLink,
-                storeLink
+                authLink,
+                storeLink,
+                errorLink,
+                httpLink
```
tjwiebell, line 170: The order of these is important to Apollo, with the terminating http link required to be at the end.

packages/venia-concept/webpack.config.js
```diff
@@ -121,12 +122,29 @@ module.exports = async env => {
...
+
     // add LimitChunkCountPlugin to avoid code splitting
     serverConfig.plugins.push(
         new LimitChunkCountPlugin({
             maxChunks: 1
         })
     );
 
-    return [config, serverConfig];
+    return config;
```
tjwiebell, line 149: Could only get one config to build at a time, complains about missing i18n generated file if this returns both configs.

packages/peregrine/lib/talons/Adapter/useAdapter.js
sirugh: Is this a bug?

tjwiebell: A bug introduced by the target branch; this change restores the link order that is present in mainline today.

packages/venia-concept/webpack.config.js
```diff
@@ -114,13 +116,32 @@ module.exports = async env => {
...
+        .add('WebpackAssetsManifest');
 
     // remove browser-only plugins
     serverConfig.plugins = serverConfig.plugins.filter(
         plugin => !browserPlugins.has(plugin.constructor.name)
     );
 
+    // remove browser-only module rules
+    serverConfig.module.rules = serverConfig.module.rules.map(rule => {
+        if (rule.oneOf) {
```
jimbo, line 130: I'm okay with this, but it's a bit fragile. Could we look for the module rule we want to remove by its `test` value rather than its `oneOf` value? Would that work?

jimbo (changes requested):

jimbo:
I'll make the requested changes to my branch.

PR #3228, opened 2021-06-10
PWA-1707: Cleanup whishlist for render empty

larsroettig:
## Description

Clean up single wishlist

## Related Issue

Closes PWA-1707

### Verification Steps

My Account > Favorites shows a "there are currently no items in this list" message
After items are added and then removed back down to zero items, the "there are currently no items in this list" message shows again
Test Plan

Navigate to My Account > Favorites with a single, empty wishlist
See that the page shows the "no items" message
Add an item to the wishlist
Remove the item from the wishlist
See that the page again shows the "no items" message

Screenshot from 2021-06-10 14-05-44

## Checklist


packages/venia-ui/lib/components/WishlistPage/wishlistPage.js
```diff
@@ -27,7 +27,10 @@ const WishlistPage = props => {
         defaultMessage: 'The wishlist is not currently available.'
     });
     const wishlistElements = useMemo(() => {
-        if (wishlists.length === 0) return null;
+        if (wishlists.length === 0) {
+            return <Wishlist data={[]} />;
```
tjwiebell, line 31: With all the destructuring happening in the component, was surprised to see it not complain about this array value. Instead of making the placeholder dependent on the developer passing some arbitrary special value, let's instead make `<Wishlist />` with no data prop behave like a placeholder. That paired with adding `defaultProps` to the component should accomplish the same thing with a better API.

Optionally, it looks like I missed adding `propTypes` on that component. Would be nice to document if you are going to pass data, what is expected in that object to render correctly.

packages/venia-ui/lib/components/WishlistPage/wishlist.css
```diff
@@ -19,6 +19,10 @@
     grid-auto-flow: column;
 }
 
+.emptyListText {
+    text-align: center;
```
tjwiebell, line 23: Minor UX observation; mocks appear to have a minimum height that isn't reflected here. Giving the grid container a `min-height` is preferable here, but a simple `line-height` here could accomplish the same thing. If you want to attempt `min-height` solution, you'll need to give the grid alignment instructions, since it appears to want to stretch. If this has been approved by UX, no big deal.

tjwiebell (changes requested):
Couple observations to look at, but this look pretty straight forward. Nice work 👍

packages/venia-ui/lib/components/WishlistPage/__tests__/wishlist.spec.js
```diff
@@ -90,6 +90,6 @@ test('hides visibility toggle', () => {
 });
 
 test('render no button when id is not set', () => {
-  const tree = createTestInstance(<Wishlist data={{}} />);
-  expect(tree.toJSON()).toMatchSnapshot();
-})
+    const tree = createTestInstance(<Wishlist data={{}} />);
```
tjwiebell, line 93: Test should match the new implementation.
```suggestion
    const tree = createTestInstance(<Wishlist />);
```

packages/venia-ui/lib/components/WishlistPage/wishlist.css
tjwiebell: My preference is to let the grid align the content, but this is fine 👍

tjwiebell (approved):
Looks good to send to QA. Please resolve linting and test failures though 🙂

supernova-at:
## QA

- [x] Verification Steps
- [x] MFTF
- [x] JIRA ticket test plan
- [x] Lighthouse
- [x] Alternate browsers
⚠️ Safari still does not work but that is a known issue not introduced by this PR ⚠️

supernova-at (approved):

PR #3229, opened 2021-06-10
Fix incorrect prop type for Icon

LordZardeck:
Currently there are prop type errors on a fresh install of venia, due to Icon expecting a function, when in all reality, it's expecting an element type. I've made this quick fix in this PR

### Resolved issues:
1. [x] resolves magento/pwa-studio#3305: Fix incorrect prop type for Icon


supernova-at:
@magento create issue

larsroettig:
Hi @LordZardeck,
can you pls add steps to verify I don't get this warning in the current development branch?

Kind regards, 

Lars

brendanfalkowski:
In our project, we were passing both raw `.svg` files (which are objects) and imports from `react-feather` (which were functions, but are now React elements) into the `<Icon>` component as `src` prop. So we had to make a customization to `<Icon>`'s prop types like `src: oneOfType([func, object])` to support both those types.

But now `elementType` seems to work for both cases without throwing warnings.

## Example

Here's an example of code that would break with `src: PropTypes.func`:

```
import { X as ResetIcon } from 'react-feather';
const resetIcon  = <Icon size={20} src={ResetIcon} />;
```

> Warning: Failed prop type: Invalid prop `src` of type `object` supplied to `Icon`, expected `function`. in Icon

I don't remember exactly, but at some point React Feather changed from exporting a function to exporting a named React element (aka an `elementType`).

## Aside

The naming/docs from React PropTypes aren't helpful explaining the the difference is between `element` and `elementType` but this helped: https://stackoverflow.com/questions/42326084/what-is-the-correct-proptypes-for-component-or-element

```
element is for an instance (<Foo />)
elementType is for a component class (Foo)
```

And also apparently PropTypes will accept a React element type-matching as `object` too, which makes sense because it's JavaScript but I guess it's not really a useful check.

larsroettig:
Hi @brendanfalkowski, thank you for your comment looks like @LordZardeck doesn't sign CLA, so that I will close this PR. There is no way to reach him. If you want, you can open a new PR with this Information.

PR #3230, opened 2021-06-10
Fixing mega menu when pathname is undefined

### LINKED ISSUE #3192: [bug]: unhandled edgecase for path "/undefined"
**Describe the bug**

When accessing a page with the path `/undefined` like for example `https://venia.magento.com/undefined`, then this leads to a list of errors till the page breaks. (was posted by someone else in slack)

**To reproduce**

see description.

**Expected behavior**

getting the expected page for this url without errors.

**Screenshots**

!https://user-images.githubusercontent.com/237319/119545818-85595b00-bd93-11eb-976e-87b90643f72f.png!

**Possible solutions**

the cause for this error is located in 

https://github.com/magento/pwa-studio/blob/9a2f1d32bfa91d05f2e8a07b6d7a137e10f203c5/packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js#L43-L50

the minifier transforms this into:
```
                 var t = e.url_path
                      , r = e.url_suffix
                      , a = "/".concat(t).concat(r || "")
                    return n.pathname === a
```
and as javascript is a bit special, it transforms an undefined variable into the string `undefined`, so: `"/".concat(undefined): "/undefined"`, which accidentially matches, if the url path is actually `/undefined`

So the solution would be an additional check if `url_path` is `undefined`

or alternativley in
https://github.com/magento/pwa-studio/blob/9a2f1d32bfa91d05f2e8a07b6d7a137e10f203c5/packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js#L107-L118

to check if `activeCategory.path` is actually defined, but as one sees, it should not actually match as an `activeCategory`, as this category does not exist.

**Debug Report**

## 

-   Device <e.g. iPhone6, PC, Mac, Pixel3]:
-   Browser [e.g. Chrome, Safari]:
-   Browser Version [e.g. 22]:
-   Magento Version [e.g Magento Commerce 2.4]:

[!-- Complete the following sections to help us apply appropriate labels! --]

-   [ ] `venia-concept`
-   [ ] `venia-ui`
-   [ ] `pwa-buildpack`
-   [x ] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ > `create-pwa`

LucasCalazans:
## Description

Fixing an error when the user goes to /undefined route.

The useMegaMenu talon was understanding that is the default category, causing the error

## Related Issue

Closes #3192.

## Verification Steps

- Access this route /undefined
- The store should show a not found page with no error

## Test Plan

- Access this route /undefined
- The store should show a not found page

- Access this route /undefined
- The store should show a not found page

No

No

image

## Checklist


supernova-at (approved):

supernova-at:
## QA

- [ ] MFTF
- [x] Cypress
- [ ] Test Plan / Verification Steps
- [x] Lighthouse
- [x] Alternate Browsers

PR #3218, opened 2021-06-04
Cypress multiple wishlist 

dpatil-magento:
## Description

Cypress test for multiple wishlist

## Related Issue

Closes [PWA-1808](https://jira.corp.magento.com/browse/PWA-1808)

### Verification Steps

1. Clone this branch and `cd venia-integration-tests/`
2. `yarn install`
3. `yarn test`
4. From cypress UI run both wishlist E2E and integration tests and make sure they pass.

## Checklist

-   I have added tests to cover my changes, if necessary.
-   I have added translations for new strings, if necessary.
-   I have updated the documentation accordingly, if necessary.

PR #3242, opened 2021-06-17
Add to wishlist bug fix.

revanth0212:
## Description

There was a bug in the wishlist dialog due to which a user would be stuck with the create dialog, unable to create a new wishlist on the PLP or the PDP page. 

## Related Issue

Closes PWA-1851

## Acceptance

Should be able to add an item to a new or an existing wishlist.

### Verification Stakeholders

@dpatil-magento 

## Test Plan

Go to PLP or the PDP page and try to add an item to a wishlist.

-   Verify user is able to add an item to a new or an existing wishlist from the PLP page.
-   Verify user is able to add an item to a new or an existing wishlist from the PDP page.

Yes

## Breaking Changes (if any)

None

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/Wishlist/WishlistDialog/wishlistDialog.js
+++ b/packages/venia-ui/lib/components/Wishlist/WishlistDialog/wishlistDialog.js
@@ -118,7 +118,7 @@ const WishlistDialog = props => {
 
 export default WishlistDialog;
 
-WishlistDialog.defaultProps = {
+WishlistDialog.propTypes = {
     classes: shape({}),
     isOpen: bool,
     isLoading: bool,


tjwiebell (approved):
Bizarre this didn't pop up until now, but a simple fix 👍  we typically like to make sure bugs are covered by a test, but that might be difficult given these are React internals. Since none of these props are marked as required, it might be good to test expected behavior when no props are passed.

We don't block approval on tests, so sending to QA.

dpatil-magento:
Checked with @revanth0212 and we cant have unit test for this case but we will soon have integration test part of https://github.com/magento/pwa-studio/pull/3218 so fix is covered.

dpatil-magento:
QA Approved.

PR #3246, opened 2021-06-23
PWA-1683 - My Account: Multi-Wishlist: clean-up 

larsroettig:
## Description
My Account: Multi-Wishlist: clean-up

## Related Issue
[PWA-1683] My Account: Multi-Wishlist: clean-up

## Acceptance
- Login toasts are rendered if user clicks to add to wishlist but is not signed in
- If EE only one wishlist is allowed, no dialog (to choose a wishlist) should appear
- The "general" Wishlist Button component takes props such as localized "buttonText" string for customization
- The "general" Wishlist Button component takes props for beforeAdd and afterAdd callbacks

### Specification

- Wishlist is enabled in Magento Admin Panel > Stores > Settings > Configuration > Customers > Wish List
- Multi-wishlist is not enabled in Stores > Settings > Configuration > Customers > Wish List > Enable Multiple Wish Lists (EE) OR Maximum number of wishlists is set to one (1)

- Create a new wishlist
- Verify that you are not prompted to set its "visibility" (pubic / private)
- Edit an existing wishlist
- Verify that you are not prompted to set its "visibility" (public / private)
- Verify that you can create as many wishlists as are set in the backend (Stores > Settings > Configuration > Customers > Wish List)
- Verify that attempting to add another wishlist after reaching that threshold presents an error message and lets you recover gracefully
- Add many products to the wishlist (20+) and verify
- Verify that the first wishlist in the list is the most recently created
- Verify that the first wishlist is expanded by default. If there is a second list it is collapsed by default.
- Verify the lists have text in the header of the List container(s) that displays the number of visible items and the total number of items in the list.
- Verify that if there are more than 20 items saved to a List, a "Load More" button appears at the bottom after the first 20 (visible) items.
- Verify that when user taps the "Load More" button the next 'batch' of saved items (1-20 items) should appear in row(s) beneath the first 20.
- Verify that text in the container header updates to include next 20 loaded items in the count - ( ex. "Showing 20 of 58 items in this list" -> "Showing 40 of 58 items in this list" )
- Verify that when there are no more items in the list the "Load More" button doesn't appear;
- Verify that items are returned in the order of most recently added items
- Verify that clicking the "load more" button loads more wishlist items
- Verify that the "load more" button remains visible (after the items) until all items have been loaded
- Verify that for each product you can add to cart and remove from wishlist

Example
-   Verify user is able to sort data after applying filters on category page
-   Verify user is able to sort data after applying filters on search page

Example
- Update default Sort value in backend and repeat above scenarios

Example
- Yes, browser testing is needed as X UI component may be impacted on <browser>
- Yes, device testing is needed as X UI component may be impacted on <mobile|desktop|etc>

Example
- Apply all filters to get 0 results and then remove filters to see respective products 

## Checklist


packages/peregrine/lib/talons/WishlistPage/useWishlist.js
```diff
@@ -1,20 +1,74 @@
...
+    }, [id, fetchMore, page]);
+
+    useEffect(() => {
+        setPage(1);
+        if (itemsCount >= 1 && isOpen === true && !data) {
+            fetchWhislistItems();
+        }
+    }, [itemsCount, isOpen, fetchWhislistItems, data]);
+
+    const items =
```
revanth0212, line 59: Can you memoize this in a `useMemo`?

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
```diff
@@ -12,6 +12,16 @@ const dialogs = {
     MORE_ACTIONS: 3
 };
 
+const SUPPORTED_PRODUCT_TYPES = ['SimpleProduct', 'ConfigurableProduct'];
+
+const mergeSupportedProductTypes = (supportedProductTypes = []) => {
+    if (supportedProductTypes.length >= 0) {
```
revanth0212, line 18: This might fail if `supportedProductTypes` is `null`. `default param` only will pick if its `undefined`.

image

Instead try something like this:

```es6
const mergeSupportedProductTypes = (supportedProductTypes = []) => {
    const newSupportedProductTypes = [...SUPPORTED_PRODUCT_TYPES]

    if (supportedProductTypes) {
       newSupportedProductTypes.push(...supportedProductTypes)
    }

    return newSupportedProductTypes;
};
```

image

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
```diff
@@ -23,19 +33,26 @@ const dialogs = {
...
     const {
         configurable_options: configurableOptions = [],
+        __typename: productType,
         image,
         sku,
         stock_status: stockStatus
     } = product;
     const { label: imageLabel, url: imageURL } = image;
 
+    const isSupportedProductType = mergeSupportedProductTypes(
```
revanth0212, line 52: Can you memoize this using a `useMemo`?

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
```diff
@@ -140,13 +173,15 @@ export const useWishlistItem = props => {
...
     const handleRemoveProductFromWishlist = useCallback(async () => {
+        setIsDeleting(true);
         try {
             await removeProductsFromWishlist();
 
             // Close the dialogs on success.
             setCurrentDialog(dialogs.NONE);
         } catch (e) {
             console.error(e);
+            setIsDeleting(false);
```
revanth0212, line 184: Shouldn't this be part of the try block as well? I understand the item will be removed but I think for aesthetics purposes it makes sense to reset it. I might be wrong, maybe it will cause a react update error because the component will be unmounted.

If you think it does cause any error and what I said makes sense, move this to the finally block so it will run irrespective of the item being removed or canceled.

I am fine either way 😄 .

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
```diff
@@ -193,6 +227,8 @@ export const useWishlistItem = props => {
         };
     }, [imageLabel, imageURL]);
 
+    const [isDeleting, setIsDeleting] = useState(false);
```
revanth0212, line 230: Please move this to the top of the file. It's usually not advised to define variables after they are used in the code. 

Also, the usual norm we follow is to, define state, context, operations, handlers, and effects in that order for consistency in a talon.

revanth0212 (changes requested):
Wonderful work Lars.

I have provided minor feedback. Please check it.

packages/peregrine/lib/talons/WishlistPage/useWishlist.js
larsroettig: Duplication of Apollo state can lead to strange issues in the app
also, Kent recommends only using useMemo if it is a complex operation, I would leave it like it here. 

**Reference:
https://kentcdodds.com/blog/usememo-and-usecallback**

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
larsroettig: memo can bring here 1-3 ms i did short benchmark but i dobt know if it wohrt it to memo it https://codesandbox.io/s/affectionate-shamir-4mm8n?file=/src/App.js

packages/peregrine/lib/talons/WishlistPage/__tests__/__snapshots__/useWishlist.spec.js.snap
```diff
@@ -2,7 +2,12 @@
 
 exports[`returns correct shape 1`] = `
 Object {
+  "error": undefined,
```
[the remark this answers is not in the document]
larsroettig: Because I use `useLazyQuery` this return `undefined`

packages/peregrine/lib/talons/WishlistPage/useWishlist.js
```diff
@@ -1,20 +1,74 @@
...
  */
-export const useWishlist = () => {
-    const [isOpen, setIsOpen] = useState(true);
+export const useWishlist = (props = {}) => {
+    const { id, itemsCount, collapsed } = props;
+    const operations = mergeOperations(defaultOperations, props.operations);
+
+    const [page, setPage] = useState(1);
+    const [isOpen, setIsOpen] = useState(!collapsed);
+    const [isFetchMore, setIsFetchMore] = useState(false);
```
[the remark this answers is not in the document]
larsroettig: isFetchMore mean the current state if the app currently fetch more data I recomand to not rename it for guessing

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
larsroettig: `mergeOperations` is a the same of cost and is also nowhere with memo would recommend don't memo this as well

packages/venia-ui/lib/components/WishlistPage/wishlistEditFavoritesListDialog.css
```diff
@@ -22,6 +22,6 @@
 }
 
 .radioRoot {
-    composes: root from '../RadioGroup/radioGroup.css';
-    grid-template-columns: auto;
+    /** schould be display if we support multiple types */
```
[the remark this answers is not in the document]
larsroettig: good catch schould remove complety

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
larsroettig: finallyBlock dont work strange UI behaivor try differnt things to get nice UI dont work well here

packages/peregrine/lib/talons/WishlistPage/useWishlist.js
revanth0212: This is a norm we follow in the codebase. We want the whole codebase to look similar. We always wrap such calculations in a `useMemo` because we don't want them to run on every single render even if the logic is trivial or complex.

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
revanth0212: If the inputs are not changing, why would it have merit to run the same logic on every render? Wouldn't it make sense to memorize it even if it is a simple calculation? 

Memoizing such calculation is a standard norm we follow in the code base.

packages/venia-ui/lib/components/WishlistPage/wishlist.js
```diff
@@ -17,36 +19,63 @@ import ActionMenu from './actionMenu';
  * @param {boolean} props.shouldRenderVisibilityToggle whether or not to render the visiblity toggle
  */
 const Wishlist = props => {
-    const { data, shouldRenderVisibilityToggle } = props;
+    const { data, shouldRenderVisibilityToggle, collapsed } = props;
```
revanth0212, line 22: Can you add `collapsed` to the JS Docs? Also, if this is a boolean value, please called it `isCollapsed` instead.

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
revanth0212: So how about having this in the try block as well?

packages/peregrine/lib/talons/WishlistPage/useWishlist.js
revanth0212:
```es6
const items = useMemo(() => {
     return data && data.customer.wishlist_v2.items_v2.items
            ? data.customer.wishlist_v2.items_v2.items
            : []
}, [data])
```

tjwiebell: I agree with @larsroettig here; this is derived state and does not need to be memoized. Apollo is already memoizing the result and we can expect this data not to change between renders unless variables change.

packages/peregrine/lib/talons/WishlistPage/useWishlistItem.js
larsroettig: ### Current behaihor 
Default Store View
### behaihor  after change
Default Store View

Problem to invalidate apollo cache it takes time

revanth0212: Cool, then lets stick to what we had before. You can ignore this comment.

revanth0212 (changes requested):
Almost there, just some changes on the wording and some tests.

packages/peregrine/lib/Apollo/policies/__tests__/index.spec.js
```diff
@@ -405,3 +405,23 @@ test('local customerWishlistProducts field returns expected data', () => {
     expect(defaultValue).toMatchInlineSnapshot(`Array []`);
     expect(existingValue).toBe(existing);
 });
+
+test('WishlistItems type merges items correctly first page', () => {
```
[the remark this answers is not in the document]
larsroettig: :+1:  fixed

revanth0212 (changes requested):
Other than danger errors, the PR looks ready for QA. Wonderful work @larsroettig 👏

revanth0212 (approved):

supernova-at:
## QA

_In progress_

- [ ] MFTF
- [ ] E2E and Integration Tests (Cypress)
- [ ] JIRA test plan
- [ ] Lighthouse
- [ ] Alternate Browsers

PR #3262, opened 2021-07-12
Add newsletter in footer

### LINKED ISSUE #3264: [Issue] Add newsletter in footer
This issue is automatically created based on existing pull request: magento/pwa-studio#3262: Add newsletter in footer

---------

## Description

Add Newsletter in the footer
## Related Issue

Closes #ISSUE_NUMBER.

image-1-732x388
image jpeg-300x145

## Checklist

shikhamis11:
## Description

Add Newsletter in the footer
## Related Issue

Closes #3264

image-1-732x388
image jpeg-300x145

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#3264: Add newsletter in footer


shikhamis11:
@magento create issue

supernova-at:
Hi @shikhamis11 , just checking in on this one - are you still working on it?

When you fix the conflicts PWAStudioBot will run again and hopefully some errors will clear.

I'd love to get this to the team for review. Thanks!

shikhamis11:
Hi @supernova-at, 
yes,  I have fixed the conflict.

packages/venia-ui/lib/components/Newsletter/newsletter.js
```diff
@@ -0,0 +1,97 @@
+import React, { Fragment, useEffect } from 'react';
```
JustinConabreeAbsolunet, line 1: Import order should be changed, and imports within venia-ui should be changed to relative paths.

Example
```
import React, { Fragment, useEffect } from 'react';
import { FormattedMessage, useIntl } from 'react-intl';
import { Form } from 'informed';
import { shape, string } from 'prop-types';

import { useNewsletter } from '@magento/peregrine/lib/talons/Newsletter/useNewsletter';
import { useToasts } from '@magento/peregrine';

import { isRequired } from '../../util/formValidators';
import FormError from '../FormError';
import Button from '../Button';
import LoadingIndicator from '../LoadingIndicator';
import TextInput from '../TextInput';
import defaultClasses from './newsletter.css';
```

packages/venia-ui/lib/components/Newsletter/newsletter.js
```diff
@@ -0,0 +1,97 @@
...
+                    id: 'newsletter.subscribeMessage',
+                    defaultMessage: 'The email address is subscribed.'
+                }),
+                timeout: 5000
+            });
+        }
+    }, [addToast, formatMessage, newsLetterResponse]);
+    if (isBusy) {
+        return (
+            <div className={classes.modal_active}>
```
JustinConabreeAbsolunet, line 43: Try passing the `global={true}` flag to LoadingIndicator to see if this removes the need for the modal wrap. This is what is used by the `fullPageLoadingIndicator`

packages/venia-ui/lib/targets/venia-ui-intercept.js
```diff
@@ -60,4 +60,18 @@ module.exports = veniaTargets => {
     });
 
     new CategoryListProductAttributes(venia);
+
```
JustinConabreeAbsolunet, line 63: Since we're inside of venia, the import and usage should be added directly to the footer.js file. No need for targetables.

JustinConabreeAbsolunet (changes requested):
Overall great job! Just some small changes to make future maintenance easier.

packages/venia-ui/lib/components/Newsletter/newsletter.js
```diff
@@ -0,0 +1,95 @@
...
+import { mergeClasses } from '../../classify';
+import FormError from '../FormError';
+import Button from '../Button';
+import LoadingIndicator from '../LoadingIndicator';
+import TextInput from '../TextInput';
+import defaultClasses from './newsletter.css';
+
+const Newsletter = props => {
+    const { formatMessage } = useIntl();
+    const classes = mergeClasses(defaultClasses, props.classes);
```
JustinConabreeAbsolunet, line 19: @shikhamis11 Sorry, I missed this one in the first review. `mergeClasses` should be changed for `useStyle`.
```
import { useStyle } from '../../classify';
// ...
const classes = useStyle(defaultClasses, props.classes);
```

shikhamis11:
@JustinConabreeAbsolunet thank you for your review. I have updated my changes Please check

packages/venia-ui/lib/components/Newsletter/newsletter.js
shikhamis11: sure

shikhamis11:
@JustinConabreeAbsolunet thank you for your review. I have updated my changes Please check

tjwiebell:
@shikhamis11 I had concerns about what would happen if newsletters were disabled, but it appears this has uncovered a GraphQL bug where that mutation will succeed regardless of backend config. Given that, two things of note:

1. I believe we discussed awhile ago converting this into an extension, so implementers can opt-in to this component if they want to support newsletters. At some point that GraphQL endpoint should start erring, and we try to avoid errors that the shopper can do nothing about. Until `storeConfig` starts returning this newsletter setting, opting in seems like the best solution.
2. Our GitHub/JIRA sync appears to have skipped a very important approval from Product and UX. We'll need to get this in front of @tkacheva and @schensley to validate if this is something we want in the product, and if so, what it should look like.

Hope to have an update and further guidance for you in the next few days.

fooman:
@tjwiebell I am all for moving every bit of functionality into separate extensions.  However as newsletter functionality is already included - see https://github.com/magento/pwa-studio/tree/develop/packages/venia-ui/lib/components/CommunicationsPage I would suggest not holding up this PR and instead create a separate task for moving everything newsletter related out to an extension.

schensley:
@shikhamis11 thanks for the work on this. I think your UI decisions are similar to my own. However in terms of desktop optimization I wonder if we might create a special form field with a textlink style action within the field, where cursors have greater accuracy, to tone down "Subscribe" in the footer, as it may appear "above the fold" in some instances. 
Then for mobile this would appear as a field and a button, preferably the secondary style, as the footer will likely never appear above the fold and the action (button) needs a larger touchable target.

At the very least I'd recommend using the secondary button, rather than the Primary Blue button. 

ContactUs_DT
ContactUs_Mobile

tjwiebell:
@shikhamis11 We can defer converting to an extension, since this doesn't cause any breakage and we are lacking GraphQL support to feature flag. Please see the feedback from @schensley about UX, and with those tiny tweaks I think we can send this to QA.

shikhamis11:
@tjwiebell, Thank you for the update. I will update the PR for the design changes and let you know here

schensley:
@shikhamis11 thank you for making those design edits. 
One last request. Since there is an odd number of items in the footer and "Subscribe" being the last, I'm wondering if in mobile size this section might span the space and the various elements assume centered justification to balance out the space?

shikhamis11:
do you also want me to make the info text `Recieve the latest news, update and special offers right to your inbox.` smaller in case of mobile view ??

schensley:
@shikhamis11 Thank you.  The text can/should match the rest of the footer text in terms of header and sub-text.  I don't think we reduce that for mobile.

shikhamis11:
@saumya1998 thank you for helping with design changes. 

@schensley Please review the design again

schensley:
@shikhamis11 Thank you for making the change to  mobile view. It looks great!

tjwiebell (approved):

JustinConabreeAbsolunet:
Hi @shikhamis11 ,

A couple of issues during QA. If you could please correct them it would be appreciated!

### Missing padding on desktop
Long email addresses get hidden behind the subscribe button. Adding padding to the right of the input on desktop should fix the issue.
image

### Accessibility regression
Missing a label for the input field
image

### Using Hn tag
It's preferable to use a non h2 tag for the "Subscribe to Venia". This is to prevent SEO penalization in Lighthouse from misorderd H tags. Changing to a span, as was done [here](https://github.com/magento/pwa-studio/pull/3412/files#diff-edf42881ef6457a6d71d3e33ada2c30ea30066bfba1a17e478a710af6077eaccR52), should do the trick (will probably need to add `display: block` to the css)

### Layout shift when submitting
There's a big layout shift of the footer when submitting which isn't great for UX. I think this also causes the form to empty when there's an error. It would be better to leave the form there but disabled while loading. I would suggest something similar to the snippet below, and adding `disabled={isBusy}` to the submit buttons
```
const maybeLoadingIndicator = isBusy ? (
            <LoadingIndicator global={true}>
                <FormattedMessage
                    id={'newsletter.loadingText'}
                    defaultMessage={'Subscribing'}
                />
            </LoadingIndicator>
) : null;

// .......

<Fragment>
    {maybeLoadingIndicator}
```

### Error handling
I don't think the error handling is working correctly. When I submit with an invalid email address, the GQL response has a specific error message that the email is invalid, but the form only displays: "An error has occurred. Please check the input and try again." which is the default form error.
image

Thanks!

shikhamis11:
@JustinConabreeAbsolunet, thanks for the update. I will check and update the changes

shikhamis11:
hi @JustinConabreeAbsolunet, 

For fixing the error handling issue I had to update the logic inside the `packages/peregrine/lib/util/deriveErrorMessage.js` file
https://github.com/magento/pwa-studio/pull/3262/commits/963483ef4f3a5fe0d03f04dea7e4e8a2d556ccf1#diff-555d016330512ae13aa6051d807f19cf047dae6ef2bcdfe1ea30d5349f717056
This will cause showing actual error at every form for example Sign-in/Create Account/Address etc

shikhamis11:
I have one more doubt, do I need to reset the newsletter form in case of `success` only or `error`, or don’t need to reset it after submission?

JustinConabreeAbsolunet:
Hey @shikhamis11 ,

Sorry for the delay, I was getting some feedback from the team. They decided to leave the generic error for now. If you could please revert what you had done to make the error message come through. So sorry about this! They're going to create another ticket to revise error handling throughout the application.

> I have one more doubt, do I need to reset the newsletter form in case of success only or error, or don’t need to reset it after submission?

Only in the case of a successful submit please! When there's an error, it will be useful for the user to see what they entered so they can correct it.

Also, I tested quickly and noticed that the padding on the input isn't enough. I think you may need to hardcode it to a specific pixel amount, since it needs to be the same amount as the "subscribe" button.

Thanks!

saumya1998:
Is there any issue coming with calculated padding which is used right now in newsletter input . can you please share the screenshot if there is any issue coming?


> Also, I tested quickly and noticed that the padding on the input isn't enough. I think you may need to hardcode it to a specific pixel amount, since it needs to be the same amount as the "subscribe" button.

JustinConabreeAbsolunet:
@saumya1998 You're right, all good now! I just had to clear my cache.

@shikhamis11 The rest of the issues are also resolved, QA passed! The branch is outdated and has a conflict. Since it's in your fork, can you please pull in the latest changes from develop and correct the conflict? We'll merge the PR once that's done

Additional notes (can ignore)
Cypress account and wishlist tests passed. MFTF has one error, but that test is being changed for a Cypress one so it's okay (`CustomerAccountAccessCest`).

JustinConabreeAbsolunet (approved):

JustinConabreeAbsolunet:
Hi @shikhamis11 I've created a new branch `shikhamis11/newsletter` and changed this to be the base. We're cutting the next PWA Studio release next week, and since we don't have any checks on whether or not the newsletter is enabled, we'll need to hold off on this feature being release. 

Merging now. Thanks for your work!

shikhamis11:
@JustinConabreeAbsolunet  Thanks for update.

PR #3278, opened 2021-07-16
Support new GraphQL Header

supernova-at:
## Description

This PR adds a new Apollo Link to our Apollo client that reads and attaches the new `X-Magento-Cache-Id` header to every GraphQL request.

## Related Issue

Closes PWA-1868.

### Verification Stakeholders

@jimbo @pdohogne-magento 

## Test Plan

You will want to point to a backend environment that sends the `x-magento-cache-id` header.

1. Load the homepage and wait for all requests to complete
2. Look for the `X-Magneto-Cache-Id` header of the last GraphQL request in the Network tab (response headers)
3. Perform an action that causes another GraphQL request (ex: click the shopping bag / cart icon)
4. Inspect the outgoing GraphQL request and verify that it has an `X-Magento-Cache-Id` request header and that the value matches the previous value sent by the server
5. Perform an operation that changes the cache id (ex: log in/out, change store, change currency, change address)
6. Verify that Venia properly updates and sends out its now-refreshed `X-Magento-Cache-Id` value

n/a

n/a

No

Test against environments that don't support the new header, perhaps.

## Checklist

### DIFF
--- a/packages/peregrine/lib/Apollo/__tests__/magentoGqlCacheLink.spec.js
+++ b/packages/peregrine/lib/Apollo/__tests__/magentoGqlCacheLink.spec.js
@@ -1,0 +1,123 @@
+import MagentoGQLCacheLink from '../magentoGqlCacheLink';
+import {
+    mockGetItem,
+    mockSetItem
+} from '@magento/peregrine/lib/util/simplePersistence';
+
+jest.mock('@magento/peregrine/lib/util/simplePersistence');
+
+test('it sets the correct context + headers on outgoing requests', () => {
+    // Arrange.
+    let setContextResult = null;
+    const operation = {
+        setContext: jest.fn().mockImplementation(implementation => {
+            const mockPreviousContext = {
+                not: 'headers',
+                headers: {
+                    something: 'else'
+                }
+            };
+            setContextResult = implementation(mockPreviousContext);
+        }),
+        getContext: jest.fn()
+    };
+    const forward = jest.fn().mockReturnValue([]);
+
+    const instance = new MagentoGQLCacheLink();
+    instance.setCacheId('unit_test');
+
+    // Act.
+    instance.request(operation, forward);
+
+    // Assert.
+    expect(operation.setContext).toHaveBeenCalled();
+
+    // It should retain the previous context.
+    expect(setContextResult.not).toEqual('headers');
+
+    // It should append to the headers.
+    expect(Object.keys(setContextResult.headers)).toHaveLength(2);
+    expect(setContextResult.headers['x-magento-cache-id']).toEqual('unit_test');
+    expect(setContextResult.headers.something).toEqual('else');
+});
+
+test('it tries to seed from local storage', () => {
+    // Arrange.
+    let setContextResult = null;
+    const operation = {
+        setContext: jest.fn().mockImplementation(implementation => {
+            const mockPreviousContext = {
+                not: 'headers',
+                headers: {
+                    something: 'else'
+                }
+            };
+            setContextResult = implementation(mockPreviousContext);
+        }),
+        getContext: jest.fn()
+    };
+    const forward = jest.fn().mockReturnValue([]);
+    mockGetItem.mockReturnValueOnce('unit_test_from_storage');
+
+    const instance = new MagentoGQLCacheLink();
+
+    // Act.
+    instance.request(operation, forward);
+
+    // Assert.
+    expect(setContextResult.headers['x-magento-cache-id']).toEqual(
+        'unit_test_from_storage'
+    );
+});
+
+test('it should save the cache id from responses', () => {
+    // Arrange.
+    const operation = {
+        setContext: jest.fn(),
+        getContext: jest.fn().mockImplementation(() => {
+            const headers = new Map();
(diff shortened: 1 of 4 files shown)


packages/peregrine/lib/Apollo/magentoGqlCacheLink.js
```diff
@@ -0,0 +1,39 @@
...
+
+            return {
+                ...previousContext,
+                headers: withCacheHeader
+            };
+        });
+
+        // Update the cache id with each response.
+        const updateCacheId = data => {
+            this.#cacheId = Math.random();
```
supernova-at, line 28: Actually read this from the response headers.

packages/peregrine/lib/Apollo/magentoGqlCacheLink.js
```diff
@@ -0,0 +1,39 @@
...
+
+            console.log('received', this.#cacheId);
```
supernova-at, line 30: Remove the `console.log` statements.

packages/peregrine/lib/Apollo/magentoGqlCacheLink.js
```diff
@@ -0,0 +1,39 @@
...
+    request(operation, forward) {
+        // Attach the cache header to each outgoing request.
+        operation.setContext(previousContext => {
+            const { headers } = previousContext;
+            const withCacheHeader = {
+                ...headers,
+                [CACHE_ID_HEADER]: this.#cacheId
+            };
+
+            console.log('sending out', this.#cacheId);
```
supernova-at, line 18: Remove the `console.log` statements.

packages/peregrine/lib/Apollo/magentoGqlCacheLink.js
```diff
@@ -0,0 +1,59 @@
...
+const storage = new BrowserPersistence();
+
+/**
+ * The Magento GraphQL Cache Link class is an ApolloLink that is responsible for
+ * Venia sending the proper `x-magento-cache-id` header with each of its GraphQL requests.
+ */
+export default class MagentoGQLCacheLink extends ApolloLink {
+    // The links get reinstantiated on refresh.
+    // If we have an existing cache id value from a previous browsing session, use it.
+    #cacheId = storage.getItem(LOCAL_STORAGE_KEY) || null;
```
pdohogne-magento, line 17: Question: When does this get instantiated to re-check local storage?

packages/peregrine/lib/Apollo/__tests__/magentoGqlCacheLink.spec.js
```diff
@@ -0,0 +1,123 @@
...
+
+    // Act.
+    instance.request(operation, forward);
+
+    // Assert.
+    expect(mockSetItem).toHaveBeenCalledTimes(1);
+    expect(mockSetItem).toHaveBeenCalledWith('magento_cache_id', 'unit_test');
+});
+
+test('it does not update its internal value when not present in the response', () => {
```
pdohogne-magento, line 100: Is there a reason for this?

fooman:
I have been working hard on our project to back out of what I consider a wrong direction; adding custom http headers to the graphQL requests. 

I haven't found the time yet to write this up as an issue report but the summary of it is that CORS will kick in (essentially doubling network activity) if a request is more than Simple (when multiple domains are involved).  Details [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS). This pull request seems to be doing more of the same when we should try to do the opposite (we should keep it a simple request if we can). If this is driven by requirements of fastly this should get implemented as a separate optional package.

In my view a better suited solution would be https://github.com/danslo/magento2-module-automatic-persisted-queries.

jimbo (approved):
@supernova-at Thanks Andy, this is exactly what I had in mind.

Older versions of PWA Studio will have to create and apply this link elsewhere, since `useAdapter` won't exist for them, but it should be easy to drop in.

pdohogne-magento:
@jimbo How old of a version?  We have a client on Magento 2.4.2 who needs this ASAP, but I don't know what version of PWA they're on

packages/peregrine/lib/Apollo/magentoGqlCacheLink.js
supernova-at: >When does this get instantiated to re-check local storage?

Every time the app is loaded: both on initial navigation to the site and on a browser refresh.

> And when does local storage get cleaned (if ever)?

I did not put any expiration date on this piece of data in local storage, it will stick around forever.

supernova-at:
>How old of a version? We have a client on Magento 2.4.2 who needs this ASAP, but I don't know what version of PWA they're on

@pdohogne-magento Version 11 will introduce `useAdapter`. This change will likely be released in an `11.1.0`. They'll just have to modify the `Adapter` component instead of `useAdapter`.

anthoula:
@supernova-at Please take a look at this new lighthouse warning. Wondering if it's relevant on my local instance. 
`Avoid enormous network payloads Total size was 3,185 KiB`

Screen Shot 2021-07-21 at 3 32 49 PM

anthoula:
✅   QA Approved

QA Verification Steps [PWA-1868](https://jira.corp.magento.com/browse/PWA-1868)

PR #3279, opened 2021-07-16
PWA-1198: Resetting option in edit cart panel and clicking on update button removes product from cart.

Rykh:
## Description
Fixed the issue with removing an item from the cart

## Steps
1. Add configurable product to cart
2. go to Cart page 
3. Go to Product Edit panel 
4. Select a new swatch color
5. reset (select) previously selected swatch color.
6. Hit update button.

## Expected 
Edit panel should close and product should exist. 

## Actual 
Edit panel is closed but product is getting removed.

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-1198.

- Run through steps, verify bug no longer exists
- Apply test change to `develop` and verify this test now correctly identifies this bug

## Checklist


supernova-at:
This makes too much sense 😅  - thank you!

larsroettig (dismissed):
:heavy_check_mark: LGTM

tjwiebell:
@Rykh - conceptually this is a great work around, but I think there's a bug in the talon that could solve this is a slightly more optimized way. Currently on option selection there's logic to select the variant, but [this line here](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/CartPage/ProductListing/EditModal/useProductForm.js#L119) is an anti-pattern, and is mutating that optionSelections `Map` when it shouldn't be.

Instead of mutating that Map state, we should create a copy and merge in selected values from the cart into our local state.

```js
const mergedOptionSelections = new Map([...optionSelections]);
cartItem.configurable_options.forEach(option => {
    if (!mergedOptionSelections.has(`${option.id}`)) {
        mergedOptionSelections.set(`${option.id}`, option.value_id);
    }
});

return findMatchingVariant({
    variants: configItem.variants,
    optionCodes: configurableOptionCodes,
    optionSelections: mergedOptionSelections
});
```

And then in the `handleSubmit` callback, only trigger `updateConfigurableOptions` - `if (selectedVariant && optionSelections.size)`. There is no use in deleting and re-adding an identical product, so this change in logic will just do nothing and close the modal.

Are you comfortable updating the PR with these suggestions? In addition, we will need to cover this with unit tests, so please let me know if you need any assistance with that as well.

revanth0212 (approved):
Approved

PR #3283, opened 2021-07-19
PWA-1593: Mega Menu re-renders position of items after initial render

Rykh:
## Description

**Acceptance Criteria**
The Mega Menu items stay stable across renders (they don't swap positions)

**Test Plan**
1. Navigate to the home page on desktop
2. Verify that the mega menu items don't switch positions
3. Refresh the page
4. Verify that the mega menu items don't switch positions

## Related Issue

Closes [http://jira.corp.magento.com/browse/PWA-1593](http://jira.corp.magento.com/browse/PWA-1593).

1. Navigate to the home page on desktop
2. Verify that the mega menu items don't switch positions
3. Refresh the page
4. Verify that the mega menu items don't switch positions

n/a

* Intentionally set colliding `position` values on other categories.

No, Mega Menu only appears on Desktop.

n/a

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js
+++ b/packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js
@@ -74,7 +74,9 @@ export const useMegaMenu = (props = {}) => {
             if (megaMenuCategory.children) {
                 megaMenuCategory.children = [...megaMenuCategory.children]
                     .filter(category => shouldRenderMegaMenuItem(category))
-                    .sort((a, b) => (a.position > b.position ? 1 : -1))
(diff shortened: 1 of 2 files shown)


tjwiebell:
@Rykh - thank you for the submission, but it unfortunately looks like we fabricated this bug with [our sample data](https://github.com/PMET-public/module-configurable-sample-data-venia/blob/master/fixtures/categories.csv#L2). There's not a way via the UI this would be reproducible, and we should be able to rely on position being unique. I'll submit a PR to the maintainers of our sample data to make sure these categories have a unique position value so this goes away, and in the meantime I'll manually resolve sample data issues on our testing environments.

Issue #3148, opened 2021-04-28
[feature]: Improved loading experience for Venia


sirugh:
**Purpose**

Determine if we can come up with an elegant way to "progressively" load page-level content.

Currently, we think we are stuck showing a full-page loading spinner until we get the results of the `urlResolver` query. After that, we can instrument the app to show page-level skeletons but they'll (usually) only be in view for a short time before actual page content appears. The combination of these two things results in a ton of layout shift for the user, as they navigate these three "stages" of the page loading.

Can we do better? What other options can we consider?

**Original Ask**

As a user I would prefer to see "shimmer loader" or an app skeleton that progressively hydrates rather than a loading spinner.

Don't worry about optimizing the display prior to receiving the response from the URL resolver.

One approach might be to render a skeleton specific to page type ie Category, Cart, PDP, etc.
 Another approach might be to have each component provide a skeleton view when loading (using nbsp, or "shimmer" components).

Jordaneisenburger:
Haven't thought about it to much but I think adding a skeleton/shimmer for each component makes more sense than a general shimmer for a whole page type. For these reasons:
- When you change the component layout you're less likely to forget you also need to update the skeleton that lives in the same folder.
- If you have components that conditionally render based on query data you're not showing a skeleton that might not even render at all.
- It's easier for 3rth party modules to supply out of the box skeletons for their components.

sirugh:
@magento export issue to JIRA project PWA as Story

fnhipster:
> Just for clarification; the demo url provided is not a pwa studio storefront right ?

That's correct.

PR #3284, opened 2021-07-19
[PWA-1878] Proof of Concept - Category Shimmer

JustinConabreeAbsolunet:
## Description
_Proof of Concept_
Adding Shimmer components while data is loading instead of a full page loader. The goal is to show a skeleton of the page, where the shimmer components take up about the same place the normal components with data would. This will improve the LCS score, and the user's perception of site speed. This Proof of Concept was done on the category page. It includes:
- Base component for all Shimmers
- Draft documentation for the Shimmer component and its use
- Category sub-component Shimmer components (which include one or more base Shimmer components)

### Short comings of this PoC
The full-screen loader is still necessary while determining the type of the page. This means that there are two loading phases which isn't ideal. SSR will correct this for the initial load. However additional strategies will need to be worked on for page-changes after the initial load.

## Related Issue
https://jira.corp.magento.com/browse/PWA-1878
https://jira.corp.magento.com/browse/PWA-1879

Relates to issue: https://github.com/magento/pwa-studio/issues/3148

### Verification Stakeholders

@tkacheva @davemacaulay @jimbo @schensley 

### Specification

Draft documentation for use available in pwa-devdocs/_drafts/venia-pwa-concept/component/shimmer/index.md

1. Visit category page
2. Once the page-type is determined, the full-screen-loader should disappear
3. The Shimmer components should then render in place of the filters, sort/filter buttons, product tiles, breadcrumbs
4. Validate that the LCS score from Lighthouse has maintained or improved

1. Validate that applying filters renders the product Item shimmer component while loading

Yes, validate that mobile shimmers appear correctly

## Breaking Changes (if any)

- Added exports to index.js files for shimmers
- Changed content of categoryContent root component so targetables may change.

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/FilterSidebar/filterSidebar.shimmer.js
+++ b/packages/venia-ui/lib/components/FilterSidebar/filterSidebar.shimmer.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import { shape, string } from 'prop-types';
+import { mergeClasses } from '../../classify';
+import Shimmer from '../Shimmer';
+import defaultClasses from './filterSidebar.css';
+
+const FilterSidebar = (props) => {
+    const classes = mergeClasses(defaultClasses, props.classes);
+
+    return (
+        <aside className={classes.root} aria-live="polite" aria-busy="true">
+            <Shimmer width="95%" height="70vh" style={{ marginBottom: 25 }} />
+        </aside>
+    )
+};
+
+FilterSidebar.defaultProps = {
+    classes: {}
+};
+
+FilterSidebar.propTypes = {
+    classes: shape({
+        root: string
+    })
+};
+
+export default FilterSidebar;
+
--- a/packages/venia-ui/lib/components/FilterSidebar/filterSidebar.shimmer.js
+++ b/packages/venia-ui/lib/components/FilterSidebar/filterSidebar.shimmer.js
@@ -1,21 +1,18 @@
 import React from 'react';
 import { shape, string } from 'prop-types';
-import { mergeClasses } from '../../classify';
+import { useStyle } from '../../classify';
+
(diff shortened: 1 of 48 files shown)


packages/venia-ui/lib/components/FilterSidebar/filterSidebar.shimmer.js
```diff
@@ -0,0 +1,24 @@
...
+
+import Shimmer from '../Shimmer';
+import defaultClasses from './filterSidebar.css';
+
+const FilterSidebar = props => {
+    const classes = useStyle(defaultClasses, props.classes);
+
+    return (
+        <aside className={classes.root} aria-live="polite" aria-busy="true">
+            <Shimmer width="95%" height="70vh" style={{ marginBottom: 25 }} />
```
larsroettig, line 13: Why `95%` I am not sure if use sizing like this

larsroettig (approved):
:heavy_check_mark: LGTM

packages/venia-ui/lib/components/FilterSidebar/filterSidebar.shimmer.js
JustinConabreeAbsolunet: @larsroettig It was to give a gap between the items and the sidebar, since padding will still have the background applied over that area. I can wrap the shimmer in an element with padding though if the `95%` is concerning for you!

eug123:
✅ Passed
- Test Plan (except lighthouse score)
- MFTF
- Cypress
- Alternate Browsers
🚫 Failed
- Validate that the LCS score from Lighthouse has maintained or improved 

Have run lighthouse several times and on average general performance score and LCS score got a bit worse compared with develop instance. There is also new notice appeared: `Avoid non-composited animations: 42 animated elements found`

PR #3288, opened 2021-07-21
[PWA-1898] Wishlist seems to be cramped in Mobile mode.

tjwiebell:
## Description

While performing regression on PWA 11.0 we have noticed that the wishlist section seems to be cramped in the mobile mode. This is reproducible on both physical devices (iOS and Android) and emulated device screens in chrome dev tools.

Steps:

Login
Add an item to wishlist
Go to wishlist page
Swtich to mobile mode

## Related Issue

* [[PWA-1898](https://jira.corp.magento.com/browse/PWA-1898)] Wishlist seems to be cramped in Mobile mode.

### Screenshots
Screen Shot 2021-07-21 at 2 40 58 PM

1. In the mobile viewport (<768px), log in and navigate to Wishlist page with items pre-populated in a list
2. Verify item count element is readable on its own row

## Checklist

### DIFF
--- a/packages/venia-ui/i18n/en_US.json
+++ b/packages/venia-ui/i18n/en_US.json
@@ -424,7 +424,5 @@
     "wishlistPage.wishlistDisabledMessage": "The wishlist is not currently available.",
     "wishlist.itemCountOpen": "Showing {currentCount} of {count} items in this list",
     "wishlist.itemCountClosed": "You have {count} {count, plural, one {item} other {items}} in this list",
-
-
     "wishlist.loadMore": "Load More"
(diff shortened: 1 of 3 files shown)


larsroettig (dismissed):
:heavy_check_mark: LGTM

revanth0212:
Fix looks straightforward. Almost ready to merge but before that a quick question.

I noticed that the count message is top-aligned but the wishlist and kabab menu is center-aligned. Is that something we need to worry about or am I just pixel peeping?

image

revanth0212 (approved):

PR #3289, opened 2021-07-21
Safari NumberFormat error fix.

revanth0212:
## Description

Safari browser has a bug related to the INTL NumberFormat polyfill. We are already handling it in the code but unfortunately, we are throwing an error as well. Since an error is thrown the whole app crashes on Safari instead of rendering the app with the handled logic.

This PR simply changes the function to not throw and warn instead to let the backup logic do its magic and render the app.

magento.pwa-venia.com
image

develop.pwa-venia.com
image

After Fix
image

## Related Issue

Closes PWA-1894

## Acceptance

App should render properly on Safari.

### Verification Stakeholders

@eug123 
@tjwiebell 

## Verification Steps

1. Open the deployed app in the safari browser. App should render properly.
2. Go to any category page and prices should render properly.
3. Switch between locales and currencies and nothing should be broken.

Yes, Browser testing is needed.

## Breaking Changes (if any)

None

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/CurrencySymbol/currencySymbol.js
+++ b/packages/venia-ui/lib/components/CurrencySymbol/currencySymbol.js
@@ -25,7 +25,7 @@ const isNarrowSymbolSupported = (() => {
         return true;
     } catch (e) {
         if (e.constructor !== RangeError) {
-            throw e;
+            console.warn(e);
         }
 
         return false;


tjwiebell (approved):
Good find, fix works perfectly 👍

supernova-at:
## QA

📝 Tested on using Safari Version 14.1.2 (14611.3.10.1.5) on macOS Mojave 10.14.6.

### To Do

- [x] Test Plan
- [ ] MFTF
- [ ] Cypress
- [ ] Lighthouse
- [ ] Alternate Browsers

### ✅  Passed

### 🚫  Failed

PR #3293, opened 2021-07-23
[FEATURE] Improve className format during development

Jordaneisenburger:
## Description
When in development finding where a component is located should be as easy as possible. When inspecting an element identifying a component by it's classname is a easy way to do so hence I improved the classname by adding the path to the file where the component is located.

## Related Issue

Closes PWA-1909.

## Verification Steps

- pull this pr
- start the watcher
- open the chrome devtools and notice the classnames now point to the actual file in a more explicit format
- close the watcher and start the stage version of the storefront, note that the classes are **not** pointing to the file structure

### Resolved issues:
1. [x] resolves magento/pwa-studio#3294: [FEATURE] Improve className format during development


larsroettig:
@magento create issue from PR

brendanfalkowski:
I like the utility of this, but I think in practice I would find it annoying because in my local dev environment it means the HTML tree as seen in the browser's dev tools will become really cluttered with (mostly) the same `/path/to/components/ComponentName` info on most nodes.

I think it's a better developer experience to keep the generated class name short with only the component name acting as the readable identifier. That's enough to search by filename in an editor because it's always a 1:1 match.

But I do like the direction of this, and I'm not sure if it's possible with Webpack and CSS Modules (probably) to generate CSS sourcemaps. This would be a real improvement for dev experience because (I think all) browser's dev tools read sourcemaps automatically if present.

Today when previewing applied CSS, we can only see `<style>` as the source on every line because of CSS Modules inlining, but I wonder if Webpack could generate browser-readable sourcemaps instead?

packages/pwa-buildpack/lib/WebpackTools/configureWebpack/getModuleRules.js
```diff
@@ -118,7 +118,10 @@ getModuleRules.css = async ({ hasFlag, mode, paths }) => ({
                 {
                     loader: 'css-loader',
                     options: {
-                        localIdentName: '[name]-[local]-[hash:base64:3]',
+                        localIdentName:
+                            mode === 'development'
```
revanth0212, line 122: That's a nice change but I am sure the cypress tests need to be updated because they depend on the classname's regex to identify an element while testing.

larsroettig: @revanth0212 maybe we should use data-attribute for cypress bind test on classNames is not recommend by cypress itself 
https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

revanth0212: We had this discussion with the team a while ago during the dev time and the team decided to use class names instead of data attributes because we dint want to pollute the app with data attributes just for cypress given cypress will mostly be used internally. 

We did have the thought of stripping off data attributes in prod mode but that will not let us run cypress against `develop`, 'release.*`, and `master` instances which we use in regression.

So we settled with class names and decided to use data attributes only if finding an element with class names is getting too complex.

Jordaneisenburger: **Just my 2 cents:**
I feel like classNames are to brittle, we should use test attributes and strip them in production. I understand your point @revanth0212 i've had those same issues but we should only strip them from the final product that is actually being deployed to customers, I assume this would need to changes from dev-ops.

Jordaneisenburger:
> I like the utility of this, but I think in practice I would find it annoying because in my local dev environment it means the HTML tree as seen in the browser's dev tools will become really cluttered with (mostly) the same `/packages/packageName/src/components/ComponentName` info on most nodes. As it is today:
> 
> Screen Shot 2021-07-23 at 10 00 34 AM
> 
> I think it's a better developer experience to keep the generated class name short (as above) with only the component name acting as the readable identifier. That's enough to search by filename in an editor because it's always a 1:1 match.
> 
> But I do like the direction of this, and I'm not sure if it's possible with Webpack and CSS Modules (probably) to generate CSS sourcemaps. This would be a real improvement for dev experience because (I think all) browser's dev tools read sourcemaps automatically if present.
> 
> Today when previewing applied CSS, we can only see `<style>` as the source on every line because of CSS Modules inlining, but I wonder if Webpack could generate browser-readable sourcemaps instead?
> 
> Screen Shot 2021-07-23 at 9 57 44 AM

@brendanfalkowski I'm pretty sure that if we have 2 components with the exact same name only the hash at the end would differ. (haven't fact checked it)

Regarding the CSS sourcemaps I totally agree, did a quick check and it should be possible https://github.com/webpack-contrib/css-loader#sourcemap

packages/pwa-buildpack/lib/WebpackTools/configureWebpack/getModuleRules.js
fooman: Having data-testid attributes in the final build is also not that bad beyond taking up a few chars they don't do much.  May even be nice to be able to run your smoketests automatically against the live site.

tjwiebell:
@Jordaneisenburger Explored the source map solution and it turned out way cooler than I anticipated. Take a look at #3407 and the attached demo video and let me know if that should solve the problem this PR was aiming to solve.

Jordaneisenburger:
> @Jordaneisenburger Explored the source map solution and it turned out way cooler than I anticipated. Take a look at #3407 and the attached demo video and let me know if that should solve the problem this PR was aiming to solve.

@tjwiebell thanks for looking into this! initially I wanted to just be able to quickly find the components in question. But this way might even be better since you can see the exact path when you hover the css filename. Personally I've never used devtool workspaces like that and I think not a lot of developers do or even know it works like this. 

So LGTM and since this could easily become a lost gem lets add this to a 'workflow/setup' section of the documentation. 

@brendanfalkowski @fooman you guys agree this issue can be closed in favor of Tommy's awesome pr ?

fooman:
Looks great to me. Still hopeful to see data-testid attributes in the future.

brendanfalkowski:
Good for me. I've never used workspaces for direct editing either, but I'm used to having sourcemaps from Gulp/Sass-era projects so that's great.

PR #3295, opened 2021-07-23
Waiting for button actions to complete mutations.

revanth0212:
## Description

Fix cypress button actions. This might not solve the issue if the network requests take even longer but it at least adds some time before mutations circle back to the browser.

AdHoc change.

## Related Issue

Closes [PWA-1914].

## Acceptance

Cypress tests should not fail.

### Verification Stakeholders

@supernova-at 

## Verification Steps

Run cypress tests, especially the wishlist ones. Nothing should fail.

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#3296: Waiting for button actions to complete mutations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,9 +115,9 @@ For example, filter options can be neatly collapsed and the product listing is a
 
 #### Pull Requests
 
-| Description                                                        | PR        |
-| ------------------------------------------------------------------ | --------- |
-| Optimized the layered navigation feature for the desktop           | [#3137][] |
+| Description                                              | PR        |
+| -------------------------------------------------------- | --------- |
+| Optimized the layered navigation feature for the desktop | [#3137][] |
 
 [#3137]: https://github.com/magento/pwa-studio/pull/3137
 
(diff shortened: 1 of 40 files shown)


supernova-at:
@magento create issue from PR

tjwiebell (changes requested):
Explicit waits like this are an anti-pattern in any testing framework, and should be avoided if possible. From my brief dive into Cypress, the better way to do this would be to alias the long running mutations and wait on those aliases. This is identical behavior to what we do in integration tests, but if you don't provide a mocked response, it's considered a spy of the actual request, and it will wait until it comes back.

I've had numerous runs where the createAccount mutation take awhile and tests fail immediately near the first step. With this little change in the [Cypress docs](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting), it's now behaving much better, regardless of what slowness might be occurring on the backend at the time of execution.

```js
cy.intercept('POST', '/graphql', req => {
    if (
        req.body.hasOwnProperty('query') &&
        req.body.query.includes('mutation')
    ) {
        req.alias = 'gqlMutation';
    }
});

cy.createAccount(
    accountAccessFixtures.firstName,
    lastName,
    accountEmail,
    accountPassword
);

cy.wait('@gqlMutation');

assertCreateAccount(firstName);
```

revanth0212:
> Explicit waits like this are an anti-pattern in any testing framework, and should be avoided if possible. From my brief dive into Cypress, the better way to do this would be to alias the long running mutations and wait on those aliases. This is identical behavior to what we do in integration tests, but if you don't provide a mocked response, it's considered a spy of the actual request, and it will wait until it comes back.

You are absolutely right @tjwiebell, explicit waits are not perfect. They solve the problem till the delay doesn't cross the wait. Also, it will be delaying the test if the network resolves immediately. 

> I've had numerous runs where the createAccount mutation take awhile and tests fail immediately near the first step. With this little change in the [Cypress docs](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting), it's now behaving much better, regardless of what slowness might be occurring on the backend at the time of execution.

We use this behavior in mocked Cypress tests in our codebase. Dint realize we can use this without mocking as well. Nice work, ill update the tests with this pattern soon. Thanks buddy.

mikhaelbois:
@tjwiebell @revanth0212 
I've updated the code with what was discussed by adding the utils suggested here: https://docs.cypress.io/guides/testing-strategies/working-with-graphql#Alias-multiple-queries-or-mutations
I had to use a timeout option since most of the mutations takes a certain amount of time to finish when the tests are executed.
Let me know what you think!

mikhaelbois:
PR can be closed in favor of https://github.com/magento/pwa-studio/pull/3343.

PR #3306, opened 2021-07-27
Release 11.0 Sign-out cache clear bug fix

revanth0212:
## Description

While performing the release regression of PWA 11.0 we figured a bug that comes up in a very weird edge case. If the TTYL of the token expires and the user tries to open/refresh the app, a side effect is causing the user to log out without clearing the apollo cache.

Because of this, the cart and customer details are leftover in the cache and UI enters a corrupted state. For instance, user has logged out but the wishlist details are left behind as shown below:

image

image

This PR fixes the issue by moving the side effect into a `useEffect` of the `UserContext` and signs the user out resetting the UI to a consistent initial state.

## Related Issue

Closes PWA-1888

## Acceptance

Cache and UI should reset to the initial state after signout or TTYL expiry.

### Verification Stakeholders

@jimbo 
@supernova-at 
@eug123 

## Verification Steps

To reproduce the error state:

1. Log in
2. Changed ttl of signing token to something small, for instance 5 (to make it expire faster)
3. Refresh the page after your expiry, in this case after 5 seconds
4. Check apollo cache, no customer or cart data should persist from before

Try the same on Develop or https://pr-3282.pwa-venia.com/ which is our release 11.0 instance.

## Breaking Changes (if any)

Need to figure out if this change breaks any old version. On top of my head, it doesnt.

## Checklist

### DIFF
--- a/packages/peregrine/lib/context/user.js
+++ b/packages/peregrine/lib/context/user.js
@@ -1,9 +1,10 @@
-import React, { createContext, useContext, useMemo } from 'react';
+import React, { createContext, useContext, useEffect, useMemo } from 'react';
 import { connect } from 'react-redux';
 
 import actions from '../store/actions/user/actions';
 import * as asyncActions from '../store/actions/user/asyncActions';
 import bindActionCreators from '../util/bindActionCreators';
+import BrowserPersistence from '../util/simplePersistence';
 
 const UserContext = createContext();
 
@@ -22,6 +23,22 @@ const UserContextProvider = props => {
         userApi,
         userState
     ]);
+
+    useEffect(() => {
+        // check if the user's token is not expired
+        const storage = new BrowserPersistence();
+        const item = storage.getRawItem('signin_token');
+
+        if (item) {
+            const { ttl, timeStored } = JSON.parse(item);
+            const now = Date.now();
+
+            // if the token's TTYL has expired, we need to sign out
+            if (ttl && now - timeStored > ttl * 1000) {
+                asyncActions.signOut();
+            }
+        }
+    }, [asyncActions]);
 
     return (
         <UserContext.Provider value={contextValue}>
--- a/packages/peregrine/lib/util/simplePersistence.js
+++ b/packages/peregrine/lib/util/simplePersistence.js
@@ -42,16 +42,12 @@ export default class BrowserPersistence {
         return this.storage.getItem(name);
     }
     getItem(name) {
-        const now = Date.now();
         const item = this.storage.getItem(name);
         if (!item) {
             return undefined;
         }
-        const { value, ttl, timeStored } = JSON.parse(item);
-        if (ttl && now - timeStored > ttl * 1000) {
-            this.storage.removeItem(name);
-            return undefined;
-        }
+        const { value } = JSON.parse(item);
+
         return JSON.parse(value);
     }
     setItem(name, value, ttl) {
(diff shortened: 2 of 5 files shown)


packages/peregrine/lib/util/simplePersistence.js
```diff
@@ -42,16 +42,12 @@ export default class BrowserPersistence {
...
         const item = this.storage.getItem(name);
         if (!item) {
             return undefined;
         }
-        const { value, ttl, timeStored } = JSON.parse(item);
-        if (ttl && now - timeStored > ttl * 1000) {
-            this.storage.removeItem(name);
-            return undefined;
-        }
+        const { value } = JSON.parse(item);
```
revanth0212, line 49: Instead of doing a side effect in the getter function, I moved it into a `useEffect` in the `UserContextProvider` which seemed the right spot for it.

packages/peregrine/lib/context/user.js
```diff
@@ -23,6 +24,22 @@ const UserContextProvider = props => {
         userState
     ]);
 
+    useEffect(() => {
```
jimbo, line 27: How often does this run? And does it always run in the necessary circumstance?

revanth0212: It only runs once on refresh and it runs on time as needed by the logic to work.

larsroettig (dismissed):
LGTM

packages/peregrine/lib/util/simplePersistence.js
eug123: What about other usages of this side effect like https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/Router/resolveUnknownRoute.js#L151 or in the possible customizations?

packages/peregrine/lib/context/user.js
supernova-at: I like this idea except for the refresh part, since that may not happen in an installed PWA. Can the token expire while the user is browsing around? If so, what happens then?

supernova-at: I actually hope this can _not_ happen since that would be a terrible user experience to be logged out in the middle of shopping 🙃

packages/peregrine/lib/util/simplePersistence.js
revanth0212: Good question @eug123. I don't think that they are related to side effects. The file you mentioned is not used in our code base. Looks like it's dead code.

eug123: Ok, I just found setting item with ttl, have not checked if this actually used. 
But it is still possible to add ttl - `setItem(name, value, ttl)`, so you will need to implement expiration logic by yourself, whenever you need it?

packages/peregrine/lib/context/user.js
supernova-at: Well it turns out that's exactly how it works

revanth0212: The use case we were trying to solve, includes a refresh or reopening of a closed tab. If the token expires mid-way while browsing, don't we have a refresh token or some mechanism to handle it?

packages/peregrine/lib/util/simplePersistence.js
revanth0212: We do have it, its part of User Context file.

packages/peregrine/lib/context/user.js
revanth0212: I tried the same without refresh and everything worked as expected. The app logged the user out and cleared cache..

larsroettig (approved):
LGTM

packages/peregrine/lib/context/user.js
supernova-at: Ah yes, there is something else for that. Nice!

JustinConabreeAbsolunet:
@revanth0212 QA passed

anthoula:
✅  MFTF Passed

Consistent test failure for `CartPageEditProductAndCheckout` test, but currently exists on release branch, and unrelated to this work.

PR #3308, opened 2021-07-29
PWA-1908 Improve loading perception while navigating between pages

JustinConabreeAbsolunet:
## Description

This PR aims to:
- Provide utilities to easily display full page shimmer effects while loading page information
- Provide a way to prefetch page information based on link URLs
- As a fallback, delay page transition until we know the next page type

To manage this, it provides:
- A new app context state value to manage the root component shimmer type (nextRootComponent)
- A new hook, useInternalLink, which returns a click function that will set the shimmer type when clicked
- A new file/folder structure for root component shimmers (venia-ui/lib/RootComponents/Shimmer/types/index.js)
- A new link component that wraps the default react-router-dom link and fetches page data when the link is visible
- Refactored page loading indicator to be a progress bar underneath the menu. The previous component has been moved to LoadingIndicator as `Spinner` it's more appropriate
- A new hook, useDelayedTransition, which takes care of blocking page transition while determining the next page's type

## Related Issue

https://jira.corp.magento.com/browse/PWA-1908

### Verification Stakeholders

@tkacheva , @jimbo, @davemacaulay 

TBD for link with prefetch (code available but not implemented)

**Root Component Shimmer**
- Set the network throttling to Fast 3G in the Network tab of Chrome devtools (so the shimmer doesn't disappear in under 1s)
- Click on either the Breadcrums (except home) or a category in the Main Navigation
- You should see a full skeleton of the category page. The page title (h1) should be a shimmer during this time. Once the page title displays, this is no longer the root component shimmer, but the shimmers inside of the normal categoryContent component

**Delayed page transition**
- Navigate to a category page
- Click on a product tile
- The loading bar should appear but the page should stay on the category for a brief moment (while the page type is being determined)
- The page should change to the product page (the normal full page loader may appear, but this is due to the root product component rendering the full page loader while it fetches data)
- When navigating to a route that is hardcoded (ex cart. search, checkout), there should be no delay
- When applying filters (or any other action that simply changes the query string) there should be no blocking

- Validate that the page always changes to the desired page
- Validate that the page transition deferral also works with error pages (doesn't revert to component displayed before error)
- Validate that 404 pages still work

Yes, device testing is needed as loading bar UI component may be impacted on mobile/desktop

## Breaking Changes (if any)

- Page loading indicator component refactored/moved to LoadingIndicator
- makeRoutes has changed to also export a list of available routes
- Any use of getUserConfirmation external of venia must be updated as we are now using this functionality

## Checklist

- [ ] ~~I have added translations for new strings, if necessary.~~


larsroettig (approved):
LGTM

packages/peregrine/lib/talons/Link/useLink.js
```diff
@@ -0,0 +1,68 @@
...
+    );
+    const linkPath = shouldPrefetch ? resourceUrl(to) : null;
+
+    useEffect(() => {
+        if (
+            !shouldPrefetch ||
+            pageTypeCalled ||
+            !runQuery ||
+            !elementRef.current ||
+            typeof intersectionObserver === 'undefined'
```
jimbo, line 30: Since you already perform this check in `useIntersectionObserver`, you can probably simplify.

```suggestion
            !intersectionObserver
```

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -13,6 +14,13 @@ export const useMagentoRoute = (props = {}) => {
     const { replace } = useHistory();
     const { pathname } = useLocation();
     const [componentMap, setComponentMap] = useRootComponents();
+    const [previousPathname, setPreviousPathname] = useState(null);
+    const [
+        { nextRootComponent },
+        {
+            actions: { setNextRootComponent, setPageLoading }
+        }
+    ] = useAppContext();
```
jimbo, line 23: In general, nested destructuring can be tough to follow. Consider a depth of 1 for readability.

```es6
const [appState, appActions] = useAppContext()
const { nextRootComponent } = appState
const { setNextRootComponent, setPageLoading } = appActions
```

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
```diff
@@ -13,6 +14,13 @@ export const useMagentoRoute = (props = {}) => {
     const { replace } = useHistory();
     const { pathname } = useLocation();
     const [componentMap, setComponentMap] = useRootComponents();
+    const [previousPathname, setPreviousPathname] = useState(null);
```
jimbo, line 17: Yes, this is probably the right place for this logic.

packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js
```diff
@@ -119,9 +120,17 @@ export const useMegaMenu = (props = {}) => {
...
+     *
+     * @returns {void}
+     */
+    const { handleClick: onNavigate } = useInternalLink('category');
+
     return {
         megaMenuData,
-        activeCategoryId
+        activeCategoryId,
+        onNavigate
```
jimbo, line 133: See the previous comment about naming conventions.

packages/peregrine/lib/talons/PageLoadingIndicator/usePageLoadingIndicator.js
```diff
@@ -0,0 +1,40 @@
...
+
+            doneTimeoutRef.current = setTimeout(() => {
+                setLoadingState(currentLoadingState => {
+                    if (currentLoadingState === 'loading') {
+                        return 'loading';
+                    }
+
+                    return 'off';
+                });
+            }, 750);
```
jimbo, line 30: If this is a magic number, we should at least extract it to a constant within this file, if not an environment variable or something.

jimbo (commented):
This approach looks good. I think the experience of going from one page to another feels better with these additions.

I made a few notes about code style as I was crawling through the code; there aren't any serious issues, but a fixup commit would be cool.

packages/peregrine/lib/hooks/useInternalLink.js
```diff
@@ -0,0 +1,22 @@
...
+        }
+    ] = useAppContext();
+    const type = `${rootType.toUpperCase()}${SHIMMER_TYPE_SUFFIX}`;
+
+    const handleClick = useCallback(() => {
+        setNextRootComponent(type);
+    }, [setNextRootComponent, type]);
+
+    return {
+        handleClick
```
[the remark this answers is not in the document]
JustinConabreeAbsolunet: Fixed!

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
JustinConabreeAbsolunet: Fixed!

packages/peregrine/lib/talons/PageLoadingIndicator/usePageLoadingIndicator.js
JustinConabreeAbsolunet: Good call! I changed it to a constant higher up in the file. Basically it serves to group together loadings that may occur right after eachother so we don't see a loading bar that bounces back and forth

packages/peregrine/lib/talons/MagentoRoute/useMagentoRoute.js
JustinConabreeAbsolunet: Nice!
There's actually a bug in this branch when navigating to routes not covered by useMagentoRoute (it holds on to the previous component). The fix for this is in [PWA-1204](https://github.com/magento/pwa-studio/pull/3353/files#diff-954aac140c96fefc0d3084918ee698a35f69001c61f320bfd641a4c3cc804d05R164), by setting it back to null when unmounting

eug123:
Verification:
:negative_squared_cross_mark: MFTF - not connected fails
:negative_squared_cross_mark: Cypress - not connected fails
:white_check_mark: Lighthouse
:white_check_mark: verification steps

Minor issues:
:negative_squared_cross_mark: Active search panel is overlapped by loading bar (the same is on mobile)
Bottoms Tops Dresses Accessories Shop The Look New Products
:negative_squared_cross_mark: When navigating to the Product Page from Category page parameter is added to the url `https://pr-3308.pwa-venia.com/carmina-earrings.html?page=1`

JustinConabreeAbsolunet:
@eug123 A refactor has been pushed for how I was delaying the page transition. The issue with the `?page=1` is now resolved, as well as the style issue with the search's autocomplete + loading bar.

This PR will need to pass code review again due to the changes to the delayed transition. Affected files from refactor:
- `useMagentoRoute` talon
- `MagentoRoute` component
- `useAdapter` talon
- `useDelayedTransition` new hook
- `useInternalLink` hook
- `makeRoutesTarget`
- `app` component

packages/peregrine/lib/hooks/useDelayedTransition.js
```diff
@@ -0,0 +1,125 @@
...
+    const operations = mergeOperations(DEFAULT_OPERATIONS);
+    const { resolveUrlQuery } = operations;
+    const [, setComponentMap] = useRootComponents();
+    const [, appApi] = useAppContext();
+    const { actions: appActions } = appApi;
+    const { setPageLoading } = appActions;
+    const unblock = useRef();
+
+    useEffect(() => {
+        // Override globalThis.addEventListener to prevent binding beforeunload while we add our blocker
```
jimbo, line 28: Can you explain more about this situation? I don't love the idea of overriding `addEventListener`, so I'd like to understand what the issue is and why this solves it.

JustinConabreeAbsolunet: @jimbo In react-router's `history.block`, they add a new event listener for the beforeunload to pop the default window.confirm to block navigation. This would mean that when you reload the page or navigate externally we'd get the message. With this we're blocking the event binding just for our block.

JustinConabreeAbsolunet: @jimbo https://github.com/remix-run/history/blob/main/docs/blocking-transitions.md#caveats

jimbo (approved):

packages/peregrine/lib/hooks/useDelayedTransition.js
jimbo: Cool, thanks for clarifying.

packages/venia-ui/lib/components/Link/link.js
```diff
@@ -0,0 +1,28 @@
...
+
+    return <RouterLink {...propsForBase} innerRef={ref} />;
+};
+
+Link.defaultProps = {
+    prefetchType: false
+};
+
+Link.propTypes = {
+    prefetchType: bool
```
luke-denton-aligent, line 25: I don't quite follow this prop. The name doesn't indicate that the type is going to be a boolean. I feel like this will be a source of a fair bit of confusion for devs. In `useLink` the destructured prop gets renamed to something that _does_ indicate a boolean type, but a developer won't see that, the exposed API for this component indicates `prefetchType`

luke-denton-aligent:
This looks like a great improvement! Did we consider checking a users connection before pre-fetching data? Using the Network Information API? I know it's experimental, but I think users who do set reduced data use on their device would be very grateful

luke-denton-aligent:
@jimbo @larsroettig Keen to get your thoughts on the above questions?

jimbo:
> @jimbo @larsroettig Keen to get your thoughts on the above questions?

Hi Luke, I realize this isn't a very timely reply but you've raised two good points.

1. Yes, the `prefetchType` prop would be less ambiguous if named `shouldPrefetch`—as we destructure it in the talon, as you pointed out.
2. Yes, it might be good to check the user's connection before prefetching. That said, the payload is small, and prefetching is set to `false` by default, so I don't think it's being misused at the moment. Also, my impression is that slow connections benefit most from prefetching the page type (at least for the most valuable destinations) as they're more likely to encounter the kind of wait time that leads a user to bounce.

I'll create a PR for the first change. While I think we can do without the second change, I would welcome a PR if you still think there's unique value the Network Information API can provide.

PR #3319, opened 2021-08-05
PWA-1278: a11y mega menu

pankhurigoel:
## Description

Verify that a user is able to navigate through the mega-menu using the keyboard.

## Related Issue

Closes [PWA-1278](https://jira.corp.magento.com/browse/PWA-1278): a11y for mega menu

Closes #ISSUE_NUMBER.

## Acceptance

User is able to navigate through the mega-menu using the keyboard.

## Verification Steps

- Navigate to the home page
- Press `Tab` unless you land on the mega-menu
- To open sub-menu press `down arrow`.
- To close sub-menu press `up arrow` or `esc` key.
- To navigate to that category click `Enter` key.

## Checklist


packages/peregrine/lib/talons/MegaMenu/useSubMenu.js
```diff
@@ -0,0 +1,26 @@
...
+        }
+    });
+
+    const isSubMenuActive = useMemo(() => {
+        if (isFocused) {
+            if (subMenuState) {
+                return true;
+            }
+        }
+        return false;
```
larsroettig, line 22:
```suggestion
       return isFocused && subMenuState;
```

Screenshot from 2021-08-20 11-28-11

larsroettig (changes requested):
Please fix my comments and the falling test 

`packages/venia-ui/lib/components/MegaMenu/__tests__/MegaMenuItem.spec.js`

packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js
```diff
@@ -119,9 +120,48 @@ export const useMegaMenu = (props = {}) => {
         }
     }, [findActiveCategory, location.pathname, megaMenuData]);
 
+    const useOutsideAlerter = ref => {
```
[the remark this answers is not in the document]
pankhurigoel: @supernova-at can you suggest a good name here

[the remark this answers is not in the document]
revanth0212: I am not sure if this is the right approach @pankhurigoel. We don't usually have talons inside of talons. You need to create a different file for it.

And for name, I think you are trying to handle blur on submenu. So maybe name it `useSubmenuBlurHandler`. I am open to opens for others on this.

revanth0212 (dismissed):
I havent reviewed the whole PR but this seems odd to me. We shouldn't be defining talons inside of talons.

packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js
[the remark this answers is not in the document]
jimbo: @revanth0212 is correct: returning a hook that the `MegaMenu` must call is not the correct approach. Instead, the `useMegaMenu` talon should perform the addition and removal of event listeners; if it needs a `navRef` to do so, it should get one from `props`.

```es6
// if this talon needs an external ref, get it from props
const { navRef } = props

// don't list refs in arrays of dependencies
const handleClickOutside = useCallback(event => {
    const { current: element } = navRef

    if (element && !element.contains(event.target)) {
        setDisableFocus(true)
        setSubMenuState(false)
    }
}, [setDisableFocus, setSubMenuState])

// import this utility hook from `peregrine/lib/hooks`
useEventListener(globalThis, "mousedown", handleClickOutside)
useEventListener(globalThis, "mouseout", handleClickOutside)
useEventListener(globalThis, "keydown", handleClickOutside)

// none of the above code needs to be returned from this talon
```

jimbo (dismissed):

packages/peregrine/lib/talons/MegaMenu/useMegaMenuItem.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+    };
+
+    return {
+        isFocused,
+        isActive,
+        handleCloseSubMenu,
+        isMenuActive,
+        a11yClick,
+        toggleSubMenu
```
tjwiebell, line 64: Can we combine these two into a callback called `handleKeyDown`?

packages/peregrine/lib/talons/MegaMenu/useSubMenu.js
```diff
@@ -0,0 +1,23 @@
...
+export const useSubMenu = props => {
+    const { isFocused, subMenuState, handleCloseSubMenu } = props;
+
+    const KEY_SHIFT = 9;
+
+    const { keyboardProps } = useKeyboard({
+        onKeyDown: e => {
+            //checking for Tab without Shift
+            if (e.keyCode == KEY_SHIFT && !e.shiftKey) {
+                e.target.addEventListener('blur', handleCloseSubMenu());
```
tjwiebell, line 13:
```suggestion
                e.target.addEventListener('blur', handleCloseSubMenu);
```

packages/venia-ui/lib/components/MegaMenu/megaMenuItem.js
```diff
@@ -14,26 +19,76 @@ import PropTypes from 'prop-types';
...
+                        size={17}
+                        aria-label={
+                            'Category: ' +
+                            category.name +
+                            '. ' +
+                            category.children.length +
+                            ' sub-categories'
+                        }
+                    />
+                ) : null}
```
tjwiebell, line 91: Consider extracting this to an element variable `moreChildrenIcon` or `downArrowIcon`. Some on the team like to indicate nullability by prefixing with `maybe<elementName>`, up to you just please extract.

packages/venia-ui/lib/components/MegaMenu/megaMenuItem.css
```diff
@@ -18,6 +18,15 @@
     display: flex;
 }
 
+.megaMenuItem_active {
+    composes: megaMenuItem;
+    display: flex;
+}
+
+.arrowDown {
+    padding-left: 8px;
```
tjwiebell, line 27: Prefer rem units here
```suggestion
    padding-left: 0.5rem;
```

packages/venia-ui/lib/components/MegaMenu/megaMenuItem.js
```diff
@@ -14,26 +19,76 @@ import PropTypes from 'prop-types';
...
                     isActive ? classes.megaMenuLinkActive : classes.megaMenuLink
                 }
                 to={categoryUrl}
             >
                 {category.name}
+                {category.children.length ? (
+                    <Icon
+                        className={classes.arrowDown}
+                        src={ArrowDown}
+                        size={17}
```
tjwiebell, line 82: Any reason for the extra pixel? I would expect this to match the font-size of the menu item, which is defaulting to 1rem (16px).

packages/venia-ui/lib/components/MegaMenu/submenu.js
```diff
@@ -11,16 +14,52 @@ import PropTypes from 'prop-types';
...
+        } else {
+            return (
+                <SubmenuColumn
+                    index={index}
+                    keyboardProps={{}}
+                    key={category.id}
+                    category={category}
+                />
+            );
+        }
```
tjwiebell, line 58:
```suggestion
            const keyboardProps = index === items.length - 1 ? talonProps.keyboardProps : {};
            return (
                <SubmenuColumn
                    index={index}
                    keyboardProps={keyboardProps}
                    key={category.id}
                    category={category}
                />
            );
```

tjwiebell (changes requested):
Some suggestions to be more consistent with the rest of the code base. This functions perfectly though, really nice work 👍

packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js
```diff
@@ -106,6 +111,25 @@ export const useMegaMenu = (props = {}) => {
         [isActive]
     );
 
+    const handleClickOutside = e => {
+        if (
+            props.mainNavRef &&
+            props.mainNavRef.current &&
```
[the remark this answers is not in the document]
pankhurigoel: Removing these lines is giving us errors as `handleClickOutside` function is called on mouse/keyboard events and the ref is attached to nav.

packages/venia-ui/lib/components/MegaMenu/megaMenuItem.js
```diff
@@ -63,32 +62,34 @@ const MegaMenuItem = props => {
...
+            }
+        />
+    ) : null;
+
     return (
         <div className={megaMenuItemClassname}>
             <Link
                 onKeyDown={e => {
-                    a11yClick(e) && toggleSubMenu(e);
+                    handleKeyDown(e);
```
tjwiebell, line 84: You shouldn't need to do your own binding here unless you're passing additional parameters. This can be:
`onKeyDown={handleKeyDown}`

tjwiebell (changes requested):
Really close, a few minor tweaks and we can send this to QA.

packages/peregrine/lib/talons/MegaMenu/useMegaMenu.js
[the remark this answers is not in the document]
tjwiebell: I figured this one out, and I'm not sure it was working as expected. You can remove these two lines but you need to fix `megaMenu.js` and make sure you're passing your props correctly. Should instead be `const talonProps = useMegaMenu({ mainNavRef });`.

packages/peregrine/lib/talons/MegaMenu/useMegaMenuItem.js
```diff
@@ -47,20 +41,23 @@ export const useMegaMenuItem = props => {
...
-            !(e.keyCode === KEY_UP || e.keyCode === KEY_ESCAPE)
+            !(e.key === 'ArrowUp' || e.key === 'Escape')
         ) {
             setIsFocused(true);
         } else {
             setIsFocused(false);
         }
     };
 
+    const handleKeyDown = e => {
```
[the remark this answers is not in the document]
pankhurigoel: We cannot do this because one function is preventing the default and if we just have one function then native `Tab` won’t work.

tjwiebell (approved):
All feedback has been addressed, just need an approval of my commit 08332ac before delivery. Sending to QA.

packages/venia-ui/lib/components/MegaMenu/megaMenuItem.js
tjwiebell: Fixed in 08332ac

packages/peregrine/lib/talons/MegaMenu/useMegaMenuItem.js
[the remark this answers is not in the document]
tjwiebell: You can see how I merged these routines and simplified the logic a bit in 08332ac.

larsroettig (approved):

PR #3329, opened 2021-08-06
[PWA-1962] Force resolution of Workbox to v6.0.2

mikhaelbois:
## Description

Workbox 6.2.0 was released yesterday, and our workbox-webpack-plugin dependency has a minor constraint on workbox-build: ^6.0.2, so it’s now picking up workbox-build@6.2.0 which is not backward compatible with 6.0.2.

This causes `yarn build` to fail in scaffolded projects, which our CI picked up.

This is a temporary fix, a better solution will be found in the future.

## Related Issue

Related to https://jira.corp.magento.com/browse/PWA-1962

### Verification Stakeholders

@tjwiebell 

`yarn build` should no longer fail in scaffolding.

### DIFF
--- a/packages/create-pwa/package.json
+++ b/packages/create-pwa/package.json
@@ -29,7 +29,7 @@
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/create-pwa#readme",
   "dependencies": {
-    "@magento/pwa-buildpack": "10.0.0-alpha.6",
+    "@magento/pwa-buildpack": "10.0.0-alpha.5",
     "@magento/venia-concept": "11.0.0-alpha.5",
     "chalk": "^2.4.2",
     "change-case": "^3.1.0",
--- a/packages/create-pwa/package.json
+++ b/packages/create-pwa/package.json
@@ -29,7 +29,7 @@
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/create-pwa#readme",
   "dependencies": {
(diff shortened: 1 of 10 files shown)


jcalcaben:
@mikhaelbois I think we also need to update: https://github.com/magento/pwa-studio/blob/develop/packages/venia-concept/_buildpack/create.js and add "resolutions" to the list of items to copy from the package.json.

jcalcaben (dismissed):
Code changes look good

Tested locally by running the scaffolding command and checking the `workbox-build` version with `yarn why workbox-build` on the project

packages/create-pwa/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/create-pwa#readme",
   "dependencies": {
-    "@magento/pwa-buildpack": "10.0.0-alpha.5",
+    "@magento/pwa-buildpack": "10.0.0-alpha.6",
```
tjwiebell, line 32: Our version bumps are automated in AWS, and I'm not sure if the jobs would break with these manual bumps. You'll likely need to undo all these, but was there a reason this was necessary?

mikhaelbois: Force of habit of updating versions when we apply a modification. I'll remove it now.

jcalcaben (approved):

anthoula:
QA Approved 👍

PR #3342, opened 2021-08-11
[PWA-1906] Prefetching and inlining data with UPWARD-PHP

JustinConabreeAbsolunet:
## Description

- Changes to the app template file and upward.yml to inline crucial data about the page (page type information, root component scripts) on page load.
- Change to webpack config to strip away inline mustache code during yarn watch
- Changing service worker strategy for html pages (NetworkFirst)
- Refactoring useMagentoRoute to use inlined data on first load, and run query on next navigations
- Rendering null during wait time when we have inlined page data to prevent content layout shift
- Adding shimmers for head switchers to prevent layout shift while loading

Note this PR depends on two UPWARD PHP pull requests
- https://github.com/magento/magento2-upward-connector/pull/34
- https://github.com/magento/upward-php/pull/86

To see the full effects of this PR, it should be tested on a Magento Cloud instance with UPWARD php modules up to date with above PRs. It should also be tested with UPWARD JS to ensure no regression.

## Related Issue

Closes [PWA-1906](https://jira.corp.magento.com/browse/PWA-1906)

### Specification

UPWARD-PHP maintainers should review UPWARD PHP pull requests

**UPWARD PHP Backend**
1. Load a category page (initial load, not navigate)
2. The page should have loaded without any loading spinner and minimal upfront delay
3. The page should transition smoothly to the category shimmer (and then to actual content)
4. Navigate to homepage
5. Loading bar should appear and then switch to homepage. There should be a delay in the transition (while it loads the page type)
6. Validate the page source. There should be a script tag that contains some of the page information in a string. There should also be the root component scripts in the source
7. Load home page
8. Revalidate page source to ensure page information reflects that of a CMS_PAGE

**UPWARD JS Backend**
1. Load a category page (initial load, not navigate)
2. The page should have loaded **with** a loading spinner
4. Navigate to homepage
5. Loading bar should appear and then switch to homepage. There should be a delay in the transition (while it loads the page type)
6. Validate the page source. There should be a script tag that contains `JSON.parse` with an empty string. There should be no root component scripts and no invalid scripts (ex no `/{{.}}` source

**Local Watch**
1. Run `yarn watch:all`
2. Build should run without error
3. Load the store front
4. Inspect the source to ensure there is no script with JSON.parse, nor any mustache code, such as `{{#webpackChunks}}`

**General**
1. Load store front
2. Navigate between pages and different page types
3. When clicking on a category in the menu or breadcrumb, the transition should be instantanious into a root component shimmer (full page skeleton)
4. When navigating to a product page or home page, the loading bar should appear, the transition should be delayed. Then the root component should appear and load data (possibly with a loading spinner, since shimmers have not been implemented for products or CMS)

- Verify account pages, cart, and checkout, to ensure loading page works correctly

- Test a page that redirects

No, functionality should not vary per device/size

## Breaking Changes (if any)

- Response from useMagentoRoute has changed for loading state. Wraps should take account for this

## Checklist

- [ ] ~~I have added translations for new strings, if necessary.~~


JustinConabreeAbsolunet:
Will create PR for PWA-1204 which is more advanced

PR #3344, opened 2021-08-11
[PWA-1974] Fix search suggested category double store code

JustinConabreeAbsolunet:
## Description

Removing createHref from useSuggestedCategory since the URL is used in a Link component which also adds the basename.

## Related Issue

Closes [PWA-1974](https://jira.corp.magento.com/browse/PWA-1974)

### Verification Stakeholders

@revanth0212 @supernova-at 

- Enable store code in URL
- Write "tanks" in search bar
- Select one of the categories (ex "Tanks in Blouses")
- Page should render without error (no Not Found page)

No

Current behaviour is that it copies the existing query string values. Could cause issues with "page" values out of range.

## Breaking Changes (if any)

None

## Checklist

- [ ]  ~~I have added translations for new strings, if necessary.~~
- [ ]  ~~I have updated the documentation accordingly, if necessary.~~


eug123:
Hi @JustinConabreeAbsolunet, fix looks good, but there is a tiny decrease in coverage. Could you please check if it possible to cover this talon with unit test, maybe this [example](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/talons/Header/__tests__/useStoreSwitcher.spec.js) could be relevant.
Than you!

JustinConabreeAbsolunet:
@eug123 Test coverage added for useSuggestedCategory! 👍

eug123 (approved):
Looks good

PR #3348, opened 2021-08-17
Background repeat not working on safari

### LINKED ISSUE #3347: [bug]: Background repeat not working on safari
**Describe the bug**

When creating a banner / column / row with a background, the background-repeat property is automatically set to true on safari

**To reproduce**

1. Create a CMS page
2. Insert a banner with an image with low width
3. Set the Background Repeat property to no
4. Save page
5. See the result in front on Safari

**Expected behavior**

The background-image should not repeat

**Possible solutions**

The problem is when creating props in parseStorageHtml.js with createContentTypeObject function, the property value returned with Safari is 'no-repeat no-repeat'. Then this value is tested in the getBackgroundImages function in utils.js like this: backgroundRepeat: node.style.backgroundRepeat !== 'no-repeat'. Which returns true because the current value is 'no-repeat no-repeat'.

First solution:

1. Change node.style.backgroundRepeat !== 'no-repeat' test in utils.js to just node.style.backgroundRepeat
2. Then change backgroundRepeat ? 'repeat' : 'no-repeat'; to backgroundRepeat; in banner.js, column.js, row.js; tabItem.js
3. Change propTypes and propTypes comments value in these same files

Second solution:

1. Change node.style.backgroundRepeat !== 'no-repeat' to node.style.backgroundRepeat !== 'no-repeat' && node.style.backgroundRepeat !== 'no-repeat no-repeat' in utils.js

-   Device : iPhone8, Mac and iPad Pro]:
-   Browser : Safari
-   Browser Version : Version 14.0.3 (16610.4.3.1.7)
-   Magento Version : Magento Commerce 2.4.2-p1

-   [ ] `venia-concept`
-   [ ] `venia-ui`
-   [ ] `pwa-buildpack`
-   [ ] `peregrine`
-   [x] `pagebuilder`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

antoine-fontaine:
## Description

When creating a banner / column / row / tabItem with a background, the background-repeat property is automatically set to true on safari

1. Change node.style.backgroundRepeat !== 'no-repeat' test in utils.js to just node.style.backgroundRepeat
2. Then change backgroundRepeat ? 'repeat' : 'no-repeat'; to backgroundRepeat; in banner.js, column.js and row.js
3. Change propTypes and propTypes comments value in the same files

## Related Issue

Closes #3347.

## Verification Steps

1. Create a CMS page
2. Insert a banner with an image with low width
3. Set the Background Repeat property to no
4. Save page
5. See the result in front on Safari, check if the background is not repeated

-   Device : iPhone8, Mac and iPad Pro]:
-   Browser : Safari
-   Browser Version : Version 14.0.3 (16610.4.3.1.7)


supernova-at:
hi @antoine-fontaine , thanks for your contribution!

I have moved this into the code review queue and someone from our team should look at it soon.
In the meantime, please address the issues from [PWAStudioBot](https://github.com/magento/pwa-studio/pull/3348#issuecomment-900307740).

If you need help with anything, just leave a comment here.

Thanks again!

supernova-at:
Assigned the `Major` version label because the API of the PageBuilder components changed (from expecting a `boolean` to `string`).

Strings were likely always already passed, but if someone was creating PageBuilder components by hand then the API has definitely changed on them.

supernova-at:
Code looks good but I have to run the verification steps and the [PWAStudioBot warnings](https://github.com/magento/pwa-studio/pull/3348#issuecomment-900307740) must be fixed before this can move on to QA.

packages/pagebuilder/lib/utils.js
```diff
@@ -12,7 +12,7 @@ export function getBackgroundImages(node) {
         backgroundSize: node.style.backgroundSize,
         backgroundPosition: node.style.backgroundPosition,
         backgroundAttachment: node.style.backgroundAttachment,
-        backgroundRepeat: node.style.backgroundRepeat !== 'no-repeat'
+        backgroundRepeat: node.style.backgroundRepeat
```
supernova-at, line 15: Pointing out this one specifically used to make a determination of whether to repeat or not when there was no data available to make the decision. In other words, it used to default to `backgroundRepeat: true`, which - after these changes - would correspond to `backgroundRepeat: 'repeat'`.

Instead, the result of this change is `backgroundRepeat` gets set to an empty string.

supernova-at:
There were a couple places that removed a block like this:

```js
        wrapperStyles.backgroundRepeat = backgroundRepeat
            ? 'repeat'
            : 'no-repeat';
```

and changed it to 

```js
wrapperStyles.backgroundRepeat = backgroundRepeat
```

Because this prop has changed to a `string` from a `boolean` there were cases where `wrapperStyles.backgroundRepeat` was being set to an empty string (especially in the tests). Because this is not a valid value for this CSS property, I defaulted the `backgroundRepeat` prop to `repeat` where applicable, which matches the browser default value for this property.

In the case where an empty string was passed in, previously `wrapperStyles.backgroundRepeat` would be set to 'no-repeat', and now it is set to 'repeat'. This is different from before but I think the correct default going forward.

anthoula (approved):

zetlen (approved):

zetlen:
Danger failure is happening for all PRs and is unrelated to this one. LGTM!

supernova-at:
I verified on my local that a newly scaffolded project using this code does pass `yarn build`. The CI failure may be related to the default values for scaffolding prompts, as PageBuilder requires an Adobe Commerce backend, not a Magento Open Source one.

PR #3349, opened 2021-08-17
PWA-1550: PB verify video

michaelyu0123:
## Description

To verify that the Page Builder video content type renders correctly on Venia for each content type property.
Only alignment is tested via Cypress automation since functionality wise it uses cross origin iframe videos which is not feasible for automation (nor needed really so if we were to test it it should be done manually).

## Related Issue

Closes [PWA-1550](https://jira.corp.magento.com/browse/PWA-1550): Verify Video

### Verification Steps

1. Cypress `video.spec.js` should pass
Steps - 
Go to venia-integration-test dir
Run `yarn install`
Run `yarn test`
Run `video.spec.js` from cypress ui

## Checklist


pankhurigoel (approved):
Looks good to me

eug123:
QA passed

PR #3354, opened 2021-08-18
Adding host headers to the GQL calls from buildpack.

revanth0212:
## Description

Adding host headers to the GQL calls from buildpack to resolve those requests.

There is another CICD PR we need to merge along with this: https://github.com/magento-commerce/pwa-studio-cicd/pull/78

## Related Issue

Closes PWA-1987

## Acceptance

Scaffolding job should not fail in the CI.

### Verification Stakeholders

@sharkySharks 

## Verification Steps

1. Check the job once this is merged for the job to consider using new code.
Meanwhile, if you need to test this code, re-run this [new temporary job that is a replica of the original scaffolding job](https://console.aws.amazon.com/codesuite/codebuild/913454440186/projects/pwa-scaffold-build-test/history?region=us-east-1&builds-meta=eyJmIjp7InRleHQiOiIifSwicyI6e30sIm4iOjIwLCJpIjowfQ#).

## Checklist


sharkySharks:
LGTM - maybe update the pr title and the branch since it seems to be out of date wit the base. I don't see any `TODO` anywhere that the Danger check is barking about.

packages/pwa-buildpack/lib/Utilities/graphQL.js
```diff
@@ -9,22 +11,35 @@ const fetchQuery = query => {
...
+        Accept: 'application/json',
+        'User-Agent': 'pwa-buildpack',
+        Host: targetURL.host
     };
 
     if (process.env.STORE_VIEW_CODE) {
         headers['store'] = process.env.STORE_VIEW_CODE;
     }
 
+    debug('Fetching query: %s', query);
```
sharkySharks, line 24: i think let's remove the debug statements now, eh? and just keep the headers.

packages/pwa-buildpack/lib/Utilities/graphQL.js
```diff
@@ -1,3 +1,5 @@
+const debug = require('../util/debug').makeFileLogger(__filename);
```
sharkySharks, line 1: do we want to keep all these debug statements in this file too?

packages/pwa-buildpack/lib/Utilities/graphQL.js
sharkySharks: this one?

packages/pwa-buildpack/lib/Utilities/graphQL.js
```diff
@@ -9,22 +11,35 @@ const fetchQuery = query => {
...
+    debug('Fetching query: %s', query);
+
     return fetch(targetURL.toString(), {
         agent: targetURL.protocol === 'https:' ? httpsAgent : null,
         body: JSON.stringify({ query }),
         headers: headers,
         method: 'POST'
     })
-        .then(result => result.json())
+        .then(result => {
```
sharkySharks, line 32: can probably just leave this as the above `.then(result => result.json())`

packages/pwa-buildpack/lib/Utilities/graphQL.js
revanth0212: These debug statements only run if the user whats them to. They are not console logs for instance. We usually have these debug lines in buildpack files to debug them when needed. For instance in this case I want to debug this file ill run the file with `DEBUG="*pwa-buildpack/lib/Utilities/graphQL*"`.

packages/pwa-buildpack/lib/Utilities/graphQL.js
revanth0212: It's a file logger object definition.

sharkySharks (approved):
lgtm

PR #3364, opened 2021-08-24
[PWA-892]: [bug] - Fix region input value not resetting after country change 

pedrochiossi:
## Description
The error when editing addresses occurs because we are not resetting the region input value properly. When we have saved addresses on Cart and we try to change countries, the useRegion talon fires a reset method on the region input field using informed useFieldApi hooks, but this method is not clearing any values, it is returning to the initial value, which is whatever is saved on Cart. This PR solves this problem by using the setValue method to set empty values on the region input after every country change.

## Related Issue

Closes PWA-892

## Acceptance

When editing addresses, users should be able to switch between UK and US without error.

## Verification Steps

1. Add product to cart and go to checkout page.
2. Add Shipping address, select shipping method and enter payment methods.
3. Uncheck billing same as shipping check box and enter any UK address.
4. Click Review order button
5. Now try to edit billing address country from UK to USA and vice-versa.
6. State input should be cleared
6. Click on update.
7. Form should show an input warning for required value on empty state.
8. Select a state from dropdown list or if not available type in a state.
9. Click on update.
10. It should update the address without any errors.

1. Perform repro steps from description above. Ensure error does not occur.
2. Attempt this same switch (UK <-> US) everywhere Venia collects an address. Ensure that all instances are fixed.
3. Attempt this same switch with a different country that has a dropdown for regions (ex: UK <-> France). Ensure the error does not occur.

## Checklist

- [x]   I have added tests to cover my changes, if necessary.
- [x]  I have added translations for new strings, if necessary.
- [x]  I have updated the documentation accordingly, if necessary.


packages/peregrine/lib/talons/Region/useRegion.js
```diff
@@ -33,24 +33,25 @@ export const useRegion = props => {
...
     // initialValues, we want to skip the first state change of the value being
     // initialized.
     useEffect(() => {
-        if (country) {
+        if (country && !loading) {
             if (hasInitialized.current) {
-                regionInputFieldApi.exists() && regionInputFieldApi.reset();
-                regionSelectFieldApi.exists() && regionSelectFieldApi.reset();
+                regionInputFieldApi.exists() && regionInputFieldApi.setValue();
+                regionSelectFieldApi.exists() &&
```
michaelyu0123, line 48: format to same line

michaelyu0123 (approved):

packages/peregrine/lib/talons/Region/useRegion.js
pedrochiossi: ok

michaelyu0123 (approved):
LGTM

michaelyu0123 (approved):

PR #3388, opened 2021-08-30
[PWA-1798] Avoid an excessive DOM size from duplication of navigation

mikhaelbois:
## Description

There is a new Lighthouse warning `Avoid an excessive DOM size` with the delivery of community contribution PWA-1682. This is a byproduct of duplicating the DOM of the layered navigation for desktop and mobile. This is the same approach that has been done with the mega-menu. Optimally, we'd like to avoid this warning, and find an alternative solution to remove unnecessary duplication.

- New `useIsInViewport` hook that uses the Intersection Observer API  to know when an element is in the viewport
- Add conditions to load navigation and layered navigation components if their in the viewport or not
- Prevent FilterList and FilterItem components creation when not visible
- Expand filter list on load when filter item is checked and above the `itemCountToShow` limit

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-1798.

- Test on Category Page and Search Page
- Lighthouse report no longer includes warning: `Avoid an excessive DOM size`
- Validate that the desktop versions of the Main Navigation and Layered Navigation are not in the DOM on mobile
- Validate that the mobile versions of the Main Navigation and Layered Navigation are not in the DOM on desktop
- Note that once rendered, they will stay rendered to improve performance.

- Test main navigation on mobile and desktop and make sure no regressions impact their functionalities
- Test filters on mobile and desktop and make sure no regressions impact their functionalities


packages/venia-ui/lib/components/Header/__tests__/header.spec.js
```diff
@@ -32,6 +34,17 @@ jest.mock('@magento/peregrine/lib/talons/Header/useHeader', () => {
     };
 });
 
+jest.mock('@magento/peregrine', () => {
+    const useWindowSize = jest.fn(() => ({
+        isDesktop: mockIsDesktop
+    }));
+
+    return {
+        ...jest.requireActual('@magento/peregrine'),
```
JustinConabreeAbsolunet, line 43: No need to import actual since we're only importing `useWindowSize`

JustinConabreeAbsolunet (changes requested):
Very small change request!
Slightly worried about performance impact when expanding large list of filters as we're now going from null -> list, instead of just changing the CSS. But since there's two layers to expand, it shouldn't have an impact.
Also a bit worried about defaulting to desktop for layered nav + SSR/Prerender for mobile devices. QA should try to cover this case to see how it reacts.
Core member should review impact of change to useWindowSize, but it looks good to me!

packages/venia-ui/lib/components/Header/__tests__/header.spec.js
mikhaelbois: @JustinConabreeAbsolunet We do use `createTestInstance` in the test file, so without extending it, we won't have access to `createTestInstance`.

For the performance, you have a good point.
Do we want to impact the user experience for a slightly better Lighthouse score?

JustinConabreeAbsolunet: @mikhaelbois right! I missed that. All good then.

I think we're okay with how you've done it. When they expand, it would show the first _x_ items. Clicking the 'show more' may then be a bit slow, but I think for improved Lighthouse score we're okay. The sub-items aren't very complex so it shouldn't be too bad of a performance hit when they render. QA will just need to check that the form behaves okay with adding/removing fields.

JustinConabreeAbsolunet (approved):

packages/peregrine/lib/hooks/useWindowSize.js
```diff
@@ -51,9 +53,10 @@ export const WindowSizeContextProvider = props => {
     // This hook has side effects of adding listeners so we only want to create it
     // once and store it in context for reference by components.
     const windowSize = useWindowSizeListener();
+    const isDesktop = windowSize.innerWidth >= WINDOW_SIZE_DESKTOP;
```
jimbo, line 56: Unfortunately, I don't think `useWindowSize` should evaluate the window size (which is why it doesn't do so already). Layout is presentation, so component code shouldn't know about layout, and that means component code shouldn't use window size to conditionally render elements. Breakpoint conditions belong in CSS.

- I realize we already have a couple of components that call `useWindowSize`, compare the result to some magic breakpoint number, and conditionally render elements based on that comparison. Ideally, they wouldn't do that; at minimum, let's not encourage more components to do so.

- I also realize that the whole reason for this PR is to move the conditional from CSS to component code in order to avoid excessive DOM size. Again, it would be great if we would not have to do that.

Have we considered any other ways to avoid rendering DOM elements? For example, could we use [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to toggle a flag in state, which could then serve as the conditional for rendering elements?

jimbo (changes requested):

packages/peregrine/lib/hooks/useWindowSize.js
mikhaelbois: I was under the impression we were okay to use the window size since it was used elsewhere.
I will get back to the task and investigate how we can change that.
Ideally, we would remove the call to window size everywhere so others don't get caught like I was.

mikhaelbois: @jimbo I've reverted all work on the useWindowSize hook and created the useIsInViewport hook instead to achieve similar results like you've proposed.

jimbo (approved):

jimbo:
Maybe roll back the `isOpen` change on mobile navigation. Probably needs a more thorough a11y treatment.

mikhaelbois:
To prevent A11Y issues and also maybe SEO issues, we reverted the change on the mobile navigation that would remove its content from the DOM if not opened.

It impacts the numbers a bit but we are still in the green.

tjwiebell:
✅  Manual tests
✅  MFTF
❓  Cypress - some flaky tests that should be resolved by #3436 and #3432. Will wait for delivery of those and re-run Cypress.

revanth0212:
Hey @tjwiebell. I would suggest to merge the PR MFTF and Manual Tests have passed. Cypress tests stabilization might take a while. If you don't see any new error apart from the errors that occur on `develop` feel free to merge it.

tjwiebell (approved):
Approving ba60462, which was requested by original reviewer and fixed.

tjwiebell:
Merging ahead of blockers after re-verifying manual tests, MFTF, and Cypress tests added in this PR all pass.

PR #3389, opened 2021-08-30
[SPIKE] Webpack 5 Upgrade and webpack plugins, working yarn watch:venia

fooman:
## Description

This is from a trial I did trying to upgrade to webpack 5. It looks like we also are using a range of webpack plugins which are in need of a version bump. I got `yarn watch:venia` to compile again with these changes, but there seems quite a bit more fallout, including tests, to address. 

## Related Issue

Closes #ISSUE_NUMBER.

## Checklist


packages/venia-concept/package.json
```diff
@@ -35,7 +35,8 @@
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/master/packages/venia-concept#readme",
   "dependencies": {
-    "@magento/pwa-buildpack": "~10.0.0"
+    "@magento/pwa-buildpack": "~10.0.0",
+    "webpack": "~5.51.1"
```
fooman, line 39: May not be needed here.

dpatil-magento:
Closing as no recent activity.

PR #3393, opened 2021-09-01
[PWA-1244]: update url_suffix queries to fetch from storeConfig

pedrochiossi:
## Description

Updating Components to use one source of truth for url_suffix of products and categories, fetching from storeConfig instead of product/category entities.

## Related Issue

Closes PWA-1244

## Acceptance

Refactor all queries that use category or product url_suffix, to use the storeConfig suffix.

1. Verify that user is able to navigate to a product page, and a category page when USE_STORE_CODE_IN_URL=true
2. Verify that user can switch stores and view products and category pages when USE_STORE_CODE_IN_URL=true
3. Verify that user can search for a product, and view search page results when USE_STORE_CODE_IN_URL=true
4. Verify that user can do all of the above when USE_STORE_CODE_IN_URL=false

## Checklist


pankhurigoel (commented):
Nice work. Looks good to me.

packages/peregrine/lib/talons/Breadcrumbs/breadcrumbs.gql.js
```diff
@@ -1,5 +1,14 @@
 import { gql } from '@apollo/client';
 
+export const GET_STORE_CONFIG_DATA = gql`
+    query GetStoreConfigData {
```
tjwiebell, line 4: Minor nit; there is an eslint rule I'd love to have enabled right now for unique named operations, which is required to start doing some fun code and type generation. Couldn't take on that scope with the latest `validate-queries` PR, so that's still in the backlog. Consider not adding to this debt and giving new operations more unique names, eg. `GetStoreConfigForBreadcrumbs`.

tjwiebell (dismissed):
This looks good, had some minor suggestions for named operations, but not a blocker to approval. Nice work 👍

packages/peregrine/lib/talons/Breadcrumbs/breadcrumbs.gql.js
pedrochiossi: thanks @tjwiebell, I didn't know about this rule, I'll change the names

mikhaelbois:
@pedrochiossi 
- ✅ Verify that user is able to navigate to a product page, and a category page when USE_STORE_CODE_IN_URL=true
- ✅ Verify that user can switch stores and view products and category pages when USE_STORE_CODE_IN_URL=true
- ✅ Verify that user can search for a product, and view search page results when USE_STORE_CODE_IN_URL=true
- ✅ Verify that user can do all of the above when USE_STORE_CODE_IN_URL=false

However, I do have one question, from my understanding the url suffix shouldn't be something that changes often.
Couldn't we get the data on build once instead of asking for it in multiple different queries?
Just like we already do in `packages/pwa-buildpack/lib/queries/getStoreConfigData.graphql` for some configs.

pedrochiossi:
Yes @mikhaelbois, I had this same idea of fetching once and persisting, but I talked to @supernova-at he mentioned that it would be better to keep things updated with Admin. What you think @supernova-at?

tjwiebell (approved):

PR #3396, opened 2021-09-02
PWA-1686: [bug]: fix quantity field warning

michaelyu0123:
- Updated TextInput and QuantityFields component to use useField hook instead of useFieldState to ensure the field is rendered prior to asking for its fieldState.

RCA: Currently the way we are using informed libraries useFieldState will cause it to fire prior to the field being rendered due to the intention is for the field already rendered and then to add logic on top of the rendered input. This in effect actually will still work as intended as per https://github.com/joepuzzo/informed/issues/287 since it will just keep checking for it until the said field is rendered since it is made to be intentionally non-blocking. 

I resolved this issue in the TextInput component and the QuantityFields component.

As a fix, I had switched out the usage of useFieldState and opted in for the useField hook so the fieldState is generated alongside the component which we can ensure the field is rendered prior to getting the field state.

## Description

*Describe the bug*

This warning is spammed several times when you load a (product?) page:

!https://user-images.githubusercontent.com/1278869/116105745-38407700-a677-11eb-84ec-99d54bcbd3f7.png!

*To reproduce*

 1. Go to {{/default/night-out-collection.html}}
 2. Look at console warnings

*Expected behavior*

Should not see this warning in console.
h3. Acceptance Criteria
 * Visiting the product page does not cause warnings to be logged to the browser console
 * Root cause analysis: does this happen on all product pages? If not, why not?

h3. Test Plan
 1. Visit a PDP for each product type (simple, configurable, virtual, bundled)
 2. {{Bundled: /night-out-collection.html}}
 3. Open the browser console and verify that no warnings of the above type appear

## Related Issue

Closes #ISSUE_NUMBER.

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/CartPage/ProductListing/useQuantity.js
+++ b/packages/peregrine/lib/talons/CartPage/ProductListing/useQuantity.js
@@ -1,5 +1,5 @@
 import { useCallback, useMemo, useState, useEffect } from 'react';
-import { useFieldState, useFieldApi } from 'informed';
+import { useFieldApi } from 'informed';
 import debounce from 'lodash.debounce';
 
 /**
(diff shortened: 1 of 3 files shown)


michaelyu0123:
deferring to a different solution

PR #3397, opened 2021-09-03
[PWA-962] Apollo V3 Cache Eviction

JustinConabreeAbsolunet:
## Description

Refactoring cache clearing of customer and cart data to use Apollo v3's `evict` and garbage cleaning.

## Related Issue

Closes [PWA-962](https://jira.corp.magento.com/browse/PWA-962)

## Acceptance

- Utilities for manual cache deletion are deprecated - deleteCacheEntry
- These calls are replaced by cache.evict(), targeting specific entries that need purged
- After checking out, all references to cart are no longer present in InMemory or persisted cache in localStorage
- After logging out, all references to cart and customer data are no longer present in InMemory or persisted cache in localStorage

- Sign-in as a customer
- Verify localStorage to see Apollo cache entries for customer data
- Add a new customer address (through checkout or through account)
- Add items to cart
- Verify cart data in localStorage
- Place order
- Entries for cart should be cleared (Cart will remain but with new ID)
- Sign out
- Entries for customer data should not longer be in cache

## Breaking Changes (if any)

Removed `deleteCacheEntry` file/function

## Checklist


packages/peregrine/lib/Apollo/clearCustomerDataFromCache.js
```diff
@@ -1,14 +1,26 @@
...
  *
  * @param {ApolloClient} client
  */
 export const clearCustomerDataFromCache = async client => {
-    await deleteCacheEntry(client, key => key.match(/^\$?Customer/i));
+    // Cached data
+    client.cache.evict({ id: 'Customer' });
+    // Remove createCustomerAddress which references newly created addresses, avoiding gc
+    client.cache.evict({
+        id: 'ROOT_MUTATION',
```
fooman, line 14: While we are changing caching behaviour just a heads up that a future version of apollo will automatically remove ROOT_MUTATION https://github.com/apollographql/apollo-client/blob/main/CHANGELOG.md#potentially-disruptive-changes - not sure if this is something we rely on.

pankhurigoel (commented):
Looks good to me.

jimbo (approved):
Looks good. 👍

packages/peregrine/lib/Apollo/clearCustomerDataFromCache.js
tjwiebell: This is great news, and would mean we can remove all the `@connection(key: "createCustomerAddress")` overrides we do to prevent sensitive data from leaking.

On that note, why was this needed @JustinConabreeAbsolunet? I see two operations that use the `createCustomerAddress` mutation, and they both have connection keys. Is this just being extra safe and removing any reference to "customer", even if it's not PII?

JustinConabreeAbsolunet: @tjwiebell Without this, if we create a new address and then sign out, the new CustomerAddress is kept in cache because the createCustomerAddress contains a ref to it. Since Apollo's garbage collection only deletes entries that have no other references (unreachable), we have to manually remove this in order for garbage collection to work.

To Fooman's point though, this can be removed once we move to V3.4 of Apollo since ROOT_MUTATION will be cleared automatically.

revanth0212 (changes requested):
Nice work @JustinConabreeAbsolunet . This PR is almost there, one small tiny tiny thing left though. Looks like we still have order details left in `ROOT MUTATION`. Everything else looks perfect.

image

JustinConabreeAbsolunet:
@revanth0212 I fixed the issue with `placeOrder` remaining in `ROOT_MUTATION`

revanth0212:
> @revanth0212 I fixed the issue with `placeOrder` remaining in `ROOT_MUTATION`

Were you able to verify the fix?

When I tried, the field seems to be in the cache.

image

revanth0212 (approved):

revanth0212:
Tagged it as a minor because there are no API changes and `deleteCacheEntry` which was removed in this PR is internal-only util.

PR #3399, opened 2021-09-03
PWA-1686: [bug]: fix quantity field warning

michaelyu0123:
-Added wrapper for informed useFieldState hook to suppress benign warning

RCA: Currently the way we are using informed libraries useFieldState will cause it to fire prior to the field being rendered due to the intention is for the field already rendered and then to add logic on top of the rendered input. This in effect actually will still work as intended as per https://github.com/joepuzzo/informed/issues/287 since it will just keep checking for it until the said field is rendered since it is made to be intentionally non-blocking.  As per conversion with @jimbo have decided to switch out the solution to suppress specific warning only on initial load since this was done intentionally to reduce code complexity and allow us to leverage the informed libraries react components.

As a fix I added a wrapper hook around the useFieldState hook to ignore this specific warning on initial render.

## Description

*Describe the bug*

This warning is spammed several times when you load a (product?) page:

!https://user-images.githubusercontent.com/1278869/116105745-38407700-a677-11eb-84ec-99d54bcbd3f7.png!

*To reproduce*

 1. Go to {{/default/night-out-collection.html}}
 2. Look at console warnings

*Expected behavior*

Should not see this warning in console.
h3. Acceptance Criteria
 * Visiting the product page does not cause warnings to be logged to the browser console
 * Root cause analysis: does this happen on all product pages? If not, why not?

h3. Test Plan
 1. Visit a PDP for each product type (simple, configurable, virtual, bundled)
 2. {{Bundled: /night-out-collection.html}}
 3. Open the browser console and verify that no warnings of the above type appear

## Related Issue

Closes [PWA-1686](https://jira.corp.magento.com/browse/PWA-1686)

## Checklist


pankhurigoel (approved):
Looks good to me.

larsroettig (approved):

PR #3401, opened 2021-09-06
Cypress Docker Support

revanth0212:
## Description

This PR created a new script in the cypress tests directory that will help one with running and updating cypress test snapshots while running locally.

### run-tests
Usage: `yarn run test:ci <URL> <UPDATE_SNAPSHOTS_FLAG>` or direct invocation `./run-tests <URL> <UPDATE_SNAPSHOTS_FLAG>`
Runs your cypress tests in headless mode against the URL provided and snapshots present in the package.

If the `UPDATE_SNAPSHOTS_FLAG` is `false` (default value), snapshots won't be updated. This is the default configuration that will be used to run tests in the CI job.

If the `UPDATE_SNAPSHOTS_FLAG` is `true`, snapshots will be updated. 

Should also merge https://github.com/magento-commerce/pwa-studio-cicd/pull/87 while merging this.

## Related Issue

Closes PWA-1975

### Verification Stakeholders

@sharkySharks 

## Verification Steps
1. Check out the branch and run the `yarn test:ci <URL>` command with URL set to https://pr-3401.pwa-venia.com/
2. All tests should pass (row video test might fail which keeps failing on me every now and then, seems like a fragile test. should be updated)

## Checklist


venia-integration-tests/cypress-tests.dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN jq --arg PR_INSTANCE_URL $PR_INSTANCE_URL '.baseUrl = $PR_INSTANCE_URL' ./cypress.json > ./cypress.json.tmp && mv ./cypress.json.tmp ./cypress.json
+
+# update snapshot updation env variable
+RUN jq --arg UPDATE_SNAPSHOT $UPDATE_SNAPSHOT '.env.updateSnapshots = $UPDATE_SNAPSHOT' ./cypress.json > ./cypress.json.tmp && mv ./cypress.json.tmp ./cypress.json
+
+# install dependencies with yarn 
+RUN yarn install --frozen-lockfile
+
+# command to run tests
+CMD ["sh", "-c", "CYPRESS_baseUrl=$PR_INSTANCE_URL yarn test:ci"]
```
sharkySharks, line 35: newline

venia-integration-tests/run-tests.sh
```diff
@@ -0,0 +1,24 @@
...
+PR_INSTANCE_URL=$1
+
+# build docker with UPDATE_SNAPSHOT=true to update failed snapshots
+docker build -f cypress-tests.dockerfile -t cypress-test --build-arg PR_INSTANCE_URL=$PR_INSTANCE_URL --build-arg UPDATE_SNAPSHOT=false .
+
+# run cypress tests and remove the container once done
+docker run cypress-test
+
+# get container ID
+conatinerId=$(docker ps -a | grep "cypress-test" | awk '{print $1}')
```
sharkySharks, line 18: `containerId`

venia-integration-tests/update-tests.sh
```diff
@@ -0,0 +1,25 @@
...
+if [ $# -eq 0 ]; then
+    echo "Please provide URL to run tests against."
+    exit 1
+fi
+
+# collect pr instance url to run tests against
+PR_INSTANCE_URL=$1
+
+# build docker with UPDATE_SNAPSHOT=true to update failed snapshots
+docker build -f cypress-tests.dockerfile -t cypress-test --build-arg PR_INSTANCE_URL=$PR_INSTANCE_URL --build-arg UPDATE_SNAPSHOT=true .
```
sharkySharks, line 12: if the only difference between this script and the one above is `UPDATE_SNAPSHOT=true` then this value should just be passed in as a parameter to this script. no need to have 2 scripts doing the exact same thing.

venia-integration-tests/package.json
```diff
@@ -5,12 +5,13 @@
     "scripts": {
         "clearCache:mac": "rm -rf ~/Library/Application\\ Support/Cypress/cy/*",
         "test": "cypress open --browser chrome --config-file cypress.config.json",
+        "test:ci": "sh run-tests.sh",
         "test:debug": "node --inspect-brk ./node_modules/cypress/bin/cypress run --browser chrome --config-file cypress.config.json",
         "test:devTest": "jest",
         "test:firefox": "cypress open --browser firefox --config-file cypress.config.json",
-        "test:headless": "cypress run --browser chrome --headless --config-file cypress.config.json",
+        "test:headless": "sh run-tests.sh",
```
sharkySharks, line 12: rather than run these commands with `sh`, you specify `bash` in the scripts itself `#!/usr/bin/env bash`, so run `chmod a+x` on both files and then you can change these commands to `./run-tests` and drop the `.sh` from the end.

venia-integration-tests/update-tests.sh
revanth0212: You are right, actually, they had different logic initially when I created 2 files. I changed the code yesterday and forgot to optimize the usage. Will do it.

venia-integration-tests/package.json
```diff
@@ -5,13 +5,12 @@
     "scripts": {
         "clearCache:mac": "rm -rf ~/Library/Application\\ Support/Cypress/cy/*",
         "test": "cypress open --browser chrome --config-file cypress.config.json",
-        "test:ci": "sh run-tests.sh",
+        "test:ci": "./run-tests",
```
sharkySharks, line 8: I see this script tags parameters, where are you passing those parameters? I don't see any here or in `test:headless`

sharkySharks (approved):

venia-integration-tests/package.json
revanth0212: They are run-time parameters and will be sent at runtime by the user.

PR #3406, opened 2021-09-08
[PWA-845] Handling Authorized Routes

mikhaelbois:
## Description

- Create new `AuthRoute` component that will return the Route or a Redirect if the path needs authentication and the user is not signed in.
- Added new options `authed` and `redirectTo` to route targets.
- `authed="true"` tests if user is signed in before routing to desired route.
- `redirectTo="/"` is the path used to redirect user to when they are not signed in.
- Created Sign In Page and Forgot Password Page to connect to the current Create Account Page.
- By default, if a user tries to access a page that needs authentification and is not signed in, they will be redirected to the new Sign In Page.
- The page they tried to access will be kept in memory so they can be redirected to that page once they either sign in or create a new account. If none is found, they will be redirected to the Order History Page.

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-845.

- As a guest, navigate to /order-history and verify you're redirected to the `/sign-in` page.
- Test every route that needs authentication and make sure you are redirected to the `/sign-in` page.
- On both mobile and desktop, signing in via the `/sign-in` page should redirect you to the `/order-history` page.
- Run the above case again, but go through the `/create-account` page.
- Test while trying to access the pages directly.
- Test when trying to access a protected route. The page you tried to access should be kept in memory when you navigate between the `/sign-in` page, `/create-account` page and the `/forgot-password` page.
- Signing in or creating an account via the menu should **not** redirect you to the `/order-history` page.

- As a customer, navigate to Order History through account menu and page loads as expected.
- Do the same as above but directly navigate, page loads as expected.
- Trying to access the `/sign-in` page or `/create-account` page should redirect you to the `/order-history` page.

- Test `/forgot-password` page.
- On mobile and desktop, test Sign In, Create Account and Forgot Password sections in Navigation.
- On mobile and desktop, test Sign In, Create Account and Forgot Password sections in Checkout.

## Checklist

- [x] I have added translations for new strings, if necessary.


larsroettig (dismissed):
LGTM but I would like to opinion about API from my point maybe not ideal if you want to customized page for signing

packages/peregrine/lib/talons/SignInPage/useSignInPage.js
```diff
@@ -0,0 +1,61 @@
...
+    const [{ isSignedIn }] = useUserContext();
+
+    // Keep location state in memory when pushing history and redirect to
+    // the `from` url instead when signing in
+    const historyState =
+        history && history.location && history.location.state
+            ? history.location.state
+            : {};
+    const fromRedirectUrl =
+        historyState && historyState.from ? historyState.from : null;
```
tjwiebell, line 34: All of this type checked derived state might be a little overkill. Can we guarantee that react-router will always return a `history.location.state` object?

packages/venia-ui/lib/components/CreateAccountPage/createAccountPage.js
```diff
@@ -1,32 +1,62 @@
...
 
 const CreateAccountPage = props => {
-    const talonProps = useCreateAccountPage();
-    const { initialValues, handleCreateAccount } = talonProps;
+    const { signedInRedirectUrl, signInPageUrl } = props;
     const classes = useStyle(defaultClasses, props.classes);
+    const { handleOnCancel, initialValues } = useCreateAccountPage({
+        signedInRedirectUrl,
+        signInPageUrl
+    });
```
tjwiebell, line 18: Some new patterns have emerged that I think could be applicable here. I know we do an awful job at documenting outside of the core team, so consider these optional but valuable improvements.

1. Start passing raw props to talons, ie. `useCreateAccountPage(props)`. Less important here since you're passing all the props anyway, but gives more power to wrapped talon targets to override original props. 
2. Prop Objects. Instead of the talon returning individual props that get composed in the component, return a prop object that we can spread onto components.

```js
// in the talon
const createAccountProps = {
    initialValues,
    isCancelButtonHidden: false,
    onCancel: handleOnCancel
};

// component
const { createAccountProps } = talonProps;
return <CreateAccount {...createAccountProps} />
```

packages/venia-ui/lib/components/CreateAccountPage/createAccountPage.js
```diff
@@ -1,32 +1,62 @@
...
+
 CreateAccountPage.propTypes = {
     classes: shape({
-        container: string
+        root: string,
+        header: string,
+        contentContainer: string
     }),
-    initialValues: shape({})
+    initialValues: shape({}),
```
tjwiebell, line 57: Might as well fix this up while we're here, this isn't a prop of the component, it comes from the talon.
```suggestion
```

packages/venia-ui/lib/components/Routes/authRoute.js
```diff
@@ -0,0 +1,38 @@
...
+                }}
+            />
+        );
+    }
+
+    return <Route {...rest}>{children}</Route>;
+};
+
+AuthRoute.defaultProps = {
+    authed: false,
```
tjwiebell, line 28: I vaguely remember this being brought up, and can't remember why we went this direction. I would expect that any `<AuthRoute>` is assumed to need authentication and you wouldn't need this property.

packages/venia-ui/lib/components/Routes/routes.js
```diff
@@ -17,7 +18,7 @@ const Routes = () => {
                  * Client-side routes are injected by BabelRouteInjectionPlugin here.
                  * Venia's are defined in packages/venia-ui/lib/targets/venia-ui-intercept.js
                  */}
-                <Route>
+                <AuthRoute>
```
tjwiebell, line 21: I think this makes the whole app require authentication, right?

packages/venia-ui/lib/targets/makeRoutesTarget.js
```diff
@@ -11,11 +11,17 @@ function makeRoutesTarget(venia) {
...
+            ? JSON.stringify(route.redirectTo)
+            : null;
+        const redirectToProp = redirectTo ? `redirectTo={${redirectTo}} ` : '';
+
         routeList.prependJSX(
             'Switch',
-            `<Route ${route.exact ? 'exact ' : ''}path={${JSON.stringify(
-                route.pattern
-            )}}><${AddedRoute}/></Route>`
+            `<AuthRoute ${exact}${authed}${redirectToProp}path={${path}}><${AddedRoute}/></AuthRoute>`
```
tjwiebell, line 24: In here I would expect to see the logic to either render `<Route>` or `<AuthRoute>` depending on the `authed` prop.

tjwiebell (changes requested):
Really nice work 👍  I don't share the API concern with @larsroettig but did have some suggestions and questions that could use clarification.

packages/venia-ui/lib/components/Routes/routes.js
mikhaelbois: Yes. But if we don't provide the authed prop, it will go straight to a Route component.
Since we inject code with makeRoutesTarget, Routes needs to have the import of AuthRoute.

But I could do something like this so it remains accessible.

```
// Import used in makeRoutesTarget
// eslint-disable-next-line no-unused-vars
import AuthRoute from './authRoute';

const Routes = () => {
    const { pathname } = useLocation();
    useScrollTopOnChange(pathname);

    return (
        <Suspense fallback={fullPageLoadingIndicator}>
            <Switch>
                {/*
                 * Client-side routes are injected by BabelRouteInjectionPlugin here.
                 * Venia's are defined in packages/venia-ui/lib/targets/venia-ui-intercept.js
                 */}
                <Route>```

**Comment by mikhaelbois** [comment]
@tjwiebell Just updated the code based on your comments.
Could you take a look?

File: packages/venia-ui/lib/components/Routes/routes.js [hunk 7]
```diff
@@ -1,10 +1,14 @@
...
 
+import { useScrollTopOnChange } from '@magento/peregrine/lib/hooks/useScrollTopOnChange';
 import { fullPageLoadingIndicator } from '../LoadingIndicator';
 import HomePage from '../HomePage';
 import MagentoRoute from '../MagentoRoute';
-import { useScrollTopOnChange } from '@magento/peregrine/lib/hooks/useScrollTopOnChange';
+
+// Import used for makeRoutesTarget
+// eslint-disable-next-line no-unused-vars
+import AuthRoute from './authRoute';
```
tjwiebell, line 11: It would be preferable to add the import to `makeRoutesTarget`, and we can remove these lines.

```js
// function addRoutes
const AuthRouteComponent = routeList.addImport(
    'import AuthRoute from "./authRoute"'
);

...

const Component = route.authed ? AuthRouteComponent : 'Route';
```

tjwiebell (changes requested):
Move that import and this is golden 🙌

packages/venia-ui/lib/components/Routes/routes.js
mikhaelbois: @tjwiebell Done. Thanks for the code!

tjwiebell: Oh no, my pseudo code didn't make this clear, sorry 🙂 the `AuthRouteComponent` should be outside of the for loop; I don't think there's any reason to have a unique import per route.

mikhaelbois: Yeah, I should have figured that out myself. 😅 
I just moved it.

tjwiebell (approved):

revanth0212:
I tried the following steps and it took me to the sign-in page after I signed out instead of the home page. Is that expected?

1. As a guest go to `/order-history` page
2. When the app takes you to `/sign-in` page, login
3. Should take you to `/order-history` page
4. Sign out and it takes you to the `/sign-in` page instead of home page

mikhaelbois:
@revanth0212 
Good question.
Since we're signing out when we're on a protected route, we will be redirected to the page defined with `redirectTo` in the defaultRoutes.
I don't know if we want something different for this case.

revanth0212:
**`e2eTests/wishList/singleWishlistAddRemoveProduct.spec.js`** cypress test seem to be failing. Can you please check?

I verified against develop and it works there. So there must be a change that was introduced in this PR.

revanth0212:
Code ✅ 
Functionality ❓ (https://github.com/magento/pwa-studio/pull/3406#issuecomment-920394391)
Cypress ❌ 
MFTF ❓

mikhaelbois:
@revanth0212 
Regarding the last comment, it was decided by the team to redirect to the `redirectTo` page.
Which is now the Sign In Page by default.

revanth0212:
@mikhaelbois can you please take care of the [failing cypress test](https://github.com/magento/pwa-studio/pull/3406#issuecomment-921041911) as well?

revanth0212:
I have run it multiple times and seems like the wishlist integration test is failing.

Command: `yarn test:headless https://pr-3406.pwa-venia.com/`

image

mikhaelbois:
@revanth0212 
Multiple Wishlist test should now be fixed.
Converted the two failing MFTF tests to Cypress tests for Customer Account Actions and Customer Address Book Actions.
`CustomerAccountAccess -> e2eTests/customerAccount/accountActions`
`VerifyAccountAddressBook -> e2eTests/customerAccount/addressBookActions`
We'll either have to skip those two MFTF tests or remove them.

revanth0212:
Merging the PR even though `pwa-pr-test` is failing because it fails on `develop` as well and it will be fixed soon.

revanth0212 (approved):

PR #3407, opened 2021-09-08
[PWA-1909] Fix CSS Module source maps

### LINKED ISSUE #3294: [Issue] [FEATURE] Improve className format during development
This issue is automatically created based on existing pull request: magento/pwa-studio#3293: [FEATURE] Improve className format during development

---------
## Description
When in development finding where a component is located should be as easy as possible. When inspecting an element identifying a component by it's classname is a easy way to do so hence I improved the classname by adding the path to the file where the component is located.

## Related Issue

Closes #Didnt create one.

## Verification Steps

- pull this pr
- start the watcher
- open the chrome devtools and notice the classnames now point to the actual file in a more explicit format
- close the watcher and start the stage version of the storefront, note that the classes are **not** pointing to the file structure

tjwiebell:
## Description

When in development finding where a component is located should be as easy as possible. When inspecting an element identifying a component by it's classname is a easy way to do so hence I improved the classname by adding the path to the file where the component is located.

## Implementation Note
This is a proposed alternative to changing class names in development mode. This PR fixes source maps, and if you correctly setup workspaces in DevTools, you can live edit CSS from the browser. Hot reloading does seem to be triggering full reload on CSS changes right now, but that appears to be an existing issue not introduced by this change.

There is still a development need to have component handles that aren't as fragile as class names, but we need to think more broadly about their applicable use.
- Testing - Cypress could definitely use some `data-testID` handles so selectors are less fragile.
- Extensibility - We would like a solid API for targeting venia-ui components, and relying on class names would again be too fragile to support until the project starts maintaining longer term releases. Re-using the data attribute method, we could add `data-venia-handle` ids on component that will strictly follow semver, and only be used at compilation.

I think source maps will solve your immediate need while we design a data attribute component handle proposal for longer term use.

## Related Issue

* [[PWA-1909](https://jira.corp.magento.com/browse/PWA-1909)] [Issue] [FEATURE] Improve className format during development
* Closes #3293 
* Closes #3294

1. `yarn watch:venia` and open DevTools. Verify that css classes are now source mapped.

source-map-demo mov

## Checklist


michaelyu0123 (approved):

michaelyu0123:
lgtm

larsroettig (approved):
LGTM @tjwiebell nice job #kudos

jimbo (approved):

JustinConabreeAbsolunet:
QA Approved! Sourcemap works in watch (all or venia), and stripped in production

PR #3410, opened 2021-09-09
[PWA-1704] UPWARD static content race condition

JustinConabreeAbsolunet:
## Description

Refactoring UPWARD JS plugin to resolve assets and configurations sequentially, instead of in parallel. This was causing a race condition which outputs different content from build to build.
Also changed the asset mapping to only take the overriden value when two configurations with the same configuration path are present.

## Related Issue

Closes [PWA-1704](https://jira.corp.magento.com/browse/PWA-1704)

### Verification Stakeholders

@supernova-at @jimbo 

**Replace venia-static**
- Scaffold a new app (with create-pwa)
- Create a `venia-static` folder in the app and add a couple of test files
- Add the following to the upward.yml
```
veniaStaticIncludes:
  resolver: directory
  directory:
    resolver: inline
    inline: './venia-static'
```
- Run `yarn build:prod`
- Validate that `dist/venia-static` only contains the test files, not the base files from venia (folder is overriden)

**Change venia-static to custom-static**
- Create a `custom-static` folder in the app and add a couple of test files
- Add the following to the upward.yml
```
veniaStaticIncludes:
  resolver: directory
  directory:
    resolver: inline
    inline: './custom-static'
```
- Run `yarn build:prod`
- Validate that `dist/custom-static` is present, and `dist/venia-static` is not

## Breaking Changes (if any)

Overriden static asset configurations now only take the overriden value, instead of both.

## Checklist


larsroettig (approved):
LGTM

zetlen (approved):
LGTM. Went over the change with @JustinConabreeAbsolunet and confirmed that this is the sane way for the plugin to work. Thanks Justin!

PR #3412, opened 2021-09-13
PWA-2071: Markup updates to improve SEO and Accessibility

larsroettig:
## Description

As a merchant, I want Venia to demonstrate the best practice implementation that scores 100 at Google Lighthouse for SEO, Accessibility and Best practices *so that* I have confidence in choosing PWA solution from Adobe Commerce and don't need to invest my own resources in customizations associated with this area. Venia should deliver this out of the box.

*Acceptance Criteria*
 * Lighthouse does not report the "Heading elements are not in a sequentially-descending order" error
 ** This error occurs on the Category Page (Follow Us in footer) and on the Product Page for configurable products (option headings)
 ** Follow the guidance here: [https://web.dev/heading-order/]
 * Lighthouse does not report the "Buttons do not have an accessible name" error
 ** This error occurs on the product tiles. The product thumbnail images are actually buttons.
 ** Follow the guidance here: [https://web.dev/button-name/]
 ** Also, we've been looking for a way to introduce [react-aria|https://react-spectrum.adobe.com/react-aria/] to the project and this scope may be perfect for it. Implement their [useButton|https://react-spectrum.adobe.com/react-aria/useButton.html] hook here.

## Related Issue

Closes PWA-2071

## Checklist


pankhurigoel (approved):
Looks good to me.

mikhaelbois:
@larsroettig 
Still getting the "Buttons do not have an accessible name" error on thumbnail buttons. ❌ 
https://pr-3412.pwa-venia.com/carina-cardigan.html
image

But the "Heading elements are not in a sequentially-descending order" error is gone. ✅ 

The "[aria-*] attributes do not match their roles" error linked to some icons is currently being corrected in another task.

JustinConabreeAbsolunet:
QA Smoke test passes (outstanding issue Mikhael pointed out is fixed). Core member will need to run Cypress/MFTF

image

mikhaelbois:
Got the following error in MFTF.
```
vendor/bin/mftf run:test GuestIsAbleToAccessCategories -r
...
[goToCategoryViaMegaMenu] ClickOnMegaMenuSubCategory
 I wait for pwa element visible "//div[contains(@class, 'megaMenuItem-meg...",60
  [waitForCategoryToLoad] wait for pwa element visible "//div[contains(@class, 'megaMenuItem-megaMenuItem-')]/a[contains(text(), 'Accessories')]",60
 I move mouse over "//div[contains(@class, 'megaMenuItem-megaMenuItem-')]/a[..."
  [hoverMouse] move mouse over "//div[contains(@class, 'megaMenuItem-megaMenuItem-')]/a[contains(text(), 'Accessories')]"
 I wait for pwa element visible "//div[contains(@class, 'submenu-submenuI...",60
  [waitForSubCategoryToLoad] wait for pwa element visible "//div[contains(@class, 'submenu-submenuItems-')]//h3[contains(text(), 'Belts')]",60
 ERROR
```
Got the following fail in MFTF.
```
vendor/bin/mftf run:test VerifySearchFunctionality -r
...
[assertNoSearchedResults] AssertNoSearchResults
 I wait for pwa element visible "div[class^='main-page-']",60
  [waitForSearchItemToLoad1] wait for pwa element visible "div[class^='main-page-']",60
 I see "No results found!"
  [seeNoResultsAreDisplayed] see "No results found!"
 FAIL
```

mikhaelbois:
New MFTF PR to update `GuestIsAbleToAccessCategories` test.
https://github.com/magento-untitled-geese/pwa-tests/pull/1

packages/venia-ui/lib/components/ProductImageCarousel/thumbnail.js
```diff
@@ -66,13 +66,14 @@ const Thumbnail = props => {
     }, [file, isDesktop, label, classes.image]);
 
     return (
-        <button
+        <span
             className={isActive ? classes.rootSelected : classes.root}
             onClick={handleClick}
             type="button"
+            aria-hidden="true"
```
JustinConabreeAbsolunet, line 73: Why aria-hidden?
Can't we just make this an AriaButton with an aria label like the others?

JustinConabreeAbsolunet (changes requested):

packages/venia-ui/lib/components/ProductImageCarousel/thumbnail.js
mikhaelbois: From my understanding, we are using `aria-hidden="true"` to prevent any accessibility interactions.
So it wouldn't be useful to use AriaButton here.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute

Also, it was transformed to a `span` since we shouldn't be using `aria-hidden` on focusable elements.

However, it shouldn't be `type="button"` but `role="button"` in this case. I will do the update.

mikhaelbois: Also, it might be useful to move the AriaButton component to the root so we can use it elsewhere.
Or maybe use it for every Button?

tjwiebell (approved):
Re-approving after mainline merge since original reviewer is out today

PR #3422, opened 2021-09-17
Added local URL support for cypress in docker

revanth0212:
## Description

Updates the cypress headless tests script to work with local Magento instances. In the previous solution, we had only worked with fully quantified URLs, like the PR instance URLs because they are publicly deployed on the internet. But the local instances are local to your machine and docker when it runs a container creates its own network which is detached from the host machine network. That's the reason why cypress when it runs on docker works to be able to resolve a local app server to run tests on.

This PR solves the problem by handling local URLs differently.

## Related Issue

PWA-2144

### Verification Stakeholders

@sharkySharks 

## Verification Steps

1. Checkout the branch
2. Run `yarn test:headless <local instance URL>` - All tests should run but you can stop after 1 test success, no need to run all
3. Run `yarn test:headless https://develop.pwa-venia.com` - All tests should run but you can stop after 1 test success, no need to run all

**_Note: To stop testing in between, open another terminal and run `docker ps` to get the ID and `docker kill <ID>` to kill the container._**

None

## Breaking Changes (if any)

None

## Checklist


larsroettig (approved):

venia-integration-tests/run-tests
```diff
@@ -17,5 +17,18 @@ PR_INSTANCE_URL=$1
 # collect update screenshots flag
 UPDATE_SCREENSHOTS=${2:-false}
 
-# run cypress tests and remove the container once done
-docker run --rm -v $PWD:/venia-integration-tests -w /venia-integration-tests --entrypoint=cypress cypress/included:8.3.1 run --browser chrome --config baseUrl=$PR_INSTANCE_URL,screenshotOnRunFailure=false --config-file cypress.config.json --env updateSnapshots=$UPDATE_SCREENSHOTS --headless
+# see if its a local instance or a pr instance
+PORT="$(echo $1 | sed -e 's,^.*:,:,g' -e 's,.*:\([0-9]*\).*,\1,g' -e 's,[^0-9],,g')"
```
fooman, line 21: Seems a bit of a stretch to assume simply because it has a PORT number for it to be local. Can't we just add
`USE_DOCKER_NETWORK=${3:-false}`
or 
`IS_LOCAL=${3:-false}`

revanth0212: This solution is built for apps built with buildpack which has a port while developing locally. You are right though buddy, I thought about adding a new arg since the second arg is already an optional arg I figured its easy to not ask for it 😅

PR #3448, opened 2021-09-27
[PWA-912] SW should be able to handle the new configured URL suffix

mikhaelbois:
## Description

The URL suffix can be anything, not just .html.

Unfortunately, Service Worker HTML handling logic depends on the suffix being .html. We need to update this and handle the other use cases as well.

This update uses the request's destination to check if it's a HTML document.
https://developers.google.com/web/tools/workbox/modules/workbox-strategies#configuring_strategies
https://developer.mozilla.org/en-US/docs/Web/API/Request/destination#value

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-912.

When the URL suffix is not .html, the service worker should still fetch and cache it.

1. Update a page's URL suffix in store config to be something other than `.html`.
2. Load Venia (Make sure you run a builded version)
3. Navigate to the page that has a different URL suffix
4. See that it gets an entry in your browser's Application > Cache Storage > workbox-runtime-{site name} cache

## Checklist


packages/venia-concept/src/ServiceWorker/__tests__/registerRoutes.spec.js
```diff
@@ -63,6 +67,76 @@ test('There should be a route for robots.txt, favicon.ico and manifest.json with
...
+
+    const mockFindSameOrLargerImage = jest.fn(() => {
+        return Promise.resolve(null);
+    });
+
+    findSameOrLargerImage.mockImplementation(mockFindSameOrLargerImage);
+
+    const mockCacheHandler = jest.fn(data => data);
+
+    createImageCacheHandler.mockImplementation(() => ({
```
JustinConabreeAbsolunet, line 122: Is there a missing `expect` for checking if the cache handler was called?

JustinConabreeAbsolunet (dismissed):
LGTM!

fooman:
We have always done away with the isHtml check as well so great to see this removed. Wondering though now that the html response can be variable via the computed response from the shimmer work and “soon” ssr if we should already switch to a different strategy like fallback only for offline.

packages/venia-concept/src/ServiceWorker/__tests__/registerRoutes.spec.js
mikhaelbois: @JustinConabreeAbsolunet You're right. I've added the missing line that checked if it was called. Thanks for the heads up!

JustinConabreeAbsolunet:
Hey @fooman , good call out! It was actually switched to `NetworkFirst` instead of `StaleWhileRevalidate` in [the shimmer PR](https://github.com/magento/pwa-studio/pull/3428/files#diff-6fd77ed71dcede1ad0bed86a9625e06ea66305fbc1ce9e7f4c123d4fcd2d70feR80). This way it will always serve new content unless it's offline. Do you have another recommendation?

JustinConabreeAbsolunet (approved):

pedrochiossi:
QA approved:
✅ SW is caching pages that don't have .html as suffix
 
Screen Shot 2021-09-28 at 9 47 55 AM
✅ Cypress Tests
✅ MFTF

fooman:
@JustinConabreeAbsolunet thanks for getting back to me. Let’s get this merged and I’ll need to get our project caught up and will then take another holistic look at what html should come from the sw when. FYI we used a pre cached navigation route so far which removed the network component completely https://developers.google.com/web/tools/workbox/modules/workbox-routing#how_to_register_a_navigation_route

PR #3449, opened 2021-09-28
CIF-1465: directly import simplePersistence in user reducer

Buuhuu:
## Description

While we set up jest unit tests on our project using peregrine we had issues configuring the redux store using the reducers exported by peregrine. Our repository does not directly depend on react-router-dom but due to the user reducer importing Util from the peregrine index we would have to provide the whole dependency tree as dev dependencies. 

In particular the following import chain is problematic:

peregrine/store/reducers/user.js 
=> peregrine/index.js
=> peregrine/hooks/usePagination
=> react-router-dom

(there may be more).

The resolution is as simple as replacing the import for the BrowserPersistence with a direct import, similar to how it is done in the auth middleware. 

## Related Issue

CIF-1465
Closes PWA-2362.

Run the unit tests to confirm no regressions were introduced.

## Checklist

- [X] Unit tests pass


mikhaelbois:
Hey @Buuhuu,
This looks good to go in QA on our side.
Do you mind filling in the "Test scenario(s) for direct fix/feature" section so we can QA this properly?
Thanks!

Buuhuu:
Thanks @mikhaelbois to get back on this. 

I am afraid but I don't have exact verification steps for that. However I believe existing unit test coverage is sufficient to QA this.

In our test cases we had to workaround this in https://github.com/adobe/aem-core-cif-components/pull/709/files#diff-f57c8d97f683ec8155c2afb113a53a9ff73f5cc497c0c9e1e7a4fbe099243640R78 . With the fix applied we would be able to use the reducers as exported by peregrine.

mikhaelbois:
Hello @Buuhuu,
No worries, we'll do an overall QA based on this change.
It's rather simple so it should be okay.
We will let you know!

ericeoeur (approved):

PR #3463, opened 2021-10-04
fix(scaffolding): find tarball with fs instead of npm output

zetlen:
## Description

- I changed the Venia scaffolding behavior in "debug mode" to find the new `.tar.gz` files using filesystem calls, instead of using the filename reported by NPM.

The Venia `_buildpack/create.js` special test flag `DEBUG_PROJECT_CREATION` breaks on NPM >=7.23.0, because it relies on specific output from `npm pack --json` and NPM changed that JSON.

JSON output from the `npm` command isn't reliable; it doesn't have an official spec or contract, so we shouldn't be relying on it. Instead, we should rely only on the assumption that `npm pack` will create a `.tar.gz` file in the working directory in which it was run (package root). The new process during the `DEBUG_PROJECT_CREATION` work flow is, for each Venia dependency in the `packages` folder:

1. Run `npm pack`
2. Get a directory listing
3. Filter it for `.tgz` files
4. Find the most recently modified one (presumably seconds ago)

This seems to work with all NPM and Yarn (classic) versions I've tested.

## Related Issue

Closes PWA-2131.

## Acceptance

An internal team member and a RAP representative, at least, should reproduce the bug using `develop` and then observe the fix from this branch.

### Verification Stakeholders

image

## Verification Steps

Best to repro right before checking the fix, to compare results.

#### Repro
1. In your PWA Studio root directory, switch to the `develop` branch.
2. Run `npm --version`. If it not version 7.x.x, run `npm install -g npm@7` to install NPM7. (You can reverse this later, or do it in NVM.)
3. Run `DEBUG_PROJECT_CREATION=true ./packages/pwa-buildpack/bin/buildpack create-project "scaffold-venia-develop" --template "packages/venia-concept" --name "scaffold-venia-develop" --author "pwa-ci <pwatest-ci@example.com>" --npm-client "yarn"`
4. The install phase should fail. You should see an error like `Package "" refers to a non-existing file`. This is because `npm pack --json` output a different filename than the one it actually created on disk.

#### Verification
5. Run `rm -rf ./scaffold-venia-develop && rm -rf ./packages/*/*.tgz` to remove scaffolding artifacts.
6. Check out this branch: `git checkout zetlen/PWA-2131-scaffold-debug-tarball-fix`
7. Run `npm --version` to make sure you're still on NPM 7
8. Run the create-project command from step 3 above

No, this is tooling-side only.

This should work on all supported versions of Node and NPM.

## Checklist

### DIFF
--- a/packages/venia-concept/_buildpack/__tests__/create.spec.js
+++ b/packages/venia-concept/_buildpack/__tests__/create.spec.js
@@ -1,8 +1,5 @@
 jest.mock('child_process');
 const { execSync } = require('child_process');
-execSync.mockImplementation((cmd, { cwd }) =>
-    JSON.stringify([{ filename: `${cwd.split('/').pop()}.tgz` }])
-);
 
 const { dirname, resolve } = require('path');
 const packagesRoot = resolve(__dirname, '../../../');
@@ -197,7 +194,8 @@ describe('when DEBUG_PROJECT_CREATION is set', () => {
                 '@magento/venia-ui': '1.0.0'
             },
             devDependencies: {
-                '@magento/peregrine': '1.0.0'
+                '@magento/peregrine': '1.0.0',
+                '@magento/pwa-buildpack': '1.0.0'
             },
             scripts: {},
             optionalDependencies: {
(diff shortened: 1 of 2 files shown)


revanth0212 (approved):
Code and Functionality look good. Another masterpiece from @zetlen 👏

PR #3482, opened 2021-10-12
Revert "[PWA-962] Apollo V3 Cache Eviction (#3397)"

justinconabree:
## Description

This reverts commit aee413191e1ee7e008da7faa2130d11959a4f22a.
Issue detected. Checkout cache is kept in apollo cache when switching store views before checkout. v3 cache eviction will only clear the cache for the current persistor.

## Related Issue

Closes PWA-2233

### Verification Stakeholders

@revanth0212 

- Login
- Switch to default store view if you are not already on it
- Checkout a product, fill out all details on the checkout page but do not complete the order yet
- Switch to the french store (or some other store)
- Logout
- Check the apollo cache data, checkout data should be gone from french store **AND** default store.
Last step was previously failing. Default store would be unchanged


revanth0212:
Functionality ✅  (Thank you @jcalcaben)
Cypress ✅  (Thank you @jcalcaben)
MFTF ✅  (Thank you @mikhaelbois)

PR #3485, opened 2021-10-13
Update size of Breadcrumbs container to allow focus ring to be visible

### LINKED ISSUE #3484: [bug]: Focus ring not visible for breadcrumbs
**Describe the bug**

Keyboard users will have a great deal of trouble determining when they're focused on breadcrumbs due the edges of the focus ring being cut off

**To reproduce**

1. Go to https://venia.magento.com/default/juno-sweater.html
2. Click on the text "Juno Sweater" in the breadcrumbs (this isn't a link)
3. Press Shift + TAB
5. TAB/Shift + TAB through the breadcrumbs to see the issues of the focus ring being cut off

**Expected behavior**

In order to preserve accessibility, a user should always be able to clearly see which element is in focus

**Screenshots**

image

**Possible solutions**

Make the containing element of the Breadcrumbs larger.
Rather than using `margin` to position the element, we should use `padding`.
We'll then need to address the hard coded height of the element, to either factor in the added padding, or otherwise remove the height in favour of another placeholder method

**Debug Report**

Command not found

-   Device [e.g. iPhone6, PC, Mac, Pixel3]: Desktops
-   Browser [e.g. Chrome, Safari]: All
-   Browser Version [e.g. 22]: All
-   Magento Version [e.g Magento Commerce 2.4]: N/A

-   [x] `venia-ui`

luke-denton-aligent:
Using padding to "position" the breadcrumbs gives the focus ring space to actually render itself.
Adding extra padding to the bottom of the component allows for the bottom of the focus ring to be visible, and to compensate for that added height (and to ensure we're not pushing the rest of the UI down as a result), we negate that padding by using a negative margin.

## Description

This PR updates the size of of the Breadcrumbs DOM container so that the focus ring is visible to keyboard users.

The change has been made such that the overall height of the Breadcrumbs component is exactly the same, so the rest of the UI isn't pushed, pulled or otherwise altered.

## Related Issue

Closes #3484.

Verify that keyboard users can now see the full focus ring for the breadcrumbs when navigating the site using their keyboard

N/A

N/A

Yes, browser testing is needed as Breadcrumb UI component may be impacted on Desktop and mobile browsers

N/A

Focus ring is now visible
image

In the following image, we can see a comparison between the fixed Breadcrumb component, and the previous implementation. The focus rings have been turned on for both, to be able to clearly see which is which, but I've also labelled. We can see that the product title still lines up perfectly. The top of the browser was used to align the two images.
image

This update has also inadvertently fixed the letter `p` in this particular breadcrumb (as it would for any other letter with a descender)

## Breaking Changes (if any)

N/A

## Checklist


jcalcaben:
@luke-denton-aligent can you resolve the conflict with the recent tailwind css changes to `breadcrumbs.module.css`?

luke-denton-aligent:
Thanks for letting me know @jcalcaben, conflict has now been resolved

JustinConabreeAbsolunet:
✅ QA passed
✅ Cypress

Waiting on scaffold to pass to merge

jcalcaben (approved):

JustinConabreeAbsolunet:
Scaffolding fixed. Merging

Thanks @luke-denton-aligent !

PR #3495, opened 2021-10-18
[PWA-2190] [bug] - Fix stale cartId on multiple tabs

pedrochiossi:
## Description

This PR fixes the problem of having staled cartId tokens saved on Redux when running the same store in multiple tabs.

In order to keep the cartId updated across tabs, this fix introduces a storage listener on cart context that fires a page reload whenever cartId is changed from another tab. The reload will force Redux to persist new cartId on storage.

 

## Related Issue

Closes PWA-2190

## Acceptance

Users should be able to run multiple tabs without having staled cartId saved on storage.

1. Add  a product to cart
2. Open a new tab
3. Place an order in a new tab
4. Open first tab
5. Click add product to cart

- Verify if product is added to cart without any errors
- Verify automated test coverage is added for this use case

## Checklist

- [x]  I have added tests to cover my changes, if necessary.
- [x]  I have added translations for new strings, if necessary.
- [x]   I have updated the documentation accordingly, if necessary.


MichaelHeinzman:
Hey, I'm just reading the different commits and I was wondering if there was a reason we reload the page instead of just updating the cart in redux using dispatch?

brendanfalkowski:
Just for reference, we also had issues with this PR causing non-stop refreshing with two tabs open.

For us, it happened because `cartState.cartId` is an integer, and `storage.getItem('cartId')` is a string. So this comparison must change:

- `(cartId && currentCartId && cartId !== currentCartId)` — always fails
- `(cartId && currentCartId && parseInt(cartId) !== parseInt(currentCartId))` — normalized typeof values

Our codebase is a PWA Studio monorepo started in 2018 (super early). We still use the REST-based cart APIs due to customizations we're maintaining. In that era, local storage `cartId` was a string. When PWA Studio switched the GQL-based cart APIs, I'm guessing they started putting `cartId` as an integer in local storage.

PR #3533, opened 2021-11-03
PWA-2355 UPWARD JS resolver for Computed type

JustinConabreeAbsolunet:
## Description
To bring the JS implementation of the UPWARD spec closer to the PHP implementation, we've added the Computed resolver to it as well. The resolver simply returns an empty string so the application can continue without error.

## Related Issue

Closes PWA-2355

### Verification Stakeholders

@zetlen

### Specification

Changes to `upward-spec` and/or `upward-js` packages must be reviewed

- Run `DEBUG=upward-js:ComputedResolver yarn stage:venia`
- Visit front-end
- Should see `upward-js:ComputedResolver Computed resolver is meant for UPWARD PHP only` in the terminal
- Site should function normally, and inlined data should be missing from source (see screenshot for reference)
image

- Test using UPWARD PHP
- Inlined data should be available (Can look for global variable `INLINED_PAGE_TYPE` in browser console)

## Checklist


mikhaelbois (approved):
LGTM ✅

packages/upward-js/lib/resolvers/ComputedResolver.js
```diff
@@ -0,0 +1,18 @@
...
+    static get resolverType() {
+        return 'computed';
+    }
+    static get telltale() {
+        return 'computed';
+    }
+    async resolve() {
+        debug('Computed resolver is meant for UPWARD PHP only.');
+
+        return '';
```
fooman, line 14: Wouldn't we be able to at least supply webpageChunks based on the page type similar to how the php implementation globs for those files?

JustinConabreeAbsolunet: @fooman Not without the page type information unfortunately. Each root component is its own chunk (sometimes more than one), and we only inline the chunks for the current page type.

PR #3571, opened 2021-11-23
PWA-2232 Updating default messages with variables to match EN

JustinConabreeAbsolunet:
## Description

Default messages whose EN version contained a variable were frequently missing the variable. This lead to messages like `Qty: ` in the mini-cart and elsewhere. This PR sets the default value for formatted texts to be equal to their EN values (containing the variables)

## Related Issue

Closes PWA-2232

- Test a store view whose locale is not EN
- Make sure translation file was not loaded
- Mini-cart item's quantity should contain the number value

## Checklist

- [ ] ~I have added tests to cover my changes, if necessary.~


pedrochiossi (approved):
All good @JustinConabreeAbsolunet.

dpatil-magento:
Verified with and without language pack and looks good. Cypress tests pass.

PR #3574, opened 2021-11-24
[PWA-1784] [bug]  - Fix SearchBar blocking click events on filter/sort buttons

pedrochiossi:
## Description

The SearchBar component has a `mousedown` event listener on the document to hide itself when clicking outside the area of its container element. This was blocking click events on `FilterItem` buttons and `ProductSort` because the page height would change after hiding searchBar and pointer location would invalidate `mouseup` events and not fire `onclick`.

This PR fixes these issues by using using `mousedown` event handlers on both Filter and Sorting buttons so it will fire always regardless of the page movement. 

## Related Issue

Closes PWA-1784

1. Search for `Tops`.
2. Click on the Price label for 100-200.
3. Ensure that filter is selected.
4. Search  for `Dress`.
5. Click on sorting dropdown.
6. Select a sorting option.
7. Ensure that product list is sorted accordingly. 

Ensure modified buttons are accessible through keyboard (Tab + Enter/Space) 

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/MagentoRoute/__tests__/useMagentoRoute.spec.js
+++ b/packages/peregrine/lib/talons/MagentoRoute/__tests__/useMagentoRoute.spec.js
@@ -37,12 +37,8 @@ jest.mock('@apollo/client', () => {
 
 const givenQueryResult = response => {
     useLazyQuery.mockReset();
-    useLazyQuery.mockImplementation((fetchUrl, { onCompleted }) => {
-        const { data = {} } = response;
-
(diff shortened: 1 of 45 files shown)


mikhaelbois:
@pedrochiossi 
Code looks good.
However, coverage could be improved a bit. Do you mind checking `packages/venia-ui/lib/components/ProductSort/sortItem.js` ?
Also, PR environment seems down at the moment.

mikhaelbois (approved):
@pedrochiossi LGTM ✅ 
PR environment is still down, but this can be tested locally.

PR #3575, opened 2021-11-24
PWA-2352 Fix intermittent load issue with category page

JustinConabreeAbsolunet:
## Description

Updating the apollo client changed the reliability of the `onComplete` option for apollo's `useLazyQuery` hook. Updating the apollo client to latest (3.5.4) fixed the intermittent issue, but caused a similar issue with the initial load when the query was served from cache.
This PR removes the usage of `onComplete` and transfers the logic into a `useEffect` based on the query response. This means more cycles through the file, but higher reliability.

## Related Issue

Closes PWA-2352

**Must test with both UPWARD JS (no inlining) and UPWARD PHP (with inlining)**

Navigate through the site ensuring that pages load correctly.
Things to test
- Page types (CMS/Homepage, PLP, PDP, cart/search/other static route)
- Test both initial load, and navigating between pages
- 404 route
- Make sure root component shimmers display correctly when navigating to PLP

## Checklist

- [ ] ~I have added tests to cover my changes, if necessary.~

### DIFF
--- a/packages/peregrine/lib/talons/MagentoRoute/__tests__/useMagentoRoute.spec.js
+++ b/packages/peregrine/lib/talons/MagentoRoute/__tests__/useMagentoRoute.spec.js
@@ -37,12 +37,8 @@ jest.mock('@apollo/client', () => {
 
 const givenQueryResult = response => {
     useLazyQuery.mockReset();
-    useLazyQuery.mockImplementation((fetchUrl, { onCompleted }) => {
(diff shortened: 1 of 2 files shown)


mikhaelbois (approved):
LGTM 💹 
Some comments to explain the logic could be a good idea here eventually.

mikhaelbois:
@JustinConabreeAbsolunet Could you add a version label?

PR #3578, opened 2021-11-25
[bugfix/3539] fixed issue #3539

### LINKED ISSUE #3539: [bug]: Environment variables not parsed in template.html when runing watch
Discovered by community member
> after the Venia 12 upgrade. Environment variables (at least in template.html) seem to have stopped working with yarn watch but fine on deployment.

Using the `templateContent` parameter of the webpack's HtmlWebpackPlugin bypasses the environment variable parsing. Most likely related to the following found in the [plugin's github](https://github.com/jantimon/html-webpack-plugin#writing-your-own-templates)
> ⚠️ templateContent does not allow to use webpack loaders for your template and will not watch for template file changes

The reason for using this to begin with was the UPWARD inlining changes. When running watch, we bypass upward, so the mustache code remained in the source unchanged. This was a way to strip out all of the inlined mustache code when running watch

Need to investigate how to keep the benefits of the mustache code being removed and keep environment variables parsed

Jira Issue: https://jira.corp.magento.com/browse/PWA-2374

**To reproduce**
1. Run yarn watch
2. Inspect front-end source
3. Variables in the html tags are unparsed

**Expected behavior**
- Variables are parsed correctly when running watch

**Possible solutions**

Easiest solution would be to have a "dev" version of the template file that is identical, minus the inline code. Can then update the webpack.config.js to swap template files instead of reading the content.

-   [x] `venia-concept`
-   [ ] `venia-ui`
-   [ ] `pwa-buildpack`
-   [ ] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

Jordaneisenburger:
## Description

This fixes the related issue :)

## Related Issue

Closes #3539

### Verification Stakeholders

@JustinConabreeAbsolunet 

Run the watcher notice that the <= tags get rendered as their actual value instead of the tag name.

no

no

no

not that I'm aware of

## Checklist


packages/venia-concept/.gitignore
```diff
@@ -13,3 +13,4 @@ merged-schema*.json
 test-report.xml
 test-results.json
 yarn-error.log
+template.generated.html
```
pedrochiossi, line 16: missing new line here

pedrochiossi (changes requested):
Code looks good, just need an extra line on .gitignore

pedrochiossi (approved):

ericeoeur:
QA Passed, LGTM. Thanks!

PR #3586, opened 2021-11-30
[PWA-2326]: Refactor deprecated type_id on ProductInterface 

pedrochiossi:
## Description

This PR updates graphQL queries and related functionality that uses deprecated `type_id` field from `ProductInterface` to use `__typename` field instead. 

## Related Issue

Closes PWA-2326

## Verification Steps

1. Ensure PWA Studio no longer uses the deprecated ProductInterface.type_id field

1. Verify that searching for all different supported product types functions correctly (simple, configurable, etc.)

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/Gallery/useAddToCartButton.js
+++ b/packages/peregrine/lib/talons/Gallery/useAddToCartButton.js
@@ -9,7 +9,7 @@ import operations from './addToCart.gql';
  * @param {String} props.item.uid - uid of item
  * @param {String} props.item.name - name of item
  * @param {String} props.item.stock_status - stock status of item
- * @param {String} props.item.type_id - product type
+ * @param {String} props.item.__typename - product type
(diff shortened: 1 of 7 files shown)


packages/peregrine/lib/talons/Gallery/useAddToCartButton.js
```diff
@@ -21,10 +21,10 @@ import operations from './addToCart.gql';
  *
  */
 const UNSUPPORTED_PRODUCT_TYPES = [
```
fooman, line 23: Would be good to swap the logic around here as product types can be custom per store.

pedrochiossi: I agree @fooman, but I think it would be out of the scope of this task.

michaelyu0123 (approved):
LGTM

dpatil-magento:
QA Approved.

PR #3580, opened 2021-11-25
PWA-2307 Update product GraphQL field from id to uid

JustinConabreeAbsolunet:
## Description

Updating all queries that fetch a ProductInteface entity to use the new `uid` instead of the deprecated `id`

## Related Issue

Closes PWA-2307

#### Test scenario(s) for direct fix/feature

**General**
A full regression of the site (alternatively: anything that touches products) needs to be performed here.

**Add to Cart Dialog**
- Sign in
- Add a configurable item to your wishlist without selecting any options
- Navigate to your wishlist
- Attempt to add the configurable product to the cart
- Verify that the Add to Cart dialog functions correctly
- Verify the steps above with a simple product

**Order History Row**
- Sign in
- Complete the checkout flow with a simple product in cart
- Go to your Order History page
- Verify that the page loads correctly
- Navigate back to the homepage
- Go offline
- Navigate back to the Order History page
- Verify that the page still loads correctly. Note that there is currently an error toast "failed to fetch" for product thumbnail images. That error is unrelated to this task.

## Checklist

- [ ] ~I have added tests to cover my changes, if necessary.~
- [ ] ~I have added translations for new strings, if necessary.~
- [ ] ~I have updated the documentation accordingly, if necessary.~

PR #3589, opened 2021-12-01
[refactor] Updated GraphQL use of UrlResolver to Route

jcalcaben:
## Description

Updates the field Query.urlResolver, which is deprecated, to use the route query instead.

## Related Issue

Closes PWA-2324

## Acceptance

Any developer

### Verification Stakeholders

Any developer

## Verification Steps

1. Run dev server: `yarn watch:venia`
2. Verify application runs with no logging errors
3. Click through home page, category page, and product page
4. Verify no logging errors
5. Use store switcher
6. Verify store switching still works as expected

## Checklist


packages/peregrine/lib/talons/Header/storeSwitcher.gql.js
```diff
@@ -11,11 +11,20 @@ export const GET_STORE_CONFIG_DATA = gql`
...
-export const GET_URL_RESOLVER_DATA = gql`
-    query getUrlResolverData($url: String!) {
-        urlResolver(url: $url) {
-            id
+export const GET_ROUTE_DATA = gql`
+    query getRouteData($url: String!) {
+        route(url: $url) {
             type
+            ... on CmsPage {
+                identifier
```
fooman, line 19: Might want to pick up the changes coming from here
https://github.com/magento/pwa-studio/pull/3580/files

packages/peregrine/lib/talons/Header/storeSwitcher.gql.js
```diff
@@ -11,11 +11,20 @@ export const GET_STORE_CONFIG_DATA = gql`
...
+            }
+            ... on ProductInterface {
+                id
```
pedrochiossi, line 22: as @fooman pointed out, these should be uid

packages/peregrine/lib/talons/Header/storeSwitcher.gql.js
```diff
@@ -11,11 +11,20 @@ export const GET_STORE_CONFIG_DATA = gql`
...
+                __typename
+            }
+            ... on CategoryInterface {
+                id
```
pedrochiossi, line 26: uid

pedrochiossi (approved):

mikhaelbois:
@jcalcaben 
Please disable linting rules in `packages/peregrine/lib/talons/Header/storeSwitcher.gql.js` and update with `develop`.
`# eslint-disable-next-line @graphql-eslint/require-id-when-available`

Root components still work as needed.
However, now that we send the `uid` to the Category Root Component, we should update the code to reflect that change.
We should also update how we get the Category data to use the `category_uid` filter instead of `ids`.
Like you've suggested, we should open a follow-up task to do those updates.

mikhaelbois:
QA ✅

jeremycharron (approved):

PR #3597, opened 2021-12-03
PWA-2319: Update CartItemInterface GraphQL field from id to uid

michaelyu0123:
- Updated cartiteminterface

## Description

The field CartItemInterface.id is deprecated. We should use the `uid` field instead.

By default, Apollo client uses the `__typename` and `id` field for caching so we will also need to update the fields that Apollo client uses to cache CartItemInterface objects from `id` to `uid`.

Found in:

peregrine/lib/talons/CheckoutPage/OrderConfirmationPage/createAccount.gql.js
peregrine/lib/talons/CreateAccount/createAccount.gql.js (x2)
peregrine/lib/talons/SignIn/signIn.gql.js
venia-ui/lib/components/SignIn/signIn.gql.js
venia-ui/lib/components/LegacyMiniCart/cartOptions.js
venia-ui/lib/components/LegacyMiniCart/product.js
Note: We don't have to fix the LegacyMiniCart queries, as that component is itself already deprecated.

Acceptance Criteria
PWA Studio no longer requests the `id` field on `CartItemInterface` objects
`CartItemInterface` objects are still cached appropriately

## Related Issue

Closes #[PWA-2319](https://jira.corp.magento.com/browse/PWA-2319)

## Verification Steps

Test Plan
Verify that creating an account and signing in still function correctly
 how to test that cart items in your cart are cached after signing in    (Merge cart flow may be? OR just go offline and reload page?)
Test upgrade flow. So either on cloud or local hit all the changed area above in 12.1 (master)  and then run staging server with this pr and check all areas in new tab of same browser. Details steps on upgrade test can be found here https://jira.corp.magento.com/browse/PWA-1418

## Checklist


mikhaelbois (approved):
LGTM ✅

dpatil-magento:
QA Aprroved

dpatil-magento (approved):

PR #3608, opened 2021-12-09
[PWA-2411] Comment out server webpack config

pedrochiossi:
## Description

PR to remove server webpack config from code until we fully implement SSR. The configuration was creating a SERVER.js chunk that was failing the bundlesize check, we are not yet using this chunk so it is safe to comment out.  

## Related Issue

Closes PWA-2411

## Acceptance

Build should pass the chunk size checks in `yarn bundlesize`

1. On project root, run `yarn bundlesize`
2. All bundles sizes should be below defined limit
 

## Checklist


fooman:
Would it make sense to instead use
excludeAssets
to then exclude `server.js`?
see https://www.npmjs.com/package/webpack-bundle-analyzer

revanth0212:
The code looks good and seems to have fixed the issue.

> Would it make sense to instead use
> excludeAssets
> to then exclude `server.js`?
> see https://www.npmjs.com/package/webpack-bundle-analyzer

@fooman I do agree with you, it would be a far easier solution to ignore the bundle instead of commenting it out but since this is a bundle that we won't be using till SSR is ready, I feel we can comment it out and reduce the compile time by a tiny bit.

What do you think about that?

revanth0212 (approved):

revanth0212 (approved):

PR #3613, opened 2021-12-10
Better cypress results logging

revanth0212:
## Description

When we added parallelization of cypress tests one of the cons was inability to see all the test results as a single unit at the end. This was because when a user decides to parallelize cypress tests, there will be `n` number of cypress threads and they wont be talking to each other to aggregate results at the end.

With this PR, we have added a new feature that will aggregate all the results and present them at the end in a readable tabular format. This functionality adds features to not only report the statues of all the tests but also aggregates the errors if any and prints them in red to catch the user's eye.

In case of all passes:

image

If something fails:

image

## Related Issue

Closes PWA-2453

### Verification Stakeholders

@dpatil-magento 

## Verification Steps

Run cypress tests as is and you should see results at the end.

`yarn test:headless -t 4 --url https://develop.pwa-venia.com/`


pedrochiossi (approved):
looks good

ericeoeur:
After multiple QA attempts, there were several errors including Webpack Compilation Errors and test hanging when attempting to connect to the browser. @revanth0212 @pedrochiossi 

image

ericeoeur:
All looking good on the QA side-- will be merging momentarily.

PR #1936, opened 2019-10-25
DONT Merge Test PR

dpatil-magento:
## Description

Test PR

## Related Issue

Closes #ISSUE_NUMBER.

### Verification Steps

1. Test PR

## Checklist

* I have updated the documentation accordingly, if necessary.
* I have added tests to cover my changes, if necessary.

PR #3619, opened 2021-12-16
[PB-PWA-1936]: Sorting product listing by custom attribute

itsyanik:
## Description

- Gets available sort methods from `products` query
- As of writing, when sorting products by a new `custom_attribute` in ASC order, the returned products from the query will list products that do not have that attribute first.
- Had to hardcode some sorting values as they were coming from the component's `defaultProps` in `productSort.js` and there is no way to merge incoming + default props.

## Related Issue

Closes [#1936](https://jira.corp.magento.com/browse/PWA-1936)

## Verification Steps
- Visit [Bottoms-page](https://pr-3619.pwa-venia.com/venia-bottoms/venia-pants.html?page=1)
- Check if sort options are alphabetically ordered
- Check that custom attribute `Fit` is in the list
- Verify that right click does not select sorting
- Sort list by clicking the `Fit` custom attribute (Products that contain this custom attribute: `Clara Wide Leg Pants (Fit: Baggy), Gloria Palazzo Pants (Fit: Baggy), Bella Eyelet Capris (Fit: Fit), Selena Pants (Fit: fit)`
- Baggy < Fit so Baggy attributes should show up first when ordering by ascending
- Check same steps above using different language/store. To test this install venia language pack extension by running `yarn venia add -D @magento/venia-sample-language-packs`, run application and select `French Store View` on header. 

Verify user is able to sort data after applying filters on category page
Verify user is able to sort data after applying filters on search page


packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
```diff
@@ -70,8 +91,75 @@ export const useCategoryContent = props => {
     const categoryDescription = categoryData
         ? categoryData.categories.items[0].description
         : null;
+    const sortingMethods = sortData
+        ? sortData?.products?.sort_fields?.options
```
pedrochiossi, line 95: I don't think you need to use optional chaining here, is there any scenario where those fields would be undefined even with returned data? Also if you use the ternary operator on `sortData` you don't need to escape it with optional chaining, since it will always return a value.

packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
```diff
@@ -58,6 +67,18 @@ export const useCategoryContent = props => {
         }
     }, [categoryId, getFilters]);
 
+    useEffect(() => {
```
pedrochiossi, line 70: You can combine this effect on the previous one, there's no need to declare a new one.

packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
```diff
@@ -70,8 +91,75 @@ export const useCategoryContent = props => {
...
+                          id: `sortItem.${method.value}`,
+                          text: `${method.label}`,
+                          attribute: method.value,
+                          sortDirection: 'ASC'
+                      };
+                  }
+
+                  return sortingProps;
+              })
+              .filter(method => method !== undefined)
```
pedrochiossi, line 146: Why do we have undefined methods here?

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
```diff
@@ -224,7 +244,33 @@ export const useSearchPage = (props = {}) => {
 
     useScrollTopOnChange(currentPage);
 
+    const sortingMethods = sortData
+        ? sortData?.products?.sort_fields?.options
```
pedrochiossi, line 248: Same thing about optional chaining here

packages/peregrine/lib/talons/RootComponents/Category/categoryContent.gql.js
```diff
@@ -31,7 +31,23 @@ export const GET_CATEGORY_CONTENT = gql`
     }
 `;
 
+export const GET_CATEGORY_AVAILABLE_SORT_METHODS = gql`
+    query getCategoryAvailableSortMethods(
+        $categoryIdFilter: FilterEqualTypeInput!
+    ) {
+        products(filter: { category_id: $categoryIdFilter }) {
```
pedrochiossi, line 38: We are using category uid now, 
```suggestion
        products(filter: { category_uid: $categoryIdFilter }) {
```

pedrochiossi (changes requested):

packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
itsyanik: I had to because the tests were failing. Doing this way made the tests work

packages/peregrine/lib/talons/SearchPage/useSearchPage.js
itsyanik: Same as before too. The tests for this were breaking until I added them.

packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
itsyanik: `sortingMethods` starts with a `null` value until data is returned from the query. I'm just making sure these values don't go in.

packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
itsyanik: I've seen this in other components, thought it would be best practice to separate in two different `useEffect`s since they call different queries.

pedrochiossi: Ok, seems like the correct way.

packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
pedrochiossi: But you are escaping falsy values on the beginning at line 130.

itsyanik:
ATM there are just some small test portions breaking, I believe it should be good to go after they're fixed. But since I won't be here to fix that, here's a summary:

- Both `useCategoryContent` and `useSearchPage` had same code for this PR
  - It was abstracted into the `useProductSort`
  - All it does is take in an array of objects with sort options as an argument and then merges it with some default values
  - The default values are the same default props that used to live in the `ProductSort` component
  - They're hardcoded atm as we can't merge defaultProps with incoming ones
  - The test for it is breaking. One of the objects in the array has it's properties set as undefined - Didn't have time to figure it out

If you want to test what is happening, check the deploy link for this PR and visit the bottoms page, as it's suggested in the description. We still need to wait for some update on GraphQL for the results to show up correctly. As of now task should work as expected once the GraphQL is updated.

packages/peregrine/lib/talons/ProductSort/productSort.js
```diff
@@ -0,0 +1,70 @@
...
+                          attribute: method.value,
+                          sortDirection: 'ASC'
+                      };
+                  }
+
+                  return sortingProps;
+              })
+              .filter(method => !!method)
+        : null;
+    console.log(sortMethodsFromQuery);
```
mikhaelbois, line 53: `console.log()` to be removed.

mikhaelbois (changes requested):

DiegoBezerra89 (dismissed):

packages/venia-ui/lib/components/FilterModal/FilterList/filterItem.js
```diff
@@ -19,20 +19,31 @@ const FilterItem = props => {
         [title, value]
     );
 
-    const handleClick = useCallback(() => {
-        toggleItem({ group, item });
+    const handleClick = useCallback(
+        e => {
+            if (e.button !== 0) return;
```
mikhaelbois, line 24: Could we had comments where we have those conditions to explain why we do this?
Also, could we update the QA Steps to make sure we test this out?

packages/venia-ui/lib/components/ProductSort/sortItem.js
```diff
@@ -11,24 +11,32 @@ const SortItem = props => {
...
+    );
 
     const activeIcon = active ? <Icon size={20} src={Check} /> : null;
 
     return (
         <button
             className={classes.root}
             data-cy={active ? 'SortItem-activeButton' : 'SortItem-button'}
-            onMouseDown={handleClick}
+            onMouseDown={e => handleClick(e)}
```
mikhaelbois, line 39: Doesn't `onMouseDown={handleClick}` already send the event?

packages/venia-ui/lib/components/ProductSort/productSort.js
```diff
@@ -15,6 +15,32 @@ const ProductSort = props => {
     const { availableSortMethods, sortProps } = props;
     const [currentSort, setSort] = sortProps;
     const { elementRef, expanded, setExpanded } = useDropdown();
+    const { locale } = useIntl();
+
+    const orderSortingList = useCallback(
+        list => {
+            return list.sort((a, b) =>
```
mikhaelbois, line 22: Could we update the QA Steps to make sure we test this out in different languages?
What if we give it a different translation in SortItem?

mikhaelbois (changes requested):

packages/venia-ui/lib/components/FilterModal/FilterList/filterItem.js
pedrochiossi: Yes sorry, I'll add some comments. The problem here is that by switching to onMouseDown events every click (right, left, middle) can trigger it, so I'm making sure we only call the action on left mouse click. This was an error from a bug fix in [PWA-1784](https://jira.corp.magento.com/browse/PWA-1784)

packages/venia-ui/lib/components/ProductSort/sortItem.js
pedrochiossi: You're right, I'll fix it.

packages/venia-ui/lib/components/ProductSort/productSort.js
pedrochiossi: Ok, no problem. I believe `localeCompare` will order here by following language rules so it's safe.

mikhaelbois (approved):
@pedrochiossi ✅ 
Just update the Verification Steps with the mouse click events and localization and you're good to go for me!

packages/venia-ui/lib/components/ProductSort/productSort.js
pedrochiossi: I had to change this to not rely on language rules, since the sort options from graphQL won't get translated.

DiegoBezerra89 (dismissed):

packages/venia-ui/lib/components/ProductSort/productSort.js
```diff
@@ -15,6 +15,43 @@ const ProductSort = props => {
...
+                const aText = formatMessage({
+                    id: a.id,
+                    defaultMessage: a.text
+                });
+                const bText = formatMessage({
+                    id: b.id,
+                    defaultMessage: b.text
+                });
+
+                return aText.localeCompare(bText, undefined, {
```
mikhaelbois, line 32: If `formatMessage` returns the localized string, why do a `localeCompare` instead of a regular alphabetical order with `sort()`?

mikhaelbois (approved):
LGTM ✅

dpatil-magento:
@itsyanik @pedrochiossi Please check following - 
1. Attribute is displayed by default in all categories instead of displaying for categories which has product associated with custom attribute.
2. When a add french language pack and switch to french store > Dont see store specific label being fetched for Sort custom attribute.
3. Go to https://magento-venia-concept-ub-0n.local.pwadev:9536/dev-p-test.html?page=1 and try sorting by `devp product attr` > 404 error page is being displayed

pedrochiossi:
@dpatil-magento, answering your feedbacks:

1. From my understandig reading the task acceptance criteria, the attribute sort options should appear in both category and search page by following the `Used for sorting in Product Listing` option in Admin.

2. @mikhaelbois fixed this by adding localized labels in the `sort_fields` response in https://github.com/magento-commerce/magento2-pwa/pull/17

3. This is expected behavior, according to [Docs](https://devdocs.magento.com/guides/v2.4/graphql/queries/products.html#sort-attribute). The `Use in Search` should be set to Yes to sort by custom attribute that is not in `ProductAttributeSortInput` type.

dpatil-magento:
@pedrochiossi As discussed https://pr-3619.pwa-venia.com/venia-tops.html?page=1  Price filter is ending up in 404 error.

dpatil-magento:
QA Approved

dpatil-magento (approved):

PR #3624, opened 2021-12-28
[PWA-1913] Viewing product attributes of type select in PDP

mikhaelbois:
## Description

Show custom attributes **of type Select** on Product Detail Page.

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-1913.

1. Needs PWA module update from https://github.com/magento-commerce/magento2-pwa/pull/14 in order to work.
2. Go to any Product Detail page and check the new "Details" section under the SKU.
3. The custom attributes listed there should be visible only when _Visible on Catalog Pages on Storefront_ is set to "yes".
4. The custom attributes should be ordered by their position.

## Checklist


fooman:
Any thoughts on why this is tied to commerce only?

mikhaelbois:
Hey @fooman
It's not a commerce only feature. What gives you this idea?
Is it because of the backend PR?
If so, it relates to the [PWA Metapackage](https://developer.adobe.com/commerce/pwa-studio/metapackages/open-source/).
Basically, that PR will give us access to the `is_visible_on_front` value.

fooman:
@mikhaelbois thanks for getting back to me. The required backend PR only being accessible in a commerce only repo gave me the idea. I unfortunately can't see what the actual change is as I do not have access so could not tell if this would work the same or different with just open source + open source pwa metapackage.

Watch3ruk:
After committing all the changes i have the same error like on the deployed version : 
Default Store View - Google Chrome 05_01_2022 10_13_51
i know its related to : productDetailFragment.gql.js

mikhaelbois:
Hey @Watch3ruk,
If you don't have the code from this [PR](https://github.com/magento-commerce/magento2-pwa/pull/14), it won't work.
At the moment, only people with access to that code can test this out.
Sorry for the inconvenience.

Watch3ruk:
Hi , @mikhaelbois.

Thanks, any idea when this will be released ?

mikhaelbois:
@Watch3ruk 
From my point of view, I don't think it will make the cut for the upcoming 12.2 release.
So it would most probably be for 13.0.

packages/venia-ui/lib/components/ProductFullDetail/customAttributes.js
```diff
@@ -0,0 +1,251 @@
...
+                    b.attribute_metadata.sort_order
+            );
+    }, [customAttributes]);
+
+    const list = useMemo(() => {
+        if (cleanAttributes.length === 0) {
+            return [];
+        }
+
+        return cleanAttributes.map(
```
tjwiebell, line 44: Including the filter and sort iterations above to init `cleanAttributes`, this would be the third pass through these attributes, which could be troublesome as this list of attributes grows. I suggest using a `reduce` call instead, and doing this in a single pass.

```js
customAttributes.reduce( (attributeList, currentAttribute) => {
  // filter
  if (!currentAttribute.attribute_metadata.is_visible_on_front) {
    const attributeContent = existingLogic();
  
    // sort
    attributeList.splice(currentAttribute.attribute_metadata.sort_order, 0, attributeContent)
  }

  return attributeList;
}, [])

File: packages/venia-ui/lib/components/ProductFullDetail/customAttributes.js [hunk 2]
```diff
@@ -0,0 +1,251 @@
...
+                        {attributeContent}
+                    </li>
+                );
+            }
+        );
+    }, [
+        classes.attributeLabel,
+        classes.listItem,
+        classes.attributeContentHtml,
+        classes.attributeContent,
```
tjwiebell, line 216: Optional, but this identity shouldn't change, can simplify dep list.
```suggestion
        classes
```

packages/venia-ui/lib/components/ProductFullDetail/customAttributes.js
```diff
@@ -0,0 +1,251 @@
...
+            }) => {
+                const key = attribute_metadata.uid;
+                const attributeLabel = attribute_metadata.label;
+                const {
+                    ui_input_type: type,
+                    is_html_allowed: isHtml
+                } = attribute_metadata.ui_input;
+                let attributeContent;
+
+                switch (type) {
```
tjwiebell, line 58: I would suggest extracting this logic into its own component, and let a supporting talon handle as much as possible. 

Optional, but definitely valuable for extensibility, I would maybe even take this a step further and apply factory pattern here (`<AttributeFactory attribute={attribute} />`), and break each of these cases up into tiny components.

tjwiebell (changes requested):
Looks good, just a few suggestions to make this more readable and efficient.

packages/venia-ui/lib/components/ProductFullDetail/customAttributes.js
mikhaelbois: Thanks for the proposition!
I've moved the sorting logic in the backend to simplify things.

packages/venia-ui/lib/components/ProductFullDetail/customAttributes.js
mikhaelbois: Done!

packages/venia-ui/lib/components/ProductFullDetail/CustomAttributes/AttributeType/Price/price.js
```diff
@@ -0,0 +1,74 @@
+import React, { Fragment } from 'react';
+import { shape, string } from 'prop-types';
+
+import { BrowserPersistence } from '@magento/peregrine/lib/util';
+import { useStyle } from '@magento/venia-ui/lib/classify';
+import { default as PriceComponent } from '@magento/venia-ui/lib/components/Price';
```
tjwiebell, line 6: The default export should already allow declaring your own variable name.
```suggestion
import PriceComponent from '@magento/venia-ui/lib/components/Price';
```

tjwiebell (approved):
Incredible work 🎉  love the structure of the new code, and I think this is a great start to opening up extensibility. One minor nit, but more of an FYI than required for approval.

tjwiebell (dismissed):

dpatil-magento:
@mikhaelbois Thanks for fixing Cypress. The cypress build got out of some error, we are waiting for github action pr to merge in order to trigger cypress build. I will test locally, for now dont worry on remaining failed test i will test and let you know.

dpatil-magento:
Deployed graphql pr on backend and cypress looks good now pr instance - ✔  All specs passed!                        37:46       80       80.
QA Approved.

dpatil-magento (approved):

aigarsbedeicis:
Hey, so this has been merged into 12.2, but I can't see required metapackage released in magento/magento2-pwa repo.
Is this intended? Currently updated our project to 12.2 and our PDP is broken. 
Am I missing something?

Watch3ruk:
update composer for magento with : 
"magento/module-upward-connector": "2.0.0",
"magento/pwa": "0.1.0"

PR #3625, opened 2021-12-28
[PWA-2456] Home page tabbing leads to page transition

mikhaelbois:
## Description

Fix keyboard events on Page Builder components to avoid history push events on other things than a click or keyboard accepted events.

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-2456.

- Navigate through on the home page until you reach Page Builder content while using `tab` on your keyboard
- Try to tab through Page Builder content
- Pressing `tab` **should not** redirect to the url
- Pressing `enter` or `space` **should** redirect to the url
- Clicking on the button or link should redirect to the url

## Checklist

### DIFF
--- a/packages/pagebuilder/lib/ContentTypes/Banner/banner.js
+++ b/packages/pagebuilder/lib/ContentTypes/Banner/banner.js
@@ -274,8 +274,8 @@ const Banner = props => {
 
     const history = useHistory();
 
-    const clickHandler = e => {
-        handleHtmlContentClick(history, e);
+    const clickHandler = (event, isClick = false) => {
+        handleHtmlContentClick(history, event, isClick);
     };
 
     let BannerFragment = (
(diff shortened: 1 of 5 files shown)


packages/pagebuilder/lib/ContentTypes/Banner/banner.js
```diff
@@ -274,8 +274,8 @@ const Banner = props => {
 
     const history = useHistory();
 
-    const clickHandler = e => {
-        handleHtmlContentClick(history, e);
+    const clickHandler = (event, isClick = false) => {
```
pedrochiossi, line 277: Instead of relying on a second parameter, can't we validate click using `event.type`? https://developer.mozilla.org/en-US/docs/Web/API/Event/type

pedrochiossi (commented):
Looks good, just had a question about the click handler implementation.

packages/pagebuilder/lib/ContentTypes/Banner/banner.js
mikhaelbois: Good catch, fixing this now!

mikhaelbois: @pedrochiossi ✅

pedrochiossi (approved):

dpatil-magento:
QA Aprroved

PR #3646, opened 2022-01-12
[PWA-2498] Rename Link's prefetchType prop

### LINKED ISSUE #3647: [Issue] Rename Link's prefetchType prop
This issue is automatically created based on existing pull request: magento/pwa-studio#3646: Rename Link's prefetchType prop
## 
## Description

For the `Link` component, rename the boolean prop `prefetchType` to `shouldPrefetch`. The former appears to indicate it will contain a type rather than a boolean, causing some confusion. The approach herein

- Deprecates the existing `prefetchType` prop to `Link` and `useLink`, but does not remove it
- Adds `shouldPrefetch` as a new prop to both component and talon
- Ignores the value of `prefetchType` unless it is `true`

## Related Issue

#3308 

### Verification Stakeholders

- @jcalcaben 
- @luke-denton-aligent 

## Test Plan
Note - Link component not being used on Venia 
1. Unit tests should pass
2. Cypress Test should pass

## Breaking Changes (if any)

These changes should be non-breaking. Eventually, we should remove the deprecated prop, which would be a breaking change.

## Checklist

jimbo:
## Description

For the `Link` component, rename the boolean prop `prefetchType` to `shouldPrefetch`. The former appears to indicate it will contain a type rather than a boolean, causing some confusion. The approach herein

- Deprecates the existing `prefetchType` prop to `Link` and `useLink`, but does not remove it
- Adds `shouldPrefetch` as a new prop to both component and talon
- Ignores the value of `prefetchType` unless it is `true`

## Related Issue

Closes #3647

### Verification Stakeholders

- @jcalcaben 
- @luke-denton-aligent 

## Verification Steps

Links with either value of `prefetchType` should work the same as they do now. That said, we don't appear to have any.

## Breaking Changes (if any)

These changes should be non-breaking. Eventually, we should remove the deprecated prop, which would be a breaking change.

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#3647: Rename Link's prefetchType prop


supernova-at:
@magento create issue from PR

packages/venia-ui/lib/components/Link/link.js
```diff
@@ -3,38 +3,40 @@ import { Link as RouterLink } from 'react-router-dom';
...
+ * @property {bool} shouldPrefetch `true` activates prefetching the target page when the link becomes visible in the viewport.
  * @property {string} to From [react-router-dom Link](https://knowbody.github.io/react-router-docs/api/Link.html). The absolute path to the target page of the link. Uses the `to` prop from the `react-router-dom Link`.
  *
  * @example <caption>Basic usage</caption>
- * <Link prefetchType={true} to="/about/">About Us</Link>
+ * <Link shouldPrefetch={true} to="/about/">About Us</Link>
  */
 const Link = props => {
-    const talonProps = useLink(props);
+    // TODO: remove `prefetchType`
```
jeremycharron, line 16: Should this TODO be removed?

jeremycharron (commented):
@jimbo 
Code looks good except single comment about TODO. I added a minor version label + fixed the "closes issue" in description. Warning should be gone next checks run.

packages/venia-ui/lib/components/Link/link.js
jimbo: No, the prop still exists so as not to be a breaking change. We can actually remove it in a future major release, and delete the comment at that time.

jeremycharron (approved):
Code looks good. Approved.

ericeoeur:
QA has passed for this ticket and will be merged momentarily. Thanks!

PR #3674, opened 2022-01-27
[PWA-2549] Magento\GraphQl\Catalog\ProductSearchTest test fails

mikhaelbois:
## Description

Putting the filters sorting logic in the backend causes too many issues with core tests.
https://github.com/magento-commerce/magento2-pwa/pull/22 will remove the logic in the backend.
Here, we apply the logic in the frontend.

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-2549.

Filters should now be sorted as follows.
- Category filter is always first.
- Then filters are sorted alphabetically if they have the same position.
- Finally, they are sorted by position.
Validate in the Category and Search Pages on desktop and mobile.

## Checklist

### DIFF
--- a/magento-compatibility.js
+++ b/magento-compatibility.js
@@ -4,6 +4,7 @@
 
 // PWA Studio version -> Magento version.
 module.exports = {
+    '12.2.0': '2.4.3',
     '12.1.0': '2.4.3',
     '12.0.0': '2.4.3',
     '11.0.0': '2.4.3',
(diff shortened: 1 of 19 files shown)


dpatil-magento:
@mikhaelbois Category filter is always [first](https://develop.pwa-venia.com/default/search.html?query=tops+dress&page=1) (if exists) followed by [price](https://develop.pwa-venia.com/default/venia-tops.html?page=1) . Please see linked `develop` queries.  But in this pr price filter does not have any special treatment, expected?

mikhaelbois:
@dpatil-magento 
We give back the position to the price filter using https://github.com/magento-commerce/magento2-pwa/blob/main/CatalogGraphQlAux/Plugin/Price.php.
This way, `price` will use their `position` config instead of being forced second.

jcalcaben (approved):

PR #3694, opened 2022-02-07
Cypress updates

dpatil-magento:
## Description
  - Update cypress dependencies 
  - Update test titles
  - Update consolidated result output to have title mapped  (mochawesome has bug  which creates file itself with wrong file path so switched to test title)
  - Make test compatible with 2.4.4
  - Update PB breaking snaps

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-2553

## Verification Steps

 - Run cypress suite locally with --t options and verify its pass.
 - Observer test results and make sure no duplication in test title

## Checklist


dpatil-magento:
Intermittently tests failing with following error because the request is already made before even cy.wait starts waiting for it. No solution for now, need to debug in follow up ticket - 

> `cy.wait()` timed out waiting `60000ms` for the 1st request to the route: `gqlMergeCartsAfterAccountCreationMutation`. No request ever occurred.

mikhaelbois (dismissed):
LGTM
The only concern I have is the use of `cy.wait()`.
But if there's no way around it and a comment explains why it's used, it should be fine.

dpatil-magento:
@mikhaelbois Was able to make registered user test work with different gql query.

mikhaelbois (dismissed):

ericeoeur (approved):
LGTM

PR #3704, opened 2022-02-14
PWA-1895: workaround Prettier mangling arbitrary values in Tailwind

brendanfalkowski:
## Description

Prettier mangles Tailwind's (v2) arbitrary values syntax, which uses `,` for spaces. Prettier requires a space after commas, so it injects one, which breaks the class-generation in Tailwind.

This will wrap all rules broken by Prettier in the 1895 PR with `prettier-ignore` comments, so they can be addressed later, and not break the UI now.

In Tailwind (v3), the "space replacement" character was changed from `,` to `_` which avoids the Prettier mangling issue.

Future work in 1895 will involve updating to TW3, and we can fully resolve the issue.

## Related Issue

Workaround for a regression that Prettier introduced in PWA-1895.

### Verification Stakeholders

@jimbo 

## Checklist

### DIFF
--- a/packages/pwa-theme-venia/tw-issues.txt
+++ b/packages/pwa-theme-venia/tw-issues.txt
@@ -87,7 +87,7 @@ Tailwind plugins may need to be installed globally (not in package only).
 
 // ----------------------------------------------
 // B6. Using CSS Modules compose (like SCSS/Less "extends") breaks Tailwind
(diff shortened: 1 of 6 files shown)


jimbo (approved):
Looks good. Will fix after upgrading to Tailwind 3.

PR #3705, opened 2022-02-14
[PWA-2471] automate Venia Multi store test (Task 2 of 2)

mikhaelbois:
## Description

Add integration tests to the multi store test to validate other features.

## Related Issue

Closes https://jira.corp.magento.com/browse/PWA-2471.

- Run Cypress test `/src/tests/integrationTests/multiStore/storeSwitching.spec.js`

## Checklist


venia-integration-tests/src/tests/integrationTests/multiStore/storeSwitching.spec.js
```diff
@@ -230,6 +223,61 @@ const interceptAccessoriesPageRequests = expectedStoreCode => {
     }).as('getMockFilterInput');
 };
 
+/**
+ * Setup network intercepts that return mocked data for requests
+ * specific to the tops category page
+ *
+ * @param {String} expectedStoreCode
+ */
+const interceptTopsPageRequests = expectedStoreCode => {
```
pedrochiossi, line 232: I think we can make these intercept setup functions more dynamic if we add a second parameter for the expected category or expected product to avoid code repetition, something like: 
```suggestion
const interceptCategoryPageRequests =  (expectedCategory, expectedStoreCode) => {
     cy.intercept('GET', getCategoriesCall, getInterceptHandler(expectedCategory, expectedStoreCode)
     .as('getMockCategories');
```
and so forth. See if we can apply this to the other ones.

venia-integration-tests/src/tests/integrationTests/multiStore/storeSwitching.spec.js
```diff
@@ -230,6 +223,61 @@ const interceptAccessoriesPageRequests = expectedStoreCode => {
...
+    }).as('getMockFilterInput');
+};
+
+/**
+ * Setup network intercepts that return mocked data for requests
+ * specific to the product 1 product page
+ *
+ * @param {String} expectedStoreCode
+ */
+const interceptProduct1PageRequests = expectedStoreCode => {
```
pedrochiossi, line 269: Same thing from above comment, see if we can reduce some code by adding a second parameter like `expectedProduct`

pedrochiossi (changes requested):
@mikhaelbois I've added a suggestion to avoid some repeating code. Apart from that all good.

venia-integration-tests/src/tests/integrationTests/multiStore/storeSwitching.spec.js
mikhaelbois: I've reworked the code and simplified things a bit like you suggested.

pedrochiossi (approved):

jeremycharron:
run cypress

jeremycharron:
run cypress

jeremycharron:
All tests are passing locally and in PR, except `verify block content` that is not related and also occurring on other PRs.
No issue detected.
QA passed ✅

PR #3706, opened 2022-02-15
[PWA-2554] - Long wishlist name length breaks layout

jasonrpoteet:
## Description

Fixed overflow of text for custom wishlist name

## Related Issue

Closes PWA-2554

### Verification Stakeholders

Project Maintainers

## Verification Steps

Wishlist should account for custom names that are too long for the viewable UI.

1. Log into Venia
2. Navigate to "Favorite List"
3. Add a custom  favorite list name that is super long
4. Notice that the name should now truncate when it reaches the end of the line, but will have a hover over with the full name

N/A

1. Verify that new and existing Jest tests pass
1. Verify that Cypress test passes

Yes

No

## Breaking Changes (if any)

No breaking changes

## Checklist


mikhaelbois:
run cypress

mikhaelbois:
Regarding the Cypress tests that are failing:
- Page Builder Column is missing images compared to the snapshot.
- Page Builder Map is different because Google seems to have changed their font color.
A follow-up task will be created to fix those.

QA ✅

PR #3709, opened 2022-02-16
Image not showing

MinhBuiCong:
Image tag from pagebuilder won't show in frontend. 
Image tag has "data-set" attribute when getting from pagebuilder, it is "src" attribute written the code file. 
We are getting wrong attribute, so we had to override this "configAggregator.js" to change ".getAttribute("src")" to ".getAttribute("data-src")" for image tag to show.

### Resolved issues:
1. [x] resolves magento/pwa-studio#3718: Image not showing


anthoula:
@magento create issue from PR

anthoula (changes requested):
@MinhBuiCong - Hello, thanks for the contribution. Would you resolve the conflicting file?

PR #3717, opened 2022-02-24
PWA-260: Make QuantityStepper @api and deprecate ProductQuantity

rwalter215:
## Description

* Deprecated QuantityFields, LegacyMiniCart
* Created QuantityStepper (to replace QuantityFields) component
* Updated imports of QuantityFields to QuantityStepper
* Updated unit and integration test ids

## Related Issue

Closes PWA-260.

## Acceptance
PM approval required (PM has approved)

## Verification Steps
Test Plan detailed in: https://jira.corp.magento.com/browse/PWA-960

No

## Breaking Changes (if any)
There shouldn't be

## Checklist


packages/venia-ui/lib/components/CartPage/ProductListing/quantity.js
```diff
@@ -11,6 +11,9 @@ import TextInput from '../../TextInput';
 import { Message } from '../../Field';
 import defaultClasses from './quantity.module.css';
 
+/**
+ * @deprecated - use components/QuantityStepper instead
```
tjwiebell, line 15: This is on the right track, but we can additionally get rid of this duplicate code now. We had a bunch of this when we migrated graphql operations from venia-ui to peregrine, and just in case a third-party dev was using them, we used this re-export and alias pattern.

https://github.com/magento/pwa-studio/blob/develop/packages/venia-ui/lib/components/ProductFullDetail/productFullDetail.gql.js

Would expect something like `export { default as QuantityFields } from '../../QuantityStepper'`. This probably means you can also cleanup style and test files, since those are redundant as well.

packages/venia-ui/lib/components/QuantityStepper/QuantityStepper.js
```diff
@@ -0,0 +1,94 @@
+import React, { Fragment } from 'react';
+import { useIntl } from 'react-intl';
+import { Minus as MinusIcon, Plus as PlusIcon } from 'react-feather';
+import { useQuantity } from '@magento/peregrine/lib/talons/CartPage/ProductListing/useQuantity';
+
```
tjwiebell, line 5: Will need to move this talon as well, and apply the re-export and alias pattern.

`export { useQuantityStepper as useQuantity } from '../QuantityStepper/useQuantityStepper'`

packages/venia-ui/lib/components/QuantityStepper/QuantityStepper.js
```diff
@@ -0,0 +1,94 @@
+import React, { Fragment } from 'react';
```
tjwiebell, line 1: Wouldn't be an inaugural code review without some trivial naming convention feedback; for consistency, suggest we switch filenames to camelCase, eg. `quantityStepper.js`.

packages/venia-ui/lib/components/LegacyMiniCart/miniCart.js
```diff
@@ -10,6 +10,9 @@ import Mask from './mask';
 import defaultClasses from './miniCart.module.css';
 import { useStyle } from '../../classify';
 
+/**
+ * @deprecated
```
tjwiebell, line 14: Keep up with the deprecation descriptions, and linking to what it was replaced with 👍

tjwiebell (changes requested):
Added some clarifications and suggestions for code consistency. Let me know if you have any questions, or want to pair on any of this.

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.js
```diff
@@ -94,7 +10,7 @@ const Quantity = props => {
                 quantity: props.initialValue
             }}
         >
-            <QuantityFields {...props} />
+            <QuantityStepper {...props} />
```
rwalter215, line 13: @tjwiebell I had some issues implementing the re-export and alias pattern for QuantityFields in this file, since I needed to use that component here, as well as export it. This was my idea for a solution, but let me know if there is a better way to tackle this.

rwalter215: * You will need to look at full file to see my approach

ericeoeur:
@rwalter215 great job! Just a two comments:

- Include a version label on this PR. My guess it is _minor_ 
- Please run _prettier_ and recommit 

Once that is done, this is LGTM and ready for QA. ✅

tjwiebell:
run lighthouse-mobile

tjwiebell:
run lighthouse-desktop

tjwiebell:
run all tests

tjwiebell:
run pr-deploy

packages/venia-ui/lib/components/CartPage/ProductListing/quantity.js
tjwiebell: Re-exporting `QuantityStepper` should make the component available for use by the default `Quantity` export. I would try this, and also go ahead and add deprecations to both of these exports.

```js
/**
 * @deprecated - moved to component directory in 12.4.0
 * @see QuantityStepper
 */
export { default as QuantityFields } from '../../QuantityStepper';
```

tjwiebell (changes requested):
Tiny suggestion, but I would accept the current solution with added @deprecated tags. Nice work 🎉

rwalter215:
@tjwiebell I added the deprecations, but when I tried the proposed solution, I am getting `Error QuantityFields is undefined`

Here is what my file has when it get's the error. 

`import React from 'react';
import { Form } from 'informed';
import { func, number, string } from 'prop-types';

/**
 * @deprecated - moved to component directory in 12.4.0
 * @see QuantityStepper
 */
export { default as QuantityFields } from '../../QuantityStepper';

const Quantity = props => {
    return (
        <Form
            initialValues={{
                quantity: props.initialValue
            }}
        >
            <QuantityFields {...props} />
        </Form>
    );
};

Quantity.propTypes = {
    initialValue: number,
    itemId: string,
    label: string,
    min: number,
    onChange: func,
    message: string
};

Quantity.defaultProps = {
    label: 'Quantity',
    min: 0,
    initialValue: 1,
    onChange: () => { }
};

export default Quantity;
`

tjwiebell (approved):

tjwiebell:
> @tjwiebell I added the deprecations, but when I tried the proposed solution, I am getting `Error QuantityFields is undefined`
I tested this assumption poorly, missed the `But where function1 and function2 do not become available inside the current module.` section of the [docs](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#re-exporting_aggregating) 🤦 

Looks good, will merge after builds.

tjwiebell:
run all tests

tjwiebell:
run lighthouse-mobile

tjwiebell:
lighthouse-mobile scores are within range of develop.pwa-venia.com, @dpatil-magento will merge.

PR #3724, opened 2022-02-28
PWA-2610: Update Magento capability table

michaelyu0123:
Adds 12.3.0 entry in compatibility table

Closes https://jira.corp.magento.com/browse/PWA-2610


anthoula (dismissed):

magento-compatibility.js
```diff
@@ -4,6 +4,7 @@
 
 // PWA Studio version -> Magento version.
 module.exports = {
+    '12.3.0': '2.4.4',
```
dpatil-magento, line 7: @michaelyu0123  Current approach is causing confusion as it does not talk about magento2-pwa. It would be nice to mention both.  Following is just an example . @supernova-at Please throw in better ways to update :) 

 12.3.0: {
    magento: "2.4.4",
    magento2-pwa : "0.2.0",
    magento2-pwa-commerce: "0.0.1"
  },

michaelyu0123: since this file will be consumed by other processes in the pipeline, not sure if this change is compatible with that. Perhaps should be done in a future ticket?

dpatil-magento: @michaelyu0123 I will create follow up ticket.  For now just change 2.4.4 ot 2.4.3 and we should be good here.

michaelyu0123:
run all tests

magento-compatibility.js
michaelyu0123: @dpatil-magento updated as per change to requirement

michaelyu0123:
run all tests

dpatil-magento (approved):

PR #3730, opened 2022-03-04
[Bugfix] [Refactor] Order confirmation page

jcalcaben:
## Description

Fixes a bug with the order confirmation page where data would not persist between a page refresh.

NOTE: This has been fixed for authenticated users, but not for guest users.

This PR also refactors the order confirmation page component to be less coupled with the checkout component.

## Related Issue

Closes PWA-2484

## Acceptance

Any core developer

### Verification Stakeholders

Any core developer

## Verification Steps

1. Go through checkout as a logged in user
2. On the order confirmation page, refresh the page
3. Verify order data is still rendered

1. Verify guest users can still go through checkout steps
2. Verify authenticated users can go through checkout steps

Authenticated users logs out and guest user tries to navigate back to the order confirmation page.

## Checklist


drpayyne:
Kind of related so I'll ask - is there any plan to sort the customer orders by most recent date? I believe it's not implemented yet. I'm aware that it's in the scope of the backend, but since work is being done on this page, I got curious. Thanks.

jcalcaben:
@drpayyne if there isn't a GitHub issue for it already, would you create one for your request about sorting the customer orders?

haal9000:
run pr-deploy

PR #3745, opened 2022-03-22
changed the reference of base docker images from dockerhub to jfrog

haal9000:
## Description

## Related Issue

Closes #ISSUE_NUMBER.

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#3751: changed the reference of base docker images from dockerhub to jfrog

### DIFF
--- a/venia-integration-tests/run-tests.js
+++ b/venia-integration-tests/run-tests.js
@@ -84,14 +84,14 @@ if (port) {
 
     dockerCommand = `docker run --rm -v ${
         process.env.PWD
-    }:/venia-integration-tests -w /venia-integration-tests --entrypoint=cypress cypress/included:8.3.1 run --browser chrome --config baseUrl=https://host.docker.internal:${port},screenshotOnRunFailure=false --config-file cypress.config.json --env updateSnapshots=${update} --headless --reporter mochawesome --reporter-options reportDir=cypress/results,overwrite=false,html=false,json=true`;
+    }:/venia-integration-tests -w /venia-integration-tests --entrypoint=cypress docker-hub-remote.dr-uw2.adobeitc.com/cypress/included:8.3.1 run --browser chrome --config baseUrl=https://host.docker.internal:${port},screenshotOnRunFailure=false --config-file cypress.config.json --env updateSnapshots=${update} --headless --reporter mochawesome --reporter-options reportDir=cypress/results,overwrite=false,html=false,json=true`;
 } else {
     // run docker on remote instance
     console.log(`Running tests on remote instance ${baseUrl}`);
 
     dockerCommand = `docker run --rm -v ${
         process.env.PWD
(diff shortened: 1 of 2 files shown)


venia-integration-tests/run-tests.js
```diff
@@ -84,14 +84,14 @@ if (port) {
 
     dockerCommand = `docker run --rm -v ${
         process.env.PWD
-    }:/venia-integration-tests -w /venia-integration-tests --entrypoint=cypress cypress/included:8.3.1 run --browser chrome --config baseUrl=https://host.docker.internal:${port},screenshotOnRunFailure=false --config-file cypress.config.json --env updateSnapshots=${update} --headless --reporter mochawesome --reporter-options reportDir=cypress/results,overwrite=false,html=false,json=true`;
+    }:/venia-integration-tests -w /venia-integration-tests --entrypoint=cypress docker-hub-remote.dr-uw2.adobeitc.com/cypress/included:8.3.1 run --browser chrome --config baseUrl=https://host.docker.internal:${port},screenshotOnRunFailure=false --config-file cypress.config.json --env updateSnapshots=${update} --headless --reporter mochawesome --reporter-options reportDir=cypress/results,overwrite=false,html=false,json=true`;
```
dpatil-magento, line 87: Local vs CI case needs to be handled.

venia-integration-tests/run-tests.js
```diff
@@ -84,14 +84,14 @@ if (port) {
...
 } else {
     // run docker on remote instance
     console.log(`Running tests on remote instance ${baseUrl}`);
 
     dockerCommand = `docker run --rm -v ${
         process.env.PWD
-    }:/venia-integration-tests -w /venia-integration-tests --entrypoint=cypress cypress/included:8.3.1 run --browser chrome --config baseUrl=${baseUrl},screenshotOnRunFailure=false --config-file cypress.config.json --env updateSnapshots=${update} --headless --reporter mochawesome --reporter-options reportDir=cypress/results,overwrite=false,html=false,json=true`;
+    }:/venia-integration-tests -w /venia-integration-tests --entrypoint=cypress docker-hub-remote.dr-uw2.adobeitc.com/cypress/included:8.3.1 run --browser chrome --config baseUrl=${baseUrl},screenshotOnRunFailure=false --config-file cypress.config.json --env updateSnapshots=${update} --headless --reporter mochawesome --reporter-options reportDir=cypress/results,overwrite=false,html=false,json=true`;
```
dpatil-magento, line 94: Local vs CI case needs to be handled.

sharkySharks:
@haal9000 I don't know that we want to be changing this to use the adobe internal artifactory instance because this docker file can be used externally to build pwa as well and this will break for public users since they do not have access. I think we should keep this generic to docker hub public images, but update the credentials for the user in our CI environment. Not sure we even need to use credentials for public images...

sharkySharks:
run lighthouse-desktop

haal9000:
> @haal9000 I don't know that we want to be changing this to use the adobe internal artifactory instance because this docker file can be used externally to build pwa as well and this will break for public users since they do not have access. I think we should keep this generic to docker hub public images, but update the credentials for the user in our CI environment. Not sure we even need to use credentials for public images...

I believe docker has introduced some throttling on the amount of data we can pull from public docker hub without having an account(ref: https://magento.slack.com/archives/C1XJ71VN3/p1608143019296000). If we have to use public docker repo then we might need a generic paid account whose credentials we can use in our CI.

anthoula:
@magento create issue from PR

PR #3757, opened 2022-04-11
CIF-2826: support setting full graphql endpoint in useAdapter

Buuhuu:
## Description

This changes allows users of Peregrine `useAdapter()` to specify the full graphql endpoint to use. 

In the context of AEM the Commerce Integration Framework uses Peregrine based components to accelerate development of transactional frontend components such as Cart, Checkout etc. In this setup a configuration maintained in AEM defines the endpoint the graphql server is running on. This endpoint is different from the default `<origin>/graphql` as it is used for "pure" Adobe Commerce deployments. Customers can even freely define the endpoint. 

In the default setup a reverse proxy is used to route the graphql traffic to the Adobe Commerce backend. This proxy is per default listening on `/api/graphql`. However, as said above customers could use any path, or even an absolute URL to an API gateway for example.

In the past we only supported Peregrine 11.x without using `useAdapter()` as we set up the context for peregrine manually (ApolloClient, Redux). This is not possible with 12.x anymore due to [`clearCacheData()`](https://github.com/magento/pwa-studio/blob/v12.3.0/packages/peregrine/lib/talons/Adapter/useAdapter.js#L272) being added to the apollo client instance, which is for example used in [SignIn](https://github.com/magento/pwa-studio/blob/v12.3.0/packages/peregrine/lib/talons/SignIn/useSignIn.js#L85-L86). The above described change is necessary for use to make use of `useAdapter()` and so pass those internal function references down to the components. 

## Related Issue

CIF-2826

## Verification Steps

The change is minor, full unit test coverage is provided and should be sufficient to verify this change. 

If manual verification is needed please ping me.

## Checklist

- [X]  I have added tests to cover my changes, if necessary.
- [ ] I have added translations for new strings, if necessary.


fooman:
Wouldn’t an environment variable be the better option here as the endpoint will likely be different between dev vs prod for example?

On another note would have loved to see the useAdapter provide an intercept target so that the whole Apollo client links can be customised for example to introduce persisted queries by endpoint etc.

Buuhuu:
> Wouldn’t an environment variable be the better option here as the endpoint will likely be different between dev vs prod for example?

This may additionally make sense, yes. However for our AEM + CIF use case the endpoint is a runtime variable that editors may change. They may for example use the same storefront implementation (AEM Components + Peregrine based components + styles) for different stores where each of the stores may have a different endpoint url. 

> On another note would have loved to see the useAdapter provide an intercept target so that the whole Apollo client links can be customised for example to introduce persisted queries by endpoint etc.

Sounds reasonable as well, but it is beyond the scope of what we want to achieve with this change.

jimbo (approved):
This looks fine as written. I agree with @fooman that we need to add targets so that people can add or modify links, but that doesn't need to be part of this PR.

Added a couple of style suggestions, you should be able to just click `commit` on them.

Buuhuu:
Thanks Jimmy, I applied the suggestions.

PR #3763, opened 2022-04-14
passing jfrog image only for CI build stage to overcome docker pull l…

haal9000:
## Description
Pulling from public docker hub causing rate limit exceptions

## Related Issue

Closes JIRA-PWA-2085

## Acceptance

Any developer
Devops engineer

## Verification Steps

1. Go through the codebuild logs to check if the public image has been replaced with jfrog image
2. Build job should not fail due to rate docker pull limit error

### DIFF
--- a/venia-integration-tests/run-tests.js
+++ b/venia-integration-tests/run-tests.js
@@ -77,7 +77,7 @@ const dockerRuns = {};
 const port = new URL(baseUrl).port;
 
 let dockerCommand = null;
-
+let dockerImage = "cypress/included:8.3.1";
 if (process.env.CI) {
     dockerImage = process.env.DockerRegistry + dockerImage
 };
--- a/venia-integration-tests/run-tests.js
+++ b/venia-integration-tests/run-tests.js
@@ -78,7 +78,7 @@ const port = new URL(baseUrl).port;
 
 let dockerCommand = null;
 let dockerImage = "cypress/included:8.3.1";
-if (process.env.CI) {
+if (process.env.DockerRegistry) {
     dockerImage = process.env.DockerRegistry + dockerImage
 };
 
--- a/venia-integration-tests/run-tests.js
+++ b/venia-integration-tests/run-tests.js
@@ -78,10 +78,9 @@ const port = new URL(baseUrl).port;
 
 let dockerCommand = null;
 
-let dockerImage = "cypress/included:8.3.1";
 if (process.env.CI) {
-    dockerImage = "docker-hub-remote.dr-uw2.adobeitc.com/" + dockerImage;
-}
+    dockerImage = process.env.DockerRegistry + dockerImage
+};
 
 if (port) {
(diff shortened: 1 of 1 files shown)


venia-integration-tests/run-tests.js
```diff
@@ -78,20 +78,25 @@ const port = new URL(baseUrl).port;
 
 let dockerCommand = null;
 
+let dockerImage = "cypress/included:8.3.1";
+if (process.env.CI) {
+    dockerImage = "docker-hub-remote.dr-uw2.adobeitc.com/" + dockerImage;
```
tjwiebell, line 83: Will this address ever change? Does this file have access to the same environment variable `$JFROG_REPO` like the cicd scripts?

tjwiebell (approved):
Looks fine to me, just a question about sourcing value from an environment variable. I would also get an approval from another DevOps engineer if possible.

venia-integration-tests/run-tests.js
```diff
@@ -78,20 +78,24 @@ const port = new URL(baseUrl).port;
 
 let dockerCommand = null;
 
+if (process.env.CI) {
```
sharkySharks, line 81: we still want to leave in the dockerImage above this line: `let dockerImage = "cypress/included:8.3.1";`

sharkySharks: let's also set a default here for the DockerRegistry. if `process.env.DockerRegistry` is not set then it will default to `undefined` which we don't want, so maybe instead of checking for `process.env.CI` we can check for `process.env.DockerRegistry` and then set the value of `dockerImage` to that value. I think `process.env.CI` is a bit too broad since many scripts will use that env var in CI environments but it won't necessarily indicate that a docker registry should be set.

OneCricketeer: > let's also set a default here for the DockerRegistry

Just an empty string? Preserve existing behavior?
```js
dockerImage = (process.env.DockerRegistry || '') + dockerImage
```

sharkySharks: also a valid way of doing it. this is the same logic as is currently in place in this pr.

sharkySharks:
run all tests

sharkySharks:
run all tests

sharkySharks:
run cypress

sharkySharks:
run pr-test

sharkySharks (approved):

venia-integration-tests/run-tests.js
```diff
@@ -77,21 +77,25 @@ const dockerRuns = {};
 const port = new URL(baseUrl).port;
 
 let dockerCommand = null;
+let dockerImage = 'cypress/included:8.3.1';
+if (process.env.DockerRegistry) {
```
fooman, line 81: All other uses of process.env are capitalized in this repo. Might be a good idea to keep it consistent.

PR #3773, opened 2022-04-20
[PWA-2742] Eventing framework

jimbo:
## Description

Add a minimal eventing framework to PWA Studio.

- Add a context provider with an API for dispatching and subscribing to a stream of events
- Add a sample extension that demonstrates how to use this API

## Related Issue

PWA-2742

### Verification Stakeholders

- @supernova-at 

1. Look for `Logging event: hello world` in the console after app load

No existing features impacted

No backend involvement

No

No

## Breaking Changes (if any)

No

## Checklist


packages/peregrine/lib/context/eventing.js
```diff
@@ -0,0 +1,57 @@
+import React, {
+    createContext,
+    useCallback,
+    useContext,
+    useMemo,
+    useRef
+} from 'react';
+import { Observable } from 'zen-observable-ts';
+
+const EVENT_NAME = 'beacon';
```
jimbo, line 10: This could be a prop, but due to the way [we render our context providers programmatically](https://github.com/magento/pwa-studio/blob/develop/packages/peregrine/lib/PeregrineContextProvider/peregrineContextProvider.js#L27-L29), there's no convenient way for people to actually supply a prop to `EventingContextProvider`.

packages/peregrine/package.json
```diff
@@ -17,7 +17,8 @@
   },
   "homepage": "https://github.com/magento/pwa-studio/tree/main/packages/peregrine#readme",
   "dependencies": {
-    "fast-glob": "~3.2.4"
+    "fast-glob": "~3.2.4",
+    "zen-observable-ts": "^1.2.0"
```
jimbo, line 21: This is not a new library. It's a dependency of Apollo Client, so we already use it indirectly.

There are better Observable implementations out there (eg, `rxjs`) but they tend to be huge. This one's free.

packages/venia-concept/package.json
```diff
@@ -56,6 +56,7 @@
     "@magento/peregrine": "~12.3.0",
     "@magento/pwa-theme-venia": "~1.2.0",
     "@magento/upward-security-headers": "~1.0.7",
+    "@magento/venia-sample-eventing": "0.0.1",
```
jimbo, line 59: Do we normally list our sample extensions here? Looks like not, actually.

packages/peregrine/package.json
jimbo: That said, we ought to upgrade Apollo-related dependencies so they all resolve to the same version of this library.

packages/extensions/venia-sample-eventing/main.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default original => props => {
+    const [observable, { dispatch }] = useEventingContext();
+
+    useEffect(() => {
+        observable.subscribe(event => {
+            console.log('Logging event:', event);
+        });
+
+        dispatch('hello world');
```
jimbo, line 12: We'll be doing most of the dispatching in `peregrine`, but it's possible to do it from an extension.

packages/extensions/venia-sample-eventing/main.js
```diff
@@ -0,0 +1,16 @@
+import { useEffect } from 'react';
+import { useEventingContext } from '@magento/peregrine/lib/context/eventing';
+
+export default original => props => {
+    const [observable, { dispatch }] = useEventingContext();
+
+    useEffect(() => {
+        observable.subscribe(event => {
+            console.log('Logging event:', event);
+        });
```
jimbo, line 10: Should probably return a function from this effect that cleans up this subscription.

packages/extensions/venia-sample-eventing/main.js
```diff
@@ -0,0 +1,16 @@
+import { useEffect } from 'react';
+import { useEventingContext } from '@magento/peregrine/lib/context/eventing';
+
+export default original => props => {
+    const [observable, { dispatch }] = useEventingContext();
+
+    useEffect(() => {
```
jimbo, line 7: Everything in this effect should probably be lazy-loaded.

packages/peregrine/lib/context/eventing.js
```diff
@@ -0,0 +1,57 @@
...
+
+                document.addEventListener(EVENT_NAME, event => {
+                    observer.next(event.detail);
+                });
+            }),
+        []
+    );
+
+    const dispatch = useCallback(detail => {
+        const event = new CustomEvent(EVENT_NAME, { detail });
```
supernova-at, line 32: This `dispatch` function supports both but when we `dispatch` from the framework are we just indicating that an event happened (ex: "signed in!"), or are we populating a bunch of contextual data along with it?

packages/peregrine/lib/context/eventing.js
```diff
@@ -0,0 +1,57 @@
...
+
+        cacheRef.current.push(detail);
+        document.dispatchEvent(event);
```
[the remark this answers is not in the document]
jimbo: Convenience. The document (as an `EventTarget`) is basically just a global context we can use for free. Honestly, we could pretty easily just skip the observable and context, put the cache directly on `document`, and export a couple of static functions:

- `dispatch`, which would `document.eventCache.push` and `document.dispatchEvent`
- `subscribe`, which would `document.eventCache.forEach` and `document.addEventListener`

The result would be the same here. I think this approach gives us little bit of framework to expand on if we need to, but I'm open to boiling it down.

packages/peregrine/lib/context/eventing.js
jimbo: I would think of these events as similar to Redux actions.

```es6
dispatch({
  type: "PAGE_VIEW",
  payload: {
    categoryId: "MTM=",
    pageType: "CATEGORY",
    pathname: "/default/venia-dresses.html",
    query: "?page=1"
  }
})
```

michaelyu0123 (approved):
LGTM

michaelyu0123:
run all tests

jcalcaben:
run all tests

michaelyu0123:
run all tests

PR #3805, opened 2022-04-27
AC-2480::Flyout panels do not trap focus while open

glo82145:
## Description

Flyout panels do not trap focus while open

Environment
Adobe Magento - Venia

Context
Windows 10; Chrome 88

Reproduction Steps
Locations (representative sample):

Global Header
Search Results
1. Using keyboard only, from a search results page activate the "Filter" button.
2. Press the Tab key to navigate to the Filters panel. Note that focus lands on elements obscured by the modal overlay.

Actual Behavior
The mobile site menu (globally) and Filters panel (on Search/list pages) add an overlay to the rest of the page content when they are expanded, but keyboard focus is not trapped within them while they are open. As a result, keyboard focus lands on elements that are obscured beneath the overlay.

Expected Behavior
When a content panel such as the mobile navigation is open and an overlay obscures other content on the screen, the dialog should be marked up as and behave as a modal dialog.

Ensure that when these panels are open and an overlay obscures the rest of the page content, keyboard focus is trapped within the panel.

Please refer to WAI-ARIA Authoring Practices Dialog (Modal) design pattern example:

https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal

## Related Issue

Closes https://jira.corp.magento.com/browse/AC-2480

## Verification Steps

✔️ QA Passed
Pre-Conditions:

1. Have Magento instance with sample data installed

2. Make sure to have pwa studio installed

3. Make sure to have a customer login for front end login

 

Manual Steps executed:

Login to venia > Navigate to global header accessories or other
Navigate to main menu and check focus navigating to the options given 
Navigate to  filter button and check focus navigating to the options given 
 

✖️ Behaviour Before The Fix : The keyboard focus is not trapped within the main menu while they are open.

✔️Behaviour After The Fix: The keyboard focus is getting trapped within the main menu while they are open.

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#3823: AC-2480::Flyout panels do not trap focus while open


supernova-at:
@magento create issue from PR

rwalter215 (changes requested):
After checking the deployment of this PR branch (https://pr-3805.pwa-venia.com/), I am not seeing the expected behavior from the mobile navigation menu. Specifically it does not focus the menu, as soon as it is opened. Separately, these changes cause an issue, once the focus is on an element of the mobile navigation menu, the focus cannot escape the mobile navigation menu, even when it is closed. Which breaks keyboard functionality for the rest of the application, until the user refreshes the browser. 

This may be because the mobile navigation menu is always present on the dom, since when it is closed, the `opacity` and `visibility` are set in order to hide it (but not remove it). I believe this is what is giving us undesirable results from the FocusScope component from react-aria. Using the `display` property, instead of `opacity` and `visibility` will hide the menu and remove from dom, or possibly using a React Portal will fix this issue.

rwalter215 (changes requested):
The focus functionality works! Unfortunately my recommendation broke the transition functionality. It works with opacity and visibility but breaks with display. So we need to find a solution that allows us to do both, trap focus properly and allow the component to transition into view properly. I will continue to look into good ways of doing this, but a good place to start may be to keep the opacity and visibility way of showing/hiding the component and imperatively handling the focus. Or perhaps react-aria has a hook that would be useful here.

glo82145:
> The focus functionality works! Unfortunately my recommendation broke the transition functionality. It works with opacity and visibility but breaks with display. So we need to find a solution that allows us to do both, trap focus properly and allow the component to transition into view properly. I will continue to look into good ways of doing this, but a good place to start may be to keep the opacity and visibility way of showing/hiding the component and imperatively handling the focus. Or perhaps react-aria has a hook that would be useful here.

@rwalter215  .. I have implemented a fix and verified on the instance created in this PR. Could you please check and let me know

rwalter215:
React Portal seems like a good solution for this issue, both focus and transition functionality work. I don't believe `tabindex` is necessary though. When I pulled up the deployed storefront for this PR (https://pr-3805.pwa-venia.com/) and removed all the tabindex that this PR added, the functionality still worked as expected. Also setting them all to `1` is not best practice, as `tabindex` is meant to explicitly set the order that the tab functionality should follow (ex. 1, 2, 3, 4, etc.). Please remove tabindex from this PR, confirm the functionality still works as expected and I will approve PR. Thank you!

glo82145:
> React Portal seems like a good solution for this issue, both focus and transition functionality work. I don't believe `tabindex` is necessary though. When I pulled up the deployed storefront for this PR (https://pr-3805.pwa-venia.com/) and removed all the tabindex that this PR added, the functionality still worked as expected. Also setting them all to `1` is not best practice, as `tabindex` is meant to explicitly set the order that the tab functionality should follow (ex. 1, 2, 3, 4, etc.). Please remove tabindex from this PR, confirm the functionality still works as expected and I will approve PR. Thank you!

@rwalter215  .. I have removed all tabIndex .. Please have a look now

rwalter215 (dismissed):

dpatil-magento:
run pr-deploy

dpatil-magento:
run cypress

dpatil-magento:
@glo82145  On Mobile view after accessing and closing Filter model , user is unable to interact with search field. Following cypress failure is related to this.

```
user should be able to filter results in Category and Search pages:
CypressError: cy.type() failed because it targeted a disabled element.
The element typed into was:

Ensure the element does not have an attribute named disabled before typing into it.

```

glo82145:
@rwalter215 .. Could you please suggest any solution for this issue .. 

cc: @dpatil-magento

glo82145:
@dpatil-magento  We have tried to reproduce the functionality manually on deployed version and We are not able to replicate the issue.
Could you please guide us more on it 

cc: @anthoula

dpatil-magento:
run cypress

glo82145:
run cypress

glo82145:
@dpatil-magento We have fixed the issue.. Could you please check

dpatil-magento:
run cypress

dpatil-magento:
run lighthouse-desktop

dpatil-magento:
run lighthouse-mobile

anthoula (approved):

dpatil-magento:
QA Approved.

sanganinamrata:
I've tested this PR in development branch & got attached console Warning:
Reason: Attributes in reactjs follow camelCase. Hence this warning is coming. So need to update all the "tabindex" to "tabIndex" in this PR.

Screen Shot 2022-08-25 at 12 49 37 PM

PR #3813, opened 2022-04-29
[TBD] Fix bug in sub-domain generation

tjwiebell:
## Description

`buildpack create-custom-origin .` uses the current directory as the seed for a unique hash, but the existing logic could result in an invalid sub-domain value.

## Related Issue

Closes #ISSUE_NUMBER.

1. Scaffold or clone into known bad path (`/path/to/fake/cwd` appears to work)
2. Run `yarn buildpack create-custom-origin .`
3. Verify that valid host and sub-domain are generated

## Checklist


glo82145:
Closing as this is very old PR.

PR #3819, opened 2022-05-04
Resolve Apollo dependencies

jimbo:
## Description

Resolve `@apollo/client` and `zen-observable-ts` to a single version across our packages. We should keep bundle size down by not serving multiple versions of a library.

## Related Issue

Follow-up to PWA-2742.

### Verification Stakeholders

@dpatil-magento 

## Verification Steps

```
yarn why @apollo/client
yarn why zen-observable-ts
yarn stats:venia
```

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,8 +91,10 @@
     "sharp": "~0.29.0"
   },
   "resolutions": {
+    "@apollo/client": "~3.5.0",
+    "@graphql-tools/prisma-loader": "6.3.0",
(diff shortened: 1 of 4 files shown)


fooman:
I have added PR #3889 as we may as well try to upgrade Apollo while we are at it. I believe my pull request has accumulated the commits from develop. That should hopefully resolve if and when this PR gets the latest commits from develop merged.

PR #3842, opened 2022-05-05
Make Apollo links customizable

jimbo:
## Description

Allow Apollo links to be customized by a `configureLinks` prop to `Adapter` / `useAdapter`. Also, while we're at it, extract the links into their own modules.

This can either follow #3757 or we can combine them here.

### New component/talon signature

```jsx
function configureLinks(linksMap) {
  // receive a map
  // return either a map or array, with modifications
  return Array.from(linksMap.values())
}

const tree = (
  <Adapter
    apiBase="https://example.com/graphql"
    configureLinks={configureLinks}
    store={store}
    styles={styles}
  />
)
```

### New package structure

```
peregrine
|- lib
  |- Apollo
    |- links
      |- authLink.js
      |- errorLink.js
      |- gqlCacheLink.js
      |- index.js
      |- mutationQueueLink.js
      |- retryLink.js
      |- storeLink.js
```

## Related Issue

CIF-2826

### Verification Stakeholders

- @dpatil-magento 
- @tjwiebell 
- @Buuhuu 

## Checklist

-   I have added tests to cover my changes, if necessary.
-   I have added translations for new strings, if necessary.
-   I have updated the documentation accordingly, if necessary.

### Resolved issues:
1. [x] resolves magento/pwa-studio#3847: Make Apollo links customizable

PR #3861, opened 2022-05-19
fix: remove caniuse-lite from dependency list

tjwiebell:
## Description

In #3842 we finally fixed the repeated `Browserslist: caniuse-lite is outdated. Please run: npx browserslist@latest --update-db` message, but the dependency was left behind in the `package.json`. If left this way, every time the browserlist is updated (quite often), our lock file will get out of sync. Running this script should add -> install -> remove this dependency so its only in the lock file, so I propose removing this dependency, and running that command as needed. I _hope_ this doesn't mean that message comes back as soon as we're one patch behind, but we can keep it the way it is if we'd prefer frequent updates to the lock over this message. 

## Related Issue

Closes #NOJIRA

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,9 +45,6 @@
     "watch:all": "node scripts/watch-all.js",
     "watch:venia": "yarn venia run watch"
   },
-  "dependencies": {
-    "caniuse-lite": "~1.0.30001335"
-  },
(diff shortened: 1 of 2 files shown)


glo82145:
Closing this as it is very old PR

PR #3882, opened 2022-06-15
AC-3522::Product search result with items count is not being read by …

glo71317:
…screen reader

## Description

product search result with items count is not being read by screen reader

**Pre Conditions:**

Have Venia setup working with Magento 2.4.5-alpha2 in the local.
Have sample data available in the instance
Have Google chrome screen reader extension enabled and working.
**Steps to reproduce:**

Keeping extension enabled, open the venia front end in the chrome browser.
Tab the focus to search button in the home page.
Press return key from the keyboard and input with the text that results with items found. (ex: carina)
Expected: Screen reader should announce the total items found in the searched result.

**Actual**: Randomly announcing the text from left side filters options sometimes or sort items dropdown, as such.

Actually issue is captured in the recording "2022-06-13 18-15-07.mp4" attached.

**Note:**

I have reported this issue, because we has already fixed issue-  [AC-2478](https://jira.corp.adobe.com/browse/AC-2478) and this fixed issue is quite similar to now what reported above.
We are doing regression testing by considering the above fixed issue [AC-2478](https://jira.corp.adobe.com/browse/AC-2478)

We had this reported issue already existing prior the fix of [AC-2478](https://jira.corp.adobe.com/browse/AC-2478) as well.

## Related Issue

Closes https://jira.corp.adobe.com/browse/AC-3522

## Checklist

### Resolved issues:
1. [x] resolves magento/pwa-studio#3921: AC-3522::Product search result with items count is not being read by …

### DIFF
--- a/packages/venia-ui/lib/components/SearchPage/searchPage.js
+++ b/packages/venia-ui/lib/components/SearchPage/searchPage.js
@@ -206,11 +206,15 @@ const SearchPage = props => {
         areaLiveVal = 'polite';
 
         if (productSortRoot) {
-            productSortRoot.setAttribute('aria-hidden', 'false');
+            setTimeout(function() {
+                productSortRoot.setAttribute('aria-hidden', 'false');
+            }, 15000);
         }
 
         if (filterSidebarRoot) {
-            filterSidebarRoot.setAttribute('aria-hidden', 'false');
+            setTimeout(function() {
+                filterSidebarRoot.setAttribute('aria-hidden', 'false');
+            }, 15000);
         }
     }
 
--- a/packages/venia-ui/lib/components/SearchPage/searchPage.js
+++ b/packages/venia-ui/lib/components/SearchPage/searchPage.js
@@ -173,11 +173,7 @@ const SearchPage = props => {
 
     const itemCountHeading =
         data && !loading ? (
-            <span
-                aria-live="polite"
-                aria-busy="true"
-                className={classes.totalPages}
-            >
+            <span aria-busy="true" className={classes.totalPages}>
                 {formatMessage(
                     {
                         id: 'searchPage.totalPages',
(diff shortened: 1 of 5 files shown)


anthoula:
@magento create issue from PR

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -194,6 +190,34 @@ const SearchPage = props => {
         .filter(Boolean)
         .join(' - ');
 
+    let areaLiveVal = 'off';
+    const productSortRoot = document.getElementById('ProductSort-root');
+    const filterSidebarRoot = document.getElementById('FilterSidebar-root');
+
+    if (productSortRoot) {
+        productSortRoot.setAttribute('aria-hidden', 'true');
```
jcalcaben, line 198: why is this set here and not directly in the `ProductSort` component?

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -194,6 +190,34 @@ const SearchPage = props => {
...
+    }
+
+    if (filterSidebarRoot) {
+        filterSidebarRoot.setAttribute('aria-hidden', 'true');
```
jcalcaben, line 202: why is this set here and not directly in the `FilterSidebar` component?

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -194,6 +190,34 @@ const SearchPage = props => {
...
+    }
+
+    if (productsCount > 0) {
+        areaLiveVal = 'polite';
+
+        if (productSortRoot) {
+            setTimeout(function() {
+                productSortRoot.setAttribute('aria-hidden', 'false');
+            }, 15000);
```
jcalcaben, line 211: this should be moved into the `ProductSort` component and done in a way such that a new timeout isn't scheduled everytime the component renders.

see: https://www.joshwcomeau.com/snippets/react-hooks/use-timeout/

packages/venia-ui/lib/components/SearchPage/searchPage.js
```diff
@@ -194,6 +190,34 @@ const SearchPage = props => {
...
+        }
+
+        if (filterSidebarRoot) {
+            setTimeout(function() {
+                filterSidebarRoot.setAttribute('aria-hidden', 'false');
+            }, 15000);
```
jcalcaben, line 217: this should be moved into the `FilterSideBar` component and done in a way such that a new timeout isn't scheduled everytime the component renders.

see: https://www.joshwcomeau.com/snippets/react-hooks/use-timeout/

jcalcaben (changes requested):

glo71317:
As @glo82145 has created different pr for the fix

PR #3969, opened 2022-09-30
PWA-2985: [bug]: Payment method always reverts to “Check / Money orde…

anthoula:
…r” when there is an error with the transaction

## Description

- Set selected payment method mutation upon onChange event
- Set initialSelectedMethod to selected_payment_method data
- Update test coverage & snapshots

## Related Issue

Closes #[PWA-2985](https://jira.corp.adobe.com/browse/PWA-2985).

## Checklist


fnhipster (approved):

dpatil-magento:
run cypress

dpatil-magento:
@anthoula Please check following observations - 
1. Cypress checkout tests failing because when there is only one payment method is available then it should be selected by default but currently even this case is waiting for user to select.

image

2. **Very minor**, horizontal separation line before selecting payment method looks bit odd. Also if its ok to have it, the position needs bit adjusted. Currently its very close to second payment method.

image

anthoula:
run cypress

dpatil-magento (approved):

PR #4039, opened 2023-02-20
Quantity value isn't automatically selected, forcing a user to have to manually clear the old value before entering a new one

### LINKED ISSUE #3767: [bug]: When on mobile, quantity value isn't automatically selected, forcing a user to have to manually clear the old value before entering a new one
**Describe the bug**

When on mobile and selecting the quantity input on the PDP, the current value isn't automatically selected, meaning a user has to use the back space key on their keyboard before entering the quantity that they would actually like.

Further, we're not able to use the up and down arrow keys on a keyboard to increase/decrease the value.

**To reproduce**

Steps to reproduce the behaviour:

1. Using a mobile device, visit https://venia.magento.com/penelope-peasant-blouse.html
2. Select the quantity input field
3. Depending on where you tap on the screen, the cursor will either be in front of the current value, or following the current value

If the cursor is placed in front of the current value, then you will have to try to tap again on the quantity input to place the cursor following the value, so that you can actually use the backspace key to clear the current value

**Expected behaviour**

Ideally when selecting the quantity input, users would be able to select their new value, rather than having to clear the value that is currently there.

**Screenshots**

Venia Quantity Input Issue

**Possible solutions**

Ideally we would use the number input. I'm aware there are accessibility issues with that element though.

That means we have to introduce JavaScript to automatically select the input's value on focus, so that typing anything into the soft keyboard (and even regular keyboard when on a desktop) will clear the current value.

We will also need to use JavaScript to implement the increase/decrease function through the up/down arrow keys on a keyboard.

-   Device [e.g. iPhone6, PC, Mac, Pixel3]: Mobile
-   Browser [e.g. Chrome, Safari]: All
-   Browser Version [e.g. 22]: All
-   Magento Version [e.g Adobe Commerce 2.4]: Any

-   [x] `venia-ui`

Sahilsks:
## Description

Quantity value isn't automatically selected, forcing a user to have to manually clear the old value before entering a new one.
Quantity value automatic selection while focusing on it.

## To reproduce

Steps to reproduce the behaviour:

Using a mobile device, visit https://venia.magento.com/penelope-peasant-blouse.html
Select the quantity input field
Depending on where you tap on the screen, the cursor will either be in front of the current value, or following the current value

## Related Issue

Closes #3767.

https://user-images.githubusercontent.com/86670457/220164765-7a36446f-7b13-4d6f-aed8-b4f257dece01.mov

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/__tests__/__snapshots__/productForm.spec.js.snap
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/__tests__/__snapshots__/productForm.spec.js.snap
@@ -262,6 +262,7 @@ https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/media/catalog/produc
                       name="quantity"
                       onBlur={[Function]}
                       onChange={[Function]}
+                      onFocus={[Function]}
                       pattern="[0-9]*"
                       value={5}
                     />
@@ -632,6 +633,7 @@ https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/media/catalog/produc
                       name="quantity"
                       onBlur={[Function]}
                       onChange={[Function]}
+                      onFocus={[Function]}
                       pattern="[0-9]*"
                       value={5}
                     />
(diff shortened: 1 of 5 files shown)


Sahilsks:
@magento run all tests

Sahilsks:
@magento run all tests

glo82145:
As ticket associated with this PR is closed, Hence closing this ticket.

PR #4054, opened 2023-03-08
Pwa 2780 remove increment decrement tab selection

### LINKED ISSUE #3766: [feature]: Remove increase/decrease buttons of quantity selector from the tab index
Currently when tabbing through a PDP, a user has to tab twice to get passed the quantity selector. If the value in the quantity selector is greater than 1, then the user will have to tab 3 times to get passed it:

First tab press: Decrease button
Second tab press: quantity input
Third tab press: Increase button
Forth Tab press: next element in the DOM

Further, the increase/decrease buttons don't have any focus state currently either, so a user isn't able to visually see that they are focused on those elements (which we don't have to fix, as we're going to remove those buttons from the tab index)

Keyboard users already have access to the keyboard, so that can enter the number for the quantity they like, without having to use the increase/decrease buttons. By removing the buttons from the tab order, we can make the usability of the app much better, without sacrificing any accessibility, as the input field is still focusable.
Unfortunately due to the current implementation of the quantity input itself, a user isn't able to use the up/down arrow keys to change the value, which I'll create another issue for.

Removing the buttons from the tab index would be done by simply adding `tabindex="-1"` to them.

-

**Additional context**

- [x] `venia-ui`

Sahilsks:
## Description

Remove increase/decrease buttons of quantity selector from the tab index

## Related Issue

Closes #3766

https://user-images.githubusercontent.com/86670457/223668422-cc0fd3c6-ae26-4234-91ef-57d9af0ec316.mov

## Checklist

### DIFF
--- a/packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/__tests__/__snapshots__/productForm.spec.js.snap
+++ b/packages/venia-ui/lib/components/CartPage/ProductListing/EditModal/__tests__/__snapshots__/productForm.spec.js.snap
@@ -215,6 +215,7 @@ https://master-7rqtwti-mfwmkrjfqvbjk.us-4.magentosite.cloud/media/catalog/produc
                   className="button_decrement"
                   disabled={false}
                   onClick={[Function]}
+                  tabindex="-1"
                   type="button"
                 >
                   <span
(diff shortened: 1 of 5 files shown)


haal9000:
run pr-deploy

haal9000:
run pr-deploy

glo82145 (approved):

glo82145:
As associated ticket with this PR is closed. hence closing this PR.

Issue #3950, opened 2022-09-01
[bug]: Payment method always reverts to “Check / Money order” when there is an error with the transaction


fdussaultAbsolunet:
**Describe the bug**

The payment method always reverts to “_Check / Money order_” when there is an error with the transaction in PWA Studio, even if _Credit Card_ mode was previously selected.

**To reproduce**

1. Activate multiple payment methods, like check money order and braintree
2. Get a product in your cart and proceed to checkout
3. Fill the Shipping Information and Shipping Method forms
4. Select "_Credit Card_"
5. Manually refresh the browser page (as a failing payment would do)

**Expected behavior**

When there is a transaction failure, the page is refreshing but the selected payment method should remain to Credit Card rather than selecting back the first available method in the list (since credit is also available).

**Screenshots**

![image](https://user-images.githubusercontent.com/112636245/187937114-7dbd48fb-4be7-4fca-abdd-b3482147f8a7.png)

**Please let us know what packages this bug is in regards to:**

-   [ ] `venia-concept`
-   [ ] `venia-ui`
-   [ ] `pwa-buildpack`
-   [x] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

anthoula:
@adobe export issue to JIRA project PWA as Bug

PR #4098, opened 2023-04-20
PWA-3029::[bug]: Selecting payment method causing graphql error

glo82145:
## Description

Describe the bug

In 12.7 [a fix was made](https://github.com/magento/pwa-studio/issues/3950) to visually select the method of payment that is saved as selected on the quote. However it also added an onChange function that executes a mutation to save the payment method.

https://github.com/magento/pwa-studio/pull/3969/files#diff-b3bc0e73e1c57b4b69ad1d1cc59b3d3e60021bd9ec46a399b73305079e2ebb11R54

This causes a graphql error for custom payment methods that require additional information. This can be seen by the need to pass in braintree data, which wouldn't be needed for non-braintree payment methods. Other payment methods require their own "braintree equivalent" data in order to be set

Example for our custom payment method for Moneris (code: mht), we now get an error when selecting the payment method radio button:
message: "Required parameter \"mht\" for \"payment_method\" is missing."

Data sent to graphql endpoint:
``{{
variables: {
cartId: "ZSNL4bWFnEaYmrtCcinryj76gCvoDcho"
paymentMethod: {
code: "mht",
braintree :

{ is_active_payment_token_enabler: false payment_method_nonce: "mht" }
}
}
}}`{{

Expected behavior

No graphql error when changing radios for payment method

Possible solutions

Don't call setPaymentMethodOnCart at a higher level than the payment method. Leave saving the payment method to the payment method (usually when going to review step, not when selecting the radio)

## Related Issue

Closes https://jira.corp.adobe.com/browse/PWA-3029

## Checklist

### DIFF
--- a/packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentMethods.js
+++ b/packages/peregrine/lib/talons/CheckoutPage/PaymentInformation/usePaymentMethods.js
@@ -45,17 +45,24 @@ export const usePaymentMethods = props => {
     const handlePaymentMethodSelection = useCallback(
         element => {
             const value = element.target.value;
-
-            setPaymentMethod({
-                variables: {
-                    cartId,
-                    paymentMethod: {
-                        code: value,
+            
+            const paymentMethodData = 
+                (value == 'braintree')
+                ? {    
+                    code: value,
                         braintree: {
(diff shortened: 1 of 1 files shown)


glo42707:
run cypress

PR #4116, opened 2023-05-11
PWA-3139::Region is required in PWA although this option is disabled …

glo80771:
…in Magento , and is not required on normal storefront

## Description

Region is required in PWA although this option is disabled in Magento , and is not required on normal storefront .

Set default country to United Kingdom , disable State is Required for UK .
Try to create order , in SHIPPING INFORMATION - state will be required ,

## Related Issue

Closes https://jira.corp.adobe.com/browse/PWA-3139

## Checklist


glo80771:
Another PR is raised for the ticket and hence closing this.

PR #4119, opened 2023-05-16
PWA-3139-V1::Region is required in PWA although this option is disabled …

glo80771:
…in Magento , and is not required on normal storefront

## Description

Region is required in PWA although this option is disabled in Magento , and is not required on normal storefront .

Set default country to United Kingdom , disable State is Required for UK .
Try to create order , in SHIPPING INFORMATION - state will be required ,

## Related Issue

Closes https://jira.corp.adobe.com/browse/PWA-3139.

## Checklist


glo82145 (dismissed):

packages/venia-ui/lib/components/CheckoutPage/ShippingInformation/AddressForm/customerForm.js
```diff
@@ -134,6 +134,18 @@ const CustomerForm = props => {
         <Text type="hidden" field="default_shipping" initialValue={true} />
     );
 
+    const createErrorMessage = JSON.stringify(
+        errors.get('createCustomerAddressMutation')
+    );
+    const updateErrorMessage = JSON.stringify(
+        errors.get('updateCustomerAddressMutation')
+    );
+    const errorMessage = 'region_id is required for the specified country code';
```
fooman, line 143: Testing for the specific error message from the admin seems very brittle (error messages could be translated and customised).

I would suggest something like this in region.js instead

```
const { validate, ...propsWithoutValidate } = regionProps;
```

and then 
```
    const regionField =
        regions.length || loading ? (
            <Select
                {...regionProps}
                field={fieldSelect}
                id={classes.root}
                items={regions}
            />
        ) : (
            <TextInput {...propsWithoutValidate} field={fieldInput} id={classes.root} optional="true" />
        );
```

glo80771:
The ticket is created with another PR and hence closing this PR.

PR #4140, opened 2023-06-27
Draft: Updated RangeFilter method

Tarun50745:
## Description

In the event that a different price filter is applied after a price filter is already applied, the query will only be for the first price filter, not the combination of the two.

This behavior might be interpreted as a bug but is actually a mis-implementation of the UI for the price filter which only allows a "range" ie {{

{ "from": "0", "to": "100" }
}}.

## Related Issue

Closes https://jira.corp.adobe.com/browse/PWA-3162

## Checklist


glo80771:
Implemented a solution to change the price filter design.

PR #4180, opened 2023-10-17
FIX#PWA-3180 Layered navigation Options are still visible even when n…

RaghavendraTirumalasetti:
…o results are available

## Description

When using layered navigation, attributes with the "Filterable (with results)" configuration are still visible on the PWA frontend even when no products are available for that search

## Related Issue

Closes #https://jira.corp.adobe.com/browse/PWA-3180.

## Checklist


packages/peregrine/lib/talons/RootComponents/Category/useCategoryContent.js
```diff
@@ -62,16 +63,71 @@ export const useCategoryContent = props => {
...
+        const fashionColor = [];
+        const fashionMaterial = [];
+        const fashionSize = [];
+        const fashionStyle = [];
+        const hasVideo = [];
+
+        //{from: "40" to: "59"}
+
+        if (filterOptions) {
+            for (const [group, items] of filterOptions) {
```
fooman, line 75: All these product attributes are custom and not present in all stores. Hard coding them this way is the wrong approach in my opinion.

glo82145 (approved):

PR #4218, opened 2023-12-14
PWA-3216 upgrading node to >=18

glo82145:
## Description

need to upgrade Node support >=18 for PWA studio package as per Node updates no more availbale for support for lower version as per link https://endoflife.date/nodejs

Prerequisite-
Node >=18 should be installed for PWA setup

Description:
as if now PWA lower support is by default >=14 as per informattion can see on https://github.com/magento/pwa-studio/blob/76fbf76fc1afce2b244d3f647164dbf60b7da070/package.json#L101

Actual Result:-
Current Node version support for PWA studio is >=14

Expected:
Default Node version need to be updated in PWA package as >=18

## Related Issue

Closes (https://jira.corp.adobe.com/browse/PWA-3216)

## Checklist


glo82145:
run pr-deploy

glo82145:
run pr-test

Issue #3716, opened 2022-02-23
[bug]: Order statuses are decoupled from the back-end


fooman:
**Describe the bug**

A refunded order will show a wrong status. Custom order statuses are not shown.

**To reproduce**

Steps to reproduce the behavior: Refund an order in the admin. Create a custom order status and assign it to an order.

1. navigate to /order-history
2. notice PWA-Studio's own take on order statuses

**Expected behavior**

The frontend should accurately show the status of the order.

**Possible solutions**

Delete the logic around [derivedStatus](https://github.com/magento/pwa-studio/blob/1fe1329b45bb584f138e351a77ddebef2ebfeb13/packages/venia-ui/lib/components/OrderHistoryPage/orderRow.js#L43) and display what the back-end provides?

**Debug Report**

Paste the output of `yarn build:report`, and any other context about the problem, here.

**Please complete the following device information:**

-   Device [e.g. iPhone6, PC, Mac, Pixel3]:
-   Browser [e.g. Chrome, Safari]:
-   Browser Version [e.g. 22]:
-   Magento Version [e.g Magento Commerce 2.4]:

**Please let us know what packages this bug is in regards to:**

-   [ ] `venia-concept`
-   [x] `venia-ui`
-   [ ] `pwa-buildpack`
-   [ ] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

fooman:
A few other things not quite right with this [implementation](https://github.com/magento/pwa-studio/blob/1fe1329b45bb584f138e351a77ddebef2ebfeb13/packages/venia-ui/lib/components/OrderHistoryPage/orderRow.js#L44). Checking for 'Complete' would not work if the language is not English and/or the label was changed.

It's also a stretch to translate Complete to Delivered. In a default Magento environment the moment an order has an invoice and a shipment its status will change to Complete. At that point in time it would not have left the merchant's warehouse.

anthoula:
@magento export issue to JIRA project PWA as Bug

fooman:
As further suggestion if the frontend should display data differently depending on how far an order has progressed relying on the order state https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Sales/Api/Data/OrderInterface.php#L29 would be the way to go. It is language independent and works in conjunction with custom order statuses.

RaghavendraTirumalasetti:
[...]

PR #4229, opened 2024-02-02
PWA-3178::Order statuses are decoupled from the back-end

glo82145:
## Description

A refunded order will show a wrong status. Custom order statuses are not shown.

To reproduce

Steps to reproduce the behavior: Refund an order in the admin. Create a custom order status and assign it to an order.

navigate to /order-history
notice PWA-Studio's own take on order statuses
Expected behavior

The frontend should accurately show the status of the order.

## Related Issue

Closes [#PWA-3178](https://jira.corp.adobe.com/browse/PWA-3178)

## Checklist


fooman:
In my opinion the solution should use the order state
https://github.com/magento/magento2-pwa/blob/main/SalesGraphQlAux/etc/schema.graphqls#L5
instead of doing any derived calculations.

Also note that "Delivered" will be inacurrate until such time that the delivery has arrived at the customer's address. The complete status will automatically get applied once all items in an order have been marked invoice and shipped. Delivered to most people would indicate they have the item in their physical posession.

fooman:
Just linking the original issue report here as well
https://github.com/magento/pwa-studio/issues/3716

glo11372 (approved):

PR #4355, opened 2024-11-21
Wrapping FormatMessage Component with <div> so that browser translate…

del15881:
…d DOM will not fail the js DOM Manipulation.

## Description
Fixing the Login and Create Account issues faced during DOM manipulation when the store front is translated to other languages using Browser Translations. 
There is also 1 more change that is needed which is at file path "packages/venia-ui/lib/components/MiniCart/miniCart.js".
This required change is already found in PR : https://github.com/magento/pwa-studio/pull/4318 .
Check Commit ID : bbd3fb3dfc494c07b173e0e0533a07ac5f94a22b 

## Related Issue

Closes #PWA-3378
https://jira.corp.adobe.com/browse/PWA-3378

## Checklist


glo82145 (approved):

glo42707:
closing since duplicate of 4318

PR #4357, opened 2024-12-05
Adding Conditions to Handle 404 redirect_code for Invalid Urls and Ad…

del15881:
…ding condition to handle Valid PWA URLs as well.

## Description
Adding conditions to venia-ui/upward.yml  veniaResponse object to handle 404 URLs as well as Valid PWA Responses.

## Related Issue

Closes # https://jira.corp.adobe.com/browse/PWA-3386

## Checklist

### DIFF
--- a/packages/venia-ui/upward.yml
+++ b/packages/venia-ui/upward.yml
@@ -23,6 +23,12 @@ veniaResponse:
         - matches:  urlResolver.redirect_code
           pattern: '(301|302)'
           use: dynamicRedirect
+        - matches: request.url.pathname
+          pattern: '^/(account-information|address-book|cart|checkout|communications|sign-in|contact-us|create-account|forgot-password|order-confirmation|order-history|customer/account/createPassword|saved-payments|search\.html|wishlist)$'
+          use: veniaAppShell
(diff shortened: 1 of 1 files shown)


glo42707 (approved):
Approved

brendanfalkowski:
@del15881 — Is there a reason for repeating all the valid PWA pathnames in `upward.yml` that align to application routes when the `default` also resolves to `veniaAppShell`? Isn't listing them the same thing as not listing them?

```
veniaResponse:
    resolver: conditional
    when:
...
...
...
              pattern: '^/(account-information|address-book|cart|checkout|communications|sign-in|contact-us|create-account|forgot-password|order-confirmation|order-history|customer/account/createPassword|saved-payments|search\.html|wishlist)$'
              use: veniaAppShell
...
...
...
default: veniaAppShell
```

We'd like to understand it how this is meant to be used. This appears to require somebody manually keeping them in sync, which will be forgotten easily.

For custom implementations of PWA Studio because we already define our `routes.js` in the application code, and it feels like we shouldn't have to repeat them in Upward config. Disabling this part of the Upward config has no impact on the application's use (it's still using the same app shell).

PR #4363, opened 2024-12-12
Fixed handle trigger click function error

del22123:
## Description

This PR contains the fix for handleTriggerClick function call error in customer login flow when customer signs in from shipping page

Closes #PWA-3381
https://jira.corp.adobe.com/browse/PWA-3381

## Verification Steps

Step 1: In PWA store view add a product to cart and then go to Checkout page -> Sign In 
Step 2: Sign in with an existing customer account

Current result:
Error is thrown 'handletriggerclick() is not a function'

Fix result:
Login will work and user will be logged in at checkout page

## Checklist


glo82145 (approved):

glo42707 (approved):
approved

PR #4469, opened 2025-06-20
PWA-4468: use the same version of `react-tabs` everywhere

### LINKED ISSUE #4468: [bug]: fix Yarn invariant violation caused by "packages/pagebuilder" requesting different version of "react-tabs"
**Describe the bug**

When running `yarn install`, Yarn throws an invariant violation and fails to install.

I traced this down to `packages/pagebuilder/package.json` which defines:

```
"devDependencies": {
  ...
  "react-tabs": "~3.0.0"
},
"peerDependencies": {
  ...
  "react-tabs": "~3.1.0"
}
```

This breaks Yarn workspaces because it can't resolve to a single version.

**Expected behavior**

Yarn can install.

**Possible solutions**

I will submit a PR.

There isn't a breaking change between `react-tabs` 3.0.0 and 3.1.0. It only added LTR locale support.

The Pagebuilder package should define both dependencies with 3.1.0, so Yarn works.

-   [ ] `venia-concept`
-   [ ] `venia-ui`
-   [x] `pagebuilder`
-   [ ] `pwa-buildpack`
-   [ ] `peregrine`
-   [ ] `pwa-devdocs`
-   [ ] `upward-js`
-   [ ] `upward-spec`
-   [ ] `create-pwa`

brendanfalkowski:
## Description

When running yarn install, Yarn throws an invariant violation and fails to install.

I traced this down to packages/pagebuilder/package.json which defines:

```
"devDependencies": {
  ...
  "react-tabs": "~3.0.0"
},
"peerDependencies": {
  ...
  "react-tabs": "~3.1.0"
}
```

This breaks Yarn workspaces because it can't resolve to a single version.

Additional context: this issue was uncovered while testing with Node Corepack enabled. This is a recent-ish mode in Node which uses a Node-bundled version of Yarn instead of globally installed Yarn. Our team ran into multiple issues converting to that, and chose to continue using globally installed Yarn. But it did help highlight this dev/peer dependency difference with the `react-tabs` dependency. It seemed like an easy fix to make all the monorepo packages ask for `3.1.0`.

## Related Issue

Closes #4468

## Checklist

### DIFF
--- a/packages/pagebuilder/package.json
+++ b/packages/pagebuilder/package.json
@@ -44,7 +44,7 @@
     "react": "~17.0.1",
     "react-dom": "~17.0.1",
     "react-slick": "~0.28.0",
-    "react-tabs": "~3.0.0",
+    "react-tabs": "~3.1.0",
     "react-test-renderer": "~17.0.1"
   },
   "peerDependencies": {


brendanfalkowski:
Note: I did not commit the lockfile, because there were too many other changes.

I'll let the Adobe team decide how to incorporate that.

